PR #3750, opened 2018-06-12
Added Quick Add Bank Tag

bolinkd:


Nightfirecat:
I'd personally like if this did not persist between bank sessions--that is, if you closed the bank, I'd like for it not to add the extra menu item. Bank menus are already long enough as-is. That said, that's a neat feature that I'd find very helpful when setting new tags. :+1:

bolinkd:
Will work on this later today!

bolinkd:
currently it does, but it will not re add it to the list,
I am including this as one of the cases to now show the menu for the next commit

bolinkd:
@Nightfirecat PR is updated with requested features as well as an updated gif 👍

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -196,6 +237,54 @@ else if (!tagSearch)
 		}
 	}
 
+	@Subscribe
+	public void onMenuOpened(MenuOpened event)
```
Adam-: I would stick with modifying the string on the script stack probably instead of this.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -134,6 +174,7 @@ public void onScriptEvent(ScriptCallbackEvent event)
 			{
 				// set menu action index so the edit tags option will not be overridden
 				intStack[intStackSize - 3] = EDIT_TAGS_MENU_INDEX;
+				intStack[intStackSize - 4] = ADD_PREVIOUS_TAG_MENU_INDEX;
```
Adam-: Why is there not another widget config flag being set for this?

Pashweetie:
I'm excited if this ever makes it to release.

bolinkd:
Ive been wanting to get the requested changes done but have been busy with other things the past couple weeks. Hoping to have some time this weekend to do it.

PR #3761, opened 2018-06-12
 chat commands: alow clearing words and chat with ctrl w/backspace 

Adam-:


deathbeam:
I would rather use Ctrl + U for clearing the line, its pretty much readline standard, works also in Vim, Emacs.

Adam-:
I would rather keep with the behavior of existing clients.

Dgc2002:
Just to throw in my 2 cents:  
  
The `ctrl+backspace` behavior is pretty counter intuitive. Most any text input that I regularly encounter will kill the previous word rather than an entire line when hitting `ctrl+backspace`.  
  
`ctrl` usually means 'apply until the next word boundary', e.g. `ctrl+right` jumps over the next word instead of the next character, `ctrl+right+shift` selects the next word instead of the next character, etc.

Issue #1728, opened 2018-04-20
Add Staff of the Dead (and Toxic Staff of the Dead) timer overlay


Nightfirecat:
http://oldschoolrunescape.wikia.com/wiki/Staff_of_the_dead#Special_attack

> The staff has a special attack, Power of Death, that halves all Melee damage of your opponent(s) (all opponents that hit you in the time of the special attack) for one minute at a cost of 100% special attack energy. This effect stacks with the PvP damage reduction effect of the Protect from Melee prayer, but is lost if the staff is unequipped.

There are two chat lines that appear during its effect: "Spirits of deceased evildoers offer you their protection." and "Your protection fades away." The latter seems to appear after the minute is up, even if you have switched weapons during that time.

PR #3762, opened 2018-06-12
Add Staff of the Dead timer

Nightfirecat:
Closes runelite/runelite#1728

P.S. I only own a normal Staff of the Dead, so could not test the charged/uncharged toxic SotD, though I assume the graphics ID is the same.
P.P.S. The wiki says the special's effect wears off on any weapon switch, so that's how I implemented it. I don't know for sure if a weapon switch between two SotDs would cause the effect to wear off, but it's an unlikely edge case--I doubt anyone is actually using two different ones at any given time.


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -110,6 +112,11 @@ protected void shutDown() throws Exception
 	@Subscribe
 	public void onVarbitChange(VarbitChanged event)
 	{
+		int equippedWeap = client.getVar(Varbits.EQUIPPED_WEAPON_TYPE);
+		if (lastEquippedWeapVarb != equippedWeap)
+		{
+			removeGameTimer(STAFF_OF_THE_DEAD);
+		}
```
deathbeam: add empty line below

Nightfirecat: Added

Issue #3708, opened 2018-06-09
[Master Clue] Hot/Cold @ East of Watson's house too far north


Jalopyy:
https://i.imgur.com/Of1ej6A.jpg

Dig site should be where the marked tile is. Current spot is a touch too north.

Nightfirecat:
So that's 4 tiles too far north?

Jalopyy:
Looks like it, yea.

deathbeam:
Fixed in #3780

Nightfirecat:
Damn, I thought mentioning `fixes so-and-so-issue` from the commits would've been enough. I'll add that to the description next time.

Issue #3713, opened 2018-06-09
[Master Clue] Hot/Cold @ West Sulfur Mine Entrance too far south


Jalopyy:
https://i.imgur.com/1UdrTT3.jpg

Move highlighted square north by 2 tiles.

PR #3780, opened 2018-06-13
Fix zeah hot cold dig locations

Nightfirecat:


Jalopyy:
This would close #3708 and #3713.

Nightfirecat:
Each of the commits references those issues

PR #3788, opened 2018-06-14
Use Object.equals for Overlays

deathbeam:
- Generated equals from Lombok can cause falsy equals checks and so match
with unrelated overlay.
- Do not add overlay twice to overlay list based on equals

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayManager.java
@@ -109,6 +109,11 @@ public class OverlayManager
 	 */
 	public synchronized boolean add(final Overlay overlay)
 	{
+		if (overlays.contains(overlay))
+		{
+			return false;
+		}
+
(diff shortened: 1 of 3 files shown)


Jasper-ketelaar:
Tested it and can confirm it fixes the issue.

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayManager.java
```diff
@@ -109,6 +109,11 @@ private OverlayManager(final ConfigManager configManager)
 	 */
 	public synchronized boolean add(final Overlay overlay)
 	{
+		if (overlays.contains(overlay))
+		{
+			return false;
+		}
+
 		final boolean add = overlays.add(overlay);
```
Adam-: This must always be true?

deathbeam: Its ArrayList, so yeah.

PR #3796, opened 2018-06-14
Ground item changes (as requested by people in Discord)

deathbeam:
- When highlight value is set to 0, disable highlight for that particular
configuration.
- Make hidden items color configurable


SomeZer0:
>When highlight value is set to 0, disable highlight for that particular configuration.

To clarify, that would mean that it uses the standard menu color for that item?

deathbeam:
It disables the highlighting, so it uses default color in right-click menu and default items color that is configurable in overlay

Issue #3592, opened 2018-06-03
Player Indicators Configuration


Vextrov:
The minimap gets quiet unclear and covered up with white texts when showing all player names in map. However I like being able to see all player names above their head without having to point my mouse. This helps me identify the people surrounding me while I'm skilling, playing events, and doing other stuff.
Therefore I'd like to suggest the ability to disable "Draw names on map" individually for Non-clan members only.

![player indicator](https://user-images.githubusercontent.com/39874941/40885017-2ab47e80-6762-11e8-8ef8-526746a887e8.png)

DanielOwen1:
![image](https://user-images.githubusercontent.com/18666558/40890991-5ec54eac-6776-11e8-8b58-c7186dfa339e.png)

I've just done this, if this is correct?

deathbeam:
This request is pretty specific. I dont think having configurable option just for "non-clan member" for minimap names makes much sense.

DanielOwen1:
Fair enough

Vextrov:
@DanielOwen1  Yes, this is the exact result I was looking for. But I also agree with @deathbeam. Perhaps you can add a "show on map" for each category? The Image below is just an example:

![mn](https://user-images.githubusercontent.com/39874941/40902396-380ac5f2-6817-11e8-9aee-c117fe9c8eb3.jpg)

Also, will I need to install an addon or something for this? or I have to wait until next update? 
Thanks :)

DanielOwen1:
I could try doing it at some point. Depends on if @deathbeam thinks there is any use for it

Nightfirecat:
I'm interested in implementing this, what's the consensus on whether this is helpful or necessary?

Pashweetie:
Personally I feel as if most users would find this clutter-some, so even if you plan on making it, please make it off by default.

Kamielvf:
Could you not make a dropdown for e.g.
`Team members` with the states `Disabled, Highlight, Minimap, Both`

This way it would be only 1 entry per player type + one color paired to it.

PR #3804, opened 2018-06-14
Expand player indicators plugin configurations

Nightfirecat:
A drop-down menu is introduced for each category of players, allowing
the choice of highlighting players, having them appear on the minimap,
or both.

The "draw names on minimap" option has also been removed.

updated-player-indicators-configuration

Closes runelite/runelite#3592


deathbeam:
You can probably just remove the "Draw names on minimap" option here.

Nightfirecat:
@deathbeam Consider it gone. :hocho:

Nightfirecat:
Should I add a config migration method to this PR to make it ready for merging? That way, it doesn't impact users who have changed default configs.

Nightfirecat:
Well now that there's also a "Tile" option to consider, the way this is written won't work at all. If I get back around to this, I'll likely rewrite this from scratch instead of picking up from how this is now.

PR #3811, opened 2018-06-15
Optional support for recoloring right-click menu

deathbeam:
Add back support for recoloring right-click menu in Ground Items, but
make it optional and disabled by default.


deathbeam:
This also fixes bug with right click menu not being recolored when item is hidden.

PR #3814, opened 2018-06-15
XP drops

trist456:
I've never used github so i'm sorry if this isn't the correct place to make a suggestion but I was wondering if it would be possible for 2012 XP drops to be added?

example: https://i.imgur.com/VvL9yaD.png
image


deathbeam:
Open issue here: https://github.com/runelite/runelite/issues/new (this is wrong place)

PR #3817, opened 2018-06-15
Add world map right click for toggling free roam camera

deathbeam:
Add option for toggling free roam camera (basically oculus orb camera)
to Camera plugin.

Preview:
peek 2018-06-15 16-45


deathbeam:
When we added this to plugin hub here: https://github.com/Adam-/runelite-plugins/tree/detached-camera wouldnt it be fine to just merge this to main client instead?

deathbeam:
As this is way more elegant solution

Nightfirecat:
I'm not sure why merging this instead would be preferred, given that external plugin was added already. This could be safely merged and updated there imo

Adam-:
The reason I left it as external is it pretty unpolished atm.

PR #3820, opened 2018-06-15
Use correct widget when hiding resizable minimap

### LINKED ISSUE #976: Minimap only hides when side panels are in "side-stones arrangement"
http://prntscr.com/irmlch
This has to be toggle off for the minimap to hide.

deathbeam:
When hiding minimap, there are 2 different widget IDs for respective
resizable mode configurations. Find proper one and hide it.

Fixes #976

peek 2018-07-20 22-32


jplsek:
I seem to be having issues with this. The minimap doesn't "unhide" after trying to enable the minimap from the options:
ezgif-5-241c73c4e1

The logout button also gets hidden when using side stones.

deathbeam:
Cant do much about logout button as it is part of minimap I guess. But I can look at the unhiding thing, idk why that is happening

deathbeam:
Fixed the bug with unhiding, should be good now

deathbeam:
Updated now it hides everything except logout button

PR #3827, opened 2018-06-15
Disable OSB prices by default

deathbeam:
Depending on 3rd party API is bad enough, having it enabled by default
is worse.


Adam-:
Is it? I thought it was a good idea to have it on by default. I guess I am indifferent.

deathbeam:
Well, its something we cant really control, so I would rather let users to enable it explicitly if they really want it, same as with CML plugin.

PR #3841, opened 2018-06-16
Render abyss rift portals as soon as visible

TheHolyWaffle:
This increases usability of running the abyss. Because if you can see the portal sooner, you know which direction/entry obstacle is closest to it.


Nightfirecat:
Do you have some screenshots that show the difference?

TheHolyWaffle:
before

after

TheHolyWaffle:
So, can this be merged?

PR #3842, opened 2018-06-16
Update cache-sourced ID files to 171-3

Abextm:
- ObjectID hit the 32k field limit, so its now separated into ObjectID and NullObjectID
 - All of the ID printing code has been moved into `IDClass`


runelite-client/src/main/java/net/runelite/client/plugins/poh/PohIcons.java
```diff
@@ -31,18 +31,24 @@
 import javax.imageio.ImageIO;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import static net.runelite.api.NullObjectID.NULL_13625;
```
deathbeam: You should probably remove either static import or wildcard import here

PR #3845, opened 2018-06-16
shows raid layout in chat & option to hide it

SohaibHN:
Shows the full raid rotation/scout in the chat box along with a plugin option to hide the result. 
Prints the result in the clan chat

image


SieBrum:
Can we change the color of the chat message?

SohaibHN:
I didn't include an option for it, just left it as highlighted in the print chat option and it defaulted to purple when I set it as a clan chat message. 

I left it instead of changing to black as all the other raid messages are purple by default.

deathbeam:
Implemented in #7545

Issue #2384, opened 2018-05-06
Remove teleblock info box when entering a safe zone (non-pvp zone)


Nightfirecat:
Currently it remains until the full duration is up.

TylerNichols:
I personally would want to still see it.

Kamielvf:
I think it makes more sense to show the box as long as the TB is affecting you.

Nightfirecat:
TB wears off as soon as you enter a safe zone.

Nightfirecat:
http://oldschoolrunescape.wikia.com/wiki/Tele_Block

> The effects of the teleblock are lifted when the affected player leaves the Wilderness by any means (jumping over the wilderness ditch, dying, etc.), or when the teleblock timer expires.

Kamielvf:
Yeah thats what I meant with having it only show the box as long as the TB is affecting you, I didn't mean to disagree with you :+1:

TylerNichols:
Oh... I didn't realize that's how tb works?? :O

PR #3847, opened 2018-06-16
Remove teleblock timer in safe zones

Nightfirecat:
Fixes runelite/runelite#2384


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -123,6 +175,35 @@ public void onVarbitChange(VarbitChanged event)
...
+				removeTbTimers();
+			}
+			lastWildernessVarb = inWilderness;
+		}
+	}
+
+	@Subscribe
+	public void onWidgetHiddenChanged(WidgetHiddenChanged event)
+	{
+		if (worldType == PVP)
```
Nightfirecat: This check is present because entering the wilderness on a normal world will trigger this condition as most of the pvp-related widgets load un-hidden and are thereafter hidden as needed. (that is, the safe zone widget is loaded un-hidden then immediately hidden before the frame renders)

With that in mind, world hopping between PVP worlds may need to be checked for as this same condition may be met there.

Nightfirecat: After testing, my suspicion here was correct--hopping between PVP worlds in a non-safe zone with this current code will register a safe zone trigger because of this phenomenon. Unfortunately, the `GameState` is `LOGGED_IN` by the time it happens, so this can't be gated behind a check for that. Does anyone have an idea of how to tackle this issue? To my knowledge, I can't add an async check for this, otherwise I'd just add some kind of short delay (say, 100ms or less) to check the widget state at that point.

runelite-client/src/main/java/net/runelite/client/plugins/timers/PvpWorldType.java
```diff
@@ -0,0 +1,51 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.timers;
+
+import net.runelite.api.WorldType;
+
+enum PvpWorldType
```
deathbeam: I would somehow reduce the need to basically have this structure what is really similar to one in XP tracker.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -505,18 +553,64 @@ public void onGameTick(GameTick event)
...
+		{
+			gameStateLoggedIn = false;
+
+			PvpWorldType worldType = worldSetToType(client.getWorldType());
+			if (lastWorldType != worldType)
+			{
+				if (worldType == DMM || lastWorldType == DMM
+					|| worldType == SDMM || lastWorldType == SDMM)
+				{
+					log.debug("Hopped to or from a DMM world, clearing Teleblock timer.");
```
deathbeam: This also probably needs to account for TOURNAMENT worlds

runelite-client/src/main/java/net/runelite/client/plugins/timers/PvpWorldType.java
Nightfirecat: Mentioned in discord, but I can't think of a good way to do this without completely rewriting WorldType, which seems like overengineering it.

deathbeam:
```java
static list <WorldType> DIFF = { DMM, SDMM, TOURNAMENT } 
static list <WorldType> PVP = { PVP }.join(DIFF)
boolean WorldType#isDifferentAccountWorld(WorldType other)
{
return 
  // hopping to different acc world
  (DIFF.contains(other) && !DIFF.contains(this)) || 
  // hopping from different acc world
  (!DIFF.contains(other) && DIFF.contains(this))
}

boolean WorldType#isPvp()
{
  return PVP.contains(this)
}
```

something like this probably should be enough

Nightfirecat: It was a little trickier than that (hopping between `DIFF` types also required clearing tbs), but I went with that approach.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: Tournament worlds are now included in the "different account type" worlds.

Nightfirecat:
Updated to remove the world hop checks as they're effectively handled upon the logout check introduced in #5090. The only two conditions the updated code now checks for are:

1. Exiting wilderness on a non-PVP world
2. Entering safe zone on a PVP world

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -720,4 +779,34 @@ private void removeGameTimer(GameTimer timer)
...
+	{
+		removeGameTimer(FULLTB);
+		removeGameTimer(HALFTB);
+		removeGameTimer(DMM_FULLTB);
+		removeGameTimer(DMM_HALFTB);
+	}
+
+	private static boolean isPvpWorld(final EnumSet<WorldType> worldTypes)
+	{
+		return listContainsAnyFromSet(PVP_WORLD_TYPES, worldTypes);
```
deathbeam: return worldTypes.stream().anyMatch(w -> PVP_WORLD_TYPES.contains(w));

Nightfirecat: Huh, didn't know you could stream sets. Will do.

deathbeam:
Needs rebase

Nightfirecat:
Rebased

Nightfirecat:
Derp, fixed that mistake.

Nightfirecat:
Added better doc comment to `IN_WILDERNESS` varbit, fixed wilderness widget IDs, and removed some spammy false positives on the "entered PVP safe zone" condition. This should be completed and ready to go.

runelite-api/src/main/java/net/runelite/api/WorldType.java
```diff
@@ -75,6 +76,13 @@
 		this.mask = mask;
 	}
 
+	public static EnumSet<WorldType> PVP_WORLD_TYPES = EnumSet.of(
```
deathbeam: Maybe this should be private? or maybe each world could have flag if its PVP and you would build this enumset based on that

Nightfirecat: Wasn't sure about exactly how to go about this--just knew I wanted to move it into WorldType since I'll actually need this for another PR. I'd be fine with making it private or making it an enum flag, though I slightly prefer just making it private.

deathbeam: Alright I guess we can go with that

Nightfirecat: Done

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -506,18 +541,35 @@ public void onGameTick(GameTick event)
...
 		{
-			lastPoint = currentWorldPoint;
 			return;
 		}
 
-		// assume movement means unfrozen
-		if (!currentWorldPoint.equals(lastPoint))
+		widgetHiddenChangedOnPvpWorld = false;
+
+		Widget widget = client.getWidget(PVP_WORLD_SAFE_ZONE.getGroupId(), PVP_WORLD_SAFE_ZONE.getChildId());
```
deathbeam: client.getWidget(PVP_WORLD_SAFE_ZONE) ?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -527,10 +579,12 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
 		if (gameStateChanged.getGameState() == GameState.HOPPING || gameStateChanged.getGameState() == GameState.LOGIN_SCREEN)
 		{
-			removeGameTimer(HALFTB);
-			removeGameTimer(DMM_HALFTB);
-			removeGameTimer(FULLTB);
-			removeGameTimer(DMM_FULLTB);
+			removeTbTimers();
+		}
+
+		if (gameStateChanged.getGameState() == GameState.LOGGED_IN)
```
deathbeam: else if (maybe you should change this entire method to do switch (gameStateChanged.getGameState()) instead, it will be cleaner

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -506,18 +541,35 @@ public void onGameTick(GameTick event)
...
-			lastPoint = currentWorldPoint;
 			return;
 		}
 
-		// assume movement means unfrozen
-		if (!currentWorldPoint.equals(lastPoint))
+		widgetHiddenChangedOnPvpWorld = false;
+
+		Widget widget = client.getWidget(PVP_WORLD_SAFE_ZONE.getGroupId(), PVP_WORLD_SAFE_ZONE.getChildId());
+		if (widget == null)
```
deathbeam: just merge this check with the next one probably

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -146,6 +158,32 @@ public void onVarbitChange(VarbitChanged event)
...
+		}
+	}
+
+	@Subscribe
+	public void onWidgetHiddenChanged(WidgetHiddenChanged event)
+	{
+		Widget widget = event.getWidget();
+		if (WorldType.isPvpWorld(client.getWorldType())
+			&& WidgetInfo.TO_GROUP(widget.getId()) == WidgetInfo.PVP_CONTAINER.getGroupId()
+			&& WidgetInfo.TO_CHILD(widget.getId()) >= WidgetInfo.PVP_CONTAINER.getChildId())
```
deathbeam: is this >= really necessary? seems wonky

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -146,6 +158,32 @@ public void onVarbitChange(VarbitChanged event)
...
+			&& client.getGameState() == GameState.LOGGED_IN
+			&& !loggedInRace)
+		{
+			if (!WorldType.isPvpWorld(client.getWorldType())
+				&& inWilderness == 0)
+			{
+				log.debug("Left wilderness in non-PVP world, clearing Teleblock timer.");
+				removeTbTimers();
+			}
+			lastWildernessVarb = inWilderness;
```
deathbeam: Add empty line above

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -146,6 +158,32 @@ public void onVarbitChange(VarbitChanged event)
 			removeGameTimer(PRAYER_ENHANCE);
 			lastRaidVarb = raidVarb;
 		}
+
+		int inWilderness = client.getVar(Varbits.IN_WILDERNESS);
+		if (lastWildernessVarb != inWilderness
```
deathbeam: add empty line above

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: I guess not. Checking the group is probably sufficient.

Nightfirecat:
@deathbeam Thanks for the double-check after the recent changes, addressed review comments.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -506,31 +542,44 @@ public void onGameTick(GameTick event)
...
-		if (gameStateChanged.getGameState() == GameState.HOPPING || gameStateChanged.getGameState() == GameState.LOGIN_SCREEN)
+		switch (gameStateChanged.getGameState())
 		{
-			removeGameTimer(HALFTB);
-			removeGameTimer(DMM_HALFTB);
-			removeGameTimer(FULLTB);
-			removeGameTimer(DMM_FULLTB);
+			case HOPPING:
+			case LOGIN_SCREEN:
+				removeTbTimers();
```
deathbeam: break; after

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -506,31 +542,44 @@ public void onGameTick(GameTick event)
...
+			case LOGGED_IN:
+				loggedInRace = true;
```
deathbeam: same here

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: Done

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -506,31 +542,46 @@ public void onGameTick(GameTick event)
...
-			removeGameTimer(HALFTB);
-			removeGameTimer(DMM_HALFTB);
-			removeGameTimer(FULLTB);
-			removeGameTimer(DMM_FULLTB);
+			case HOPPING:
+			case LOGIN_SCREEN:
+				removeTbTimers();
+				break;
+			case LOGGED_IN:
+				loggedInRace = true;
```
deathbeam: When do you actually set this variable to false?

Nightfirecat: I think this got lost during some rebasing or rework... This was meant to be set to false on game tick, so I've re-added that.

PR #3848, opened 2018-06-16
Adding font support for Lato-Heavy

msb217:
I have a feeling that this might be a longshot for instant approval, however, I am adding support for a cleaner, open-source font called Lato-Heavy. I will continue work on this branch in order to provide support for both the original RuneLite fonts and Lato-Heavy.


deathbeam:
I dont think straight replacing the runescape fonts with something else will ever be added. You would have better chance with just adding support for new fonts. But also, it is design decision to use RuneScape fonts everywhere to stay with the look of the game, using other font will just look way off.

PR #3875, opened 2018-06-18
ScreenshotPlugin: Capture theatre of blood rewards

MagicfTail:


MagicfTail:
Updated to use the widgetLoaded event, resulting in much more consistent screenshots. Worked 5/5 times at ToB and Barrows

MagicfTail:
5 more ToB completions and still 100% success. Just need testing at Raids 1

PR #3885, opened 2018-06-18
Fix sotd attack style hiding and permanently hidden defensive casting option after plugin shutdown

### LINKED ISSUE #3865: Staff of the Dead Defensive Cast
Unsure if there are any others but the Staff of the Dead and Toxic Staff of the Dead do not hide their defensive casting option when you have defensive options set to hide. I know Master Wand, Ancient Staff, and Ahrim's Staff all hide it properly.

Nightfirecat:
Fixes #3865

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -319,7 +319,8 @@ public class AttackStylesPlugin extends Plugin
 			// Magic staves defensive casting mode
 			// Note: Staff of  the Dead is weapon type 21, rather than the expected type 18
 			if (equippedWeaponType == WeaponType.TYPE_18
-				|| equippedWeaponType == WeaponType.TYPE_21)
+				|| equippedWeaponType == WeaponType.TYPE_21
+				|| !enabled)
 			{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -317,7 +317,10 @@ private void hideWarnedStyles(boolean enabled)
 			}
 
 			// Magic staves defensive casting mode
-			if (equippedWeaponType == WeaponType.TYPE_18)
+			// Note: Staff of  the Dead is weapon type 21, rather than the expected type 18
+			if (equippedWeaponType == WeaponType.TYPE_18
+				|| equippedWeaponType == WeaponType.TYPE_21
+				|| !enabled)
```
deathbeam: Why was this !enabled line added?

Nightfirecat: It's part of the second commit, which is to fix the issue of the defensive casting style being permanently hidden if you have a non-staff equipped and disable the plugin or disable style hiding.

Nightfirecat:
Adam has a cleaner way to fix this, so I'll hold off on this fix and wait for his fix or work off of it in a few hours.

PR #3892, opened 2018-06-18
Replace "Send notifications when focused" global setting with plugin specific settings

HSJ-OSRS:
its nice to have the ability to have some of these plugins notify you even when the client is focused while having some of them not notify you while focused


deathbeam:
The whole purpose of notifier class was to group all these notification related settings into 1 class that will be simply used for sending notifications. I disagree with reverting part of this change, imo user should just use the plugin configuration and adjust it based on what is he doing currently.

PR #3901, opened 2018-06-19
Replace Hook with MethodHook where possible

deathbeam:
In order to cleanup the Hooks class a bit, replace runescape-client
Hook with MethodHook in mixins where possible/convenient.

Depends on !13
Depends on !14

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
+++ b/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
@@ -41,7 +41,6 @@ import java.awt.image.VolatileImage;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GraphicsObject;
-import net.runelite.api.ItemComposition;
 import net.runelite.api.MainBufferProvider;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MessageNode;
@@ -55,7 +54,6 @@ import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.GraphicsObjectCreated;
 import net.runelite.api.events.MenuOptionClicked;
-import net.runelite.api.events.PostItemComposition;
 import net.runelite.api.events.ProjectileMoved;
 import net.runelite.api.events.SetMessage;
 import net.runelite.api.widgets.Widget;
@@ -427,13 +425,6 @@ public class Hooks
 		// but having the game tick event after all packets
 		// have been processed is typically more useful.
 		shouldProcessGameTick = true;
-	}
-
-	public static void postItemComposition(ItemComposition itemComposition)
-	{
-		PostItemComposition event = new PostItemComposition();
-		event.setItemComposition(itemComposition);
-		eventBus.post(event);
 	}
 
 	public static void onGraphicsObjectCreated(GraphicsObject go, int var1, int var2, int var3, int var4, int var5, int var6, int var7)
--- a/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
+++ b/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
@@ -240,14 +240,6 @@ public class Hooks
 		keyManager.processKeyTyped(keyEvent);
 	}
 
-	public static void focusGained(KeyFocusListener l, FocusEvent focusEvent)
-	{
-		FocusChanged focusChanged = new FocusChanged();
-		focusChanged.setFocused(true);
-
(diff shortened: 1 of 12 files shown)


Adam-:
This should probably have one commit per method moved. I suspect there will be things which no longer have to be exposed in rl-api as a result of this. Also the entire `@Hook` annotation can go away.

deathbeam:
Alright. So that means for the performance-critical methods that cannot be just sent through eventBus i should just call methods in Hooks directly from mixins? In similar way as mouse/key listeners do atm?

deathbeam:
Split each hook change to single commit, updated !13 with removal of `@Hook` annotation and used constructor injection for remaining hooks, so I call this update feature complete :d

runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -467,31 +464,6 @@ public static void menuOpened(Client client, int var1, int var2)
 		eventBus.post(event);
 	}
 
-	/**
```
Adam-: Move the javadoc too.

runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -1054,4 +1055,18 @@ public static void updateNpcs(boolean var0, PacketBuffer var1)
 		// have been processed is typically more useful.
 		shouldProcessGameTick = true;
 	}
+
+	@Inject
+	@FieldHook("focused")
```
Adam-: Any reason why this isnt a methodhook like before?

runelite-mixins/src/main/java/net/runelite/mixins/RSItemCompositionMixin.java
```diff
@@ -73,4 +76,13 @@ public void resetShiftClickActionIndex()
 	{
 		shiftClickActionIndex = DEFAULT_CUSTOM_SHIFT_CLICK_INDEX;
 	}
+
+	@Inject
+	@MethodHook(value = "post", end = true)
+	public void post()
+	{
+		final  PostItemComposition event = new PostItemComposition();
```
Adam-: Fix formatting

runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -383,27 +380,6 @@ public static void addChatMessage(int type, String name, String message, String
 		eventBus.post(chatMessage);
 	}
 
-	/**
```
Adam-: Do not remove javadoc.

runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -367,19 +365,6 @@ public static boolean menuActionHook(int actionParam, int widgetId, int menuActi
 		return menuOptionClicked.isConsumed();
 	}
 
-	public static void addChatMessage(int type, String name, String message, String sender)
-	{
-		if (log.isDebugEnabled())
```
Adam-: Any reason this is removed? The Hooks logger is accessible. We could also maybe just inject a logger.

runelite-mixins/src/main/java/net/runelite/mixins/RSMainBufferProviderMixin.java
```diff
@@ -42,6 +42,6 @@
 	@Replace("draw")
 	final void draw(Graphics graphics, int x, int y)
 	{
-		Hooks.draw(this, graphics, x, y);
+		Hooks.draw(this, graphics);
```
Adam-: What is the point of this commit?

runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
deathbeam: No reason for it to be methodHook when that field is properly modified in both focusLost and focusGained, this felt like simpler approach.

runelite-mixins/src/main/java/net/runelite/mixins/RSItemCompositionMixin.java
deathbeam: For the annotation?

runelite-mixins/src/main/java/net/runelite/mixins/RSMainBufferProviderMixin.java
deathbeam: Consistency with other draw-related methods, it does not need x/y so only needed fields should be exposed

PR #3905, opened 2018-06-19
Feature/idle overlay

NathenSample:
Exposes a configuration option to enable a overlay utilising the logout door sprite, with a timer representing MM:SS until logout due to idle.

Also has an option for `seconds until logout` representing the time at which the plugin should render the overlay.

Revisions to name and description of the configuration options highly appreciated, because I can't word things to save my life apparently.

runelite_2018-06-19_23-40-01


NathenSample:
Will rectify travis CI failure condition's this evening

NathenSample:
travis style issues rectified

NathenSample:
Going to close this for now and re-evaluate.
The idle plugin has problems out of the scope of this PR, namely that the edge cases mentioned above are present in the base plugin, not just my infobox implementation.

Will take a look at the plugin itself, rather than simply adding to the problem.

PR #3174, opened 2018-05-23
Add droplogger plugin

WooxSolo:
Tracks drops received from NPCs, Players, Barrows, Chambers of Xeric and Clue scrolls, and how many times it happened (i.e. killcount).

- Tracks which drops were picked up and which were left on the ground and supports juggling of drops
- Option to show only PvP drops/only PvM drops/both
- Option to show drops picked up/drops left on ground/both
- Option to group kills by name/name and level/name, level and id
- Option to show only drops from a subset of events, e.g. only from chickens as seen in the image below even though both goblins and cows were also tracked

No support for permanent logging yet (all logs disappear after the client is closed), but I plan on adding that in another pull request at some point in the future.

It also doesn't log clue scroll loot if you 0-tick the clue scroll (i.e. receive the loot without opening the interface), but hopefully people don't need that.

Would be nice if we can get some people to test the drop logger at different locations. Particularly at slayer creatures that can only be killed on task as I wasn't able to test these myself.

Fixes #1235

image

PR #3908, opened 2018-06-20
Add Loot Tracking events

TheStonedTurtle:
I took the core functionality from @WooxSolo's droplogger plugin (#3174) and added it to the API (slightly modified).

### Depends on !16

`NpcLootReceived` is triggered when you receive loot from an NPC
`PlayerLootReceived` is triggered when you receive loot from a Player
`EventLootReceived` is triggered when you receive loot from event/activity (Barrows, Clue Scrolls, etc). 
`ItemPickedUp` is triggered when you pick an item up from the floor
`ItemDropped` is triggered when you drop an item

I figure there are probably a lot of optimizations that can be made so any feedback is welcome.

@ Woox if you are already working on this let me know and I'll close this PR and scrap my work.

## Known Issues

<details>
<p>

#### ItemPickedUp Not Firing
* Item Spawning on tile as item is looted
#### ItemPickedUp Mis-fires:
* Item Despawning on tile under you as you gain the item via another method (banking/tele grabbing)
#### ItemDroppedEvent Not Firing:
* Same Item Despawning on tile under you as you drop the item
#### ItemDroppedEvent Mis-Fires:
* Same Item Spawning on tile under you as you remove item via another method (banking/eating/wearing/using)

</p>
</details>

### Subscribe Code Examples

<details>
<p>

```java
	@Subscribe
	protected void onNpcLootReceived(NpcLootReceived l)
	{
		log.info(String.valueOf(l));
		log.info("NPC Name: " + l.getComposition().getName());
		log.info("Received Loot: " + l.getItems().toString());
	}
```
NpcLootReceived(npcId=3272, composition=jq@1ab1757, location=WorldPoint(x=2953, y=3378, plane=0), items=[Item(id=882,quantity=1), Item(id=526,quantity=1)])

```java	
	@Subscribe
	protected void onPlayerLootReceived(PlayerLootReceived l)
	{
		log.info(String.valueOf(l));
                Player p = l.getPlayer();
		log.info("Player Name: " + p.getName());
		log.info("Received Loot: " + l.getItems().toString());
	}
```
PlayerLootReceived(player=bu@13e739f, location=WorldPoint(x=3098, y=3512, plane=0),  items=[ Item(id=526,quantity=1)])

```java
	@Subscribe
	protected void onEventLootReceived(EventLootReceived l)
	{
		log.info(String.valueOf(l));

		String type = l.getEvent();
		List<Item> loot = l.getItems();
		if (type.equals(LootTypes.BARROWS))
		{
			log.info("Barrows Loot: " + loot.toString());
		}
		else if (type.equals(LootTypes.CLUE_SCROLL_HARD))
		{
			log.info("Hard Clue Loot: " + loot.toString());
		}
	}
```
EventLootReceived(event=Barrows, items=[Item(id=560,quantity=121)])

```java
	@Subscribe
	protected void onItemPickedUp(ItemPickedUp i)
	{
		log.info(String.valueOf(i));
	}
```
ItemPickedUp(id=562, qty=1, location=WorldPoint(x=3098, y=3512, plane=0))

```java
	@Subscribe
	protected void onItemDropped(ItemDropped i)
	{
		
		log.info(String.valueOf(i));
	}
```
ItemDropped(id=561, qty=1, location=WorldPoint(x=3098, y=3512, plane=0))

</p>
</details>


runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -132,6 +132,14 @@
 	public static final int DEMONIC_GORILLA_PRAYER_SWITCH = 7228;
 	public static final int DEMONIC_GORILLA_DEFEND = 7224;
 	public static final int IMP_DEATH = 172;
+	public static final int THROW_CHINCHOMPA = 7618;
```
Nightfirecat: Did these animation IDs get in by accident?

TheStonedTurtle: I pulled them from Woox's PR, but off the top of my head I know the Chinchompa and Burst one are used for helping determine if you are using AOE attack

WooxSolo:
I was planning on doing this myself at some point, but if you already got it working I suppose I don't need to do it any more

runelite-client/src/main/java/net/runelite/client/game/loot/events/NpcLootReceived.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+
+/**
+ * An event where you killed an {@link NPC} and received loot
+ */
+@Data
+public class NpcLootReceived
+{
+	@Getter
+	private final int npcId;
```
HSJ-OSRS: i would probably replace the npcId and npccomposition with just the NPC

runelite-client/src/main/java/net/runelite/client/game/loot/events/PlayerLootReceived.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+
+/**
+ * An event where you killed a {@link Player} and received loot
+ */
+@Data
+public class PlayerLootReceived
+{
+	@Getter
+	private final String name;
```
HSJ-OSRS: same for this replace the just passing the name with the player imo

runelite-client/src/main/java/net/runelite/client/game/loot/events/NpcLootReceived.java
Jasper-ketelaar: Actor compositions are set to null if they despawn. This would effectively mean you don't have access to their id or name anymore.

HSJ-OSRS: it worked when i tested it changed it to pass the npc and still could get the name and id

Jasper-ketelaar: Huh. I'm pretty sure originally @TheStonedTurtle wanted to have the implementation you suggested originally but it turned out not to be possible due to what I described.

TheStonedTurtle: I just tested this, again, passing just the NPC and this is the result:
https://i.imgur.com/5Wobo73.png
And here's the subscribe event
https://i.imgur.com/wq0UriT.png

HSJ-OSRS: yea i retried it but came to the same result but im sure i had it working earlier then again i have been awake like 20+ hours i will continue messing around and see if i can figure out how i got it to work if i even did

runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -132,6 +132,14 @@
...
+	public static final int CAST_BARRAGE = 1979;
+	public static final int CAST_BURST = 1979;
+	public static final int CAVE_KRAKEN_DEATH = 3993;
+	public static final int AIR_WIZARD_DEATH = 2553;
```
deathbeam: These anims can be just WIZARD_DEATH? As they have all the same id

runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -132,6 +132,14 @@
 	public static final int DEMONIC_GORILLA_PRAYER_SWITCH = 7228;
 	public static final int DEMONIC_GORILLA_DEFEND = 7224;
 	public static final int IMP_DEATH = 172;
+	public static final int THROW_CHINCHOMPA = 7618;
+	public static final int CAST_BARRAGE = 1979;
```
deathbeam: CAST_BURST_BARRAGE? maybe, same ID again

runelite-client/src/main/java/net/runelite/client/game/loot/LootTypes.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package net.runelite.client.game.loot;
+
+public enum LootTypes
+{
+	UNKNOWN_EVENT,
+	BARROWS,
+	RAIDS,
+	CHAMBERS_OF_XERIC,
+	RAIDS_2,
```
deathbeam: RAIDS_2 and THEATRE_OF_BLOOD? What is the difference, lol

runelite-client/src/main/java/net/runelite/client/game/loot/LootLogger.java
```diff
@@ -0,0 +1,1117 @@
...
+	private boolean completedTheatreOfBloodThisTick = false;
+	private boolean hasOpenedTheatreOfBloodRewardChest = false;
+
+	@Inject
+	private LootLogger(EventBus eventBus)
+	{
+		this.eventBus = eventBus;
+
+		// Init
+		this.interactedActors = new HashMap<>();
```
deathbeam: Initialize when declaring the fiedls

runelite-client/src/main/java/net/runelite/client/game/loot/LootLogger.java
```diff
@@ -0,0 +1,1117 @@
...
+				log.debug("Dropped {} of itemId: {} at {}", qty, id, location);
+				eventBus.post(new ItemDropped(id, 1, location));
+			}
+		}
+		else
+		{
+			log.debug("Dropped {} of itemId: {} at {}", qty, id, location);
+			eventBus.post(new ItemDropped(id, qty, location));
+		}
+
```
deathbeam: Remove this empty line

runelite-client/src/main/java/net/runelite/client/game/loot/LootLogger.java
```diff
@@ -0,0 +1,1117 @@
...
+	}
+
+	/**
+	 * Converts the drops Map into a List of ItemStacks
+	 *
+	 * @param drops A Integer, Integer map of ItemIDs and Quantities
+	 * @return A list of ItemStack's
+	 */
+	private List<ItemStack> createItemList(Map<Integer, Integer> drops)
```
deathbeam: Why not use ItemStack list in first place?

TheStonedTurtle:
@deathbeam I've made the requested file changes but I'm not entirely sure what i need to change to accommodate your most recent comment:

> Also, reducing the super extensive creation of new collections for every single thing almost every other tick would be nice. Also, less collection transformations for no reason. That is probably the biggest thing in this PR that needs to be changed.

If you could show me a few specific cases I should be able to take it from there.

deathbeam:
Well its just what I noticed from going through the code quickly. I am sure most of it can be reduced, but that requires  studying the PR a bit more. As, this will be events, they would not be able to be turned off so the acompanying overhead should be minimal, so that is where the focus here should go.

TheStonedTurtle:
@deathbeam I'm not sure what else i can optimize, can you review in more depth and give me some input? I've wrote up an explanation of what exactly this is doing which will hopefully speed up your review of how this is working.

<details>
<p>

In order to properly track loot we need to know what items were on the ground last tick. We will also want to account for the player dropping/picking up items on the same tick which is why we are tracking the players inventory.

### How Ground Item Memorization Works:
* 1) Listen for `ItemLayerChanged` and memorize the tiles that were changed.
* 2) Every tick loop over changed tiles and add tile content to memory
* 3) Clear map when region ID changes

### How Player Inventory Events Works:
* 1) Listen for `ItemContainerChanged` and store new inventory if it was an inventory change
* 2) Every tick check if we memorized a new inventory, this would indicate an item was added/removed from the inventory.
* 3) Calculate if the item(s) were dropped/picked up or added/removed by other means.

### How Interacted Actors work:
* 1) Every Tick memorize actors we've interacted with.
* 2) Check for cannon placement/pickup, if placed check and memorize npcs cannonballs hit.
* 3) If AOE attack check for adjacent targets (Only checks for same Actor type, supports NPC/Players)

### How Loot Tracking Works:
* 1) Listen for `ActorDespawned` and add to list if we've interacted with it. (SEE NOTE BELOW)
* 2) Every tick we loop over the despawned actors list and determine where their loot spawned. We compare the current ground items with the old ground items and account for the player picking up/dropping loot.
* 3) Trigger correct event for Actor type with loot items

**NOTE:** Certain NPCs drop their loot or determine where to drop it before despawning. These NPCs are added to this list via `AnimationChanged` instead of `ActorDespawned`

### How Event Loot Tracking Works:
* 1) Listen for `WidgetLoaded` and set the event flag to true if its GroupId matches.
* 2) Listen for `onItemContainerChanged` and store the data if its one of the tracked events item container.
* 3) Every tick we check each `WidgetLoaded` flag, if true we trigger a LootReceived event for this Activity

### Game Tick
To accomplish all of this, excluding the events we subscribe to as mentioned above, we must do the following (in order) every tick:
* 1) Memorize new interacted actors
* 2) Check for and trigger Event rewards
* 3) Check for player inventory changes
* 4) Loop over Dead actor and determine what loot they dropped
* 5) Loop over changed tiles and memorize updated Ground Items
* 6) Store current inventory data in previous tick variable
* 7) Store current player world point in previous tick variable

**NOTE** All `Item`s (Ground, Loot, & Inventory) are stored using the ItemStack class

</p>
</details>

psikoi:
@deathbeam i guess this too

PR #3917, opened 2018-06-20
Add debuff/buff change indicators

### LINKED ISSUE #3983: Preserve next stat change even if not boosted
Would it be possible to have a timer to keep track of the servers clock for potions even when your not using a potion ? I'm not sure how the mechanics of the timer work, but at the least i'd imagine it possible to keep it ticking after a potion has been used, if not from log in. 
Thanks

deathbeam:
- Add support for dynamically updating infoboxes
- Create infoboxes only once and then determine if they should be drawn
- Add indicators for both buff and debuff timers. By default only buff
timer is enabled.
- Add support for showing boost indicators after no stat is boosted based
on last accurate change.

Closes #3983

Preview:
peek 2018-06-21 18-29
peek 2018-06-21 18-21


Abextm:
- I can't get the Next + restore to open, ever
 - The Next - restore resets when you drink something, even if the timer is already running

Kamielvf:
What's the restore useful for anyway?

deathbeam:
Updated the PR and retested now it should be working. I did some last minute fixes yesterday before going to sleep in belief that it will reset the timers properly, but apparently it just broke it so added reset logic to GameTick instead.

deathbeam:
Added extra support for preserving the change indicators as that was original goal of this.

deathbeam:
Apparently the accuracy of the boosts timer while nothing is boosted is in mutlitudes of 15 for w/e reason. It mostly works, idk how to determine which quarter is the correct one, so its better than nothing.

deathbeam:
Ignore last comment. The timers were accurate but I missed that sometimes I was getting 14 second countdown instead of 60 for some reason and that made the overall timer off. When that was not happenign the timers were mostly tick perfect without any boost active (as long as there was starting point that was accurate, e.g boost change). I need to figure that out, as I have no idea.

deathbeam:
Updated and fixed

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -224,24 +321,52 @@ public void addStatChangeIndicator()
 	 *
 	 * @return integer value in seconds until next boost change
```
Abextm: Mismatched docs

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -224,24 +321,52 @@ public void addStatChangeIndicator()
...
-		return (timeSinceChange > 60) ? 75 - timeSinceChange : 60 - timeSinceChange;
+		return (ticksSinceChange > 100) ? 125 - ticksSinceChange : 100 - ticksSinceChange;
 	}
 
-	private int timeSinceLastChange()
+	int getChangeDownTime()
 	{
-		return (int) Duration.between(lastChange, Instant.now()).getSeconds();
+		final int changeDownTicks = getChangeDownTicks();
+		return changeDownTicks != -1 ? (int)(changeDownTicks * 0.6) : changeDownTicks;
```
Abextm: This should probably account for the time between call and when it the last tick was so the timer decreases at a regular cadence.

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -224,24 +321,52 @@ public void addStatChangeIndicator()
...
 	}
 
-}
+	/**
+	 * Restoration from debuff is separate timer as restoration from buff, and is always ticking, so just find
+	 * diff between last change up and now and limit it to cycles of 60.
+	 *
+	 * @return integer value in seconds until next stat restoration up
```
Abextm: Doc mismatch

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -224,24 +321,52 @@ public void addStatChangeIndicator()
...
+		}
+
+		int ticksSinceChange = client.getTickCount() - lastChangeUp;
+		return 100 - ticksSinceChange;
+	}
+
+	int getChangeUpTime()
+	{
+		final int changeUpTicks = getChangeUpTicks();
+		return changeUpTicks != -1 ? (int)(changeUpTicks * 0.6) : changeUpTicks;
```
Abextm: Same thing

deathbeam:
Updated based on comments

PR #3919, opened 2018-06-20
Adds a chat message that shows the layout when starting the raid

Matsyir:
image
Little QOL for Chambers of Xeric so you can keep track of the rotation and know what's next. I usually write it in the chat but this is a bit better. I tested it on 20+ raids like in the picture, writing the rotations beforehand. Included a toggle which is default to on. I'm not a big fan of the text color either but it's to keep it consistent with the rest of the raid's text, and it will contrast with the regular game text once you actually start doing stuff.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsConfig.java
@@ -81,6 +81,17 @@ public interface RaidsConfig extends Config
 
 	@ConfigItem(
 		position = 4,
+		keyName = "layoutChatMsg",
+		name = "Show layout in chat on raid start",
+		description = "Shows the raid layout in chat when starting the raid, similarly to the overlay."
+	)
+	default boolean layoutChatMsg()
+	{
+		return true;
+	}
+
+	@ConfigItem(
+		position = 5,
 		keyName = "whitelistedRooms",
 		name = "Whitelisted rooms",
 		description = "Display whitelisted rooms in green on the overlay. Separate with comma (full name)"
@@ -91,7 +102,7 @@ public interface RaidsConfig extends Config
 	}
 
 	@ConfigItem(
-		position = 5,
+		position = 6,
 		keyName = "blacklistedRooms",
 		name = "Blacklisted rooms",
 		description = "Display blacklisted rooms in red on the overlay. Separate with comma (full name)"
@@ -102,7 +113,7 @@ public interface RaidsConfig extends Config
 	}
 
 	@ConfigItem(
-		position = 6,
+		position = 7,
 		keyName = "enableRotationWhitelist",
 		name = "Enable rotation whitelist",
 		description = "Enable the rotation whitelist"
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -270,6 +271,57 @@ public void onChatMessage(ChatMessage event)
...
+					// overlay displays it. The layout, then combat & puzzle rooms.
+					String layout = getRaid().getLayout().toCode().replaceAll("#", "").replaceAll("¤", "");
+					ChatMessageBuilder msgBuilder = new ChatMessageBuilder()
+							.append(ChatColorType.HIGHLIGHT)
+							.append("[")
+							.append(layout)
+							.append("]: ");
+
+					boolean firstElement = true;
+					for (Room layoutRoom : getRaid().getLayout().getRooms())
```
Jasper-ketelaar: I would personally use streams for this as it makes the code quite a bit easier to read:
```
getRaid().getLayout().getRooms().stream()
			.map(e -> getRaid().getRoom(e.getPosition()))
			.filter(Objects::nonNull)
			.filter(e -> e.getType() == RaidRoom.Type.COMBAT || e.getType() == RaidRoom.Type.PUZZLE)
			.map(e -> e.getType() == RaidRoom.Type.COMBAT ? e.getBoss().getName() : e.getPuzzle().getName())
			.reduce((s1, s2) -> s1 + ", " + s2)
			.ifPresent(msgBuilder::append);
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsConfig.java
```diff
@@ -81,6 +81,17 @@ default boolean scoutOverlayAtBank()
 
 	@ConfigItem(
 		position = 4,
+		keyName = "layoutChatMsg",
+		name = "Show layout in chat on raid start",
```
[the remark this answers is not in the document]
Matsyir: Somewhat agree but I did make sure it was entirely visible without hovering. Initially I didn't put the word "raid" in there but it seemed a bit more clear with it.

Kamielvf:
Shouldn't this just show when the raid gets scouted instead?

Matsyir:
Hmm I could add the option for that, but I think its pointless since the scouting overlay is already present until the start, also it would really spam your chat when looking for a layout

Kamielvf:
People might prefer the text message over the overlay (which you can already toggle off in the settings) so I think it should be done when the raid gets scouted.

My plan is to make the overlay be able to stay active inside of the raid but I have to rewrite some stuff which got me stuck with an error I was unable to fix so far.

Matsyir:
Understandable, once I get home (around Monday, uncertain) I'll add the option for that and implement Jasper's stream solution since it is much nicer. My initial idea was to keep the overlay there aswell but I figured it would be too intrusive, though I see why someone would prefer it over a chat message. I didn't plan on adding whitelist/blacklist colors, or Puzzle/Combat indicators for the unknown rooms (but it can be inferred from the scfpcc.. layout), so as it stands the chat message would not quite be as fully featured as the overlay, but I will look into implementing those as well.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Kamiel
+ * Copyright (c) 2018, Kamiel, Matsyir, Jasper-ketelaar
```
Nightfirecat: New copyrights should be separate entries on new lines.

deathbeam:
Implemented in #7545

Issue #1727, opened 2018-04-20
Notify on full special attack


lulwut:
A toggle for the notification sound and a notification pop up for when the special attack returns to 100% would be great. Maybe to prevent it being overpowered only notify on skill weapon 100%s (D pick, Harpoon etc.). Also if it is possible to not have the notification play when one returns their spec to full using clan wars/dying/house pool as it would be redundant in these situations and would get annoying.

PR #3922, opened 2018-06-21
Feature implementation for request #1727: Notify on full special attack.

dicrypt:
References issue [#1727](https://github.com/runelite/runelite/issues/1727)
Renamed Special Attack Counter menu option to Special Attack in order to house more related options later.
Moved Special Attack Counter into configuration page of Special Attack option.
Added a new preference to Special Attack configuration page to get notified when reaching 100% special attack.


runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -246,4 +268,19 @@ private int getHit(SpecialWeapon specialWeapon, int deltaExperience)
...
+	private void notifySpecialChange(int special)
+	{
+		// Not enabled, break
+		if (!config.enableSpecialAttackNotification())
+		{
+			return;
+		}
+		// Notify if we are now 100% and the last updated percentage was not default, 0%, or 100%
+		// Logging in with full special simultaneously triggers 0% and 100%
+		if (special == 1000 && this.specialPercentage > 0 && this.specialPercentage < 1000)
```
Jasper-ketelaar: As suggested in the issue of which this PR is based. This will also trigger when restoring from the house pool. Would suggest only notifying if the user has restored their special attack according to the natural restore rate. This would probably still cause some cases where the notification is sent when you are restoring at 90%-95% in your house pool but it would eliminate all the other instances.

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterConfig.java
```diff
@@ -0,0 +1,35 @@
...
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup(
+	keyName = "specialAttack",
+	name = "Special Attack",
+	description = "Configuration for special attack options"
+)
+public interface SpecialCounterConfig extends Config
```
Jasper-ketelaar: Shouldn't all of the classes now be renamed to SpecialAttack* instead of SpecialCounter* as it's not solely a special counter anymore?

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterConfig.java
```diff
@@ -0,0 +1,35 @@
+package net.runelite.client.plugins.specialcounter;
```
Jasper-ketelaar: Header

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -246,4 +268,19 @@ private int getHit(SpecialWeapon specialWeapon, int deltaExperience)
 			return (int) Math.round(damageOutput);
 		}
 	}
+
+	private void notifySpecialChange(int special)
```
Jasper-ketelaar: notifySpecialChange might be a bit of a deceiving name as that is not always what it does. I think naming it something like checkFullSpecial.

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -246,4 +268,19 @@ private int getHit(SpecialWeapon specialWeapon, int deltaExperience)
...
+	{
+		// Not enabled, break
+		if (!config.enableSpecialAttackNotification())
+		{
+			return;
+		}
+		// Notify if we are now 100% and the last updated percentage was not default, 0%, or 100%
+		// Logging in with full special simultaneously triggers 0% and 100%
```
Jasper-ketelaar: I know it's not your naming but since you do make use of it. Can the class field specialPercentage be renamed to previousSpecialPercentage as it's a bit confusing to see it be used as such but not have a clear name.

Adam-:
I think this should probably be part of the idle notifier instead of the spec attack (counter) plugin.

PR #3928, opened 2018-06-21
Fix NPC indicators to not consume tags clicks on non-NPCs

Kamielvf:
Fix an issue introduced by #3778 that would prevent the player from clicking the "Tag" option on pillars in the pinball random event.


Nightfirecat:
That's an oddly specific bug. :+1:

Kamielvf:
Also related: agility arena

Adam-:
Usually when we add our own actions they are of type RUNELITE - can you not just check that?

Kamielvf:
It calls `menuManager.addNpcMenuOption` which loops over the available NPC actions of that npc's composition to look for an empty spot, then fills it with the new action.

On another note, if all options are already taken there won't be a new one added. I'm not sure if there's NPCs that have that many actions (5) but it is likely possible to happen at some point.

Alexsuperfly:
menaphite thugs cant get tagged because of that

PR #3930, opened 2018-06-21
Revert "kourendlibrary: Hide navbutton when not in the library"

### LINKED ISSUE #3929: Arc Library Plugin Update
Hello, I know that today's update makes the arc library dissapear when you are away from the area for those that only search for books for a limited time. There are many ironmen that have been hurt in terms of xp rates from this. We use Woox's method which we catch imps with magic box's and use them to bank bulk amount of books. Yesterday I was getting 60k rc xp per hr at 68 rc with 6k books banked total. I used the plugin to figure out the amount of books I need since there are 15 books and we use 2 of each besides the one on ground floors. We use one of each of those. Now we have to walk towards the library if we forgot what books are called which given the names, its common to get them mixed up. I ask that you add in a toggle option for it not to dissapear or atleast include the region of the bank to help accomodate a very good RC method. Also we have to always manually click on the plugin again upon entering.

MagicfTail:
Ironmen were using the tab being visible when not being at the library when doing the "Woox method". If people don't use the panel they can simply turn off the plugin.
Also, set the plugin to be disabled by default.

This reverts commit 64b9b7a1686c150249a9a3c17fe81c77e24e1e07.

Fixes #3929

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
@@ -53,6 +53,7 @@ public class KourendLibraryOverlay extends Overlay
 {
 	private final static WorldPoint LIBRARY_CENTER = new WorldPoint(1632, 3807, 1);
 	private final static int MAXIMUM_DISTANCE = 24;
+	private final static int ROUGH_ENABLE_DISTANCE = 45;
 
 	private final Library library;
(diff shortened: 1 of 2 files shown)


deathbeam:
I dont think this should be reverted, because we have the plugin on by default and we are adding panel that is used in only 1 specific area by default. We dont do such thing with any other "activity" plugin, so current default behaviour is correct and should not be changed. Probably adding configurable option for it would be better.

Adam-:
Fixed in #3934

PR #3933, opened 2018-06-21
NPC Composition

TheStonedTurtle:
`npc.getComposition()` will now return a cached version of the composition if an updated composition can't be found


TheStonedTurtle:
Closed per discussion with Adam

Issue #3946, opened 2018-06-22
MTA plugin: Telekinetic solver fails to update 


abextm:
On commit 26fbf78399e328b73fee72d6a1d7f1841e528b50

Sometimes the telekinetic solver fails to update after the room has changed state. I don't know exactly what causes it, but it happens every 10 rooms on average.

![2018-06-22_02-06-11](https://user-images.githubusercontent.com/12366911/41765279-da077a4c-75c0-11e8-8b3d-e754a3e21876.png)

Jasper-ketelaar:
Should also be fixed now in #3947

Adam-:
Do you know why #3947 fixes this or what the issue is?

troypressley1:
I wonder if this'll ever be able to be fixed

Spedwards:
Doesn't look like this is getting fixed anytime soon. Shame.

MalakaJones:
Still an issue.

Adam-:
Sorry for the wait. This is now fixed.

PR #3947, opened 2018-06-22
Mta plugin fix + Telekinetic room rework

### LINKED ISSUE #3945: MTA plugin: Hint arrow does not get cleared
On commit 26fbf78399e328b73fee72d6a1d7f1841e528b50

The hint arrow does not get cleared sometimes after leaving the enchanting room

Jasper-ketelaar:
This fixes #3945 and #3946. The telekinetic room has been rewritten to be more robust, more efficient and easier to understand.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
@@ -66,7 +66,6 @@ public class EnchantmentRoom extends MTARoom
 			if (!inside())
 			{
 				dragonstones.clear();
-				client.clearHintArrow();
 			}
 		}
 	}
@@ -121,6 +120,7 @@ public class EnchantmentRoom extends MTARoom
 			boolean removed = dragonstones.remove(worldPoint);
 			if (removed)
 			{
+				clearArrowIfPresent(worldPoint);
 				log.debug("Removed dragonstone at {}", worldPoint);
 			}
 			return;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
```diff
@@ -66,6 +66,7 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
 			if (!inside())
 			{
 				dragonstones.clear();
+				client.clearHintArrow();
```
Adam-: This would probably break all game content which uses the hint arrow

Jasper-ketelaar: Yeah, completely did not realize. I've changed it to clear the hint arrow whenever a dragonstone despawns and the dragonstone location was the current hint arrow location.

delps1001:
Can this be re-reviewed at some point? The telekinetic room is still broken.

Adam-:
There is a difference in "fix" and "rewrite completely". Eg, this claims to fix a bug but does not detail what the bug even is. It's not possible to tell from the commit history.

Jasper-ketelaar:
Well you say that Adam but we had a discussion on discord about what it fixed at the time of making this pull request, you just said you didn't have the time to properly test the rewrite. 

This fixes issue #3946 which should be detailed enough and it also suggests the next move when you cast telekinetic grab instead of when the guardian stops moving.

ShaunDreclin:
Hey if you're still around could you rebase/update this?

Nightfirecat:
Fixed in #9328.

PR #3950, opened 2018-06-22
Bug fix binding necklace destroy

### LINKED ISSUE #3874: Binding necklace- Destroy issue
| Operating System | /Windows 10

When I use binding necklace and destroy it, the client doesn't seem to register it's been destroyed so the charges stay the same. It changes the display on the necklace when I use a new binding necklace to '?' untill the next one goes to the number of charges the previous binding necklace was at. Kind of hard to track when i'm trying to use the necklace to disintegrate with every two duelling rings for Runecrafting.

honeyhoney:
Adds widgets for the destroy item interface.

image

Resolves #3874


runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -158,6 +165,22 @@ public void onChatMessage(ChatMessage event)
...
+	public void onMenuOptionClicked(MenuOptionClicked event)
+	{
+		Widget widgetDestroyItemName = client.getWidget(WidgetInfo.DESTROY_ITEM_NAME);
+
+		if (widgetDestroyItemName == null || !widgetDestroyItemName.getText().equals("Binding necklace"))
+		{
+			return;
+		}
+
+		if (event.getWidgetId() == WidgetInfo.DESTROY_ITEM_YES.getId())
```
Adam-: Probably move this check first since it is cheaper.

PR #3959, opened 2018-06-22
Slayer overlay items refresh on render

JHPinto:
Currently the slayer plugin checks for slayer items on each game tick, which on occasion causes text to be persist longer than expected when swapping between the inventory and equipment instances.
The queries are now run in the render call in the SlayerOverlay class, preventing this issue from occurring.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
@@ -34,8 +34,11 @@ import java.awt.Rectangle;
 import java.util.Collection;
 import java.util.Set;
 import javax.inject.Inject;
+
+import com.google.common.eventbus.Subscribe;
 import net.runelite.api.ItemID;
 import net.runelite.api.Query;
+import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.queries.EquipmentItemQuery;
 import net.runelite.api.queries.InventoryWidgetItemQuery;
 import net.runelite.api.widgets.WidgetInfo;
@@ -51,7 +54,7 @@ class SlayerOverlay extends Overlay
 {
 	private final SlayerConfig config;
 	private final SlayerPlugin plugin;
-        
+	private Collection<WidgetItem> items;
 	@Inject
 	private QueryRunner queryRunner;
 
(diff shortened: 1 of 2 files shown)


Jasper-ketelaar:
Running two queries on every render tick is quite expensive. I would instead suggest subscribing to the ItemContainerChanged event and using this to get the WidgetItems that need to be displayed.

JHPinto:
@Jasper-ketelaar Does not fix the problem when switching between the inventory and equipment screens.

deathbeam:
Rebase this properly

JHPinto:
Rebased

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
```diff
@@ -24,27 +24,36 @@
...
 import net.runelite.client.ui.overlay.components.TextComponent;
+import net.runelite.client.util.QueryRunner;
 
 class SlayerOverlay extends Overlay
 {
 	private final SlayerConfig config;
 	private final SlayerPlugin plugin;
 
+	@Inject
+	private QueryRunner queryRunner;
```
deathbeam: Inject this in Overlay constructor

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
```diff
@@ -76,6 +85,18 @@
 		ItemID.EXPEDITIOUS_BRACELET
 	);
 
+	private ImmutableList<WidgetItem> checkInventory()
```
deathbeam: Why is this method above contructor?

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
```diff
@@ -76,6 +85,18 @@
...
+	{
+		Query inventoryQuery = new InventoryWidgetItemQuery();
```
deathbeam: Use .idEquals here

JHPinto: What benefit would that bring in practice?

deathbeam: It will not query all inventory items but only needed ones

Issue #3911, opened 2018-06-20
Improvement: Encapsulation of plugins


NathenSample:
**Problem**
Access modifiers in some plugins are too permissive, meaning there's potential down the road to end up in weird scenarios where plugin A tries to use logic, or utilise internal classes from plugin B. This was originally brought to my attention by @deathbeam 

**Solution**
Gradually update plugins to respect proper encapsulation, in theory only the plugin itself, as well as overriden methods, need be more permissive than `package-private`

I'll be opening PR's against this issue on a plugin-by-plugin basis

PR #3962, opened 2018-06-23
Update prayer plugin encapsulation

NathenSample:
Relating to #3911

Simply reducing public methods/class to package-private to prevent re-use of plugin logic/classes external to the plugin itself

Tested post-change
Boost Indicator/Overhead indicator

Flick Helper

Dose Indicator
runelite_2018-06-23_01-45-37

prayer stats

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerFlickOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerFlickOverlay.java
@@ -46,7 +46,7 @@ class PrayerFlickOverlay extends Overlay
 	private Instant startOfLastTick = Instant.now();
 
 	@Inject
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerFlickOverlay.java
```diff
@@ -39,14 +39,14 @@
...
-public class PrayerFlickOverlay extends Overlay
+class PrayerFlickOverlay extends Overlay
 {
 	private final Client client;
 	private boolean prayersActive = false;
 	private Instant startOfLastTick = Instant.now();
 
 	@Inject
-	public PrayerFlickOverlay(Client client)
+	PrayerFlickOverlay(Client client)
```
deathbeam: This should be private

PR #3963, opened 2018-06-23
Even more hot-cold clue fixes

Nightfirecat:


deathbeam:
This is some weird diff

Nightfirecat:
Not sure why there were all those non-breaking spaces in there to begin with. Just changed those in a separate commit. I can back that out if needed. :man_shrugging:

Issue #3969, opened 2018-06-23
Feature Request: Inventory slots notifier


SwampyApple:
**Is your feature request related to a problem? Please describe.**
The current implementation of the idle checker for various skills works fine, but I believe it could be a lot more efficient for many different tasks if it would notify the user _before_ the inventory is completely filled up or before the last action is performed. Because the character will already stand idle by the time the user is notified in the current solution.

**Describe the solution you'd like**
There could be a plugin which you enable, in which you toggle that you either want to check for a certain amount of free spaces or a certain amount of filled spaces. For example:

Warn me when the inventory has 25 or more filled slots.
Warn me when the inventory has 10 or fewer empty slots. 

This allows for both resource consuming and generating tasks to get a preemptive warning, allowing users to continue their task more efficiently, before the character becomes idle. 

It would be applicable to many different use cases and I believe it not to be an overly complex concept.

Jasper-ketelaar:
I think this is a great idea. Implementing it right now I have a question though: Would you want the plugin to warn you only for when you reach x filled slots or for every slot you fill after you reach the notifier threshold too?

SwampyApple:
Thank you @Jasper-ketelaar for your quick contribution. I glanced over #3972 and it appears only an increased inventory count is checked. Would it be possible to add an option to check for a decreasing inventory count as well? This way the plugin can have added value for item consuming tasks like crafting. Warning the user that the inventory is about to run out, before it actually happens.

Jasper-ketelaar:
Done. Anytime the inventory amount is now reach (either by increase or decrease), the notification is triggered.

Nightfirecat:
This should be possible [via the Advanced Notifications hub plugin](https://runelite.net/plugin-hub/show/advanced-notifications).

PR #3972, opened 2018-06-23
Idle Notifier Plugin: Add inventory slot notifier

Jasper-ketelaar:
Feature was requested in #3969. Sends you a notification when your specified inventory threshold is reached. Example use is setting the threshold to 27 in Motherlode Mine so you know you need to pay attention as your inventory is about to fill up.


Nightfirecat:
Closed [per my comment on the target issue](https://github.com/runelite/runelite/issues/3969#issuecomment-1637451395)

Issue #3968, opened 2018-06-23
Chat-command for clue scrolls completed (+ Rank)


J-Leg:
**Is your feature request related to a problem? Please describe.**
A simple addition to the chat-commands, because right now the list looks a bit lonely :smiley:.

**Describe the solution you'd like**
Uses the ```HiscoreClient``` to fetch the relevant data.  

Something like this:
![screenshot from 2018-06-23 15-18-34](https://user-images.githubusercontent.com/30682722/41806071-8fb197e0-76f9-11e8-9edf-cc52d1887b16.png)

Nightfirecat:
@J-Leg I think you implemented this with #3973?

In the future, [you can close the issue from a PR using specific keywords](https://help.github.com/articles/closing-issues-using-keywords/) so it is closed automatically upon merge.

deathbeam:
Implemented in #3973

J-Leg:
Ah yeah, was wondering how to do that. Thanks!

PR #3973, opened 2018-06-23
Clue chat commands

J-Leg:
#3968
Small addition to chat commands: 

```!clues```  - Number of clues completed, any level
```!clues easy``` - Number of easy clues completed
for all clue levels etc...

These commands can also be toggled the same way ```!lvl``` are toggled.


J-Leg:
Reformatting.

PR #3992, opened 2018-06-24
Add RuneLite Actions HotKey

jplsek:
Add a "RuneLite hotkey" to the RuneLite settings. This allows changing the "ALT" key to anything the user wants. I partly did this because WMs for Linux use ALT for window management actions (such as dragging the window), and I didn't like that I had to modify my WM key (granted META kind of makes more sense, but some WMs make remapping to META difficult). This may also add extra accessibility implications with this change. A common configuration I would expect users to use would be "CTRL".

I updated all usages of the hardcoded "ALT" to use this hotkey, such as dragging overlays, ground items, and GE searching.

There ended up being a few bugs with the GE plugin since it did not consume all events and would lose game focus while searching, but that should be fixed now.

There will probably be other unforeseen consequences we may find over time, but I think it's mostly from not "consuming" events properly.


Caia-OSRS:
Any chance this makes it in this weeks update? :)

PR #3997, opened 2018-06-24
Remove title from corp panel

deathbeam:
This is completely unnecessary as it is obvious and is just using screen
estate for no real reason. Also we dont do this for any other plugin.


Adam-:
I thought the title looked nice. I guess I do not care much either way though.

PR #4000, opened 2018-06-24
runecrafting: add color overlay to runepouches

### LINKED ISSUE #1451: colours for Runecrafting pouches
Green represents full (there are rune/pure essence inside it)
Yellow/orange represents empty (there are no rune/pure essence inside it)
red represents degrading (it needs to be restored by talking to dark mage by the npc contact spell)

image
something similar to this sorry don't have the red colour.

jplsek:
runepouch
runepouch2

Red = Degraded
Yellow = Not full
Green = Full

Not sure why the original PR was closed. But this PR simplifies it by only acting like other clients.

Would close #1451 if it wasn't mistakenly closed ^_^


deathbeam:
With new toBufferedOutline method this should probably use item outlines like inventory tagging instead.

jplsek:
Alright, updated to use outlines.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftOverlay.java
```diff
@@ -53,6 +58,8 @@
 
 	private final RunecraftConfig config;
 
+	private final Map<String, BufferedImage> storedOutlines = new HashMap<>();
```
deathbeam: As this is second plugin using outline API and basically duplicating the logic, maybe extending sprite manager with it would make sense

jplsek:
Added outline caching to item manager and switched both plugins to utilize it.

honeyhoney:
I massively prefer the implementation of an overlay seen in #1451
It's much more immediate than an outline.

jplsek:
Added a chooser to select between either option (or none at all).

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/OverlayStyle.java
```diff
@@ -0,0 +1,45 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.runecraft;
+
+public enum OverlayStyle
```
LeviSchuck: Use required args constructor, no need to write your own

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftConfig.java
```diff
@@ -37,19 +37,32 @@
...
+		description = "Configures the colored overlay style",
+		position = 0
+	)
+	default OverlayStyle overlayStyle()
+	{
+		return OverlayStyle.OUTLINE;
+	}
+
+	@ConfigItem(
+		keyName = "showPouchCount",
```
LeviSchuck: Changing the key name will forget all existing settings people have, are you sure you want that?

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftOverlay.java
```diff
@@ -77,36 +86,83 @@ public Dimension render(Graphics2D graphics)
...
 
 		return null;
 	}
+
+	private void renderOverlay(final Graphics2D graphics, final WidgetItem item, final Color outlineColor, final Color backgroundColor)
+	{
+		if (config.overlayStyle() == OverlayStyle.OUTLINE)
+		{
+			final BufferedImage outline = itemManager.getItemOutline(item.getId(), item.getQuantity(), outlineColor);
+			graphics.drawImage(outline, item.getCanvasLocation().getX() + 1, item.getCanvasLocation().getY() + 1, null);
```
LeviSchuck: Are you trying to do a drop shadow effect? How do you explain this plus one offset?

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftConfig.java
jplsek: Whoops, that was probably a find/replace issue. Granted, the behavior is technically different. But I'll change it back.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftOverlay.java
jplsek: Yep, as @deathbeam said in Discord, this is just how the outline effect works to be aligned properly. It's the same for the inventory tag plugin.

LeviSchuck: Would be nice for a comment explaining the plus one.
The plus sixteen above too.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/OverlayStyle.java
```diff
@@ -0,0 +1,43 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.runecraft;
+
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public enum OverlayStyle
```
Nightfirecat: Does this need to be public? Should probably be package-private.

jplsek: Yes it does. Otherwise an IllegalAccessError gets thrown at runtime.

oplosthee:
While runecrafting I made something similar to your plugin, except mine changes colors one "iteration" before a pouch degrades. I think we can probably combine these two plugins into one, where the pouch is colored green by default, yellow when it is about to degrade and red when it has degraded. 

I'm not sure if showing the whether the pouch is full really adds much since the number in the top left pretty much already shows this. Would like to hear some opinions on this.

jplsek:
IMO, signifying whether the pouch is full is somewhat necessary in a QoL standpoint. I don't need to know the max essence each pouch holds and it goes along with not knowing, at a glance, which pouch is which.

The "status" that could be removed could be "Not Full" because players only really need to know if the pouch is full or degraded.

So the Not Full status could be removed in place of "about to degrade", or, add a new color, like orange, to signify that.

But I think that's out of the scope of this PR.

oplosthee:
Thought about it some more and I think I agree with you. I think I would however change the colors a slight bit, making red indicate an empty pouch over a degraded pouch. The sprite already indicates whether a pouch has degraded, and by changing red to empty you can disable the numbers in the top left and still see whether a pouch is empty or not.

I think I'll address the "about to degrade" stuff in a separate PR.

PR #4006, opened 2018-06-25
runelite-client: Fix JRE bug to allow OSX to fallback to the system font

Abextm:
This is implemented primarily for Hotkey's toString, because modifier keys on OSX use special characters instead of words.

https://youtrack.jetbrains.com/issue/JRE-599


deathbeam:
Now instead of wrong characters it is displaying no characters at all for modifiers (or well, rather, the binding button remains on Not Set).

Abextm:
Hmm, that sounds unrelated. Is the config value actually getting set to a new value, or did this somehow break string concatenation?

PR #4013, opened 2018-06-25
runelite-client: Don't use system specific modifier key names

Abextm:
Without this OSX shows a replacement character, see #4006. When/If #4006 is merged, this should be reverted.


deathbeam:
Looks like this sorts the issue for modifier keys. Cant really bind some combinations, because the key gets sent with modifier applied, about what idk if we can do much. E.g if i would want to bind Alt + S, it binds ALT + ß (because modifier being applied). Also, backspace still shows bad character.

PR #3143, opened 2018-05-23
Add kills left to xp tracker and globe

DaanG96:
I closed #1645 and implemented the feature after a few changes were made in the xp tracker code. Also added the kills left to the xp globe now.

screenshot
closes #1594

Issue #1594, opened 2018-04-16
Add "Kills Left" on the XP Tracker for combat kills


kleinaeinar:
Right now the combat XP Tracker shows "### Action Left" until a level up.  If the client could be updated with "### Kills Left" for the combat skills that would help out a lot.  Because right now I am 62 ranged, and I look at the XP tracker and it shows "1612 Actions Left" and than a few second later "1323 Actions Left", few seconds after that "1852 Actions Left".  It is not very accurate when it comes to combat skills.

king-koda:
i think thats because the actions left is calculated based on how much xp is given for that particular hit eg. 40xp for hit of 10, 80xp for hit of 20, that's why it isn't accurate. The kills left should be implemented or at the minimum, actions left removed because of it's misinterpretation.

MrHoward190:
Hows progress on this going? Was just talking to a friend about how useful it would be.

Nightfirecat:
@deathbeam I think this one was closed in error.

MrHoward190:
Any update on progress? Does this look doable?

IanMcNelly:
Opened PR #4015 to address this, based off the work in #3143

Kingdavid337:
Is this closed, or has it just not been accepted by anyone?

IanMcNelly:
I had it working finally locally, and posted a few times in discord/github asking for more feedback on my PR #4015 , and it was working fine locally but gave up on keeping it up to date because i couldnt get a resposne. On https://github.com/runelite/runelite/pull/3143 @LeviSchuck posted he would pick it up again

Adam-:
I'm sorry it's been neglected. I've left some review on #4015 now.

IanMcNelly:
@Adam- awesome thanks for the feedback! I'll take a look asap and see if i can get it into a state where its mergable :)

Titowam:
Any updates? This would be a great addition!

IanMcNelly:
@Titowam this was added in october: https://github.com/runelite/runelite/pull/6163

PR #3939, opened 2018-06-21
Add Player Comparison table to Opponent Information plugin

Nightfirecat:
Player Comparison table

This adds a configuration to the opponent information plugin which, if
enabled, will show a table comparing the stats of your player and
players you interact with, highlighting any differences with green and
red.

Fixes runelite/runelite#2961

PR #4015, opened 2018-06-26
Issue #1594 kills left to lvl

IanMcNelly:
Took the commits made in https://github.com/runelite/runelite/pull/3143 and reworked based on the PR feedback. Tested with all Melee/Range styles and interacting with players and NPC's, no errors are raised and kill counts look correct. Don't have an XP lamp to test with however, so not sure if that switch works correctly


runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -287,4 +296,13 @@ private void drawTooltipIfMouseover(Graphics2D graphics, XpGlobe mouseOverSkill,
 
 		xpTooltip.render(graphics);
 	}
+
+	private String getActionKillsText(Skill skillName)
+	{
+
+		return (combat.contains(skillName) && client.getLocalPlayer().getInteracting() != null)
```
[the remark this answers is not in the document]
Jasper-ketelaar: Also might just be me but I think using large ternary statements like this one only adds clutter and it doesn't save that much space from just using an if else block.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -287,4 +296,13 @@ private void drawTooltipIfMouseover(Graphics2D graphics, XpGlobe mouseOverSkill,
...
+				? "Kills left:"
+				: "Actions left:";
+	}
+
```
Jasper-ketelaar: Remove whiteline

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -117,6 +132,60 @@ private int getActionsRemaining()
...
+				killsRemaining = (int) Math.ceil(getXpRemaining() / actionExp);
+			}
+		}
+
+		return killsRemaining;
+	}
+
+	private double getCombatXPModifier(Client client)
+	{
+		final double longRangedXPModifier = 2.0;
```
Jasper-ketelaar: These can be constants

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -74,9 +78,15 @@
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
+	private final Map<String, Integer> oppInfoHealth = OpponentInfoPlugin.loadNpcHealth();
```
Jasper-ketelaar: I'm pretty sure you're not supposed to call methods from other plugins as they're designed not to rely on each other.

Jasper-ketelaar: Besides that, you're not using oppInfoHealth in this class.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -74,9 +78,15 @@
...
+
+	private static final List<Skill> COMBAT = Arrays.asList(Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE, Skill.RANGED, Skill.HITPOINTS);
+
+	private int killsRemaining = Integer.MAX_VALUE;
```
Jasper-ketelaar: Not being used

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -74,9 +78,15 @@
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
+	private final Map<String, Integer> oppInfoHealth = OpponentInfoPlugin.loadNpcHealth();
+
+	private static final List<Skill> COMBAT = Arrays.asList(Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE, Skill.RANGED, Skill.HITPOINTS);
```
Jasper-ketelaar: Can just put one list instance in the main plugin class and give it package access level

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -72,6 +74,8 @@
 
 	private static final int TOOLTIP_RECT_SIZE_X = 150;
 
+	private static final ImmutableList<Skill> combat = ImmutableList.of(Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE, Skill.RANGED, Skill.HITPOINTS);
```
Jasper-ketelaar: You're missing magic

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -175,7 +185,10 @@ private void rebuildAsync(boolean updated, XpSnapshotSingle xpSnapshotSingle)
 			// Update information labels
 			expGained.setText(htmlLabel("XP Gained: ", xpSnapshotSingle.getXpGainedInSession()));
 			expLeft.setText(htmlLabel("XP Left: ", xpSnapshotSingle.getXpRemainingToGoal()));
-			actionsLeft.setText(htmlLabel("Actions: ", xpSnapshotSingle.getActionsRemainingToGoal()));
+			actionsLeft.setText((COMBAT.contains(skill) && client.getLocalPlayer().getInteracting() != null)
```
Jasper-ketelaar: Refer to my previous comment about using getInteracting

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -27,18 +27,32 @@
...
+import net.runelite.client.plugins.opponentinfo.OpponentInfoPlugin;
 
 @Slf4j
 @RequiredArgsConstructor
 class XpStateSingle
 {
 	private final Skill skill;
 
+	private final Map<String, Integer> oppInfoHealth = OpponentInfoPlugin.loadNpcHealth();
+	private static final List<Skill> COMBAT = Arrays.asList(Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE, Skill.RANGED, Skill.HITPOINTS);
```
Jasper-ketelaar: Once again should just keep one list not three lists.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -53,6 +67,7 @@
 	private int[] actionExps = new int[10];
 	private int actionExpIndex = 0;
 
+
```
Jasper-ketelaar: Remove whiteline

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -117,6 +132,60 @@ private int getActionsRemaining()
...
 	}
 
+	private int getKillsRemaining(Client client)
+	{
+		Player local = client.getLocalPlayer();
+		if (local == null)
+		{
+			return Integer.MAX_VALUE;
+		}
+		Actor opponent = local.getInteracting();
```
Jasper-ketelaar: Once again this would only work during combat and as soon as you're out of combat it would stop working.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -247,15 +249,22 @@ private void drawTooltipIfMouseover(Graphics2D graphics, XpGlobe mouseOverSkill,
...
+			}
 			if (actionsLeft != Integer.MAX_VALUE)
 			{
 				String actionsLeftString = decimalFormat.format(actionsLeft);
 				xpTooltip.getChildren().add(LineComponent.builder()
-					.left("Actions left:")
-					.leftColor(Color.ORANGE)
-					.right(actionsLeftString)
-					.build());
+						.left(getActionKillsText(selectedSkill))
```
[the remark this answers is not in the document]
IanMcNelly: Had a bit of trouble with the rebase, I'll take care of the whitespace issues

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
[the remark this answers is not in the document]
IanMcNelly: I like that idea, I'll take a look at the idle notifier for combat and rework to be more similar to that. One thing though is that during testing it stayed on kills remaining even when switching styles (and so skills), so I think it might only be refreshed when that particular skill gains xp. As for timeout, is there a standard amount of time? or something like 30 seconds would be fine?

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
IanMcNelly: I definitely need opponentinfo in the xpstatesingle, is there a way to declare that it has a dependency on that plugin? Can definitely remove this one from here though

NathenSample: Under no circumstances should you rely on the internal logic of another plugin, and I'm gradually updating all plugin access modifiers to make this impossible.

If you find yourself logically needing something another plugin already uses internally, then move it out of the plugins to a more logical location within the client and reference it from there.

Nightfirecat: FYI, I'm reworking some of the opponent health info plugin as part of #3939, and will absolutely be making this method private.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -91,6 +96,10 @@ XpTrackerConfig provideConfig(ConfigManager configManager)
 		return configManager.getConfig(XpTrackerConfig.class);
 	}
 
+	@Getter(AccessLevel.PACKAGE)
+	public static final List<Skill> COMBAT = Arrays.asList(Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE, Skill.RANGED, Skill.HITPOINTS, Skill.MAGIC);
```
Adam-: Make this private or just package protected with no getter. Also use `ImmutableList` or set.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -163,6 +253,17 @@ else if (durationHours > 0)
 		return String.format("%02d:%02d", durationMinutes, durationSeconds);
 	}
 
+	private static Map<String, Integer> loadNpcHealth()
```
Adam-: If npc health is going to be required for >1 plugin it should be moved somewhere central (`net.runelite.client.game`?) - or perhaps exposed via a service in the oppoinfo plugin.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
Adam-: This is a valid point. Also the `XPStateSingle` should not be responsible for calculating something like this in real time. This logic should be within the xptracker plugin. Maybe the best event to use for this would be `NpcLootReceived` since you can kill things you are not interacting with (cannon?). Get the health of the NPC which died and calculate it into the running average.

LeviSchuck: I am not too comfortable with so much logic here. This is meant to be a domain object that stores mostly data.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -163,6 +253,17 @@ else if (durationHours > 0)
...
+	{
+		Gson gson = new Gson();
+		Type type = new TypeToken<Map<String, Integer>>()
```
LeviSchuck: please don’t make new Gson instances for every function call

Is there one that can be dependency injected?

This type should also be done once instead if each function call.

Also IO really does not belong in this class

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -254,6 +355,7 @@ XpSnapshotSingle snapshot()
 			.actionsInSession(actions)
 			.actionsRemainingToGoal(getActionsRemaining())
```
deathbeam: Actions remaining and actions/hr should simply return kills remaining and kills/hr when needed. same with actions in session.

PR #3555, opened 2018-06-01
Add reset others to XpInfoBox right click menu

evansloan:
[Preview](https://user-images.githubusercontent.com/8287750/40848319-db9bb74a-658c-11e8-8831-ad2de9996f64.gif)

PR #4019, opened 2018-06-26
Add "Reset all" to XP tracker plugin

Shingyx:
Similar to https://github.com/runelite/runelite/pull/3555 but resets everything instead. Useful for tracking exp rates when I'm training combat (which affects multiple skills) and I'm trying to compare different approaches.


Jasper-ketelaar:
We already have this on the total xp counter.

Shingyx:
I'm clearly blind. Thanks!

PR #4026, opened 2018-06-26
Add temporary measure to fix the alkharid course.

### LINKED ISSUE #4020: Agility Plugin - Al Kharid's First Tightrope Counts as Lap
| Question    | Answer
| ------------| ---------------
| Java Version | 1.8.0_171-b11
| Operating System | Microsoft Windows 10 Pro (Version 1803, OS Build 17134.112)

**Description:** When successfully crossing the first tightrope (_TIGHTROPE_10284_) on the Al Kharid agility course, it counts as a completed lap on the _Total Laps_ counter. The counter does increment normally when jumping the final gap (_GAP_10352_). This causes the addition of two total laps during one full lap, or one lap if a later obstacle is failed.

**Reproduction:** Cross the first tightrope on the Al Kharid agility course.

ScottMCarr:
Added a temporary fix for the Al Kharid agility course. A better solution would be to check the player has started the lap and then check the total experience gained matches the lap total. This would, however, change the behavior should they run the course after logging in.

Fixes #4020

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -106,6 +106,7 @@ public class AgilityPlugin extends Plugin
 
 	private int lastAgilityXp;
 	private WorldPoint lastArenaTicketPosition;
+	private boolean pastTightrope;
 
 	@Provides
 	AgilityConfig getConfig(ConfigManager configManager)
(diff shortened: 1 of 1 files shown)


Adam-:
I would maybe just store the position of the end of the course and check that in addition.

Kamielvf:
This way would break when logging in on top of the course as well

ScottMCarr:
Ah didn't think of that. That's very true, I guess Adam's solution is the best way to go about it.

Issue #3920, opened 2018-06-20
Auto-completion of highscore lookup


Twiglet1022:
| Question    | Answer
| ------------| ---------------
| Java Version | Java Version 8 Update 171
| Operating System | Windows 10

When typing a name into the highscore search bar it will find a name from your friends list and place the rest of the letters from their name after what you have typed. After typing a second letter the letters from auto-complete remain there, tagged onto the end of what you just typed.

GIF demonstration below

![runelive-v1 4 3 2-highscore-search](https://user-images.githubusercontent.com/29353990/41688740-389ada40-74e5-11e8-9d4c-04fb9dfd36c7.gif)

pettenge:
I'm able to reproduce this when I type two letters in quick succession. Is that what you did here?

pettenge:
I think the problem is that calling `SwingUtilities.invokeLater` is causing a race condition somewhere down the line. I'm going to look into this.

Twiglet1022:
It didn't seem to have anything to do with that, no. I can't actually reproduce it now. It's as if the client was in a glitched state and after I restarted it again it was fine. Not sure what initially caused it, but I didn't need to type two characters quickly to get the result I did there. In fact you can see after type the first letter it duplicates the auto-fill twice. Looks like the same thing as issue #4154

pettenge:
I wonder if the key listener could somehow be getting added twice?

Adam-:
Fixed in 30c3ce9dc17f799aec59a177d82360becf6a664b

PR #4030, opened 2018-06-27
3920 hiscores lookup bug

ScottMCarr:


Adam-:
I cannot reproduce the original issue of #3920

ScottMCarr:
I could not either, so I believe it might be system dependent. This should fix the issue as it's shown in the image however without changing the functionality for anyone else as well.

pettenge:
I still run into the bug with this change if I type characters in quick succession (basically pressing the keys the same type).

Adam-:
Fixed in 30c3ce9dc17f799aec59a177d82360becf6a664b

Issue #3974, opened 2018-06-23
Add shift-click menu swapping to spellbook


Jalopyy:
IE click = tele to GE, shift-click = varrock center

deathbeam:
https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

honeyhoney:
Is this really the same as below? (I assume this is why you closed this request)
```Spellbook Re-ordering: Other 3rd party clients do not have this feature and it failed a Jagex poll (see #1038)```

As far as I know this is regarding the menu options such as:
![image](https://user-images.githubusercontent.com/11140337/41810307-51821580-76f4-11e8-9cf0-b2fbdbc08e55.png)
Instead of having to configure which one is left click (i.e. Varrock or Grand Exchange) you would simply be able to have left click Varrock and shift+left click Grand Exchange.

Adam-:
Added in #10039

PR #4031, opened 2018-06-27
Add feature #3974 to the menu entry swapper plugin.

ScottMCarr:


Nightfirecat:
Just as a bit of feedback for these various PRs you've opened, I would recommend a better format for your commit titles and messages.

`"Add feature #3974 to the menu entry swapper plugin"` doesn't really explain what you changed or why.

You could change it to something like the following instead to be much more clear about what you changed:

```
Add shift-click menu swapping for teleport spells

This change adds a configuration option to the menu entry swapper plugin
to swap the position of the normal spellbook teleports with their
alternate locations when Shift is held.

Closes runelite/runelite#3974
```

In general, I always recommend following the guidelines in the [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) article if the project does not have an explicit guideline for formatting commits.

ScottMCarr:
Thank you for the criticism. I will try to be better at writing commit messages from now on.

Adam-:
Added in #10039

Issue #3960, opened 2018-06-22
Clear freeze/entangle timers on movement


Nightfirecat:
When a freeze timer is added by an entity which despawns or runs outside of interaction range, the timer will remain even after the freeze expires. Since we do not have a way of tracking the source of a freeze, timers could be reset upon local player movement.

Edge cases:

* If you teleport while frozen, but are still within range of the freeze's source, you are still frozen
* Planting mithril seeds will cause your character to remain frozen and still move one tile
* Being hit by a dragon spear special attack will cause your character to remain frozen and still move one tile
* ???

ScottMCarr:
I believe the only way to fix this would be to add a hook for the players' movement and check that they are moving with it while also not having the dspear animation present.

Nightfirecat:
Fixed in 720c5f018465f4e854dd0da8b056f735542c0b12:

https://github.com/runelite/runelite/blob/89b8bc52ca28d445d52747bc34ef7cbc477a80f6/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java#L515-L521

PR #4033, opened 2018-06-27
Issue #3960 - Clear freeze/entangle timers on movement

ScottMCarr:
Addresses issue #3960 while leaving possible edge cases open. Would need to have the ability to check the actual players movement either by event or the actor class and implementing it with that.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -26,8 +26,8 @@ package net.runelite.client.plugins.timers;
 
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
+import java.util.function.Predicate;
 import javax.inject.Inject;
-
 import net.runelite.api.Actor;
 import net.runelite.api.AnimationID;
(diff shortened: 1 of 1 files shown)


Infinitay:
You'll have to revert your import changes as we don't allow wildcard imports

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -578,6 +581,30 @@ public void onItemContainerChanged(ItemContainerChanged itemContainerChanged)
 		}
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick tick)
+	{
+		if (!config.showFreezes())
```
Nightfirecat: This seems to be reversed? Should be `if (config.showFreezes())`

Also, would be better as a guard clause:

```
if (!config.showFreezes())
{
    return;
}
```

Nightfirecat:
Fixed in 720c5f018465f4e854dd0da8b056f735542c0b12:

https://github.com/runelite/runelite/blob/89b8bc52ca28d445d52747bc34ef7cbc477a80f6/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java#L515-L521

PR #4048, opened 2018-06-27
Allow alt search while grand exchange inventory is active

### LINKED ISSUE #4040: Alt lookup for items doesn't work when the grand exchange offer interface is open.

jplsek:
Rewrote the logic to use menu entries instead of purely a mouse listener. This simplifies the code and fixes several bugs.

Fixes #4040

Some other things fixed now:
- Spam clicking the alt search on an item no longer crashes the applet (however, the exception is still thrown)
- The applet no longer loses focus when using alt search, which caused the tooltip to go away (and may have been the reason for the crash?)
- Now works in the shop interface


jplsek:
Yep, turns out this is a problem for other plugins, like the examine plugin. Gonna try to see if making a class to catch these "edge" cases can be applied to these plugins.

Kamielvf:
Why is this whole plugin using a mouse listener? You should be able to do all this using the menu entries

Issue #4054, opened 2018-06-28
Change the Decimal and Thousand Separators to the UK/US format


Saer-Dev:
Currently, the Decimal Separator is in the non-UK/US format (where the Decimal Separator is a comma - [see image](https://prnt.sc/k0ekvs)).

**Is your feature request related to a problem? Please describe.**
It's not a problem as such, but may be considered confusing to some UK/US users.  Especially considering that it is not the format used in OSRS itself.  **EDIT:** From reading chatter regarding this in the Discord, I've found out that the formatting is inconsistent, with the ground items using the UK/US format.  So that could be considered a _problem_.

**Describe the solution you'd like**
I would prefer that this is switched to the UK/US format (where the Decimal Separator is a full-stop/period and the Thousand Separator is a comma) to bring it in line with the format used by OSRS.

**Describe alternatives you've considered**
I would also be happy to have a toggle put in place to allow non-UK/US users to maintain the current format.

**Additional context**
- [Image linked at the top of the issue](https://prnt.sc/k0ekvs)
- [Wikipedia article regarding the Decimal Separator](https://en.wikipedia.org/wiki/Decimal_separator)

NathenSample:
We should probably be using a generic formatter thoughout, such as `Stackformatter` (With a more generic name to represent that it isn't just for formatting stacks)

deathbeam:
Fixed in #4059

PR #4059, opened 2018-06-28
#4054 Represent decimals as .'s not ,'s

NathenSample:
Addresses #4054 in that it represents stacks with decimals as . rather than ,

I believe the other formatters format logically, since they don't maintain a decimal, and instead produce output like `100,000`

Test updated to reflect the `10.5k` format rather than `10,5k`

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
@@ -147,7 +147,7 @@ public class StackFormatter
 	 */
 	public static String quantityToRSDecimalStack(int quantity)
 	{
-		if (quantity == 0 || String.valueOf(quantity).length() <= 4)
+		if (String.valueOf(quantity).length() <= 4)
 		{
 			return String.valueOf(quantity);
 		}
--- a/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.util;
 
+import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParseException;
 import java.util.regex.Matcher;
@@ -50,6 +51,10 @@ public class StackFormatter
 	 */
 	private static final NumberFormat NUMBER_FORMATTER = NumberFormat.getInstance();
 
+	/**
+	 * A decimal number formatter
+	 */
+	private static final NumberFormat DECIMAL_FORMATTER = new DecimalFormat("#,###.#");
 	/**
 	 * Convert a quantity to a nicely formatted stack size.
 	 * See the StackFormatterTest to see expected output.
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -50,6 +51,10 @@
 	 */
 	private static final NumberFormat NUMBER_FORMATTER = NumberFormat.getInstance();
 
+	/**
+	 * A decimal number formatter
+	 */
+	private static final NumberFormat DECIMAL_FORMATTER = new DecimalFormat("#,###.#");
 	/**
```
deathbeam: Add empty line above

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -133,54 +138,22 @@ else if (quantity < 10_000_000)
...
 	 */
 	public static String quantityToRSDecimalStack(int quantity)
 	{
-
-		if (quantity < 10_000)
-		{
-			return Integer.toString(quantity);
-		}
-		else if (quantity < 1_000_000)
+		if (String.valueOf(quantity).length() <= 4)
```
deathbeam: Dont use String.valueOf twice

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -133,54 +139,23 @@ else if (quantity < 10_000_000)
...
-		{
-			if (quantity % 1_000_000_000 == 0)
-			{
-				return quantity / 1_000_000_000 + "B";
-			}
-			return NUMBER_FORMATTER.format(quantity).substring(0, Integer.toString(quantity).length() - 7) + "B";
+			return quantityStr;
 		}
+
+		int power = (int) StrictMath.log10(quantity);
```
Adam-: Is it necessary to use `StrictMath` here?

NathenSample: That being said I'll update the PR with `Math` when I get home tonight, as there is no scenario in which we need `StrictMath`s strict floating point accuracy

NathenSample:
`StrictMath` updated to `Math` - Since the use-case doesn't call for floating point accuracy

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -133,54 +139,23 @@ else if (quantity < 10_000_000)
...
-			if (quantity % 1_000_000_000 == 0)
-			{
-				return quantity / 1_000_000_000 + "B";
-			}
-			return NUMBER_FORMATTER.format(quantity).substring(0, Integer.toString(quantity).length() - 7) + "B";
+			return quantityStr;
 		}
+
+		int power = (int) Math.log10(quantity);
+		return DECIMAL_FORMATTER.format(quantity / (Math.pow(10, (power / 3) * 3))) + SUFFIXES[power / 3];
```
reubent: `SUFFICES` surely, @NathenSample?

Issue #4061, opened 2018-06-28
Inventory Tags do not take quantity based icons into consideration


NathenSample:
For items which have a inventory icon which varies depending on stack size inventory tags will outline based upon the 1 stack size, rather than taking current stack size into consideration
![runelite_2018-06-28_22-11-35](https://user-images.githubusercontent.com/9803552/42060887-410ceeca-7b20-11e8-8c3e-c2e66ef27b1d.png)

Adam-:
Fixed in 2e336cd5d193e2a790cd892e01589c1fe9fcc4a5

PR #4062, opened 2018-06-28
inventory tags: Fix outlines for stackable items

Nightfirecat:
Fixed inventory tags for stackable items

Because the stored outlines were saved only once for each item, items
with different sprites for different quantities such as coins, ranged
ammunition, seeds, etc. would only show the outline of the first
quantity displayed while tagged. This adds the quantity to the key used
to store cached outlines.

Fixes runelite/runelite#4061


deathbeam:
This is also already fixed in #4000 i believe

PR #4065, opened 2018-06-28
agility plugin: Fix Al Kharid course lap counter

Nightfirecat:
Fixes runelite/runelite#4020

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -26,6 +26,7 @@ package net.runelite.client.plugins.agility;
 
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Working on updating to use a stored `WorldPoint[]` in the courses enum, but don't have it quite working yet.

Nightfirecat:
Updated PR to store a `WorldPoint[]` with each course of possible ending locations and check that on XP drops matching the last obstacle XP drop amount.

Adam-:
I would probably only want the end point added to the alkharid course since that is the only course with this issue... maybe unless youve tested each course?

Nightfirecat:
I have tested each course. Each course has only one end point except the following:

* The gnome course has two pipes for the final obstacle, each will complete a lap
* Varrock's course allows you to jump down each tile of a 5-tile-wide end obstacle
* The penguin course grants the lap experience when opening the door, which means you can be on either the tile inside or outside the door depending on where your character was standing when clicking on the door
* The wilderness course has three rock climb obstacle options for the final obstacle

Issue #3736, opened 2018-06-10
Show Runecrafting altars on the map


Gamer1120:
It would be nice to know where the Runecrafting altars are when looking at the minimap. I suggest placing a marker on each mysterious ruins to indicate which altar it is.

Abextm:
Why?

Gamer1120:
Why not? I mean, there is no reason for there not to be markers on the altars.

Abextm:
Because its something we would have to write, maintain, and provide support for.

Gamer1120:
Alright. Feel free to close if you believe this adds nothing to the client.

PR #4083, opened 2018-06-29
World Map: RC Altar Locations

basicDamen:
Adds RC Altar Locations to World Map plugin with corresponding rune icons. Requested in #3736.
air_altar_demo


deathbeam:
These world map icons should probably be in similar style as other world map icons (e.g in the stylish circle)

basicDamen:
If someone wants to make new icons, they can.. but.. going any smaller than that and stuffing it into a circle won't look too good.. At least the plugin will be there and someone would just need to update the png files. @deathbeam

Nightfirecat:
@basicDamen You may want to check what has been done to make the rune icons for the rune pouch overlay. I agree with deathbeam, using the 3d icon looks very out of place.

PR #4088, opened 2018-06-29
Return preferredSize.width from TitleComponent

deathbeam:
To return parent defined dimensions properly, return preferredSize.width
instead of text width, as this is block component so it should act like
block and not return custom sizes.


Nightfirecat:
LGTM

PR #4097, opened 2018-06-30
Add API to make AccountType from varbit and name

deathbeam:
- Revert "Add ironman status and AccountType to API"
- Add API to make AccountType from varbit and name
- Use this new API in mixins and in ChatCommandsPlugin


deathbeam:
Fixed in #4098

PR #4098, opened 2018-06-30
 chat commands: fix to use exiting accounttype api 

Adam-:


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -432,19 +431,34 @@ private static String sanitize(String lookup)
...
 	/**
 	 * Returns the ironman status based on the symbol in the name of the player.
 	 * @param name player name
 	 * @return hiscore endpoint
 	 */
-	private static HiscoreEndpoint getIronmanStatusByName(final String name)
+	private static HiscoreEndpoint getHiscoreEndpointByName(final String name)
 	{
-		return toEndPoint(AccountType.fromName(name));
+		if (name.contains("<img=2>"))
```
deathbeam: This does not need to be chain of ifs as every condition has early return

Issue #4103, opened 2018-06-30
Skill icon scaling in XP Globes


hydrarguru:
**_Hello!_**
I really like how I can easily scale down or up the size of the experience globes in RuneLite, I'm often using a smaller size. A problem I'm seeing with smaller sizes is that the skill icon starts to clip out of the globe when the globe size is around 31 and smaller, I would love to see the skill icon to scale with the size of the globe or that I can change the skill icon size separately.

_this is the globe with a size of 40._
![bild](https://user-images.githubusercontent.com/17914439/42124088-16502628-7c5d-11e8-90f8-95b6c5aaf539.png)

_this is a globe that is size 25:_
![bild](https://user-images.githubusercontent.com/17914439/42124107-56f361f4-7c5d-11e8-8fea-eb885a0292f9.png)

I don't think this should be hard to implement but I do not have any experience in Java.

Cheers
hydrarcurum

PR #4112, opened 2018-07-01
Add XP_Globe - Dynamic icon sizing

J-Leg:
PR for #4103.
UPDATE:
_All commits squashed. Tested and all seems good._

Add Dynamic sizing of xpglobe icons using Swing.Util.resize(). 
BufferedImages of icons are stored in instances of xpGlobes.
XpGlobe - icon ratio used for scaling is based on the defaults.

~~Added a new Maven dependency ```Thumbnailator``` (apparently this is faster than Java's ```Image.getScaledInstance```). I adjusted the Globe to Skill_icon ratio slightly.~~

_globe size 40_
screenshot from 2018-07-01 11-56-23

_globe size 25_
screenshot from 2018-07-01 11-57-47


Kamielvf:
There's a few things I'd like to mention:
- Most likely shouldn't need a new maven dependency
- You would only have to scale the images once, after the size has changed in config, and then probably save those scaled images inside of the plugin
- Below a certain size it can probably use the small skill images

Just my thoughts on this

J-Leg:
@Kamielvf For the small icon set, I don't think there is much reason to use it (besides the negligible performance benefits of scaling an image with less pixels) since the plugin allows a globe size of anything above 0.

_This is with the small_icon_set, and I suppose nobody would actually use a globe size this small, but if they did, scaling again would be necessary_
screenshot from 2018-07-01 13-21-33
_scaled_
screenshot from 2018-07-01 13-22-58

LeviSchuck:
I agree that a dependency here is not needed, the skills don't need to be re-scaled on every frame as they do not change. I hope that the circle progress bar is drawn after scaling, not before. 
(Will review code)

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -191,21 +196,52 @@ private Ellipse2D drawEllipse(Graphics2D graphics, int x, int y)
...
-		if (skillImage == null)
+		BufferedImage toDraw;
+		if (xpGlobe.getSkillIcon() != null && xpGlobe.getGlobeSize() == config.xpOrbSize())
 		{
-			return;
+			toDraw = xpGlobe.getSkillIcon();
+		}
+		else
+		{
+			toDraw = generateScaledIcon(xpGlobe.getSkill());
```
LeviSchuck: This doesn't need to be done on every frame.

J-Leg: Currently this else block will only execute if ```xpGlobe``` is not set to anything, or if the globe size has recently been changed (requires scaling again).

After which, the scaled image is saved in the xpGlobe instance, so the next time we need to draw the icon, it will just set ```toDraw = xpGlobe.getSkillIcon()```, rather than generating a new one.

J-Leg:
I'll get rid of ```thumbnailtor``` now.

Abextm:
fwiw Thumbnailinator is just a (overly complex) wrapper around java2d, so you can just use `SwingUtil.resizeImage` for exactly the same thing.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -23,12 +23,7 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 package net.runelite.client.plugins.xpglobes;
-import java.awt.BasicStroke;
-import java.awt.Color;
-import java.awt.Dimension;
-import java.awt.Graphics2D;
-import java.awt.RenderingHints;
-import java.awt.Stroke;
+import java.awt.*;
```
Nightfirecat: Your IDE settings probably did this, but project code convention is to generally avoid wildcard imports.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -227,18 +220,17 @@ private BufferedImage generateScaledIcon(Skill skill)
...
-				.size((int)(config.xpOrbSize() * GlOBE_HEIGHT_RATIO), (int)(config.xpOrbSize() * GLOBE_WIDTH_RATIO))
-				.asBufferedImage();
-			return scaled;
-		}
-		catch (Exception e)
-		{
-			log.debug("Unable to scale skill icon {}", e);
-		}
+
+		int height = (int)(config.xpOrbSize() * GlOBE_HEIGHT_RATIO);
```
Nightfirecat: final

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -227,18 +220,17 @@ private BufferedImage generateScaledIcon(Skill skill)
...
+		int width = (int)(config.xpOrbSize() * GLOBE_WIDTH_RATIO);
```
Nightfirecat: final

pom.xml
```diff
@@ -154,7 +154,7 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
-		</dependencies>
+        </dependencies>
```
ScottMCarr: Spaces are used and also it's an artifact left over from your previous dependency addition.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -191,21 +197,56 @@ private Ellipse2D drawEllipse(Graphics2D graphics, int x, int y)
...
+		}
+		else
+		{
+			toDraw = generateScaledIcon(xpGlobe.getSkill());
+			if (toDraw == null)
+			{
+				return;
+			}
+			xpGlobe.setSkillIcon(toDraw);
+			xpGlobe.setGlobeSize(config.xpOrbSize());
```
ScottMCarr: Couldn't this just be moved inside the else condition above?

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -191,21 +197,56 @@ private Ellipse2D drawEllipse(Graphics2D graphics, int x, int y)
...
-			x + (config.xpOrbSize() / 2) - (skillImage.getWidth() / 2),
-			y + (config.xpOrbSize() / 2) - (skillImage.getHeight() / 2),
+			toDraw,
+			x + (config.xpOrbSize() / 2) - (toDraw.getWidth() / 2),
+			y + (config.xpOrbSize() / 2) - (toDraw.getHeight() / 2),
 			null
 		);
 	}
 
+	private BufferedImage generateScaledIcon(Skill skill)
```
ScottMCarr: Seems like this can just be an Image class removing the need to create a BufferedImage and drawing to it.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobe.java
```diff
@@ -36,6 +37,8 @@
 	private int goalXp;
 	private Instant time;
 	private double skillProgressRadius;
+	private BufferedImage skillIcon;
+	private int globeSize;
```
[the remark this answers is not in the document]
J-Leg: I thought it would be better to keep it consistent with the rest of the file. I guess I can do use them for the others too.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
Abextm: It is preferred to draw to a BufferedImage because it only has to scale once. Simply returning the result of `getScaledInstance` makes it rescale everytime the image is drawn.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -191,21 +197,56 @@ private Ellipse2D drawEllipse(Graphics2D graphics, int x, int y)
...
+
+		final int height = (int)(config.xpOrbSize() * GlOBE_HEIGHT_RATIO);
+		final int width = (int)(config.xpOrbSize() * GLOBE_WIDTH_RATIO);
+
+		if (width == 0 || height == 0)
+		{
+			return null;
+		}
+
+		Image toolkit = icon.getScaledInstance(width, height, Image.SCALE_REPLICATE);
```
Abextm: This should just be a call to SwingUtil.resizeImage, rather than copying the code here

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
J-Leg: I'm a bit confused what you mean by this. It is part of the else condition, and is only set if the image is properly created.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
ScottMCarr: Didn't realize that was the behavior, that makes sense then.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
J-Leg: Is this method supposed to be private? I have changed it to public.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
ScottMCarr: Oh my, I read that wrong because of the above comment. Sorry about that!

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobe.java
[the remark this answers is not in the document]
ScottMCarr: If the whole file uses getter and setter methods then you could leave it the way it is, but I do know Lombok is used in the project so it might better be changed to be consistent among the project as a whole.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
J-Leg: nws :smile: 
Always good to double check things anyway

LeviSchuck:
I would suggest you update the main PR description and squash all your commits down to one.

J-Leg:
Will do that now.

PR #4114, opened 2018-07-01
Add pause skill timing feature to the XP Tracker

### LINKED ISSUE #3113: XP tracker: only use in-game time for ETAs
If I let runelite stay on overnight while I am gone, when I am back the next day and **continue** smithing, it has really bad estimates for time til next level/goal since it includes all the time I was gone from the game.

Suggested solution: use game ticks to add time to each skill, and only if the delta between ticks is within say 10 seconds (to account for connectivity?).

I plan to address this one myself, making an issue to document my intent.

### LINKED ISSUE #3598: Missing pause button?
It's not a problem we just need pause and play button for skills just like konduits features. I was only trying to take break to eat my lunch and want my grind to stay. But couldn't.

A play and a pause button to stop calcuating.

N/A 

**Additional context**
 
image

LeviSchuck:
Add the ability to manually pause a skill, automatically pause a skill after logout, or after a timeout of no XP changes. Similarly, skills will be auto unpaused when XP is gained or when (if paused due to logout) logged back in.

## Motivation
Some players would like to pause their expectations on skill ETAs when having lunch, going on a break, or sleeping for the night. By only counting effective time in game, players don't need to account for time outside of game or between skills when training.

## Configuration changes
image

Added a setting for pausing skills without any XP changes, default value: 0 which behaves as no auto pausing by timeout
Added a setting for auto pausing skills in relation to logout, default value: Never

## Changes to default behavior
No changes to default behavior

## UI Changes
ProgressBar now extends a new Component class which comes with dimming of the foreground and background color.
ProgressBar will also dim the left and right label, with the expectation that the center label be used for an alternate display text.
image

Within the XP Tracker panel, right clicking on a skill will bring up a "Pause" menu item.
image
When a skill is paused, it will be labeled as "Unpause"

## Example
logout-pausing

Resolves #3598 
Resolves #3113

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
@@ -54,6 +54,15 @@ import net.runelite.client.util.SwingUtil;
 @Slf4j
 class XpInfoBox extends JPanel
 {
+	// Templates
+	private static final String HTML_TOOL_TIP_TEMPLATE =
+		"<html>%s actions done<br/>"
+			+ "%s actions/hr<br/>"
+			+ "%s till goal lvl</html>";
+	private static final String HTML_LABEL_TEMPLATE =
+		"<html><body style='color:%s'>%s<span style='color:white'>%s</span></body></html>";
+
+	// Instance members
 	private final JPanel panel;
 
 	@Getter(AccessLevel.PACKAGE)
@@ -74,6 +83,9 @@ class XpInfoBox extends JPanel
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
+	private final JMenuItem pauseSkill = new JMenuItem("Pause");
+
+	private boolean paused = false;
 
 	XpInfoBox(XpTrackerPlugin xpTrackerPlugin, Client client, JPanel panel, Skill skill, SkillIconManager iconManager) throws IOException
 	{
@@ -98,12 +110,16 @@ class XpInfoBox extends JPanel
 		final JMenuItem resetOthers = new JMenuItem("Reset others");
 		resetOthers.addActionListener(e -> xpTrackerPlugin.resetOtherSkillState(skill));
 
+		// Create reset others menu
+		pauseSkill.addActionListener(e -> xpTrackerPlugin.pauseSkill(skill, !paused));
+
 		// Create popup menu
 		final JPopupMenu popupMenu = new JPopupMenu();
 		popupMenu.setBorder(new EmptyBorder(5, 5, 5, 5));
 		popupMenu.add(openXpTracker);
 		popupMenu.add(reset);
 		popupMenu.add(resetOthers);
+		popupMenu.add(pauseSkill);
 
 		JLabel skillIcon = new JLabel(new ImageIcon(iconManager.getSkillImage(skill)));
 		skillIcon.setHorizontalAlignment(SwingConstants.CENTER);
@@ -138,6 +154,7 @@ class XpInfoBox extends JPanel
 		progressBar.setMaximumValue(100);
 		progressBar.setBackground(new Color(61, 56, 49));
 		progressBar.setForeground(SkillColor.values()[skill.ordinal()].getColor());
+		progressBar.setDimmedText("Paused");
 
 		progressWrapper.add(progressBar, BorderLayout.NORTH);
 
@@ -157,12 +174,12 @@ class XpInfoBox extends JPanel
 		panel.revalidate();
 	}
 
-	void update(boolean updated, XpSnapshotSingle xpSnapshotSingle)
+	void update(boolean updated, boolean paused, XpSnapshotSingle xpSnapshotSingle)
 	{
-		SwingUtilities.invokeLater(() -> rebuildAsync(updated, xpSnapshotSingle));
+		SwingUtilities.invokeLater(() -> rebuildAsync(updated, paused, xpSnapshotSingle));
 	}
 
-	private void rebuildAsync(boolean updated, XpSnapshotSingle xpSnapshotSingle)
+	private void rebuildAsync(boolean updated, boolean skillPaused, XpSnapshotSingle xpSnapshotSingle)
 	{
 		if (updated)
 		{
@@ -171,6 +188,8 @@ class XpInfoBox extends JPanel
 				panel.add(this);
 				panel.revalidate();
 			}
+
+			paused = skillPaused;
 
 			// Update information labels
 			expGained.setText(htmlLabel("XP Gained: ", xpSnapshotSingle.getXpGainedInSession()));
@@ -181,28 +200,42 @@ class XpInfoBox extends JPanel
 			progressBar.setValue(xpSnapshotSingle.getSkillProgressToGoal());
 			progressBar.setCenterLabel(xpSnapshotSingle.getSkillProgressToGoal() + "%");
 			progressBar.setLeftLabel("Lvl. " + xpSnapshotSingle.getStartLevel());
-			progressBar.setRightLabel("Lvl. " + (xpSnapshotSingle.getEndLevel()));
+			progressBar.setRightLabel("Lvl. " + xpSnapshotSingle.getEndLevel());
 
-			progressBar.setToolTipText("<html>"
-				+ xpSnapshotSingle.getActionsInSession() + " actions done"
-				+ "<br/>"
-				+ xpSnapshotSingle.getActionsPerHour() + " actions/hr"
-				+ "<br/>"
-				+ xpSnapshotSingle.getTimeTillGoal() + " till goal lvl"
-				+ "</html>");
+			progressBar.setToolTipText(String.format(
+				HTML_TOOL_TIP_TEMPLATE,
+				xpSnapshotSingle.getActionsInSession(),
+				xpSnapshotSingle.getActionsPerHour(),
+				xpSnapshotSingle.getTimeTillGoal()));
+
+			progressBar.setDimmed(skillPaused);
 
 			progressBar.repaint();
 		}
+		else if (!paused && skillPaused)
+		{
+			// React to the skill state now being paused
+			progressBar.setDimmed(true);
+			progressBar.repaint();
+			paused = true;
+			pauseSkill.setText("Unpause");
+		}
+		else if (paused && !skillPaused)
+		{
+			// React to the skill being unpaused (without update)
+			progressBar.setDimmed(false);
+			progressBar.repaint();
+			paused = false;
+			pauseSkill.setText("Pause");
+		}
 
-		// Update exp per hour seperately, everytime (not only when there's an update)
+		// Update exp per hour separately, every time (not only when there's an update)
 		expHour.setText(htmlLabel("XP/Hour: ", xpSnapshotSingle.getXpPerHour()));
 	}
 
 	static String htmlLabel(String key, int value)
 	{
 		String valueStr = StackFormatter.quantityToRSDecimalStack(value);
-
-		return "<html><body style = 'color:" + SwingUtil.toHexColor(ColorScheme.LIGHT_GRAY_COLOR) + "'>" + key + "<span style = 'color:white'>" + valueStr + "</span></body></html>";
+		return String.format(HTML_LABEL_TEMPLATE, SwingUtil.toHexColor(ColorScheme.LIGHT_GRAY_COLOR), key, valueStr);
 	}
-
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2018, Levi <me@levischuck.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.xptracker;
+
+import java.util.EnumMap;
+import java.util.Map;
+import net.runelite.api.Skill;
+
+class XpPauseState
+{
+	// Internal state
+	private final Map<Skill, XpPauseStateSingle> skillPauses = new EnumMap<>(Skill.class);
+	private boolean cachedIsLoggedIn = false;
+
+	boolean pauseSkill(Skill skill)
+	{
+		return findPauseState(skill).manualPause();
+	}
+
+	boolean unpauseSkill(Skill skill)
+	{
+		return findPauseState(skill).unpause();
+	}
+
+	boolean isPaused(Skill skill)
+	{
+		return findPauseState(skill).isPaused();
+	}
+
+	void tickXp(Skill skill, int currentXp, int pauseAfterMinutes)
+	{
+		final XpPauseStateSingle state = findPauseState(skill);
+
+		if (state.getXp() != currentXp)
+		{
+			state.xpChanged(currentXp);
+		}
+		else if (pauseAfterMinutes > 0)
+		{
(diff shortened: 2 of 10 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -74,6 +74,12 @@
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
+	private final JMenuItem openXpTracker = new JMenuItem("Open online tracker");
```
deathbeam: Any reason why those were extracted from constructor when they are used only there?

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
```diff
@@ -0,0 +1,154 @@
+package net.runelite.client.plugins.xptracker;
```
deathbeam: License header

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseStateSingle.java
```diff
@@ -0,0 +1,72 @@
+package net.runelite.client.plugins.xptracker;
```
deathbeam: License header

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseStateSingle.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.time.Instant;
+import java.util.EnumSet;
+import java.util.Set;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import net.runelite.api.Skill;
+
+
```
deathbeam: Remove this empty line

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseStateSingle.java
```diff
@@ -0,0 +1,72 @@
...
+	{
+		return pauseReasons.add(XpPauseReason.PAUSED_TIMEOUT);
+	}
+
+	public boolean onManualPause()
+	{
+		return pauseReasons.add(XpPauseReason.PAUSE_MANUAL);
+	}
+
+	public boolean onXpChanged()
```
deathbeam: Is onXpChanged and onUnPause really necessary as they are same methods? I would just expose clearAll

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -0,0 +1,35 @@
+package net.runelite.client.plugins.xptracker;
```
deathbeam: License header

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+	@ConfigItem(
+		position = 0,
+		keyName = "pauseSkillAfter",
+		name = "Auto Pause minutes",
+		description = "Configures how many minutes passes before pausing a skill while in game and there's no XP"
+	)
+	default int pauseSkillAfter()
+	{
+		return 60;
```
deathbeam: By default this should never pause

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpAutoPauseLogoutBehavior.java
```diff
@@ -0,0 +1,20 @@
+package net.runelite.client.plugins.xptracker;
+
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public enum XpAutoPauseLogoutBehavior
+{
+	NEVER("Never"),
+	ON_LOGOUT("On Logout"),
+	AFTER_LOGOUT_1("Logout + 1m"),
```
deathbeam: Is this option really useful?

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
```diff
@@ -0,0 +1,154 @@
...
+
+	public boolean isPaused(Skill skill)
+	{
+		return findPauseState(skill).isPaused();
+	}
+
+	// Private internal functions
+	private void tickXp(Instant now)
+	{
+		final int pauseAfterSeconds = Math.max(60, config.pauseSkillAfter() * 60);
```
deathbeam: Why is this limited at 60 seconds?

runelite-client/src/main/java/net/runelite/client/ui/components/DimmableJPanel.java
```diff
@@ -0,0 +1,82 @@
+package net.runelite.client.ui.components;
```
deathbeam: License header

runelite-client/src/main/java/net/runelite/client/ui/components/DimmableJPanel.java
```diff
@@ -0,0 +1,82 @@
...
+			super.setBackground(dimmedBackground);
+			super.setForeground(dimmedForeground);
+		}
+		else
+		{
+			super.setBackground(undimmedBackground);
+			super.setForeground(undimmedForeground);
+		}
+		// Allow user behavior here
+		onDimmedChanged(dimmed);
```
deathbeam: I dont think this is really necessary, as user using this class can just override this method and call super.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
LeviSchuck: No functional reason at least.

When I added one of the menu items to it, I felt it was inconsistent to have only some of the menu items be inside the constructor and some outside.

runelite-client/src/main/java/net/runelite/client/ui/components/DimmableJPanel.java
LeviSchuck: In this case, I am hiding that I deduplicate the dim state changes. I don't want extenders to keep redundant state.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
Jasper-ketelaar: Also does this mean it can't be turned off? You have to accept that your xp counter is going to pause its timer after an hour?

LeviSchuck: It's a lower bound of 60 seconds, since the user can select `0` as their config value, which isn't valid.

LeviSchuck: As stated in the PR description, it is a change to default behavior that skills will be automatically paused if they don't receive any XP for a whole hour.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpAutoPauseLogoutBehavior.java
LeviSchuck: It seemed like a compromise for discussions on issues and in discord I think.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
deathbeam: I think you should just make the 0 valid value and make it default to keep current master behaviour and behaviour of all other clients.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseStateSingle.java
LeviSchuck: I've simplified a few things in the parent `XpPauseState` class, this function will have slightly different behavior in the next revision.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
LeviSchuck: Changing to 0, which will act as if disabled.
Will update PR document accordingly.

LeviSchuck:
Changes addressed for @deathbeam

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
deathbeam: I believe best practice is to reduce scope and visibility as much as possible and even when it is inconsistent, this is doing opposite

runelite-client/src/main/java/net/runelite/client/ui/components/DimmableJPanel.java
deathbeam: I see, I dont think that should really be concern here or anywhere else as the setBack/Foreground is not expensive at all so even you dont need to handle this state

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
LeviSchuck: Reverted.

LeviSchuck:
Style changes made.

Can squash when everything looks good.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -70,17 +76,43 @@
...
+	}
+
+	@Provides
+	XpState provideInternalState()
+	{
+		return new XpState();
+	}
+
+	@Provides
+	XpPauseState provideXpPauseState(XpTrackerConfig xpTrackerConfig, Client client)
```
deathbeam: I dont think this is correct. The XpPauseState looks like it should not need to have access to entire client and entire config. I would rather just propagate to it what it really needs and not use guice for creating it at all

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -70,17 +76,43 @@
...
 
+
+	@Provides
+	XpTrackerConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(XpTrackerConfig.class);
+	}
+
+	@Provides
+	XpState provideInternalState()
```
deathbeam: Why are you using provider for this when it has no dependencies? You can just construct it and send it around

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -401,4 +426,51 @@ private static VarPlayer endGoalVarpForSkill(final Skill skill)
...
+		final Instant now = Instant.now();
+		xpPauseState.tick(now);
+
+		if (lastTick == null)
+		{
+			lastTick = now;
+			return;
+		}
+
+		final Duration tickDuration = Duration.between(lastTick, now);
```
deathbeam: I dont really like this ticking and sending diffs between ticks here. Cant you just send actual time and handle it properly?

runelite-client/src/main/java/net/runelite/client/ui/components/DimmableJPanel.java
```diff
@@ -0,0 +1,93 @@
...
+			super.setForeground(dimmedForeground);
+		}
+		else
+		{
+			super.setBackground(undimmedBackground);
+			super.setForeground(undimmedForeground);
+		}
+	}
+
+	public boolean isDimmed()
```
deathbeam: Use lombok

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
LeviSchuck: Ultimately all calculations are based on a `Duration`.

If I want to pause things between skills independently, then I can't pass time, I have to pass durations.

LeviSchuck: An alternate way would be to pass milliseconds, so `Duration` is out of the picture entirely except for string formatting.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
LeviSchuck: I need:
1. Client game state
2. Access to all current skills XP values
3. Configuration settings on pausing (That's currently the only thing in the configuration)

LeviSchuck: An alternate route is to include these on the tick function, rather than at construction.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
LeviSchuck: By bringing this up, I noticed a bug where I forgot to update the lastTick to now.
That would have been embarrassing.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
LeviSchuck: I'll simplify these so construction time requirements is no longer necessary.

LeviSchuck:
Changes addressed for @deathbeam

LeviSchuck:
Rebased, altered again towards @deathbeam 's wishes.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -74,6 +74,15 @@
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
+	private final JMenuItem pauseSkill = new JMenuItem("Pause");
+
+	private boolean paused = false;
+	private static final String HTML_TOOL_TIP_TEMPLATE =
```
deathbeam: these private static final fields should probably be at the top of the class

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
```diff
@@ -0,0 +1,130 @@
...
+{
+	// Internal state
+	private final Map<Skill, XpPauseStateSingle> skillPauses = new EnumMap<>(Skill.class);
+	private boolean cachedIsLoggedIn = false;
+
+	/**
+	 * Tick the pause status for all skills, taking into account skill no-xp timeouts
+	 * and logout pause settings
+	 */
+	public void tick(Client client, XpTrackerConfig config)
```
deathbeam: I would probably send the skill experience here and config.pauseSkillAfter directly

deathbeam: And probably move some of the tickLogout logic to tickXp and just call onLogout directly when game state changes to LOGIN_SCREEN (and rename the onLogout() to just logout(), probably)

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
```diff
@@ -0,0 +1,130 @@
...
+			if (state.getXp() != currentXp)
+			{
+				state.onXpChanged(currentXp, now);
+			}
+			else
+			{
+				final Instant lastChange = state.getLastChange();
+				// When config.pauseSkillAfter is 0, it is effectively disabled
+				if (pauseAfterMinutes > 0 && lastChange != null
+					&& Duration.between(lastChange, now).getSeconds() >= pauseAfterSeconds)
```
deathbeam: Instead of creating duration all the time you can probably just compare milliseconds.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -229,15 +228,14 @@ boolean update(int currentXp, int goalStartXp, int goalEndXp)
...
-		// If this is first time we are updating, we just started tracking
-		if (skillTimeStart == null)
-		{
-			skillTimeStart = Instant.now();
-		}
-
 		return true;
 	}
 
+	public void tick(long millis)
```
deathbeam: I would probably rename this millis to `delta`

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
LeviSchuck: I personally don't like renaming these methods, because they don't **cause** logout, they react to it.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("xpTracker")
+public interface XpTrackerConfig extends Config
+{
+	@ConfigItem(
+		position = 0,
+		keyName = "logoutPausing",
+		name = "Logout Pausing",
+		description = "Configures whether skills should unpause on logout"
```
Adam-: Do  you mean "paused on logout"?

LeviSchuck: You are right.

LeviSchuck:
Squashed into two commits, one for sharable UI changes, one for the pausing feature.
@Adam- @deathbeam

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPauseState.java
```diff
@@ -0,0 +1,107 @@
...
+			final long lastChangeMillis = state.getLastChangeMillis();
+			// When config.pauseSkillAfter is 0, it is effectively disabled
+			if (lastChangeMillis != 0 && (now - lastChangeMillis) >= pauseAfterMillis)
+			{
+				state.timeout();
+			}
+		}
+	}
+
+	void tickLogout(boolean pauseOnLogout, GameState gameState)
```
deathbeam: You should probably just send loggedIn as param here instead of GameState. But having method named tickLogout that handles if player is logged in or not is slightly confusing imo.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("xpTracker")
+public interface XpTrackerConfig extends Config
+{
+	@ConfigItem(
+		position = 0,
+		keyName = "logoutPausing",
+		name = "Logout Pausing",
```
Adam-: "Pause on Logout" maybe?

Adam-:
If you enable pause on logout while logged out it doesn't pause.

PR #4116, opened 2018-07-01
Xerics talisman plugin

tstaples:
This is a pretty simple plugin that displays the actual teleport location for the different options on the Xeric's talisman (ie. "Hosidius House" instead of "Xeric's Glade").

The purpose of this plugin is to avoid needing to remember or lookup where all the different options go.

There are currently two config options (with pretty horrible names):
* **hideOriginalTeleportName** - the original location text will be hidden and only the 'friendly' name will be shown (while respecting original formatting). If this option is false (default) the 'friendly' name will be shown in parenthesis after the original text.
* **hideMenuActionTargetForTeleports** - every menu option in a right-click menu (aside from cancel) is suffixed with the item's name. This option removes the item name from the teleport options to prevent the tooltip from being massive.

Here is an image showing both the right-click menu and the 'rub' menu with the different config variations:
config variations

I don't write much Java so if I'm doing anything weird or there are nicer ways to do things let me know :)


runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -102,6 +102,7 @@
 	public static final int MTA_TELEKINETIC_GROUP_ID = 198;
 	public static final int CORP_DAMAGE = 13;
 	public static final int DESTROY_ITEM_GROUP_ID = 584;
+	public static final int XERICS_TALISMAN_MENU_GROUP_ID = 187;
```
Abextm: Is this interface only used for the xeric's talisman? I would think it is the same one as the spirit tree one.

tstaples: Ah you're correct, it is used for the spirit tree as well. What is a more appropriate generic name for that group ID?

Nightfirecat: Along those same lines, I wonder if the wilderness obelisk menu (after Hard wildy diary) uses the same widget. Maybe `SCROLL_TELEPORT_MENU_GROUP_ID`?

deathbeam:
I suggested this on the Pharaoh sceptre plugin, but I can mention it also here, but maybe we could have some "Menu Entry Renamer" plugin where stuff like this would live

tstaples:
@deathbeam That's a good idea. I'll refactor this to fit that approach.

raiyni:
This is covered by the Easy Teleports hub plugin now.

Issue #3102, opened 2018-05-22
Reorder XP Tracker bars


kittyk4y:
Maybe drag and drop or group by skill type along with the option to group it by the current default?

deathbeam:
Duplicate of #2029

Issue #783, opened 2018-03-02
Add the ability to reorder the sidebar tabs


rune3132:
I feel like the logout tab should be at the bottom, other people might want another order. Being able to reorder it would be nice.

ShaunDreclin:
duplicate of #582

deathbeam:
As Shaun said

rune3132:
From my understanding, that issue is requesting the reordering of the ingame tiles. I would like to reorder the clients tabs, like the Xp Tracker, Info, Kurend Library plugin and so on.

deathbeam:
Ah, yes, sorry, misunderstood the issue. Calling button tab is very confusing :( Anyway, for the login/logout problem, there is #646 PR open currently.

ShaunDreclin:
Ohh I misunderstood too, yeah being able to re-order the client's tab strip would be nice.

Possibly related to that, #449

rune3132:
No worries, i could have been more precise when describing my request. Fel free to reword anythign if there is some kind of standart that i didn't follow :)

yoghurtis:
Yes please!

RickertdeH:
This would be useful, especially with all the new external plugins available.

wigglepop:
![image](https://user-images.githubusercontent.com/77034846/103752787-1e861000-5045-11eb-8ed3-8466aa8ca986.png)
if the ability to re-arrange sidebars is not going to be put, please add a scrolling ability for those who don't play full screen. it's hard to put the client to full screen again and again just so we could open up a plugin that's at the bottom of the list

jevonsp:
Is this ever going to be revisited? with the plugin hub it is a larger annoyance

mooseweener:
Agreed, this needs to be revisited. I play full screen and I just hit the max icons.

jonathanjosephduarte:
scrolling or the ability to rearrange the icons would be excellent!

Issue #2029, opened 2018-04-27
Able to re-arrange skills in the skill tracker tab. 


Stickaplex:
Currently if you switch to training a different skill the new skill will appear at the bottom of the list. Would be nice if you could drag the skills around to have it in the order you want. Or honestly however is easiest to do as long as you can re-arrange them. I just think a dragging mechanic there would be easiest

PR #4118, opened 2018-07-01
Reorder xp tracker bars

Shingyx:
Implement a custom MouseAdapter which watches for mouse drag events to move child components within a JPanel. Include a workaround for child popup menus, since setComponentPopupMenu consumes the required MouseEvents. Also remove redundant line in `XpInfoBox#reset`.

Resolves https://github.com/runelite/runelite/issues/3102 and https://github.com/runelite/runelite/issues/2029. There could be others too, I didn't look that hard.

The mouse adapter should be generic enough so that it could be applied to some of the other plugins in the future.


Shingyx:
I think I've caught most of the edge cases. Would be interested in your thoughts now. Honestly, I think the best way to review this is pull it down locally and experiment with it. Swing doesn't make drag and drop easy, unless I completely missed something in my Google searches.

Gif demonstrating the feature:
drag

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -69,13 +69,14 @@
...
 	private final ProgressBar progressBar = new ProgressBar();
+	private final JPopupMenu popupMenu = new JPopupMenu();
 
 	private final JLabel expGained = new JLabel();
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
 
-	XpInfoBox(XpTrackerPlugin xpTrackerPlugin, Client client, JPanel panel, Skill skill, SkillIconManager iconManager) throws IOException
+	XpInfoBox(XpTrackerPlugin xpTrackerPlugin, Client client, XpInfoBoxPanel panel, Skill skill, SkillIconManager iconManager) throws IOException
```
Shingyx: Any reason why this constructor declares `throws IOException`? I can't see anything that could throw one.

Shingyx:
I don't know of any more issues from testing on Windows. However, I tried this out on Ubuntu (and Kubuntu) and it works great in most cases, except for one issue I noticed where it can get into a weird state when trying to drag an info panel while a pop up menu for another info panel is open, where the dragging panel gets "stuck" when the mouse is released. This is because the `MouseReleased` event doesn't seem to get received when this happens. The overall UI recoverable by clicking somewhere else in the XP plugin panel.

On that note, should the reordering be configurable? I don't know how I'd workaround this issue...

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -69,13 +70,14 @@
...
 	private final ProgressBar progressBar = new ProgressBar();
+	private final JPopupMenu popupMenu = new JPopupMenu();
 
 	private final JLabel expGained = new JLabel();
 	private final JLabel expHour = new JLabel();
 	private final JLabel expLeft = new JLabel();
 	private final JLabel actionsLeft = new JLabel();
 
-	XpInfoBox(XpTrackerPlugin xpTrackerPlugin, Client client, JPanel panel, Skill skill, SkillIconManager iconManager) throws IOException
+	XpInfoBox(XpTrackerPlugin xpTrackerPlugin, Client client, JComponent panel, Skill skill, SkillIconManager iconManager) throws IOException
```
Shingyx: Does `throws IOException` need to be here?

Shingyx:
Rebased and squashed.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,302 @@
...
+import javax.swing.JLayeredPane;
+
+/**
+ * Pane which allows reordering its components via drag and drop.
+ * <p>
+ * For child components' popup menus, try avoid using setComponentPopupMenu as this consumes the MouseEvents required
+ * for drag and drop reordering. Instead, implement the PopupMenuOwner interface in the children components to enable
+ * both child popup menus and drag and drop reordering.
+ */
+public class DragAndDropReorderPane extends JLayeredPane
```
Nightfirecat: If this is going to be a public class, this should not be introduced as a plugin class. This should be added with the other UI elements in runelite/runelite-client/src/main/java/net/runelite/client/ui/.

Shingyx: Done. I moved both DragAndDropReorderPane and PopupMenuOwner. `net.runelite.client.ui.components` seemed like the most appropriate package.

Shingyx:
I've rebased again. Is there anything else I can do to help move this PR forwards? It's been almost a year now.

Personally, the XP tracker is my most used feature of RuneLite and I've always found it annoying when the only way to reorder the tracker bars is to reset the ones which I want to move to the bottom, especially since this also resets the XP rate calculation.

I understand that a lot of code was added, but Swing's mouse event handling made adding drag-and-drop while maintaining popup menus difficult.

Nightfirecat:
I'll try to give this a review this weekend. I appreciate you've kept this PR up to date, and I'm sure we can find some time to properly review this and get it merged.

Shingyx:
That's awesome to hear! Please let me know if there are any issues or if anything needs clarification.

Thanks!

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * Pane which allows reordering its components via drag and drop.
+ * <p>
+ * For child components' popup menus, try avoid using setComponentPopupMenu as this consumes the MouseEvents required
+ * for drag and drop reordering. Instead, implement the PopupMenuOwner interface in the children components to enable
+ * both child popup menus and drag and drop reordering.
+ */
+public class DragAndDropReorderPane extends JLayeredPane
+{
+	private Point dragStartPoint = null;
```
Nightfirecat: Non-primitive fields are initialized to `null` by default; there is no need to explicitly set them.
```suggestion
	private Point dragStartPoint;
```

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,250 @@
...
+import java.awt.event.MouseEvent;
+import java.util.HashMap;
+import java.util.Map;
+import javax.swing.BoxLayout;
+import javax.swing.JLayeredPane;
+
+/**
+ * Pane which allows reordering its components via drag and drop.
+ * <p>
+ * For child components' popup menus, try avoid using setComponentPopupMenu as this consumes the MouseEvents required
```
Nightfirecat: This doesn't feel like entirely the right place for this comment. This should probably be moved to the PopupMenuOwner interface so it can be more easily found.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -197,7 +199,6 @@
 	void reset()
 	{
 		canvasItem.setText(ADD_STATE);
-		container.remove(statsPanel);
```
Nightfirecat: Why is this line removed?

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,250 @@
...
+	private void moveDraggingComponent(Point point)
+	{
+		// place the center of the dragging component onto the mouse cursor
+		int y = point.y - draggingComponent.getHeight() / 2;
+		// clamp the height to stay within the pane
+		y = Math.max(y, 0);
+		y = Math.min(y, getHeight() - draggingComponent.getHeight());
+
+		Point location = new Point(0, y);
+		draggingComponent.setLocation(location);
```
Nightfirecat: Remove the `location` variable above and set the location with the new Point directly.
```suggestion
		draggingComponent.setLocation(new Point(0, y));
```

Nightfirecat:
Also, it looks like the second commit can be squashed into the first? This change looks like it can be contained in one commit.

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
Shingyx: Removed

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
Shingyx: `statsPanel` was moved to be inside `headerPanel` at some point, so this line had no effect.

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
Shingyx: Done

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
Shingyx: I've split the comment out a bit. Is it better now?

Shingyx:
> 
> 
> Also, it looks like the second commit can be squashed into the first? This change looks like it can be contained in one commit.

I've squashed and rebased

raiyni:
You don't need to comment on every request. Just click the mark as resolved button.

Nightfirecat:
Code LGTM now, but I'll freely admit that I'm not well-versed in Swing, so some other contributors should probably weigh in on that. Thanks for responding so quickly to feedback. :+1:

raiyni:
1. If this doesn't save the order of the panels, what is the point?
2. This maybe could be used to resolve #783 if implemented correctly

Shingyx:
Personally, I don't think that saving the order of the panels between sessions is necessary for this to be useful. One example use case: I want to keep all XP trackers active for the whole session, so that I can see how much XP I gained in each skill at the end, but at the same time I want the skill I'm actively working on to be at the top.

https://github.com/runelite/runelite/issues/783 looks out of scope to me. I think the component I added could be used for that though.

Shingyx:
I've rewritten basically half of this PR for the following:
* Removed the pop-up menu hacks since I've found a relatively clean way to forward mouse drag events to parent components. Replaced with a `MouseDragEventForwarder` instead.
* Improved how the XP info box follows the mouse, avoiding the "jump" when the dragging starts.

Please let me know if there's anything else I can do to move this forward. The linked issue has been duplicated several times now, suggesting that some people are interested in this feature.

Shingyx:
Rebased and updated gif.

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,203 @@
...
+
+		@Override
+		public void mouseDragged(MouseEvent e)
+		{
+			if (SwingUtilities.isLeftMouseButton(e) && dragStartPoint != null)
+			{
+				Point point = e.getPoint();
+				if (!contains(point))
+				{
+					finishDragging();
```
deathbeam, line 181: Imo this should not stop dragging when OOB. Mainly when OOB means the size of parent container what is size of the actual listed xp trackers, so the dragging just stops there and it is pretty annoying.

Shingyx: Yeah that makes sense. Updated.
sample4

abextm:
Why doesn't this use swing's drag and drop stuff?

Shingyx:
There were limitations with all of the native Swing implementations I could find. None of them were designed for **reordering** components (like the `XpInfoBox`) inside a `JComponent`/`JPanel` with drag and drop. The best examples I found involved writing a custom implementation similar to what I have done here.

There is a native drag and drop implementation for [JList](https://docs.oracle.com/javase/7/docs/api/javax/swing/JList.html#setDragEnabled(boolean)), but rewriting the XP panel to be a `JList` holding `XpInfoBox` components doesn't sound like a good idea. The other [Swing drag and drop stuff](https://docs.oracle.com/javase/tutorial/uiswing/dnd/intro.html) I could find mainly involved dragging and dropping **data** in the application.

So from what I could find, [JLayeredPane](https://docs.oracle.com/javase/7/docs/api/javax/swing/JLayeredPane.html) is the correct choice here. It defines a `DRAG_LAYER` layer for rendering components being dragged around. I admit I'm not the best at Swing, so there could be _other_ ways to implement this, but IMO this already works really well as-is.

runelite-client/src/main/java/net/runelite/client/ui/components/DragAndDropReorderPane.java
```diff
@@ -0,0 +1,199 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.ui.components;
+
+import javax.swing.BoxLayout;
+import javax.swing.JLayeredPane;
+import javax.swing.SwingUtilities;
```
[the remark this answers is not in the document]
Shingyx: I must've wrote that before configuring IntelliJ properly 😅

PR #4119, opened 2018-07-01
Change Mahogany Tree growth time from 106.6 hours to 85.3 hours.

brett-taylor:
Current mahogany tree time is incorrect and out by around a day changed the expected growth time to match the wiki from 106 hours to 85.


Adam-:
Fixed in #4125

PR #4122, opened 2018-07-01
Use any installed font with runelite (client/overlays/tooltips)

l2-:
Defaults back to runescape small if the selected font in the config wasn't found on the system.

The loading of the config now happens before the clientUI.init() so that the set font in the config can be applied to the UI. The client will need to be restarted after changing the client font to have effect. I tried to work around this and make it work live but it resulted in a web of component.repaint mess so i stuck with the restart req.

Had to change the text positioning of text in some components because of incorrect font metrics use but this should work for any font.
I have tested this with some fonts but there are many to test so feedback always welcome.

I know y'all like your runescape font but I think it's awesome to use any desired font :)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -65,7 +65,7 @@
 	// so we replace any item quantity higher with "Lots" instead.
 	private static final int MAX_QUANTITY = 65535;
 	// The 15 pixel gap between each drawn ground item.
-	private static final int STRING_GAP = 15;
+	private static final int STRING_GAP = 18;
```
deathbeam: This should stay at original 15 gap.

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
```diff
@@ -36,37 +39,53 @@
 	private static final Font runescapeSmallFont;
 	private static final Font runescapeBoldFont;
 
+	// By using a 2way map we're not dependant on the toString() method so we have control over the font naming in the UI.
+	private static final HashMap<String, Font> fontLookupMap;
```
deathbeam: These both fields should be initialized here

deathbeam: I dont think 2 maps are necessary at all for this

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
```diff
@@ -93,4 +112,14 @@ public static Font getRunescapeBoldFont()
 	{
 		return runescapeBoldFont;
 	}
+
+	public static LinkedHashMap<Font, String> getFontMap()
```
deathbeam: You probably should not expose entire font map

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -381,18 +382,22 @@ private void safeRender(Client client, Overlay overlay, OverlayLayer layer, Grap
...
+		// Set font based on configuration
+		if (position == OverlayPosition.TOOLTIP)
 		{
-			subGraphics.setFont(runeLiteConfig.tooltipFontType().getFont());
+			subGraphics.setFont(runeLiteConfig.tooltipFont());
 		}
 		else
 		{
-			subGraphics.setFont(FontManager.getRunescapeFont());
+			subGraphics.setFont(runeLiteConfig.overlayFont());
```
deathbeam: Why was support for changing font for dynamic overlays separately removed? It should be put back

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -381,18 +382,22 @@ private void safeRender(Client client, Overlay overlay, OverlayLayer layer, Grap
...
 
-		// Set font based on configuration
-		if (position == OverlayPosition.DYNAMIC)
+		// Set font rendering properties like the OS's font rendering
+		Toolkit tk = Toolkit.getDefaultToolkit();
+		Map desktopHints = (Map)(tk.getDesktopProperty("awt.font.desktophints"));
+		if (desktopHints != null)
 		{
-			subGraphics.setFont(runeLiteConfig.fontType().getFont());
+			graphics.addRenderingHints(desktopHints);
```
deathbeam: This should be done for root graphics and not here

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/InfoBoxComponent.java
```diff
@@ -38,7 +38,6 @@
 @Setter
 public class InfoBoxComponent implements LayoutableRenderableEntity
 {
-	private static final int SEPARATOR = 3;
```
deathbeam: This was fixed-size separator from bottom, why was it removed?

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TooltipComponent.java
```diff
@@ -104,7 +104,7 @@ public Dimension render(Graphics2D graphics)
 					textComponent.setColor(nextColor);
 					String text = line.substring(begin, j);
 					textComponent.setText(text);
-					textComponent.setPosition(new Point(lineX, textY + (i + 1) * textHeight - textDescent));
+					textComponent.setPosition(new Point(lineX, textY + (i + 1) * metrics.getMaxAscent() + i * metrics.getMaxDescent()));
```
deathbeam: You are positioning the components differently as the original tooltip height is calculated, use same logic in both places

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TooltipComponent.java
```diff
@@ -141,7 +141,7 @@ else if (subLine.startsWith("img="))
 						textComponent.setColor(nextColor);
 						String text = line.substring(begin, j + 1);
 						textComponent.setText(text);
-						textComponent.setPosition(new Point(lineX, textY + (i + 1) * textHeight - textDescent));
+						textComponent.setPosition(new Point(lineX, textY + (i + 1) * metrics.getMaxAscent() + i * metrics.getMaxDescent()));
```
deathbeam: See my previous comment

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TooltipComponent.java
```diff
@@ -155,7 +155,7 @@ else if (subLine.startsWith("img="))
 			final TextComponent textComponent = new TextComponent();
 			textComponent.setColor(nextColor);
 			textComponent.setText(line.substring(begin, line.length()));
-			textComponent.setPosition(new Point(lineX, textY + (i + 1) * textHeight - textDescent));
+			textComponent.setPosition(new Point(lineX, textY + (i + 1) * metrics.getMaxAscent() + i * metrics.getMaxDescent()));
```
deathbeam: See my previous comment

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
```diff
@@ -36,37 +39,53 @@
...
 					FontManager.class.getResourceAsStream("runescape_bold.ttf"))
 					.deriveFont(Font.PLAIN, 16);
-			ge.registerFont(boldFont);
 
-			runescapeBoldFont = StyleContext.getDefaultStyleContext()
-					.getFont(boldFont.getName(), Font.PLAIN, 16);
+			fontMap = new LinkedHashMap<>();
+			fontLookupMap = new HashMap<>();
+
+			fontMap.put(runescapeSmallFont, "rs small");
```
deathbeam: Why are these short abbreviations used? just use normal name

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -189,28 +191,6 @@ default boolean sendNotificationsWhenFocused()
 		return false;
 	}
 
-	@ConfigItem(
-		keyName = "fontType",
-		name = "Dynamic Overlay Font",
-		description = "Configures what font type is used for in-game overlays such as player name, ground items, etc.",
-		position = 30
-	)
-	default FontType fontType()
```
deathbeam: FontType looks unused now

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -243,4 +223,37 @@ default int infoBoxSize()
...
+
+	@ConfigItem(
+		keyName = "overlayFont",
+		name = "Overlay Font",
+		description = "Configures what font type is used for in-game overlays such as player name, ground items, etc.",
+		position = 31
+	)
+	default Font overlayFont()
+	{
+		return FontManager.getRunescapeSmallFont();
```
deathbeam: Shouldnt this use new API?

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -243,4 +223,37 @@ default int infoBoxSize()
...
+
+	@ConfigItem(
+		keyName = "tooltipFont",
+		name = "Tooltip Font",
+		description = "Configures what font type is used for in-game tooltips such as food stats, NPC names, etc.",
+		position = 31
+	)
+	default Font tooltipFont()
+	{
+		return FontManager.getRunescapeSmallFont();
```
deathbeam: Shouldnt this use new API?

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -243,4 +223,37 @@ default int infoBoxSize()
...
+
+	@ConfigItem(
+		keyName = "clientFont",
+		name = "Client Font",
+		description = "Configures what font type is used for the client",
+		position = 30
+	)
+	default Font clientFont()
+	{
+		return FontManager.getRunescapeFont();
```
deathbeam: See my previous comment

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
```diff
@@ -36,37 +39,53 @@
...
+
+			// Get all available fonts on the system
+			Font[] availableFonts = ge.getAllFonts();
+			//build bidirectional map
+			Arrays.stream(availableFonts).sorted(Comparator.comparing(Font::getFontName)).forEach(f ->
+			{
+				if (!fontLookupMap.containsKey(f.getFontName()) && !fontMap.containsKey(f))
+				{
+					// I believe font size 11 or 12 is for many fonts about the same size as font size 16 for the rs fonts
+					f = f.deriveFont(11f);
```
deathbeam: you should never use uneven size as default, mainly when there is no other option. Also i believe for a lot of monospace fonts they will be pretty small with this

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
l2-: Mainly because of fitting inside the drop down but alright boss

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
l2-: Im out of the loop. What new API do you mean?

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
deathbeam: Probably, as you are removing the possibility to choose smaller font for tooltips than for panels with this change, what was one of the issues before why the option for dynamic overlays was added.

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/InfoBoxComponent.java
l2-: I thought you came up with this because you didn't seem to use metrics.getdescent which needs to be used to get the base y coordinate for rendering text. I replaced it with metrics.getMaxDescent() so i figured it isn't needed anymore but i can understand if you want a defined seperator (it won't have to be 3 anymore though).

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
deathbeam: Well you are using it just to map font to string, so I guess getFontName(Font) or something like that

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
deathbeam: I found this: https://stackoverflow.com/questions/13236927/sorting-a-guava-bimap

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
l2-: Can you give a source which explains why you never should? For example ms word starts with size 11 calibri for as long as i can remember. Also do you have an example for a monospace font? i realize 11 is not gonna work for every font but default size doesn't work for most fonts so do you have another suggestion?

deathbeam: Because for monospaced fonts they tend to get pretty ugly when they are not using even sizes. E.g try Terminus with 11px and 12px

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
l2-: Yes i think that might work i'll get on that.

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
deathbeam: I meant they can be blurry, its safer to use even number. Maybe 10 then? i dont think size-per-font woud be reasonable. Also, do you have any kind of fallback when font no longer exists in system? because that can easily happen.

l2-: 10 is too small for all. Settings should default back to runescape font but that's only when you open the runelite settings when the set font doesn't exit. There is already a size per font but you can't change it (every rs font is 16 while all other fonts are huge at that point).

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TooltipComponent.java
l2-: i think i got all your point except this one and the 2 similar ones below, can you clarify a bit more what's wrong here?

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
deathbeam: probably somewhere at the top of render method where also antialiasing is set

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/InfoBoxComponent.java
deathbeam: No this was intentional to add fixed padding from bottom (as 2px is border size + 1 px is separator)

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TooltipComponent.java
deathbeam: I mean that textHeight should be dfeined using this logic here

l2-: Unless I misunderstand you, I think that's not the case. Font is drawn from its base y position not from the top or bottom so textHeight is something different than this logic. This logic is for the Font base y.

07Arab:
@l2- could you please fix up the conflicting files? would love for this to be implemented.

l2-:
> @l2- could you please fix up the conflicting files? would love for this to be implemented.

done

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -197,36 +199,47 @@ default boolean sendNotificationsWhenFocused()
...
-		return FontType.SMALL;
+		return FontManager.getRunescapeFont();
 	}
 
 	@ConfigItem(
-		keyName = "interfaceFontType",
-		name = "Interface Overlay Font",
-		description = "Configures what font type is used for in-game interface overlays such as panels, opponent info, clue scrolls etc.",
-		position = 32
+		keyName = "overlayGroup3Font",
```
raiyni: Looks like you need to rebase this config file. Shouldn't need to change every config message and name.

l2-: This is intentional and not due to rebase. The old names weren't really accurate anymore. At first i based the font selection on tooltip positioning but this didn't really work out because 99% has position top-left/dynamic positioning while they definitely should have the option to select different fonts. I generalized it to groups and added descriptions to tell the user which group has what kind of components. Overlay x isn't really i good name, i realize that, but you can only really fit 1 word in the name or it will be cut off in the settings screen.

l2-:
> you can also just add 1 new config option that will fit perfectly into what we have now, something like Base font and default option will be RuneScape. then, you will prechache small, reg and bold version of your system font, and voila, you are done, the entire PR then would be just changing the 3 getters in FontManager, adding 1 config option to RuneLite settings, and loading of the system fonts.

> the 3 dropdowns will stay, and their values too, they will just use regrs, smallrs and boldrs in case the base font is RuneScape, otherwise they would just use normal-sized version of your system font, small-sized one and bold one (or well, FontManager will)

by deathbeam

Most of the PR needs to be rewritten

deathbeam:
Are you planning to work on that still @l2- ? (with way I suggested to do it)

07Arab:
@l2- please revive this plugin <3 @deathbeam method seems more efficient as well and would be great for it to be implemented ^.^

raiyni:
He closed it because of conflict of opinion and didn't want to implement feedback. You are welcome to pull it and make the changes.

PR #4125, opened 2018-07-01
farmingtracker: Correct several patch timings

Abextm:
There were several places where the autogenerated patch implementations would double count a stage, and miss another. PatchImplementationTest now checks for missed stages, excluding the final stage because that is often missed on plants which do not have a "Check-health" stage. There were also a couple produces with incorrect tick counts.

Supersedes #4119

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/farmingtracker/PatchImplementationTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/farmingtracker/PatchImplementationTest.java
@@ -24,6 +24,8 @@
  */
 package net.runelite.client.plugins.farmingtracker;
 
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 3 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/farmingtracker/PatchImplementationTest.java
```diff
@@ -24,16 +24,21 @@
  */
 package net.runelite.client.plugins.farmingtracker;
 
+import java.util.HashMap;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
```
Nightfirecat: Unused import

Abextm: Its used.

Issue #4089, opened 2018-06-29
Notification when Rune Pouch degrades


Gamer1120:
It would be nice to have a notification when a Rune Pouch degrades.

Gamer1120:
I think that it warrants its own notification, since doing it with the chat notifier is not intuitive. I will try to see if there is a message.

Gamer1120:
![image](https://user-images.githubusercontent.com/9074448/42124144-d76ae6e0-7c5d-11e8-8f45-80ec67742408.png)
This is the message in the chat. "Your pouch has decayed through use."

Adam-:
Added in #4155

PR #4155, opened 2018-07-03
Add a notification when a runepouch degrades

Ishadijcks:
Implements #4089

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
@@ -167,7 +167,8 @@ public class RunecraftPlugin extends Plugin
 				bindNeckOverlay.bindingCharges = 17;
 			}
 		}
-		if (config.degradingNotification()) {
+		if (config.degradingNotification())
+		{
 			if (event.getMessage().contains("Your pouch has decayed through use."))
 			{
 				notifier.notify(POUCH_DECAYED_MESSAGE);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
@@ -67,7 +67,8 @@ import net.runelite.client.util.QueryRunner;
 public class RunecraftPlugin extends Plugin
 {
 	private static Pattern bindNeckString = Pattern.compile("You have ([0-9]+) charges left before your Binding necklace disintegrates.");
-	private static final String POUCH_DECAYED_MESSAGE = "Your rune pouch has decayed.";
+	private static final String POUCH_DECAYED_NOTIFICATION_MESSAGE = "Your rune pouch has decayed.";
+	private static final String POUCH_DECAYED_MESSAGE = "Your pouch has decayed through use.";
 	private static final int DESTROY_ITEM_WIDGET_ID = WidgetInfo.DESTROY_ITEM_YES.getId();
 
 	@Getter(AccessLevel.PACKAGE)
@@ -170,16 +171,16 @@ public class RunecraftPlugin extends Plugin
 			{
 				//set it to 17 because this message is triggered first before the above chat event
 				bindNeckOverlay.bindingCharges = 17;
+				return;
 			}
-			return;
 		}
 		if (config.degradingNotification())
 		{
-			if (event.getMessage().contains("Your pouch has decayed through use."))
+			if (event.getMessage().contains(POUCH_DECAYED_MESSAGE))
 			{
-				notifier.notify(POUCH_DECAYED_MESSAGE);
+				notifier.notify(POUCH_DECAYED_NOTIFICATION_MESSAGE);
+				return;
 			}
-			return;
 		}
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
@@ -54,6 +54,7 @@ import net.runelite.api.queries.InventoryItemQuery;
 import net.runelite.api.queries.NPCQuery;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.client.Notifier;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.plugins.Plugin;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -133,36 +138,41 @@ public void updateConfig(ConfigChanged event)
 	@Subscribe
 	public void onChatMessage(ChatMessage event)
 	{
-		if (event.getType() != ChatMessageType.SERVER || !config.showBindNeck())
+		if (event.getType() != ChatMessageType.SERVER && config.showBindNeck())
```
Nightfirecat: Not sure why this was changed in this way. At the very least, a guard clause for non-server messages should still remain:

```
if (event.getType() != ChatMessageType.SERVER)
{
    return;
}
```

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -133,36 +138,41 @@ public void updateConfig(ConfigChanged event)
...
-				|| event.getMessage().contains("dust")
-				|| event.getMessage().contains("smoke")
-				|| event.getMessage().contains("mist"))
+			if (event.getMessage().contains("Your Binding necklace has disintegrated."))
 			{
-				bindNeckOverlay.bindingCharges -= 1;
-				return;
+				//set it to 17 because this message is triggered first before the above chat event
+				bindNeckOverlay.bindingCharges = 17;
 			}
```
Nightfirecat: There should be a `return` statement after the above line.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -133,36 +138,41 @@ public void updateConfig(ConfigChanged event)
...
-
-		if (event.getMessage().contains("Your Binding necklace has disintegrated."))
+		if (config.degradingNotification())
 		{
-			//set it to 17 because this message is triggered first before the above chat event
-			bindNeckOverlay.bindingCharges = 17;
+			if (event.getMessage().contains("Your pouch has decayed through use."))
+			{
+				notifier.notify(POUCH_DECAYED_MESSAGE);
+			}
```
Nightfirecat: There should be a `return` statement after the above line.

Ishadijcks: Done

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -166,13 +171,15 @@ public void onChatMessage(ChatMessage event)
 				//set it to 17 because this message is triggered first before the above chat event
 				bindNeckOverlay.bindingCharges = 17;
 			}
+			return;
```
Nightfirecat: This should be within the above `if (match.find())` block, after the line where `bindNeckOverlay.bindingCharges` is set.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -166,13 +171,15 @@ public void onChatMessage(ChatMessage event)
...
 		}
 		if (config.degradingNotification())
 		{
 			if (event.getMessage().contains("Your pouch has decayed through use."))
 			{
 				notifier.notify(POUCH_DECAYED_MESSAGE);
 			}
+			return;
```
Nightfirecat: Same as other comment, this should be within the pouch decayed `if` block, after the `notifier.notify()` call.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -133,36 +138,48 @@ public void updateConfig(ConfigChanged event)
...
+				//set it to 17 because this message is triggered first before the above chat event
+				bindNeckOverlay.bindingCharges = 17;
+			}
+			return;
+		}
+		if (config.degradingNotification())
 		{
-			//set it to 17 because this message is triggered first before the above chat event
-			bindNeckOverlay.bindingCharges = 17;
+			if (event.getMessage().contains("Your pouch has decayed through use."))
```
Nightfirecat: Use a constant for this string. This should probably be `POUCH_DECAYED_MESSAGE`, and the notification string should probably be something like `POUCH_DECAYED_NOTIFICATION_MESSAGE`.

Ishadijcks: Fair enough, wasn't sure as it wasn't done with the binding necklace.

Gamer1120:
Maybe it's an idea to say which pouch degraded? Like Large or Giant? I don't know whether that is worth the effort.

Ishadijcks:
I'm not sure that is detectable. Also I think there is already a plugin that highlights broken pouches

Gamer1120:
The degraded pouch changes color in inventory, so I'm pretty sure it's possible.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
```diff
@@ -133,36 +139,48 @@ public void updateConfig(ConfigChanged event)
...
+			if (event.getMessage().contains("Your Binding necklace has disintegrated."))
+			{
+				//set it to 17 because this message is triggered first before the above chat event
+				bindNeckOverlay.bindingCharges = 17;
 				return;
 			}
 		}
-
-		if (event.getMessage().contains("Your Binding necklace has disintegrated."))
+		if (config.degradingNotification())
```
NathenSample: `if(config.degradingNotification() && event.getMessage().contains(POUCH_DECAYED_MESSAGE))`

NathenSample:
It is possible, this class already has logic which detects the presence of a broken pouch in your inventory, look at the logic around setting `degradedPouchInInventory` to true to see how it's done. Similar logic can be used to tell the user which of his pouches are decayed.

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/RunecraftPlugin.java
Ishadijcks: Ah yeah I overlooked that during refactoring, good catch

PR #4171, opened 2018-07-05
Add toggle for tile shading in Ground Marker plugin

Matsyir:
Adds a toggle for tile shading in the Ground Marker plugin. I thought it made some things harder to see, particularly the crystal/spike attack on Olm but many other of its abilities too. Without the shade, I will still be able to make use of the highlight without hindering my ability to see everything properly. Initially overloaded renderPolygon() to simply have an optional parameter, but decided to change it to a separate method in order to not ever so slightly slow down every other call to renderPolygon() due to a boolean comparison.

Shaded (current):
image
Unshaded (new toggle):
image

### DIFF
--- a/cache-client/pom.xml
+++ b/cache-client/pom.xml
@@ -29,7 +29,7 @@
 	<parent>
 		<groupId>net.runelite</groupId>
 		<artifactId>runelite-parent</artifactId>
-		<version>1.5.17</version>
+		<version>1.5.18-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>cache-client</artifactId>
--- a/cache-updater/pom.xml
+++ b/cache-updater/pom.xml
@@ -28,7 +28,7 @@
 	<parent>
 		<groupId>net.runelite</groupId>
 		<artifactId>runelite-parent</artifactId>
-		<version>1.5.17</version>
+		<version>1.5.18-SNAPSHOT</version>
 	</parent>
 	
 	<name>Cache Updater</name>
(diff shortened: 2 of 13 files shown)


Nightfirecat:
The contents of `renderPolygon(Graphics2D graphics, Polygon poly, Color color)` should be changed to a one-line call to `renderPolygon(graphics, poly, color, true)`.

Matsyir:
That's the way I had written it the first time around, but I thought it was a little more efficient to separate it so renderPolygon doesn't have to do a boolean comparison if it would always be true, which would be everywhere else it is used currently (which is what I meant in my first commit), but if you'd like me to do that I will. Perhaps I don't know enough about compilers, it may not even use the if in that situation as far as I know.

deathbeam:
Its more about writing duplicate code for no reason, you should avoid doing that.

devLotto:
This should be referred to as ’fill shadow’ instead of ’shading’.

Matsyir:
Everywhere in the code/plugin titles/desc etc it is only referred to as "shade" (basically the verb of shadowing something) never "shading", it seems weird to say "shadow the tile" rather than just "shade the tile" in my opinion

Matsyir:
I remember talking with deathbeam or lotto awhile back and you guys mentioned waiting for the transparency setting to make its way into colors in order to have a more customizable/elegant version of this. I recently noticed the color transparency was added so I'll soon be remaking this by instead adding a 2nd color config for the fill color, and a checkbox to toggle fill, default to what it currently is: filled with transparent black. Let me know if there are any issues with that.

PR #4181, opened 2018-07-05
BankValuePlugin: Add option to display exact value

MagicfTail:


runelite-client/src/main/java/net/runelite/client/plugins/bankvalue/BankTitle.java
```diff
@@ -92,12 +92,26 @@ void update(long gePrice, long haPrice)
 
 		if (config.showGE() && gePrice != 0)
 		{
-			strCurrentTab += " (EX: " + StackFormatter.quantityToStackSize(gePrice) + ")";
+			if (config.showExact())
+			{
+				strCurrentTab += " (EX: " + StackFormatter.formatNumber(gePrice) + ")";
```
Nightfirecat: Since the `(EX: ... )` parts of the string will always be added, they should be added outside of the if statement.

runelite-client/src/main/java/net/runelite/client/plugins/bankvalue/BankTitle.java
```diff
@@ -92,12 +92,26 @@ void update(long gePrice, long haPrice)
...
+				strCurrentTab += " (EX: " + StackFormatter.quantityToStackSize(gePrice) + ")";
+			}
 		}
 
 		if (config.showHA() && haPrice != 0)
 		{
-			strCurrentTab += " (HA: " + StackFormatter.quantityToStackSize(haPrice) + ")";
+			if (config.showExact())
+			{
+				strCurrentTab += " (HA: " + StackFormatter.formatNumber(haPrice) + ")";
```
Nightfirecat: Same as above.

runelite-client/src/main/java/net/runelite/client/plugins/bankvalue/BankValueConfig.java
```diff
@@ -49,4 +49,13 @@ default boolean showHA()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "showExact",
+		name = "Show exact bank value",
+		description = "Show exact bank value")
```
Nightfirecat: `)` on new line.

MagicfTail: I did think it was a bit weird, but it was what all the other configs were doing, so I just went with it. I'll fix the already existing configs too

Nightfirecat: Strange... Must be whoever added this plugin formatted all of them that way. :man_shrugging:

Issue #4174, opened 2018-07-05
!clue should provide feedback rather than failing silenty


NathenSample:
As it stands `ChatCommandPlugins` logic means that if the player is not ranked, or an `IOException` occurs when attempting to get highscore information, the command will fail silently.

This is counter intuitive and leads to support requests from users questioning why their clue plugin is not working.

Suggestion - 
When the player is not ranked, and so the number of clues cannot be determined - Tell them.
If something goes wrong with their connection - Tell them.

The user has inputted a command, and expects text returns from it, this lets them know that the command, and plugin - are indeed working, just something other than the intended execution path is occurring.

Notables -
[`ChatCommandPlugin:386`](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java#L386) - Returns when no value is returned by the endpoint.
[`ChatCommandPlugin:412`](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java#L412) - on io exception we log an error, but don't tell the user

Nightfirecat:
For the record, this affects all chat commands, not just `!clues`.

J-Leg:
I'll have my PR include others too then.

PR #4184, opened 2018-07-05
Add error handling for chat-commands

J-Leg:
PR for #4174 - Providing feedback for users
Add some checks that will notify the user.

_Sanity of the input/output_
screenshot from 2018-07-05 22-30-22

_IO Exception_ 
(Can someone think of a better message to put here?)
screenshot from 2018-07-05 22-28-53


NathenSample:
Not being ranked doesn't necessarily mean you haven't completed any clues, so "none" isn't strictly accurate - Rather the amount you've done simply cannot be determined.

If you were to stick with that logic, none breaks from the rest of the formatting and should instead be 0

There's no need to perform the rank logic if the return has no level, as that means it can't find you on the high scores and inherently have no rank.

J-Leg:
Ok. Made the appropriate changes.

Cases:

- (First 3) The player is not on the HiScores at all, so we cannot determine
- ```!clues master``` When the player is on the HiScores but has not completed any master clues (Hence also unranked)


image

PR #4185, opened 2018-07-05
Add cape rack to menuentryswapper settings.

thesnell:
This is for the Achievement Gallery cape hanger. It swaps Admire for teleport (normal skill capes) and Perks (Max Cape).


Nightfirecat:
This looks like it should be closed by the already-merged #5994

deathbeam:
Yeah this is already added. Or well, the perks isnt, so just rebase and add only that

xDemoN:
The first half of this was done in #5994 as previously mentioned, and the second half was done by db in #6180 - this can be closed

PR #4196, opened 2018-07-06
Prefer Ipv4 for OkHttpClient

deathbeam:
This resolves some issues with people using client and trying to use
Ipv6 instead of Ipv4 what our API does not support. This is alternative
solution to https://github.com/runelite/launcher/pull/24.


deathbeam:
Solved in https://github.com/runelite/launcher/pull/24

PR #4198, opened 2018-07-06
Add Cinematic Mode Plugin

MagicfTail:
Add a "cinematic mode" for taking pictures without the ui and runelite overlays. The plugin only works in resizable mode.


Kamielvf:
Isn't this what the orb of occulus is for? (aside from our own overlays)

NathenSample:
The orb isn't freely available in OSRS

MagicfTail:
^You can more or less recreate the orb of occulus using this and the detached camera dev tool

PR #4205, opened 2018-07-07
Follow GitHub templates for bug report and issue

deathbeam:
Based on points made in #4160, update GitHub issue templates to be more
versatile, do not include any inline comments and be straightforward,
while pointing out reporter in correct direction. Also, one of the
crucial things that previous template for bug report was missing was
RuneLite and Launcher versions and steps to reproduce the bug.

Supersedes #4160


jplsek:
I originally removed the "descriptions" from the templates ("A clear and concise description of what the bug is.") because I didn't like when the issuer didn't remove them. The boilerplate description does not need to be in the issue imo. Maybe that could stay as an inline comment?

I also like https://github.com/runelite/runelite/search?type=Issues for issue searching since it shows closed, open issues and PRs for people unfamiliar with GitHub. (Sometimes a PR doesn't have an issue related to it, but if an issuer see's a PR, then they don't need to make a new issue.)

I'm probably slightly biased about this, but I also like linking to the FAQ page, since as shown by the most recent new issues reported, some of them could have been avoided if they were pointed to the new FAQ page.

deathbeam:
I dont like putting comments very much in the issues, and clear and concise descriptions under each issue field are better than just empty fields, so it directs user to think about what he is writing there more. 

I dont think linking FAQ page is necessary as issues already contains most of the stuff in FAQ.

PR #4226, opened 2018-07-08
Model Viewer Improvements and Additions

ScottMCarr:
-Fixed objects of variable width and length from being positioned incorrectly on the map.
-Renamed LocationKey to ModelKey and added another field so that cached values are correct for objects using the same models.
-Implemented shading to the scene.
-Implement full SceneTilePaint with variable colors and textures.
-Implement textures while rendering SceneTileModel
-Enable maps to be loaded by their region id instead of just their coordinates.
-Implement full object model creation (replacement colors, replacement textures, resizing, translation)
-Added two new methods to ModelDefinition used in object model creation.

### DIFF
--- a/cache/src/main/java/net/runelite/cache/definitions/ModelDefinition.java
+++ b/cache/src/main/java/net/runelite/cache/definitions/ModelDefinition.java
@@ -601,4 +601,15 @@ public class ModelDefinition
 
 		}
 	}
+
+	public void move(int xOffset, int yOffset, int zOffset)
+	{
+		for (int i = 0; i < this.vertexCount; i++)
+		{
+			this.vertexPositionsX[i] += xOffset;
+			this.vertexPositionsY[i] += yOffset;
+			this.vertexPositionsZ[i] += zOffset;
+		}
+		this.reset();
+	}
 }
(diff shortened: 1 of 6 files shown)


cache/src/main/java/net/runelite/cache/definitions/ModelDefinition.java
```diff
@@ -601,4 +601,30 @@ public void retexture(short var1, short var2)
...
+		for (int i = 0; i < this.vertexCount; i++)
+		{
+			this.vertexPositionsX[i] += xOffset;
+			this.vertexPositionsY[i] += yOffset;
+			this.vertexPositionsZ[i] += zOffset;
+		}
+		this.reset();
+	}
+
+	public void method2663(int var1)
```
Adam-: this is just `rotate`

model-viewer/src/main/java/net/runelite/modelviewer/ModelViewer.java
```diff
@@ -843,6 +909,21 @@ public static int RS2HSB_to_RGB(int RS2HSB)
...
+		int rgb = Color.HSBtoRGB((float) decode_hue / 63, (float) decode_saturation / 7, (float) decode_brightness / 127);
+		return adjustRGB(rgb & 0xFFFFFF, 0.6);
+	}
+
+	private static int adjustRGB(int rgb, double brightness)
+	{
+		double rp = (double) (rgb >> 16) / 256.0D;
+		double gp = (double) (rgb >> 8 & 255) / 256.0D;
+		double bp = (double) (rgb & 255) / 256.0D;
+		rp = Math.pow(rp, brightness);
```
Adam-: This noticeably destroys performance. Maybe use precomputed colors like the client.

ScottMCarr: Added a new method to use in place of this one. I have kept it for reference though.

PR #1598, opened 2018-04-16
Adding percentage to raids plugin overlay

JonnoUK:
- Toggle in configuration 
- Percentage displays on same line as player total points

NOTE: I personally haven't tested this IN Raids, however I see no reason for the percentage to be incorrect. (If somebody could try this, it'd be amazing)
The toggle works in real time, and doesn't need reloading etc.

Closes Issue #1499

Issue #1499, opened 2018-04-13
Show percentage of total points while in the raid


Denny966:
Currently, it only shows the % of total points at the end, add an option to show the % of total points while raiding

winterdaze:
This has been implemented in the "raids points overlay" plugin on the plugin hub

PR #4239, opened 2018-07-09
Adds Point Percentage to Chambers of Xeric

AtlassianRS:
Shows the point percentage in the points overlay just under personal points to one decimal place. 

Related to
Issue [#1499](https://github.com/runelite/runelite/issues/1499)
Issue [#1598](https://github.com/runelite/runelite/pull/1598)


Trevor159:
This should be closed because we do not use a custom points overlay anymore, it was moved to an external plugin and he can request this feature or PR it there if it is wanted.

Nightfirecat:
We switched to making the vanilla interface movable in #8018, and as Trevor said the previous custom panel exists as an external plugin now.

PR #4241, opened 2018-07-09
Fixed line breaking on right side

AtlassianRS:
This should correct the calculation when multiple lines appear on the right side. 
Has been tested in some scenarios.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/LineComponent.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/LineComponent.java
@@ -78,12 +78,6 @@ public class LineComponent implements LayoutableRenderableEntity
 				leftSmallWidth -= rightSmallWidth;
 			}
 
-			System.out.println("leftFullWidth: " + leftFullWidth);
-			System.out.println("rightFullWidth: " + rightFullWidth);
-			System.out.println("leftSmallWidth: " + leftSmallWidth);
-			System.out.println("rightSmallWidth: " + rightSmallWidth);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
The code looks fine at a glance, will test this further tomorrow.

AtlassianRS:
Found a bug in the original pull request. If the left text was too long, it would compute a negative size for the right maxwidth. I've reverted the calculation of the rightSmallWidth to preferredsize / 3. It looks like the only calculation issue was in the position of the right text.

Nightfirecat:
fixed

:tada:

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/LineComponent.java
```diff
@@ -74,7 +74,7 @@ public Dimension render(Graphics2D graphics)
 
 			if (!Strings.isNullOrEmpty(right))
 			{
-				rightSmallWidth = (preferredSize.width / 3);
+				rightSmallWidth = preferredSize.width / 3;
```
Nightfirecat: If there's no functional change, can the changes be reverted on this line?

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/LineComponent.java
```diff
@@ -170,5 +170,4 @@ private static int getLineWidth(final String line, final FontMetrics metrics)
 
 		return wrapped.toString().split("\n");
 	}
-}
-
+}
```
Nightfirecat: Revert the newline at EOF change.

PR #4247, opened 2018-07-09
Hiscore autocomplete: Fix edge case bug

pettenge:
This fixes a separate bug I discovered while investigating #3920 . When two or more keys are typed at the same time, some characters could get placed at the end of the auto-completed name.


Adam-:
This probably is the same bug as #3920. I believe that the awt keylistener is always called on the awt thread, so the invokeLater is unnecessary?

Adam-:
Fixed in 30c3ce9dc17f799aec59a177d82360becf6a664b

PR #4255, opened 2018-07-10
ChatCommandsPlugin: Update boss kc when looking at RoW

MagicfTail:


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -237,6 +244,48 @@ public void onChatMessage(ChatMessage chatMessage)
 		}
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick event)
```
deathbeam: So this is just to delay the loading 1 game tick? I am starting to feel like we do this way too often we should probably have some util method for it, like scheduleLater or something, maybe :d

MagicfTail:
This will not work in instances to make it so it won't update your kc by looking at other players' adventure logs. It could be made to work by looking at the previous widget, displayed below, and making sure it's "The exploits of *your username*", but that would be janky to make work reliable.

[Image](https://i.imgur.com/a1HESxy.png)

PR #4264, opened 2018-07-11
cache: Relax label name constraints

abextm:
Allows labels to be any reasonable identifier, not just `LABEL-?[0-9+]`

INSTRUCTION was changed to IDENTIFIER because if the LABEL rule is relaxed the lexer cannot tell the difference between a LABEL and a INSTRUCTION.

### DIFF
--- a/cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
+++ b/cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
@@ -42,25 +42,25 @@ string_var_value: INT ;
 
 line: instruction | label | switch_lookup ;
 instruction: instruction_name instruction_operand ;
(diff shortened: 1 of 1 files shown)


cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
```diff
@@ -42,25 +42,25 @@ string_var_value: INT ;
 
 line: instruction | label | switch_lookup ;
 instruction: instruction_name instruction_operand ;
-label: 'LABEL' INT ':' ;
+label: IDENTIFIER ':' ;
 
 instruction_name: name_string | name_opcode ;
-name_string: INSTRUCTION ;
+name_string: IDENTIFIER ;
```
Adam-: Instruction names are also numbers if the opcode is not known.

Issue #3673, opened 2018-06-07
slayer plugin tooltip highlight not accurate/doesn't work


abra238:
https://i.imgur.com/80xbUXz.png

Abextm:
What is incorrect?

abra238:
mousing over the task icon shows points and streak. streak shows 0, and
points shows 8/9.

Alexsuperfly:
I've spoken to a friend of mine and he has the same issue.  He says it shows `Points: 17` `Streak: 0`.  he currently has over 17 thousand slayer points and told me that when he was at ~15k it showed `Points:15` and when he broke into the 16 thousands it showed `Points: 16`.  This seems to me like the regex is mis matching when the points total is over one thousand points.

PR #4271, opened 2018-07-11
Slayer tooltip fix

Alexsuperfly:
Fixes the regex matching to eliminate all commas in the numbers parsed before converting them to Integers.
Updates the REWARD_POINTS pattern to take into account possibility of commas.

fixes: #3673 

Before: 
After:


Nightfirecat:
Can you squash your commits and basically put everything you wrote in the PR into your commit message? "remove comma from points and streak capture" and "update REWARD_POINTS regex to account for commas" don't explain what the commits fix.

Alexsuperfly:
cleaned up the regex to look better and added a test for a streak over 1,000 and points total over 1,000,000

PR #4291, opened 2018-07-12
Update widget used for retrieving Kourend Library book ID

trimbe:
Child IDs for sprite and text in sprite dialogs were changed and broke finding the id for the found book.


Abextm:
This should change the widgetID, not the reference to it

trimbe:
Right, I missed that, fixed.

Adam-:
duplicate of #4301

PR #4296, opened 2018-07-13
timers plugin: Use item and sprite IDs for timers

Nightfirecat:


runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -92,28 +99,69 @@
 		this.duration = Duration.of(time, unit);
 	}
 
+	GameTimer(int itemId, int quantity, String description, Integer graphicId, long time, ChronoUnit unit)
```
deathbeam: shouldnt this quantity be just always 1?

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -92,28 +99,69 @@
...
+		this.spriteId = spriteId;
+	}
+
 	GameTimer(String imageResource, String description, long time, ChronoUnit unit)
 	{
 		this(imageResource, description, null, time, unit);
 	}
 
-	public BufferedImage getImage()
+	GameTimer(int itemId, int quantity, String description, long time, ChronoUnit unit)
```
deathbeam: See my previous comment. If you need to differentiate i would use some enum or something probably

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimerTimer.java
```diff
@@ -33,9 +35,12 @@
 {
 	private final GameTimer timer;
 
+	static ItemManager itemManager;
+	static SpriteManager spriteManager;
+
 	public TimerTimer(GameTimer timer, Plugin plugin)
```
deathbeam: plugin.getItemManager and plugin.getSpriteManager. And remove the static assigns. or even send both to constructor

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
Nightfirecat: I guess? I wrote this constructor with two reasons in mind:

1. The item and sprite constructors can't both have the same arguments (if there's a better way to handle that, I'm all ears)
2. It's unlikely, but possible someone might want to have a timer with a stackable item as the displayed image, like a pile of coins.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
Nightfirecat: It's enums all the way down!

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimerTimer.java
Nightfirecat: I went ahead with the constructor approach.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimerImageType.java
```diff
@@ -0,0 +1,31 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.timers;
+
+enum GameTimerImageType
```
Kamielvf: I feel like this whole enum isn't needed.

PR #4299, opened 2018-07-13
timers plugin: Fix DMM normal spellbook freezes

Nightfirecat:
In Deadman mode, protect from magic does not half the duration of normal
spellbook freeze spells (bind, snare, entangle) unlike non-Deadman
worlds.

Source: https://www.runescape.com/oldschool/deadman/season


Nightfirecat:
According to the "Binding spells" section of this dev blog, this appears to only affect SDMM worlds, and not the permanent DMM world. I will change this PR accordingly.

Issue #3062, opened 2018-05-21
Freeze timer currently resets after every succesful cast


Tomato3s:
It's a bug that should be fixed.

deathbeam:
This should be fixed now

PR #4300, opened 2018-07-13
timers plugin: Don't reset freezes on every cast

Nightfirecat:
Instead of resetting the ice spell timers on every successful cast, they
are only set when no timer already exists.

Note: This behavior is still not fully correct, as there are ~5 ticks of
immunity to freezes once an existing timer expires.

Fixes runelite/runelite#3062


Nightfirecat:
Closing in favor of #4551, that PR incorporates some of the changes here but has much better implementation to handle edge cases.

PR #4301, opened 2018-07-13
ScreenshotPlugin: Fix widget ids after update

MagicfTail:


MagicfTail:
Causing no screenshots being taken on level up

MagicfTail:
This should now fix all the used widget ids that were changed

PR #4313, opened 2018-07-13
Include apostrophe and hyphen in kc regex

trimbe:
Fixes kc detection for bosses such as Kree'ara and TzKal-Zuk


Kamielvf:
For reference

Zuk kc message:
image

Arma kc message:
image

Vet'ion kc message:
image

Adam-:
Incorporated in 17f87669eb212510c0925ce2ded6b7df91808921

PR #4322, opened 2018-07-14
Slayer overlay code structure and query refactoring.

JHPinto:
QueryRunner is now injected in the constructor.
Method checkInventory moved bellow the constructor.
Restructured queries inside checkInventory to utilize idEquals.
Renamed method checkInventory to getSlayerItems to better reflect the new business logic.
Removed the now unnecessary check in the render call.
Changed Set<Integer> of slayer equipment to ImmutableIntArray, and expanded it to include jewelry to prevent the necessity of repeated concat calls when calling idEquals.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
@@ -64,7 +64,7 @@ class SlayerOverlay extends Overlay
 		ItemID.SLAYER_RING_8
 	);
 
-	private final ImmutableIntArray allSlayerItems = ImmutableIntArray.of(
+	private final static ImmutableIntArray ALL_SLAYER_ITEMS = ImmutableIntArray.of(
 		ItemID.SLAYER_HELMET,
 		ItemID.SLAYER_HELMET_I,
 		ItemID.BLACK_SLAYER_HELMET,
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
```diff
@@ -65,7 +64,7 @@
 		ItemID.SLAYER_RING_8
 	);
 
-	private final Set<Integer> slayerEquipment = ImmutableSet.of(
+	private final ImmutableIntArray allSlayerItems = ImmutableIntArray.of(
```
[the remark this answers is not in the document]
JHPinto: Would it make sense to do the same to the Set of jewelryItems?

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerOverlay.java
```diff
@@ -65,7 +64,7 @@
 		ItemID.SLAYER_RING_8
 	);
 
-	private final Set<Integer> slayerEquipment = ImmutableSet.of(
+	private final static ImmutableIntArray ALL_SLAYER_ITEMS = ImmutableIntArray.of(
```
deathbeam: as it is static now it should be at the top of the class. Also, while at it, slayerJewelry should receive same treatment

JHPinto: Roger that

PR #4338, opened 2018-07-15
adds Grotesque Guardians instance close timer

basicDamen:
Adds a Boss Timer for the Grotesque Guardians after killing the final form of Dusk. (2 minutes until the instance closes after final form killed.)


MagicfTail:
This should imo be in the timers plugin instead since it works nothing like any of the other boss timers

basicDamen:
^ It would be odd seeing that it's an actual boss. Sure, it doesn't work the same way, but the instance is also handled a little bit differently.

Nightfirecat:
I agree with Magic. The boss timers plugin is pretty specifically only for boss respawn timers, not for other boss-related timers.

deathbeam:
Imo it fits less in the timers plugin code-wise as we have already existing API for that in Boss Timers, also noone ever said Boss Timers are specifically for boss respawn timers, so it being boss timer too applies. Also its countdown from dying too to where you need start paying attention so its arguable about it being "respawn" timer or nah.

basicDamen:
The instance has a 2 minute window of either respawning the NPCs or leaving the instance, otherwise it removes you. This timer gives you a sense of how much time you have until you are going to be forcibly removed from the instance. I personally utilized this because once you become more efficient at that boss, you leave your loot on the ground and continue a multi-kill trip, due to inventory space. This lets me know how much time I have to sort through loot, pick everything up, or start another kill. (Also helps if you have to look away for a second in-between kill sets.)

Nightfirecat:
That's fine, but I'd like to see the timer removed when you leave the instance or spawn the bosses, as it's not useful after either of those events.

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/Boss.java
```diff
@@ -57,7 +57,8 @@
 	CERBERUS(NpcID.CERBERUS, 8400, ChronoUnit.MILLIS, ItemID.HELLPUPPY),
 	THERMONUCLEAR_SMOKE_DEVIL(NpcID.THERMONUCLEAR_SMOKE_DEVIL, 8400, ChronoUnit.MILLIS, ItemID.PET_SMOKE_DEVIL),
 	KRAKEN(NpcID.KRAKEN, 8400, ChronoUnit.MILLIS, ItemID.PET_KRAKEN),
-	KALPHITE_QUEEN(NpcID.KALPHITE_QUEEN_965, 30, ChronoUnit.SECONDS, ItemID.KALPHITE_PRINCESS);
+	KALPHITE_QUEEN(NpcID.KALPHITE_QUEEN_965, 30, ChronoUnit.SECONDS, ItemID.KALPHITE_PRINCESS),
+	DUSK(NpcID.DUSK_7889, 120, ChronoUnit.SECONDS, ItemID.NOON);
```
Adam-: `2, ChronoUnit.MINUTES`?

basicDamen: Meant to change that as I was going to take a different approach, though, I didn't.

PR #4342, opened 2018-07-15
Add search support for skill calculators

### LINKED ISSUE #2976: Skill calculator (specifically construction) isn't organzied and hard to find out what method to use
See this image for more context: http://prntscr.com/jk1m7a

In the image I was to know all the methods from lv 70-78 construction, but when I do that it lists a million options with no way to filter. If there was a way to type in a search bar and search something like "oak larders" or "mahogany tables" that would make it a bit more organized.

rbbi:
image

Fixes #2976


Nightfirecat:
This would also fix  #5286

deathbeam:
Needs rebase

rbbi:
@deathbeam rebased

deathbeam:
Awesome. Only thing I would point out in this change is the SearchBar class. Wasn't the icon text field enough? Or can the search bar be used on other places where we have SearchBar? I dont mind the component, but if it only works in skill calc interface there is no point in having it.

rbbi:
@deathbeam Yep, i'm writing implementations for other interfaces too - will push to this PR when i'm finished

deathbeam:
So now it is basically plugin-specific logic inside the IconTextField? That probably seems worse than before. You should probably just leave the plugin specific appearance inside plugins and leave only the common stuff in IconTextField

rbbi:
@deathbeam This logic is encapsulated in an enum (Formatter) _within_ the IconTextField class - would you prefer if it was moved into another physical file?

deathbeam:
I would prefer if the plugin-related logic would stay in respective plugin's packages

rbbi:
@deathbeam updated

deathbeam:
Looks good. I would probably split the refactoring and the addition of the search bar to tracker to separate commits tho

rbbi:
@deathbeam rebased due to merge conflict

rbbi:
@deathbeam any update on this?

deathbeam:
No, my last comment still applies

PR #4350, opened 2018-07-16
runelite-client: Consume hotkeys in HotkeyListener

Abextm:
KeyTyped consumer any non-modifier key while the hotkey is being pressed because there is no good way to determine if a KeyTyped event matches a Keybind due to it's keyCode being zeroed.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
@@ -110,6 +110,11 @@ public class Keybind
 			keyCode = KeyEvent.VK_UNDEFINED;
 		}
 
+		if (e.getID() == KeyEvent.KEY_RELEASED && keyCode != KeyEvent.VK_UNDEFINED)
+		{
+			return this.keyCode == keyCode;
+		}
+
 		return this.keyCode == keyCode && this.modifiers == modifiers;
 	}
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/config/Keybind.java
```diff
@@ -39,15 +40,15 @@
 @EqualsAndHashCode
 public class Keybind
 {
-	private static final BiMap<Integer, Integer> modifierToKeyCode = new ImmutableBiMap.Builder<Integer, Integer>()
+	private static final BiMap<Integer, Integer> MODIFIER_TO_KEY_CODE = new ImmutableBiMap.Builder<Integer, Integer>()
```
deathbeam: .inverse is not very cheap afaik, you should probably declare static MODIFIER_TO_KEY_CODE_INVERSED or something

Abextm: It is cached internally

deathbeam: I see, alright

runelite-client/src/main/java/net/runelite/client/util/HotkeyListener.java
```diff
@@ -42,15 +42,28 @@
...
 
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
 		if (keybind.get().matches(e))
 		{
+			boolean wasPressed = isPressed;
 			isPressed = true;
-			hotkeyPressed();
+			if (wasPressed)
```
whitehooder: It should be `if (!wasPressed)`, or else it still repeats when the key is held down and even skips the first press.

PR #4360, opened 2018-07-17
Add WASD Camera Plugin (Stable)

Abexlry:
Plugin now uses client var strings to check if input is going into chat or not. If input is locked and input is being made in chat entry then it will try to handle camera. Text entry interfaces are not affected and should all work as intended.

WASD Camera Plugin:
-Listed under "WASD Camera"
-Allows use of WASD keys to move camera
-Chatbox is disabled with a text overlay that says "Press Enter to Chat..."
-Can recolor overlay in plugin option
-Hitting slash enables chat for easy clan chat
-Does not affect/disable F keys, arrow keys, etc.
-Can use dialog options (1, 2, 3...) when talking with npc while input is locked
-Can use text entry interfaces while input is locked


TheStonedTurtle:
You know you don't need to make new Pull Requests to update it? Just commit to the branch the PR is on.

Abexlry:
Thanks for letting me know :) Obviously still learning git.

Adam-:
It looks like you can just see if 162.0 has a keylistener or not to tell whether or not the chatbox input is active.

Adam-:
Added in #4874

PR #4364, opened 2018-07-17
Monk (level 3) health correction

ltvill:


ltvill:
Ardougne monastery. Accidentally hit one and realised the health is wrong.

Kamielvf:
Seems correct 👍

PR #4354, opened 2018-07-16
Added Copy-To-Clipboard for Raid Scouting

TimLehner:
When enabled, will copy the raid layout + puzzle/combat names to clipboard for saving during the raid in e.g. notes or similar

During raids as soon as the raid has begun the overlay disappears. This can be annoying if later trying to remember what order and what is next, particularly where scav/farm rooms are

PR #4365, opened 2018-07-17
Add !layout command to Chambers of Xeric plugin

deathbeam:
Add support for !layout command to raids plugin that will stream current
raid layout to everyone who will see the message.

The code for displaying the chat message is partially from #3919 and from #2992 so thanks for the work done in those PRs.

Depends on #4698

Closes #4354 
Closes #3919 
Closes #2992

screenie


MagicfTail:
!layout would probably be a more descriptive command name

Matsyir:
I'd like to point out that I would still like to have the option of an automatic message on start from #3919 however at this point it would be a very negligible change, this will do the trick albeit not automatically. For team raids this will be very useful though.

deathbeam:
Changed this PR to depend on #4698, what removes any duplicate code here from ChatCommandsPlugin

deathbeam:
Closing this as it needs @Adam- 's review and idk if he even likes it, and I am not going to maintain this forever

Issue #4366, opened 2018-07-17
Barrows brother tracker


rosscmcd:
**Is your feature request related to a problem? Please describe.**
The barrows brother tracker overlay on the side of the screen seems to take up quite a bit of space especially when playing on fixed mode.

**Describe the solution you'd like**
Would be nice if the unnecessary extra space was removed to make the plugin look more flush. 

**Describe alternatives you've considered**
An alternative would be to make the overlay toggle-able. Currently the only way to turn it off is to turn off the entire barrows plugin which makes the tunnels much more difficult.

**Additional context**
https://imgur.com/dBW3sEN

PR #4367, opened 2018-07-17
Barrows overlay switch

ltvill:
Also disables original overlay.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -43,6 +43,9 @@ public class BarrowsBrotherSlainOverlay extends Overlay
 	private final PanelComponent panelComponent = new PanelComponent();
 
 	@Inject
+	private BarrowsConfig config;
+
+	@Inject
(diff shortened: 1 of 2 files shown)


ltvill:
#4366

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
```diff
@@ -62,11 +65,16 @@ public Dimension render(Graphics2D graphics)
 
 		// Hide original overlay
 		Widget barrowsBrothers = client.getWidget(WidgetInfo.BARROWS_BROTHERS);
-		if (barrowsBrothers != null)
+		if (barrowsBrothers != null || !config.showBrotherOverlay())
```
Adam-: this isn't right

iVilux: What’s wrong? It worked fine when I tested it.

ltvill: What’s wrong? It worked fine when I tested it, unless you want to keep the original overlay.

PR #4373, opened 2018-07-17
Change skill calc checkboxes to radio buttons

### LINKED ISSUE #2935: Skill calculator prayer multipliers
I think it would be better if it's changed to a radiobutton instead of a checkbox, since you can only use one of the two options anyway .
how it is

rbbi:
giphy

Fixes #2935

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -25,7 +25,6 @@
  */
 package net.runelite.client.plugins.skillcalculator;
 
-import com.google.common.collect.Lists;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -67,6 +68,7 @@
 
 	private UICombinedActionSlot combinedActionSlot = new UICombinedActionSlot();
 	private ArrayList<UIActionSlot> combinedActionSlots = new ArrayList<>();
+	private List<BonusCheckBox> bonusCheckBoxes = Lists.newArrayList();
```
deathbeam: Make this final, and just use new ArrayList<>() probably.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -173,27 +181,64 @@ private void renderBonusOptions()
...
+	 * Adjusts and deselects all <b>other</b> checkboxes
+	 * contained within {@link SkillCalculator#bonusCheckBoxes}
+	 * except the checkbox provided as the parameter
+	 *
+	 * @param except The checkbox to keep unchanged
+	 */
+	private void adjustAllExcept(JCheckBox except)
+	{
+		bonusCheckBoxes.stream()
+			.filter(checkbox -> !checkbox.equals(except))
```
deathbeam: you should probably use != instead of equals

PR #4375, opened 2018-07-17
ui: Add Table Component

Nightfirecat:
Test overlay used to test multiple table types and sizes: https://gist.github.com/Nightfirecat/a1e2187ac0a1bd7b5ef40de76d96aad9

test table 1
test table 2
test table 3
test table 4
test table 5

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/PlayerComparisonOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/PlayerComparisonOverlay.java
@@ -40,9 +40,11 @@ import net.runelite.client.ui.overlay.OverlayLayer;
 import static net.runelite.client.ui.overlay.OverlayManager.OPTION_CONFIGURE;
 import net.runelite.client.ui.overlay.OverlayMenuEntry;
 import net.runelite.client.ui.overlay.OverlayPosition;
-import net.runelite.client.ui.overlay.components.LineComponent;
 import net.runelite.client.ui.overlay.components.PanelComponent;
+import net.runelite.client.ui.overlay.components.TableComponent;
+import static net.runelite.client.ui.overlay.components.TableComponent.TableAlignment;
 import net.runelite.client.ui.overlay.components.TitleComponent;
+import net.runelite.client.util.ColorUtil;
 import net.runelite.client.util.Text;
 import net.runelite.http.api.hiscore.HiscoreResult;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/components/table/TableAlignment.java
```diff
@@ -0,0 +1,32 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.ui.overlay.components.table;
+
+public enum TableAlignment
```
deathbeam: This should probably follow the panel component and live inside the TableComponent class

Nightfirecat: Good point, I've moved it into TableComponent.

deathbeam:
This is awesome. Do you think you could incorporate this to panel instead/replace panel component with this? Panel component was originally meant to have grid, but it was too annoying to implement so I gave up.

Nightfirecat:
Mentioned in discord, but copied here for visibility--in its current state, this PR was intended exclusively for laying out text. Re-implementing it for component layout would be a big effort and would need more thought put into it, but I may work on that in the future. (a possible Grid component to take the place of Panel component later on?)

Nightfirecat:
Also, if it would be preferred for this PR to have some use before being merged, I do plan on writing a PR for the player comparison table to use this. (4th image from original post) I can just add that commit here if that's preferred though.

Nightfirecat:
I've updated the opponent info plugin's player comparison table to use this component, as it looks vastly better than what I had initially created using `LineComponent`.

Before:
before

After:
after

Nightfirecat:
Rebased to fix build after getBounds addition.

PR #4072, opened 2018-06-29
Add color utility class

Nightfirecat:
This adds `ColorUtil::colorTag`, `ColorUtil::prependColorTag`, and `ColorUtil::wrapWithColorTag`, and implements them in existing plugins, as well as moving some color-related functions from `SwingUtil` into this new class.

Uses of `<col` in strings have been maintained as it is entirely proprietary to our code. (specifically the chat notifications plugin) Some uses of `<col=` ([1](https://github.com/Nightfirecat/runelite/blob/37407976a529cffc4cf34ee8fef5216c750a76b2/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java#L53), [2](https://github.com/Nightfirecat/runelite/blob/37407976a529cffc4cf34ee8fef5216c750a76b2/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPlugin.java#L199), [3](https://github.com/Nightfirecat/runelite/blob/cfd06f7c9a855592982de318dbb4fed0f54edc6c/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java#L93-L116), [4](https://github.com/Nightfirecat/runelite/blob/37407976a529cffc4cf34ee8fef5216c750a76b2/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TextComponent.java#L39)) in strings have been maintained for readability and have been moved as needed to constants.

PR #4378, opened 2018-07-18
add AttackStyle Chat Message Warnings

basicDamen:
Was a TODO sitting in `AttackStyle` for a while, so, I figured I'd give it a shot.

**TODO:** "chat message to warn players that their weapon switch also caused an unwanted attack style change"

This handles login, weapon change, and attackstyle change notifications.

All logic is commented. (HANDLES COMPLETELY DIFFERENT NOW.)

// Utilizes GameTicks, seeing as variables are set PRIOR to a GameTick actually being ran.
// The messages are guarded by this tick-set, therefore, the messages aren't spammed upon
// variable setting and then once all variables are set, this code block runs and sends
// the corresponding login message(s).

attackstyle - combined messages

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -338,21 +338,24 @@ public class AttackStylesPlugin extends Plugin
 			{
 				if (!messageTickGuarded)
 				{
-					// If the login message has been sent, it will run the corresponding messages if it's a weaponSwitch or just a combatStyle change.
-					if (loginMessageSent)
+					if (builder.length() > 0)
 					{
-						if (weaponSwitch)
+						// If the login message has been sent, it will run the corresponding messages if it's a weaponSwitch or just a combatStyle change.
+						if (loginMessageSent)
 						{
-							sendChatMessage("This weapon's attack style will grant XP in: " + builder);
+							if (weaponSwitch)
+							{
+								sendChatMessage("This weapon's attack style will grant XP in: " + builder);
+							}
+							else
+							{
+								sendChatMessage("This attack style will grant XP in: " + builder);
+							}
 						}
 						else
 						{
-							sendChatMessage("This attack style will grant XP in: " + builder);
+							sendChatMessage("You logged in wielding a weapon that will grant XP in: " + builder);
 						}
-					}
-					else
-					{
-						sendChatMessage("You logged in wielding a weapon that will grant XP in: " + builder);
 					}
 				}
 			}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -310,9 +310,7 @@ public class AttackStylesPlugin extends Plugin
 	private void updateWarning(boolean weaponSwitch)
 	{
 		warnedSkillSelected = false;
-		String[]styleBuilderArray = new String[4];
-		String styleBuilder = "";
-		int count = 0;
+		StringBuilder builder = new StringBuilder();
 		if (attackStyle != null)
 		{
 			for (Skill skill : attackStyle.getSkills())
@@ -324,8 +322,12 @@ public class AttackStylesPlugin extends Plugin
 						// If all variables have been set, this guard will be set to false, therefore, continuing the code block pertaining to messages.
 						if (!messageTickGuarded)
 						{
-							count++;
-							styleBuilderArray[count] = getAttackColor(skill);
+							if (builder.length() > 0)
+							{
+								builder.append(", ");
+							}
+
+							builder.append(getAttackColor(skill));
 						}
 					}
 					warnedSkillSelected = true;
@@ -336,45 +338,21 @@ public class AttackStylesPlugin extends Plugin
 			{
 				if (!messageTickGuarded)
 				{
-					for (int i = 1; i <= count; i++)
-					{
-						if (i == 1)
-						{
-							styleBuilder = styleBuilderArray[i];
-							continue;
-						}
-						else if (i == 2 && count != 2)
-						{
-							styleBuilder += (", " + styleBuilderArray[i]);
-							continue;
-						}
-						else if (i == 2 && count == 2)
-						{
-							styleBuilder += (" and " + styleBuilderArray[i]);
-							continue;
-						}
-						else if (i == 3)
-						{
-							styleBuilder += (", and " + styleBuilderArray[i]);
-							continue;
-						}
-					}
-
 					// If the login message has been sent, it will run the corresponding messages if it's a weaponSwitch or just a combatStyle change.
 					if (loginMessageSent)
 					{
 						if (weaponSwitch)
 						{
-							sendChatMessage("This weapon's attack style will grant you " + styleBuilder + " XP.");
+							sendChatMessage("This weapon's attack style will grant XP in: " + builder);
 						}
 						else
 						{
-							sendChatMessage("This attack style will grant you " + styleBuilder + " XP.");
+							sendChatMessage("This attack style will grant XP in: " + builder);
 						}
 					}
 					else
 					{
-						sendChatMessage("You logged in wielding a weapon that will grant you " + styleBuilder + " XP.");
+						sendChatMessage("You logged in wielding a weapon that will grant XP in: " + builder);
 					}
 				}
 			}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -310,33 +310,72 @@ public class AttackStylesPlugin extends Plugin
 	private void updateWarning(boolean weaponSwitch)
 	{
 		warnedSkillSelected = false;
+		String[]styleBuilderArray = new String[4];
+		String styleBuilder = "";
+		int count = 0;
 		if (attackStyle != null)
 		{
 			for (Skill skill : attackStyle.getSkills())
 			{
 				if (warnedSkills.contains(skill))
 				{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -196,10 +220,19 @@ public void onEquippedWeaponTypeChange(VarbitChanged event)
...
+				loginWarningMessageCount++;
+			}
+
 			equippedWeaponTypeVarbit = client.getVar(Varbits.EQUIPPED_WEAPON_TYPE);
 			updateAttackStyle(equippedWeaponTypeVarbit, client.getVar(VarPlayer.ATTACK_STYLE),
 				client.getVar(Varbits.DEFENSIVE_CASTING_MODE));
-			updateWarning(true);
+
+			if (loginWarningMessageCount == 3 || loginWarningMessageCount == 4)
+				updateWarning(true);
```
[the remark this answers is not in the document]
basicDamen: I'll comment the entire file to hopefully create a better understanding for other minds.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -260,21 +260,27 @@ public void onConfigChanged(ConfigChanged event)
 			switch (event.getKey())
 			{
 				case "warnForDefensive":
+					sendChatMessage("You have " + (enabled ? "enabled" : "disabled") + " the warning for " + returnAttackColor(Skill.DEFENCE) + " XP." );
```
Nightfirecat: This should definitely not be an ingame chat message, and there is already a debug statement generated upon config changes.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -323,17 +329,20 @@ private void updateWarning(boolean weaponSwitch)
 				{
 					if (loginWarningMessageCount == 3)
 					{
-						sendChatMessage("<col=871A1A>[ATTACK STYLE]</col> You logged in wielding a weapon that will grant you " + returnAttackColor(skill) + " XP. #" + loginWarningMessageCount);
+						if (config.warningMessages())
```
Nightfirecat: There should not be three `if (config.warningMessages())` checks for a block which will always print a chat message of some kind. Ideally, this check would be done in `sendChatMessage` (which should probably be renamed accordingly) so it does not always need a paired config check to determine whether it should print.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -367,6 +416,41 @@ private void hideWidget(Widget widget, boolean hidden)
...
+				QueuedMessage.builder()
+					.type(ChatMessageType.GAME)
+					.runeLiteFormattedMessage(message)
+					.build());
+		}
+	}
+
+	private String returnAttackColor(Skill skill)
+	{
+		String color = "<col=ff0000>" + skill + "</col>";
```
Nightfirecat: Just make `color` the leading tag, and append the skill name and ending tag after determining what the color is.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -177,6 +188,29 @@ public void onGameStateChange(GameStateChanged event)
...
+		{
+			loginMessageSent = false;
+			messageTickGuarded = true;
+		}
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick e)
+	{
+		// Utilizes GameTicks, seeing as variables are set PRIOR to a GameTick actually being ran.
```
Nightfirecat: I'm inclined to say this should be part of the commit body rather than a comment. It should definitely be stated somewhere, but the full explanation probably doesn't need to be here. A short `// One tick delay to ensure vars are set`` comment here should suffice if this is moved to the commit message.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -177,6 +188,29 @@ public void onGameStateChange(GameStateChanged event)
...
+		// The messages are guarded by this tick-set, therefore, the messages aren't spammed upon
+		// variable setting and then once all variables are set, this code block runs and sends
+		// the corresponding login message(s).
+		if (messageTickGuarded)
+		{
+			messageTickGuarded = false;
+			updateWarning(false);
```
Nightfirecat: Why is this new `updateWarning()` call added?

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -367,6 +416,41 @@ private void hideWidget(Widget widget, boolean hidden)
...
+
+			chatMessageManager.queue(
+				QueuedMessage.builder()
+					.type(ChatMessageType.GAME)
+					.runeLiteFormattedMessage(message)
+					.build());
+		}
+	}
+
+	private String returnAttackColor(Skill skill)
```
Nightfirecat: This method name is kind of awkward... `getAttackColor()`? `attackStyleColor()`?

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
basicDamen: It's to send the login messages, seeing as all variables are set and it can NOW dictate what to output.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
basicDamen: Okay. I figured someone wouldn't want it there, but, it was mainly for someone to just understand why it utilizes that method. I'll move it.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
basicDamen: 👍

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -434,21 +431,21 @@ private void sendChatMessage(String chatMessage)
...
+				color = "<col=D44A4A>";
 			case RANGED:
-				return color = "<col=259443>" + skill + "</col>";
+				color = "<col=259443>";
 			case MAGIC:
-				return color = "<col=369EB3>" + skill + "</col>";
+				color = "<col=369EB3>";
 		}
-		return color;
+		return color  + skill + "</col>";
```
Nightfirecat: Remove double space between `color  + skill`

basicDamen: 👍

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -285,12 +316,27 @@ private void updateWarning(boolean weaponSwitch)
...
+								sendChatMessage("This attack style will grant you " + getAttackColor(skill) + " XP.");
+							}
+						}
+						else
+						{
+							sendChatMessage("You logged in wielding a weapon that will grant you " + getAttackColor(skill) + " XP.");
+						}
 					}
 					warnedSkillSelected = true;
-					break;
```
Nightfirecat: Given the chattiness of adding multiple chat messages for a single login or weapon switch event, I would like to see messages looking more like `You logged in wielding a weapon that will grant you Attack, Strength, and Defence XP.` This would likely require ditching the loop over the attack styles and trying to find an intersection between warned and active styles instead.

basicDamen: I thought about it, but.. it wouldn't be that difficult.. I have an idea in mind on how to make that.. But, I'll look into it.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -336,45 +338,21 @@ private void updateWarning(boolean weaponSwitch)
...
-						}
-					}
-
 					// If the login message has been sent, it will run the corresponding messages if it's a weaponSwitch or just a combatStyle change.
 					if (loginMessageSent)
 					{
 						if (weaponSwitch)
 						{
-							sendChatMessage("This weapon's attack style will grant you " + styleBuilder + " XP.");
+							sendChatMessage("This weapon's attack style will grant XP in: " + builder);
```
Nightfirecat: Maybe these `sendChatMessage` calls should be assignments of a printf-formatted string instead:

```
attackStyleWarningFormat = "This weapon's attack style will grant XP in: %s";
```

and a single `sendChatMessage(String.format(attackStyleWarning, builder))` outside of the `if (loginMessageSent)` if-block.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -367,6 +435,46 @@ private void hideWidget(Widget widget, boolean hidden)
...
+				color = "<col=259443>";
+				break;
+			}
+			case MAGIC:
+			{
+				color = "<col=369EB3>";
+				break;
+			}
+		}
+		return color + skill.getName() + "</col>";
```
Nightfirecat: Now that #4072 has been merged, use `ColorUtil::wrapWithColorTag` and store an actual _`Color`_ object in `color` to change this to `return ColorUtil.wrapWithColorTag(skill.getName(), color);`

PR #4393, opened 2018-07-19
Stop screenshot plugin from overwriting pictures

### LINKED ISSUE #4374: Screenshots taken within the same second overwrite previous screenshots in that second
If I click the screenshot hotkey or button multiple times within a second, the last screenshot taken within that second will be the only one present in the screenshots folder because it overwrites any previous screenshots of the same name already present.

Note: This is true in any other scenario where an existing screenshot has the same name as a new one. (DMM level-up screenshots, for instance)

Noremac201:
fixes #4374

Simply added more precise seconds to the file name.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -648,6 +648,14 @@ public class ScreenshotPlugin extends Plugin
 		{
 			File screenshotFile = new File(playerFolder, fileName + ".png");
 
+			// To make sure that screenshots don't get overwritten, check if file exists,
+			// and if it does create file with same name and suffix.
+			int i = 1;
+			while (screenshotFile.exists())
+			{
+				screenshotFile = new File(playerFolder, fileName + String.format("(%d)", i++) + ".png");
+			}
+
 			ImageIO.write(screenshot, "PNG", screenshotFile);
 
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This fixes the linked issue, but not the more general "Screenshots of same name overwrite each other" issue. Personally, I'd like to see a fix that works like most filesystems handle duplicate file names--appending a counter to the screenshot, like  `Screenshot Name (1).png`

Noremac201:
Right you are, 
I've updated the PR to fix the more general issue.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -107,7 +107,7 @@
 	private static final HttpUrl IMGUR_IMAGE_UPLOAD_URL = HttpUrl.parse("https://api.imgur.com/3/image");
 	private static final MediaType JSON = MediaType.parse("application/json");
 
-	private static final DateFormat TIME_FORMAT = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss");
+	private static final DateFormat TIME_FORMAT = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss.SSS");
```
Nightfirecat: With the change to loop over filenames, this probably isn't needed.

Noremac201:
Removed the unnecessary code, and updated the commit.

Thanks for taking a look.

PR #4420, opened 2018-07-19
TimerPlugin: Fix prayer enhance length

### LINKED ISSUE #4409: Prayer Enhance (raids) timer is off
The timer reaches 00:00 and then approximately 10 seconds later, the enhance effect goes off.
EDIT: 14-15 seconds to be exact.

MagicfTail:
Fixes #4409


Adam-:
I do not think the time is constant:

https://github.com/runelite/runelite/commit/d684bef65c0b4107eaefde77a2396d31a3b8f4eb#diff-fdeef3e1c6d3c4aa9740b0c76e40e2a1
https://github.com/runelite/runelite/commit/56a4aecb6a2dba5b25757bd4106e253db4c14a0c#diff-fdeef3e1c6d3c4aa9740b0c76e40e2a1

Kamielvf:
It might have changed due to some jagex mistake (it used to give a warning before running out, this was removed for whatever reason) and that's when we shortened our timer.
If they re-added the 15 seconds before running out warning, the 290 second timer would be correct like before.

deathbeam:
Can someone confirm is this timer is still correct? If yes we should merge it.

Kamielvf:
Yes, this is still correct. Did a raid earlier

PR #4428, opened 2018-07-20
Abyssal Sire Stun Countdown

### LINKED ISSUE #4132: Add 30sec Abyssal Sire Stun Countdown to bosstimers
If you hit a Shadow barrage spell on the Abyssal Sire you he is stunned for 30 seconds, i'd just like to have a boss timer on that.

every shadow spell on the ancient spell book has an increased chance on stunning him, Rush 25% /Burst 50% /Blitz 75% /Barrage 100%, If he gets hit for any damage with the Shadow Barrage spell he is stunned, others are just RNG. 

When he is stunned, he is always stunned for 30 seconds.

pettenge:
Add Abyssal Sire stun countdown to Timers plugin. Closes #4132.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -27,7 +27,6 @@ package net.runelite.client.plugins.timers;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
-import java.util.Arrays;
 import javax.inject.Inject;
 import net.runelite.api.Actor;
 import net.runelite.api.AnimationID;
@@ -102,15 +101,12 @@ public class TimersPlugin extends Plugin
 {
 	private static final int ABYSSAL_SIRE_STUNNED_NPC_ID = NpcID.ABYSSAL_SIRE_5887;
 
-	private static final int[] ABYSSAL_SIRE_NOT_STUNNED_NPC_IDS = new int[]
-	{
-		NpcID.ABYSSAL_SIRE,
-		NpcID.ABYSSAL_SIRE_5888,
-		NpcID.ABYSSAL_SIRE_5889,
-		NpcID.ABYSSAL_SIRE_5890,
-		NpcID.ABYSSAL_SIRE_5891,
-		NpcID.ABYSSAL_SIRE_5908
-	};
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_1 = NpcID.ABYSSAL_SIRE;
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_2 = NpcID.ABYSSAL_SIRE_5888;
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_3 = NpcID.ABYSSAL_SIRE_5889;
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_4 = NpcID.ABYSSAL_SIRE_5890;
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_5 = NpcID.ABYSSAL_SIRE_5891;
+	private static final int ABYSSAL_SIRE_NOT_STUNNED_NPC_ID_6 = NpcID.ABYSSAL_SIRE_5908;
 
 	private int lastRaidVarb;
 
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,17 +469,32 @@ public void onAnimationChanged(AnimationChanged event)
 	{
 		Actor actor = event.getActor();
 
-		if (actor != client.getLocalPlayer())
-		{
-			return;
-		}
-
 		if (config.showVengeanceOther()
+			&& actor == client.getLocalPlayer()
```
Nightfirecat: Revert the guard clause change, and move the sire check above the guard clause

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,17 +469,32 @@ public void onAnimationChanged(AnimationChanged event)
...
 		{
 			createGameTimer(VENGEANCEOTHER);
 		}
+
+		String actorName = actor.getName();
+
+		if (config.showAbyssalSireStun()
+			&& actor instanceof NPC
+			&& actorName != null
+			&& actorName.equals("Abyssal Sire"))
```
Nightfirecat: Remove the `actorName` assignment and null check, and change this check to `"Abyssal Sire".equals(actor.getName())`

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -94,12 +95,16 @@
...
-	tags = {"combat", "items", "magic", "potions", "prayer", "overlay"}
+	tags = {"combat", "items", "magic", "potions", "prayer", "overlay", "abyssal", "sire"}
 )
 public class TimersPlugin extends Plugin
 {
+	private static final int ABYSSAL_SIRE_STUNNED_NPC_ID = NpcID.ABYSSAL_SIRE_5887;
+
 	private int lastRaidVarb;
 
+	private boolean abyssalSireStunTimerShown;
```
Nightfirecat: This variable seems useless; it is only set within the same method that checks it--just default to removing the stun timer if the changed NPC doesn't match the stunned sire ID.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,17 +469,32 @@ public void onAnimationChanged(AnimationChanged event)
...
+			&& actor instanceof NPC
+			&& actorName != null
+			&& actorName.equals("Abyssal Sire"))
+		{
+			if (((NPC)actor).getId() == ABYSSAL_SIRE_STUNNED_NPC_ID)
+			{
+				createGameTimer(ABYSSAL_SIRE_STUN);
+				abyssalSireStunTimerShown = true;
+			}
+			else if (abyssalSireStunTimerShown)
```
Nightfirecat: As mentioned in the comment on the variable, just do `else { ...`

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,6 +467,20 @@ public void onAnimationChanged(AnimationChanged event)
 	{
 		Actor actor = event.getActor();
 
+		if (config.showAbyssalSireStun()
```
deathbeam: You should move this block after the actor != localPlayer

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,6 +467,20 @@ public void onAnimationChanged(AnimationChanged event)
...
+			&& actor instanceof NPC
+			&& "Abyssal Sire".equals(actor.getName()))
```
deathbeam: Use NpcID instead of strings

Nightfirecat: Agreed, now that I think about it, there's no real need to call `removeGameTimer` for this.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: If it was moved there, it wouldn't ever work.

deathbeam: oh, right true, oops

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
pettenge: I'm calling removeGameTimer for the case when the Sire exits the stun phase and leaves the throne.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -464,6 +478,21 @@ public void onAnimationChanged(AnimationChanged event)
...
+		if (config.showAbyssalSireStun()
+			&& actor instanceof NPC)
+		{
+			final int npcId = ((NPC)actor).getId();
+
+			if (npcId == ABYSSAL_SIRE_STUNNED_NPC_ID)
+			{
+				createGameTimer(ABYSSAL_SIRE_STUN);
+			}
+			else if (Arrays.stream(ABYSSAL_SIRE_NOT_STUNNED_NPC_IDS).anyMatch(id -> id == npcId))
```
deathbeam: Just use list and use contains? But I would just do || checks here

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -94,10 +95,19 @@
 @PluginDescriptor(
 	name = "Timers",
 	description = "Show various timers in an infobox",
-	tags = {"combat", "items", "magic", "potions", "prayer", "overlay"}
+	tags = {"combat", "items", "magic", "potions", "prayer", "overlay", "abyssal", "sire"}
 )
 public class TimersPlugin extends Plugin
 {
+	private static final int ABYSSAL_SIRE_STUNNED_NPC_ID = NpcID.ABYSSAL_SIRE_5887;
```
deathbeam: Just use the NpcID class directly no need to store the IDs as constants here

pettenge: I was trying to make it more readable since NpcID doesn't explicitly say which state the id represents.

deathbeam: Comment is enough, the constants were not very helpful as well.

deathbeam:
Also, thinking about this again, shouldn't this be in NPC spawned instead of animation changed? Not sure how sire works

pettenge:
The spawned/despawned events don't get raised when the Sire's id changes. I found that the animation change event is raised whenever the npc id changes (but an animation change does not necessarily mean the npc id changed). The alternative was to check the npc id on every tick, which isn't very efficient.

deathbeam:
I see. That is weird, but alright

PR #4429, opened 2018-07-20
Add support for RS movable widgets

deathbeam:
Add support for RS movable widgets

Add support for adding RS widgets to RuneLite overlay system and make
them moveable with persisted positions during restarts.

Depends on #3820 for widget ids

Preview:
peek 2018-07-21 00-00


Adam-:
`worldToMiniMap` would need to be adjusted to account for this.

PR #4431, opened 2018-07-20
Overhead API Changes

Infinitay:
- Changed OverheadIcon to PrayerOverhead
  - Updated Plugins using OverheadIcon
- Added SpriteOverhead to Player
- Added isSkulled method to Player

I should note I haven't tested this yet as I do not have an injected client containing these changes.


Infinitay:
> Can the new methods be named OverheadPrayer and OverheadSprite, like OverheadIcon

Sure I'll change it, I thought it both sounded better, and would be better to name the _type_ of overhead beforehand

KeeganKGB:
Any chance of this being implemented in the near future? isSkulled()  at the very least would be quite useful. As far as I can tell there's no way to detect it currently, as  getOverhead() only returns overhead prayers and not other overhead icons.

Infinitay:
Noticed that these changes have been approved, like the original post says: 
> I should note I haven't tested this yet as I do not have an injected client containing these changes.

IIRC, someone provided me image files for the different overheads and I assumed that the name of the image correlated with the overhead's id. Once again, I couldn't test it because I didn't have an injected client with the changes

Infinitay:
image

PR #4443, opened 2018-07-22
Inventory Value Overlay Plugin

DirkBijleveld:
Essentially the ground item calculator but for items in your inventory. Uses 5 configurable categories with colour overlays to indicate how much that stack is worth. 

This could be useful for extended activities with many different drops, especially to players that do not yet know the exact value of every item. 

This is my first plugin/github pull request so all criticism is welcome; I'd love to learn.


raiyni:
This is available in a hub plugin [Inventory Value Overlay](https://runelite.net/plugin-hub/show/inventory-value-overlay)

PR #4463, opened 2018-07-23
Add support for making XP globes moveable

deathbeam:
Add support for integration of XP globes with DETACHED overlay system,
e.g making them moveable.

Preview:
peek 2018-07-23 17-41

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
@@ -28,6 +28,7 @@ import java.awt.BasicStroke;
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.awt.Rectangle;
 import java.awt.RenderingHints;
 import java.awt.Stroke;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -54,84 +54,109 @@
...
+		final int startDrawX;
+		final int startDrawY;
+		if (getPreferredLocation() == null)
 		{
-			return null;
+			//check the width of the client if we can draw properly
+			final int clientWidth;
+			switch (config.centerOrbs())
+			{
+				case MIDDLE_CANVAS:
```
Nightfirecat: This seems to share code with the default--if that's the plan going forward for this configuration, remove the code and tack it onto the default case:

```
case MIDDLE_CANVAS:
default:
    ...
```

deathbeam: Well i just copy-pastad the code so idc mostly about it. I can clean it up later, the entire plugin is mess atm anyway

PR #1952, opened 2018-04-26
Add config for showing username in window title

SoyChai:
Fixes #1918 

Solved my own issue as @rherms was working on #1951. Either solution is probably equally effective, this one just includes a config to make the behaviour optional.

image

PR #4465, opened 2018-07-23
Add toggle for displaying username in title

### LINKED ISSUE #1918: Display username as part of the window title
This would help identify different RuneLite instances in streaming apps, and also in the taskbar / dock. An option to choose between appending and prepending the text would be valuable.

deathbeam:
Closes #1918
Closes #1952

Preview:
screenie

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.ui;
 
+import com.google.common.base.Strings;
 import com.google.common.eventbus.Subscribe;
 import java.applet.Applet;
 import java.awt.Canvas;
@@ -42,6 +43,7 @@ import java.awt.event.KeyEvent;
 import java.awt.image.BufferedImage;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import javax.inject.Provider;
 import javax.inject.Singleton;
 import javax.swing.BoxLayout;
 import javax.swing.ImageIcon;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -231,6 +235,40 @@ public void onNavigationButtonRemoved(final NavigationButtonRemoved event)
 		});
 	}
 
+	@Subscribe
+	public void onGameTick(final GameTick event)
```
Nightfirecat: Why is this done every game tick instead of on game state changed?

deathbeam: Because on login you dont have username yet

Nightfirecat: Huh.. Would've thought you could get it once game state changed to LOGGED_IN. Alright then. :man_shrugging:

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -231,6 +235,40 @@ public void onNavigationButtonRemoved(final NavigationButtonRemoved event)
...
+	{
+		if (!(client instanceof Client))
+		{
+			return;
+		}
+
+		if (lastName != null && !config.usernameInTitle())
```
Nightfirecat: Why is `lastName != null` part of this check?

deathbeam: To only update the frame title once

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -680,6 +726,20 @@ private void updateFrameConfig(boolean updateResizable)
...
+			final Player player = ((Client)client).getLocalPlayer();
+
+			if (player != null && player.getName() != null)
+			{
+				frame.setTitle(properties.getTitle() + " - " + player.getName());
+			}
+		}
+		else
+		{
+			frame.setTitle(properties.getTitle());
```
Nightfirecat: I'd like to see the string defined before doing the `config.usernameInTitle()` check, and this else block removed. That way, only one `frame.setTitle()` call is needed.

deathbeam: but in first if the .setTitle is not always called, because title do not needs updating (as there are no valid data to update the title with)

Nightfirecat: Hm... I suppose that's true. Would be nice if we could find a way to avoid the double-call, but I guess that's not possible here.

PR #4468, opened 2018-07-23
Fix level 126 next goal to be max XP (200,000,000) instead of -1 xp

### LINKED ISSUE #3658: Xp tracker doesn't track xp gained after hitting virtual level 126 in the skill
You can see in the pic I'm gaining smithing xp but it's not appearing on the tracker. This happened after hitting virtual level 126.

image

LeviSchuck:
This is a fix for endLevelExp being set to -1, which then causes an exception later when creating a snapshot for the GUI. That exception is thrown in `Experience.getLevelForXp` on a `< 0` xp amount.

The fix is to use the max XP value, which also is convenient for those that want to hit the max XP in the game, namely 200,000,000.

image

image

Used `::addxp fishing <big number here>` to test this.

Resolves #3658

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
@@ -220,7 +220,9 @@ class XpStateSingle
 		if (goalEndXp <= 0 || currentXp > goalEndXp)
 		{
 			int currentLevel = Experience.getLevelForXp(currentXp);
-			endLevelExp = currentLevel + 1 <= Experience.MAX_VIRT_LEVEL ? Experience.getXpForLevel(currentLevel + 1) : -1;
+			endLevelExp = currentLevel + 1 <= Experience.MAX_VIRT_LEVEL
+				? Experience.getXpForLevel(currentLevel + 1)
+				: Experience.MAX_SKILL_XP;
 		}
 		else
 		{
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -220,7 +220,9 @@ boolean update(int currentXp, int goalStartXp, int goalEndXp)
 		if (goalEndXp <= 0 || currentXp > goalEndXp)
 		{
 			int currentLevel = Experience.getLevelForXp(currentXp);
-			endLevelExp = currentLevel + 1 <= Experience.MAX_VIRT_LEVEL ? Experience.getXpForLevel(currentLevel + 1) : -1;
+			endLevelExp = currentLevel + 1 <= Experience.MAX_VIRT_LEVEL
+				? Experience.getXpForLevel(currentLevel + 1)
+				: 200000000;
```
Nightfirecat: `200_000_000`

Nightfirecat:
It would be kind of nice to have a visual indicator that you're working toward `MAX` instead of the start and end level goals being identical.

MagicfTail:
200 would be a nice max ^

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -248,6 +252,8 @@ XpSnapshotSingle snapshot()
 			.actionsRemainingToGoal(getActionsRemaining())
 			.actionsPerHour(getActionsHr())
 			.timeTillGoal(getTimeTillLevel())
+			.startLevelLabel("Lvl. " + startLevel)
```
deathbeam: I would rather propagate startLevelXp and endLevelXp and create labels from UI, I dont think creating of labels should be located here

LeviSchuck:
Made updates @deathbeam

PR #3668, opened 2018-06-07
Improve client creation

deathbeam:
## Move rs-client related classes to own package

To properly differentiate between client loading logic, move these
classes to own package.

## Allow RS loading classes to be used with Guice

- Provide OkHttpClient through Guice instead of using it from static
variable
- Chain the RS classes with Guice and inject ClientLoader to RuneLite
class
- Cleanup RS loading classes a bit

## Inject and load RS client and applet through Guice

Instead of requiring to create client from RuneLite just use Guice for
it and separate applet provider.

## Remove recursive dependency on RuneLite

Extract RuneLite#shutdown to separate class so RuneLite do not depends
on ClientUI and ClientUI do not depends on RuneLite.

Signed-off-by: Tomas Slusny <slusnucky@gmail.com>

PR #3497, opened 2018-05-30
Add a splash screen to RuneLite

jplsek:
## About
Although the client is still fast when starting up, "nothing" happens for several seconds. Adding a splash screen at least lets the user know that "something" is happening, and not try to open the client multiple times. This is more noticeable when the api is acting slow or when users have a slow connection. The current implementation does not seem to affect launch time.

## API
- All public methods are non-blocking.
- There is an argument to disable the splash screen.

I'm not sure how the API should act. Right now, it's part of `init`. Should the `setMessage` calls be inside the individual methods?

## UI
Updated:
updated4

Before:
<details>

Here is a full video of the splash screen (the inconsistent "RuneScape" capitals have been fixed):
normal

Slow connection simulation (1000ms delay):
slow
</details>

Feedback welcome. Thanks!

PR #4470, opened 2018-07-24
runelite-client: Open window after plugins are started

Abextm:
#3668 reordered client startup to show the window before plugins are loaded, which makes the client startup about 600ms slower. Moving pack() later also gives us a noticeable improvement in startup time, totaling roughly 1000ms

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -241,7 +241,7 @@ public class RuneLite
 		sessionManager.loadSession();
 
 		// Initialize UI
-		clientUI.open(this);
+		clientUI.init(this);
 
 		// Initialize Discord service
 		discordService.init();
@@ -272,6 +272,9 @@ public class RuneLite
 
 		// Start plugins
 		pluginManager.startCorePlugins();
+
+		// Show the window
+		clientUI.show();
 	}
 
 	public void shutdown()
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -280,7 +280,7 @@ public class ClientUI
 	 * @param runelite runelite instance that will be shut down on exit
 	 * @throws Exception exception that can occur during creation of the UI
 	 */
-	public void open(final RuneLite runelite) throws Exception
+	public void init(final RuneLite runelite) throws Exception
 	{
 		SwingUtilities.invokeAndWait(() ->
 		{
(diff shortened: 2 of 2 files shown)


deathbeam:
600ms is not much when you have visual indicator sooner instead. But, simple reorder of the startup sequence causing slower startup do not makes much sense, any idea why that happens?

Abextm:
I would assume its mostly redraw and relayout

Abextm:
And #3497 is a much better way to show that something is happening than showing a partially built UI.

deathbeam:
The only redraw happening should be sidebar and that one should be only partial and there should be no relayout, so that is kind of weird. Maybe but showing client ui sooner was less effort.

Abextm:
I don't think moving the code to where it was before is much effort, and
even if it was more complicated that copy and paste I think it would still
be worth it.

deathbeam:
No i mean showing the ClientUI sooner is less effort than adding loading screen :d Anyway, I tried this and its working, maybe it would solve the issue instead of partially reverting the change? 

```diff
diff --git a/runelite-client/src/main/java/net/runelite/client/ui/ClientPluginToolbar.java b/runelite-client/src/main/java/net/runelite/client/ui/ClientPluginToolbar.java
index fb7a112607..8e7c7b6539 100644
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientPluginToolbar.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientPluginToolbar.java
@@ -66,7 +66,6 @@ public class ClientPluginToolbar extends JToolBar
                                add(component, index);
                        }

-                       revalidate();
                        repaint();
                }
        }
@@ -78,7 +77,6 @@ public class ClientPluginToolbar extends JToolBar
                if (component != null)
                {
                        remove(component);
-                       revalidate();
                        repaint();
                }
        }
```

Abextm:
It still needs to be revalidated for when plugins add panels after startup

deathbeam:
Well i tested adding panels after startup and it works even without revalidation,

deathbeam:
Both #4479 and this PR have approximately same effect for me:

```
2018-07-24 19:06:23 [main] INFO  net.runelite.client.RuneLite - Startup time 2413
2018-07-24 19:06:35 [main] INFO  net.runelite.client.RuneLite - Startup time 2395
2018-07-24 19:06:45 [main] INFO  net.runelite.client.RuneLite - Startup time 2228
2018-07-24 19:06:56 [main] INFO  net.runelite.client.RuneLite - Startup time 2379
```

Both are measured using simple diff between 2 times in RuneLite in affected area:

```diff
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Locale;
 import javax.annotation.Nullable;
 import javax.inject.Provider;
@@ -240,6 +242,8 @@ public class RuneLite
                // Load the session, including saved configuration
                sessionManager.loadSession();

+               Instant start = Instant.now();
+
                // Initialize UI
                clientUI.init(this);

@@ -275,6 +279,9 @@ public class RuneLite

                // Show the window
                clientUI.show();
+
+               Instant end = Instant.now();
+               log.info("Startup time {}", Duration.between(start, end).toMillis());
        }
```

PR #3879, opened 2018-06-18
Add prayer bar to prayer plugin

deathbeam:
Add optional prayer bar under HP bar to prayer plugin.
Reimplemented from and closes #2549

Preview:
screenie

screenshot 2018-07-09 00 24 40

PR #4472, opened 2018-07-24
Add prayer bar color configuration options

nfiniteecho:
Adds options to #3879 to configure fill and background color of prayer bar.  

GIF

Add new default colors of cyan (fill) and black (background) for ease of visibility. The old default colors of cyan and white were hard to see with high brightness and with colorblindness.


deathbeam:
The prayer was was not made configurable on purpose, because its showing prayer and it should have specific color. But I like the black background with cyan here as default more, looks better.

nfiniteecho:
@deathbeam Yeah that makes sense generally to make it one color, but I think color customization options would be nice to have for those that may want them.  Prayer doesn't really have one defined color: it's meter is blue in the status orb, green potion for prayer pots, and a pink potion for super restores.

Not the biggest deal to leave out these customization options because I think a default black/cyan color scheme will fix most visibility problems.  What do you think?

deathbeam:
Sounds good.

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerConfig.java
```diff
@@ -32,87 +32,87 @@
 public interface PrayerConfig extends Config
 {
 	@ConfigItem(
-		position = 0,
-		keyName = "prayerflick",
-		name = "Prayer flick helper",
-		description = "Enable the prayer flick helper"
+			position = 0,
```
deathbeam: Do not reindent the fields here, see https://github.com/runelite/runelite/wiki/Code-Conventions

Nightfirecat:
This was added in #5635.

PR #4476, opened 2018-07-24
Remove .revalidate() from ClientPluginToolbar

deathbeam:
This call is completely unnecessary and removing it is saving some
milliseconds from startup time.


deathbeam:
See also #4470

Abextm:
This has no effect on startup time for me

Adam-:
If it is unnecessary it should be removed, regardless, right?

deathbeam:
Closing as #4507 has this

PR #4477, opened 2018-07-24
Split the Reset menu entry to two seperate checkboxes

Telavus:
Split the Reset menu entry to two seperate checkboxes, one for Reset and one for Lay to better cater for seperate training methods.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -443,7 +443,7 @@ public class MenuEntrySwapperPlugin extends Plugin
 		{
 			swap("reset", option, target, true);
 		}
-		else if (config.swapBoxTrap() && option.equals("take"))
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -443,7 +443,7 @@ else if (config.swapBoxTrap() && (option.equals("check") || option.equals("disma
 		{
 			swap("reset", option, target, true);
 		}
-		else if (config.swapBoxTrap() && option.equals("take"))
+		else if (config.layBoxTrap() && option.equals("take"))
```
Nightfirecat: :-1: If `layBoxTrap` is set off, the old behavior of `swapBoxTrap` being enabled is not maintained.

Telavus: Will fix that. It was kinda sloppy and hastily made for my own needs, so I didnt care about testing 👎

PR #4478, opened 2018-07-24
Properly determine fullscreen for OSX

deathbeam:
Properly determine if window is fullscreen on OSX

### DIFF
--- a/runelite-client/pom.xml
+++ b/runelite-client/pom.xml
@@ -122,7 +122,6 @@
 			<groupId>net.runelite</groupId>
 			<artifactId>orange-extensions</artifactId>
 			<version>1.0</version>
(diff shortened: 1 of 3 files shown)


runelite-client/pom.xml
```diff
@@ -122,7 +122,6 @@
 			<groupId>net.runelite</groupId>
 			<artifactId>orange-extensions</artifactId>
 			<version>1.0</version>
-			<scope>provided</scope>
```
Adam-: This will probably explode if not `provided`

deathbeam: It explodes when provided too, not sure what to do about it.

deathbeam:
I guess this is not so important atm

PR #4481, opened 2018-07-24
Add simple info log with startup time

deathbeam:

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -32,6 +32,8 @@ import com.google.inject.Guice;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
 import java.io.File;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


Adam-:
https://google.github.io/guava/releases/18.0/api/docs/com/google/common/base/Stopwatch.html

deathbeam:
I dont think stopwatch is really necessary?

Abextm:
Seems like it would be better to use `RuntimeMXBean::getUptime`

deathbeam:
Doesnt that include also startup time of JVM? Never actually used that property and there is not much documentation for it.

Abextm:
I would assume it includes jvm startup time. That is part of our startup time too.

deathbeam:
So it would also include launcher startup time when the reflection launcher will be used, I am not sure that would be ideal. Mainly because when JVMLauncher will be used it will not account for that and so the results will vary.

Abextm:
It seems like you want to measure init time, rather than startup time

deathbeam:
I guess. I just want to measure how long it takes to create entire RuneLite module and start it.

Abextm:
Both seem like useful metrics, so maybe it should be both.

deathbeam:
Added also display of uptime

PR #4486, opened 2018-07-24
Added Guard pickpocket entry swap

Nalerient:
> edit: jagex compliances... Swapped talk-to with pickpocket for H.A.M. guards, should be 69% compliant in the current format. Once Npc attack option is disabled via ingame options, left click pickpocketing should work.

untitled

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -143,7 +143,7 @@ public interface MenuEntrySwapperConfig extends Config
 
 	@ConfigItem(
 		position = 10,
-		keyName = "swapPickpocket", 
+		keyName = "swapPickpocket",
 		name = "Pickpocket on H.A.M.",
 		description = "Swap Talk-to with Pickpocket on H.A.M members"
 	)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -143,9 +143,9 @@ public interface MenuEntrySwapperConfig extends Config
 
 	@ConfigItem(
 		position = 10,
-		keyName = "swapPickpocket",
-		name = "Pickpocket H.A.M Guards and Members",
-		description = "Swap Talk-to and Attack with Pickpocket for H.A.M Guards and members"
+		keyName = "swapPickpocket", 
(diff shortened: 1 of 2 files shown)


Nightfirecat:
From https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features:

> ####Menu Entry Swapping
> - Pickpocket on non-H.A.M. members: Not supported by other 3rd party clients, and absurdly powerful when blackjacking

ShaunDreclin:
It's supposed to be swapping the talk-to option, not attack.

Wouldn't it be better to just do `if (config.swapPickpocket() && target.contains("h.a.m.") || target.contains("guard")` ?

Nalerient:
My bad, wasn't fully familiar with Jagex compliances. Thanks for pointing that out.

Nightfirecat:
What is the left-click option for HAM guards when NPC attack option is hidden? The reason this feature was previously limited to HAM members is because they were the only NPC which were frequently pickpocketed and did not have a talk-to option which became the left-click when NPC attack options were disabled. (eg. It became a convenience feature to not require changing NPC attack options rather than a feature which provided something unattainable in the vanilla client)

ShaunDreclin:
If you hide attack, left click on ham guards becomes talk-to

I don't think extending the functionality to ham guards would be that overpowered, the main concern was with blackjacking

Nalerient:
Still, kinda would love the left click pickpocket for H.A.M. guards tho.

Nalerient:
Sounds good to me. Got an elegant implementation under that furcout of yours? I know very little of runelite api, so yeah...

Nalerient:
Correct me if im wrong but there are only 2 types of H.A.M. npc's with pickpocket option, rendering your `(target.contains("h.a.m.")` part pointless under `if (option.equals("attack"))` statement returning us back to square one :^

ShaunDreclin:
I mean change that line instead of the change you made to the file, cause the option is to swap talk-to with pickpocket, not attack with pickpocket

Nalerient:
My misunderstanding. Thanks for pointing those things out, perhaps this solution will please mod ash

Adam-:
Pickpocket swapping is iffy since it impacts (afk?) exp rates. So far we have only swapped what other clients swap. I probably do not want to swap anything else pickpocket related.

ShaunDreclin:
>You can get left-click pickpockets with HAM members by changing attack settings.

You actually can't. That's why we have it in the plugin in the first place. In the vanilla client it's not possible to left click pickpocket ham members

jsmoa921:
Would be nice to have this feature

Trevor159:
This should be closed as it falls under `Modifies menu options for pickpocketing, such as swapping 'Pickpocket' and 'Talk-to'`  here https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1

PR #4507, opened 2018-07-26
Use consistent API and order for title and sides

deathbeam:
Use same API for ordering title and sidebar and order them in the swing
components and not in events as that is unreliable.

Closes #4476


Adam-:
Does this change the titlebar button order back to pre-1.4.9 or just make it deterministic?

Abextm:
Is there a reason that the title toolbar buttons are no longer separated from the plugin toolbar buttons?

deathbeam:
In sidebar with custom window chrome disabled? Hmm.., I can fix that. Should be same as pre-1.4.9 @Adam-

deathbeam:
Updated

runelite-client/src/main/java/net/runelite/client/ui/ClientPluginToolbar.java
```diff
@@ -53,33 +61,38 @@
...
 	}
 
-	public void removeComponent(final NavigationButton button)
+	private void update()
 	{
-		final Component component = componentMap.remove(button);
+		removeAll();
 
-		if (component != null)
+		final AtomicBoolean isDelimited = new AtomicBoolean(false);
```
Abextm: Using AtomicX it typically a good hint that you shouldn't use a lambda

deathbeam:
Updated

PR #3581, opened 2018-06-02
Adds in support for birdhouse timers for #1224

jlajoie:
## Description
Implemented for #1224

Adds in a timer based system utilizing seconds since epoch when adding seeds to a placed birdhouse based on the birdhouse type, and a constant of 50 minutes to calculate when birdhouses are done collecting.

Currently does not support adding only partial seeds (anything other than 10), though could likely add that based on the number of "Hops" seeds in the inventory and doing a diff when the appropriate VarP changes.

## Notes
Keeps the Tab model to allow for any additional Hunter timer based activities in the future.

Waits 10 ticks before trying to detect changes to Birdhouse Space VarP values, as it takes a second to update on logging in. (e.g. logging in at any of the regions would result in a VarP value of 0, before transitioning to the "full" state which would reset the timer(
This could also be solved by:
* Verifying seed decrements in inventory when going to a "Full" state
* Disallowing a 0 -> "Full" state transition, as a "Built" step is required in between
* Disallowing a "Full" to 0 state transition if no Hunter XP is gained.

Modeled after the Farming Tracker Plugin, follows same style hierarchy that results in
* World
  * Region
    * Space
      * Implementation
        * State
          * Birdhouse (type)

PR #259, opened 2017-12-15
Add a stopwatch plugin

XrioBtw:
A stopwatch plugin displayed on the report button
img001

PR #4510, opened 2018-07-26
Add time tracking plugin

### LINKED ISSUE #1224: Hunter Plugin: Add support for birdhouses
Either timers based on when you created them, or use varbits if there are any

### LINKED ISSUE #1612: Request: Add user-configurable timers/stopwatches
I've found much use of a plugin that provided user-configurable stopwatches and timers in OSBuddy. See image below for their implementation.

OSBuddy timers and stopwatches sidebar plugin

takuyakanbr:
**Changes**

- Convert farming tracker to time tracking plugin
  - Flower, hops, and grape patches now have their own tabs
- Add bird house tracker and a new bird house tab
  - The plugin can optionally send a notification when all bird houses are ready to be dismantled
- Add a new clock tab that contains user-defined timers and stopwatches
  - Timers and stopwatches continue 'running' when client is closed
  - Timers can optionally send a notification when they finish counting down
- Add a new overview tab that provides a quick summary of the various tabs
- Add tooltips to tab icons

**Screenshots**

Plugin icon
image0

Overview tab
overview

Timers and stopwatches
image1

Bird house tracker
image2

**Notes**

All existing farming tracker data will be migrated over to the new plugin. No data is expected to be lost.

I would suggest that the migration code (`FarmingTracker#migrateConfiguration`) should be removed after around 2 to 4 weeks.

Closes #259
Closes #3581
Fixes #1224
Fixes #1612


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/Tab.java
```diff
@@ -32,12 +32,17 @@
...
+	HERB("Herb Patches", ItemID.GRIMY_RANARR_WEED),
+	TREE("Tree Patches", ItemID.YEW_LOGS),
+	FRUIT_TREE("Fruit Tree Patches", ItemID.PINEAPPLE),
+	FLOWER("Flower Patches", ItemID.RED_FLOWERS),
+	HOPS("Hops Patches", ItemID.BARLEY),
+	BUSH("Bush Patches", ItemID.POISON_IVY_BERRIES),
+	GRAPE("Grape Patches", ItemID.GRAPES),
+	SPECIAL("Special Patches", ItemID.MUSHROOM);
+
+	public static final Tab[] FARMING_TABS = {ALLOTMENT, HERB, TREE, FRUIT_TREE, FLOWER, HOPS, BUSH, GRAPE, SPECIAL};
```
Adam-: Prefer using `Tab.values()` instead

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/hunter/BirdHouse.java
```diff
@@ -0,0 +1,67 @@
...
+	OAK("Oak Bird House", ItemID.OAK_BIRD_HOUSE),
+	WILLOW("Willow Bird House", ItemID.WILLOW_BIRD_HOUSE),
+	TEAK("Teak Bird House", ItemID.TEAK_BIRD_HOUSE),
+	MAPLE("Maple Bird House", ItemID.MAPLE_BIRD_HOUSE),
+	MAHOGANY("Mahogany Bird House", ItemID.MAHOGANY_BIRD_HOUSE),
+	YEW("Yew Bird House", ItemID.YEW_BIRD_HOUSE),
+	MAGIC("Magic Bird House", ItemID.MAGIC_BIRD_HOUSE),
+	REDWOOD("Redwood Bird House", ItemID.REDWOOD_BIRD_HOUSE);
+
+	static final BirdHouse[] BIRD_HOUSES = {NORMAL, OAK, WILLOW, TEAK, MAPLE, MAHOGANY, YEW, MAGIC, REDWOOD};
```
Adam-: `Birdhouse.values()`

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
```diff
@@ -108,18 +121,44 @@ protected void startUp() throws Exception
 
 		pluginToolbar.addNavigation(navButton);
 
+		timer = new Timer();
+		timer.scheduleAtFixedRate(new TimerTask()
+		{
+			@Override
+			public void run()
+			{
+				updatePanel();
```
Adam-: We do have a `@Schedule` (https://static.runelite.net/api/runelite-client/net/runelite/client/task/Schedule.html) although I do not think it provides sub-tick accuracy currently. I would either maybe modify that, or at least inject the scheduler and schedule it with that. Each java `Timer` is always backed by a java `Thread` which is not great.

Updating swing panels often is a performance issue and even now we have users report that when the farming panel is open (thus, when it updates) it causes noticeable performance issues. Can we make this not update if there are no timesr running? Or if it isn't open?

Abextm:
- Partially seeded birdhouses varp%3==2 should probably show as BUILT, not UNKNOWN
 - Birdhouses should probably use the same color code as farming patches to be consistent
 - When I was messing with birdhouses they did tick similar to farming patches, but at a relatively fast rate, so simply adding 50 minutes to the setup time was not entirely accurate, but I'm not sure if there is anything we can really do about that. 
 - Logging in on fossil island resets the birdhouse times (this bug also applies to farming patches, but it doesn't really matter because they just get fixed when you finish logging in). The best way to fix this I think would be to make sure the Click here to continue screen widget isn't showing, and wasn't showing the previous tick. It would also be nice to add the plane check to farming patches because they need it too.
 - The timer UI is rather clunky, perhaps it would be better to make the shown time a input. If this isn't possible I would still like to have something that doesn't open a dialog box
 - Same thing with rename, it could just be a text field that activates on focus without being as modal

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
takuyakanbr: Even if I decrease the tick interval of the `Scheduler`, I can't currently use it because `Hooks#clientMainLoop` does not run when idling on the login screen. 

Only 1 `Timer` is created and used when the plugin is active, which imo is insignificant in terms of resource consumption. (If and when another plugin requires a timer, it can perhaps be moved into `Scheduler` or another class.)

None of the tabs and their contents are being updated when the tab isn't open. (Previously, all patch panels were being updated when updating the UI.) I have also updated it such that while on the clock tab, no UI updates are being done if there are no active clocks.

takuyakanbr:
Added an overview tab (see first comment for screenshot) and addressed above comments.

deathbeam:
I would put birdhouses before farming as its single-entry compared to x amount of different entries for farming

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockTabPanel.java
```diff
@@ -0,0 +1,235 @@
...
+
+			synchronized (ImageIO.class)
+			{
+				deleteIcon = ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("delete_icon.png"));
+				lapIcon = ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("lap_icon.png"));
+				pauseIcon = ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("pause_icon.png"));
+				resetIcon = ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("reset_icon.png"));
+				startIcon = ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("start_icon.png"));
+				ADD_ICON = new ImageIcon(ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("add_icon.png")));
+				ADD_ICON_HOVER = new ImageIcon(ImageIO.read(TimeTrackingPlugin.class.getResourceAsStream("add_hover_icon.png")));
```
Nightfirecat: What's the reason for two separate images here instead of using `SwingUtil::grayscaleOffset` like the other hovered icon images?

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
Adam-: I would maybe prefer if you left the scheduler alone and scheduled it on the `ScheduledExecutorService` which you can inject.

takuyakanbr:
Added notifications for bird house tracker and farming tracker. (Most of the necessary infrastructure was already added to support the overview tab anyway.) Also addressed all pending comments.

For the bird house tracker, I am now blocking transitions from 3 or 4 non-empty bird houses to 4 empty bird houses, since it seems that the client might occasionally not receive the varps in a timely manner.

The code should be ready for reviewing now.

Abextm:
- There is definitely something wrong with the prediction now, my trees which should be done on Sunday are saying Thursday, thought their progress bar is correct. My birdhouses I just planted (on Saturday) say they will be done on Wednesday.
 - My tree patches say ready, even though my hardwoods are not (spirit trees are all done, and normal trees are not planted)
 - Bird houses still can reset during the login process, the issue is not that the server doesn't send the varps in a consistent time, it is that it only sends them the tick after you exit the "click here to continue" splash screen
 - There doesn't seem to be any feedback when a timer expires, which makes them somewhat useless imo
 - Timers should automatically reset when you start them.

takuyakanbr:
Updated to resolve the above issues.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/hunter/BirdHouseData.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package net.runelite.client.plugins.timetracking.hunter;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+/**
+ * Contains data about the state of a particular {@link BirdHouseSpace}, at a particular point in time.
+ */
+@AllArgsConstructor
```
Abextm: `@AllArgsConstructor` is redundant here

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
```diff
@@ -180,13 +222,19 @@ public void checkCompletion()
...
 	}
 
-	@Schedule(period = 10, unit = ChronoUnit.SECONDS)
-	public void updatePanel()
+	private void updatePanel()
 	{
-		SwingUtilities.invokeLater(panel::update);
+		long unitTime = Instant.now().toEpochMilli() / 200;
+
+		boolean clockDataChanged = unitTime % 5 == 0 && clockManager.checkCompletion();
```
Abextm: Non functional methods should not be called with short circuiting operators.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockTabPanel.java
```diff
@@ -0,0 +1,239 @@
...
+	/**
+	 * Clears and recreates the components of this panel.
+	 * This should be done whenever a clock is added or removed.
+	 */
+	void rebuild()
+	{
+		removeAll();
+		clockPanels.clear();
+
+		GridBagConstraints c = new GridBagConstraints();
```
Abextm: DynamicGridLayout is preferred if possible

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -56,6 +58,15 @@
...
 	private FarmingWorld farmingWorld;
 
+	/**
+	 * The time at which all patches of a particular type will be ready to be harvested,
+	 * or {@code -1} if we have no data about any patch of the given type.
+	 *
+	 * Each value is set to {@code 0} if all patches of that type have already completed
+	 * when updating the value.
+	 */
+	private Map<PatchImplementation, Long> completionTimes = new HashMap<>();
```
Abextm: EnumMap

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/hunter/BirdHouseTracker.java
```diff
@@ -0,0 +1,210 @@
...
+	 * or {@code -1} if we have no data about any of the bird house spaces.
+	 *
+	 * This is set to {@code 0} if the bird houses have already completed
+	 * when updating it.
+	 */
+	private long completionTime = -1;
+
+	public BirdHouseTabPanel createBirdHouseTabPanel()
+	{
+		return new BirdHouseTabPanel(itemManager, this, config);
```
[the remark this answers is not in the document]
takuyakanbr: That will cause the UI to screw up. Looks like guice creates it before the UI theme has been set?
a1

takuyakanbr:
Made requested changes.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/hunter/BirdHouseTracker.java
```diff
@@ -0,0 +1,223 @@
...
+	private ConfigManager configManager;
+
+	@Inject
+	private TimeTrackingConfig config;
+
+	@Inject
+	private Notifier notifier;
+
+	@Getter(AccessLevel.PACKAGE)
+	private final Map<BirdHouseSpace, BirdHouseData> birdHouseData = new HashMap<>();
```
Abextm: This is accessed from multiple threads

takuyakanbr: Switched to `ConcurrentHashMap`

Abextm: Its still not thread safe because you iterate over it

takuyakanbr: All updates and iterations happen on the same (client) thread though. The UI thread uses only `Map#get`.

deathbeam:
Needs rebase

deathbeam:
Also, the grayscaleOffset etc methods were moved to ImageUtil, so that needs to be updated here too.

takuyakanbr:
Made requested changes.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
```diff
@@ -0,0 +1,240 @@
...
+
+		clientToolbar.addNavigation(navButton);
+
+		panelUpdateFuture = executorService.scheduleAtFixedRate(this::updatePanel, 200, 200, TimeUnit.MILLISECONDS);
+	}
+
+	@Override
+	protected void shutDown() throws Exception
+	{
+		lastTickLocation = null;
```
deathbeam: set lastTickPostLogin to false here too

takuyakanbr: Updated

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockTabPanel.java
```diff
@@ -0,0 +1,191 @@
...
+	{
+		BufferedImage deleteIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "delete_icon.png");
+		BufferedImage lapIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "lap_icon.png");
+		BufferedImage pauseIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "pause_icon.png");
+		BufferedImage resetIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "reset_icon.png");
+		BufferedImage startIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "start_icon.png");
+		BufferedImage addIcon = ImageUtil.getResourceStreamFromClass(TimeTrackingPlugin.class, "add_icon.png");
+
+		DELETE_ICON = new ImageIcon(deleteIcon);
+		DELETE_ICON_HOVER = new ImageIcon(ImageUtil.grayscaleOffset(deleteIcon, -100));
```
Nightfirecat: The behavior of `grayscaleOffset` has changed somewhat due to my PR. Double-check to see if these should still be using this method (which now explicitly reduces RGB values, *and not alpha*) or if they should use `alphaOffset` instead.

PR #4547, opened 2018-07-28
Prevent crash when updating the fairy ring filter

Abextm:
Its amazing that this ever worked, and that I didn't think that even though the other callback breaks that this one might too.


deathbeam:
Needs rebase

PR #4551, opened 2018-07-29
Ancient timers revisited

devOSRS:
Ancient timers don't reset per https://github.com/runelite/runelite/pull/4300.

The following has been added:

1. All freeze timers are triggered upon "You have been frozen!" message, instead of graphic. Two situations where graphic shows up and you have not been frozen are: (1) barrage clumps- people running close by and immune targets that have been recently unfrozen, (2) when someone freezes you off the rendered map (i.e.: opponent runs in the opposite direction).
2. Graphics act as a filter. Upon successful freeze/triggering of timers in (1) above. Then the graphic event is checked and only the matching graphic/timer is kept. If no graphic shows up, **all timers are kept**. This was a design choice to keep all timers running as reference as we don't really know which spell was casted. After this all timers are kept until they run out
3. In the case that you have been frozen and the caster runs off the map (setting you free), we have two options to reset the timer: (1) check if the list of opponents attacking the player in the rendered map is empty, or (2) if the player is able to move. I chose the latter.

timers

EDIT: some typos


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -318,12 +314,10 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 	@Subscribe
 	public void onChatMessage(ChatMessage event)
 	{
-
 		if (event.getType() != ChatMessageType.FILTERED && event.getType() != ChatMessageType.SERVER)
 		{
 			return;
 		}
-
```
deathbeam: Why was this empty line removed?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -28,19 +28,9 @@
...
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemContainer;
-import net.runelite.api.ItemID;
-import net.runelite.api.NPC;
-import net.runelite.api.NpcID;
-import net.runelite.api.Prayer;
-import net.runelite.api.Varbits;
+
+import net.runelite.api.*;
```
deathbeam: Fix your imports

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -458,6 +452,31 @@ public void onChatMessage(ChatMessage event)
 		{
 			removeGameTimer(STAFF_OF_THE_DEAD);
 		}
+
+		if (config.showFreezes() && event.getMessage().equals("<col=ef1020>You have been frozen!</col>"))
+		{
+			createGameTimer(ICERUSH, false);
```
deathbeam: So this will create 4 different timers? That seems weird

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -458,6 +452,31 @@ public void onChatMessage(ChatMessage event)
...
+			createGameTimer(ICEBURST, false);
+			createGameTimer(ICEBLITZ, false);
+			createGameTimer(ICEBARRAGE, false);
+		}
+	}
+
+	@Subscribe
+	public void onGameTickChanged(GameTick event)
```
deathbeam: This is wrong naming for the method

devOSRS: `onGameTick`? If you are going to say its wrong could at least point me in the right direction? Saying what you said vs. "Change it to _______" takes the same amount of characters and is more productive.

Nightfirecat: Yes, `onEventName` is the convention. (almost) All existing examples should be in that format as well.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: Please read the PR

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
deathbeam: There is like 137 events that are wrongly named, but otherwise yeah :D , on + `EventClassName`

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: Will do

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: No idea

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: I have to agree with db here... I'd prefer a solution that assumes a freeze is of one type (Ice Barrage would probably be best) and corrects itself should a graphic of another type appears.

devOSRS: @Nightfirecat @deathbeam Freeze message defaults to barrage. If graphic appears, barrage timer is removed and correct freeze time is created. If no graphic appears, default timer is kept (barrage).

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -28,6 +28,7 @@
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
 import javax.inject.Inject;
+
```
deathbeam: Remove this newline (https://github.com/runelite/runelite/wiki/Code-Conventions check xml for intellij idea here)

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -91,15 +95,20 @@
...
 import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
+import lombok.extern.slf4j.Slf4j;
+
 
 @PluginDescriptor(
 	name = "Timers",
 	description = "Show various timers in an infobox",
 	tags = {"combat", "items", "magic", "potions", "prayer", "overlay", "abyssal", "sire"}
 )
+@Slf4j
```
deathbeam: This looks unused

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -563,24 +594,28 @@ public void onGraphicChanged(GraphicChanged event)
...
 			{
-				createGameTimer(ICEBLITZ);
+                removeGameTimer(ICEBARRAGE, 1);
+                createGameTimer(ICEBLITZ, false);
 			}
 
 			if (actor.getGraphic() == ICEBARRAGE.getGraphicId())
 			{
-				createGameTimer(ICEBARRAGE);
+				;
```
deathbeam: ? remove this condition now I guess, if it is not necessary

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
```diff
@@ -96,4 +96,9 @@ public boolean cull()
 		return timeLeft.isZero() || timeLeft.isNegative();
 	}
 
+	public long secondsPassed()
```
deathbeam: I dont think this method is necessary, you can just do that outside from method in plugin

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: Sure, wanted to be explicit as other people might look at the code

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
devOSRS: Agreed, don't want to change Timer.java. I am having trouble extracting time left without adding another function to timer. I still need to somehow get `Duration.between(Instant.now(), endTime).getSeconds()` from the Timer. I could call the get text and convert the value back into seconds?

deathbeam: Yeah, getter for endTime seems fine

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -458,6 +464,28 @@ public void onChatMessage(ChatMessage event)
 		{
 			removeGameTimer(STAFF_OF_THE_DEAD);
 		}
+
+		if (config.showFreezes() && event.getMessage().equals("<col=ef1020>You have been frozen!</col>"))
```
deathbeam: After #4072 has been merged, this string now should be constant at the top of the file

Nightfirecat: Personally, I'd just be happy with the string being a constant at the top of the file.

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
```diff
@@ -96,4 +96,9 @@ public boolean cull()
 		return timeLeft.isZero() || timeLeft.isNegative();
 	}
 
+	public Instant getEndTime()
```
deathbeam: Use lombok `@Getter`

devOSRS: Done :)

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -92,6 +98,7 @@
 import static net.runelite.client.plugins.timers.GameTimer.VENGEANCEOTHER;
 import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
 
+
```
deathbeam: Remove this empty line

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -125,6 +133,9 @@
 	private static final String SUPER_ANTIVENOM_DRINK_MESSAGE = "You drink some of your super antivenom potion";
 
 	private int lastRaidVarb;
+	private int lastWorldX;
+	private int lastWorldY;
+	private HashSet<GameTimer> freezeTimers;
```
deathbeam: Change HashSet<> to Set<> and use ImmutableSet.of here, and change this to static final FREEZE_TIMERS

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +702,32 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
 		}
 	}
 
-	private void createGameTimer(GameTimer timer)
+	/**
+	 * Adds a game timer to the infobox manager to render. Accepts a `replaceExisting` option to
```
Nightfirecat: Doc should be updated to explain the `checkActiveFreezeTimers` aspect of it.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -483,6 +494,28 @@ public void onChatMessage(ChatMessage event)
...
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		Player player = client.getLocalPlayer();
+		WorldPoint currentWorldPoint = player.getWorldLocation();
+
+		if (lastWorldX != currentWorldPoint.getX() || lastWorldY != currentWorldPoint.getY())
+		{
+			removeGameTimer(ICERUSH);
```
Nightfirecat: Make a method to remove ancient freeze timers, and don't call `removeGameTimer` every tick unless it's needed. (Pollutes debug log pretty heavily)

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -679,8 +736,35 @@ private void createGameTimer(GameTimer timer)
...
+		createGameTimer(timer, true);
+	}
+
 	private void removeGameTimer(GameTimer timer)
 	{
-		infoBoxManager.removeIf(t -> t instanceof TimerTimer && ((TimerTimer) t).getTimer() == timer);
+		removeGameTimer(timer, -1);
+	}
+
+	private void removeGameTimer(GameTimer timer, long timePassedThreshold)
```
Nightfirecat: This could use doc.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -588,25 +621,25 @@ public void onGraphicChanged(GraphicChanged event)
 				}
 			}
 
+			// threshold to allow frozen graphic to register and remove redundant timers
 			if (actor.getGraphic() == ICERUSH.getGraphicId())
 			{
-				createGameTimer(ICERUSH);
+				removeGameTimer(ICEBARRAGE, 1);
```
Nightfirecat: `1` should be a constant.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -588,25 +621,25 @@ public void onGraphicChanged(GraphicChanged event)
...
+				createGameTimer(ICERUSH, false, true);
 			}
 
 			if (actor.getGraphic() == ICEBURST.getGraphicId())
 			{
-				createGameTimer(ICEBURST);
+				removeGameTimer(ICEBARRAGE, 1);
```
Nightfirecat: `1` should be a constant.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -588,25 +621,25 @@ public void onGraphicChanged(GraphicChanged event)
...
+				createGameTimer(ICEBURST, false, true);
 			}
 
 			if (actor.getGraphic() == ICEBLITZ.getGraphicId())
 			{
-				createGameTimer(ICEBLITZ);
+				removeGameTimer(ICEBARRAGE, 1);
```
Nightfirecat: `1` should be a constant.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +702,32 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+	 * timer.
+	 *
+	 * @param timer           			The timer to be added to the infobox manager.
+	 * @param replaceExisting 			Whether to remove existing timer of the same type. `true` to replace
+	 *                        			it, `false` to preserve it if present.
+	 * @param checkActiveFreezeTimers	Whether to check for active freeze timers before creating a new one,
+	 *                                  potentially of a different spell type (i.e.: blitz vs. barrage)
+	 */
+
+	private void createGameTimer(final GameTimer timer, final boolean replaceExisting, final boolean checkActiveFreezeTimers)
```
Nightfirecat: Since this isn't a general createGameTimer method (can't be applied to other timer types), this might be better-suited as a differently-named method, and the `removeExisting` code moved back into `createGameTimer(GameTimer, boolean)`.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -679,8 +736,35 @@ private void createGameTimer(GameTimer timer)
...
+		long totalTimerSeconds = timer.getDuration().getSeconds();
+		if (timePassedThreshold >= 0)
+		{
+			infoBoxManager.removeIf(
+				t -> t instanceof TimerTimer
+				&& ((TimerTimer) t).getTimer() == timer
+				&& (totalTimerSeconds - Duration.between(Instant.now(), ((TimerTimer) t).getEndTime()).getSeconds()) <= timePassedThreshold
+			);
+		}
+		else
```
Nightfirecat: Is there any way we can remove this `if` statement? We have duplicated `t -> t instanceof TimerTimer && ((TimerTimer) t).getTimer() == timer` code as a result of it.

devOSRS:
@deathbeam I can make the edits @Nightfirecat suggests, want to make sure you are in agreement before I get started

deathbeam:
Yeah

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: We could, but that would come at an expense of computing `(totalTimerSeconds - Duration.between(Instant.now(), ((TimerTimer) t).getEndTime()).getSeconds()) <= timePassedThreshold` overtime we remove timers instead of only when a `timePassedThreshold` is defined, correct?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: Agreed. 

You could also argue that the signature `replaceExisting` is also only used for freeze timers as well. However, in the future we may want similar behavior for non-freezing timers and nothing in that function specifically checks for freezes, so doesn't need to be changed. Thoughts?

Nightfirecat: In that case, I'd want to see this become something more generic, like accepting a set of GameTimers that should be checked against.

devOSRS: Ok, about to push some code up!

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: How do you feel about checking `infoBoxManager` for existing timers, this way we don't call `removeTimer` unless there is something to remove? 

We could also keep a struct of active timers within `TimersPlugin` although this would require some housekeeping.

These may be an overkill since we don't want to do too much on every game tick, but not sure.

Nightfirecat: The first sentence is kind of what I was thinking--rather than directly calling `removeGameTimer` on each ice timer, check to see if they're active first, then remove them if they exist. The overhead should be fairly minimal since `removeGameTimer` would perform the same checks anyway.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -702,6 +704,30 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
 		}
 	}
 
+	/**
+	 * Adds a FREEZE game timer to the infobox manager to render. Accepts a `replaceExisting` option to
```
Nightfirecat: This doc needs updating.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: Ok cool, before pushing I tested the code and noticed a slight bug with this function. When running and frozen the timer doesn't show up until the next freeze cast (if at all). This could be an issue for hybrid fights especially. Not sure if this is because of WorldPoint (server/delayed) vs. LocalPoint (no delay). I'll try it out. 

We could also check a larger distance. If a player is immune for the remainder of the in game freeze timer (say if caster runs off screen) then we can entirely remove this function.

devOSRS: Local point sees same behavior

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +706,48 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+	 * @param replaceExisting 				Whether to remove existing timer of the same type. `true`
+	 *                                      to replace it, `false` to preserve it if present.
+	 * @param mutuallyExclusiveGameTimers	Mutually exclusive set of timers to be checked. If the timer
+	 *                                      at hand is in the set. It will not be created.
+	 */
+
+	private void createGameTimer(final GameTimer timer, final boolean replaceExisting, final Set<GameTimer> mutuallyExclusiveGameTimers)
+	{
+		if (!mutuallyExclusiveGameTimers.isEmpty()
+			&& infoBoxManager.getInfoBoxes().stream().anyMatch(t -> t instanceof TimerTimer && mutuallyExclusiveGameTimers.contains(((TimerTimer) t).getTimer())))
```
deathbeam: I would rather you to not use the .getInfoboxes() like this anywhere in this class. Instead add method .has(Function<InfoBox, Timer>) to infoboxManager and use that

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: To my knowledge, that behavior is caused by PID. I believe that if you have it, you will be able to run for one additional tick before the freeze actually takes effect.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
devOSRS: done!

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +706,48 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
 		}
 	}
 
-	private void createGameTimer(GameTimer timer)
+	/**
+	 * Adds a FREEZE game timer to the infobox manager to render. Accepts a `replaceExisting` option to
```
Nightfirecat: Doc needs updating.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -679,8 +756,44 @@ private void createGameTimer(GameTimer timer)
...
+	}
+
 	private void removeGameTimer(GameTimer timer)
 	{
-		infoBoxManager.removeIf(t -> t instanceof TimerTimer && ((TimerTimer) t).getTimer() == timer);
+		removeGameTimer(timer, NO_TIME_PASSED_THRESHOLD);
+	}
+
+	/**
+	 * Adds a game timer to the infobox manager to render. Accepts a `replaceExisting` option to
```
Nightfirecat: This doc also looks wrong... `removeGameTimer` doesn't add game timers, and doesn't accept a `replaceExisting` option.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -679,8 +756,44 @@ private void createGameTimer(GameTimer timer)
...
+	 *                                  events.
+	 *
+	 *                                  i.e.: default freeze timer is BARRAGE upon
+	 *                                  frozen message, but graphic indicates BLITZ. This
+	 *                                  threshold allows for the removal of BARRAGE. On subsequent
+	 *                                  freeze spell casts the number of seconds passed should
+	 *                                  be greater than this threshold and should not allow the
+	 *                                  current freeze timer to be removed, as it is in effect.
+	 */
+
```
Nightfirecat: Remove this empty line.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +706,48 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+	/**
+	 * Adds a game timer to the infobox manager to render. Accepts a `replaceExisting` option to
+	 * preserve existing timers of the same type instead of replacing them with the new passed
+	 * timer.
+	 *
+	 * @param timer           			The timer to be added to the infobox manager.
+	 * @param replaceExisting 			Whether to remove existing timer of the same type. `true` to replace
+	 *                        			it, `false` to preserve it if present.
+	 */
+
```
Nightfirecat: Remove this empty line.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -669,8 +706,48 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+	 * timer. Accepts a `checkActiveFreezeTimers` option to preserve existing FREEZE timers of a certain
+	 * freeze spell instead of replacing or adding additional freeze timers of a different freeze spell.
+	 *
+	 * @param timer           				The timer to be added to the infobox manager.
+	 * @param replaceExisting 				Whether to remove existing timer of the same type. `true`
+	 *                                      to replace it, `false` to preserve it if present.
+	 * @param mutuallyExclusiveGameTimers	Mutually exclusive set of timers to be checked. If the timer
+	 *                                      at hand is in the set. It will not be created.
+	 */
+
```
Nightfirecat: Remove this empty line.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -588,25 +625,25 @@ public void onGraphicChanged(GraphicChanged event)
...
+				removeGameTimer(ICEBARRAGE, FREEZE_TIME_PASSED_THRESHOLD);
+				createGameTimer(ICEBURST, false, FREEZE_TIMERS);
 			}
 
 			if (actor.getGraphic() == ICEBLITZ.getGraphicId())
 			{
-				createGameTimer(ICEBLITZ);
+				removeGameTimer(ICEBARRAGE, FREEZE_TIME_PASSED_THRESHOLD);
+				createGameTimer(ICEBLITZ, false, FREEZE_TIMERS);
 			}
```
Nightfirecat: Remove this empty line.

Nightfirecat:
Also, this PR would close #3062

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -483,6 +496,30 @@ public void onChatMessage(ChatMessage event)
...
+		}
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		Player player = client.getLocalPlayer();
+		WorldPoint currentWorldPoint = player.getWorldLocation();
+
+		if (lastWorldX != currentWorldPoint.getX()
```
Nightfirecat: `&&` and `||` are not of the same precedence; `&&` is evaluated first.

```
java> true || true && false
java.lang.Boolean res0 = true
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -679,8 +754,41 @@ private void createGameTimer(GameTimer timer)
...
+	 *                                  events.
+	 *
+	 *                                  i.e.: default freeze timer is BARRAGE upon
+	 *                                  frozen message, but graphic indicates BLITZ. This
+	 *                                  threshold allows for the removal of BARRAGE. On subsequent
+	 *                                  freeze spell casts the number of seconds passed should
+	 *                                  be greater than this threshold and should not allow the
+	 *                                  current freeze timer to be removed, as it is in effect.
+	 */
+	private void removeGameTimer(GameTimer timer, long secondsPassedThreshold)
```
Nightfirecat: `long secondPassedThreshold` may be better off being `Duration timePassedThreshold`, but I think that's minor enough not to matter at the moment.

Adam-:
This logic could be simplified by just storing a `private GameTimer freezeTimer` in the plugin and referencing that.

Adam-:
Here is my attempt at this - https://github.com/Adam-/runelite/commit/abf27e7762507c523646015403614a1e3daadac5. Can you verify this still behaves as you expect?

devOSRS:
@Adam- Code looks good. Let me double check behavior.

devOSRS:
@Adam- works as expected in all the cases I talked about above! Way simpler solution :) still getting a hang of all the available classes to make use of!

Adam-:
Added in 720c5f018465f4e854dd0da8b056f735542c0b12

PR #4552, opened 2018-07-29
Fix Assembler tests when autocrlf=true on Windows

Shingyx:
Tests were failing on Windows for me when running `mvn install`. Fixed by replacing \r\n with \n when asserting the `expected` value.

Another workaround is disabling autocrlf in git for this project, but I feel it's not reasonable to expect all contributors on Windows to do this.

### DIFF
--- a/cache/src/test/java/net/runelite/cache/script/assembler/AssemblerTest.java
+++ b/cache/src/test/java/net/runelite/cache/script/assembler/AssemblerTest.java
@@ -75,7 +75,7 @@ public class AssemblerTest
 		in = AssemblerTest.class.getResourceAsStream(this.script);
 		Assert.assertNotNull(in);
 
-		String original = new String(IOUtils.toByteArray(in));
+		String original = new String(IOUtils.toByteArray(in)).replaceAll("\r\n", "\n");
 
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Alternatively, we could use a gitattributes file to set `eol=lf` for files where it's necessary.

deathbeam:
I dont think adjusting line endings just so our tests will pass makes sense, I like this solution way more.

deathbeam:
I would prefer to not use gitattributes as the rs2asm file should support cross-platform line endings. Just the test does not support it.

Shingyx:
Oh sorry I misunderstood. I didn't realise the rs2asm files should support cross-platform line endings. I think it's unlikely that other tests should be reading rs2asm like AssemblerTest does, so I think I'll revert my last change for now.

Abextm:
This should probably happen in the grammar, not as some prepossessing step

deathbeam:
what do you mean by that? for the dissasembler to preserve \r\n instead of just using \n?

Abextm:
Nevermind me, I didn't realize we fixed that the assembler didn't work with newlines at all

Issue #4326, opened 2018-07-14
Add Leather hide shields to crafting calculator


Solburst:
Leather hide shields don't appear in the crafting skilling calculator.

TheEnigmaThatIsMe:
Where are the item icon numbers found? Doesn't appear that LEATHER_SHIELD is in ItemID.java, unless its supposed to be HARD_LEATHER_SHIELD?

Solburst:
Sorry i don't understand code at all, just letting people who do code know
that it doesn't appear on the skilling calculator

On Mon, Jul 16, 2018 at 10:52 PM, George Gilmartin <notifications@github.com

rbbi:
@Solburst have you checked out the leatherhide-shields branch on my fork?

deathbeam:
Added in #4331

PR #4331, opened 2018-07-15
Add leather shields to the crafting calculator #4326

rbbi:
screenshot at jul 15 02-46-05

PR #4553, opened 2018-07-29
Add Leather hide shields to crafting calculator #4326

ryancatlin:
Added Hard Leather Shield & Snakeskin Shield to crafting skill calculator


Nightfirecat:
Isn't this basically just a subset of #4331?

PR #4555, opened 2018-07-29
Add loot tracker

### LINKED ISSUE #1235: Add monster loot logger
A plug in on the side that lists all loot you obtained. It will be useful for tracking boss loot and slayer task loot.

deathbeam:
Add loot tracker plugin that will track monsters killed in this session.

Closes #1235

Preview:
screenie2
screenie


Kamielvf:
What do you think of "Total kills" and "Total value" instead of the killed/profit.
I would prefer to not use profit here at least.

deathbeam:
Maybe even total count and total value as it is also supposed to count clues etc.

deathbeam:
Updated and changed `Profit` and `Killed` to `Total count` and `Total value`

deathbeam:
Made the plugin disabled by default as such extensive tracking can probably add non-insignificat overhead (in swing panel) so users should enable it only if they really want to track something.

TheStonedTurtle:
If you kill 2 dust devils on same tick you get two events with identical loot throwing tracker off.

Potential Fixes: 
* Trigger 1 event per tile
* Send additional events with no items

Option 1 will throw off the kills tracker while option 2 will show a blank kill. Both fix the loot issue though.
IIRC Adam said he'd prefer option 1.

DarkSlaayer:
What if the loot tracker hooked into the valuable drop chat notification. Have it set to 0gp and hide the messages that come through the valuable drop and allow the user to set the valuable drop message in the drop logger.

deathbeam:
I dont think how valuable drop notification is related to drop logger, also its not possible.

@TheStonedTurtle imo its fine for it to be added with that 1 problem as neither of the proposed behaviours will be accurate anyway.

MagicfTail:
@deathbeam A clue and a kill https://i.imgur.com/RFRWAM0.png

Issue #3097, opened 2018-05-22
Spawn timers for ores


Unmoon:
To track when the ore is going to respawn after mining, useful for runite, adamantite etc.

PR #4557, opened 2018-07-29
Shows mining locations & respawn times on mining rocks

### LINKED ISSUE #4392: Add ore respawn timers
Its annoying to grind out mining, and it sucks waiting for ores to respawn. sometimes you just end up walking away and it respawns as soon as you leave.

I believe it would be nice to have a plugin that adds timers above rocks that tells you how long it would be until the ores respawn. however this might not be possible to do on clientside, so a secondary feature that would be nice would to be able to see what specific ores and rocks are available at any mining given mining site by adding some sort of overlay to the world map that displays a list of ores when you hover over a mining sight or such.

DarkSlaayer:
Shows mining locations & respawn times on mining rocks
would close #4392 #3097


Meaglin:
Is there work done on this recently, would be nice if this gets merged :)?

Adam-:
This is covered by #8973

PR #2895, opened 2018-05-18
Shift Click walk here

Reasel:
This will allow users to left click attack and then Shift + left click walk under enemies.

This is only triggered when the attack option is available.

PR #4559, opened 2018-07-30
Corp Plugin Option Enhancement (Left Clickable Core)

j2wick:
A similar plugin to this existed with OSBuddy (or still does -  I don't use OSBuddy any more so I don't know).

In scenarios where a player doesn't opt to kill the core and just wants to run back and forth to hug it, it can be annoying to constantly have to right click to it's current position as it has a left click attack option (by default). 

I have added an additional setting (that can be toggled) that swaps the first option of the core to be "walk here" so players have an easier time running the core.

java_2018-07-29_23-06-14

I sort of copied some existing code from the menu entry swapper plugin since all the methods there are private (if there's some more elegant way to do this let me know and I'd be more than happy to fix it per Runelite standards).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpPlugin.java
@@ -30,25 +30,20 @@ import java.util.Set;
 import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
-import net.runelite.api.Actor;
-import net.runelite.api.ChatMessageType;
-import net.runelite.api.Client;
-import net.runelite.api.GameState;
-import net.runelite.api.NPC;
-import net.runelite.api.NpcID;
-import net.runelite.api.Varbits;
-import net.runelite.api.events.GameStateChanged;
-import net.runelite.api.events.HitsplatApplied;
-import net.runelite.api.events.InteractingChanged;
-import net.runelite.api.events.NpcDespawned;
-import net.runelite.api.events.NpcSpawned;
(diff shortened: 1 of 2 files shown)


Adam-:
Can someone with OSB pro confirm this is still a thing?

j2wick:
So I actually just quickly downloaded OSBuddy to take a look at their corporeal beast plugin. This feature is still in fact in their client, and the option is called "Require right-click to attack Dark energy core". This option is essentially the same thing as what this plugin does, aka it sets the attack option to a right click option so players can just left click to walk under the core when they hug it. Pic below:

javaw_2018-08-01_20-31-37

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpPlugin.java
```diff
@@ -202,4 +206,48 @@ public void onInteractingChanged(InteractingChanged interactingChanged)
...
+	{
+		if (client.getGameState() != GameState.LOGGED_IN)
+		{
+			return;
+		}
+		String option = Text.removeTags(event.getOption()).toLowerCase();
+		String target = Text.removeTags(event.getTarget()).toLowerCase();
+		if (option.equals("attack"))
+		{
+			if (config.leftClickableCore() && target.contains("dark energy core"))
```
Kamielvf: I would put this config check with the gamestate check, to prevent from even running anything at all when disabled.

deathbeam:
If we are fine implementing this I would rather do something more generic, like #2895.

Adam-:
Added in 1fb5365e7c7a8ff5d3f65604c25c9e8358a0fefd

Issue #4056, opened 2018-06-28
Implement option to change Grand Exchange tab redirect to OSBuddy's Exchange History as opposed to Runescape's


vedver:
Currently if you click on an item in the Grand Exchange tab it will redirect you to the item's history such as [Wine of Zamorak](http://services.runescape.com/m=itemdb_oldschool/Wine_of_zamorak/viewitem?obj=245). However, I think it would be beneficial to users who have a preference to allow them to configure the exchange tab to change the redirect to [OSBuddy's exchange history](https://rsbuddy.com/exchange?id=245&).

Hydrox6:
With the move to Wiki prices, this has been implemented for them in 088cb19752169acefef5ac0df88dcbfcd6645f39

PR #4566, opened 2018-07-30
Add OSBuddy links and popup menus to GE plugin

Shingyx:
Add popup menus to both the offers and search tabs, allowing links to either the official GE prices or the OSBuddy GE prices. Add option to prefer OSBuddy prices when left clicking search results.

Resolves https://github.com/runelite/runelite/issues/4056.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeItemPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeItemPanel.java
@@ -41,7 +41,6 @@ import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
 import net.runelite.client.util.AsyncBufferedImage;
 import net.runelite.client.ui.ColorScheme;
-import net.runelite.client.util.LinkBrowser;
 import net.runelite.client.util.QuantityFormatter;
 
 /**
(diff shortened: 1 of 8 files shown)


deathbeam:
You should probably just use 1 link, based on ```	@ConfigItem(
		position = 3,
		keyName = "enableOsbPrices",
		name = "Enable OSB actively traded prices",
		description = "Shows the OSBuddy actively traded price at the GE"
	)
	default boolean enableOsbPrices()
``` and not add right-click menu. E.g with OSB prices enabled, the link will be OSB one, otherwise OSRS one. Also, probably you should also factor the OSB price in search

Shingyx:
This could just be me, but I've always wanted a way to lookup prices for an item in the Offers tab to see if prices changed since I made the offer. I feel like the best way to do this is with a popup menu. I also compare OSBuddy and official prices quite often. I'm not sure if I agree with only having one link.

I think the `preferOsbuddyGeLinks` setting I added is redundant though and I should use the existing `enableOsbPrices` setting more. Things like only showing the Lookup OSBuddy GE price menu item if it's enabled, etc. I'll investigate more tomorrow.

deathbeam:
Well, in GE plugin there should not be anything referring to OSB by default, so make sure that everything OSB related requires explicit enabling. Using the enableOsbPrices config option for that is ideal. I guess if you want, keep the right click menu and make OSB primary menu when the option is enabled, but when disabled there should not be any right click menu and it should simply point to GE.

Shingyx:
Changed behaviour to be around the `enableOsbPrices` setting now. More details in commit messages. Haven't factored OSB price in search yet, but I feel that could probably be it's own separate PR.

deathbeam:
Looks like this needs rebase

Shingyx:
Done.

Shingyx:
Rebased again.

Shingyx:
Needs rework now that https://github.com/runelite/runelite/pull/11940 is merged. Closing for now.

PR #4568, opened 2018-07-30
Properly shutdown everything on window close

deathbeam:
- Properly shutdown applet on window close
- Properly remove system tray icon on close
- Shutdown executorService on shutdown
- Stop all plugins on shutdown


Adam-:
This makes it so it takes ~5 seconds to close the client.

deathbeam:
Most of that time is shutdown time of the applet, I cannot do anything about it.

deathbeam:
!22 in internal needs to be merged before I can continue working on this @Adam-

deathbeam:
Updated the PR to instead of Applet#destroy call GameEngine#shutDown()

PR #4570, opened 2018-07-30
Remember equipment

Joepeke3:
Someway somehow to remember your last equipment.
so you can do a farm run inbetween thieving, that if you comeback from the run you can set your last equipment back faster.


forsco:
This is suppose to be an 'issue' not a pull request.

deathbeam:
https://github.com/runelite/runelite/issues/new/choose ^

Issue #3504, opened 2018-05-30
Add Tar to the Herblore Skill calculator


Hydrox6:
While usually used in tick manipulations methods, making Guam, Marrentill, Tarromin, and Harralander tar is the best experience that can be gotten out of these low level herbs (except for Harralander), at the cost of using 15 tar per herb. Being able to use them in the calculator would be very useful.

PR #4581, opened 2018-07-30
Add Tar to the Herblore Skill calculator #3504

ryancatlin:
Added Guam, Marrentill, Tarromin and Harralander tars to herblore skill calculator


psikoi:
All of these items have been added to the skill calc, not sure in which commit.

https://github.com/runelite/runelite/blob/master/runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_herblore.json

But this should be closed.

@Hydrox6

Issue #925, opened 2018-03-10
Idle support to AFK combat scenarios


FrozenState:
The current implementation of the idle notifier pings the client every 4 minutes and 40 seconds if a user is splashing / in constant combat (NMZ).

It would be helpful to have an option implemented to have the client ping every 19 minutes in these scenarios.

deathbeam:
Fixed in #5222

PR #4584, opened 2018-07-30
Splash timer - request - idea

Sethjhe:
Hey I have no idea how this site works or where to post this, so if an admin reads this, could you please move it to the appropriate area?

Anway, my suggestion/idea for RuneLite is to maybe add a splash timer, so that you get a notification when you're out of combat. (Similair exists in NightMareZone but not outside it) 

So I was hoping someone could make a plugin for it! Thanks in advance lads, hopefully this reaches someone!


Abextm:
Duplicate of #925
FWIW This should have been an issue, Pull Requests are for code changes

PR #4586, opened 2018-07-31
Add Logout delay timer

MagicfTail:
Add timer showing how long until you can log out after getting attacked


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -165,6 +168,22 @@ public void onVarbitChange(VarbitChanged event)
...
 
+	@Subscribe
+	public void onHitsplatApplied(HitsplatApplied event)
+	{
+		if (event.getActor() != client.getLocalPlayer())
+		{
+			return;
+		}
+
+		if (config.showLogoutDelay() &&
```
Nightfirecat: This project's line continuation style is to move operators to the beginning of the next line, like so:

```
true
    || false
    && true
```

PR #4587, opened 2018-07-31
Mounted glory reorder

beardhub:
Add ability to reorder teleports for mounted glory in poh


devLotto:
Is this something we want?

raiyni:

PR #4592, opened 2018-07-31
Fix order of loading session settings

### LINKED ISSUE #4681: prayer plugin not working

**Describe the bug**
since the last update the prayer plugin has stopped working for me and when i try to open the settings the menu freezes until i click the wrench twice to close and open the side menu then the prayer plugin menu is open but its empty

**To Reproduce**

**Expected behavior**

**Screenshots**

 - OS: windows 
 - RuneLite version: 1.4.10
 - Launcher version: [e.g 1.5.2]

**Additional context**

deathbeam:
Currently the session simply overwrites default settings and if they
were not persisted before, they will be just null. Load session before
loading default config to fill empty places.

Fixes #4681

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -241,6 +241,9 @@ public class RuneLite
 		// Load user configuration
 		configManager.load();
 
+		// Load the session, including saved configuration
+		sessionManager.loadSession();
+
 		// Tell the plugin manager if client is outdated or not
 		pluginManager.setOutdated(isOutdated);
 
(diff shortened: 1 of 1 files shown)


deathbeam:
I can reproduce this consistently one my PC where I did not reset settings, because of new Enum option in Prayer plugin, this bug basically stopped loading of the config panel, so the prayer panel was empty (well did not opened, and it was empty after going to another tab and back). It was because parsing of Enum failed, because Name was null (as the default config got overwritten). After this patch, Prayer panel is loading fine.

deathbeam:
@Adam- So can this be added to next release? I saw that you said its ok to merge yesterday but I am not sure if it was related to this one.

Adam-:
This is now a regression from https://github.com/runelite/runelite/commit/cda98fc50e2490f3c124a0b3cbff109ae01b60ee where before having the plugin manager registers on the eventbus prior to plugins starting allowed it to load default plugin config when plugins were loaded, which prevented this.

deathbeam:
Removed the clientSessionManager move around

PR #4605, opened 2018-08-01
Remove unused @Slf4j annotations and imports

Nightfirecat:
The following script was used to remove the offending instances, run from the root of the project:

```bash
foo="$(git ls-files ./*.java)"
while IFS='' read -r file || [[ -n "$file" ]]; do
  if grep --quiet -F '@Slf4j' "$file"; then
    if ! grep --quiet -F 'log.' "$file"; then
      sed -i -r -e '/^@Slf4j/d' -e '/^import lombok\.extern\.slf4j\.Slf4j;/d' "$file"
      git add "$file"
    fi
  fi
done <<< "$foo"
```

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/session/SessionClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/session/SessionClient.java
@@ -29,14 +29,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.UUID;
-import lombok.extern.slf4j.Slf4j;
 import net.runelite.http.api.RuneLiteAPI;
 import okhttp3.HttpUrl;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
 
-@Slf4j
 public class SessionClient
 {
 	public UUID open() throws IOException
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/alchemy/AlchemyRoomTimer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/alchemy/AlchemyRoomTimer.java
@@ -26,12 +26,10 @@ package net.runelite.client.plugins.mta.alchemy;
 
 import java.awt.image.BufferedImage;
 import java.time.temporal.ChronoUnit;
-import lombok.extern.slf4j.Slf4j;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.ui.overlay.infobox.Timer;
 import net.runelite.client.util.ImageUtil;
 
-@Slf4j
 public class AlchemyRoomTimer extends Timer
 {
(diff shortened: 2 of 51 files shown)


deathbeam:
Can you double check if this is still catching all of the unused ones after recent changes?

Nightfirecat:
Good call, looks like the last commit of the image removal PR removed a lot of log statement calls from IOException handling. I've rebased for the most recent run of that script.

runelite-client/src/main/java/net/runelite/client/plugins/mta/alchemy/AlchemyRoomTimer.java
```diff
@@ -1,57 +1,55 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.mta.alchemy;
```
deathbeam: Why does this file have line endings changed?

http-api/src/main/java/net/runelite/http/api/session/SessionClient.java
```diff
@@ -1,100 +1,98 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.UUID;
+import net.runelite.http.api.RuneLiteAPI;
+import okhttp3.HttpUrl;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+public class SessionClient
```
deathbeam: Why does this file have line endings changed?

runelite-client/src/main/java/net/runelite/client/plugins/mta/alchemy/AlchemyRoomTimer.java
Nightfirecat: Because it was apparently using `\r\n` line endings and `sed` changed them.

PR #2449, opened 2018-05-07
Update NPC health data

Kamielvf:
Only NPCs which have an "Attack" option were dumped.
- This updates the NPC health data to include all the NPCs that are in the cache as of the update on [August 09 2018](http://services.runescape.com/m=news/looting-bag-improvements-pid-changes-and-bounty-hunter-targets?oldschool=1).

The health information is gathered from the OSRS Wikia. This might cause some data to be inaccurate if the wiki is missing it or has the incorrect values.
For a list of currently missing NPCs: [missing.json](https://puu.sh/Befao/a874c774d2.json)

Closes #3387
Closes #1790
Closes #4364
Closes #4766

PR #4618, opened 2018-08-01
Add remaining baby dragons to NPC health

Erveon:
Added baby green, baby blue and non-POH baby red dragon to the NPC health list.
Moved the POH baby red dragon with the other baby dragons so it wouldn't be lonely.


Erveon:
Closed because https://github.com/runelite/runelite/pull/2449 was merged, which contains these changes

PR #4630, opened 2018-08-01
Account for linked item id, format tooltip

### LINKED ISSUE #4644: Loot tracker showing noted items with 0 value
Snapdragons and Adamantite bars from Zulrah show as 0 GP value in the loot tracker.

deathbeam:
- Account for linked item id when getting item price
- Format item price tooltip
- Create new data structure for loot tracker and precompute it
to avoid calling methods that are unsafe outside of client thread

Fixes #4644

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerItemEntry.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerItemEntry.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2018, Tomas Slusny <slusnucky@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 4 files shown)


TheStonedTurtle:
Mentioned it in discord but:

Add this to `ItemManager.java`
```java
    /**
     * Look up an item's composition. If the noted variant of the item
     * is passed, this will return the composition for the unnoted version.
     *
     * @param itemId item id
     * @return item composition
     */
    public ItemComposition getUnnotedItemComposition(int itemId)
    {
        ItemComposition composition = getItemComposition(itemId);
        if (composition.getNote() == -1)
        {
            return composition;
        }
        return getItemComposition(composition.getLinkedNoteId());
    }
```
and then update `calculatePrice()` so `itemPrice` is this:
```java
ItemPrice itemPrice = itemManager.getItemPrice(itemManager.getUnnotedItemComposition(itemStack.getId()).getId())
```

deathbeam:
Because that call is unsafe when not on client thread.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerItemEntry.java
```diff
@@ -0,0 +1,37 @@
...
+package net.runelite.client.plugins.loottracker;
+
+import java.awt.image.BufferedImage;
+import lombok.Value;
+
+@Value
+class LootTrackerItemEntry
+{
+	private final String name;
+	private final BufferedImage image;
```
Adam-: The `getImage` is safe since it uses an AsyncBufferedImage so you probably should just leave that in the UI.

Adam-:
Also maybe the getItemComposition should assert being on the corret thread to prevent this in the future.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -250,4 +250,33 @@ public void onChatMessage(ChatMessage event)
...
 		}
 	}
+
+	private LootTrackerItemEntry[] buildEntries(final Collection<ItemStack> itemStacks)
+	{
+		return itemStacks.stream().map(itemStack ->
+		{
+			final ItemComposition itemComposition = itemManager.getItemComposition(itemStack.getId());
+			final int realItemId = itemComposition.getNote() != -1 ? itemComposition.getLinkedNoteId() : itemStack.getId();
+			final ItemPrice itemPrice = itemManager.getItemPrice(realItemId);
```
Adam-: This call is also safe on any thread, so probably I would move that back to the UI aswell.

deathbeam: Yeah, but then I would need to set the "realItemId" instead to the entry and then send itemManager back there. I dont see any benefit in doing either, so why do you prefer that?

PR #4632, opened 2018-08-01
instance map: properly remove mouse listener on shutdown

### LINKED ISSUE #3473: Instance Map strange behavior
This really wouldn't affect many people at all, and it's the kind of thing that would likely fix itself after a client restart anyway, but...

If you have the Instance Map plugin enabled and open the instance map, then disable the plugin without first closing the map, it remains "open" but invisible, and you can't click on that top-left portion of the screen anymore.

It confused me for a bit until I figured out what was causing the problem. I guess it might be worth fixing; if anything, just for the sake of avoiding weird behavior like that :p

Kamielvf:
Fixes #3473


Joopi:
Should be applied for when logging out due to idling as well. Whenever you get back it blocks most of the Red button in lobby screen.

PR #4637, opened 2018-08-02
Add option to make screen markers not click through

MagicfTail:
Add an option to make the screen markers not click through, similar to osbuddys screen markers.

Note, the icons are very much placeholders


runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
```diff
@@ -124,6 +124,8 @@ protected void startUp() throws Exception
 		clientToolbar.addNavigation(navigationButton);
 
 		mouseListener = new ScreenMarkerMouseListener(this);
+
+		mouseManager.registerMouseListener(mouseListener);
```
Kamielvf: You would need to unregister this mouse listener in shutdown.

runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
```diff
@@ -229,6 +225,18 @@ void resizeMarker(Point point)
 		overlay.setPreferredSize(bounds.getSize());
 	}
 
+	public boolean shouldEatClick(Point point)
+	{
+		for (ScreenMarkerOverlay screenMarker : screenMarkers)
+		{
+			if (screenMarker.getBounds().contains(point) && !screenMarker.getMarker().isClickThrough())
```
Kamielvf: check if the marker is not click trough first

Kamielvf:
Have you tested how this behaves when loading previously stored screen markers?

MagicfTail:
It sets all existing screen markers to no click through, which is a bit unfortunate

Kamielvf:
In that case I would make it so that it is click trough by default to not change the existing behavior.

MagicfTail:
Existing screen markers are now click through

Kamielvf:
I think it would be better to locate this button next to the visible toggle

runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
MagicfTail: Done

MagicfTail:
It's now moved to the right and rebased

PR #4667, opened 2018-08-03
Move isOverlayOutdated to IndexDatabase

deathbeam:
Instead of creating new callback and then simply setting boolean on
RuneLite to dump it somewhere, expose this boolean from IndexDatabase
and remove injection of RuneLite.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/IndexDataBase.java
+++ b/runelite-api/src/main/java/net/runelite/api/IndexDataBase.java
@@ -29,4 +29,8 @@ package net.runelite.api;
  */
 public interface IndexDataBase
 {
+	/**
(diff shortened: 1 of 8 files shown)


runelite-api/src/main/java/net/runelite/api/IndexDataBase.java
```diff
@@ -29,4 +29,8 @@
  */
 public interface IndexDataBase
 {
+	/**
+	 * Returns true if any cache overlay is outdated due to hash mismatch
```
Abextm: `Returns true if any cache overlay in this index  is outdated due to hash mismatch`

deathbeam: Updated

PR #4670, opened 2018-08-03
Fix for noted items having no value in loottracker

Solution6996:
- Adds a check for noted items when calculating prices


Solution6996:
I'm and idiot.

Solution6996:
I'm an idiot.

PR #4673, opened 2018-08-03
Remove unnecessary headers from plugin panels

deathbeam:
Now all plugin panels have tooltips so the headers are not necessary and
most of plugins do not have them already.

Preview:
screenie4
screenie3
screenie2
screenie


psikoi:
Not a big of the huge reset button, but i must say, they all look way better without the headers, niiice

deathbeam:
Yeah, it was either leave the kourend library plugin name there as only exception, or leave the small button on right without any header or just do huge ass button, I chose the best way I think :D

Issue #1639, opened 2018-04-17
Menu Entry Swapper does not work on Zanaris fairy ring


Jimbimanim:
Menu Entry Swapper plugin normally puts "Last Destination" as the left click option, but on the Zanaris fairy ring "Configure" is the left click option as it would be with the plugin off.
I think this maybe has something to do with the fact that this particular fairy ring only has the options Configure/Last Destination/Examine rather than the Zanaris/Last Destination/Configure/Examine that every other fairy ring has

This probably isn't needed for this issue, but as per your submission guidelines i am on linux mint 18.3 using java 1.8.0_162 and runelite 1.3.4.1

Gamer1120:
I will work on this for a bit. I think we want a separate option for the Zanaris fairy ring, since I could see some people preferring Configure and some people preferring Last Destination.

Gamer1120:
This needs more feedback from the community. The specific question to answer is: "Should the Zanaris fairy ring have a separate option in the Menu Entry Swapper plugin?"
A more detailed explanation is: "The options on the Fairy ring in Zanaris are "Configure" and "Last-destination". The options on all other Fairy rings are "Zanaris", "Configure" and "Last-destination". In other words, would you want to be able to select "Last-destination" on the Fairy ring in Zanaris and select "Zanaris" on all other Fairy rings (or Configure on the Zanaris Fairy ring and Last-destination on all others)?"
A scenario in which it would be useful to be separate is when you want to fish Karambwans and bank in Zanaris. You'd want Zanaris on the regular Fairy ring, and Last-destination on the Zanaris Fairy ring.

Jimbimanim:
I say yes, doing karambwan fishing is actually what prompted me to report this in the first place and having both fairy rings able to be left clicked would be great. A separate option for Zanaris ring sounds great

Neokolzia:
I believe this is still open and unfufillled in the game if someone is willing to tackle it!  it would certainly help individuals looking to use Zanaris for banking (ensouled heads, karambwans etc)

Gamer1120:
This issue is still not fixed. There is still no distinction between the Zanaris Fairy ring and the other fairy rings. That is what this was about. I checked out master, and saw it for myself. I made a quick MSPaint mock-up of what I mean::
![image](https://user-images.githubusercontent.com/9074448/43659064-1f89b952-975b-11e8-802d-10f6191708ae.png)

deathbeam:
[...]

PR #4680, opened 2018-08-03
Add separate option for Zanaris Fairy ring to Menu Entry Swapper

### LINKED ISSUE #4677: Separate option for Zanaris Fairy Ring on the Menu Entry Swapper
image
I would like to request an additional option in the Menu Entry Swapper menu for the Zanaris Fairy Ring. This can be useful when doing skilling near fairy rings and banking in Zanaris, for example fishing Karambwans.

Gamer1120:
Fixes #4677


Abextm:
IMO this doesn't need its own option. I would just make it so that If you have Zanaris configured on the Zanaris ring it goes to Last-Destination.

Gamer1120:
@Abextm Please check the discussion in #1639 about this. If you then still don't want the option, just close it. It's a GREAT QOL for semi-AFK Karambwan fishing with banking in Zanaris.

Abextm:
Thats why I said to swap it directly to last-destination. I don't see a use case where you want 1 click to zanaris, but configure to leave.

Gamer1120:
@Abextm Because once you have an inventory of Karambwans, you'd have to right-click to access the Zanaris option.

Abextm:
If the config is set to `Zanaris`:
 - The zanaris ring will say `Last-destination(XXX)`
 - Every other ring will say `Zanaris`

Gamer1120:
Really? I didn't know that. I'm testing that.

Abextm:
No, that is what I am saying you should implement instead of an option.

Gamer1120:
That makes sense. I'll do that.

Gamer1120:
Fixed and tested.

CameronDeLone:
As of July 2020 having the option set to "Zanaris" still defaults to "configure" when in using the zanaris fairy ring

PR #4698, opened 2018-08-04
Add generic chat command API with generic data storage

deathbeam:
## Add chat data service

- Add new HTTP service /chat to what various temporary chat data can be
submitted and queried from.
- Make kill count command use ChatDataService now instead of
KillCountService and remove entire KillCountService
- Add ChatDataType enum to what new data types will be added in future

## Add ChatCommandManager

- Adds ChatCommandManager to what plugins will be able to register their
commands and execute/postpone them.
- Convert ChatCommandsPlugin to use ChatCommandManager


deathbeam:
As #4365 is closed, this is not needed

PR #4708, opened 2018-08-04
Ground item timers

### LINKED ISSUE #2383: Item drop despawn and visibility timers
Suggestion
- Ability to track the time remaining until an item on the ground despawns
- Ability to track the time remaining until an item becomes visible to other players
- Ability to limit tracking to only certain items by name or by type
  - For example tracking only food and pot despawns

Things to discuss
- How to go about displaying the counter 
  - Pie chart or maybe inline with ground indicator information
  - Want to minimize clutter
- How to handle stacks of the same item
- Special Cases that affect timers
  - Items dropped in an instanced area
  - Items dropped on death
  - Are there other special circumstances that effect item despawns?

This would be very helpful for juggling items in areas such as GWD where you might have gotten more supplies from drops than you can actually hold

### LINKED ISSUE #4535: Ground Item Timers
When doing safe spot combat, it kind of sucks running out to grab loot from the ground every time something nice drops. It would be wonderful if players could have the option to see a timer next to the ground item labels that tell the player how long until the items despawn. This could potentially be limited to items that are above (x) threshold. Doing this would allow player to only have to go pick things up once they are close to despawning.

### LINKED ISSUE #4609: Dropped item timer
Hello,

Can we please get a dropped item timer, maybe switches colors or something to show when the item is visible to public? 

Example:
Blue from 1:59 to 1:01 and red from 01:00 till it disappears at zero?

Thank you.

ms813:
Closes #2383 and closes #4535 and closes #4609 

Known issues:
* items already on ground when you see them assume the public timer starts from 60 seconds and may disappear before that
* wilderness not handled correctly
* items dropped under you by another player are assumed to be yours so show the private timer instead of the public timer
* items dropped on player death show the 1 min public timer instead of a 1 hour timer
* stacks of items that are dropped one at a time (e.g when shift dropping an inventory) are not handled well

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -49,26 +49,6 @@ public interface GroundItemsConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "privateDurationColor",
-		name = "Private timer color",
-		description = "Configures the color for the private duration timer"
-	)
-	default Color privateDurationColor()
-	{
-		return Color.GREEN;
-	}
-
-	@ConfigItem(
-		keyName = "publicDurationColor",
-		name = "Public timer color",
-		description = "Configures the color for the public duration timer"
-	)
-	default Color publicDurationColor()
-	{
-		return Color.YELLOW;
-	}
-
-	@ConfigItem(
 		keyName = "highlightedItems",
 		name = "Highlighted Items",
 		description = "Configures specifically highlighted ground items. Format: (item), (item)",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -32,10 +32,42 @@ import net.runelite.client.config.ConfigItem;
 import net.runelite.client.plugins.grounditems.config.ItemHighlightMode;
 import net.runelite.client.plugins.grounditems.config.MenuHighlightMode;
 import net.runelite.client.plugins.grounditems.config.PriceDisplayMode;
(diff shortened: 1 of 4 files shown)


ms813:
Green = only visible to you
Yellow = visible to everyone
Colours are customisable in settings.
Can set it so that timers are visible Always, When Hotkey is pressed (alt), or Never
Clue scroll is untradeable and green timer lasts 150 seconds
image

Popedigoz:
Is this live?

forsco:
> Is this live?

Hasn't been reviewed as of yet and not merged.

azzawoody:
is this in the works for up coming updates plz add asap Keep up the good work

JacoboMoral:
Any progress on this plugin? Really interested on it

Adam-:
No. There are a lot of edge cases mentioned in the OP. In addition to those it also:

1) incorrectly marks static spawns as despawning when walking into a scene
2) incorrectly marks static spawns as despawning when they respawn
3) shows the incorrect time when walking into a scene with an item already on the ground

I do not think it would be that well received if we released something as incorrect as this.

Maybe if we reduced it to only show timers for things that are more likely to be correct such as:

1) items you drop yourself
2) items you get from killing a npc or player

It would be better?

Popedigoz:
> 
> 
> No. There are a lot of edge cases mentioned in the OP. In addition to those it also:
> 
>     1. incorrectly marks static spawns as despawning when walking into a scene
> 
>     2. incorrectly marks static spawns as despawning when they respawn
> 
>     3. shows the incorrect time when walking into a scene with an item already on the ground
> 
> 
> I do not think it would be that well received if we released something as incorrect as this.
> 
> Maybe if we reduced it to only show timers for things that are more likely to be correct such as:
> 
>     1. items you drop yourself
> 
>     2. items you get from killing a npc or player
> 
> 
> It would be better?

Yes release plugin with only those and if and when possible just update to have the edge cases

Adam-:
added by #9423

Issue #4668, opened 2018-08-03
Barrows plugin crypt dig tiles


eggbrow:
It would be nice to have the tiles that you can dig to enter barrows crypts highlighted. The sizes are different for each barrows brother and it would be useful to some players. A toggle for this feature would be good as not everybody would like to see so much tile highlighting.

[Image of highlighted tiles concept](https://i.imgur.com/FwmEpxI.png)

psikoi:
@deathbeam good first contribution

deathbeam:
I would not highlight EVERY tile as that just looks horrible, I would maybe use center tile highlighting with spade icon ALA clue scroll plugin

psikoi:
@deathbeam that defeats the purpose of this PR, everyone knows you can dig in the middle, this is supposed to highlight all the possible dig locations, i believe

deathbeam:
Maybe, but it is better than nothing I guess, and highlighting entire area just looks bad no matter what you would do (also the fact that we cannot draw under players/npcs do not helps too)

psikoi:
Hum i guess so.. Maybe if the opacity was low enough it wouldnt be as ugly

Spudlee:
We could highlight the borders of the dig area, maybe not even put the tiles, but just an outline. Still might not look great, but it would be a happy medium between the two.

Reasel:
Might be nice to change the color based on if you have cleared that brother yet or if it is the entrance.

jachymb:
How about only highlight the outline, similar to what the NPC Aggression Timer plugin does

Frosty-J:
[...]

PR #4714, opened 2018-08-05
Tiles for barrows crypts

Bjdufre1:
Implements #4668 

This highlights all tiles that can be dug on at barrows. Not sure how useful this is but someone requested it and I wanted some experience so this is what I came up with. Initially I wanted to draw a single rectangle but with how the mounds are, it is super inaccurate and deceiving.

barrows


psikoi:
I'd personally make the tiles lower in opacity, it might be too distracting this way

runelite-client/src/main/java/net/runelite/client/plugins/barrows/CryptLocations.java
```diff
@@ -0,0 +1,265 @@
...
+			new WorldPoint(3558, 3298, 0),
+			new WorldPoint(3558, 3299, 0),
+			new WorldPoint(3558, 3300, 0),
+			new WorldPoint(3559, 3295, 0),
+			new WorldPoint(3559, 3296, 0),
+			new WorldPoint(3559, 3297, 0),
+			new WorldPoint(3559, 3298, 0),
+			new WorldPoint(3559, 3299, 0),
+			new WorldPoint(3559, 3300, 0)),
+	KARILS(new WorldPoint(3563, 3273, 0),
```
ollieatkinson: This code might benefit from creating a `WorldRect` structure which takes in a origin and size and can output all the points inside.

cfoust: Could probably just add a method to `WorldArea`

ollieatkinson:
Yes, I agree, return a standard struct which represents a rect

runelite-client/src/main/java/net/runelite/client/plugins/barrows/CryptLocations.java
psikoi: Maybe a constructor that takes in two tiles? Northwest and Southeast? that should be enough

ollieatkinson:
Yes, or a coordination and a width/height

deathbeam:
See my reply here: https://github.com/runelite/runelite/issues/4668#issuecomment-423778889

Issue #4715, opened 2018-08-05
Add option to show quantity on ground items as x 15


GoEasyImBad:
Adding an option to either show x amount of items on ground vs the ( x ) option. 
would probably be an easy add. the Parantheses are clunky in my opinion.

GoEasyImBad:
yes you are right. instead of (100) make it x 100.. maybe Blood Rune (100)
can be swapped to Blood Rune x 100

adm1n1stra1tor:
@GoEasyImBad 
@deathbeam 
Further suggestion in this thread.

Make the quantity color changeable. So you get further contrast, based on what color you like,

adm1n1stra1tor:
Additional thoughts;
Making the quantity text, automatically a bit different from the original color you chose for that value-bracket.

Example: (Blue)Coins (Ligther Blue)x 100

deathbeam:
As we now have collapsing, this will be confusing, closing

PR #4716, opened 2018-08-05
Option to switch notation for groud objects

Bjdufre1:
Addresses #4715 
Adds the option to switch from the format of Coins (100) to Coins x 100 for ground items

paren
x


deathbeam:
See the resolution on the issue

PR #4718, opened 2018-08-05
LootLogger: Change right click options from reset to remove

MagicfTail:


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -151,7 +151,7 @@ void addLog(final String eventName, final int actorLevel, LootTrackerItemEntry[]
 		updateOverall();
 
 		// Create reset menu
```
Nightfirecat: The comments above the changed lines should be updated to reflect the verbiage change.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -96,7 +96,7 @@
 		overallPanel.add(overallInfo, BorderLayout.CENTER);
 
 		// Create reset all menu
```
Nightfirecat: Here as well.

PR #1774, opened 2018-04-21
Inventory Highlight While Dragging Plugin (Somewhat WIP)

jplsek:
This is somewhat a work in progress. Although it works and is mostly feature complete, the way dragging is detected can be improved. I have a feeling that there is a way to detect when an item is being dragged in the runescape client (similar to widget dragged, but that doesn't work for items). Figuring out the render layer would also be nice so the overlays don't draw on top of the item.

Closes #1333 

Examples of different options:
From left to right: The default option (renders an item underneath), highlight option, highlight and grid option.
ezgif-5-7f125a901c

Suggestions welcome.

I chose to make the plugin is disabled by default just so players are not surprised or get distracted.

PR #4722, opened 2018-08-05
Inventory Grid

DrizzyBot:
Created a plugin that shows a grid in your inventory to differentiate item slots


psikoi:
why?

jplsek:
Related #1774

Issue #4713, opened 2018-08-05
ItemPricesPlugin: Hide tooltips on Inventory items no longer works in the bank


MagicfTail:

deathbeam:
Looks like this is intentional behaviour, changing this to feature request

PR #4725, opened 2018-08-05
Price tooltip hidden when in bank

nieka:
When the item price plugin is enabled and the setting hide in inventory is set the price tooltip is now also hidden in the invertory when the bank is open. Which fixes issue #4713

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
@@ -113,8 +113,13 @@ class ItemPricesOverlay extends Overlay
 							return null;
 						}
 						// intentional fallthrough
-					case WidgetID.BANK_GROUP_ID:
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -113,8 +113,13 @@ public Dimension render(Graphics2D graphics)
 							return null;
 						}
 						// intentional fallthrough
-					case WidgetID.BANK_GROUP_ID:
 					case WidgetID.BANK_INVENTORY_GROUP_ID:
+						if (config.hideInventory())
```
Nightfirecat: Just add this to the `WidgetID.INVENTORY_GROUP_ID` case above so the code is not duplicated.

nieka: Move the case statment so the invetory check code isn't duplicated

PR #4728, opened 2018-08-05
Add herb sack price check

### LINKED ISSUE #4573: Herb Sack price check
Once you "Check" Herb Sack could we get also price check included with that, similar to the "Examine" on items.

aquivers:
When clicking "Check" on Herb Sack, the player gets a chat message which contains the total value of the herbs in the sack. 

Fix #4573 

herbsvalue


realm1930:
Why is this not being merged? It seems ready and it fixes a real issue?

PR #4733, opened 2018-08-06
Set the default off all entity hider options to false

MagicfTail:
Helps the people who turn on random plugins and don't know why they suddenly can't see any other players.


deathbeam:
I dont think having plugin off by default and make all its options when enabled off by default too makes sense.

MagicfTail:
The people who want to use the plugin will know to look in the settings for what they want, while it will be a big help for the people who turn on random plugins, which is a surprisingly large amount of people going by the discord

PR #4736, opened 2018-08-06
Add account-specific configuration

### LINKED ISSUE #249: Slayer plugin - Task carrying over to other accounts
**Detailed Problem:**
Slayer task amount is only saved for the client, not per account. As a result, when a user switches accounts it will still show the slayer task from the previous account, and even progress the task if not checked on the new account.
**Cause/Limitations:**
Slayer task amount is saved as an extra variable in the saved plugin settings. There is no distinction between different accounts, because plugins settings are global between all accounts, Plugin settings currently do not support saving Maps because the database is not structured to handle them.

### LINKED ISSUE #2220: Multiple plugin setting profiles
Depending on the activity I'm doing, I use vastly different plugin settings. Right now I have to manually enable/disable and change a dozen plugins settings multiple times every day.

deathbeam:
Add possibility to make plugin configurations account specific and so
load different configuration based on logged in user.

Closes #2220 and closes #249

Preview:
screenie


takuyakanbr:
IMO there should be a separate storage for account/player data (e.g. slayer, farming, bird house, kc).

Then the plugin configuration system can be extended with a profile/preset system, which allows mixing and matching between accounts and profiles (i.e. you can choose to create multiple profiles for 1 account - this seems to be what #2220 is requesting for).

deathbeam:
Well the config system is already well-suited for account-specific data as we already use it for that in farming tracker, I just made this functionality more generic. And for #2220 for the use-case there just account-specific settings would be enough.

deathbeam:
I guess there is no good way atm to do this without putting the usernames to the group, so closing this for now, no reason to keep this up

PR #4737, opened 2018-08-06
Add Examine Placeholder Price Lookup

DevDennis:
- Allows you to lookup the price of placeholders in your bank by examining


deathbeam:
I would probably add new method to handle both notes and placeholders to itemManager and make also ItemPrices use this API (for tooltips in bank).

Issue #4527, opened 2018-07-27
Highlight birdhouse being empty


Invalids:
I'm a dumb person who often forgets to fill my birdhouses with seeds (Even after 100+ runs I'm still doing it...).

A red overlay in the square of a Birdhouse (Empty) would be fantastic for me. It wouldn't have to do much, just highlight a birdhouse that isn't filled and disappear when the birdhouse is filled.

I've considered just getting gud, but that hasn't worked so far.

deathbeam:
Looks like we will not be implementing this

PR #4741, opened 2018-08-06
Highlight empty birdhouses plugin

kristian-welsh:
I haven't been able to test redwood birdhouses due to my hunter lvl but if the state value continues to follow the pattern it followed from normal to magic, it should work.
resolves issue #4527


runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,157 @@
...
+)
+public class BirdhousesPlugin extends Plugin
+{
+	private static final int[] BIRDHOUSE_REGIONS = {
+		14906,
+		14652,
+		14651
+	};
+	
+	public static final Map<Integer, VarPlayer> ID_TO_VAR = createIdMap();
```
Nightfirecat: You can initialize this using the following syntax:

```
public static final Map<Integer, VarPlayer> ID_TO_VAR = new HashMap<Integer, VarPlayer>()
{{
    put(30565, VarPlayer.BIRDHOUSE_1);
    put(30566, VarPlayer.BIRDHOUSE_2);
    // ...
}};
```

[Here is an example of this type of initialization being used in ColorUtil.java](https://github.com/runelite/runelite/blob/master/runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java#L35-L45).

kristian-welsh: ooh, handy

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+import net.runelite.api.events.GameStateChanged;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.ui.overlay.OverlayLayer;
+import net.runelite.client.ui.overlay.OverlayManager;
+
+@PluginDescriptor(
+	name = "Birdhouses",
+	description = "highlights the positions of empty birdhouses",
+	tags = {"birdhouse", "fosil island", "highlight"}
```
Kamielvf: `fossil island` typo here

Kamielvf:
I would use Guava's ImmutableMap for this. ref: https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableMap.html#of-- and get the object IDs from the ID files we have.

```
public static final Map<Integer, VarPlayer> ID_TO_VAR = ImmutableMap.of(
		NULL_30565, VarPlayer.BIRDHOUSE_1,
		NULL_30566, VarPlayer.BIRDHOUSE_2,
		NULL_30567, VarPlayer.BIRDHOUSE_3,
		NULL_30568, VarPlayer.BIRDHOUSE_4
);
```

Same goes for your other list, you can make use of Guava's ImmutableList.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesOverlay.java
```diff
@@ -0,0 +1,87 @@
...
+		25
+	);
+	
+	private final BirdhousesPlugin plugin;
+	
+	@javax.inject.Inject
+	private Client client;
+	
+	@Inject
+	public BirdhousesOverlay(BirdhousesPlugin plugin)
```
Nightfirecat: `private`

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesOverlay.java
```diff
@@ -0,0 +1,87 @@
...
+		13,
+		16,
+		19,
+		22,
+		25
+	);
+	
+	private final BirdhousesPlugin plugin;
+	
+	@javax.inject.Inject
```
Nightfirecat: Remove this `@Inject` and set it as a parameter in the constructor.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesOverlay.java
```diff
@@ -0,0 +1,87 @@
...
+	{
+		setPosition(OverlayPosition.DYNAMIC);
+		setLayer(OverlayLayer.ABOVE_SCENE);
+		this.plugin = plugin;
+	}
+	
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		if (!plugin.isNearBirdhouses())
```
Nightfirecat: If `plugin.getBirdhouses()` is empty whenever a player is outside of the birdhouse regions, this check won't be needed since the TileObject loop below won't be executed. This seems unnecessary.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesOverlay.java
```diff
@@ -0,0 +1,87 @@
...
+		{
+			return null;
+		}
+		
+		for (TileObject birdhouse : plugin.getBirdhouses())
+		{
+			VarPlayer var = BirdhousesPlugin.ID_TO_VAR.get(birdhouse.getId());
+			int state = client.getVar(var);
```
Nightfirecat: This is only used in the next line, remove the variable assignment and just use `EMPTY_STATES.contains(client.getVar(var))`.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+)
+public class BirdhousesPlugin extends Plugin
+{
+	private static final int[] BIRDHOUSE_REGIONS = {
+		14906,
+		14652,
+		14651
+	};
+	
+	public static final Map<Integer, VarPlayer> ID_TO_VAR = new HashMap<Integer, VarPlayer>()
```
Nightfirecat: Package private

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+	{{
+		put(30565, VarPlayer.BIRDHOUSE_1);
+		put(30566, VarPlayer.BIRDHOUSE_2);
+		put(30567, VarPlayer.BIRDHOUSE_3);
+		put(30568, VarPlayer.BIRDHOUSE_4);
+	}};
+	
+	@Getter
```
Nightfirecat: `@Getter(AccessLevel.PACKAGE)`

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+	private List<TileObject> birdhouses;
+	
+	@Getter
```
Nightfirecat: `@Getter(AccessLevel.PACKAGE)`

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+	@Override
+	protected void startUp() throws Exception
+	{
+		overlayManager.add(overlay);
+		overlay.setLayer(OverlayLayer.ABOVE_SCENE);
+		nearBirdhouses = inBirdhouseArea();
+	}
+
+	@Override
+	protected void shutDown() throws Exception
```
Nightfirecat: Birdhouses list should be cleared as well.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+	public static final Map<Integer, VarPlayer> ID_TO_VAR = new HashMap<Integer, VarPlayer>()
+	{{
+		put(30565, VarPlayer.BIRDHOUSE_1);
+		put(30566, VarPlayer.BIRDHOUSE_2);
+		put(30567, VarPlayer.BIRDHOUSE_3);
+		put(30568, VarPlayer.BIRDHOUSE_4);
+	}};
+	
+	@Getter
+	private List<TileObject> birdhouses;
```
Nightfirecat: Since it is not initialized, isn't it possible to encounter a NPE by enabling the plugin while already logged in, then triggering a `GameObjectSpawned` with a birdhouse?

kristian-welsh:
I've made the changes, but am a little uncertain what the null ids are for. Is it an auto-generated list of IDs that are guaranteed to reference some object in game?

Nightfirecat:
NullObjectID is just a list of objects that have null names. So yes, they are references of ingame objects.

Kamielvf:
The ID files are auto generated from the cache. They correspond to either an object's ID, item ID or NPC ID. The reason some are identified as  `NULL` is because their name is null.

kristian-welsh:
Great, thank you.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
kristian-welsh: good catch

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+		NullObjectID.NULL_30566, VarPlayer.BIRDHOUSE_2,
+		NullObjectID.NULL_30567, VarPlayer.BIRDHOUSE_3,
+		NullObjectID.NULL_30568, VarPlayer.BIRDHOUSE_4
+	);
+	
+	@Getter(AccessLevel.PACKAGE)
+	private List<TileObject> birdhouses = new ArrayList<TileObject>();
+	
+	@Getter(AccessLevel.PACKAGE)
+	public boolean nearBirdhouses;
```
Nightfirecat: This looks unused now. It is assigned, but never read from anymore.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	@Override
+	protected void shutDown() throws Exception
+	{
+		overlayManager.remove(overlay);
+		birdhouses.clear();
+	}
+	
+	private boolean inBirdhouseArea()
```
Nightfirecat: Similarly, this will be unused once `nearBirdhouses` is removed.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+import net.runelite.client.ui.overlay.OverlayManager;
+
+@PluginDescriptor(
+	name = "Birdhouses",
+	description = "highlights the positions of empty birdhouses",
+	tags = {"birdhouse", "fossil island", "highlight"}
+)
+public class BirdhousesPlugin extends Plugin
+{
+	private static final int[] BIRDHOUSE_REGIONS = {
```
Nightfirecat: I believe this will also be unused once `inBirdhouseArea()` is removed.

runelite-client/src/main/java/net/runelite/client/plugins/birdhouses/BirdhousesPlugin.java
```diff
@@ -0,0 +1,136 @@
...
+		NullObjectID.NULL_30566, VarPlayer.BIRDHOUSE_2,
+		NullObjectID.NULL_30567, VarPlayer.BIRDHOUSE_3,
+		NullObjectID.NULL_30568, VarPlayer.BIRDHOUSE_4
+	);
+	
+	@Getter(AccessLevel.PACKAGE)
+	private List<TileObject> birdhouses = new ArrayList<TileObject>();
+	
+	@Inject
+	private Client client;
```
Nightfirecat: This is unused now as well.

Adam-:
I don't really understand why this is useful? We have the new timer plugin already which also tracks birdhouse state.

kristian-welsh:
I was unaware the timer plugin tracks state, this likely isn't useful given that.

kristian-welsh:
Does that plugin draw the state to the screen? The issue referenced forgetting to put seeds in and having no visual indicator of that. If the state is only in the sidebar it would still be easy to miss.

Adam-:
No. I don't think this is a big enough of an issue to require its own plugin.

kristian-welsh:
I agree, but couldn't think of any other place to put it. Perhaps this could be integrated into the timers plugin or some other appropriate place.

kristian-welsh:
Rebased to master and moved code to the hunter plugin.

runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
```diff
@@ -24,26 +24,21 @@
...
 package net.runelite.client.plugins.hunter;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import java.time.Instant;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
```
deathbeam: Fix your imports (see code guidelines wiki page)

runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
```diff
@@ -296,6 +303,32 @@ public void onGameObjectSpawned(GameObjectSpawned event)
...
 	}
 
+	@Subscribe
+	public void onGameObjectChanged(GameObjectChanged event)
+	{
+		onGameObject(event.getPrevious(), event.getGameObject());
+	}
+
+	@Subscribe
+	public void onGameObjectDeSpawned(GameObjectDespawned event)
```
deathbeam: onGameObjectDespawned

runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
```diff
@@ -387,4 +420,21 @@ public void onConfigChanged(ConfigChanged event)
 			overlay.updateConfig();
 		}
 	}
+
+	@Subscribe
+	public void onGameStateChange(GameStateChanged event)
+	{
+		switch (event.getGameState())
```
deathbeam: just clear them on hopping and login_screen, and make the birdhouse array final and initialize it right away

runelite-client/src/main/java/net/runelite/client/plugins/hunter/TrapOverlay.java
```diff
@@ -27,15 +27,21 @@
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.util.List;
 import java.util.Map;
 import javax.inject.Inject;
+
+import com.google.common.collect.ImmutableList;
```
deathbeam: Fix your imports

kristian-welsh: Is the issue here the blank line?

Nightfirecat: Yes, they should all be in the same block without blank lines separating them.

kristian-welsh:
The review is resolved, but doesn't seem to be marked as such. Could someone take a look at this?

runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
```diff
@@ -296,6 +318,32 @@ public void onGameObjectSpawned(GameObjectSpawned event)
...
+	@Subscribe
+	public void onGameObjectChanged(GameObjectChanged event)
+	{
+		onGameObject(event.getPrevious(), event.getGameObject());
+	}
+
+	@Subscribe
+	public void onGameObjectDespawned(GameObjectDespawned event)
+	{
+		onGameObject(event.getGameObject(), null);
```
NathenSample: why not just remove the object from the collection here? Rather than invoking the method with a null. 

Alternatively overload `onGameObject`.

You're having to add null guards to `onGameObject` just because you want to call it with a null.

kristian-welsh:
Good to know. I'll close this since it's unwanted. Someone should close issue#4527 too.

PR #4746, opened 2018-08-06
Remove our custom raid point overlay and make RS one moveable instead

deathbeam:
With ability to make RS widgets moveable, our own custom raid point
overlay is not necessary.


deathbeam:
I will reopen this maybe when I would not be lazy to rebase this :d

Issue #1617, opened 2018-04-17
Right click hop worlds on friendslist/friendschat


Stickaplex:
Nothing to add. Title pretty much says it

deathbeam:
We are not doing this until there will be way to world hop without faking user actions.

PR #4761, opened 2018-08-07
Add world hopper plugin

devLotto:
When this plugin is enabled, pressing ctrl + shift + left/right "quick-hops" you to the previous/next world of the same type as the one you're currently in. 

This means that you hop to a members world if you're in a members world, you hop to a F2P world if you're in a F2P world, you hop to a Deadman world if you're in a Deadman world, and so on.. 

You never hop into a PVP or high risk world, and a special case exists for PVP and high risk worlds, where hopping from one always hops you to a non-PVP world. It also only tries to quick-hop to total level restricted worlds if your total level is adequate.

This pull request only introduces the quickhopping ability, and not a side panel where you can list all worlds. That functionality will most likely be added in the future.

Quickhopping: 
https://gfycat.com/LegitimateVacantGonolek
Using Hop-to: 
https://gfycat.com/MelodicUnsungArgentinehornedfrog
Quickhopping in Tournament worlds (same type): 
https://gfycat.com/LimpingEasygoingLeopardseal

Clan Hop-to: 

Friend Hop-to (only appears if the target is online): 

Chat Hop-to (only appears if you're a friend or in the same clan as the target):


Janzzze:
From the Rejected or Rolled Back Features wiki page
"World Hopping Shortcuts: Not possible to implement without faking user input (see #1617 )"
Has the stance on this changed?

Adam-:
We have mostly mitigated this internally by making it non-trivial to fake widget interactions other than the ones required for the world hop - so I am okay with it.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+	@Override
+	protected void startUp() throws Exception
+	{
+		keyManager.registerKeyListener(inputListener);
+
+		lookupFuture = executorService.scheduleAtFixedRate(() ->
+		{
+			try
+			{
+				WorldResult worldResult = new WorldClient().lookupWorlds();
```
deathbeam: You can just use client API for this, no need to use WorldClient at all

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+		{
+			int menuEntryCount = newMenu.length;
+			ArrayUtils.swap(newMenu, menuEntryCount - 1, menuEntryCount - 2);
+		}
+
+		client.setMenuEntries(newMenu);
+	}
+
+	@Subscribe
+	public void onHopToMenuClicked(PlayerMenuOptionClicked event)
```
deathbeam: onPlayerMenuOptionClicked

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		if (worldToHopTo == -1)
+		{
+			return;
+		}
+
+		if (client.getWidget(WidgetInfo.WORLD_SWITCHER_LIST) == null)
```
deathbeam: I would not open the world hopper explicitly and simply require users to open it at least once

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+		if (client.getGameState() != GameState.LOGGED_IN)
+		{
+			return;
+		}
+
+		int totalLevel = client.getTotalLevel();
+
+		World currentWorld = worldResult.findWorld(client.getWorld());
+
+		// Make it so you always hop out of PVP, high risk and bounty worlds
```
deathbeam: What about people aready on PVP worlds? I would assume if someone is already on PVP world, that person wants to hop to another PVP world.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperInputListener.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	@Override
+	public void keyTyped(KeyEvent e)
+	{
+
+	}
+
+	@Override
+	public void keyPressed(KeyEvent e)
```
c01dc0ffee: Possibly change this to a configurable keybind setting?

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: Why?

runelite-mixins/src/main/java/net/runelite/mixins/WorldHoppingMixin.java
```diff
@@ -0,0 +1,51 @@
...
+public abstract class WorldHoppingMixin implements RSClient
+{
+	@Inject
+	@Override
+	public void openWorldHopper()
+	{
+		// The clicked x & y coordinates (the last arguments) are not processed in the game or sent to Jagex, so they don't have to be real.
+		// Don't use a WidgetInfo, to prevent being able to pass in arbitrary group & child ids.
+		// Also don't use WidgetInfo.PACK to prevent being able to edit what that method returns to pass in an arbitrary packed widget id.
+		menuAction(-1, WidgetID.LOGOUT_PANEL_ID << 16 | WidgetID.LogoutPanel.WORLD_SWITCHER_BUTTON, 57, 1, "World Switcher", "", 658, 384);
```
Adam-: Generally we would access the WidgetInfo and `.getId()` here. See also `PACK`. Even if you get the group/child id off of the widgetinfo, that is preferable to this.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: That is not the route to go if we are planning to have our own world hopper sidebar. Players will complain for days on end at the fact that you must open the in-game world hopper once for our hopper to work at all.

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -129,8 +130,9 @@
 		static final int TEXT = 4;
 	}
 
-	static class LogoutPanel
+	public static class LogoutPanel
```
Adam-: Make this private

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -590,4 +592,9 @@
 		static final int KILLS = 14;
 		static final int STREAK = 15;
 	}
+
+	public static class WorldSwitcher
```
Adam-: Private

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
Adam-: I am pretty sure the client API for world list doesn't work unless the world switcher menu is open.. so we can't use that. This seems fine. Additionally we have seen that they cannot change the world list without restarting everything so I probably would not bother scheduling this at all.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+	void hop(boolean previous)
+	{
+		if (client.getGameState() != GameState.LOGGED_IN)
+		{
+			return;
+		}
+
+		int totalLevel = client.getTotalLevel();
+
+		World currentWorld = worldResult.findWorld(client.getWorld());
```
Adam-: This should account for maybe worldResult being null if it fails to load on plugin start.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,359 @@
...
+		// Make it so you always hop out of PVP, high risk and bounty worlds
+		// Also don't regard skill total worlds as a type that must be hopped between
+		EnumSet<WorldType> currentWorldTypes = currentWorld.getTypes().clone();
+		currentWorldTypes.remove(WorldType.PVP);
+		currentWorldTypes.remove(WorldType.PVP_HIGH_RISK);
+		currentWorldTypes.remove(WorldType.BOUNTY);
+		currentWorldTypes.remove(WorldType.SKILL_TOTAL);
+
+		int currentWorldIndex = worldResult.getWorlds().indexOf(currentWorld);
+		CyclicListIterator<World> worldIterator = new CyclicListIterator<>(worldResult.getWorlds(), currentWorldIndex);
```
Adam-: This entire iterator is just `current +/-= 1; current = min(worldList.size(), max(0, current));`

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
Adam-: Keep the behavior of other clients.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: I'll check exactly what other clients do tomorrow. I was thinking that quickhopping out of PVP would be something people want, but I'll check out the behaviour of other clients.

For the bounty world, I was thinking that you wouldn't want to hop into a world where a lot of people are PKing if you're in the wilderness, but maybe it's not really that problematic.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: Added a nullcheck above

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
deathbeam: Idk, i just feel like it is very-bot like and this is really pushing it.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,394 @@
...
+	{
+		keyManager.registerKeyListener(previousKeyListener);
+		keyManager.registerKeyListener(nextKeyListener);
+
+		worldResultFuture = executorService.submit(() ->
+		{
+			try
+			{
+				WorldResult worldResult = new WorldClient().lookupWorlds();
+				worldResult.getWorlds().sort(Comparator.comparingInt(World::getId));
```
deathbeam: WorldResult can be null so add null-check for it.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,394 @@
...
+				this.worldResult = worldResult;
+			}
+			catch (IOException ex)
+			{
+				log.error(null, ex);
```
deathbeam: I would add some decription to this. Also, make this just warning, error usually means app cannot continue working, this will just make the plugin not work correctly.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,394 @@
...
+	private ChatMessageManager chatMessageManager;
+
+	@Inject
+	private ScheduledExecutorService executorService;
+
+	@Inject
+	private WorldHopperConfig config;
+
+	@Provides
+	WorldHopperConfig getConfig(ConfigManager configManager)
```
deathbeam: Move method below fields

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,394 @@
...
+		});
+	}
+
+	@Override
+	protected void shutDown() throws Exception
+	{
+		keyManager.unregisterKeyListener(previousKeyListener);
+		keyManager.unregisterKeyListener(nextKeyListener);
+
+		worldResultFuture.cancel(true);
```
deathbeam: Reset worldHopTo, worldResult and worldResultFuture here too.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
deathbeam: Some people probably want one and some another, I can definitely see it being useful in both cases, so maybe configuration option.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,394 @@
...
+
+		chatMessageManager
+				.queue(QueuedMessage.builder()
+				.type(ChatMessageType.GAME)
+				.runeLiteFormattedMessage(chatMessage)
+				.build());
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick event)
```
deathbeam: Why is this onGameTick? Can't you just use ClientThread.invokeLater instead where you setting the world to hop to and hop directly there?

devLotto: Hopping in ClientThread.invokeLater doesn't work without waiting anything since the world hopper was opened. I think the source of the issue is that if both openWorldHopper and hopToWorld are parsed in the same tick on the server, it doesn't let you hop.

deathbeam: You can return true from ClientThread.invokeLater and have the method rescheduled for next tick. So you can do same logic as here, but instead of having worldToHopto reset to -1, you just return false from .invokeLater (and true when opening worldHopper)

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: OSBuddy does exactly this, it's ok.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: Okay so OSBuddy's quick hopper always hops out of PvP worlds, and never into them. Bounty is regarded as a completely regular world, so BOUNTY will have to be removed from both currentWorld.types and the types of the world that the current is being compared to.

I'll add a config option for enabling only hopping between PvP worlds (if your current world is a pvp world). If you're in a non-PvP world, never hop into one.

devLotto: Changed it to work like OSBuddy's hopper and added a config option for toggling "always hop out of pvp worlds".

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: I think that doesn't work if you're on world 301 and want to go to the previous world. I made a thing that works for both going to the previous from 301 and going to the next from 424. Maybe it can be improved but it works.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
devLotto: Changed it to do that now, seems to work good.

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -1426,4 +1433,8 @@
 	 * @param speed speed
 	 */
 	void setOculusOrbNormalSpeed(int speed);
+
+	void openWorldHopper();
```
deathbeam: Documentation

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -1426,4 +1433,8 @@
...
+
+	void hopToWorld(int world);
```
deathbeam: Documentation

deathbeam: Also, to prevent hopping to non-existant world, this could maybe take `World` instead. Everywhere we use it, we have `World` at hand already

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldTableHeader.java
```diff
@@ -0,0 +1,120 @@
...
+	private final JLabel textLabel = new JLabel();
+	private final JLabel arrowLabel = new JLabel();
+
+	// Determines if this header column is being used to order the list
+	private boolean ordering = false;
+
+	static
+	{
+		ARROW_DOWN = new ImageIcon(ImageUtil.getResourceStreamFromClass(WorldHopperPlugin.class, "arrow_down.png"));
+		ARROW_UP = new ImageIcon(ImageUtil.getResourceStreamFromClass(WorldHopperPlugin.class, "arrow_up.png"));
```
deathbeam: Remove these images, and use ImageUtil to create them from ARROW_DOWN

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,538 @@
...
+					int totalRequirement = Integer.parseInt(world.getActivity().substring(0, world.getActivity().indexOf(" ")));
+
+					if (totalLevel >= totalRequirement)
+					{
+						types.remove(WorldType.SKILL_TOTAL);
+					}
+				}
+				catch (NumberFormatException ex)
+				{
+					log.warn(null, ex);
```
deathbeam: Maybe more decriptive message?

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,538 @@
...
+		}
+
+		navButton = NavigationButton.builder()
+			.tooltip("World Switcher")
+			.icon(icon)
+			.priority(3)
+			.panel(panel)
+			.build();
+
+		clientToolbar.addNavigation(navButton);
```
deathbeam: This entire panel should be disable-able, as I would at least see features that reimplement parts of game be optional. The shortcuts are still nice to use even without the panel.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -0,0 +1,538 @@
...
+				{
+					return clanMember;
+				}
+			}
+		}
+
+		return null;
+	}
+
+	private static EnumSet<net.runelite.api.WorldType> toWorldTypes(final EnumSet<net.runelite.http.api.worlds.WorldType> apiTypes)
```
deathbeam: This is duplicate of DefaultWorld code, make it not, and create common method for it.

deathbeam:
Where did the `Protect` annotation went :thinking:

runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -382,7 +382,10 @@
 	/**
 	 * Corp beast damage
 	 */
-	CORP_DAMAGE(999);
+	CORP_DAMAGE(999),
+
+	WORLDHOPPER_FAVROITE_1(4597),
+	WORLDHOPPER_FAVROITE_2(4598);
```
oplosthee: Typo, should be favo(u)rite

PR #4766, opened 2018-08-07
Show superior monster health

ltvill:


ltvill:
Some of them might have to be rounded up because a few individual wiki pages show slightly different health.

ltvill:
Some of the lower level superior creatures give XP 10 times bigger than their health, last fixes.

PR #4771, opened 2018-08-08
Add World Switcher Plugin

psikoi:
Adds the full UI implementation for a world switcher plugin.

This includes:

- World list, with fancy new country icons, players and activity indicators.
- World sorting, by world, players or activity
- Adding world to favorites
- Removing world from favorites
- Double clicking to change world
- PVP / Tournament highlighting
- Current world highlighting


runelite-client/src/main/java/net/runelite/client/plugins/worldswitcher/WorldSwitcherPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		navButton = NavigationButton.builder()
+			.tooltip("World Switcher")
+			.icon(icon)
+			.priority(3)
+			.panel(panel)
+			.build();
+
+		clientToolbar.addNavigation(navButton);
```
deathbeam: This entire sidebar should be disable-able in plugin config as if this plugin is also going to have the world hop shortcuts and hop-to, it should not require to have the sidebar present when there is already better version in-game.

deathbeam:
For the arrow-down and arrow-faded, use ImageUtil. All resources you need here is arrow pointing up and then just transform it to other variants.

runelite-client/src/main/java/net/runelite/client/plugins/worldswitcher/WorldSwitcherPlugin.java
psikoi: What's the solution here, then?

psikoi:
Didn't know ImageUtil existed tbh

runelite-client/src/main/java/net/runelite/client/plugins/worldswitcher/WorldSwitcherPlugin.java
deathbeam: Config option for disabling the sidebar I guess

psikoi: Which is?

Nightfirecat:
It's new. :man_shrugging:

Gamer1120:
It might be useful if this takes the login limit into account, such that it will show a warning that hopping would exceed the login limit and how much time it takes before you can hop again. Saves the trouble of re-typing your password. This could also be a separate issue.

psikoi:
@Gamer1120 is this timeout an exact amount of time?

Gamer1120:
@psikoi I'm not 100% sure which time it is exactly, but I think it's that you can login X times per X seconds. I'll try to Google if someone already figured that out.

Gamer1120:
https://www.reddit.com/r/2007scape/comments/533qdl/world_hop_limit/ is not 100% sure, but this could be used as a basis. I'll try to test it.

Adam-:
Merge as part of #4761

PR #4785, opened 2018-08-08
Prevent idle notification after using the animation machine in the Warriors' Guild

crai9:
Animation machine uses the same animations as smithing cannonballs, so added a check to see if the player is at Warriors' Guild before sending an idle notification.


MagicfTail:
The smithing animation is used everywhere for stuff like going down ladders and interacting with objects. It should, therefore, check if you are actually smithing, not have special cases for all the times you are not

crai9:
@MagicfTail didn't know that, are you aware of any other skilling animations that are reused and trigger the idle notifier? I'll add skilling checks instead of location checks if so

PR #4795, opened 2018-08-09
Supplies Tracker

DWhite72:
Rewrote detection method to use onMenuOptionsClicked and onInventoryChanged.
Everything seems to work fine other than if you quickly(same tick) decant a potion and spam click on the new potion it adds an extra dose to the tracker.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
+++ b/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
@@ -111,7 +111,7 @@ public class WidgetID
 
 	static class WorldMap
 	{
-		static final int OPTION = 42;
+		static final int OPTION = 41;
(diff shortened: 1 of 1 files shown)


forsco:
> Zamorak brew shares same event message as saradomin brew so appears as Sara brew.
> Antidotes all share same event message so they all show as antidote++.

Couldn't you just add a check for which item the player is using instead of utilizing chat box messages?

DWhite72:
That would be better I was using what boost tracker. I'll stay rewriting plugin class

DWhite72:
@psikoi @forsco onItemContainerChanged goes off everywhere. If you drop a manta and pick it up it will count as a manta used when dropped and not change when picked up.

forsco:
Check the menu action used on the item

PR #4802, opened 2018-08-09
ChatCommands: Add Clipboard Support

ConnorVG:
Adds CTRL+V and CTRL+C support to the in-game chatbox.

Fixes an issue with errors being thrown when shortcuts were used when not currently logged in.

Extracts clipboard logic (a couple places already had a sort of "copy" functionality) to a simple clipboard util.


MagicfTail:
This is autotyping and should probably not be in the game

ConnorVG:
@MagicfTail yeah... I was wondering if that would be the case, honestly.

Trevor159:
We do not want something like this in the client. It is too similar to an autotyping.

PR #4804, opened 2018-08-09
Implement Client.getWeight() and add RunEnergyPlugin

### LINKED ISSUE #4666: Show time until run energy full and time until depletion

seandewar:
This plugin shows the following tooltip when the user mouses over the run energy orb:
image

And has the option to replace the run energy percentage next to the orb with the remaining run time:

This plugin gives an estimated time until run energy has been depleted and until run energy recharges to 100%. It also shows the user's weight (as shown in the equipment screen).

The plugin considers the Graceful set effect (with the Agility and Max Cape substitution) and stamina boosts. Weights under 0 kg have no effect on the estimated depletion time as expected.

Closes #4666.


runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyOverlay.java
```diff
@@ -0,0 +1,225 @@
...
+		GRACEFUL_BOOTS_13677, GRACEFUL_BOOTS_13678, GRACEFUL_BOOTS_21076, GRACEFUL_BOOTS_21078
+	);
+
+	// Agility skill capes also count for the Graceful set effect
+	private static final ImmutableSet ALL_GRACEFUL_CAPES = ImmutableSet.of(
+		GRACEFUL_CAPE, GRACEFUL_CAPE_11853, GRACEFUL_CAPE_13581, GRACEFUL_CAPE_13582,
+		GRACEFUL_CAPE_13593, GRACEFUL_CAPE_13594, GRACEFUL_CAPE_13605, GRACEFUL_CAPE_13606,
+		GRACEFUL_CAPE_13617, GRACEFUL_CAPE_13618, GRACEFUL_CAPE_13629, GRACEFUL_CAPE_13630,
+		GRACEFUL_CAPE_13669, GRACEFUL_CAPE_13670, GRACEFUL_CAPE_21064, GRACEFUL_CAPE_21066,
+		AGILITY_CAPE, AGILITY_CAPET, AGILITY_CAPE_10648, AGILITY_CAPE_13340, AGILITY_CAPET_13341
```
MagicfTail: Also max cape

seandewar: Good catch! This only includes the non-cosmetic versions, right? (e.g. doesn't include the max fire cape, for example)

MagicfTail: Ye. Also, why are there 42 different agility capes? Are you sure some of them aren't mounted versions or something?

seandewar: Is there any way to check?

MagicfTail: Don't think so no, but it looks like `AGILITY_CAPE` and `AGILITY_CAPET` are the correct items

seandewar: Okay, cheers! I'm also not sure about the huge amount of Graceful item variations but I'm unsure of how to check those. :(

MagicfTail: Ignore that comment lol

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyOverlay.java
```diff
@@ -0,0 +1,226 @@
...
+
+	private String getEstimatedRecoverTimeRemaining()
+	{
+		// Calculate the amount of energy recovered every second
+		double recoverRate = (48 + client.getBoostedSkillLevel(Skill.AGILITY)) / 360.0;
+
+		if (isWearingFullGraceful())
+		{
+			// 30% recover rate increase from Graceful set effect
+			recoverRate *= 1.3;
```
Kamielvf: Graceful doesn't work like this. It only applies the effect if you are standing still

seandewar: Cheers. I'll have a look in a sec. :)

seandewar: If I understand correctly, run energy only ever regenerates when the player's not running anyway (and not busy doing something such as crossing agility obstacles).

In this case, maybe it would be useful to hide the recovery time while running?

seandewar:
I've changed things so that the recovery time displays N/A when running to a destination (moving faster than 1 tile per tick with a waypoint set).

Also, I've added the config ability to replace the run orb text with the seconds of run remaining.

I might reword the tooltip contents to make it less verbose and maybe add a color shift to the tooltip text when less than 15s of run energy remains.

Let me know what you think. :)

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+
+@PluginDescriptor(
+	name = "Run Energy",
+	description = "Show various information related to run energy",
+	tags = {"overlay"}
+)
+public class RunEnergyPlugin extends Plugin
+{
+	// TODO It would be nice if we have the IDs for just the equipped variants of the Graceful set items.
+	private static final ImmutableSet ALL_GRACEFUL_HOODS = ImmutableSet.of(
```
Adam-: `ImmutableSet<Integer>`

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+
+	private boolean isStaminaEffectActive()
+	{
+		final Widget runOrbIcon = client.getWidget(WidgetInfo.MINIMAP_RUN_ORB_ICON);
+
+		return (runOrbIcon != null &&
+			runOrbIcon.getSpriteId() == SpriteID.MINIMAP_ORB_RUN_ICON_SLOWED_DEPLETION);
+	}
+
+	public String getEstimatedRunTimeRemaining(boolean inSeconds)
```
Adam-: this should be package private

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+
+		return (ALL_GRACEFUL_HOODS.contains(items[EquipmentInventorySlot.HEAD.getSlotIdx()].getId()) &&
+			ALL_GRACEFUL_TOPS.contains(items[EquipmentInventorySlot.BODY.getSlotIdx()].getId()) &&
+			ALL_GRACEFUL_LEGS.contains(items[EquipmentInventorySlot.LEGS.getSlotIdx()].getId()) &&
+			ALL_GRACEFUL_GLOVES.contains(items[EquipmentInventorySlot.GLOVES.getSlotIdx()].getId()) &&
+			ALL_GRACEFUL_BOOTS.contains(items[EquipmentInventorySlot.BOOTS.getSlotIdx()].getId()) &&
+			ALL_GRACEFUL_CAPES.contains(items[EquipmentInventorySlot.CAPE.getSlotIdx()].getId()));
+	}
+
+	public String getEstimatedRecoverTimeRemaining()
```
Adam-: package private

seandewar:
Changes added. Also changed the recovery time left text to be less verbose. :)

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyOverlay.java
```diff
@@ -0,0 +1,87 @@
...
+		if (bounds.contains(mousePosition.getX(), mousePosition.getY()))
+		{
+			final String tooltip = "Weight: " + client.getWeight() + " kg</br>" +
+				"Run Time Remaining: " + plugin.getEstimatedRunTimeRemaining(false) + "</br>" +
+				"100% Energy In: " + plugin.getEstimatedRecoverTimeRemaining();
+
+			tooltipManager.add(new Tooltip(tooltip));
+		}
+
+		return new Dimension((int) bounds.getWidth(), (int) bounds.getHeight());
```
deathbeam: You can just return bounds.getSize() or simply null as this is not layoutable overlay.

seandewar: Thanks. I've fixed this. 👍

ShaunDreclin:
👀👌🏻

PR #4809, opened 2018-08-09
Use working sprite for puzzle solver arrows

deathbeam:
For some reason 422.1 do not works for puzzle solver arrows (throws NPE,
e.g is non-existant) so just use different sprite (BH) and scale it
down.

preview

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
@@ -42,7 +42,7 @@ import net.runelite.api.GameState;
 import net.runelite.api.InventoryID;
 import net.runelite.api.Item;
 import net.runelite.api.ItemContainer;
-import static net.runelite.api.SpriteID.MINIMAP_DESTINATION_FLAG;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -425,7 +426,11 @@ private void solve(int[] items)
 
 	private BufferedImage getDownArrow()
 	{
-		return spriteManager.getSprite(MINIMAP_DESTINATION_FLAG, 1);
+		if (downArrow == null)
+		{
+			downArrow = ImageUtil.resizeImage(spriteManager.getSprite(SpriteID.MINIMAP_GUIDE_ARROW_YELLOW, 0), 16, 16);
```
Nightfirecat: Should be 15x15 to match the intended image size, but not a big deal.

Adam-:
Do we know why the other one doesn't work?

Kamielvf:
@Abextm might know, he looked into it a bit yesterday.

From discord:
`The primary issue is that I didn't actually read the cache code, I just assumed it was the same as something else. Look at method2870 vs getIndexedSprites`

PR #4823, opened 2018-08-11
"Automatically look North after teleportation" - Function

adm1n1stra1tor:
"Look North"
Enabling this function makes you automatically face north after each teleportation.


deathbeam:
Open up issue for this

PR #4856, opened 2018-08-11
Add the ability to display remaining prayer time in orb

seandewar:
Adds a configuration option for showing the remaining prayer time next to the orb in seconds:

image

If the user is not praying, the percentage will be shown instead as normal.

Similar to the feature proposed in PR #4804 for run energy.


MagicfTail:
This is already a feature? https://i.imgur.com/vIfXzkG.png

MagicfTail:
Nvm, i see

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerPlugin.java
```diff
@@ -265,6 +306,53 @@ private boolean isAnyPrayerActive()
 		return false;
 	}
 
+	double getPrayerDrainRate(Client client)
```
deathbeam: This can be private

deathbeam: Also remove Client from parameters

dekvall:
Added in #9599

PR #4861, opened 2018-08-11
[WIP] Add Loot Tracker Group View

psikoi:
Adds a new viewing mode, grouped view. This mode groups the loot by NPC/Event name and can be toggled on/off in the plugin configs.

TO-DO: Figure out how to reassemble single logs from a grouped log.


psikoi:
Re-writting this, it's bad.

PR #4864, opened 2018-08-12
idlenotifier: Use an interaction check for fishing

Abextm:
This is nearly instant and works alot better than just waiting for animation to change

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -40,6 +40,7 @@ import net.runelite.api.Varbits;
 import net.runelite.api.events.AnimationChanged;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
+import net.runelite.api.events.InteractingChanged;
 import net.runelite.client.Notifier;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.plugins.Plugin;
@@ -72,6 +73,7 @@ public class IdleNotifierPlugin extends Plugin
 	private boolean notifyPrayer = true;
 	private boolean notifyIdleLogout = true;
 	private boolean notify6HourLogout = true;
+	private boolean wasFishing = false;
 
 	private Instant sixHourWarningTime;
 	private boolean ready;
@@ -110,11 +112,11 @@ public class IdleNotifierPlugin extends Plugin
 			case WOODCUTTING_DRAGON:
 			case WOODCUTTING_INFERNAL:
 			case WOODCUTTING_3A_AXE:
-			/* Cooking(Fire, Range) */
+				/* Cooking(Fire, Range) */
 			case COOKING_FIRE:
 			case COOKING_RANGE:
 			case COOKING_WINE:
-			/* Crafting(Gem Cutting, Glassblowing, Spinning, Battlestaves) */
+				/* Crafting(Gem Cutting, Glassblowing, Spinning, Battlestaves) */
 			case GEM_CUTTING_OPAL:
 			case GEM_CUTTING_JADE:
 			case GEM_CUTTING_REDTOPAZ:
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -263,6 +252,25 @@ public void onGameTick(GameTick event)
 		}
 	}
 
+	@Subscribe
+	public void onInteractingChanged(InteractingChanged ev)
+	{
+		Actor source = ev.getSource();
+		if (source != client.getLocalPlayer())
```
Nightfirecat: Need to add `fishingIdle` check as well; currently the config isn't used.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -263,6 +252,30 @@ public void onGameTick(GameTick event)
...
+		}
+
+		Actor target = ev.getTarget();
+		boolean isFishing = target != null && target.getName().contains("Fishing spot");
+
+		if (wasFishing && !isFishing)
+		{
+			notifier.notify("[" + source.getName() + "] has stopped fishing!");
+		}
+		wasFishing = isFishing;
```
deathbeam: WasFishing probably also needs to be reset in resetTimers

deathbeam: Oh right, just noticed that this is not using the idle delay at all. it should.

Abextm:
Why? The point of this is that it doesn't need a delay.

Abextm:
The previous one needed delay for if you stopping animating but were still
actually doing the action

deathbeam:
Well alright. But as I said, it is not about "needing" it, but to have option to delay the notification. What would this change remove from fishing.

Abextm:
I have always seen and used it as a workaround. If you don't want it to
notify you while you are active use the notify when focused global option.
If you want I can add a separate option for things that do not need a
workaround, but that seems confusing.

deathbeam:
I was always using the idle notifier with delay of 3 to 5 seconds to AFK slightly more, as otherwise sometimes the notifications can be quite annoying. I would just keep existing behaviour and just trigger the delay earlier with your change. Then you can just set the delay to 0 and it will fire immediately.

Nightfirecat:
The delay seemed like a workaround for idle animation detection. Since this will not depend on idle animations, I don't see the need for artificially delaying the notification.

deathbeam:
Well this is still removing feature that I use and plan to use in future too. Splitting the config so you dont have to reconfigure it does not belongs to this PR as well as that is different issue and would be weird if there was special delay just for fishing. So you should simply add this and take existing config into consideration and make it behave like it behaved before this change by default.

Abextm:
I still don't even understand why you want that delay.

deathbeam:
Mutliple reasons:
1. When I set bigger delay, I can AFK slightly more at the cost of XP/hr
2. When the fishing spot switches too fast it can be annoying to receive the notification straight away

Abextm:
1. Perhaps rather than delaying the notification, just don't respond to it immediately?
2. So you just want it to not be accurate pseudo-randomly?

deathbeam:
1. Its easy to actually forget to respond to notification completely if I do that
2. No, I just want the configurable delay to avoid that, the new start of the fishing idle is fine

Abextm:
So what you really want is:
1) for it to re-notify you if you haven't responded?
2) not notify while you have the client focused?

deathbeam:
1. No, because I would not want the first notification in such scenario. I simply want to delay every AFK notification I have by X seconds, so exactly how is it working on master rn
2. No because I sometimes switch desktops immediatelly after clicking so having the client focused or not does not matter.

Abextm:
So it seems like what we should have is 
a) The old option renamed to ` Animation Hysteresis`
b) A new (global?) option named `Notification Delay`
and have this only apply the second delay, but the animation ones use both

deathbeam:
That sounds good, but the thing is users wouldn't know that the fishing idle alert is no longer animation-related as it will be after this change only special case in idle notifier for skilling-related alerts, and I am unsure how to present that so it will not cause confusion.

Nightfirecat:
FWIW I think your use case is not that of most players. (and maybe not of other clients?) I think most people would want a notification immediately on idle, and they would respond to it on their own time.

deathbeam:
Implemented in #6457

PR #4869, opened 2018-08-12
Agility: fix brimhaven timer when falling

### LINKED ISSUE #4859: Brimhaven agility arena timer resets when falling off obstacles
Brimhaven agility arena timer resets when falling off obstacles.

**Expected behavior**
Timer should continue to count down.

 - OS: Windows
 - RuneLite version: 1.4.11
 - Launcher version: 1.6.0

Justsofun:
Fixes #4859

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -242,16 +242,21 @@ public class AgilityPlugin extends Plugin
 			if (!Objects.equals(lastArenaTicketPosition, newTicketPosition))
 			{
 				// We don't want to notify when players first enter the course
-				if (lastArenaTicketPosition != null && newTicketPosition != null)
+				if (lastArenaTicketPosition != null)
 				{
-					if (config.notifyAgilityArena())
+					// We want to ignore the plane so that the timer won't reset when falling and climbing back up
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -242,16 +242,21 @@ public void onGameTick(GameTick tick)
 			if (!Objects.equals(lastArenaTicketPosition, newTicketPosition))
 			{
 				// We don't want to notify when players first enter the course
-				if (lastArenaTicketPosition != null && newTicketPosition != null)
+				if (lastArenaTicketPosition != null)
 				{
-					if (config.notifyAgilityArena())
+					// We want to ignore the plane so that the timer won't reset when falling and climbing back up
+					if (!(lastArenaTicketPosition.getX() == newTicketPosition.getX()
```
Adam-: How does the hint arrow work in agility arena when you fall? It moves it down a plane? And I don't think it is safe to assume it is never null.

Justsofun: Yeah it moves to the plane 0, as far as I'm avare it's never null when the player is in the arena but I can add the check back

Adam-:
fixed in d1d6d26b60396244e6420b22cb8bd467bbdb0564

PR #4871, opened 2018-08-12
Actually add special case for Grotesque Guardians

MagicfTail:


Adam-:
Has anyone tested this?

MagicfTail:
Tested and actually works this time

Kamielvf:
So is `DUSK_7888` completely redundant at this point here?

MagicfTail:
It looks like it isn't necessary, but same as with normal gargoyles, he changes into another npc during his death animation, and gargoyles have both ID's in there. So i would keep it there unless someone can confirm it isn't necesserary for gargoyles either.

PR #4872, opened 2018-08-12
Add plugin to track and display stats from your performance in Castle Wars

cw-dev:
Keeps track of various stats while playing a Castle Wars game and prints a post-game summary in the chatbox: https://i.imgur.com/NqDwdnJ.png


deathbeam:
Cleaned up the PR a bit.

1. Removed tracking of stuff that required way too much logic and was hard to follow and understand or was working only in specific scenarios (e.g removed all barricade tracking, and freeze/splash tracking, as that one worked mostly only for ice barrage).
2. Removed unused stuff like checking if player is wearing CW brace etc
3. Properly encapsulated everything

I will post screenie in a bit, I just need to play some CW game

deathbeam:
Updated screenshot:

Adam-:
This is so niche, should it be off by default?

runelite-client/src/main/java/net/runelite/client/plugins/castlewars/data/CWTeam.java
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	public CWTeam opposite()
+	{
+		return this == ZAM ? SARA : this == SARA ? ZAM : NONE;
+	}
+
+	public CWFlag getFlag()
+	{
+		return FLAGS.get(this);
```
Adam-: why not just store the flag on the team?

Unmoon:
Idea on how to display the stats:
```
Castle Wars - 24v24 - World 334
Zamorak 0 - 0 Saradomin
Captures: 0 Saves: 0 Deaths: 7 Kills: 9 Stuns: 0 Freezes: 0
Damage dealt: 2460 (Max 62) Damage taken: 1728 (Max 55)
```
Kills could be removed if it can't be reliably tracked.
Stuns instead of Spears since it's a stun spec, even though OSRS doesn't really have other stun sources.
No need for all the `|`, `~` and `.` business.

I kinda like the advanced stats that were in the initial PR. Those would only be useful for more serious CW players, but would provide valuable information to improve one's play. Casting stats not so much.

deathbeam:
I don't think so @Adam- , it makes the end of the CW game way more interesting, and with the stuff I left/removed it is not so niche anymore imo.

@Unmoon hmmm that looks a bit better. Yeah I don't think that kills are possible to track. I removed the advanced stats because they were unreliable and added unnecessary complexity.

deathbeam:
Updated image after @Unmoon 's requests

deathbeam:
As per @Adam- 's request disabled the plugin by default (even though I still disagree)

Adam-:
I don't like this at all. We have much, much less niche plugins that are off by default like the loot tracker. Having a config option for "do more things" is silly. I'm fine with it if 5ea5f204de31c85b0c7e7410758951b2f794ac45 is reverted.

deathbeam:
Well yeah but loot tracker works everywhere, this works only in this particular activity

PR #4874, opened 2018-08-12
Add wasd camera plugin

Adam-:


devLotto:
I just came to think of French keyboards (AZERTY), as they have ZQSD in place of WASD, and DVORAK keyboards that has <AOE. Maybe Keybind should be used with WASD as default?

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+	@Inject
+	private ClientThread clientThread;
+
+	@Inject
+	private KeyManager keyManager;
+
+	@Inject
+	private WASDCameraListener inputListener;
+
+	boolean typing;
```
deathbeam: private and `@Getter`

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+	{
+		clientThread.invoke(() ->
+		{
+			if (client.getGameState() == GameState.LOGGED_IN)
+			{
+				unlockChat();
+			}
+		});
+
+		keyManager.unregisterKeyListener(inputListener);
```
deathbeam: reset `typing` to false

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraListener.java
```diff
@@ -0,0 +1,242 @@
...
+	public void keyPressed(KeyEvent e)
+	{
+		if (!plugin.chatboxFocused())
+		{
+			return;
+		}
+
+		if (!plugin.typing)
+		{
+			if (config.up().matches(e))
```
deathbeam: Most of this method is duplicated in keyReleased and keyTyped, extract this to common method and use it in all listeners

Issue #4820, opened 2018-08-11
Player indicators highlight recolors level when it should only recolor player name


Nightfirecat:
Previously (see gif in #1154) only the player name was recolored, not the player level. I believe this is an unintended side effect of a9d29a70f4f3b8b6096e27cc9b6ec3fdb7708e6c. (perhaps there are two color tags in the vanilla menu entry--one for the player and another for the level indicator?)

![2018-08-11_00-06-34](https://user-images.githubusercontent.com/2199511/43989251-ec39df6a-9d35-11e8-82d6-a3d598148036.png)

Adam-:
fixed in 3c34b5f338089d8d70399d8d553e6bfab6aa1269

PR #4875, opened 2018-08-12
player indicators: Fix menu coloring

Nightfirecat:
This fixes a regression from a9d29a7 where instead of coloring only a
player's name with the "Colorize player menu" option enabled, the full
menu entry (name and level) was colored instead.

Fixes runelite/runelite#4820


Adam-:
Since the original change was not fixing anything (that I know of?) I just reverted it, which is much nicer.

Nightfirecat:
That's fine, though it brings back the mostly-wrong comment in that code. :/

PR #4884, opened 2018-08-13
Fix exp tracker UI margin incosistency

psikoi:
This reduces the top margin of the exp tracker UI to match the other sides and the similarly designed Loot Tracker.

Before:

After:


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
```diff
@@ -67,7 +67,7 @@
 	{
 		super();
 
-		setBorder(new EmptyBorder(10, 6, 10, 6));
+		setBorder(new EmptyBorder(5, 6, 10, 6));
```
deathbeam: Shouldn't this be 6 and not 5?

PR #4886, opened 2018-08-13
Add friends list title counter

forsco:
On behalf of @ConnorVG

Adds a friends list counter to the tabs title.


runelite-client/src/main/java/net/runelite/client/plugins/friendchat/FriendChatPlugin.java
```diff
@@ -0,0 +1,74 @@
...
+		final String title = new StringBuilder("Friends - W")
+			.append(world)
+			.append(" (")
+			.append(friends)
+			.append("/")
+			.append(limit)
+			.append(")")
+			.toString();
+
+		friendListTitleWidget.setText(title);
```
deathbeam: You should revert the widget to original text on shutdown

ConnorVG: Which would make more sense here, do you think?
1) We set it to what it would be manually (ie build the "Friends List - World X" string;
2) We store the previous value on GameStateChange and re-set that in onShutdown.

Being that it's a simple fixed format, 1 makes most sense to me but I'm wondering what you'd prefer.

deathbeam: I think we do 1. with combat info plugin, so probably that, as its really simple scenario.

forsco: If it's set manually, wouldn't that persist after a world hop?

forsco: This has now been addressed.

runelite-client/src/main/java/net/runelite/client/plugins/friendchat/FriendChatPlugin.java
```diff
@@ -0,0 +1,92 @@
...
+@Slf4j
+public class FriendChatPlugin extends Plugin
+{
+	@Inject
+	private Client client;
+
+	@Override
+	protected void shutDown()
+	{
+		if (client.getGameState() != GameState.LOGGED_IN)
```
deathbeam: No need to do this check, the widget will simply be null

runelite-client/src/main/java/net/runelite/client/plugins/friendchat/FriendChatPlugin.java
```diff
@@ -0,0 +1,86 @@
...
+	@Inject
+	private Client client;
+
+	@Override
+	protected void shutDown()
+	{
+		setFriendsListTitle("Friends List - World " + client.getWorld());
+	}
+
+
```
deathbeam: Remove this empty line

forsco: Has now been removed

runelite-client/src/main/java/net/runelite/client/plugins/friendchat/FriendChatPlugin.java
```diff
@@ -0,0 +1,85 @@
...
+import net.runelite.api.widgets.Widget;
+import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+
+@PluginDescriptor(
+	name = "Friend Chat",
+	description = "Add extra information to the friends chat"
+)
+@Slf4j
```
deathbeam: This looks unused, remove

runelite-client/src/main/java/net/runelite/client/plugins/friendchat/FriendChatPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+import net.runelite.api.Client;
+import net.runelite.api.VarPlayer;
+import net.runelite.api.events.GameTick;
+import net.runelite.api.widgets.Widget;
+import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+
+@PluginDescriptor(
+	name = "Friend Chat",
```
deathbeam: I would probably rename this to friend list from friend chat, as friend chat is basically another name for clan chat, can be confusing

forsco: Pushed.

Issue #3839, opened 2018-06-16
Make the clue overlay only show on hover-over


Grabbin:
**Is your feature request related to a problem? Please describe.**
Especially when in fixed mode, the clue overlay can take up a lot of your screen (pic for reference: https://i.imgur.com/9Vg2Gn2.png )

**Describe the solution you'd like**
Instead of the whole box, just show a small icon with a clue, when you hover over that, it brings up the whole text.

**Describe alternatives you've considered**
Maybe make it a toggle? Not sure.

**Additional context**
/

Jasper-ketelaar:
I like having it open at all times. For fixing instances like this I think a better solution would be to hide the clue text considering it adds very little helpful information.

Grabbin:
I can see how some people would probably like to keep it the way it is right now, so maybe an option in the clue scroll plugin to toggle the display?

deathbeam:
With #6203 this request is no longer necessary

Issue #3644, opened 2018-06-05
Large clue scroll overlay takes up too much screen real estate 


RuneliteFan:
Certain clue scroll steps have a large overlay that overtakes the screen (example here: https://imgur.com/27xB69J). Three possible solutions discussed include 1. Condensing the amount of information displayed to include only relevant information  2. Introduce a clue scroll helper in the sidebar so that information does not interfere with the screen or 3. Add the ability to hide/unhide clue steps

RuneliteFan:
https://imgur.com/a/o5G2CLQ

I am going to compile a list of steps that are too large on fixed screen.

laagone:
Support for making a panel on the sidebar, that way it's always available if you want it but it doesn't take up screen space on the game. It's distracting and can get there even when you just pick up a clue scroll and haven't opened it (https://i.imgur.com/A8IE4TR.png), and turning off the plugin is the only way to get rid of it but that's not a very elegant method.

deathbeam:
This was solved by removing full cryptic clue descriptions from overlay

Issue #1871, opened 2018-04-24
Option to have raids overlay stay up during the raid


Stickaplex:
Have the option to have the raids overlay (the scouting helper) stay up permanently during the raid.

-Would fill out unknowns as the raids load more.
-Maybe have an option to 'minimize' it so its not fully in the way in additon to bring it back to its normal size

Kamielvf:
I have this almost fully working, I'm just rewriting some of it and cleaning code up. I hope to PR it soon.

chsami:
Awesome keep up the great work!

Op di 24 apr. 2018 11:06 schreef Kamiel <notifications@github.com>:

Stickaplex:
@Kamielvf does this include the 2nd part of this?

Kamielvf:
Yes, plus some more features.

Aciz:
@Kamielvf I don't know if you're still working on this, but I have a few suggestions:

* Option to print the raids scouter layout & rooms into chatbox, similar to how OSBuddy does, rather than keeping the overlay always open
* Option to recolor the letters in layout string, to make it easier to follow the layout of the raid. A quick mockup I made, colors could be different or even user configurable.

![image](https://user-images.githubusercontent.com/14221121/42628840-fefb1ce0-85d9-11e8-975a-7e377d24df14.png)

PR #4911, opened 2018-08-15
cluescrolls: Declutter clue overlay

raiyni:
Can maybe close https://github.com/runelite/runelite/issues/3839 and https://github.com/runelite/runelite/issues/3644.

Add options to disable parts of the clue overlay.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
@@ -34,10 +34,67 @@ public interface ClueScrollConfig extends Config
 	@ConfigItem(
 		keyName = "displayHintArrows",
 		name = "Display hint arrows",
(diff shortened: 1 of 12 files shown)


deathbeam:
`@Inject` will not work here

MagicfTail:
If anything, this should be a button on the overlay itself to minimize the clue overlay, similar to the raid overlay in the comment of #1871 . Toggling the overlay on and off in the config isn't a good solution imo.

raiyni:
@deathbeam [here](https://github.com/runelite/runelite/pull/4911/commits/f3913ff4c0288d409c306aca297f6b898e2361de#diff-8064a68b8189af2375e716bd01fbe36eR53) or for all files? Worked fine when I tested it.

deathbeam:
In all files except plugin, idk how it could've worked, but it is definitely wrong as Guice is not creating the classes

raiyni:
@MagicfTail I feel like some people would still want to see relevant info and hide clutter.

Issue #3323, opened 2018-05-25
Discord Integration displays 'Training Agility' while barbarian fishing


Fyrecean:
Should probably display whichever skill is gaining the most xp over time if multiple skills are increasing

Infinitay:
I think it might be displaying agility because its the last skill to gain xp in (fishing -> strength -> agility), although I'm not too positive

deathbeam:
Fixed in #9061

Issue #1151, opened 2018-03-27
Discord Training Plugin defaults to strength when using controlled


Stickaplex:
Using controlled defaults to strength on the discord plugin

This: basically https://i.imgur.com/SwjeENe.png

Nightfirecat:
What does it look like if that line extends beyond the size of that tooltip? Let's say you're training 7 skills at once (attack, strength, defense, and hitpoints via the controlled attack style; ranged from a cannon; prayer from a bonecrusher; and slayer from the monsters being killed)--I can't imagine that we could display `Training: Attack, Strength, Defence, Ranged, Prayer, Hitpoints, Slayer` and not have the text wrap or be clipped somehow.

PR #4913, opened 2018-08-15
Display Three Skills at once, and Order Skills Based on Xp Gained

SquirrelHub:
These can maybe close issues #3323, and #1151.
Slayer is by default set to have xp at 200,000,000 to maintain priority over all skills now that xp is the first check.
Only three skills are displayed because state has a maximum length of 128 characters.


SquirrelHub:
The way I handled this problem previously seems to be no longer viable. I would need to revisit this problem again in order to solve it, but unfortunately, I don't have as much time for this project as I did when I submitted the request.

PR #4924, opened 2018-08-16
More World hopper bug fixes

psikoi:
Bugs reported by @devLotto.

- Add to favorite/Remove from favorites menu options now update when clicked
- When you hop off a world it will now have a correct coloring on the activity label
- Remove default world hightlighting to prevent issues with the default world plugin running after the list has been painted


devLotto:
These fixes are ok and don't introduce any new bugs as far as I can see.

PR #4925, opened 2018-08-16
Loot tracker grouping loot by NPC/Event

psikoi:
This introduces a new view mode: Grouping by NPC/Event

- Reversible view mode, meaning you can go back to single loots, or, go from single to grouped.
- Adds kill count indicator to grouped logs
- Fixes a small ui margin that was annoying me.
- Adds detailed view for npc/event

peek 2018-09-12 16-04


DannysPVM:
That looks fantastic @psikoi, keep up the great work man.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -160,7 +183,7 @@ public void onPlayerLootReceived(final PlayerLootReceived playerLootReceived)
 		final String name = player.getName();
 		final int combat = player.getCombatLevel();
 		final LootTrackerItemEntry[] entries = buildEntries(stack(items));
-		SwingUtilities.invokeLater(() -> panel.addLog(name, combat, entries));
+		SwingUtilities.invokeLater(() -> panel.addEntry(name, combat, entries, config.groupLoot()));
```
deathbeam: I would just do setGroupLoot on the panel from configChanged instead of adding new method param

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -149,7 +172,7 @@ public void onNpcLootReceived(final NpcLootReceived npcLootReceived)
 		final String name = npc.getName();
 		final int combat = npc.getCombatLevel();
 		final LootTrackerItemEntry[] entries = buildEntries(stack(items));
-		SwingUtilities.invokeLater(() -> panel.addLog(name, combat, entries));
+		SwingUtilities.invokeLater(() -> panel.addEntry(name, combat, entries, config.groupLoot()));
```
deathbeam: See my next comment

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -203,7 +226,7 @@ public void onWidgetLoaded(WidgetLoaded event)
 		if (!items.isEmpty())
 		{
 			final LootTrackerItemEntry[] entries = buildEntries(stack(items));
-			SwingUtilities.invokeLater(() -> panel.addLog(eventType, -1, entries));
+			SwingUtilities.invokeLater(() -> panel.addEntry(eventType, -1, entries, config.groupLoot()));
```
deathbeam: See my previous comment

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -141,6 +153,17 @@ protected void shutDown()
...
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged ev)
+	{
+		if (!LootTrackerConfig.GROUP_KEY.equals(ev.getGroup()))
+		{
+			return;
+		}
+
+		panel.rebuildLogs(config.groupLoot());
```
deathbeam: See my comment about config.groupLoot()

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -131,39 +142,105 @@ private static String htmlLabel(String key, long value)
...
-		overallGp += box.getTotalPrice();
+		overallGp += log.getTotalPrice();
 		overallKills += 1;
 		updateOverall();
 
+		logs.add(log);
+		logsContainer.add(log, 0);
+	}
+
+	void rebuildLogs(boolean groupLoot)
```
deathbeam: rename this to refresh, remove method param

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -40,13 +43,29 @@
 import net.runelite.client.util.StackFormatter;
 
 @Getter
-class LootTrackerBox extends JPanel
+class LootTrackerLog extends JPanel
```
deathbeam: LootTrackerBox was better name as it indicated more that it was component, this is more confusing. Use the original name

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -40,13 +43,29 @@
...
+	private final JLabel priceLabel = new JLabel();
+	private final JLabel subTitleLabel = new JLabel();
+
+	private final ItemManager itemManager;
+
+	private long totalPrice;
+	private int totalKills;
+
+	@Getter
+	private String title;
```
deathbeam: Just override equals and compare 2 lootTrackerBoxes instead of exposing title just for that

deathbeam: Also, rename this to id or something

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -102,8 +146,52 @@
 			itemContainer.add(slotContainer);
 		}
 
-		add(logTitle, BorderLayout.NORTH);
-		add(itemContainer, BorderLayout.CENTER);
+		itemContainer.repaint();
+	}
+
+	private List<LootTrackerItemEntry> getCombinedItems()
```
deathbeam: Remove this method and just move it to rebuildItems() and calculate total in same loop

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -102,8 +146,52 @@
...
+				list.add(entry);
+			}
+		}
+
+		list.sort((i1, i2) -> i1.getPrice() < i2.getPrice() ? 1 : -1);
+
+		return list;
+	}
+
+	private List<LootTrackerItemEntry> getAllItems()
```
deathbeam: just iterate the entries in the rebuildItems, no need for new list

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -61,36 +80,61 @@
...
+			priceLabel.setText(StackFormatter.quantityToStackSize(totalPrice) + " gp");
 		}
 
+		if (totalKills > 1)
+		{
+			subTitleLabel.setText("x " + totalKills);
+		}
+	}
+
+	private void rebuildItems()
```
deathbeam: rename this to buildItems

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -61,36 +80,61 @@
...
+
+	void addEntry(LootTrackerEntry entry)
+	{
+		entries.add(entry);
+		totalKills++;
+		totalPrice = calculatePrice();
+
+		rebuildItems();
 
 		if (totalPrice > 0)
```
deathbeam: Just always append the price unconditionally to make the look consistent

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -61,36 +80,61 @@
...
 		{
-			final JLabel priceLabel = new JLabel(StackFormatter.quantityToStackSize(totalPrice) + " gp");
-			priceLabel.setFont(FontManager.getRunescapeSmallFont());
-			priceLabel.setForeground(ColorScheme.LIGHT_GRAY_COLOR);
-			logTitle.add(priceLabel, BorderLayout.EAST);
+			priceLabel.setText(StackFormatter.quantityToStackSize(totalPrice) + " gp");
 		}
 
+		if (totalKills > 1)
```
deathbeam: same here, make it x 1

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -40,13 +43,29 @@
...
+
+	private long totalPrice;
+	private int totalKills;
+
+	@Getter
+	private String title;
+
+	private List<LootTrackerEntry> entries = new ArrayList<>();
+
+	LootTrackerLog(final ItemManager itemManager, LootTrackerEntry entry)
```
deathbeam: Having both initial construction from entry and method addEntry seems weird.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
```diff
@@ -115,12 +203,12 @@ private String buildToolTip(LootTrackerItemEntry item)
 		return name + " x " + quantity + " (" + StackFormatter.quantityToStackSize(price) + ")";
 	}
 
-	private static long calculatePrice(final LootTrackerItemEntry[] itemStacks)
+	private long calculatePrice()
```
deathbeam: see my comment about calculating total

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -131,39 +142,105 @@ private static String htmlLabel(String key, long value)
 		return String.format(HTML_LABEL_TEMPLATE, ColorUtil.toHexColor(ColorScheme.LIGHT_GRAY_COLOR), key, valueStr);
 	}
 
-	void addLog(final String eventName, final int actorLevel, LootTrackerItemEntry[] items)
+	void addEntry(final String eventName, final int actorLevel, LootTrackerItemEntry[] items, boolean groupLoot)
```
deathbeam: Maybe you should just send the LootTrackerEntry as param.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -131,39 +142,105 @@ private static String htmlLabel(String key, long value)
...
+	{
+		final String subTitle = actorLevel > -1 ? "(lvl-" + actorLevel + ")" : "";
+
+		LootTrackerEntry entry = new LootTrackerEntry(eventName, subTitle, System.currentTimeMillis(), items);
+		entries.add(entry);
+
+		if (!groupLoot || !stackEntry(entry))
```
deathbeam: Use list.contains. Also, this logic is very cryptic. 
I would just make the entries map, and do:
```
existingEntry = map.get(e.getname());
if (entry != null)
{
  existingEntry.combineWith(entry);
}

overallGp = map.values().stream().map(Entry::getTotalPrice()).sum();
```

also probably I would do that by using map.compute or something.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -131,39 +142,105 @@ private static String htmlLabel(String key, long value)
...
+	void rebuildLogs(boolean groupLoot)
+	{
+		logsContainer.removeAll();
+		logs.clear();
+
+		overallGp = 0;
+		overallKills = 0;
+		updateOverall();
+
+		for (LootTrackerEntry entry : entries)
```
deathbeam: see my previous comment, also this logic looks duplicated

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
psikoi: Because every log contains atleast one entry, the initial one, new ones can be added later

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
psikoi: I was trying to avoid having that sort of stuff in the UI classes

deathbeam: I dont mean of you adding configChanged even to panel. But to make setter and set it on panel in configChanged event from plugin. I dont see how adding parameter to multiple methods for param that is basically global is better.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
psikoi: You sure? It is only used to re-construct the logs when the grouping config changes (from single to grouped or vice versa), not to refresh any view, rebuild sounds more appropriate

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
psikoi: Won't I have to expose the id then? to compare between two boxes

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
psikoi: But rebuild items is only executed when the config changes, not everytime a new item/log is added, so the price would always be oudated until the config changed

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
psikoi: Then single logs would not show combat, even if grouping was disabled.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
deathbeam: Nope, you would compare them based on title. `@EqualsAndHashCode(of='title')` or something like that

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
deathbeam: then just rebuild i guess

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerLog.java
psikoi: Ah, didn't know about that annotation

psikoi: Still, why would I compare two boxes if I don't need to create a second in the first place, I can just search for that ID, and if a box with that Id already exists, add items to it, instead of creating a new box and comparing it, only to then throw it away

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -0,0 +1,45 @@
...
+	String GROUP_KEY = "loottracker";
+
+	@ConfigItem(
+		keyName = "groupLoot",
+		name = "Group loot by NPC/Activity",
+		description = "Groups the loot received from a NPC or Activity into one single log."
+	)
+	default boolean groupLoot()
+	{
+		return true;
```
Adam-: This should probably default to the existing behavior?

deathbeam: The group layout is more useful than per-kill layout in almost every scenario, I think it should be default, and a lot of people want just that.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -199,11 +234,19 @@ private void buildBox(LootTrackerRecord record)
...
-		final JPopupMenu popupMenu = new JPopupMenu();
-		popupMenu.setBorder(new EmptyBorder(5, 5, 5, 5));
 		popupMenu.add(reset);
+
+		if (groupLoot)
+		{
+			// Create details menu
+			final JMenuItem details = new JMenuItem("View details");
+			details.addActionListener(e ->
+			{
```
deathbeam: remove the unnecessary { } around and make this 1 line

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -199,11 +234,19 @@ private void buildBox(LootTrackerRecord record)
...
+				showDetailedView(box);
```
deathbeam: Just set the detailedKey here, instead of exposing the ID from box (here you already have access to ID)

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -50,6 +50,8 @@
 	private final JLabel priceLabel = new JLabel();
 	private final JLabel subTitleLabel = new JLabel();
 	private final ItemManager itemManager;
+
+	@Getter
```
deathbeam: Revert this

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -56,6 +56,7 @@
 
 	// Handle loot boxes
 	private final JPanel logsContainer = new JPanel();
+	private final JPanel layoutPanel = new JPanel();
```
deathbeam: Move this back to constructor? It is unused from outside of it

Nightfirecat:
It's a little jarring to have the scrollbar appear and disappear when switching from a view that doesn't take all the vertical space to one which does. Could we have a scrollbar permanently there, and disabled/taking the full height when it's not needed, or some other way to avoid that little jump when switching between groupings?

Also, the grouping buttons don't have tooltips so there's no way to tell exactly what they represent.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -143,8 +169,13 @@ void add(final String eventName, final int actorLevel, LootTrackerItem[] items)
 	{
 		final String subTitle = actorLevel > -1 ? "(lvl-" + actorLevel + ")" : "";
 		final LootTrackerRecord record = new LootTrackerRecord(eventName, subTitle, items);
+
```
deathbeam: Any reason for this empty line?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -189,6 +220,10 @@ private void buildBox(LootTrackerRecord record)
 		final LootTrackerBox box = new LootTrackerBox(itemManager, record.getTitle(), record.getSubTitle());
 		box.combine(record);
 
+		// Create popup menu
+		final JPopupMenu popupMenu = new JPopupMenu();
```
deathbeam: Why was this moved here?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -36,10 +36,19 @@
 	@ConfigItem(
 		keyName = "groupLoot",
 		name = "Group loot by NPC/Activity",
-		description = "Groups the loot received from a NPC or Activity into one single log."
+		description = "Groups the loot received from a NPC or Activity into one single log.",
+		hidden = true
 	)
 	default boolean groupLoot()
```
deathbeam: change this to get, and description for the field is unnecessary anymore

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -64,18 +76,39 @@
 	private final JLabel overallGpLabel = new JLabel();
 	private final JLabel overallIcon = new JLabel();
 
+	private final JPanel actionsContainer = new JPanel();
```
deathbeam: Are you sure all those need to be instance fields? And not created only where they are used?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerRecord.java
```diff
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.loottracker;
 
+import java.util.ArrayList;
```
deathbeam: Unused import

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -257,6 +358,15 @@ private void buildBox(LootTrackerRecord record)
 		updateOverall();
 	}
 
+	void onGroupingModeChanged(boolean group)
```
deathbeam: on? rename this to changeGroupingMode, onX/Y would mean this is some sort of event, but it is not

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -281,11 +282,12 @@ void add(final String eventName, final int actorLevel, LootTrackerItem[] items)
 	/**
 	 * Rebuilds all the boxes from scratch using existing listed records, depending on the grouping mode.
 	 */
-	void rebuild()
+	private void rebuild()
 	{
 		logsContainer.removeAll();
 		boxes.clear();
-		records.stream().filter(r -> detailedKey == null || r.getTitle().equals(detailedKey)).forEach(this::buildBox);
+
```
deathbeam: any reason for the new line

deathbeam:
Added tooltips

PR #4936, opened 2018-08-16
worldhopper: fix spam in chatbox when trying to open switcher in bank

devLotto:
Makes it retry 3 times before it gives up.


devLotto:
To expand on this, it seems like it sends that "please finish what you're doing before using the world switcher" message if widget 164.13 has children.

devLotto:
Updated to stop trying to open the switcher if the "please finish what you're doing before using the world switcher" message is received.

PR #4943, opened 2018-08-16
WASD camera not consume CTRL keypress

JorgenRe1:
CTRL click = run to target even if run is turned off.


deathbeam:
Looks like this was already solved

PR #4952, opened 2018-08-16
wasd plugin: check if chatbox input is hidden instead of possible open widgets

Kamielvf:
…n widgets


deathbeam:
Looks like this has been already implemented

Issue #4896, opened 2018-08-14
Change farming timer to AM/PM


shogunreaper:
Is this a thing? I looked through settings but i didn't see anything like it.

if not can an option to change from 24 hour to am/pm time be added?

PR #4959, opened 2018-08-17
12 hour clock for timers

GandalfTheGoat:
for #4896 
timers


Nightfirecat:
If possible, I'd like to see this just read from the user profile's setting of time format (eg. the time format their computer uses) instead of being a config.

Abextm:
When I wrote the plugin originally I was going to make it do system config, but there isn't a system independent way of doing so, so on windows you have to read the registry, which I don't really want to do.

PR #4980, opened 2018-08-17
Update CrypticClue.java

merlindima:
Updated the Ulizius clue to have fairy ring CKS in help box


deathbeam:
Implemented in #4976

PR #4991, opened 2018-08-18
WIP - Add support for remapping numbers to FKeys in WASD mode plugin

deathbeam:
E.g 1 -> F1, 0 -> F10

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraListener.java
@@ -95,11 +95,26 @@ class WASDCameraListener extends MouseListener implements KeyListener
 					case KeyEvent.VK_SLASH:
 						// refocus chatbox
 						plugin.setTyping(true);
-						clientThread.invoke(() ->
-						{
-							plugin.unlockChat();
-						});
+						clientThread.invoke(() -> plugin.unlockChat());
 						break;
+					case KeyEvent.VK_1:
+					case KeyEvent.VK_2:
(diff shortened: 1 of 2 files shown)


Adam-:
Why is this useful? Also this looks like maps numbers to f-keys and not the reverse?

Unmoon:
Many laptop have have F-keys behind function key.

deathbeam:
Ah yeah, the title is wrong, yeah. so e.g you can use numbers instead of FKeys in WASD mode. I saw some people request it in past and the numbers are closer to home row so it is more convenient. Also, on Macs, positioning of FKeys is abyssmal

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraListener.java
```diff
@@ -95,11 +95,26 @@ else if (config.right().matches(e))
...
+					case KeyEvent.VK_6:
+					case KeyEvent.VK_7:
+					case KeyEvent.VK_8:
+					case KeyEvent.VK_9:
+					case KeyEvent.VK_0:
+						if (config.numbers())
+						{
+							final int newKeyCode = e.getKeyCode() == KeyEvent.VK_0
+								? KeyEvent.VK_F10
+								: e.getKeyCode() + 0x40 - 1; // This is the difference between num key
```
Nightfirecat: Shouldn't this be `e.getKeyCode() + 0x3f`? Also, this could probably be defined as a constant:

```
private static final int NUMBERS_F_KEY_DIFFERENCE = KeyEvent.VK_F1 - KeyEvent.VK_1;
```

PR #5000, opened 2018-08-18
add support for esc and Fx hotkeys, rename from wasdcamera

Groctel:
Added support to the wasd camera plugin to redistribute the top Fx keys and Esc shortcuts to the whole keyboard. Since it has more options, I renamed it to "Key Redistribution".

A message for the future: This is my first commit and @melchor629 guided me through the whole process. This is my thank you note. IOU a meal!


deathbeam:
But there is already #4991 I made earlier. I dont see how also making to what the fkeys map to is useful.

TaxoRubio:
I saw #4991 after the pr :s

I would love to be able to map the camera to WASD and the most used fkeys to QERF123. It would add a more wheel-like style to the tab switch and would require less space. It would also let anyone configure the tabs emulating other control schemes they're comfortable with.

TaxoRubio:
Fixed!

PR #5003, opened 2018-08-18
Fix KBD respawn timer

MagicfTail:
Timed using NPC tagger


deathbeam:
But wiki says its ten 🤔

MagicfTail:
Then the wiki is wrong since the npc tagger is supposed to be tick perfect 🤷‍♂️

ShaunDreclin:
You sure it said 9 and not 9.9?

MagicfTail:
Yes, because i made it log the time and npc when starting a new timer. I should probably pr that since it's really useful

ShaunDreclin:
Oh, yea that sounds useful ;o I'll go update the wiki

MagicfTail:
https://i.imgur.com/fyfIYBw.png

ShaunDreclin:
>https://i.imgur.com/fyfIYBw.png

dountdown

deathbeam:
Alright now wiki says 9 so i guess its ok :p

PR #5014, opened 2018-08-19
TimeTracking: Add support for intuitive time notations like "1h 30m 10s"

### LINKED ISSUE #4921: Timers plugin should accept "0h", "0m", and "0h00m" style lengths for timers and stopwatches
When creating timers and/or stopwatches, you must specify a colon-separated length of time: "hh:mm:ss". This format is lengthy and is not easy to type (pressing and releasing shift to enter `:` between numbers)

It would be more convenient and intuitive to input something like "2h" instead of "2:00:00", or "15m" instead of "15:00". This type of time specification is used across numerous sites (Youtube and Twitch for video timestamps, for instance) and is much easier to type.

jamyc:
Fixes #4921 

- Tries to parse the input using the ISO-8601 Durations format if it contains either `h`, `m` or `s`, else it defaults to the `hh:mm:ss` logic
- Supports notations like:
  - 1h 30m 20s
  - 1h30m20s
  - 3000s
- Does **not** support notations like:
  - 20s 20m

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockPanel.java
@@ -32,6 +32,8 @@ import java.awt.event.FocusEvent;
 import java.awt.event.FocusListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.time.Duration;
+import java.time.format.DateTimeParseException;
 import javax.swing.BorderFactory;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This seems like another opportunity to write some tests. Functionality looks good though! :+1:

deathbeam:
Do you want to work on the tests @jamyc? I am fine merging this either way

jamyc:
I'm afraid I won't be able to find time to do this in the upcoming days. I could submit a separate PR later this week with some tests.

deathbeam:
Sounds good.

PR #5015, opened 2018-08-19
Implement Refocus plugin. Closes #5011

### LINKED ISSUE #5011: Remember mouse position on alt-tab
I am willing to implement this. 

I play two accounts at once (ironman btw) and moving my mouse between windows is really tiresome, to the point where I feel fatigue in my hands. I would love to be able to alt-tab between two RuneLite windows and have the cursor position either remembered in each window separately or to recenter it over the viewport.

Before I implement this, I wanted to get some other opinions. I desperately want this functionality to save my hands. I think what I've described is pretty limited, but moving the cursor automatically (even to maintain the illusion that it never moved) is edging towards cheating.

JustAPerson:
This plugin saves mouse position when the window focus is lost while the
mouse is over the viewport. It will restore this position when focus is
regained.

This is intended to relieve the strain caused by constantly moving the
mouse around while playing more than one RuneLite client simultaneously
(which is permissible under RuneScape rules).

See #5011 

Feed back desired on calculating the correct viewport absolute position. An alternative is to grab some lock and climb the entire awt.Component heirarchy. Also please advise about awt.Robot. It was the first thing I found for moving the cursor, but it has some other sketchy methods and I wouldn't be surprised if this is the type of thing Jagex tries to detect.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/refocus/RefocusPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/refocus/RefocusPlugin.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2018, JustAPerson <https://github.com/JustAPerson>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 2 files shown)


Adam-:
RuneLite is not a bot, and I do not think it should be moving the mouse for you. Also if I had to guess, this would probably set off the bot detection.

deathbeam:
I dont see how positioning mouse for player is permissible under RuneScape rules.

JustAPerson:
>I dont see how positioning mouse for player is permissible under RuneScape rules.

That was pretty clearly in reference to the "playing multiple accounts simultaneously".

I don't see how this gives an advantage. Jagex has stated in the past that accessibility tools that move/click are fine so long as it's only one action per keypress. There's one keypress and one movement involved. OSBuddy has a MouseKeys plugin that will move the cursor a configurable distance (e.g. to quickly drop before shift-drop).

deathbeam:
I dont see how moving your mouse to X/Y position is not botting. Also, OSB mousekeys are not even allowed, they specifically stated that you can only use your OS mouse keys, and none of them allow custom distances.

JustAPerson:
This doesn't automate the game, it automates moving between windows. From the perspective of a window that loses focus, the mouse hasn't moved at all. It wont receive mouse events until the mouse is back within the window bound, and when it returns, it is in the exact same position it last was.

>they specifically stated that you can only use your OS mouse keys, and none of them allow custom distances.

[The RS3 Wiki shows how pretty configurable most OS implementations are.](http://runescape.wikia.com/wiki/Mouse_keys#Dropping_items) Xfce looks particularly ez to calibrate to exact pixel offsets.

deathbeam:
1. Your mouse enters window at X/Y
2. Then plugin will move your mouse at different X/Y

ShaunDreclin:
>tools that move/click are fine so long as it's only one action per keypress

Iirc that was never actually their policy, it was just repeated so many times in the community that people took it as fact. Regardless though, the current policy is that anything beyond rebinding a button to a different button is not allowed. Moving your mouse cursor to a specific x/y coordinate would definitely not be kosher. ([Source](http://services.runescape.com/m=news/mouse-keys---changes--clarification?oldschool=1))

deathbeam:
See #5011

PR #5016, opened 2018-08-19
Omit empty strings in NPC indicators

### LINKED ISSUE #5012: NPC Indicators config has an issue with trailing commas
If your list of NPCs ends with a comma, Dairy cows get a small triangle under them highlighted.

screenshot

deathbeam:
Do no try to parse the list to strings including empty names.

Closes #5012


Nightfirecat:
Seems like this would be good to write a few tests for.

Nightfirecat:
Seems like we should add a test for empty strings then. It currently doesn't check that.

deathbeam:
I will add some tests for NPC indicators later today probably then

deathbeam:
Added the test

PR #5019, opened 2018-08-20
Add 2007 interface style with modified compass

ShaunDreclin:
For people (like me) who keep getting east and west mixed up. 🙄

screenshot


kongoslice:
This would be a great addition to Runelite

Adam-:
I don't like that this adds an entire new interface style called "2007" that only just changes the compass to one which has never actually existed at any point. It seems very arbitrary. Maybe just a checkbox to use the 2005/6 compass sprite?

ShaunDreclin:
Yeah I wasn't sure the best way of putting this in, a checkbox makes more sense. Does it have to use the 2005 sprite though? That one kinda clashes with the other two styles IMO

Popedigoz:
Ain't this a thing already?

PR #5021, opened 2018-08-20
Add XpTracker toggle for tracking post-99 XP

evansloan:
Add a config option to opt out of tracking XP for stats that have
reached level 99.

[Preview](https://streamable.com/t5t7z)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
@@ -52,4 +52,15 @@ public interface XpTrackerConfig extends Config
 	{
 		return 0;
 	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "trackSkillAfter99",
+		name = "Track skill after level 99",
+		description = "Configures whether or not skills will be tracked after level 99"
+	)
+	default boolean trackSkillAfter99()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -52,4 +52,15 @@ default int pauseSkillAfter()
...
 		return 0;
 	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "trackSkillAfter99",
+		name = "Track skill after level 99",
+		description = "Configures whether or not skills will be tracked after level 99"
+	)
+	default boolean trackSkillAfter99()
```
deathbeam: I would inverse this, e.g rename to Do not tracker after 99, and make it disabled by default

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -52,4 +52,15 @@ default int pauseSkillAfter()
...
 		return 0;
 	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "doNotTrackSkillAfter99",
+		name = "Do not track skill after level 99",
+		description = "Configures whether or not skills will be tracked after level 99"
+	)
+	default boolean doNotTrackSkillAfter99()
```
psikoi: I think trackAfter99 would be simpler

raiyni: I agree with this.

evansloan: Yes, I'll change it in a bit.

deathbeam:
Implemented in #5741

PR #5022, opened 2018-08-20
Fixed OSB display bug

Asymons:
There was a bug where if you clicked on an item in the GE, pressed back (or bought) and went back to the item screen the OSB price would not display.

Reproducible Steps:
1. Open GE
2. Look at a Zulrah Scale
3. Press back
4. Look at a Zulrah Scale (notice no OSB)


Adam-:
Fixed in 8c3a3cdaa2e03c8f256fff24ccf03f595eab2e0c

PR #5025, opened 2018-08-20
Added GE Buy and Sell history on items

Asymons:
Adding recording of prices when bought/sold and have their recorded timestamp.


runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
```diff
@@ -63,4 +63,33 @@ default boolean enableOsbPrices()
...
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default String geBuySellData()
+	{
+		return "{}";
+	}
+
+	@ConfigItem(
```
ollieatkinson: is using `ConfigItem` like this the preferred way of persisting key-value data in runelite?

Asymons: I'm not sure what the preferred way is, but this code is based off how NotesConfig.java is persisted.
Is there a better way?

ollieatkinson: Probably a question for one of the core maintainers - @deathbeam are you able to chime in?

deathbeam: For simple data yes, but this is probably misuse of the config system as it will contain A LOT of data eventually

ollieatkinson: That was my concern, is that this is essentially using configuration to store a large data set. Should we look at writing this sort of information to the `.runelite` directory? It would mean the information is not persisted across machines but I don't see that being too much of an issue.

deathbeam: We had similar discussion in regards of storing loot tracker data and the outcome from @Adam- was that it should be persisted only remotely and it would require new table in the remote RuneLite DB

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -318,6 +350,52 @@ public void onGameTick(GameTick event)
...
+			{
+				build.append(StackFormatter.formatNumber(geBuySell.getBought()));
+			}
+			else
+			{
+				build.append(unknownText);
+			}
+
+			// Handles sold price lookup
+			build.append(soldText);
```
ollieatkinson: This would read better using commas to separate the "Bought", "Sold" and "Last Trade" items.

```
Bought: 100gp Sold: 10:00:12 Last Trade: 09:53:43
```

-->

```
Bought: 100gp, Sold: 10:00:12, Last Trade: 09:53:43
```

Asymons: The commas do help separate it well. Thanks for the suggestion!

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -193,6 +207,24 @@ public void onGrandExchangeOfferChanged(GrandExchangeOfferChanged offerEvent)
 		boolean shouldStack = offerItem.isStackable() || offer.getTotalQuantity() > 1;
 		BufferedImage itemImage = itemManager.getImage(offer.getItemId(), offer.getTotalQuantity(), shouldStack);
 		SwingUtilities.invokeLater(() -> panel.getOffersPanel().updateOffer(offerItem, itemImage, offerEvent.getOffer(), offerEvent.getSlot()));
+		if (!geBuySellHistory.containsKey(offer.getItemId()))
+		{
+			geBuySellHistory.put(offer.getItemId(), new GrandExchangePriceHistory(0, 0, "Unknown"));
```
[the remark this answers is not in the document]
Asymons: You're thinking we should not show the line if both buy/sell is not recorded?
We'll need some sort of indicator for when one over the other is recorded or we can just simply remove anything that doesn't have required info.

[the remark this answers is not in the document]
ollieatkinson: What are the scenarios we would need to handle?

#### Sold is `Unknown`
```
Bought: 100gp, Last Trade: 09:53:43
```

#### Bought is `Unknown`
```
Sold: 102gp, Last Trade: 09:53:43
```

#### Both are available
```
Bought: 100gp, Sold: 102gp, Last Trade: 09:53:43
```

or none are available - in this instance the row could be hidden.

--- 

Forgive me if I have got the wrong end of the stick, but is the above along the right lines of how you intended it to work?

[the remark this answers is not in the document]
Asymons: You're exactly right here.
I believe I made it easy to refactor this behavior into what we currently have too.

PR #5026, opened 2018-08-20
Fix hover inconsistency in screenmarker plugin

MagicfTail:
Thanks to FluffyRabbit for the image


MagicfTail:
https://i.imgur.com/y3VyBhU.gifv

Nightfirecat:
Isn't this new icon just a darker version of the existing icon? Use `ImageUtil.grayscaleOffset` for that instead of adding a new file.

MagicfTail:
I assumed there was a reason that would work since fill_colour_icon is going the same thing, i'll look into it if it there isn't.

deathbeam:
Any update on this?

MagicfTail:
You can maybe get the exact same colour if you tinker a lot with it 🤷‍♀️ , but I couldn't get it the exact same

PR #5030, opened 2018-08-20
Added boss skill screenshots

### LINKED ISSUE #5010: Screenshot boss kills
I think it would be useful for an auto-screenshot on boss kill feature to be added to the screenshot plugin. Just like how it automatically screenshots on PvP kills but for bosses.

ItsSebas:
Takes a screenshot if boss is killed (default disabled)
image
image

Closes #5010

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -120,11 +120,23 @@ public interface ScreenshotConfig extends Config
 		return false;
 	}
 
+
 	@ConfigItem(
-			keyName = "boss",
-			name = "Screenshot Boss Kills",
-			description = "Configures whether or not screenshots are automatically taken of boss kills",
-			position = 8
+		keyName = "kills",
+		name = "Screenshot PvP Kills",
+		description = "Configures whether or not screenshots are automatically taken of PvP kills",
+		position = 8
+	)
+	default boolean screenshotKills()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "boss",
+		name = "Screenshot Boss Kills",
+		description = "Configures whether or not screenshots are automatically taken of boss kills",
+		position = 9
 	)
 	default boolean screenshotBossKills()
 	{
@@ -132,22 +144,10 @@ public interface ScreenshotConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "kills",
-		name = "Screenshot PvP Kills",
-		description = "Configures whether or not screenshots are automatically taken of PvP kills",
-		position = 9
-	)
-	default boolean screenshotKills()
-	{
-		return false;
-	}
-
-
-	@ConfigItem(
-			keyName = "hotkey",
-			name = "Screenshot hotkey",
-			description = "When you press this key a screenshot will be taken",
-			position = 10
+		keyName = "hotkey",
+		name = "Screenshot hotkey",
+		description = "When you press this key a screenshot will be taken",
+		position = 10
 	)
 	default Keybind hotkey()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -121,14 +121,14 @@ public interface ScreenshotConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "hotkey",
-		name = "Screenshot hotkey",
-		description = "When you press this key a screenshot will be taken",
-		position = 10
+			keyName = "boss",
+			name = "Screenshot Boss Kills",
+			description = "Configures whether or not screenshots are automatically taken of boss kills",
+			position = 8
 	)
-	default Keybind hotkey()
+	default boolean screenshotBossKills()
 	{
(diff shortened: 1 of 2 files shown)


Kamielvf:
I think you should change it so that the file name is `<Boss name> kill <kill count>`. You can probably copy the regex from the killcount stuff in the chat commands plugin.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
```diff
@@ -121,14 +121,14 @@ default boolean uploadScreenshot()
 	}
 
 	@ConfigItem(
-		keyName = "hotkey",
-		name = "Screenshot hotkey",
-		description = "When you press this key a screenshot will be taken",
-		position = 10
+			keyName = "boss",
```
Nightfirecat: Don't rearrange this just to insert boss kills; it is probably fine to be after the PvP kills entry.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -344,6 +344,12 @@ public void onChatMessage(ChatMessage event)
 			String fileName = "Kill " + format(new Date());
 			takeScreenshot(fileName);
 		}
+
+		if (config.screenshotBossKills() && chatMessage.contains("kill count is:"))
```
Nightfirecat: This will not capture all activities with kill counts. [See the chat commands plugin.](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java#L85-L88) Moving those to a common location for shared use among plugins would probably be a good way to approach that.

MagicfTail: We do imo not want to screenshot raids kc since we are already taking screenshots when the loot is received, which i assume is the reason people would want to screenshot boss kc.

deathbeam: And screenshotting e.g wintertodt kc is useless

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
deathbeam: Also wrong indentation. See code guidelines wiki page

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
Nightfirecat: That's fair, though we should still be [reusing the boss kill count regex](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java#L85-L88).

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -474,6 +480,20 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
 		return skillName + "(" + skillLevel + ")";
 	}
 
+	String parseBossKill(String killMessage)
```
deathbeam: Make this method private

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -474,6 +480,20 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
+	{
+
```
deathbeam: Remove this empty line

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -344,6 +344,12 @@ public void onChatMessage(ChatMessage event)
 			String fileName = "Kill " + format(new Date());
 			takeScreenshot(fileName);
 		}
+
+		if (config.screenshotBossKills() && chatMessage.startsWith("Your") && chatMessage.contains("kill count is"))
```
Nightfirecat: This should use `BOSSKILL_MESSAGE_PATTERN` instead of `chatMessate.startsWith` and `chatMessage.contains`.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -474,6 +481,25 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
 
+	/**
+	 * Receives a String and parses it into a shortened string for filename usage.
+	 * @param killMessage killMessage String received after killing a boss,
+	 *                     with the format "Your (Boss Name) kill count is: (killcount)."
+	 * @return Shortened string in the format "(Boss Name) kill (killcount)"
+	 */
+	private String parseBossKill(String killMessage)
+	{
+		Matcher m = BOSSKILL_MESSAGE_PATTERN.matcher(killMessage);
```
deathbeam: You should reuse this created matcher (what you created earlier to check if message matches_

PR #5041, opened 2018-08-21
Kourend library customer book overlay

dekvall:
Adds a panel in the top left corner for showing the book requested by the customer. The text will turn green when the book is in the player's inventory indicating you can return to the customer. Since you can only have one of each book in your inventory, the bookcase in which the book is located will also turn red when acquired.
This also adds the help option of the library customers to the menuentryswapper. 
ininvy
notininvy

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -235,8 +235,13 @@ public interface MenuEntrySwapperConfig extends Config
 		name = "Decant",
 		description = "Swap Talk-to with Decant for Bob Barter and Murky Matt at the Grand Exchange."
 	)
-	default boolean swapDecant()
-	{
-		return false;
-	}
+	default boolean swapDecant() { return false; }
+
+	@ConfigItem(
+	    position = 18,
+		keyName = "swapLibrary",
+		name = "Help",
+		description = "Swap Help on the library customers"
+	)
+	default boolean swapLibrary() { return true; }
 }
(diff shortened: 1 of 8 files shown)


forsco:
Doesn't the side bar plugin already turn green when you have the requested book?

deathbeam:
It does. I guess this is to not require sidebar being open?

forsco:
I guess so long as the overlay is a toggle it should be fine? I never do the books without the sidebar open anyway.

dekvall:
Yes precisely, this allows the player to do library without the sidebar open. 
I simply forgot to add a toggle, will do.

In general i have found this overlay easier to use than the sidepanel. You can still batch turn-in by collecting books along the way to the requested book but you don't have to find the requested book in the sidepanel list.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -235,8 +235,13 @@ default boolean swapAssignment()
 		name = "Decant",
 		description = "Swap Talk-to with Decant for Bob Barter and Murky Matt at the Grand Exchange."
 	)
-	default boolean swapDecant()
-	{
-		return false;
-	}
+	default boolean swapDecant() { return false; }
```
forsco: Checkstyle:

```
default boolean swapDecant()
	{
		return false;
	}
```
Let IntelliJ do the collapse.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -235,8 +235,13 @@ default boolean swapAssignment()
...
-	}
+	default boolean swapDecant() { return false; }
+
+	@ConfigItem(
+	    position = 18,
+		keyName = "swapLibrary",
+		name = "Help",
+		description = "Swap Help on the library customers"
+	)
+	default boolean swapLibrary() { return true; }
```
forsco: As above.

PR #5044, opened 2018-08-21
[WIP] pest control: overlay for boat and ingame 

raiyni:
Note: still wip, needs wrapping and proper y placement with different image/text sizes and unused methods need removed

This PR introduces a new PanelComponent child type: LineImageComponent
Allows drawing images and text on the same line inside of a panel

A debug screenshot showing multiple lines with mixed text and images

PC ingame overlay. I couldn't find an image ID for the ingame PC icons (they use modelids)


runelite-client/src/main/java/net/runelite/client/plugins/pestcontrol/PestControlInfoOverlay.java
```diff
@@ -0,0 +1,164 @@
...
+	@Inject
+	public PestControlInfoOverlay(QueryRunner queryRunner, Client client)
+	{
+		setPosition(OverlayPosition.TOP_LEFT);
+		setLayer(OverlayLayer.UNDER_WIDGETS);
+		setPriority(OverlayPriority.LOW);
+		this.queryRunner = queryRunner;
+		this.client = client;
+
+		this.hpIcon = ImageUtil.getResourceStreamFromClass(PestControlPlugin.class, "knight_hp_icon.png");
```
Nightfirecat: Both of these icons should be available through `SpriteManager.getSprite`.

raiyni: They don't have spriteid's though. They have modelid's.

Nightfirecat: That's fine then... In the future, we may want to consider adding some way to get an image from a model so we don't have to do this.

psikoi:
What is that first screenshot supposed to be? I can't seem to understand it

raiyni:
The multiple ways it can render, with and would text ignoring the past control information. Probably needs a more clear example and the ability to render the image after the text.

raiyni:
Switched to using a list for left and right side. By default, it auto expands the line. Will add line breaks later and update screenshot.

deathbeam:
Also you can probably achieve what you want to do with already existing code, e.g by using PanelComponent VERTICAL with bunch of PanelComponent HORIZONTAL as children

PR #5112, opened 2018-08-25
Add client focus to top on windows 10

### LINKED ISSUE #2335: Idle notifications doesn't bring client to the front on Windows 10
Windows 10 plays a toast during the idle notification but doesn't switch to the client or bring it to the front of my screen

'Send notifications when focused' and 'Request focus on notification' are both selected under Runelite settings

jkybtw:
Currently on windows 10, requesting focus only makes the taskbar flash, and the game does not pop up to the top. On other OS (MacOS at least), the client comes to the top when requesting focus.

This workaround makes the client pop up to the top on Windows 10.

Fixes #2335

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -512,19 +512,13 @@ public class ClientUI
 		{
 			OSXUtil.requestFocus();
 		}
+		else if (OSType.getOSType() == OSType.Windows && System.getProperty("os.name").contains("10"))
+		{
+			frame.toFront();
+		}
 
-		if (System.getProperty("os.name").equals("Windows 10"))
-		{
-			frame.setAlwaysOnTop(true);
-			frame.requestFocus();
-			giveClientFocus();
-			frame.setAlwaysOnTop(config.gameAlwaysOnTop());
-		}
-		else
-		{
-			frame.requestFocus();
-			giveClientFocus();
-		}
+		frame.requestFocus();
+		giveClientFocus();
 	}
 
 	/**
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -513,7 +513,7 @@ public class ClientUI
 			OSXUtil.requestFocus();
 		}
 
-		if (System.getProperty("os.name").contains("10"))
+		if (System.getProperty("os.name").equals("Windows 10"))
 		{
 			frame.setAlwaysOnTop(true);
 			frame.requestFocus();
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -512,13 +512,16 @@ public class ClientUI
 		{
 			OSXUtil.requestFocus();
 		}
-		else if (OSType.getOSType() == OSType.Windows && System.getProperty("os.name").contains("10"))
+		else if (OSType.getOSType() == OSType.Windows)
 		{
 			frame.toFront();
+			frame.setAlwaysOnTop(true);
+			frame.setAlwaysOnTop(config.gameAlwaysOnTop());
 		}
 
 		frame.requestFocus();
 		giveClientFocus();
+
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -513,8 +513,18 @@ public void requestFocus()
 			OSXUtil.requestFocus();
 		}
 
-		frame.requestFocus();
-		giveClientFocus();
+		if (System.getProperty("os.name").contains("10"))
```
Nightfirecat: Is this really the best way to detect windows 10? Among other things, can't you at least do `OSType.getOSType() == OSType.Windows`?

What's the full string of `System.getProperty("os.name")` anyway? Seems like you should be matching its exact value instead of doing `.contains("10")`...

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -513,8 +513,18 @@ public void requestFocus()
...
+		{
+			frame.setAlwaysOnTop(true);
+			frame.requestFocus();
+			giveClientFocus();
+			frame.setAlwaysOnTop(config.gameAlwaysOnTop());
```
Nightfirecat: Why is `frame.setAlwaysOnTop()` called twice in this block?

jkybtw: The first call brings the client to the front, and then the second call sets it back to whatever the config was set to. 

In the case of true, the client would always be on top anyway, but in the case of false, when the client is hidden behind another application, the client is called to the top, client is given focus and then the client can be hidden away again.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
jkybtw: The full string is "Windows 10", i will fix this.

ProditorMagnus:
It does not come to top with linux mint/cinnamon either, it only receives all keypresses without being visible.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -513,8 +513,18 @@ public void requestFocus()
 			OSXUtil.requestFocus();
 		}
 
-		frame.requestFocus();
-		giveClientFocus();
+		if (System.getProperty("os.name").equals("Windows 10"))
```
deathbeam: We have OSType.getOSType() what is literally 2 lines above and I would prefer for you to use that, also use either switch or elseif

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -513,8 +513,18 @@ public void requestFocus()
...
+		{
+			frame.setAlwaysOnTop(true);
+			frame.requestFocus();
```
deathbeam: You should also add frame.toFront() above and probably call this on EDT

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -512,6 +512,10 @@ public void requestFocus()
 		{
 			OSXUtil.requestFocus();
 		}
+		else if (OSType.getOSType() == OSType.Windows && System.getProperty("os.name").contains("10"))
```
deathbeam: Why only on windows 10?

deathbeam: Also does this still work? I wanted to use .toFront together with the rest of stuff you had before

jkybtw: Ok I will just do a check for all windows versions.

jkybtw:
@DannysPVM Yeah I'm on W10 right now. The PR right now brings the client to the front, but it doesnt focus :l

Adam-:
From googling, maybe try 
https://stackoverflow.com/a/7435722?

danwbyrne:
@deathbeam Any word on this PR? It is still an issue for windows 10 users. I've tested the PR as it is now and it definitely fixes the problem.

If some work needs to be done cleaning it up let me know and I'll do it.

deathbeam:
Well, the PR needs to be updated. It needs to use **setExtendedState AND toFront AND setAlwaysOnTop**

melkypie:
Option to force client to front was added in 7c9e193f80cfca491909ec71bd9796198b807b5e

Issue #5121, opened 2018-08-25
Let us right-click hop to cc mates in chat


JeljoGb:
Let us right-click hop to cc mates, this is faster than going to the cc tab, searching for the person, then right-clicking hop.

deathbeam:
Fixed in 6eb34a752f023c15083edc59922c31a047745562

PR #5125, opened 2018-08-25
Add Hop-To option in chat

Retoxified:
This implements feature-request #5121, the Hop-To option is added before the Report option in chat.


deathbeam:
What if player has the report option disabled?

Retoxified:
That is a fair point. I will see how other plugins handle this and revise the solution.

Retoxified:
notremoved

The report option is not removed from the chat menu when it is toggled off, as seen in the screenshot above.

PR #5126, opened 2018-08-25
Add pause all button to xp tracker

### LINKED ISSUE #5120: Add pause all option to XP tracker
### Scope:
Currently, you need to pauze them one by one, though it would be more efficient (for example when going afk) to pauze everything at once.

### Solution:
Right-click on the total counter and add the option to the dropdown to pauze all tracking.

Rdeneve:
Fixes #5120

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
@@ -63,8 +63,6 @@ class XpPanel extends PluginPanel
 	/* This displays the "No exp gained" text */
 	private final PluginErrorPanel errorPanel = new PluginErrorPanel();
 
-	private boolean paused = false;
-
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
```diff
@@ -89,11 +91,28 @@
 		final JMenuItem reset = new JMenuItem("Reset All");
 		reset.addActionListener(e -> xpTrackerPlugin.resetAndInitState());
 
+		// Create pause all menu
+		final JMenuItem pauseAll = new JMenuItem("Pause All");
+		pauseAll.addActionListener(e ->
+		{
+			xpTrackerPlugin.pauseAllSkills(!paused);
+			if (paused)
```
deathbeam: You could maybe drop the boolean and just check for text of pauseAll

PR #5127, opened 2018-08-25
Add plugin to hide world

Adam-:


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -578,20 +583,23 @@ private void hop(int worldId)
...
 
-		String chatMessage = new ChatMessageBuilder()
-			.append(ChatColorType.NORMAL)
-			.append("Quick-hopping to World ")
-			.append(ChatColorType.HIGHLIGHT)
-			.append(Integer.toString(world.getId()))
-			.append(ChatColorType.NORMAL)
-			.append("..")
-			.build();
+		if (!pluginManager.isPluginEnabled(StreamingPlugin.class))
```
deathbeam: I dont think we should add implicit dependencies on plugins like this. You can just alter the message in streaming plugin, or add config option for not sending this message to this plugin.

PR #5139, opened 2018-08-26
#5068 fix: cant quick hop to w26

### LINKED ISSUE #5068: Using the quick-hop feature from world 35 causes a hop to world 16, rather than to world 26.
**To Reproduce**
1. Go to world 35
2. Use the <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>↑</kbd> quick-hop shortcut
3. See that you will be hopping to w16 rather than 26.

**Expected behavior**
As w26 is not a pvp world, I would not have expected it to be skipped.

Karolcz125:
Fixes #5068

Now it is possible to quick hop into (and out) w26

deepin-screen-recorder_sun-awt-x11-xframepeer_20180826161640

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -471,6 +471,8 @@ public class WorldHopperPlugin extends Plugin
 		currentWorldTypes.remove(WorldType.SKILL_TOTAL);
 		// Allow hopping from a high risk world to a non-high risk world
 		currentWorldTypes.remove(WorldType.PVP_HIGH_RISK);
+		// Allow hopping from a LMS world to a non-LMS world
+		currentWorldTypes.remove(WorldType.LAST_MAN_STANDING);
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -471,6 +471,8 @@ private void hop(boolean previous)
 		currentWorldTypes.remove(WorldType.SKILL_TOTAL);
 		// Allow hopping from a high risk world to a non-high risk world
 		currentWorldTypes.remove(WorldType.PVP_HIGH_RISK);
```
deathbeam: What is this? Not related to this PR but what? why is it in config.quickhopOutOfDanger too?

ollieatkinson: Seems like that this line's a mistake and should be removed. But, I will make an assumption that `config.quickhopOutOfDanger()` is a setting someone can enable to be able to quick-swap to these risky worlds

PR #5154, opened 2018-08-27
Plugin - Item Information Tooltips

BenDol:
Thought I'd submit this PR of something I put together over the weekend. I was finding it frustrating having to google search all the OSRS items during my adventures, so I figured since I couldn't find anything like this (I might have missed it) I'd create one.

**How it works:**
Currently this plugin is using osrsbox.com item database to retrieve all the item information via the REST API asynchronously and applies them to the item tooltip. The item search is standoff so it won't show information for an item thats name isnt an exact match. It might be worth while transitioning this database over to the runelite http service db and relying on that. For now this works fine.

**Demo:** https://youtu.be/_LbPpwzwN1U

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
@@ -34,9 +34,21 @@ public interface OpponentInfoConfig extends Config
 	@ConfigItem(
 		keyName = "lookupOnInteraction",
 		name = "Lookup players on interaction",
-		description = "Display a combat stat comparison panel on player interaction. (follow, trade, challenge, attack, etc.)"
+		description = "Display a combat stat comparison panel on player interaction. (follow, trade, challenge, attack, etc.)",
+		position = 0
 	)
 	default boolean lookupOnInteraction()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "showPercent",
+		name = "Show percent",
+		description = "Shows hitpoints as a percentage even if hitpoints are known",
+		position = 1
+	)
+	default boolean showPercent()
 	{
 		return false;
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
@@ -55,6 +55,7 @@ class OpponentInfoOverlay extends Overlay
 
 	private final Client client;
 	private final OpponentInfoPlugin opponentInfoPlugin;
+	private final OpponentInfoConfig opponentInfoConfig;
 	private final HiscoreManager hiscoreManager;
 
 	private final PanelComponent panelComponent = new PanelComponent();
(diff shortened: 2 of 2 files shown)


osrsbox:
Glad to see someone using the item database. Feel free to use the service or grab the data from my repo and use it however you please. FYI, I am in the middle of updating the database contents as a bunch of items are currently missing. Can provide further details if needed, just ask.

BenDol:
Great stuff! I was wondering if items level requirements were going to be added at any point in the future, that would be a useful piece of information. Thanks for the great work!

Adam-:
If it isn't too much data I would just ship it with the client resources. Otherwise I'd make a RL web service for it.  I do not want a dependency in the client on 3rd party web services.

Also we have an existing item stats plugin which this might fit into instead of being its own plugin.

BenDol:
It didn't fit nicely into the existing one, I did that initially.

osrsbox:
@BenDol Great idea about the item level requirements. Will look into how to easily get this info and look at adding it in the future. 
@Adam- I completely agree with not having a third party dependency in the client, but am happy to share the data if you, the RuneLite development team or anyone else wants to use it or integrate it. 

Like I mentioned before, I think **the item database really needs an update before any serious use**, as it has not been updated since 25/12/2017 and items have been released since them. I am also worried about the accuracy of some items too. To combat this, I am rewriting my original tools, and hope to have it up-to-date in the next week or so.

Just a FYI regarding the size, the minified JSON is approximately 5MB, but this includes a lot of item metadata that is already available from the RuneLite client through the `ItemComposition` class (e.g., stackable, tradeable). I am not sure on size restrictions, but removing duplicate properties seems like good practice. Maybe limiting the contents to only an item ID and the item bonuses (stats) array would be a good option, and would be a significant size reduction (guessing >60%). Also, it should be merged into one file, as they are currently individual JSON files for the API (but that should be easy).

BenDol:
@Adam- Any idea when this can be reviewed? Its stable and the osrsbox API is just like using the osbuddy API. We could have it disabled by default and people can turn it on initially? Would be good to get this into the client release though. I use it a lot and it's super helpful.

raiyni:
Do we really want to depend on the osbox API?

BenDol:
@raiyni well we don't **need** to, people can just turn the plugin off. Also in the future when we can host the API ourselves it would literally be like 2 lines of code to change.

raiyni:
I didn't ask if we need to. But do we want to. This has been discussed in other channels and the general consensus was that we didn't want to depend on the osbox API fetching stats. A better solution would to be just serve our own data. Adam even said that exact thing in this thread.

BenDol:
@raiyni yes, using the osrs data if I understood him correctly. My point is that "depending" on osrsbox service for this plugin initially isn't a big deal with the intention of changing it later. There's nothing wrong with that API and we could ship the plugin initially as "opt in" so people have to enable it to use it?

If we are going the route of creating a web service/database I don't have the time to develop it so someone else would have to take that on.

raiyni:
Should do it right from the start. So we aren't stuck in a bind if we **need** to migrate from it because of an issue--be it throttling, site goes down, or laziness.

BenDol:
Recreated in a new branch here: https://github.com/runelite/runelite/pull/6261

PR #5151, opened 2018-08-27
Add player death check, ToB region check and 1 loot per-tile per-tick check to loot tracker events

TheStonedTurtle:
Only triggering one event per location per tick should fix #4839
Player Death check Should fix #4653
Checking for onion in ToB inventory fixes #4695

PR #5155, opened 2018-08-27
Instance correct WorldPoint from LocalPoint

TheStonedTurtle:
Title.
Creates `WorldPoint.fromLocalInstance(Client, LocalPoint)` to accomplish this.


TheStonedTurtle:
#5151 will add these changes

PR #5157, opened 2018-08-27
Add alternate bank tag search string '#'

ShaunDreclin:


runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -181,11 +183,16 @@ public void onScriptEvent(ScriptCallbackEvent event)
 				}
 
 				boolean tagSearch = searchInput.startsWith(TAG_SEARCH);
+				boolean tagSearchAlt = searchInput.startsWith(TAG_SEARCH_ALT);
```
Nightfirecat: This would probably be better implemented by changing `TAG_SEARCH` into a Pattern (`^(tag:|#)`) and performing the substring operation based on the length of the match. This way we don't introduce these extra variables/if-branches to perform the exact same behavior again.

deathbeam: I would rather do one or another, but we had this discussion before, I dont think we came to any reasonable conclusion

ShaunDreclin: # is easier/faster to type, I'd be in favour of that if we only have one

ShaunDreclin: changed it to use regex

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -180,16 +180,9 @@ public void onScriptEvent(ScriptCallbackEvent event)
...
-				String search;
-				if (tagSearch)
-				{
-					search = searchInput.substring(TAG_SEARCH.length()).trim();
-				}
-				else
-				{
-					search = searchInput;
-				}
+
```
Nightfirecat: Remove this extra line.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -59,13 +59,13 @@
...
-		"(To show only tagged items, start your search with 'tag:')";
+		"(To show only tagged items, start your search with 'tag:' or '#')";
 
 	private static final String SEARCH_BANK_INPUT_TEXT_FOUND =
 		"Show items whose names or tags contain the following text: (%d found)<br>" +
-			"(To show only tagged items, start your search with 'tag:')";
+			"(To show only tagged items, start your search with 'tag:' or '#')";
 
-	private static final String TAG_SEARCH = "tag:";
+	private static final String TAG_SEARCH = "^(?:tag:|#)(.+)";
```
Nightfirecat: Typically using a pre-compiled `Pattern` is preferred to using a regex `String`. Can you change this to use a `Pattern` instead?

ShaunDreclin: How do you do string replacement with a `Pattern`? `String.matches()` and `String.replaceAll()` both expect regexes as strings

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
ShaunDreclin: this part of the code is gone

ShaunDreclin: done

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -180,16 +180,8 @@ public void onScriptEvent(ScriptCallbackEvent event)
...
-				if (tagSearch)
-				{
-					search = searchInput.substring(TAG_SEARCH.length()).trim();
-				}
-				else
-				{
-					search = searchInput;
-				}
+				boolean tagSearch = searchInput.matches(TAG_SEARCH);
+				String search = tagSearch ? searchInput.replaceAll(TAG_SEARCH, "$1").trim() : searchInput;
```
raiyni: I don't think this should be a ternary for clarity's sake.

If anything, the search section should be moved up to here.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
raiyni: You get the group, not string replace.

ShaunDreclin: done

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
ShaunDreclin: done

deathbeam:
Maybe this is not really necessary as you can also search for tags even without typing tag:

ShaunDreclin:
Depending on the name of the tag that can give you a bunch of unrelated items too, making it less useful

PR #5168, opened 2018-08-28
defaultworld: store the last world you were connected to

BenDol:
store the last world you were connected to and apply on startup.


deathbeam:
This should probably be part of default world plugin, it is pretty unrelated to world hopper.

runelite-client/src/main/java/net/runelite/client/plugins/defaultworld/DefaultWorldPlugin.java
```diff
@@ -54,6 +54,9 @@
 	@Inject
 	private DefaultWorldConfig config;
 
+	@Inject
+	private ConfigManager configManager;
```
deathbeam: Just use hidden config property, like slayer plugin does

BenDol:
ping

Adam-:
This seems fine. I would probably name it `Use Last World` instead of `Use Last Connected World` since the phrasing there would probably confuse most people.

PR #5169, opened 2018-08-28
Added config sliders.

FusionLord:
Adds sliders as a config option for `int` types.
Simply add a `range` parameter to the `ConfigItem` annotation.
```Java
@ConfigItem(
	keyName = "sampleslider",
	name = "Sample Slider",
	description = "Just a Sample",
	range = { 17, 35 }
)
default int sampleSlider()
{
	return 35;
}
```


deathbeam:
As we do not need them atm for anything there was no reason for this PR to be merged. Maybe you can make use of it for your PR?

raiyni:
Is this satisfied by the `@Range` parameter?

deathbeam:
Yes

PR #5170, opened 2018-08-28
Fixes BackgroundComponent's border to match background alpha.

FusionLord:
All in the title.


deathbeam:
Yes, it then do not matches with in-game xp tracker interface.

PR #5182, opened 2018-08-28
Reduce default plugin noise for regular users

deathbeam:
### Disable blast furnace clickboxes by default

A lot of people visit blast furnace without intention to actually do the
activity (e.g to buy ores) so simply disable the clickbox noise by
default and leave to user if they want them.

### Make KourendLibrary overlay show only on interact

Make KourendLibrary plugin overlay show only on interaction with
bookcase or with customer after login, as by default the noise caused by
the overlays is horrendous.

### Add common hidden items to GroundItemsPlugin

To reduce default plugin noise add some common hidden items to
GroundItemsPlugin (basically same default hidden list as OSBuddy has).

### Disable lit/unlit burner indicators by default

Majority of players do not care about these and it is just permanent
noise in POH for them, leave the enabling of this setting to minority.


deathbeam:
Alright remove the logout reset

PR #5186, opened 2018-08-29
Shorten devtool button labels

### LINKED ISSUE #4458: Several dev tool labels are cut off
screenshot

Either larger buttons or adding tooltips to the buttons would be great.

ShaunDreclin:
Resolves #4458

screenshot


deathbeam:
This is probably not very good solution, as instead of dots now you have things like "Cam position", "Tile Loc Tooltip" etc. what is hardly better. Maybe instead adjust the layout a bit so everything fits?

ShaunDreclin:
I dunno, "Loc" and "Cam" seem like pretty clear abbreviations to me. I also tried making each button two lines but it looked bulky and ugly to me

deathbeam:
"Better" solution was implemented in #9224

PR #5193, opened 2018-08-29
Add automatic bounty hunter target lookup

noelkeijzer:
Added option to the hiScore plugin to automatically lookup a bounty hunter target when it gets assigned to you.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
@@ -74,4 +74,15 @@ public interface HiscoreConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 5,
+			keyName = "bountylookup",
+			name = "Bountylookup",
+			description = "Automatically lookup the stats of your bounty hunter target"
+	)
+	default boolean bountylookup()
+	{
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
@@ -31,13 +31,18 @@ import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
 import java.lang.reflect.InvocationTargetException;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.swing.SwingUtilities;
+
+import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
+import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.PlayerMenuOptionClicked;
@@ -83,6 +88,7 @@ public class HiscorePlugin extends Plugin
 
 	private NavigationButton navButton;
 	private HiscorePanel hiscorePanel;
+	private Pattern bountyPattern;
 
 	@Inject
 	private NameAutocompleter autocompleter;
@@ -117,6 +123,8 @@ public class HiscorePlugin extends Plugin
 		{
 			hiscorePanel.addInputKeyListener(autocompleter);
 		}
+
+		bountyPattern = Pattern.compile("<col=ff0000>You've been assigned a target: (.*)</col>");
 	}
 
 	@Override
@@ -219,25 +227,45 @@ public class HiscorePlugin extends Plugin
 	{
 		if (event.getMenuOption().equals(LOOKUP))
 		{
-			executor.execute(() ->
-			{
-				try
-				{
-					SwingUtilities.invokeAndWait(() ->
-					{
-						if (!navButton.isSelected())
-						{
-							navButton.getOnSelect().run();
-						}
-					});
-				}
-				catch (InterruptedException | InvocationTargetException e)
-				{
-					throw new RuntimeException(e);
-				}
+			lookupPlayer(Text.removeTags(event.getMenuTarget()));
+		}
+	}
 
-				hiscorePanel.lookup(Text.removeTags(event.getMenuTarget()));
-			});
+	private void lookupPlayer(String playerName) {
+		executor.execute(() ->
+		{
+			try
+			{
+				SwingUtilities.invokeAndWait(() ->
+				{
+					if (!navButton.isSelected())
+					{
+						navButton.getOnSelect().run();
+					}
+				});
+			}
+			catch (InterruptedException | InvocationTargetException e)
+			{
+				throw new RuntimeException(e);
+			}
+
+			hiscorePanel.lookup(playerName);
+		});
+	}
+
+	@Subscribe
+	public void onChatMessageReceived(ChatMessage event)
+	{
+		if (config.bountylookup() && event.getType().equals(ChatMessageType.SERVER))
+		{
+			String message = event.getMessage();
+			if (message.startsWith("<col=ff0000>You've been assigned a target: "))
+			{
+				Matcher m = bountyPattern.matcher(message);
+				if (m.matches()) {
+					lookupPlayer(m.group(1));
+				}
+			}
 		}
 	}
 }


runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -83,6 +87,7 @@
 
 	private NavigationButton navButton;
 	private HiscorePanel hiscorePanel;
+	private Pattern bountyPattern;
```
Nightfirecat: This should be `private static final`, and defined up here.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -117,6 +122,8 @@ protected void startUp() throws Exception
 		{
 			hiscorePanel.addInputKeyListener(autocompleter);
 		}
+
+		bountyPattern = Pattern.compile("<col=ff0000>You've been assigned a target: (.*)</col>");
```
Nightfirecat: Per previous comment, define this as a static variable.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -219,25 +226,47 @@ public void onLookupMenuClicked(PlayerMenuOptionClicked event)
 	{
 		if (event.getMenuOption().equals(LOOKUP))
 		{
-			executor.execute(() ->
+			lookupPlayer(Text.removeTags(event.getMenuTarget()));
+		}
+	}
+
+	private void lookupPlayer(String playerName)
```
Nightfirecat: Try not to mix subscribers and defined functions. This should be defined below the subscriber methods.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
deathbeam: private static final and initialize it in-place

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -219,25 +226,47 @@ public void onLookupMenuClicked(PlayerMenuOptionClicked event)
...
+			{
+				throw new RuntimeException(e);
+			}
+
+			hiscorePanel.lookup(playerName);
+		});
+	}
+
+	@Subscribe
+	public void onChatMessageReceived(ChatMessage event)
```
deathbeam: onChatMessage

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -219,25 +226,47 @@ public void onLookupMenuClicked(PlayerMenuOptionClicked event)
...
+	{
+		if (config.bountylookup() && event.getType().equals(ChatMessageType.SERVER))
```
deathbeam: flip this if and do early return

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -219,25 +226,47 @@ public void onLookupMenuClicked(PlayerMenuOptionClicked event)
...
+		{
+			String message = event.getMessage();
+			if (message.startsWith("<col=ff0000>You've been assigned a target: "))
```
deathbeam: make this constant at the top

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
deathbeam: move this together with other subscribed methods

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
```diff
@@ -74,4 +74,15 @@ default boolean autocomplete()
...
+
+	@ConfigItem(
+			position = 5,
+			keyName = "bountylookup",
+			name = "Bountylookup",
+			description = "Automatically lookup the stats of your bounty hunter target"
+	)
+	default boolean bountylookup()
+	{
+		return true;
```
deathbeam: default this to false

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
Nightfirecat: Reverse these conditions--check for message type before parsing the text.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
```diff
@@ -74,4 +74,15 @@ default boolean autocomplete()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 5,
```
deathbeam: Fix your indentation to match rest of the file

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
```diff
@@ -74,4 +74,15 @@ default boolean autocomplete()
...
+			keyName = "bountylookup",
+			name = "Bountylookup",
```
deathbeam: spaces in name?

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
Nightfirecat: Is there a reason this can't use `bountyPattern` to check if it matched?

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscoreConfig.java
Nightfirecat: Actually, I disagree with this one. Since most users won't interact with bounty hunter to begin with, I feel it would be better for PvP qol to enable this for those users who do play bounty hunter.

deathbeam: Well this lookups on the chat message. And you can be on bounty hunter world without caring about their targets but they still get them assigned

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
noelkeijzer: Wouldn't it be more efficient to first check if messages need to be parsed at all before checking what kind of chatmessage it is?

Nightfirecat: Doing a regex parse is almost certainly more expensive than an equality check.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
noelkeijzer: No there is not, I changed it to make use of bountyPattern, testing that now to see if I can find any issues with doing it that way.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
noelkeijzer: I'm not sure I understand how reversing the conditions would influence that. I have pushed an updated version of the code, does that fix the problem?

Nightfirecat: [Because of short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation), if the first conditional is false, it would not evaluate the second condition, hence the cheaper and/or more-likely-to-fail operation should be evaluated first.

Nightfirecat: That said, I did misread the original code--looks like it's ordered correctly, and the code looks fine now. I had thought for some reason that the first check was a regex matcher instead of the config check. My bad there. :upside_down_face:

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -201,6 +206,31 @@ else if (BEFORE_OPTIONS.contains(option))
 		}
 	}
 
+	@Subscribe
+	public void onLookupMenuClicked(PlayerMenuOptionClicked event)
```
Nightfirecat: I know you didn't write this code originally, but change the method name to `onPlayerMenuOptionClicked`.

PR #5216, opened 2018-08-30
Fixed issue with locale when the tests were run from a Dutch computer.

CC007:
Failed tests before the commit:
```
Failed tests: 
  StackFormatterTest.quantityToRSDecimalStackSize:49 expected:<21[.]7K> but was:<21[,]7K>
  StackFormatterTest.quantityToStackSize:84 expected:<1[,]000> but was:<1[.]000>
  StackFormatterTest.stackSizeToQuantity:107 expected:<10500> but was:<10>
```
After the commit, the tests succeeded.


deathbeam:
You should probably instead set the default locale from tests, as we already set the locale from RuneLite#main function

CC007:
@deathbeam The locale is set in the test in the setup, but this seems to be ignored by the cases that I fixed.

deathbeam:
Hmm, alright. Maybe we could remove the setting of locale from `main` then as I think we use only few nubmer formats

CC007:
That will not solve the issue. The reason that the locale isnt properly set is because the variables in question are static and therefore initialized statically. For the test suite, this happens before the setUp method is called, which is why the wrong locale is used. I expect that this is also the case when the locale is set in the main method.

deathbeam:
Yeah I mean to keep it as you set it now, but there is no reason to both set the locale from there and also globally from main method, so I would remove the setting from `main`

CC007:
Ah check. I dont know if there are other methods that depend on locales, but at least the tests seem to succeed with the main locale deletion. I'll post the commit in a sec.

CC007:
Ok that should do it.

deathbeam:
Hmm actually looks like we use locale stuff at quite a lot of places so probably the setting of Locale in main should stay

CC007:
I reverted the removal of setting the locale in the main class

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -68,6 +61,14 @@
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
+import javax.annotation.Nullable;
```
deathbeam: Revert these import changes

CC007: Hmm this is something that my IDE changed. Is there a linter or anything available that keeps the code in the style that runelite wants?

deathbeam: https://github.com/runelite/runelite/wiki/Code-Conventions we have link to XML that you can just import to IntelliJ and then it will format the imports correctly

PR #5222, opened 2018-08-30
Fix in combat idle logout notification and add config option for notification

MagicfTail:
The idle logout timer is 20 min instead of 5 min when in combat

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -37,6 +37,7 @@ import net.runelite.api.AnimationID;
 import static net.runelite.api.AnimationID.*;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
+import net.runelite.api.Hitsplat;
 import net.runelite.api.NPC;
 import net.runelite.api.NPCComposition;
 import net.runelite.api.Player;
@@ -45,6 +46,7 @@ import net.runelite.api.Varbits;
 import net.runelite.api.events.AnimationChanged;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
+import net.runelite.api.events.HitsplatApplied;
 import net.runelite.api.events.InteractingChanged;
 import net.runelite.client.Notifier;
 import net.runelite.client.config.ConfigManager;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -318,6 +352,8 @@ public void onGameTick(GameTick event)
 		}
 	}
 
+
```
deathbeam: remove these 2 new lines

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -275,19 +281,47 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
 		}
 	}
 
+	@Subscribe
+	public void onHitsplat(HitsplatApplied event)
```
deathbeam: onHitsplatApplied

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -275,19 +281,47 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
+	{
+		final Hitsplat hitsplat = event.getHitsplat();
```
deathbeam: this should be after the if as it serves no point before if you return early from the method

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -275,19 +281,47 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
+			return;
+		}
+
+		final Hitsplat hitsplat = event.getHitsplat();
+
+		if (hitsplat.getHitsplatType() == Hitsplat.HitsplatType.DAMAGE
+			|| hitsplat.getHitsplatType() == Hitsplat.HitsplatType.BLOCK)
+		{
+			inCombat = true;
+			countdown = HIGHEST_MONSTER_ATTACK_SPEED;
```
deathbeam: You should reset both inCombat and countdown in resetTimers

deathbeam:
ALso you should probably add unit tests for this (now that idle notifier has some)

francomidence:
you say this was fixed but it isn't @deathbeam 
Any setting I should change or something?

Nightfirecat:
You should join our Discord server and ask in the `#support` channel. Given that this PR was merged 33 years ago, it's more likely a config issue than a problem with the plugin.

PR #5228, opened 2018-08-31
added solution to ACE MATCH ELM", "Cam The Camel"

merlindima:
added the solution "6" to clue step for ACE MATCH ELM


Abextm:
Are you 100% sure this is correct? The wiki says this should be a puzzle box.

merlindima:
you are right this is for MACHETE CLAM

merlindima:
wrong title

Kamielvf:
Cam is the solution to the hard anagram clue: ACE MATCH ELM. She will ask you to complete a Puzzle box.

She is also the solution to the elite anagram clue: MACHETE CLAM. The solution to the elite anagram clue is 6.

PR #5237, opened 2018-08-31
Add option to require control to be held for zooming

MagicfTail:


runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -117,21 +133,108 @@ public void onScriptEvent(ScriptCallbackEvent event)
...
+	}
+
+	private void updateControlZoom(Boolean value)
+	{
+		if (!loggedIn)
+		{
+			return;
+		}
+
+		if (value)
```
Nightfirecat: This whole if-else block can become:

```
client.getWidget(WidgetInfo.SETTINGS_TAB_SCROLL_WHEEL_INDICATOR).setHidden(value);
client.getVarbitValue(client.getVarps(), Varbits.SCROLL_WHEEL_ZOME.getId(), value ? 0 : 1);
```

MagicfTail: Yep, nice catch.

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomConfig.java
```diff
@@ -54,10 +56,22 @@ default boolean innerLimit()
...
+		position = 3
 	)
 	default boolean relaxCameraPitch()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "controlZoom",
+		name = "Control Zoom",
```
deathbeam: I would name this like "Require Control down" or something, "Control Zoom" seems weird

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -41,14 +48,23 @@
...
 	private ZoomConfig zoomConfig;
 
+	@Inject
+	private KeyManager keyManager;
+
+	@Inject
+	private EventBus eventBus;
+
+	private boolean loggedIn;
+	private boolean hotkeyDown;
```
deathbeam: You should reset hotkeyDown on focus change. Also, you can just use the Hotkey class we have instead of listener

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -117,21 +133,100 @@ public void onScriptEvent(ScriptCallbackEvent event)
...
+
+	private void updateControlZoom(Boolean value)
+	{
+		if (!loggedIn)
+		{
+			return;
+		}
+
+		client.getWidget(WidgetInfo.SETTINGS_TAB_SCROLL_WHEEL_INDICATOR).setHidden(value);
+		client.setVarbitValue(client.getVarps(), Varbits.SCROLL_WHEEL_ZOOM.getId(), value ? 0 : 1);
```
Adam-: this is supposed to be only for devtools

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -117,21 +133,100 @@ public void onScriptEvent(ScriptCallbackEvent event)
...
 	}
 
 	@Subscribe
 	public void onConfigChanged(ConfigChanged ev)
 	{
+		if (!zoomConfig.controlZoom())
```
Adam-: This will break all other uses of zoom.

MagicfTail: What do you mean?

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
MagicfTail: This was the simplest way I could think off while still allowing scroll to be used for other things than banking, and keep the "zoom bar" working

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -41,14 +48,23 @@
...
 	private Client client;
 
 	@Inject
 	private ZoomConfig zoomConfig;
 
+	@Inject
+	private KeyManager keyManager;
+
+	@Inject
+	private EventBus eventBus;
```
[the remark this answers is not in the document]
MagicfTail: This was from testing, should be removed

PR #5244, opened 2018-08-31
worldhopper: add default world hotkey

### LINKED ISSUE #4964: World Hopper: Default world keybind
Allow us to set a keybind to world hop to our default world so if we're using the default Ctrl+Shift+left/right a bunch we can just do something like Ctrl+Home and be sent to our default world if one is set.

devLotto:
Adds a hotkey to the world hopper for hopping to your default world.
The default hotkey is Ctrl+Shift+Home.

Closes #4964

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -55,10 +55,8 @@ import net.runelite.api.GameState;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
 import net.runelite.api.Varbits;
-import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.ConfigChanged;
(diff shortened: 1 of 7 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -159,6 +166,15 @@ public void hotkeyPressed()
 			hop(false);
 		}
 	};
+	private final HotkeyListener defaultKeyListener = new HotkeyListener(() -> config.defaultKey())
+	{
+		@Override
+		public void hotkeyPressed()
+		{
+			// Worlds start at 301, so add 300
+			hop(defaultWorldConfig.getWorld() + 300);
```
Adam-: Is the default world always configured to something valid?

devLotto: hop(...) checks if the world exists so I think it's ok if .getWorld() returns an invalid id

PR #5247, opened 2018-08-31
Plugin - Climb Switcher

killje:
A plugin that puts the climb-up (climb-down when holding shift) option on top for stairs and ladders that go up and down


killje:
The last build resulted in a ``java.net.SocketTimeoutException: connect timed out`` and does not seem to be the cause of the commit itself

Adam-:
This could probably just go into the menu entry swapper plugin?

killje:
Maybe it could, but it does uses a shift modifier that non of the menu entries use in the swapper plugin

killje:
I can add it no problem, i just don't know if i should add a "new" feature to the swapper plugin with different actions than it normally has

PR #5254, opened 2018-09-01
Added the maven-compiler-plugin with the useIncrementalCompilation set to false

CC007:
This change will make sure that when recompiling your code, it doesnt have to recompile all the files when it detects a change in only one file. Now it only has to compile that one file. This significantly speeds up the compilation time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,9 @@
 				<configuration>
 					<source>${java.version}</source>
 					<target>${java.version}</target>
+					<!-- Enable incremental compilation so that unchanged files dont have to be recompiled.
+					     False seems to actually enable incremental compilation, due to a bug in this mvn plugin:
+					     See: https://issues.apache.org/jira/browse/MCOMPILER-209 -->
 					<useIncrementalCompilation>false</useIncrementalCompilation>
 				</configuration>
 			</plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,7 @@
 				<configuration>
 					<source>${java.version}</source>
 					<target>${java.version}</target>
+					<useIncrementalCompilation>false</useIncrementalCompilation>
 				</configuration>
 			</plugin>
 			<plugin>


Adam-:
It looks like theres a maven bug open for this: https://issues.apache.org/jira/browse/MCOMPILER-209 where false is true and true is false. This should be put in a comment in the pom.xml.

CC007:
Is there a reason why this MR is not accepted? It would speeds up compilation time significantly without any downside to my knowledge. The RuneLite client alone already has over 700 classes and it takes over 10 seconds to compile. This hampers the development speed a lot. 

Think of having a bug in your code. If you try to find where it is with some log statements, it would be nice to not have to wait more than 10 seconds every time you add a log statement to a different part of your code.

raiyni:
You can use hot reloading.

CC007:
That would work in most circumstances, but that doesn't help if issues only happen during startup. For instance, if I start Runelite in debug mode and then add `log.info("Hello hot reloading");` to the main method, this doesn't get executed.

CC007:
Therefore, in situations where you actually need to compile again, this change would be beneficial.

(Ps: still thanks for the tip. This hot reloading does still help in a lot of situations.)

raiyni:
It just reloads the method bodies, not the class variables and imports.

CC007:
Comparison: 
Compile time using maven's compile goal with this change: 15.807 s
Compile time using maven's compile goal without this change: 46.594 s

CC007:
@raiyni I understand that. And therefore it would be nice to also be able to compile fast if you can't use your hotswapping trick.

CC007:
You sure keep your approved PRs open for a long time...

CC007:
Ok fair point. When adding or changing methods without changing their signature, useIncrementalCompilation can be off, but when changing the method signature or when removing the method, this isn't safe. Got it. Maybe a profile with this turned on could be added instead for those that know what it does and know when it is safe to use. Because it does speed up compilation a whole lot.

CC007:
I'll close this PR for now. Feel free to add such a functionality in a profile if you feel like you need it.

PR #5267, opened 2018-09-02
Show cannon spot overlay only with cannon

### LINKED ISSUE #1092: Cannon spots overlay should only show when you have a cannon in your inventory
Should probably depend on #1070

deathbeam:
This makes cannon spot overlays show only when cannon is present in
player inventory.

Closes #1092

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -38,6 +38,8 @@ import net.runelite.api.AnimationID;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameObject;
+import net.runelite.api.InventoryID;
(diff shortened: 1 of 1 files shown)


deathbeam:
Made the plugin require all cannon pieces in order to show overlay

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -137,6 +139,50 @@ protected void shutDown() throws Exception
...
+				continue;
+			}
+
+			switch (item.getId())
+			{
+				case ItemID.CANNON_BASE:
+				case ItemID.CANNON_STAND:
+				case ItemID.CANNON_BARRELS:
+				case ItemID.CANNON_FURNACE:
+					partsFound++;
```
TheStonedTurtle: May want to use a Set instead since this can be fooled by duplicate items. Not sure if its worth the effort though.

PR #5275, opened 2018-09-03
Add support for JSR 250 annotations

CC007:
This includes the annotations `@PostConstruct` and `@PreDestroy`

### DIFF
--- a/runelite-client/pom.xml
+++ b/runelite-client/pom.xml
@@ -149,6 +149,11 @@
 			<version>${guice.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>ru.vyarus</groupId>
+			<artifactId>guice-ext-annotations</artifactId>
+			<version>1.3.0</version>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -67,6 +67,7 @@ import net.runelite.client.ui.overlay.tooltip.TooltipOverlay;
 import net.runelite.client.ui.overlay.worldmap.WorldMapOverlay;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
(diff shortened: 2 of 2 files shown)


Adam-:
Is there existing code which can be improved with this?

CC007:
I have been using it in my branches and added it locally when I found out I needed it. I could have included it in the pull request of  the plugin that I am working on, but I wanted to make sure that this design choice would be accepted, before having to rewrite a lot of code.

CC007:
It could also be beneficial for other developers, since you cannot use the injected variables in constructors without it.

Adam-:
Generally for that you just inject the variables as parameters to the constructor.

CC007:
That is true but that would create a local variable that you dont need. Also there are multiple other uses, like when you need to use the variable `this` in a method call during the setup of your object.

deathbeam:
You can already do that with constructor as well. And, it is generally preferred to always use constructors for injection, to create mostly immutable classes (as the constructor is private, and injected fields final).

CC007:
When Ivwas talking about `this` in constructors, I was talking about the following: http://www.javapractices.com

This is not a problem in a method with the `@PostConstruct` annotation

deathbeam:
Do you have better link? That one is just poiting to jp homepage

CC007:
My bad: http://www.javapractices.com/topic/TopicAction.do?Id=252

CC007:
An example would be if you want to add a listener for a certain class automatically. You cannot safely do this from the constructor, because as my previous link states, it is not safe to pass `this` out of the constructor. That is no problem if you do it from the method with the `@PostConstruct` annotation, because there the object has already been properly instantiated, so it is safe to use the `this` keyword.

CC007:
Comparison: 
Compile time using maven's compile goal with this change: 15.807 s
Compile time using maven's compile goal without this change: 46.594 s

deathbeam:
Well the thing is I don't see how it would be useful in current codebase

CC007:
Ok, lets leave it out until it is actually used in a pull request. I'll reopen the PR then.

PR #5295, opened 2018-09-04
Make sprite and widget sprite overrides more flexible

deathbeam:
- Expose NodeCache interface in RuneLite
- Expose Client.getWidgetSpriteCache and reset method on it
- Change setters for sprite and widget overrides to be getters

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/interfacestyles/InterfaceStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/interfacestyles/InterfaceStylesPlugin.java
@@ -32,8 +32,6 @@ import java.awt.image.BufferedImage;
 import java.awt.image.PixelGrabber;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.Map;
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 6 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/interfacestyles/InterfaceStylesPlugin.java
```diff
@@ -145,7 +145,9 @@ private void overrideSprites()
 			}
 		}
 
-		client.setSpriteOverrides(overrides);
+		client.getWidgetSpriteCache().reset();
+		client.getSpriteOverrides().clear();
```
raiyni: would you really want to clear the entire map if someone else is using sprite overrides?

deathbeam: I mostly just made the change to do what it was doing till now, but now there is possibility to alter this to make it work with more plugins

deathbeam:
Probably not that one, because it is quite related to interface styles and it's resource folder. But I think in my original iteration where I actually needed this change I extracted the imageSpritePixels outside of the plugin. But, it probably do not has place in this change as it would be used at only at 1 place atm in master, and it is also very trivial to extract whenever you would need it (I decided to implement the not clearing of everything to the interface styles already here because it was not so trivial).

PR #5305, opened 2018-09-05
widgets: expose border thickness

raiyni:
Needed to make itemid widgets have a border


runelite-api/src/main/java/net/runelite/api/widgets/Widget.java
```diff
@@ -557,4 +557,14 @@
 	 * Sets the archive id of the font
 	 */
 	void setFontId(int id);
+
+	/**
+	 * Returns the border thickness of the widget
+	 */
+	int getBorderThickness();
```
deathbeam: If you are exposing new fields, you should probably make widget inspector know about them

raiyni: Sure, looks like I missed checking that file on commit

PR #5309, opened 2018-09-05
widget:expose setchildren

raiyni:
Some widgets set children to null vs just hiding all of the children when switching views.


runelite-api/src/main/java/net/runelite/api/widgets/Widget.java
```diff
@@ -124,6 +124,11 @@
 	 */
 	Widget[] getChildren();
 
+	/**
+	 * sets the widget children
```
deathbeam: Upper-case for Sets

raiyni: ok

Issue #4773, opened 2018-08-08
Postion/Stat Effect Timers on Death/Etc.


tgeene:
At the moment, when you die with potion/stat effects on your character, the timers keep ticking down like nothing happened. This can be confusing if you are doing things like learning bosses.

It would be wonderful if these overlays would be removed anytime your character does something that would reset/remove these effects from your character.

johnhamilto:
#5316 fixes this

PR #5316, opened 2018-09-05
remove timers on death

johnhamilto:
add removeall timers method.

automatically removes timers on death that are no longer relevant


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -45,16 +45,7 @@
...
-import net.runelite.api.events.AnimationChanged;
-import net.runelite.api.events.ChatMessage;
-import net.runelite.api.events.ConfigChanged;
-import net.runelite.api.events.GameTick;
-import net.runelite.api.events.GameStateChanged;
-import net.runelite.api.events.GraphicChanged;
-import net.runelite.api.events.ItemContainerChanged;
-import net.runelite.api.events.MenuOptionClicked;
-import net.runelite.api.events.NpcDespawned;
-import net.runelite.api.events.VarbitChanged;
```
raiyni: Why did these get removed?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -263,55 +254,55 @@ public void updateConfig(ConfigChanged event)
 	public void onMenuOptionClicked(MenuOptionClicked event)
 	{
 		if (config.showAntidotePlusPlus()
-			&& event.getMenuOption().contains("Drink")
-			&& (event.getId() == ItemID.ANTIDOTE1_5958
-			|| event.getId() == ItemID.ANTIDOTE2_5956
-			|| event.getId() == ItemID.ANTIDOTE3_5954
-			|| event.getId() == ItemID.ANTIDOTE4_5952))
+				&& event.getMenuOption().contains("Drink")
+				&& (event.getId() == ItemID.ANTIDOTE1_5958
```
raiyni: Indention levels are wrong

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -720,4 +717,9 @@ private void removeGameTimer(GameTimer timer)
 	{
 		infoBoxManager.removeIf(t -> t instanceof TimerTimer && ((TimerTimer) t).getTimer() == timer);
 	}
+
+	private void removeAllGameTimer()
+	{
+		infoBoxManager.removeIf(t -> t instanceof TimerTimer);
```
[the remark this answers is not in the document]
johnhamilto: right, forgot about those, going to only remove timers that are no longer relevant

johnhamilto:
closes #4773

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -482,6 +483,43 @@ public void onChatMessage(ChatMessage event)
 		}
 	}
 
+	@Subscribe
+	public void onLocalPlayerDeath(LocalPlayerDeath death)
+	{
+		removeGameTimer(ANTIDOTEPLUS);
```
NathenSample: Add these to a collection and iterate the collection rather than using 100 lines of code.

Honestly I'm not sure this is the correct way to go about this; Future proofing will be a nightmare, as people will create new timers which could be cleared on death, but don't know/think to add to, this class.

johnhamilto:
Ok I’ll try fixing that

deathbeam:
Any update on this?

johnhamilto:
@deathbeam Yes, just pushed.  Unsure if anti-venom is removed on death.  If anyone knows, lmk.

deathbeam:
Seems fine as it is, we can probably adjust it later as this PR is adding the necessary functionality to do that, should be simple enough

Nightfirecat:
I'll be honest, I feel like it would make more sense to maintain a list of timers to remove on death in the plugin than it does to add a field to the timers enum, as it's not really a property of the timer itself.

johnhamilto:
@Nightfirecat I think I actually disagree and believe it's in the right place.  The timer controls the time that the icon is up on the screen, and most notably, removes it when the time is up.  The only other thing that can remove a potion effect or buff is clan wars or death.  (May actually want to add clan wars).  It seems as much of a property of the timer as the actual length of the timer.

deathbeam:
Yeah I agree with @johnhamilto

Nightfirecat:
I don't see it that way--the timer enums are just definitions of lengths associated with certain actions. The logic around adding and removing them is wholly contained within the plugin. That said, given that this code works and looks fine, don't let my disagreement be a blocker for merging this.

deathbeam:
This needs rebasing now looks like

PR #5323, opened 2018-09-05
Add item charges display for ogre bellows

### LINKED ISSUE #5319: Add uses to ogre bellows
Add a number to ogre bellows that shows how many uses are left, like with teleport jewelry

MackBryan:
Closes #5319

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeType.java
@@ -26,6 +26,7 @@ package net.runelite.client.plugins.itemcharges;
 
 enum ItemChargeType
 {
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeType.java
```diff
@@ -26,6 +26,7 @@
 
 enum ItemChargeType
 {
+	BELLOW,
```
deathbeam: Why is this named BELLOW and not BELLOWS?

MackBryan: I was thinking the same thing when I was working on it, a fix will be pushed in a few minutes.

Issue #4624, opened 2018-08-01
Add help hint to !kc when client does not know the kill count


Nightfirecat:
**Is your feature request related to a problem? Please describe.**
Many players do not know how to set their kill counts in Runelite, and will reach out to our support channel to find this information.

**Describe the solution you'd like**
If a valid `!kc` command is entered but no kill count is known by the client yet, enter a warning/notice message in the user's chat (and don't send a public message) informing them of the issue and explaining how to correct it--by getting a new kill or checking RoW.

PR #5327, opened 2018-09-06
chat commands: Add kill count lookup error message

Nightfirecat:
For kill count lookups which fail, "[boss] kill count: unknown" will be
displayed. Additionally if it was the local player's kill count which
was unknown, "get kill or check ring of wealth", will be prepended as a
hint for the user to fix the issue.

Fixes runelite/runelite#4624


Nightfirecat:
fwiw this might not be needed if we decide to fall back to the (new since last PR update) hiscore boss KCs.

PR #5328, opened 2018-09-06
widget: borderThickness -> borderType

raiyni:
pedantic rename of method


deathbeam:
I dont think this will work as the export inside the internal branch is still named "borderThickness"

raiyni:
Yeah, someone has to rename it. Consensus was to rename when asked.

PR #5329, opened 2018-09-06
added swapping of the climb-up and climb-down options when clicking a ladder

johnhamilto:
Adds an option to the menu swapper to swap the default climb option on ladders to climb-up or climb-down when shift is held

+ various cleanup/bugfix in menuentryswapper


killje:
Probably also want it for target stairs

johnhamilto:
>Probably also want it for target stairs

@killje I had considered that.  I forgot about the spiral staircases in lumbridge castle for example.  I might just remove that check completely because if it can’t find an option to swap then it won’t be able to swap them anyway.

killje:
Yea, that is was i did in my plugin. Have not found it to break on something

PR #5346, opened 2018-09-06
Fix Achievment Diary widget ids

deathbeam:
This makes the achievment diary plugin work again.


Adam-:
It sounds like this should be reverted since it prevents being able from reading the full diary.

deathbeam:
Probably not as the widget IDs are correct, we just need to fix the scrollbar.

PR #5357, opened 2018-09-06
Box trap entry swap improvement

### LINKED ISSUE #5335: Split "Reset" option in menu entry swapper to ground laying and box resetting
Currently, enabling "reset traps" under menu entry swapper applies to 2 things:
1. It makes traps on the ground relayable with left click 
2. It makes the left click for failed traps reset them 
The left click lay box (that is on the ground) is very useful and handy,.
However, the reseting trap option is very time consuming, and pretty terrible overall. 

At the moment, enabling the swapper option puts them both on, while only ground laying is useful.

I would like an option to select only 1. ground laying or 2. box resetting.

A split in the option of the menu entry swapper, to choose ground laying seperate to box resetting.

MackBryan:
Closes #5335 
There are now three options that relate to different states on a box trip while hunting. It should give a user the ability to swap every useful entry when using box traps.
The config options do as follows:

- Check trap: Swap "Check" with "Reset" on box trap when a chin is successfully caught.
- Take trap: Swap "Take" with "Lay" on box trap when a box is on the ground and needs to be built.
- Dismantle trap: "Swap" Dismantle with "Reset" on box trap when a chin has disabled a trap when unsuccessfully caught.

*There is currently no feature to swap "Investigate" and "Dismantle" when a trap is active and waiting to be tripped. It can be added but seems unnecessary and not useful.*

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -165,17 +165,39 @@ public interface MenuEntrySwapperConfig extends Config
 
 	@ConfigItem(
 		position = 12,
-		keyName = "swapBoxTrap",
-		name = "Reset",
+		keyName = "swapBoxTrapCheck",
+		name = "Reset trap",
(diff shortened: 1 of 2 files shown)


raiyni:

PR #5358, opened 2018-09-06
Add Quest to quest screenshots

MagicfTail:
Add Quest to quest screenshots so all screenshots of quest completions are grouped


ShaunDreclin:
Windows no like `:`.

screenshot 1393

MagicfTail:
This doesn't exactly look the best, but it kind of follows the current naming scheme and should be fine on all platforms 🤷‍♀️

deathbeam:
Yeah with Quest() it is fine I guess

PR #5362, opened 2018-09-07
Expand checkstyle

Nightfirecat:
This introduces the charset property to help avoid issues on Windows machines, and adds rules regarding braces (required except for empty-body loops) and import wildcards (disallowed, except static imports).

Depends on !25

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -26,6 +26,7 @@
 <!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
 <module name="Checker">
+	<property name="charset" value="UTF-8"/>
 	<module name="TreeWalker">
 		<module name="LeftCurly">
 			<property name="option" value="nl"/>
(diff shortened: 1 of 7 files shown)


checkstyle.xml
```diff
@@ -44,6 +48,9 @@
 			<property name="tokens" value="COMMA"/>
 		</module>
 		<module name="UnusedImports"/>
+		<module name="AvoidStarImport">
+			<property name="allowStaticMemberImports" value="true"/>
```
deathbeam: Oh accidentally commented on wrong PR. Is it possible to make this check for > 10 static imports as we have defined in checkstyle?

Nightfirecat: We don't have that defined in checkstyle, only in the intellij settings. To my knowledge, there's no way to limit this module to below or above any specific number of imports.

deathbeam: Yeah I meant code style: `By changing Names count to use static imports with ''*' to 10 (static imports for f.e. ItemIDs with wildcards are preferred)`. But if it is not possible it is fine I guess

Nightfirecat:
Seems like this isn't wanted.

PR #5375, opened 2018-09-07
Add option to ignore mouse activity checks in idle notifier

### LINKED ISSUE #5374: An option to display idle notifications regardless of user input
The new Idle Notifications have one flaw, that isn't technically a flaw, but it could be a little better:
a notification is not displayed when the client detects user input, for example, when moving the mouse as the mining animation stops. 

The solution to this would be making it(preferably an option, since I'm sure not everyone would want this) so that the notification would be sent regardless if the user is active in the client or not. Consider such a case: 

The mouse is being dragged through the client as the skilling animation is finished. The user is not paying attention and the idle notification is not displayed, since the client detects mouse movement and therefore thinks that the user is not afk. The user now either notices this sometime after it happens or gets the notification that the account is about to log out, both of which may take some time and defeat the purpose of idle notifications.

**Additional context**
Ideally, there could be a checkbox in the Idle Notifcations plugin settings such as this:
"Display notifications regardless of user input". 
While having this on, the idle notifications are displayed no matter what. 
While having this off the idle notifications are not displayed when there is mouse activity in the client.

ShaunDreclin:
Resolves #5374

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -32,10 +32,10 @@ import net.runelite.client.config.ConfigItem;
 public interface IdleNotifierConfig extends Config
 {
 	@ConfigItem(
+		position = 1,
 		keyName = "animationidle",
 		name = "Idle Animation Notifications",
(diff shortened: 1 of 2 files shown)


deathbeam:
After #5403 this PR will probably not be necessary

ShaunDreclin:
#5403 makes this unnecessary

PR #5402, opened 2018-09-09
Do not forget last idle state when in IDLE already + add unit tests

deathbeam:
Do not forget last state when we are already in IDLE state and there
was mouse activity

Depends on and closes #5369 


deathbeam:
This is dependant on #5369 because as after resetting the idle timer it is never set on master again as player is in idle state already.

deathbeam:
Even though this is sort of working, it will just bring back falsy alerts. #5402 will prevent this as well, otherwise this is simply bug with how MLM works, so I am closing this.

PR #5409, opened 2018-09-09
runelite-api: Update widget field names

Abextm:
Updates all of the names from #5394 

Depends on !26
Closes #5328


runelite-api/src/main/java/net/runelite/api/widgets/Widget.java
```diff
@@ -566,14 +538,14 @@
...
+	 * 2 -> 1px black under 1px white border (selected item)
 	 */
-	int getBorderThickness();
+	int getBorderType();
 
-	/**
-	 * Sets the border thickness of the widget
-	 */
-	void setBorderThickness(int thickness);
+	void setBorderType(int thickness);
```
deathbeam: You should at least add `@see` annotation and reference the getter. In generated javadocs they will probably not be close to eachother

runescape-api/src/main/java/net/runelite/rs/api/RSWidget.java
```diff
@@ -274,21 +271,11 @@
...
 
-	@Import("paddingX")
-	@Override
-	int getPaddingX();
-
-	@Import("paddingX")
-	@Override
-	void setPaddingX(int paddingX);
+	@Import("xPitch")
+	int getXPitch();
```
deathbeam: So i assume this was removed from exports because it should not be messed with?

Abextm: I removed it because:
1) It seems very useless
2) Nothing used it
3) The docs were wrong anyway

PR #5415, opened 2018-09-10
Add overlay to help people with the light box

MagicfTail:
Add an overlay making it clear what button they should press to get it to show the solution, and showing them to only click one button each tick.

https://i.imgur.com/Bcm3mFD.gifv


runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverPlugin.java
```diff
@@ -73,11 +76,18 @@
...
+	private PuzzleSolverOverlay puzzleOverlay;
+
+	@Inject
+	private LightBoxOverlay lightBoxOverlay;
 
 	@Inject
 	private Client client;
 
+	@Getter
+	private ArrayList<String> unclickedButtons = new ArrayList<>(Arrays.asList("A", "B", "C", "D", "E", "F", "G", "H"));
```
Adam-: This seems like it could be greatly simplified just by using a bitfield?

devLotto:
This would be neat to have.

PR #5426, opened 2018-09-10
Fix WorldMapPlugin

deathbeam:
- Merge fairy ring icons and tooltips to 1 toggle (as one cannot work
without eachother, its confusing)
- Merge agility icons and tooltips to 1 (same reason)
- Make plugin work correctly during restarts


runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapConfig.java
```diff
@@ -38,40 +38,18 @@
 		description = "Display the code for fairy rings in the icon tooltip",
 		position = 1
 	)
-	default boolean fairyRingTooltips()
-	{
-		return true;
-	}
-
-	@ConfigItem(
-		keyName = WorldMapPlugin.CONFIG_KEY_FAIRY_RING_ICON,
```
Nightfirecat: Why not fix these features instead of requiring them to be dependent?

Nightfirecat:
Also, I'm not seeing any issues when turning the plugin off and on when running release, even if I start Runelite with the plugin off and turn it on after logging in. What issue is that fixing exactly?

deathbeam:
Fixed tooltips and icons working together

ShaunDreclin:
this will resolve #4455

PR #5429, opened 2018-09-10
http-service: add caching for hiscore lookups

Adam-:


http-service/src/main/java/net/runelite/http/service/hiscore/HiscoreService.java
```diff
@@ -45,12 +50,25 @@
 @Slf4j
 public class HiscoreService
 {
-	public HiscoreResultBuilder lookupUsername(String username, HiscoreEndpoint endpoint) throws IOException
+	private LoadingCache<HiscoreKey, HiscoreResultBuilder> hiscoreCache = CacheBuilder.newBuilder()
```
deathbeam: final

Issue #1518, opened 2018-04-14
Add links to wiki page to plugins


JerwuQu:
There are a lot of users asking questions for help with plugins so I propose we add a help button in the plugin options that leads to their wiki-page. 

![image](https://user-images.githubusercontent.com/3710677/38767096-4a5b24b2-3fdc-11e8-82a3-ea64a5fa37dc.png)

PR #5430, opened 2018-09-10
plugins panel: Add wiki link to name label

Nightfirecat:
This adds a hover color effect and a popup menu linking to the plugin's
wiki page to each plugin name label both in the plugin panel and
configuration panels. In the plugin panel, for plugins which have
configurable settings, it also adds a "Configure" menu item to the
added popup menu.

Plugin name label hover effect and wiki popup menu example

Configuration panel example

Fixes runelite/runelite#1518

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListItem.java
@@ -26,30 +26,43 @@ package net.runelite.client.plugins.config;
 
 import java.awt.BorderLayout;
 import java.awt.Color;
+import java.awt.Component;
 import java.awt.Dimension;
 import java.awt.GridLayout;
+import java.awt.MouseInfo;
+import java.awt.Point;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.awt.image.BufferedImage;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import javax.annotation.Nullable;
 import javax.swing.ImageIcon;
 import javax.swing.JLabel;
+import javax.swing.JMenuItem;
 import javax.swing.JPanel;
+import javax.swing.JPopupMenu;
+import javax.swing.SwingUtilities;
+import javax.swing.border.EmptyBorder;
 import lombok.AccessLevel;
 import lombok.Getter;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigDescriptor;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.PluginPanel;
 import net.runelite.client.ui.components.IconButton;
 import net.runelite.client.util.ImageUtil;
+import net.runelite.client.util.LinkBrowser;
 import org.apache.commons.text.similarity.JaroWinklerDistance;
 
 class PluginListItem extends JPanel
 {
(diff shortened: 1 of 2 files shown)


ShaunDreclin:
I'd suggest "Help" or "About This Plugin" or "Info" rather than Wiki, might be a bit clearer what it means

Nightfirecat:
I'm of the opposite mindset--if they didn't know what it meant, they'd learn after clicking it once, and that could help drive more attention to the wiki for new/curious players.

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListItem.java
```diff
@@ -143,6 +156,38 @@ private PluginListItem(ConfigPanel configPanel, @Nullable Plugin plugin, @Nullab
...
+				Component source = (Component) mouseEvent.getSource();
+				Point location = MouseInfo.getPointerInfo().getLocation();
+				SwingUtilities.convertPointFromScreen(location, source);
+				popupMenu.show(source, location.x, location.y);
+			}
+
+			@Override
+			public void mouseEntered(MouseEvent mouseEvent)
+			{
+				PluginListItem.this.lastForeground = nameLabel.getForeground();
```
deathbeam: you can just store lastForeground inside this anon impl right. that seems better

Nightfirecat: That does seem better. Didn't know that was an option. Done!

Nightfirecat:
I've added the same effect to the plugin label at the top of the configuration panel.

Configuration panel example

killje:
Should the wiki link maybe be overridable in the plugin descriptor to future proof it.

deathbeam:
Imo nope, forcing consistent naming for wiki is good thing

Nightfirecat:
The only future-proofing that's really necessary would be updating the wiki link function to properly handle non-word characters--the proper replacement is `s/[\W]/-/g`, but we're unlikely to have plugins with other characters in them anytime soon, if ever.

deathbeam:
Oh, now that @killje brought this up, where does "RuneLite" config lead to?

Nightfirecat:
https://github.com/runelite/runelite/wiki/RuneLite

killje:
Maybe add a test that gives a warning when the wiki page does not exists

deathbeam:
But the wiki page always exists, it will just prompt you to create it

killje:
When the wiki page is not created then -_-

killje:
Oww, no not a warning in the client. Add a warning to the tests so that it shows us in travis

Nightfirecat:
I'd be fine with that, but that's kind of out of scope for this change.

Abextm:
No one would ever see it unless it failed the build, and it shouldn't fail the build because the wiki page shouldn't be there until the build has passed and the pr has been merged.

RiddleTime:
@Nightfirecat would be it be an idea to have a wiki link or partial link to be required for every plugin, that so they could be part of the plugin description/config/requirements?

I don't want to sound harsh but my only suggestion would be the above including a change in the private static String wikiUrl wikiUrl() function.

Nightfirecat:
We have a very consistent URL schema for our wiki pages, so I'm not sure I understand your question. Why would we want to add a wiki URL requirement to the plugin descriptor?

RiddleTime:
@Nightfirecat then it's not a problem, for the current UI design what would be needed to integrate this helpful addition?

Nightfirecat:
I've updated this PR to add a "Configure" menu entry for plugin panel listings with configurable settings. The behavior is otherwise unchanged.

PR #5444, opened 2018-09-11
Add WorldHoppingManager

devLotto:
Allows for reuse of the world hopping logic.
Will allow for the default world plugin to call the world hopper for #4964.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2018, Lotto <https://github.com/devLotto>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
```diff
@@ -0,0 +1,115 @@
...
+	private Consumer<WorldHoppingResult> resultConsumer;
+	private int displaySwitcherAttempts = 0;
+
+	@Inject
+	public WorldHoppingManager(Client client)
+	{
+		this.client = client;
+	}
+
+	public synchronized void hop(net.runelite.api.World targetWorld, Consumer<WorldHoppingResult> resultConsumer)
```
deathbeam: You could maybe rather result future? That way you would be able to also properly cancel previous hop and the listener will be aware of it in case of multiple calls of this method (and you would not need custom 3-state enum)

runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
```diff
@@ -0,0 +1,115 @@
...
+	private static final int DISPLAY_SWITCHER_MAX_ATTEMPTS = 3;
+
+	private final Client client;
+
+	private net.runelite.api.World targetWorld;
+	private Consumer<WorldHoppingResult> resultConsumer;
+	private int displaySwitcherAttempts = 0;
+
+	@Inject
+	public WorldHoppingManager(Client client)
```
deathbeam: private

runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
```diff
@@ -0,0 +1,115 @@
...
+
+	public synchronized void hop(net.runelite.api.World targetWorld, Consumer<WorldHoppingResult> resultConsumer)
+	{
+		this.targetWorld = targetWorld;
+		this.resultConsumer = resultConsumer;
+		this.displaySwitcherAttempts = 0;
+	}
+
+	@Subscribe
+	public synchronized void onGameTick(GameTick event)
```
deathbeam: You should assume that event bus is thread-safe, making subscribers synchronized seems incorrect

runelite-client/src/main/java/net/runelite/client/game/WorldHoppingManager.java
deathbeam: You would create the future. And as you would resolve it in game thread, it will be run there. The 3 retries will be implemented in same way, i guess, you would just check for the future not being already cancelled each game tick. I guess you would need to make the retries volatile too. Rest I mostly posted in Discord already, but I can copy-paste it here:

```java
private volatile Future<> future;
...
hop()
...
if (future != null && !future.isCancelled)
{
  future.cancel()
};

...

future = new CompletableFuture<>;
addHopCallBackToFuture(future);
return future; // receiver will add callbacks as it pleases
```

PR #5445, opened 2018-09-11
entityhider: make hidden entities not occupy the tile they're on

### LINKED ISSUE #1979: If there is entity under hidden entity the entity will be hidden as well
When selected to **not** render players, if there is an NPC underneath of them the NPC will also not render.

devLotto:
Fixes #1979 

Only works internally until it's merged.

Result: https://gfycat.com/EverlastingLargeAlbino

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
@@ -29,7 +29,11 @@ import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import javax.inject.Inject;
 import net.runelite.api.Client;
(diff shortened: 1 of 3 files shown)


AppleSocks:
A large part of competitive castlewars is being able to quickly rightclick explode/bucket/tinder barricades underneath players the tick after they're set. If this is merged, players could toggle the "Hide Players" option to make these actions a leftclick, massively reducing the skillcap. 

Instead, maybe `shouldDraw` could return true if this NPC is in a blacklist of NPC ids? (inside a mixin so its not editable?)

devLotto:
I might be able to make it so nothing is hidden if you're inside Castle Wars. Would that be fine?

AppleSocks:
If you figure that'd be less expensive or hacky then yeah that'd work

devLotto:
Updated so that it disables the hiding if you're in a Castle Wars game.

PR #5449, opened 2018-09-11
Add option to always show fishing spots

MagicfTail:


MagicfTail:
Idk if it's really necessary, but I enjoyed always seeing the fishing spots. It also kinda (not really) gets around the problem of barehanded fishing.

deathbeam:
Needs rebase

Issue #5431, opened 2018-09-11
Hiscores lookup slow


humphrjm1:
Please check if your issue is not a duplicate by [searching existing issues](https://github.com/runelite/runelite/search?type=Issues)

**Describe the bug**
Hiscores lookup is unnecessarily slow
**To Reproduce**
Right click chat box, look up any player

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. Windows, Ubuntu, macOS] Win 10
 - RuneLite version: [e.g 1.4.6] 1.4.16
 - Launcher version: [e.g 1.5.2] 1.6.0

**Additional context**
Konduit is much faster. Not sure if they cache on their own website or cache locally somehow but it's much quicker.

Adam-:
I did https://github.com/runelite/runelite/pull/5429 today for this. I don't know how much it will help. It looks like Jagex is throttling us on their side.

Adam-:
We will see what #5429 does next release. If it doesn't improve the situation enough I think we will have to move the hiscore lookups to be all client side.

runelitepk:
Still an ongoing issue... now it also shows percentage of health instead of the exact health the opponent has. It's definitely more difficult to pk without the exact numbers. So not only does opponents hp show as percent, i also can't look up ppl so i won't know what their stats are, what their max hit will be. @Adam- What exactly do you mean by #5429 and "if it doesn't improve the situation enough i think we will have to move the hiscore lookups to be all client side"? please elaborate. I'm new to this whole thing and registered today to adress this issue. It's incredibly frustrating and i'm losing my patience. This has been an ongoing issue for a few days now. Thanks for addressing this issue @humphrjm1. Hp needs to be fixed too as it shows in percent now for some weird reason which it didn't do before.

runelitepk:
Player interaction setting doesn't work as well to show stats...

Adam-:
Fixed in #5473

PR #5473, opened 2018-09-13
Make HiscoreClient call the OSRS hiscore API directly

deathbeam:
See #5431 


deathbeam:
Looks like the OSRS API is really slow atm, so not sure if this will actually help, I was still getting timeouts and failed hiscore lookups even with this.

Adam-:
This is for #5431

PR #5480, opened 2018-09-13
ConfigManager: don't post an update if the values are the same

raiyni:


raiyni:
added the same check for unset

PR #5483, opened 2018-09-14
Loot tracker - Add ability to ignore items

### LINKED ISSUE #4661: "Omit value" option for loot tracker
frustrated when [...]

No related problems to what I would like to suggest.  Loot tracker doesn't have an option for not counting value of a specific loot, i.e. if I kill zulrah and receive 237 scales, 2 calquat seeds and 12 noted dragon bones I may consider leaving the calquat seeds as they are a very low value.  Other times you have possibly received the drop but were unable to collect it due to dying in an instance for example, and thus can't count that loot anymore.  Loot tracker only finds what was dropped and adds them together with no way to remove in certain circumstances for more precise measurement of gain.

The most user-friendly option I could think of personally would consist of a user being able to simply left click one of the loots displayed from a recent kill.  It would act as a toggle to remove the item's value from the overall value displayed.  It should also have some clear visual feedback, such as turning the panel behind that item specifically red to show that the loot's value has been omitted.

A clear and concise description of any alternative solutions or features you've
considered.

A possible solution might be to include a minimum value threshold for drops received in the player's name.  This would help players set the value to act as a catch all for loots under a specific amount to just not be added to the tracker.  This wouldn't offer the functionality of removing drops that were unable to be secured as previously described.

**Additional context**

Example of what visual feedback what look like if player selected to omit noted flax from loot:
https://i.imgur.com/Bq8QWaY.png

### LINKED ISSUE #5060: Shift click remove loots
frustrated when [...]
N/A

N/A

A clear and concise description of any alternative solutions or features you've
considered.

It'd be good to be able to Shift+click the loots inside the Loot Tracker to remove them (but not for remove all), I find myself wanting to keep some loots but remove certain ones. For example; when I'm at reves I want to remove the coins drop so I have to keep right clicking and pressing remove. I went to the abyss and had to remove about 30 loots but wanted to keep my previous 20 from revs so I had to right click remove the ones from the abyss while getting my pouch.

**Additional context**
N/A

psikoi:
Fixes #5060
Fixes #4661

This adds the ability to toggle items in the Loot Tracker, when an item is toggled, its price won't count towards the price check, and if the "Show ignored items" config option is not enabled, it won't be displayed on the loot tracker UI either.

Toggling items:

Hiding ignored items:

Ignored tooltip:

Ignored items config:


deathbeam:
This do not looks very intuitive. You should put the show ignored items at the top maybe, because otherwise it will be very annoying to ignore items then hide them then unignore later. Also, why red? You should probably grey out the box, red usually indicates error.

Nightfirecat:
What happens if you kill a monster which only drops ignored items? (Say you ignore bones and kill some monster that only drops bones)

deathbeam:
Looks like it will just display empty box

psikoi:
Displays the darker bar of each box, that displays:

[ Cow x 1      132 gp]

Nightfirecat:
That behavior feels inconsistent to me--if you kill a monster which has no drops, no loot tracker entry is created. I would expect the same behavior to happen if a monster dropped only ignored items.

deathbeam:
I partially agree with @Nightfirecat but also what if you want to unignore the items and the entry is simply not there

Nightfirecat:
I would think that ignoring the items would only take effect when building the entries, so un-ignoring them would not impact previous loot entries, only ones created after that point. That's just my take on this though. :woman_shrugging:

deathbeam:
But that means you are losing data. Atm it works just as filter, same as grouping/per-kill views, I dont think it should work otherwise.

Nightfirecat:
Then entries should just be hidden if they have no items? Just would feel wrong to show an entry with no items.

deathbeam:
Yeah probably that is good solution. And "unhide" them when it suddenly has some items based on updated filter

deathbeam:
So I guess simply check if item size > 0, if not, setHidden(true) on the box?

aussietinask:
Hi, would like to bump this feature request.

I like to track how much money I'm making, but there are some items that I will just not pick up, like bones and grimy marrentills/guams, or non-alchable equipment. This function would make it easier to track how much I'm actually earning.

An alternative is to use the GE Guide in the game to price check, but that makes it hard to see total income after considering things like multiple bank trips, or looting bag interactions.

Another way to do this might not be to actually ignore items, but have the option to 'delete' certain drops like bones after a session is complete? So I might:
- remove 100 bones from the loot tracker of cows that i've killed, which decreases gp by ~6k
- But if i kill another cow, the loot would come up again starting from 1.

I don't code, so I'm not sure how viable these ideas are, but would be awesome if they could be implemented! Thanks for developing such a great app!

aussietinask:
Honestly, I could also just go to loot tracker, open up a calculator, and subtract the total price of all the items I don't pick up, but it would be cool if the process could be streamlined. Thanks again!

psikoi:
This feature is only a few hours away from being done, i just need to find the time to complete it

psikoi:
Updated @deathbeam

Perelan2008:
Was this removed?

Issue #4213, opened 2018-07-07
Agility pyramid laps to level is inaccurate


Nightfirecat:
Ref: #4158

Because of the bonus experience gained on every lap, the lap counter to next level is inaccurate when training at the agility pyramid.

PR #5498, opened 2018-09-14
agility plugin: Fix Agility pyramid laps to level

Nightfirecat:
fix proof

```
$ python -c 'print(78147 / (722.0 + (80 * 8 + 300)))'
47.0198555957

$ python -c 'import math; print(math.ceil(78147 / (722.0 + (80 * 8 + 300))))'
48.0
```

Fixes runelite/runelite#4213


apex-meme-lord:
+1

Nightfirecat:
This has merge conflicts and if I update this, I'd rather make it consider the progress needed to the goal level (taking into account the exp increases between levels) instead of assuming a constant agility level to the goal. For now, I don't think I'm going to update this PR.

Issue #5504, opened 2018-09-15
Widget/Text/overlay additions/toggles


GoEasyImBad:
Would it be possible to add an option to “alt” or “ctrl” toggle for ground items to appear/disappear. Also adding more text options that are less bulky for widgets and overlays such as clue scrolls,mouse over text etc. 
being able to fill in the background of these overlays and widgets would along with the text changes would be a nice customization option.

deathbeam:
Maybe. It would help to close the issues as duplicates if you made 1 issue per request.

PR #5505, opened 2018-09-15
Adds an option to GroundItems plugin to always show +/- on ground items.

HotdogCombo:
Fills part of request: #5504


Kamielvf:
I think the issue is not for the +/- options, rather for the whole overlay so you could toggle viewing ground item labels or not.

deathbeam:
Yeah looks like this PR is misunderstandement of the original issue, closing

PR #5507, opened 2018-09-15
Herbiboar Plugin: Disable left-click "Pick"

ltvill:
image

Taken from Corporeal Beast plugin.


deathbeam:
I dont think we should do this. Also, it has nothing to do with Herbiboar?

deathbeam:
https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features actually, we have it listed here

ltvill:
I did Herbiboar for the first time today, some of the mushrooms are close to inspect objects. Can easily cause misclicks.

PR #5525, opened 2018-09-17
menu entry swapper: Swaps wield/release option for salamanders. 

bowlcutmullet:
The option is disabled by default.

Not sure if there is already another way to do this, but i couldn't find one. Makes hunting salamanders a bit more chill.


Alexsuperfly:
you can sticky keys shift

bowlcutmullet:
> you can sticky keys shift

Couldn't you then customize shift click to bury bones (for example) and then sticky key shift. I don't think many people know you can do that, it's quite unintuitive. 

Another solution would be to have a "customize left click" that works like customize shift click. But idk if that would be possible/allowed by jagex.

deathbeam:
For this we have shift-click configuration. Most of remaining options in menu-entry swapper that swap inventory are mostly from before we had shift-click.

Nightfirecat:

PR #2509, opened 2018-05-09
More accurate NPC hp overlay when max hp is known.

henke96:
I figured out how the server calculates the healthRatio it sends the client (which is healthRatio = 1 + (healthScale - 1) * health / maxHealth for health > 0 and 0 for health = 0).

I have tested this on quite a lot of different npcs, for some i've confirmed every single hp value. I've also tested on bosses with large hp bars and have not been able to disprove it at all.

Knowing how the server calculates the ratio means we can reverse the interval of possible hp that could've resulted in it. Which is exactly what i've done, i then show the middle between the max/min possible hps.

Note that the current hp display is incorrect even for NPCs with a 1:1 relation between possible hp and healthratio (those with max hp <= health scale).

PR #5528, opened 2018-09-17
More accurate NPC hp overlay when max hp is known

### LINKED ISSUE #58: Opponent info health off by 1
When an opponent is damaged, the health will be 1 lower than it actually is. This results in the opponent info showing 0/maxhp when in reality the opponent is 1hp. I have tested this on cows, but I had noticed it before.

raiyni:
https://github.com/runelite/runelite/pull/2509 appears to be stale.

Rebase of it. 

Closes #58
Closes #2509


Nightfirecat:
Closes #58
Closes #2509

Issue #375, opened 2018-01-14
Display virtual levels after 99 in the skill tab


Dreyri:
For us nerds who love xp.

Maybe a special kind of flair when 200m is hit.

deathbeam:
+ Highscores

Adam-:
Hiscore panel virtual levels implemented in #1525

PR #5534, opened 2018-09-18
Add Virtual Levels plugin

Nightfirecat:
This plugin displays the virtual level in the denominator of given
skills, and the virtual total level on the skills tab.

Virtual Levels plugin demonstration

Fixes runelite/runelite#375


runelite-client/src/main/java/net/runelite/client/plugins/virtuallevels/VirtualLevelsConfig.java
```diff
@@ -0,0 +1,56 @@
...
+@ConfigGroup("virtuallevels")
+public interface VirtualLevelsConfig extends Config
+{
+	@ConfigItem(
+		position = 0,
+		keyName = "skillTab",
+		name = "Show virtual levels on skill tab",
+		description = "Configures whether the skill tab should show virtual levels."
+	)
+	default boolean showOnSkillTab()
```
deathbeam: What is the reason for this config to exist? Isnt this basically what enabling the plugin should do? or you can just make both configs enum, I dont think you should ever end up in state where you would have plugin enabled but it would actually do nothing.

PR #5574, opened 2018-09-21
New Player Information Plugin

xandayn:
A new plugin for replicating the information provided by the Minimap Orbs as InfoBoxes.

Specifically useful for Steamers/Content Providers who's overlay blocks the minimap orbs but still want their viewers to be able to see them easily.

Also fixes an issue with net.runelite.api.SpriteID.MINIMAP_ORB_RUN_ICON having an incorrect value (1070 instead of 1069)

Preview


raiyni:
This would be fixed with status bars. Also they should put the webcam somewhere else then imo

xandayn:
True that could be a solution just moving it, but not every setup really allows for that, plus it's a nice feature to have regardless since we can't move the orbs separately from the minimap, and it's optional anyways. Also what do you mean with "status bars"?

PR #5598, opened 2018-09-23
Add security stronghold answers to Puzzle Solver

RSNinja:


runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverPlugin.java
```diff
@@ -60,14 +60,15 @@
 import net.runelite.client.util.Text;
 
 @PluginDescriptor(
-	name = "Puzzle Solver",
-	description = "Show you where to click to solve puzzle boxes",
-	tags = {"clues", "scrolls", "overlay"}
+		name = "Puzzle Solver",
```
Nightfirecat: These indents should not be changed.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/SecurityStrongholdAnswer.java
```diff
@@ -0,0 +1,99 @@
...
+    QUESTION_27("Virus scan my device then change my password."),
+    QUESTION_28("No, you should never buy a RuneScape account."),
+    QUESTION_29("No way! I'm reporting you to Jagex!"),
+    QUESTION_30("No way! You'll just take my gold for your own! Reported!"),
+    QUESTION_31("Politely tell them no and then use the 'Report Abuse' button."),
+    QUESTION_32("No."),
+    QUESTION_33("Report the player for phishing."),
+    QUESTION_34("Don't give them the information and send an 'Abuse Report'.");
+
+    private static final List<String> MATCHES = new ArrayList<>();
```
Nightfirecat: Use `SecurityStrongholdAnswer.values()` (or just `values()` for code within the enum) instead of this.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverPlugin.java
```diff
@@ -174,6 +192,7 @@ else if (widgetId == LIGHT_BOX_BUTTON_G.getId())
 			combination = Combination.G;
 		}
 		else if (widgetId == LIGHT_BOX_BUTTON_H.getId())
+
```
Asymons: i think you have an extra space here

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/SecurityStrongholdAnswer.java
```diff
@@ -0,0 +1,103 @@
...
+	QUESTION_27("Virus scan my device then change my password."),
+	QUESTION_28("No, you should never buy a RuneScape account."),
+	QUESTION_29("No way! I'm reporting you to Jagex!"),
+	QUESTION_30("No way! You'll just take my gold for your own! Reported!"),
+	QUESTION_31("Politely tell them no and then use the 'Report Abuse' button."),
+	QUESTION_32("No."),
+	QUESTION_33("Report the player for phishing."),
+	QUESTION_34("Don't give them the information and send an 'Abuse Report'.");
+
+	private static final List<String> MATCHES = new ArrayList<>();
```
Asymons: Would it be better to use a HashSet instead of a list? Lookup would be O(1) instead of O(n), correct me if I'm wrong tho.

NathenSample: Are you sure you don't work at my office? We were having this exact discussion last week. You stalker.

HashSet is generally viewed as being `O(1)` (And in this case will be) but can, in very edge cases, get performance as bad as `log(n)` (But how often do hash collisions actually occur?)

An ArrayList is indeed `O(n)` since it just uses `.indexOf()`

With such a tiny collection we're into tiny amounts of nano-second difference between the two.

So tiny difference, but given the use-case, no real reason to NOT use the set over the list.

Asymons: Yep that's completely understandable, for this particular use case it doesn't really matter in terms of optimization; however, when a newer person is looking at your code in the future for example, it may be best to write it in the best way in case that user does copy pasta but for a larger data set or is looking at your code as a learning opportunity. So I'd still stand by switching it over to a HashSet, but it's understandable if you don't switch it because in terms of optimization it barely contributes.

RSNinja: I will change it to HashSet, the order isn't important in this case anyway.

PR #5599, opened 2018-09-23
Reduce tab size in notes panel to 2

### LINKED ISSUE #5591: Adjust tab distance in notes plugin
I keep a lot of things in my note pad, but it be dreadfully hard to read and keep organized.

I think decreasing the tabbing spacing would go a long way to aid in creating clear notes.

It would also be awesome to add bullet lists with a hot-key to activate.

9474f8fc586c3ff816c40f34d174814c

bmoyer:
Fixes #5591 

New tab size:
image


deathbeam:
Hmm 1 is maybe too low, 2 is probably good middle ground

bmoyer:
Changed to 2.

PR #5603, opened 2018-09-23
Interface plugin: Added 2005 chatbox

### LINKED ISSUE #3201: 2005 Interface - Chat Box not changed
Great update today 5/24/2018. 

The option for the **2005 interface** should also change the text box because that is major part of the theme itself. Can the 2005 interface setting be updated to **change the chat box** as well.

SebastiaanVanspauwen:
Added and modified the 2005 chatbox so it doesn't clip with chat

Fixes #3201
Closes #5493


troypressley1:
Yay, finally!

robmonte:
Can we not get the torn page look along the bottom of the background in the 2005 version?

PR #5611, opened 2018-09-24
Corp chat message

steffenhauge:
Added toggle for chat messages and changed the color to pink as suggested.


deathbeam:
Wait now thinking about this, looks like the color is already configurable via Chat Color plugin..

steffenhauge:
But wouldn't that also change the kc chat color?

deathbeam:
Yeah

deathbeam:
Actually, the in-game one? probably not

steffenhauge:
So, should I just close this PR and open another one for the previous commit?

deathbeam:
you can just remove the color related changes I guess

steffenhauge:
There :)

PR #5619, opened 2018-09-24
Fix clipping in fixed mode for edge snapping map points

### LINKED ISSUE #5618: Clue map markers getting clipped in fixed mode
https://gyazo.com/ee99649ab8ddb060f5207ef21b1fcbaf

https://gyazo.com/efdc018da55d212eb9ced0ff7cf28293

trimbe:
The bounds for the canvas in fixed mode are offset to the right by the size of the border. This translates the canvas bounds back to the left so we don't clip the icon if it's on the left side.

Fixes #5618

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
@@ -92,7 +92,11 @@ public class WorldMapOverlay extends Overlay
 
 		final Rectangle worldMapRectangle = widget.getBounds();
 		final Area mapViewArea = getWorldMapClipArea(worldMapRectangle);
-		final Area canvasViewArea = getWorldMapClipArea(client.getCanvas().getBounds());
+		final Rectangle canvasBounds = client.getCanvas().getBounds();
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Oh, so it's the black borders when the Runelite client is larger than the fixed mode size that affects this? I see. Will test this and confirm the behavior is correct later today.

runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
```diff
@@ -92,7 +92,23 @@ public Dimension render(Graphics2D graphics)
 
 		final Rectangle worldMapRectangle = widget.getBounds();
 		final Area mapViewArea = getWorldMapClipArea(worldMapRectangle);
-		final Area canvasViewArea = getWorldMapClipArea(client.getCanvas().getBounds());
+		Area canvasViewArea;
```
Nightfirecat: Looks like this can be final? It will only be assigned to once.

deathbeam:
Tbh, wouldn't simply setting X/Y of the canvas bounds unconditionally to 0 here solve the issue as well?

Issue #5570, opened 2018-09-21
Hide specific monsters opponent info if they have in-game ones


Beckxr:
Monsters such as ToB bosses and Zuk for example have in-game hp bars already, there is no need to show runelite hps.

hide the runelite overlay when attack specific npcs.

Felanbird:
This was added in late 2021 7b6074f0f86320efe644db08b40633b362e10bfb

PR #5639, opened 2018-09-25
Option to remove opponent info on bosses who already have a vanilla one

robinwithes:
Added a toggle-able feature to disable health bars on bosses who have a vanilla one, the npc's that are blacklisted are in opponentinfooverlay.java all the way at the bottom.

Im not sure if there is a easier way to do this. If so tell me and i'll close this pr.

Solves https://github.com/runelite/runelite/issues/5570


deathbeam:
You can probably just detect if widget is visible and do not display opponent stats in that case

robinwithes:
> You can probably just detect if widget is visible and do not display opponent stats in that case

I assumed so but then if you attack other npcs in the area then they wont show either.

robinwithes:
Im going to close this one to try and remake it in a different way. Thanks for the input Nightfire

PR #5650, opened 2018-09-26
HP Regeneration notifier option

robinwithes:
This is a personal addition i made a while back to aid with nightmare zone afk-ing and i figured it might be useful to some other people as well.

Q:
Why didn't i make it into idle notifiers or the nightmare zone plugin?
A:
2 Reasons: Either i would have to remake what regen meter does in a different place or i would have to rely on regen meter being enabled, as well as this being useful in situations like giant mole as well. thus not fitting solely in the nightmare zone area.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
@@ -61,7 +61,7 @@ public interface RegenMeterConfig extends Config
 	@ConfigItem(
 		keyName = "regenWarning",
 		name = "Show a warning when about to regenerate health.",
-		description = "0 Means disabled, the value can be 0 - 99 with the value equalling the percentage of hp regeneration.")
+		description = "0 Means disabled, the value can be 0 - 99 with the value being the percentage of hp that triggers a warning.")
 	default int regenWarning()
 	{
 		return 0;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
@@ -57,4 +57,13 @@ public interface RegenMeterConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "regenWarning",
+		name = "Show a warning when about to regenerate health.",
+		description = "0 Means disabled, the value can be 0 - 99 with the value equalling the percentage of hp regeneration.")
+	default int regenWarning()
+	{
+		return 0;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018 Abex
+ * Copyright (c) 2018 Robin Withes
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
@@ -60,8 +60,8 @@ public interface RegenMeterConfig extends Config
 
 	@ConfigItem(
 		keyName = "regenWarning",
-		name = "Show a warning when about to regenerate health.",
-		description = "0 Means disabled, the value can be 0 - 99 with the value being the percentage of hp that triggers a warning.")
+		name = "Health Regeneration Notification Threshold",
+		description = "The amount of remaining health regeneration % to send a notification at. A value of 0 will disable notification.")
 	default int regenWarning()
 	{
 		return 0;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
@@ -154,11 +154,7 @@ public class RegenMeterPlugin extends Plugin
 			// Show it going down
 			hitpointsPercentage = 1 - hitpointsPercentage;
 		}
-		else if (maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && (Math.round(hitpointsPercentage * 100)) == config.regenWarning() && !client.isPrayerActive(Prayer.RAPID_HEAL))
-		{
-			notifier.notify("You have reached your HP Regeneration Threshold!");
-		}
-		else if (!messageSend && maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1 || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() ) && client.isPrayerActive(Prayer.RAPID_HEAL))
+		else if (!messageSend && maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1))
 		{
 			notifier.notify("You have reached your HP Regeneration Threshold!");
 			messageSend = true;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
```diff
@@ -57,4 +57,13 @@ default boolean showWhenNoChange()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "regenWarning",
+		name = "Show a warning when about to regenerate health.",
+		description = "0 Means disabled, the value can be 0 - 99 with the value equalling the percentage of hp regeneration.")
```
raiyni: spelling error `equalling`

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -149,5 +154,16 @@ else if (currentHP > maxHP)
 			// Show it going down
 			hitpointsPercentage = 1 - hitpointsPercentage;
 		}
+		else if (maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && (Math.round(hitpointsPercentage * 100)) == config.regenWarning() && !client.isPrayerActive(Prayer.RAPID_HEAL))
```
raiyni: These blocks seems overly complex and can be reduced to 1 statement.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
```diff
@@ -57,4 +57,13 @@ default boolean showWhenNoChange()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "regenWarning",
+		name = "Show a warning when about to regenerate health.",
```
deathbeam: You should probably follow the naming for notification alers in Idle Notifier (e.g for HP etc)

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -149,5 +154,12 @@ else if (currentHP > maxHP)
 			// Show it going down
 			hitpointsPercentage = 1 - hitpointsPercentage;
 		}
+		else if (!messageSend && maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1))
+		{
+			notifier.notify("You have reached your HP Regeneration Threshold!");
```
deathbeam: You should probably follow the notification pattern in idle notifier here

robinwithes: I am not a 100% sure what you meant with this but i think my last commit fixes the issue you had with it?

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -149,5 +155,13 @@ else if (currentHP > maxHP)
 			// Show it going down
 			hitpointsPercentage = 1 - hitpointsPercentage;
 		}
+		else if (!messageSend && maxHP > currentHP && config.regenWarning() > 0 && config.regenWarning() < 100 && ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1))
+		{
+			Player local = client.getLocalPlayer();
+			notifier.notify("[" + local.getName() + "] has reached the regeneration threshold!");
```
deathbeam: regeneration threshold sounds too debug-like, maybe [ username ] HP is about to regen

or something like that?

robinwithes: Yeah good idea, pushed now.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -74,6 +80,8 @@
 	private int ticksSinceSpecRegen;
 	private int ticksSinceHPRegen;
 	private boolean wasRapidHeal;
+	private boolean messageSend;
+	private boolean checkHealthPercentage;
```
deathbeam: Do not make this var instance var, just declare it in method

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -140,6 +148,7 @@ private void onTick(GameTick event)
 
 		int currentHP = client.getBoostedSkillLevel(Skill.HITPOINTS);
 		int maxHP = client.getRealSkillLevel(Skill.HITPOINTS);
+		checkHealthPercentage = ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1);
```
deathbeam: reuse Math.round(hitpointsPercentage * 100). Also why the check for config.regenWarning() - 1?

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -149,5 +158,13 @@ else if (currentHP > maxHP)
 			// Show it going down
 			hitpointsPercentage = 1 - hitpointsPercentage;
 		}
+		else if (!messageSend && maxHP > currentHP && checkHealthPercentage && config.regenWarning() > 0)
```
deathbeam: you should check config.regenWarning() > 0 before everything

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -140,6 +148,7 @@ private void onTick(GameTick event)
 
 		int currentHP = client.getBoostedSkillLevel(Skill.HITPOINTS);
 		int maxHP = client.getRealSkillLevel(Skill.HITPOINTS);
+		checkHealthPercentage = ((Math.round(hitpointsPercentage * 100)) == config.regenWarning() || (Math.round(hitpointsPercentage * 100)) == config.regenWarning() - 1);
 		if (currentHP == maxHP && !config.showWhenNoChange())
```
deathbeam: move this !config.showWhenNoChange() inside the currentHp == maxHp check

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
robinwithes: Because when using rapid heal it can be on a different tick.

NathenSample:
Rather than checking the config every tick could you implement an `onConfigChanged`?

i.e

```
@Subscribe
public void onConfigChanged(ConfigChanged event)
{
  if (event.getGroup().equals("regenWarning") && (Integer.valueOf(event.getNewValue) > 0))
  {
    warningEnabled = true;
  }
}
```
I'm sure the performance overhead is entirely negligible, but I still dislike the idea of pulling a value out of the config every tick.

From a readability standpoint I believe it makes it more obvious the circumstances in which the notification is disabled.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -140,7 +147,9 @@ private void onTick(GameTick event)
 
 		int currentHP = client.getBoostedSkillLevel(Skill.HITPOINTS);
 		int maxHP = client.getRealSkillLevel(Skill.HITPOINTS);
-		if (currentHP == maxHP && !config.showWhenNoChange())
+		long hpPercentage = (Math.round(hitpointsPercentage * 100));
+		boolean checkHealthPercentage = (hpPercentage == config.regenWarning() || hpPercentage == config.regenWarning() - 1);
+		if (!config.showWhenNoChange() && currentHP == maxHP)
```
deathbeam: by moving this inside if I meant inside if.

e.g

```
if (currentHP == maxHP)
{
  if (!config.showWHenNoChange())
}
else (regenStuff)

**Comment by raiyni** [comment]
There is probably 0 performance impacts and it looks cleaner for it to just be a method call vs having it setup in config changed.

**Comment by robinwithes** [comment]
> 
> 
> There is probably 0 performance impacts and it looks cleaner for it to just be a method call vs having it setup in config changed.

Ye i agree,
would just add a whole chunk of code to do the same that's already been done.

File: runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java [hunk 11]
```diff
@@ -140,14 +147,27 @@ private void onTick(GameTick event)
 
 		int currentHP = client.getBoostedSkillLevel(Skill.HITPOINTS);
 		int maxHP = client.getRealSkillLevel(Skill.HITPOINTS);
-		if (currentHP == maxHP && !config.showWhenNoChange())
+		long hpPercentage = (Math.round(hitpointsPercentage * 100));
+		boolean checkHealthPercentage = (hpPercentage == config.regenWarning() || hpPercentage == config.regenWarning() - 1);
+		if (currentHP == maxHP)
```
deathbeam: add empty line above

Nightfirecat:
This has been implemented in #8789.

PR #5672, opened 2018-09-27
Impling notification

bowlcutmullet:
Option in Implings plugin to get notifications about specific impling types when they appear/spawn.

imps


Trevor159:
This is in the client already.

PR #5686, opened 2018-09-28
WIP - Swap NPC names in clue scrolls with NPC ids

deathbeam:
- Make NpcClueScroll use npc id instead of name
- Replace NPC names in AnagramClue with NPC ids
- Make CipherClue use NPC ids instead of names
- WIP - Change NPC names to ids in CrypticClue


deathbeam:
This is too annoying to do, mainly because the size of the clue scroll files lags my computer :d

Issue #5689, opened 2018-09-28
Super antifire is missing from herblore skill calculator


Majst:
**Describe the bug**
Super antifire (non extended) isn't calculated in the herblore skill calculator pane.

PR #5708, opened 2018-09-29
Added super antifire to skill calculator

robinwithes:
fixes https://github.com/runelite/runelite/issues/5689


runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_herblore.json
```diff
@@ -329,6 +329,12 @@
       "icon": 21978,
       "name": "Extended Super Antifire (4)",
       "xp": 160
+    },
+    {
+      "level": 92,
```
Nightfirecat: This should be placed between super combat and anti-venom potions so it is in ascending level order.

PR #5731, opened 2018-10-01
Remove 'Take' option from items on the ignored list.

Reasel:
This option is added under the grounditems plugin. Appears as

Option

This allows the user to hide items on the hidden list. Making it easier to loot and generally ignore people dropping junk at various activities.

Demo


trimbe:
This is a rejected feature.

1337umad:
runeliteplus accepts rejected features ;)
https://github.com/runelite-extended/runelite/
https://runelitepl.us

PR #5734, opened 2018-10-01
Make raids scout overlay not show at Olm

### LINKED ISSUE #5721: Remove raid room overlay for Olm
Could the overlay showing the raid room order be removed during Olm? It takes up a bunch of room (on fixed mode) and provides no obvious benefit after you've finished the other rooms.

One possible way of going about this, if the game message log can be read, is to look for the pink "lower level complete!" message as seen below, and remove the overlay at that point. The only time this message should be seen is when the final room has been completed and olm prep is going on or the olm fight is about to start.

Kamielvf:
This will prevent the scout overlay from rendering while at Olm as the information there is obsolete.

Closes #5721

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
@@ -29,6 +29,7 @@ import java.awt.Dimension;
 import java.awt.Graphics2D;
 import javax.inject.Inject;
 import lombok.Setter;
+import net.runelite.api.Client;
 import net.runelite.client.plugins.raids.solver.Room;
 import net.runelite.client.ui.overlay.Overlay;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
```diff
@@ -47,18 +49,20 @@
...
+		this.client = client;
 		this.plugin = plugin;
 		this.config = config;
 	}
 
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		if (!config.scoutOverlay() || !scoutOverlayShown)
+		//Do not show the scout overlay while we are down at olm (plane == 0)
```
Nightfirecat: Maybe make `private static final int OLM_PLANE = 0` and use that to get rid of this comment?

PR #5740, opened 2018-10-01
Implemented NPC name truncation in loot tracker

### LINKED ISSUE #5673: Loot Tracker: Truncate NPC name rather than kill amount
If the loot tracker is tracking an NPC with a particularly long name (Thermonuclear Smoke Devil) with over a 3 digit kill count (100+), the amount of NPCs killed in the title gets turned into a set of ellipses, obscuring your kill count for that npc.

Rather than ellipsing the full line of text in the header, truncate the NPC's name, so that the kill amount is always visible.

Patriq:
Fixes #5673.

Only truncates the title of the log. Which means both subtitle and price will show up totally.
Because the name can now be truncated, I also added a tool tip for usability sake.

Example of it working:
example

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
@@ -28,6 +28,7 @@ package net.runelite.client.plugins.loottracker;
 import com.google.common.base.Strings;
 import java.awt.BorderLayout;
 import java.awt.Color;
+import java.awt.Dimension;
 import java.awt.GridLayout;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


deathbeam:
I am not sure if this is correct solution. This could probably be done with messing around with BorderLayout a bit, maybe use WEST for label, and EAST for subtitle and gp value and dont use center? not sure

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -143,9 +146,21 @@ void combine(final LootTrackerRecord record)
...
 		}
 
+		resizeLogTitle();
+
 		repaint();
 	}
 
+	private void resizeLogTitle()
+	{
+		double maxWidth = itemContainer.getMinimumSize().getWidth() - 14;
```
Nightfirecat: This should get the border width instead of using a magic number.

PR #5741, opened 2018-10-01
Implemented the ability to hide both globes and xp tracker after 99

### LINKED ISSUE #2631: Disable tracking of skills after 99 (Option)
Give us the option to hide this - https://gyazo.com/87c5c950595288c8f0191a48923103a4 

I'd rather not see skills that I've already got 99 in.
87c5c950595288c8f0191a48923103a4

Patriq:
Fixes #2631.

Not sure if two different configurations was the best way of do it. I could have used the XP Tracker config for the XP Globals, but thought mixing configs was a mess.

Let me know if I should change this.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Experience.java
+++ b/runelite-api/src/main/java/net/runelite/api/Experience.java
@@ -43,6 +43,11 @@ public class Experience
 	public static final int MAX_SKILL_XP = 200_000_000;
 
 	/**
+	 * Maximum effective skill level at 13,034,431 experience.
+	 */
+	public static final int MAX_REAL_LEVEL = 99;
+
+	/**
 	 * The total experience required for each skill level.
 	 */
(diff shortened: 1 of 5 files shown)


runelite-api/src/main/java/net/runelite/api/Experience.java
```diff
@@ -42,6 +42,11 @@
 	public static final int MAX_VIRT_LEVEL = 126;
 	public static final int MAX_SKILL_XP = 200_000_000;
 
+	/**
+	 * Maximum effective skill level at 13,034,431 experience.
+	 */
+	public static final int MAX_REAL_LEVEL = 99;
```
Nightfirecat: I'd actually place this at the beginning of the class, above the virtual level/xp definitions.

psikoi:
What about https://github.com/runelite/runelite/pull/5021 ? Probably should be closed

deathbeam:
As long as this PR is open the other one is valid too i guess

PR #5742, opened 2018-10-01
Screenshot Plugin: Add option to hide Runelite overlays while screenshot is taken

### LINKED ISSUE #5552: Make all custom interfaces & overlays dissapear temporarily when taking screenshots
Currently when taking screenshots, all custom overlays like stat boots, targets and other map markers are shown in the screenshot itself. If I want to take a 'clean' screenshot I have to first disable all overlays, and then take the screenshot (potentially missing the moment).

Could it be made possible that all overlays are temporarily hidden when a screenshot is taken? Maybe it could be made optional in the screenshots tab as I could imagine it might take a game tick before the overlays are hidden and some players won't like it (it might also cause screenshots for level-ups to be missed mid-combat).

Disabling all overlays manually (a pain in the ass)

trimbe:
Adds a config option to hide our overlays while screenshots are taken.

Closes #5552


deathbeam:
Cant you just do the check at the beginning of the render method? Instead of doing it at 2 places

trimbe:
The screenshot overlay needs to render if the user wants the date to be shown.

deathbeam:
Oh totally missed that. It would make sense if it would not render, also referencing plugin overlay from OverlayManager seems like bad idea (the overlay should not have public modifier in first place)

deathbeam:
Looks like travis is failing

PR #5761, opened 2018-10-02
Add support for making the boost colour yellow if below boost threshold

### LINKED ISSUE #4767: Have boosted level turn to red when it reaches your set notification level in settings
When training combat, it would be great if your boosts didn't just stay green until they reset to normal. The notification is a bit more clunky as you'd have to enable all notifications while client is focused for it to give you the desired result. This way, once your eyes notice the change in color, boom, re-pot or re-spec. 

Thank you!

MagicfTail:
Fixes #4767

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
@@ -76,12 +76,12 @@ public class BoostIndicator extends InfoBox
 		int boosted = client.getBoostedSkillLevel(skill),
 			base = client.getRealSkillLevel(skill);
 
-		if (boosted > base)
+		if (boosted < base)
 		{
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
```diff
@@ -76,12 +76,12 @@ public Color getTextColor()
 		int boosted = client.getBoostedSkillLevel(skill),
 			base = client.getRealSkillLevel(skill);
 
-		if (boosted > base)
+		if (boosted < base)
```
deathbeam: This should be boosted <= base

deathbeam: I guess it does not really matter, the color is not applied when not boosted anyway

PR #5765, opened 2018-10-02
Only add 1 search option

### LINKED ISSUE #2094: Multiple occurencies of "Search x" on inventory item
image

raiyni:
Closes #2094 

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
@@ -35,6 +35,7 @@ import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
 import javax.inject.Inject;
@@ -57,6 +58,7 @@ import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.GrandExchangeOfferChanged;
 import net.runelite.api.events.MenuEntryAdded;
+import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetID;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -266,10 +268,32 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
 				menuEntry.setOption(SEARCH_GRAND_EXCHANGE);
 				menuEntry.setType(MenuAction.RUNELITE.getId());
+				entries = Arrays.copyOf(entries, 2);
+				entries[1] = menuEntry;
 				client.setMenuEntries(entries);
 		}
 	}
 
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked event)
```
deathbeam: Does this actually work? Because you cant click when holding alt

raiyni: Yes, you release alt.

deathbeam: But the menu option is not added when not holding alt, or? At least that is the purpose of the keybind, for it only searching when ALT is pressed

raiyni: The menu still shows up when release alt.

deathbeam: That sounds like bug with it not being removed then

raiyni: If you hold alt and click on the menu, it still searches.

Kamielvf: I cannot even open the right click menu whilst holding alt.

raiyni: it's a very specific timing of right clicking and hitting alt at the same time

deathbeam: It is a bug, so I would not add more stuff to make the bug work more, I would just fix the issue you are referencing in this PR rather than doing extra stuff

Kamielvf: So for anyone to use this feature they have to try ~20 times to get the timing right? This seems stupid, I just managed to get it but this seems wrong.

PR #5766, opened 2018-10-02
Fix KitType index for the head slot

### LINKED ISSUE #4866: getEquipmentId() returns -1 for HEAD KitType, even when target player has something equiped
**Bug description**
I have noticed that the `getEquipmentId()` function is returning -1 (no itemID or nothing equipped) for the head slot of a target player when the target player is actually wearing something in the head slot.

**To Reproduce**
This can be reproduced in a plugin. The following code should reproduce the same bug, where `p` is a `Player` that has already been targeted. In my case, the player is targeted using an additional menu entry when right-clicking a player in-game.

```
PlayerComposition playerComposition = p.getPlayerComposition();
for (KitType kitType : KitType.values())
{
     int itemId = playerComposition.getEquipmentId(kitType);
     log.debug("itemId: {}", itemId);
}
```

What is interesting, is that I am getting different results using the following two functions:

- `getEquipmentIds()`
- `getEquipmentId()`

The output I am getting from each are:

- `getEquipmentIds()`: `[1091, 1533, 0, 0, 1089, 0, 0, 1523, 386, 3454, 9517, 266]` 
- `getEquipmentId()`: `[1021, -1, -1, 577, -1, 1011, -1, 2942, 9005, -1]`

Please note, that the `getEquipmentIds()` output above is not taking into account minus 512 from any value greater than 512 (as documented in the API). When you do perform that, you will get the same ID values. However, the returned item ID value 1091 (aka actual item ID number 579) - which is the head slot that is missing... in this case a blue wizard hat (579).

**Expected behavior**
I expected the head slot to be returned when using `getEquipmentId()`.

**Screenshots**
Not applicable

Windows 10, IntelliJ with JDK version 8u181, git version object hash: `6ef32bf91`

**Additional context**
Not applicable

Happy to help with any additional questions (and hope I am not missing something basic which is causing this problem!). Thanks!

trimbe:
Closes #4866

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/kit/KitType.java
+++ b/runelite-api/src/main/java/net/runelite/api/kit/KitType.java
@@ -36,13 +36,14 @@ import net.runelite.api.PlayerComposition;
  */
 public enum KitType
 {
+	HEAD(0),
 	CAPE(1),
 	AMULET(2),
 	WEAPON(3),
 	TORSO(4),
 	SHIELD(5),
 	LEGS(7),
-	HEAD(8),
+	HAIR(8),
 	HANDS(9),
 	BOOTS(10),
 	JAW(11);


Adam-:
HEAD is different from HELMET I think?

Joshua-F:
Yeah, it appears that `0` is `HAT` (or something relating to still showing the rest of the head) where as `8` is covering the whole head.

trimbe:
`8` appears to be hair. I've tested by getting a haircut and updated the PR.

PR #4591, opened 2018-07-31
Add Progress arc color based on skill to Xp Globes plugin.

Monsterxsync:
-Relocated SkillColor.Java from xptracker plugin to game folder (fits in well here as its where the icons get loaded for orbs.)

A screenshot of this feature in use: 
screenshot

As you can see it uses the already implemented xp tracker colors for consistancy.

PR #5786, opened 2018-10-03
Add default skill color to XP globes

### LINKED ISSUE #2888: Different coloured XP globes for each skill
I would love to have each skill's xp globes in a different colour, just like in konduit, I think it makes them look a lot nicer, customisable colours would be even better.

psikoi:
The previous similar PR has been abandoned for months even tho I supplied most of the code for it (wasn't a lot anyways).

This is a simple improvement to the exp globes, something that I'm sure a lot of people want, including me.

**If any of you native english speakers have some ideas on how to improve the wording of the config item, i'd appreciate it, I had a bit of trouble with that.**

Should probably close #4591
Closes #2888

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
@@ -55,10 +55,21 @@ public interface XpGlobesConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "enableCustomArcColor",
+		name = "Enable custom arc color",
+		description = "Enables the custom coloring of the globe's arc instead of using the skill's default color.",
+		position = 2
+	)
+	default boolean enableCustomArcColor()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "Progress arc color",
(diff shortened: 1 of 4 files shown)


Kamielvf:
I think this would be acceptable to have enabled by default, or make the toggle `Use custom orb color` instead.

psikoi:
I think it makes more sense to be the default behaviour, but I'd like more opinions about it just to make sure. Also, wouldn't "custom" mean the custom user selected colors?

Kamielvf:
Yes, so the toggle would be to enable the custom colour instead of having the default behaviour of the skill-selective colouring.

psikoi:
@deathbeam any thoughts on the matter?

psikoi:
Updated the config item as such:

**Enable custom arc color**
Enables the custom coloring of the globe's arc instead of using the skill's default color.

psikoi:
Can I get a merge on this? @deathbeam

Monsterxsync:
Im confused this is exactly the same as what i have done months ago (its up to date and not abandoned btw) ???

deathbeam:
@Monsterxsync I will squash the changes and set psikoi as co-author then and you as author

Monsterxsync:
@deathbeam sounds good thank you.

PR #5789, opened 2018-10-04
Added a remaining experience field to skill calculator

### LINKED ISSUE #4816: Show difference in XP in Skill Calculator
For the skill calc when you put in your current level and target lvl can you also show the difference in xp between the two? thats what im always trying to find and it would be nice if it could be included #suggestions

patriq:
Fixes #4816. Thanks @ConnorVG for the patch he provided, did half the work.
Did some UI tweaks to his solution.

Example:
Example

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -51,6 +51,7 @@ import net.runelite.client.ui.DynamicGridLayout;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.ui.PluginPanel;
 import net.runelite.client.ui.components.IconTextField;
+import net.runelite.client.util.StackFormatter;
 
 class SkillCalculator extends JPanel
 {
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Not gonna lie, I don't like it being a full-width input field when the two rows above aren't.

Nightfirecat:
I'm envisioning it being on the right like the other two experience fields, but with no field on the left. Don't know if that would look wrong though

Patriq:
Well originally I had it on the left and didn't really like it, that is why I went full width. I could try to put it on the right and left and show a screenshot so you could compare.

Patriq:
@Nightfirecat Does this look better?
Example

Nightfirecat:
In my opinion, yes.

ConnorVG:
I completely forgot about this, surprised someone kept the patch for so long 🤔

psikoi:
I'd keep it full width, but instead of it looking like an input field, just make it a label, that way it wouldn't be confusing to the user

raiyni:
Added in https://github.com/runelite/runelite/commit/2d351768fbd6ed4d4437cad4f34fa2c943b503ae

PR #5790, opened 2018-10-04
Add ultimate force notification trigger to NMZ Plugin

### LINKED ISSUE #5787: Adding Ultimate Force Support to NMZ Plugin
Despite what most guides tell you in saying Ultimate force is not useful, it actually is for slayer. If you have a troll or black demon task you can afk it in NMZ. Ultimate force is helpful in the fact that it will kill the monsters for you yet still grant slayer xp, and if there is no slayer monster spawned you can clear all of the current ones out in hopes to spawn one.

A checkmark added for ultimate force powerup that works the same as the others, can be defaulted to disabled since it's not a very popular powerup.

mikek2:
Closes #5787 

Makes sense to have native support for it in the NMZ plugin rather than having to purposely use another plugin (chat notification etc)


deathbeam:
I think the ultimate force was purposely left out because it is useless

PR #5792, opened 2018-10-04
update FORLUN anagram clue tip to point to dungeon entrance

### LINKED ISSUE #5644: Runolf Elite clue not showing on worldmap
image
image
It does say "Miscellania & Etceteria Dungeon", but the dungeon entrance is not marked on the map.

Brodan:
Fixes #5644.

This clue is not currently showing up on the map because it's underground, specifically in the Miscellania and Etceteria Dungeon. This PR updates the `WorldPoint` for this `AnagramClue` to point to the coordinates of the entrance to that dungeon. Runolf is within a couple steps of that dungeon entrance, so he should not be difficult to find.


Abextm:
While this isn't a great solution to all clues, it works good enough for this one. By the time you get into the cave the NPC will be in view and that should take over.

It would be nice to have a way to specify multiple points for varying locations so things like the dragon eye clue work right, but it isn't necessary for this.

forsco:
Would it not then be better to leave it how it was so that it doesn’t require changing back in the future? Especially if someone intends to look at multi Point handling?

Abextm:
I would rather fix a subset of the problem trivially than block it indefinitely. If this was a 50+ line change I would be more worried, but its small and easy to revert should we gain the feature needed to fix it correctly.

Nightfirecat:
After thinking about it again, I agree with abex. This one will almost always show the NPC immediately after going down the ladder. I'd like to see a more comprehensive multi-point clue solver for these problematic clue steps (dragon eye, mausoleam) for anything that's not so trivial.

Brodan:
@Nightfirecat circling back on this. based on the discussion it seems like this is good to go. can this get merged still?

PR #2872, opened 2018-05-18
Remove custom chrome

deathbeam:
In order to maintain somewhat healthy ClientUI implementation and to not
break native OS functionality, remove custom window decorations and
move everything from previous title bar to sidebar.

The main points why this should be implemented:
1. We are facing with users asking why their client is not behaving like other windows do. E.g, not snapping to window corners, not working when using some applications that manipulate/detect windows (OBS for example) and many other
2. The amount of workaround work that will need to be done to solve all issues that come with having undecorated window that basically detaches from native OS window manager can get unmaintainable really quick, and at the moment we tend to tell people to just disable custom window chrome to solve their issues

I am opening this PR also to see people's reactions and get some discussion going on this topic.

Closes #1953 and closes #1426 and closes #956 and closes #2335 and closes #3138  

## New
peek 2018-04-28 12-26
screenshot 2018-04-28 13 02 20
screenshot 2018-04-28 13 02 15
screenshot 2018-04-28 13 02 09

## New (with Obsidian)
screenshot 2018-05-25 18 30 38

## Old:
screenshot 2018-04-26 21 53 07
screenshot 2018-04-26 21 52 58
screenshot 2018-04-26 21 52 47

PR #5797, opened 2018-10-04
Add support for hiding/showing sidebar button with custom chrome disabled

deathbeam:
Draw hide/show sidebar button in top-right corner using Game's draw
hook. This will not work in Vanilla and so it is disabled in it.

screenshot 2018-05-25 18 30 38

peek 2018-06-06 10-34

This is extracted from #2872

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
@@ -26,13 +26,10 @@ package net.runelite.client.util;
 
 import java.awt.AWTException;
 import java.awt.Color;
-import java.awt.Dimension;
 import java.awt.Font;
 import java.awt.Frame;
 import java.awt.Image;
-import java.awt.Rectangle;
 import java.awt.SystemTray;
-import java.awt.Toolkit;
 import java.awt.TrayIcon;
 import java.awt.Window;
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Why are the SwingUtil changes included? It doesn't even seem to be used in the sidebar arrow code.

Nightfirecat:
Also, how does this look in fixed mode with the FPS plugin enabled?

deathbeam:
I was using that before, I saw unused method, I removed it. And, in fixed mode it looks exactly like in resizable mode with FPS plugin enabled

runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
```diff
@@ -186,20 +183,6 @@ public void mouseClicked(MouseEvent e)
...
 	}
 
-	/**
-	 * Check if point is in screen bounds.
-	 *
-	 * @param x the x
-	 * @param y the y
-	 * @return the boolean
-	 */
-	public static boolean isInScreenBounds(final int x, final int y)
```
Nightfirecat: As I understand it, the changes in this file were from a cherry-picked commit from the "Remove custom UI" PR, but I'd like to see them be removed or in a separate commit since they aren't used in the substantive part of this change.

PR #5800, opened 2018-10-04
Add new Al-Kharid palace agility shortcuts

forsco:
Depends on #5799

Adds the new Al-Kharid palace shortcuts to the Agility plugins obstacles list.

notes marker


forsco:
Will rebase upon the merge of 5799.

PR #5805, opened 2018-10-04
Add Baby Chinchompa Metamorphosis swapper

### LINKED ISSUE #1729: Menu Swap Metamorphosis For Chinchompa Pet
Would be great if we're able to menu swap to the metamorphosis option, so we're able to spam click that option to obtain golden skin rather than having to right click & select.

e9c70fff957d08e866a4829926026daa

forsco:
Closes #1729


raiyni:
This can be accomplished with the customizable left click now

Issue #5294, opened 2018-09-04
Chat commands should support different ways to type commands


Unmoon:
Currently chat commands only work with specific strings. For example, level command only supports `!lvl`, but not `!level`.

`total` -> `total, overall`
`price` -> `price, ge, value(?)`
`lvl` -> `lvl, level`
`cmb` -> `combat, cmb, cb`

Unmoon:
As a side note, if anyone looks in to chat commands, the short boss names should probably be changed to not be a big pile of switch.

astaninger:
what would you want instead of the switches?

Unmoon:
Probably something similar to `SkillAbbreviations` (also part of chat commands plugin).

deathbeam:
I think we decide to not add abbrevs because other clients will not work with those (e.g OSB)

PR #5810, opened 2018-10-04
Added alternatives for chat commands

failip:
Implements the feature requested in #5294

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -219,11 +219,11 @@ public class ChatCommandsPlugin extends Plugin implements ChatboxInputListener
 				log.debug("Running killcount lookup for {}", search);
 				executor.submit(() -> killCountLookup(setMessage.getType(), setMessage, search));
 			}
-            else if (config.lvl() && LEVEL_COMMAND_SET.contains(command))
-            {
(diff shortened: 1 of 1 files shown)


raiyni:
This and https://github.com/runelite/runelite/pull/5801

look like blatant copy and paste for hacktoberfest, they are even split out into 2 prs.

failip:
Where exactly did I copy this from?

NathenSample:
Both this PR, and #5801, when combined together, implement the same changeset that #5303 attempts to introduce.

Your modifications of `ChatCommandsPlugin` are incredibly close to the modifications #5303 makes, but you use sets rather than regex, which could be pure coincidence.

However in #5801 it is painfully obvious that `BossAbbreviations.java` is a direct copy and paste from #5303 which draws further attention to the similarities that this PR has to how it was performed there.

failip:
I know you guys wont believe me for whatever reason but I actually just looked at #5294 and the replies, where Unmoon suggested to do the boss names like the SkillAbbreviations (I just compared mine and the one I 'copied' and I actually changed a subtle spelling mistake which I definitely wouldn't have noticed if I just copy pasted). 

I split the PRs into two separate ones because I though that would be the proper way to do it, as this is my first time contributing to an OSS. 

Just found it a bit rude to imply ill intent, if I just wanted PRs I could've just opened my own public repo.

So should I combine the two PRs? And any actual comments :D?

Nightfirecat:
Per comments on #5294, we're not looking to add this feature anymore in order to maintain consistent cross-client command names.

Issue #4960, opened 2018-08-17
Runecrafting Altar menu-entry swapping


TheArbalist:
Could we have the Menu Entry Swapper get the ability to make the left-click option for all RC Altars be "Walk here", instead?

If you're crafting combination runes, it's easy to accidentally click the "Craft-rune" option on the altar.
I think that a lot of people would be glad to see a plugin that helps prevent this, as it incurs a rather grating penalty due to the delay and the much-reduced experience for that inventory.

LlemonDuck:
This can be done by configuration with hub plugin [Combo Runes Only](https://runelite.net/plugin-hub/show/combo-runes-only) or with the new shift-click custom swaps.

PR #5818, opened 2018-10-05
Runecrafting Altar menu-entry swap

p0tata:
This is for the feature-request https://github.com/runelite/runelite/issues/4960

Makes it so a left click on the runecrafting altar is walk here, preventing you from accidentally crafting normal runes while doing combination tunes.


raiyni:
You can do this with the menu swapper right click config options now

PR #5825, opened 2018-10-05
Added herb sack content value to examine plugin

TheLonelyDev:
This PR adds the functionality to display the value of the herbs within the herb sack. The values are displayed when the user uses the 'check' option.

image

image

The way it works is by checking the chat messages and checking if said message is a SERVER message and contains the keyword 'Grimy'. After this it just extracts the herb's name and quantity.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
@@ -162,13 +162,16 @@ public class ExaminePlugin extends Plugin
 		{
 			String name = new String(event.getMessage().replaceAll(".* x Grimy ", "").trim().toLowerCase());
 
-			PendingExamine pendingExamine = new PendingExamine();
-			pendingExamine.setWidgetId(CHATBOX.getId());
-			pendingExamine.setActionParam(Integer.parseInt(event.getMessage().replaceAll( " x Grimy .*", "")));
-			pendingExamine.setType(ExamineType.ITEM_BANK_EQ);
-			pendingExamine.setId(herbs.get(name));
-			pendingExamine.setCreated(Instant.now());
-			pending.push(pendingExamine);
+			if (herbs.containsKey(name))
+			{
+				PendingExamine pendingExamine = new PendingExamine();
+				pendingExamine.setWidgetId(CHATBOX.getId());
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Would a drop notice for noted grimy herbs not trigger this then? Seems like you need a more strict trigger for the message.

runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -75,6 +77,24 @@
 		.maximumSize(128L)
 		.build();
 
+	private static final HashMap<String, Integer> herbs = new HashMap<String, Integer>()
+	{{
+		put("guam leaf", 199);
```
raiyni: Use the ItemID class.

runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -138,6 +158,22 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 	@Subscribe
 	public void onChatMessage(ChatMessage event)
 	{
+		if (event.getType() == ChatMessageType.SERVER && event.getMessage().contains("Grimy"))
+		{
+			String name = new String(event.getMessage().replaceAll(".* x Grimy ", "").trim().toLowerCase());
```
raiyni: Why are you replacing Grimy? They usually have a different price from clean.

TheLonelyDev:
> Would a drop notice for noted grimy herbs not trigger this then? Seems like you need a more strict trigger for the message.

I tested some cases and it did not for the things that were tested, I will commit some more tests and supply the results.

runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
Nightfirecat: This should be replaced with a Pattern for performance.

Alexsuperfly:
this closes #4573

TheLonelyDev:
The drop notice does not trigger the code
image
image

Nor does examine for both the grimy and clean variant (eg for a second time/infinite loop). And I have been using the plugin for some time while playing and I have not noticed any false triggers so far. I am happy to test some more cases if there are any other cases that need to be tested.

PR #5828, opened 2018-10-05
Fix rs2asm parser grammar to correctly handle EOF

deathbeam:
If the file do not ends with newline, it is still valid script but
grammar assumed that newline at the end of each statement is always
required. Also add checking of EOF to correctly wait for end of file
instead of trying to process even unfinished input.

### DIFF
--- a/cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
+++ b/cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
@@ -24,7 +24,7 @@
  */
 grammar rs2asm;
 
-prog: NEWLINE* (header NEWLINE+)* (line NEWLINE+)+ ;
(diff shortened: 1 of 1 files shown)


cache/src/main/antlr4/net/runelite/cache/script/assembler/rs2asm.g4
```diff
@@ -24,7 +24,7 @@
  */
 grammar rs2asm;
 
-prog: NEWLINE* (header NEWLINE+)* (line NEWLINE+)+ ;
+prog: NEWLINE* (header NEWLINE+)* (line NEWLINE*)+ EOF;
```
Adam-: I probably don't want newlines optional on this.

deathbeam: I did not found better way how to make the last newline optional what would not freeze the assembler. Any tips?

deathbeam:
As I have no better ideas, closing

PR #5830, opened 2018-10-05
Refactor high alchemy price calculations

psikoi:
5 different plugins (was about to 6 with the Loot Tracker) used the same copy and pasted code to calculate high alchemy price of an item, decided to create a method in ItemManager for that and edit all the plugin to reference that method.

There was also introduced a new exception to the calculations that would make sure items with 0 gp alch wouldn't falsely return 1 as the price.

However, there still seems to be a problem in the high alchemy value calculation as some items return wrong values when Math.round is used, and others return wrong values when it is not used, **this problem already existed before this PR**, I just hope that by bringing all those calculations to one single place will make the fix easier, in the future, if one exists.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
@@ -259,7 +259,7 @@ public class ItemManager
 		return price;
 	}
 
-	public int getAlchValue(int itemID)
+	public long getAlchValue(int itemID)
 	{
 		if (itemID == ItemID.COINS_995)
 		{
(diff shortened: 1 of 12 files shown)


runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -232,7 +235,7 @@ public void onGameStateChanged(final GameStateChanged event)
 	 * @param itemID item id
 	 * @return item price
 	 */
-	public int getItemPrice(int itemID)
+	public long getItemPrice(int itemID)
```
deathbeam: these both methods can be int, because both GE and HA price can never exceeed max int

psikoi:
[](url)

PR #5840, opened 2018-10-05
Add toggle to enable/disable the team-cape image styled overlay

odxs:
Allows for the user to define what type of overlay they prefer (text-style or image-based) for the team-cape plugin.

scdg7kj


deathbeam:
Better solution would be simply to just add tooltip for each image

odxs:
> Better solution would be simply to just add tooltip for each image

The entire point of this PR was to provide two different styling options for the plugin. I am in a pure clan (we use this plugin all the time) and many of the members have told me that they do not find the image-style overlay to be appealing and preferred the older style as it's easier to look at. I agree with them which is why I wanted to give people the option of what styling they wanted.

Issue #1785, opened 2018-04-22
Infobox grouping


MagicfTail:
Since you can't move our infoboxes separately, you are forced to have your boost icons, prayer icons, chest icon, and so on in the same place. I don't think most players like this since they, by the looks of it, place their non-combat icon by the chat, but boosts in the top left corner.
This could be fixed by making icon groups, such as combat, potion, non-combat, or by allowing contributors to create custom icon groups.

Dupe of comment from #1722

deathbeam:
Maybe each plugin should have it's own movable part of the infoboxes, that would be cleanest solution and easy to implement, I think.

Hydrox6:
Closed by #12550

Issue #1369, opened 2018-04-11
Split Kingdom Overlay and Potion Overlay


Stickaplex:
When trying to move the potion overlay the kingdom overlay "sticks" to it so they have to be in the same place. Would be nice if you could have these as 2 different things

deathbeam:
That is because miscellania overlay is infobox so you can move them as whole only, same applies to slayer infobox and when stats are displayed as infoboxes.

Stickaplex:
Unfortunate but still would be nice especially considering potions are typically a cb thing and kingdom not so much

PR #5868, opened 2018-10-07
Moveable Infobox groups

raiyni:
Closes https://github.com/runelite/runelite/issues/1785
Closes https://github.com/runelite/runelite/issues/1369

Note: There is a 1 frame flicker that happens with this config enabled. Since overlay positioning is 1 frame behind, I'm not sure how to solve that.


Nightfirecat:
One future improvement I could imagine beyond this would be finding a way to separate individual infoboxes from within a group--for example, separating one specific timer (or a small subset of timers) from the rest.

PR #5874, opened 2018-10-08
Kingdom Plugin: fix throne completion state check

### LINKED ISSUE #1980: Kingdom of Miscellania Favor using wrong quest requirement
Does not display Favor during and after quest complete of throne of miscellania even if standing in miscellania

trimbe:
Now uses the correct varp from the quest list completion state script (1358).

Additionally, allows the infobox to show if the quest is in progress to allow users to check favor during the last part of the quest. 

Closes #1980


runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -32,6 +32,8 @@
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import static net.runelite.api.ItemID.TEAK_CHEST;
+
```
deathbeam: Remove this empty line

Issue #617, opened 2018-02-17
Follow Guice best practices


deathbeam:
Currently, the model we are creating when using Guice and it's injectors is not really following Guice best practices, mainly:

* [Minimize mutability](https://github.com/google/guice/wiki/MinimizeMutability)
* [Inject only direct dependecies](https://github.com/google/guice/wiki/InjectOnlyDirectDependencies)
* [Keep constructors hidden](https://github.com/google/guice/wiki/KeepConstructorsHidden)

For first point, changing all the injections from `@Inject` annotation on fields to `@Inject` annotations
on constructor should be sufficient.

For second point, f.e. ClientUI should somehow provide the plugin panel, then `net.runelite.api.Client` needs to be split to multiple interfaces and each of them needs to be provided separately, and plugins should use only the interface they really need (for example to get camera values, they should really just inject something like ClientCamera, instead of injecting Client that contains everything).

For third point, just making the constructors private should be good enough (depends on first point).

If anyone have any more ideas on how to improve the current Guice model, feel free to post them here.

Desetude:
I think the best way to improve the current Guice model is by seperating Plugin into Plugin and Module, where the module instantiates the plugin and provides the config, etc..
This new Plugin structure will allow constructor injection, rather than field, which is evidently favourable, and the structure will generally be more flexible, e.g. with how a plugin is instantiated. Also, this new model will be a lot less hacky to implement (https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java#L403 is definitely not 'natural' guice injection).

NathanielPather:
This is still despite a solution being merged.

deathbeam:
No, there are still quite a few cases in source where people are ignoring best practices, mainly for 1st and 3rd point.

Nightfirecat:
(1) was addressed in 7ceb3c9e020aa8316ddc5809a2d409d5d2efb697, and (2) is as far as I'm aware, not an issue in this codebase. That means (3) should be the only point needing to be addressed, which I plan to open a PR to fix shortly.

PR #5876, opened 2018-10-08
Better chat input

Abextm:
Adds a text menu input that allows the user to select out of a predefined list in the gameframe:

Adds a text input that supports arrow keys, keyboard and mouse selection, and copy/cut/paste.
full edit input

Also gives us a framework to add basically whatever we want to that panel.

Currently nothing uses this, but I would like to convert all uses of `ChatboxInputManager` to this in the future.


runelite-api/src/main/java/net/runelite/api/FontID.java
```diff
@@ -0,0 +1,52 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.api;
+
+public abstract class FontID
```
deathbeam: We use public final class at other places with constants. + This class should have some documentation

runelite-api/src/main/java/net/runelite/api/FontTypeFace.java
```diff
@@ -0,0 +1,32 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.api;
+
+public interface FontTypeFace
```
deathbeam: Same here, this class is missing documentation

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxInput.java
```diff
@@ -0,0 +1,36 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.game.chatbox;
+
+public abstract class ChatboxInput
```
deathbeam: Maybe this class should have some documentation as well, but its not really necessary as it is not in api

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+	private final MouseManager mouseManager;
+
+	private final Provider<ChatboxTextMenuInput> chatboxTextMenuInputProvider;
+	private final Provider<ChatboxTextInput> chatboxTextInputProvider;
+
+	@Getter
+	private ChatboxInput currentInput = null;
+
+	@Inject
+	public ChatboxPanelManager(EventBus eventBus, Client client, ClientThread clientThread,
```
deathbeam: this ctor should be private

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+public class ChatboxPanelManager
+{
+	private final Client client;
+	private final ClientThread clientThread;
+	private final EventBus eventBus;
+
+	private final KeyManager keyManager;
+	private final MouseManager mouseManager;
+
+	private final Provider<ChatboxTextMenuInput> chatboxTextMenuInputProvider;
```
deathbeam: Why are these 2 providers? Afaik only optional dependency we have that requires providers is Client

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
```diff
@@ -0,0 +1,663 @@
...
+	private int fontID = FontID.QUILL_1;
+
+	// This is a lambda so I can have atomic updates for it's captures
+	private ToIntFunction<MouseEvent> getCharOffset = null;
+	private Predicate<MouseEvent> isInBounds = null;
+
+	private boolean built = false;
+
+	@Inject
+	protected ChatboxTextInput(ChatboxPanelManager chatboxPanelManager, ClientThread clientThread)
```
deathbeam: private again

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
Abextm: Because they are not singletons, I need to provide multiple instances

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
Abextm: This class is designed to be subclassed

runelite-api/src/main/java/net/runelite/api/FontID.java
Abextm: IDK what I would document it with other than `Font IDs`. I'm not going to describe each font.

deathbeam: I would probably just mention in what method these FontIDs can be used

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
deathbeam: I do not see where it is subclassed. And if that is the case, maybe it should not use injection

Abextm: It is not subclassed, it is __designed__ to be subclassed if in the future we want more complicated input boxes. I can manually inject it, but that seems several times worse than just letting guice do it.

deathbeam: Yeah, I just dont think that designing classes that are supposed to be created through Guice to be subclassed in future is good idea. The constructor has only like 2 fields anyway, so creating it without Guice might be better.

Abextm: I don't see the advantage to adding redundant boilerplate.

deathbeam: The advantage is that it will not create more technical debt for this issue: https://github.com/runelite/runelite/issues/617

Abextm: I don't see how this has anything to do with #617.

deathbeam: From what I've understood from their wiki pages is that if Guice is managing the class, you are not supposed to call the constructor unless you are unit testing the class

Abextm: Why are you not supposed to?

deathbeam: To prevent accidental misuse of the classes. In this case it is not so easy I guess, but the most common case where people do stupid things with non-private ctors for Guice classes is that they try to create the class by themselves and `@Inject` to it at same time, in what case the injected fields are null.

Abextm: If you have no `@Inject` fields then you can't have that problem. I don't see how calling a `@Inject` constructor is worse than calling the exact same method without `@Inject`.

deathbeam: Yeah it is mostly only problem with `@Inject` fields. I guess this is fine then

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
```diff
@@ -0,0 +1,663 @@
...
+		}
+		return this;
+	}
+
+	public ChatboxTextInput cursorAt(int index)
+	{
+		return cursorAt(index, index);
+	}
+
+	public ChatboxTextInput cursorAt(int s, int e)
```
NathenSample: Could we get real variable names here? Even after looking over the code I'm not 100% certain what on earth is going on here, which would likely be massively remedied by informative names.

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
```diff
@@ -0,0 +1,663 @@
...
+
+		String tsValue = value.toString();
+		isInBounds = ev -> bounds.contains(ev.getPoint());
+		getCharOffset = ev ->
+		{
+			int cx = ev.getX() - canvasX;
+
+			int c = (tsValue.length() * cx) / fullWidth;
+
+			for (; c >= 0 && c <= tsValue.length(); )
```
NathenSample: Is there a rationale for this over a while loop? Besides making me think I'm going insane whilst reviewing it

Abextm: I just always use `for` loops. Its a Go thing.

NathenSample: It feels dirty, whilst I recognised the construct,  you're bastardising a for `loop` into a `while` even though the language has a `while` implementation.

Functionality wise, they're identical, but for the sake of the minds of sweet innocent java developers(and future maintainability) wouldn't it be better to use a while.

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+		this.clientThread = clientThread;
+		this.eventBus = eventBus;
+
+		this.keyManager = keyManager;
+		this.mouseManager = mouseManager;
+
+		this.chatboxTextMenuInputProvider = chatboxTextMenuInputProvider;
+		this.chatboxTextInputProvider = chatboxTextInputProvider;
+
+		eventBus.register(this);
```
deathbeam: You should register this class to EventBus from `RuneLite` like we register every other subscriber

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	public void close()
+	{
+		clientThread.invokeLater(this::unsafeCloseInput);
+	}
+
+	void unsafeCloseInput()
```
deathbeam: maybe both these unsafe close/open methods should be private

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+		{
+			mouseManager.registerMouseListener((MouseListener) input);
+		}
+		if (input instanceof MouseWheelListener)
+		{
+			mouseManager.registerMouseWheelListener((MouseWheelListener) input);
+		}
+
+		currentInput = input;
+		client.setVar(VarClientInt.INPUT_TYPE, -3);
```
deathbeam: What is -3? I believe we have now constant class for input type, probably put this -3 there

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+		return chatboxTextInputProvider.get()
+			.prompt(prompt);
+	}
+
+	@Subscribe
+	public void onScriptCallbackEvent(ScriptCallbackEvent ev)
+	{
+		switch (ev.getEventName())
+		{
+			case "resetChatboxInput":
```
deathbeam: if this switch is supposed to be extendable in future, maybe add break at the end. otherwise, just if (!ev.getEventName().equals(blahblah)) at the top of the method to reduce unnecessary nesting

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+				if (currentInput != null)
```
deathbeam: You should maybe flip this if to reduce nesting

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+				currentInput = null;
+		}
+	}
+
+	public Widget getContainerWidget()
+	{
+		return client.getWidget(WidgetInfo.CHATBOX_CONTAINER);
+	}
+
+	public static void clearDynamicChildren(Widget c)
```
deathbeam: even though this makes sense being static, you should make it method anyway so you would need to inject the manager instead of calling static method on it (we do it in similar way in ItemManager)

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
```diff
@@ -0,0 +1,191 @@
...
+	{
+		if (c.getChildren() != null)
+		{
+			Arrays.fill(c.getChildren(), null);
```
deathbeam: Wouldn't c.setChildren(null) be better?

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
deathbeam: I agree with Nathen, this is not Go and you should use actual Java language features even though they are missing in `Go`

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
Abextm: It may work, this is what the cs2 opcode does to I would prefer to do that. Also this saves a few allocations.

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
Abextm: How does this help? We do this in multiple other places, why is it not ok here?

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextInput.java
Abextm: Well you are in luck, I realized there is an edge case so its a real for loop now 🤷‍♂️

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxPanelManager.java
deathbeam: Its nice to have overview what is registered to EventBus in 1 class, and majority of stuff registered in EventBus is in RuneLite. Also, classes self-registering to EventBus are annoying because you need to actually inject the class somewhere so it's constructor will be called so they will be registered to EventBus.

Adam-:
Please rebase

PR #5887, opened 2018-10-09
Feature: added an extra field to the skill calculator so you can calculate custom xp actions

evdm97:


raiyni:
have a screenshot?

raiyni:
and an issue?

deathbeam:
What is this change trying to achieve? I dont understand how is this useful at all

evdm97:
[It does with super glass make](https://oldschool.runescape.wiki/w/Molten_glass) but that's besides the point. 

The skill calculator can't possibly cover every edge case xp drop and this allows people to calculate things that aren't on there.

Is it the feature you don't like or the form it takes. I could try and think of a different appearance that doesn't leave that extra field just dangling below.

raiyni:
To be honest, the xp remaining pr will probably make this near useless.

evdm97:
> 
> 
> To be honest, the xp remaining pr will probably make this near useless.

How so? Wouldn't a user still have to grab the calculator themselves if they knew the xp remaining. Isn't the point of the tab to have it calculated for you

ollieatkinson:
Maybe it's better to add "Goal"'s to the 'XP tracker' plugin - which can show actions, experience and time left.

PR #5888, opened 2018-10-09
Add "rescale to multiple" config option

widget-:
Adds option next to client dimensions to scale to a particular multiple of the base client size.

Mostly this is because making stretched fixed mode look nice is rather difficult, as you have to calculate the dimensions want every time.

The intent was to have it also update the dimensions in the panel but it looks like RuneLite can't do that. The values are still set, and will be visible when you navigate back to the list of plugins and back into the preferences though.


ollieatkinson:
So setting this makes the client ignore the game size defined in the option above? It might be a little confusing. 

Is there a way we can make this clearer? Because I really like the feature.

widget-:
It overwrites it but the value isn't visually updated until you make it refresh the panel. If there's a way to do that I'd like to add it too.

Issue #3395, opened 2018-05-27
Multiple coloured ground markers


Allemaalbaas:
Hi, I want to suggest multiple coloured ground tiles. How it would work is that you have a slider or checkbox to choose of you want to change all ground tiles to this colour or just one. This way a ground tile  would automatically be another colour if you change the colour (with the colour picker). If you would slide or check the box it would change all the tiles to the chosen colour.

ElevatedLife:
Lots of support for this feature.

https://github.com/runelite/runelite/issues/5812#issue-366864846
https://github.com/runelite/runelite/issues/5187
https://github.com/runelite/runelite/issues/4242

Adam-:
Reopening as the merged PR could not stabilize prior to release.

eggbrow:
> I'm starting work on this, but it'll probably be a little while before I have this ready for review.
How likely is it that there'll be support for different-coloured unique object markers and npc indicators?

Issue #4157, opened 2018-07-03
Use consistent distance cap for overlay rendering across plugins


Nightfirecat:
**Is your feature request related to a problem? Please describe.**
When writing an overlay which can render tiles/actors/objects at a distance, it should only render that overlay when a clickbox is available. This is particularly relevant for GPU plugin users, as it is frequently possible for an actor or object to have a clickbox generated for it, but not have an overlay drawn for it, or vice versa if using hard-coded LocalPoint or WorldPoint distance caps since clickbox rendering is dependent on camera position, not player position.

Nightfirecat:
Updated issue title/description.

Nightfirecat:
Alternately, as has been mentioned in Discord before, we could just do away with these MAX_DISTANCE constants because they're not really useful. :man_shrugging:

dekvall:
This seems like a good idea.

Nightfirecat:
Following some discussion in Discord, I've updated this enhancement request to be more in line with an ideal solution.

abextm:
imo having a `Client::get(Min|Max)LocalInteraction[XY]` + `Client::isInteractable(LocalPoint)` would be a good api for this

PR #5890, opened 2018-10-09
Add support for colored ground markers

Nightfirecat:
This adds a color indicator per marker, saved to config, allowing multiple markers to be of different colors.

This also adds a deprecated configuration migration method which will convert any existing markers to the updated schema, setting their colors to be the current configuration color.

ground markers usage demonstration

rainbow ground markers

Fixes runelite/runelite#3395


Adam-:
This needs rebase for #7788

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
```diff
@@ -69,15 +71,16 @@
...
-	private static final Gson gson = new Gson();
+	private static final Gson GSON = new Gson();
 
 	@Getter(AccessLevel.PACKAGE)
 	@Setter(AccessLevel.PACKAGE)
 	private boolean hotKeyPressed;
 
 	@Getter(AccessLevel.PACKAGE)
-	private final List<WorldPoint> points = new ArrayList<>();
+	private final Map<GroundMarkerPoint, Collection<WorldPoint>> points = new HashMap<>();
```
Adam-: This should probably instead just be a list of some private class which contains scene x/y/z and Color

runelite-api/src/main/java/net/runelite/api/coords/WorldPoint.java
```diff
@@ -191,6 +195,58 @@ public static WorldPoint fromLocalInstance(Client client, LocalPoint localPoint)
...
+	 *
+	 * @param client  The client, used to determine whether it is running in an instanced Scene
+	 * @param sceneId The ID of the Scene
+	 * @param sceneX  The X point within the Scene
+	 * @param sceneY  The Y point within the Scene
+	 * @param z       The Z point within the Scene
+	 * @return A {@link Collection} of {@link WorldPoint}s which can correspond to the given
+	 *         {@link net.runelite.api.Scene} values. Should return a singleton list when run outside of an instance.
+	 */
+	public static Collection<WorldPoint> fromScene(final Client client, final int sceneId, final int sceneX, final int sceneY, final int z)
```
Adam-: Use the method added in #7788 instead. Also `sceneId` isn't a thing we have defined anywhere.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
```diff
@@ -314,21 +228,63 @@ protected void markTile(LocalPoint localPoint)
...
 	}
-}
\ No newline at end of file
+
+	/**
+	 * Updates configuration data for ground markers to store color information.
+	 * This method should be removed after a reasonable amount of time.
+	 */
+	@Deprecated
+	void setColorConfiguration()
```
Adam-: I don't see any need for this at all. If it has no color then it has no color, so use whatever is configured.

Adam-:
I added this to the 1.5.13 milestone since it apparently is getting some [use](https://www.reddit.com/r/2007scape/comments/ard417/xzacts_new_strategy_to_dodge_zuk_healer_damage/egmfz8q/).

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
Nightfirecat: My thought process behind this is that that this PR removes the ability to have existing markers change upon configured plugin color changing, so it would be kind of weird for markers existing before this is merged to change when any new ones would be of a fixed color. I can change it like you requested anyway, but it just seems weird to introduce inconsistent behavior that way.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
Nightfirecat: At that point, it seems easier to add serialization for worldpoints to configmanager and use a worldpoint directly since that removes the need to handle scene logic within the plugin. At this point, I'm inclined to rewrite GroundMarkerPoint as a wrapper for a worldpoint and color and go from there.

Adam-: I think it would be better to use a separate class.

Nightfirecat: I think we're talking about the same thing. Let me know if what I'm doing now is different than what you had in mind.

Nightfirecat:
@deathbeam I've added a toggle to switch between the new behavior and the old (existing) behavior.

deathbeam:
@Nightfirecat thanks

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerOverlay.java
```diff
@@ -42,6 +42,8 @@
 
 public class GroundMarkerOverlay extends Overlay
 {
+	private static final int MAX_DRAW_DISTANCE = 32;
```
deathbeam: I am pretty sure we have max distance defined in constants, so if you are already changing this, you should maybe use that.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
```diff
@@ -69,8 +69,9 @@
 	private static final String CONFIG_GROUP = "groundMarker";
 	private static final String MARK = "Mark tile";
 	private static final String WALK_HERE = "Walk here";
+	private static final String REGION_PREFIX = "region_";
 
-	private static final Gson gson = new Gson();
+	private static final Gson GSON = new Gson();
```
deathbeam: Maybe use RuneLiteAPI.GSON (I think its even injectable via Guice, we should maybe start migrating to that)

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
deathbeam: Imo this is not necessary too, or it can be done in better way? E.g on point loading, check if point color is null and if yes, set it to current color and save. Would be a lot simpler than this "abomination" 😄

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/OldGroundMarkerPoint.java
```diff
@@ -0,0 +1,42 @@
...
+
+import lombok.Value;
+
+/**
+ * Used for serialization of old ground marker points.
+ * This class should be removed after a reasonable amount of time.
+ */
+@Deprecated
+@Value
+class OldGroundMarkerPoint
```
deathbeam: You dont need separate class to just not load optional field, it will be just null in guice

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPoint.java
```diff
@@ -25,13 +24,18 @@
...
+import net.runelite.api.coords.WorldPoint;
 
 @Value
 class GroundMarkerPoint
 {
-	private int regionId;
-	private int regionX;
-	private int regionY;
-	private int z;
+	private WorldPoint worldPoint;
```
deathbeam: There is no reason for this worldPoint change, so unless its absolute necessary, either create your own serializer/deserializer for Guice or simply dont do this.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerOverlay.java
Nightfirecat: Doesn't look like it. [Constants](https://github.com/runelite/runelite/blob/master/runelite-api/src/main/java/net/runelite/api/Constants.java) doesn't have anything like this, and #4157 hasn't been closed so I don't think that's the case.

deathbeam: Ah yeah I guess I never PRed that after the agility plugin revert

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
Nightfirecat: Based on Discord discussion, I'll pass on doing that for this PR. We might look at making a plugin-specific GSON instance later on for use among plugins for config serialization.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
Nightfirecat: I got myself confused on separating a private encapsulation of `WorldPoint` and `Color` and thought at the time that migrating configs to use worldpoint would be the best solution there. Your suggestion makes sense, and that's what I've done in my recent push--serialization is no longer converted and plugin logic now uses the new `ColorTileMarker` encapsulation for rendering.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPoint.java
Nightfirecat: I've moved encapsulation of `WorldPoint` + `Color` to a separate class and left this alone for serialization.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerConfig.java
```diff
@@ -45,4 +44,14 @@ default Color markerColor()
 	{
 		return Color.YELLOW;
 	}
+
+	@ConfigItem(
+		keyName = "usePerTileColors",
+		name = "Use per-tile colors",
+		description = "Color tiles using the color from time of placement"
+	)
+	default boolean rememberTileColors()
```
deathbeam: This should be defaulted to false, imo.

Nightfirecat: I disagree, but I'm not going to revert your change.

deathbeam:
Fixed small issue:

```diff
-                       points.add(point);
+
+                       // Add point back only if we are remembering tile colors, otherwise simply remove it
+                       if (config.rememberTileColors())
+                       {
+                               points.add(point);
+                       }
```

Nightfirecat:
Mentioned in discord, but that "small issue fix" doesn't do what you intended, it just blocks people from marking new tiles when they have that option disabled.

Issue #2011, opened 2018-04-26
Add Freeze time on the opponents


mrpker9:
Is it possible to create a freeze time on the opponent so when someone gets binded or frozen it shows on him how many seconds left till he can move?

devOSRS:
Are opponent freeze/bind timers are off the table? Don't want to pursue a feature if there are issues with the feature being added to runelite. (i.e.: https://github.com/runelite/runelite/pull/604)

deathbeam:
https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features Yeah, they are on rejected list now based on previous discussions we had about them.

TwiNaga:
Since unfair advantages when PKing seems to be the main concern, would it be possible to have this added only for NPCs?  It would greatly help with bosses such as Scorpia.

PR #5894, opened 2018-10-10
Opponent freezes

devOSRS:
Added freeze timers similar to Exilent. This is a very rough PR for now. Looking for any constructive feedback.

For example, thinking of maybe using a Guava Cache instead of a map, making more efficient use of FreezeInfo class, better overlay design, etc.

I find this feature extremely useful and it's been heavily asked for! Hopefully we can work together.
Thanks!


devOSRS:
ref: https://github.com/runelite/runelite/issues/2011

devOSRS:
Very cool!
freeze-timers

NathenSample:
Freeze timers are in the rejected list.
See: https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

BrianJMRogers:
@devOSRS what ever happened to this? I see it in a list of "not being considered" but to me that doesn't mean it can't be added.

Trevor159:
this falls under `Indicates how long an opponent is frozen for` in https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1

PR #5898, opened 2018-10-10
Implement locking and corrupted file sanitization to ConfigManager

### LINKED ISSUE #4853: Client settings reset after power outage shut down pc
The streamer Sick Nerd mentioned earlier on his stream that his RuneLite settings were reset after a power outage made his pc shut down. He wasn't willing to check the settings file, so I don't know if it was the usual corruption of the settings file, but it might have something to do with peoples settings getting reset?

deathbeam:
## Remove invalid properties from map on property load

In order to not have config file flooded with invalid and corrupted
properties in case of file corruption, remove all these invalid
properties from property map on configuration load.

## Use UTF-8 instead of ISO 8859-1 when loading/saving config

In order to prevent failures on reading/writing incorrectly escaped
characters or writing special UTF-8 characters on save as gibberish,
explicitly specify UTF-8 encoding when loading and writing properties.

## Lock configuration file when writing to filesystem

Use OS-wide exclusive lock when writing configuration file to filesystem
to prevent config file corruption when multiple concurrent writers try
to write to same config file at same time.

Closes #4853

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
@@ -41,6 +41,7 @@ import java.io.OutputStreamWriter;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Proxy;
+import java.nio.channels.FileLock;
 import java.nio.charset.Charset;
 import java.time.Instant;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -179,11 +183,11 @@ private synchronized void loadFromFile()
 
 		try (FileInputStream in = new FileInputStream(propertiesFile))
 		{
-			properties.load(in);
+			properties.load(new InputStreamReader(in, Charset.forName("UTF-8")));
 		}
 		catch (FileNotFoundException ex)
 		{
-			log.debug("Unable to load settings - no such file");
+			log.debug("Unable to load settings - no such file", ex);
```
Adam-: exception here was intentionally not logged since this is part of normal operation

PR #5901, opened 2018-10-10
Option to show username when chat input is locked (by WASDCamera plugin).

Karn:
**Proposed Changes**:
- Add a toggle to prefix the chat prompt with the username when the chat input is disabled.

**Assets**:
Additional config entry to toggle username (disabled by default).
screen shot 2018-10-10 at 6 53 14 am

Expected result:
screen shot 2018-10-10 at 6 56 01 am

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
@@ -28,6 +28,7 @@ package net.runelite.client.plugins.wasdcamera;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import java.awt.Color;
+import javax.annotation.Nonnull;
 import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
@@ -36,6 +37,7 @@ import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.VarClientInt;
 import net.runelite.api.VarClientStr;
+import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.ScriptCallbackEvent;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetInfo;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
```diff
@@ -182,4 +198,20 @@ void unlockChat()
...
 		}
 	}
+
+    /**
+     * Helper function used to update the chatbox input field once the chat
+     * has been locked.
+     *
+     * @param chatboxInput Non-null chatbox input widget.
+     */
+	private void hideChatInput(@Nonnull Widget chatboxInput) {
```
forsco: new line {

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
```diff
@@ -182,4 +198,20 @@ void unlockChat()
...
+		String prefix = "";
+
+		if (config.showUsername()) {
```
forsco: new line {

forsco:
Checkstyle: Indents must use tabs, not spaces

Karn:
@forsco Comments have been addressed. You know what they say about old habits. 😓

deathbeam:
I am not sure if this even needs a toggle

widget-:
I personally like having it hidden. I could see the option going somewhere else though.

Karn:
I can do one of two things that I think might work:
1. Remove the option and make the username prefix default.
2. Default the `show username` value to `true`

I'd prefer the first option as there is less code to maintain.

raiyni:
It would go in the streamer mode plugin.

deathbeam:
Yeah idk what is best, how was it in rs2?

Karn:
The username was always included from what I remember. Circa 2010?

Karn:
@raiyni Perfect! In that case, I'll leave it toggleable for now.
@deathbeam Are you okay with leaving it as disabled by default?

Karn:
The issue is that the chat input locking functionality is bundled with the WASDCamera plugin. Maybe then it makes sense instead to modify the WASDCamera plugin to instead be a subset of a `HotKey` plugin, there you could map the keys to the functionality. E.g WASD for Camera, 0-9 for bank tabs, etc.

This decouples the chat input locking from the camera and leaves the ability to extend keys presses. If we then introduce a streamer plugin, the username should be prefixed by default and only hidden when the streamer plugin is enabled.

cc: @deathbeam

Adam-:
Sorry! Didn't realize this was here.

PR #5908, opened 2018-10-10
Use TAG_SEARCH const instead of hardcoding "tag:" in Bank Tags

deathbeam:


runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -243,10 +244,9 @@ public void update()
 
 		str = Text.standardize(str);
 
-		if (str.startsWith("tag:"))
+		if (str.startsWith(BankTagsPlugin.TAG_SEARCH))
```
raiyni: Don't need to reference the class

PR #5910, opened 2018-10-10
Don't limit bank search and title explicitly to 15 characters

raiyni:
Before

After


Adam-:
Definitely comment up the rs2asm extensively since they are generally hard to follow and also we need to be able to identify what it is we changed since this would get missed.

PR #5921, opened 2018-10-11
Add support for Kotlin

DylanLange:
Hey everyone!
My apologies if this is the wrong place to do this. I'm quite new here and new to open source contributions in general.
I have been looking around for usages or even mentions of Kotlin in this project, and haven't seemed to find anything about it so I thought I'd just go ahead and create this PR.

I come from an Android development background and I'm a huge advocate for using Kotlin in general.

For those who don't know, Kotlin is a not-so-new language that has recently had a huge boom in the Android development community mainly for the amazing benefits it provides over Java **_(whilst still being fully interoperable! (See [here](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html) and [here](https://kotlinlang.org/docs/reference/java-interop.html)))_** I won't go in to details of them but I'm certain if you do a quick Google search you'll find plenty of examples and explanations.

Anyways I'd love to help contribute to RuneLite, but I definitely saw an opportunity here to have a prod and see what people think about supporting Kotlin development in this project.

**For a quick look at how they compare visually:**
**Java**:
image
image

**Kotlin**:
image

These screenshots are really just for an idea of how similar they are rather than to highlight cases where Kotlin really shines. 

[Here's a list of points of potential interest.](https://kotlinlang.org/docs/reference/comparison-to-java.html)

Thanks for your time! Looking forward to seeing what you guys think or what questions you might have.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,13 +193,17 @@
 					<execution>
 						<id>compile</id>
 						<phase>compile</phase>
-						<goals> <goal>compile</goal> </goals>
+						<goals>
+							<goal>compile</goal>
+						</goals>
 					</execution>
 
 					<execution>
 						<id>test-compile</id>
 						<phase>test-compile</phase>
-						<goals> <goal>test-compile</goal> </goals>
+						<goals>
+							<goal>test-compile</goal>
+						</goals>
 					</execution>
 				</executions>
 			</plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,8 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>1.8</java.version>
 		<lombok.version>1.16.22</lombok.version>
+		<kotlin.version>1.2.71</kotlin.version>
+		<kotlin.compiler.incremental>true</kotlin.compiler.incremental>
 
 		<maven.javadoc.skip>true</maven.javadoc.skip>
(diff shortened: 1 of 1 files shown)


Adam-:
My opinion on this is it adds more trouble than it is worth. The client is currently delivered using many Java-isms that would quickly become not-great Kotlin-isms and that would just be a disaster to fix. Also this would probably bloat up the runtime considerably.

DylanLange:
Fair call - it would definitely throw a spanner in the mix. But supporting use of Kotlin doesn't mean best practice would have to go out the window - it would just require a bit more learning to see how things should be written to remain elegant in itself and also with Java interop'. I'm happy to provide as much as I can on how exactly things work together going between Java source and Kotlin source if that is desired :)

Also I think you're right about there being some unforeseen consequence on the runtime adding this in. I'd be happy to run some tests to see if they are considerable.

jlmcdonnell:
The good thing about Kotlin is you don't need to write 'idiotmatic Kotlin' to get a lot of its benefits. In terms of runtime performance, Kotlin compiles to bytecode and runs on the JVM just like Java so there would not be an impact on that front. I also come from an Android background and have worked on plenty of projects that share both Kotlin and Java (you do not need to convert everything to Kotlin, as both languages are interoperable!).

deathbeam:
As Adam said. We dont need Kotlin and most of contributors are Java devs, and mixed code is usually horrible, no matter how good Kotlin is or is not.

PR #5922, opened 2018-10-11
menuentryswapper: Disable swapping on clue scroll target NPCs

Abextm:
I'm sick of `travel`ing rather than `talk`ing to the NPC for my clue.


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollService.java
```diff
@@ -0,0 +1,41 @@
...
+import net.runelite.api.NPC;
+
+public class ClueScrollService
+{
+	@Inject
+	private ClueScrollPlugin plugin;
+
+	public List<NPC> getNpcsToMark()
+	{
+		return plugin.getImmutableNpcsToMark();
```
deathbeam: You can just do Collections.immutableList(plugin.getNpcsToMark()) here instead of adding new getter

Abextm: This is to not alloc in a getter.

deathbeam: Isnt immutable list just view to original one? should not be very costly and I would personally prefer that small alloc than new field in plugin who do not needs it at all just to avoid it. Or alternatively you can move the pre-allocated field here.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollService.java
```diff
@@ -0,0 +1,41 @@
...
+package net.runelite.client.plugins.cluescrolls;
+
+import com.google.inject.Inject;
+import java.util.List;
+import net.runelite.api.NPC;
+
+public class ClueScrollService
+{
+	@Inject
+	private ClueScrollPlugin plugin;
```
deathbeam: Use constructor injector with private ctor

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollService.java
Abextm: IMO any method labed `get` should be a) very cheap to call and b) pure. `Collections.immutableList` is not pure

deathbeam: Well then move the field to the ClueScrollService class as it is relevant only here

deathbeam:
Needs rebase

deathbeam:
Implemented in: <https://github.com/runelite/runelite/pull/6437> what was more straightforward solution without any dependency

PR #5931, opened 2018-10-11
Update to BA plugin:

Jacoblairm:
- Fixed this: https://i.imgur.com/APUYSyL.jpg by nulling the GameTimer when the game is complete.
- Added an option (by popular demand) that shows the round time (in seconds) instead of the call change time (off by default).
- Fix a rounding issue with the timer, by rounding Duration.toMillis() instead of just using the Duration.getSeconds() (which was always rounded down)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultConfig.java
@@ -43,6 +43,16 @@ public interface BarbarianAssaultConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "secondsTimer",
+		name = "Show elapsed seconds",
+		description = "Instead of showing the call change timer, the wave elapsed time will be shown (in seconds)"
+	)
+	default boolean secondsTimer()
+	{
+		return false;
+	}
+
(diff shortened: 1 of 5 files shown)


deathbeam:
Maybe you should 1. split commits to commit fixing bug and commit adding new feature, 2. display both timers and not one or another if both are enabled?

Jacoblairm:
So having them side by side if they're both enabled like this  -  '00:12 (18') or '(18) 00:12' ?

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultConfig.java
```diff
@@ -42,6 +42,16 @@ default boolean showTimer()
 		return true;
 	}
 
+	@ConfigItem(
+		keyName = "secondsTimer",
+		name = "Show elapsed seconds",
+		description = "Instead of showing the call change timer, the wave elapsed time will be shown (in seconds)"
```
MagicfTail: This description needs to be changed now that you no longer remove the other timer

Jacoblairm: Ah yea good point

PR #5939, opened 2018-10-11
Zoom fix

MagicfTail:
- Remove now useless outer zoom limit expansion
- Change exponent to make zoom feel better after extra 'notch' was added
- Put fixed and resizable zoom together since they now use the same scale
- Change script to correctly fetch range
- Fix control zoom


MagicfTail:
That extra zoom bit was basically nothing, having an option to zoom out that much extra would be a joke

Adam-:
What about a setting for restoring the old zoom limit? There has been several requests for this.

Aborts:
add our own default camera limit 💯💯💯

MagicfTail:
I was going to add that, but I feel like it should be a separate pr since this is focused on fixing stuff, rather than adding new features

deathbeam:
I think making presets or something like someone suggested way before this zoom limit fix would make more sense anyway

Aborts:
i agree with deathbeam 💯

PR #5951, opened 2018-10-11
Add option to remove attack option when warned style is selected

Reasel:
This will add an option in the attack styles plugin to remove the option to attack anything when the warned attack style is selected.

demo


Nightfirecat:
What's the use case for this? It seems like simply removing the widget to select warned styles should always be enough to prevent unwanted attacks with warned styles.

deathbeam:
Sometimes when you switch weapons the selected style can be changed.

Reasel:
@Nightfirecat The style option is the same when you switch weapons so for example I no longer wish to train strength post 99 and it just so happens that rapid on a rune cbow is the same slot as slash...so if I switch to my cbow and choose rapid then switch to a whip it will be on strength. I just wanted to not have the attack option and instead be forced to switch.

PR #5959, opened 2018-10-12
Add support for mouse tooltip icons

deathbeam:
- Add support for setting image to Tooltip and rendering it before text.
- Add support for showing item image tooltips next to text tooltip to
MouseHighlightingPlugin (disabled by default)

Preview:
peek 2018-10-13 01-41


deathbeam:
On second thought, I actually dont like this feature at all :D

PR #5960, opened 2018-10-12
Fix cave kraken boss task name

### LINKED ISSUE #5958: Small bug with Kraken boss task counter and no icon
image

image

image

deathbeam:
Closes #5958


ollieatkinson:
Actually, Will this affect normal Kraken tasks?

deathbeam:
Nope

Nightfirecat:
This should've been added as another test case imo. We should probably just add a new test case whenever we need to update a parsing regex.

PR #5967, opened 2018-10-12
GameEventManager: Simulate Actor spawn events

Nightfirecat:
I don't think this will affect many (if any) plugins currently, but it will be useful for some work I'll soon be submitting to remove deprecated `Query` usage.


deathbeam:
You dont need to filter the location, its just all players in scene, also, you can just use client.getCachedNpcs and .getCachedPlayers to avoid list allocation (and filter for nonnull)

Issue #3767, opened 2018-06-13
Bank tags for charged items


rsekelsky:
I think it would be helpful if bank tags for charged items like enchanted jewelry would carry through their charges, i.e. if you tag a Ring of dueling (8) with tag "x" then all items Ring of dueling (y) with 1 <= y <= 8 show up with tag "x".

I'm not familiar with how Runelite is coded; in fact, I'm quite clueless on its inner workings in general. So I'm not sure if this is even possible, but the example I showed above would be how I imagine this feature could be implemented. If a tagged item ends with (x) for x an integer, then tag every item with the same name but x-1, x-2, ..., 1 at the end. In the case of re-enchantable items (like dragonstone jewelry, elf crystal, etc.) it may be tougher to tag the 0 charge item. Perhaps a blanket item for all charged item types can help this, like the same method I just described, but also tag the item without the (x) at the end.
So overall:

Ex: tag "amulet of glory(6)" with tag "x": tags "amulet of glory(6)", "amulet of glory(5)", ... "amulet of glory(1)", and "amulet of glory" with "x".

Cheers, and thanks for considering this!
RS

EDIT: This could also work for potions as well. I feel this would be best as a toggle, as a lot of people like to only see 4-dose (and similarly, full-charged jewelry) as their tagged item.

PR #5972, opened 2018-10-12
Bank tags: tag charge/degradable items

raiyni:
Holding shift adds the ability to edit variation tags on an item.

peek 2018-10-30 09-04

Closes https://github.com/runelite/runelite/issues/3767


runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -58,6 +60,18 @@
 @Slf4j
 public class ItemManager
 {
+	private static final List<Pattern> normalizePatterns = new ArrayList<Pattern>()
+	{{
+		// cakes, pizzas, pies
+		add(Pattern.compile("^(?:[0-9]/[0-9]|half an?) ([^(0-9]+)|$"));
```
Nightfirecat: [0-9]/[0-9] could probably be [1-9]/[1-9]? There's no items that are "0/3" for instance, and I doubt there will ever be 0 in the denominator either.

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -58,6 +60,18 @@
 @Slf4j
 public class ItemManager
 {
+	private static final List<Pattern> normalizePatterns = new ArrayList<Pattern>()
```
Nightfirecat: Is there any way we can run a test for this? Seems like we'd want to make sure this doesn't break with future updates.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -307,4 +332,28 @@ public MouseWheelEvent mouseWheelMoved(MouseWheelEvent event)
 		clientThread.invokeLater(() -> tabInterface.handleWheel(event));
 		return event;
 	}
+
+	@Override
+	public void keyTyped(KeyEvent e)
+	{
+
```
Nightfirecat: Remove this empty line I guess

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
deathbeam: also, UPPER_CASE

raiyni:
I'm writing tests

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -58,6 +60,18 @@
 @Slf4j
 public class ItemManager
 {
+	private static final List<Pattern> UNCHARGE_PATTERNS = new ArrayList<Pattern>()
```
deathbeam: ImmutableList.builder

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -108,6 +112,11 @@
 	@Inject
 	private TabInterface tabInterface;
 
+	@Inject
+	private KeyManager keyManager;
+
+	private boolean shiftPressed = false;
```
deathbeam: You should reset shiftPressed on focus lost and shutdown

runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
```diff
@@ -85,20 +131,48 @@ public void addTag(int itemId, String tag)
...
+		return tags.stream().anyMatch(tag -> tag.contains(Text.standardize(search)));
 	}
 
 	public void removeTag(String tag)
 	{
 		final String prefix = CONFIG_GROUP + "." + ITEM_KEY_PREFIX;
 		configManager.getConfigurationKeys(prefix).forEach(item -> removeTag(Integer.parseInt(item.replace(prefix, "")), tag));
+
+		final String chargePrefix = CONFIG_GROUP + "." + VARIATION_KEY_PREFIX;
+		configManager.getConfigurationKeys(prefix).forEach(item -> removeVariationTag(Integer.parseInt(item.replace(chargePrefix, "")), tag));
```
deathbeam: you can get the configuration keys only once and reuse

deathbeam: or, do getConfigurationKeys(chargePrefix) as that looks like correct behaviour here

raiyni: yeah it's a typo

raiyni:
now with passing tests

raiyni:
Rebased to use an enum

Nightfirecat:
So the drag-onto-tab behavior for item variations is only changed when holding shift, right? I don't know that I'd often want to tag variations this way so I'd be against the default behavior changing.

raiyni:
Yes, only when holding shift

deathbeam:
I am afraid this will not work with import/export very well and that I cant really make it work reliably without the exported string being some insanity and the website handling code as well

raiyni:
Updated to use current config tag key. Should be compatible with import/export. Up to the website to display it right.

deathbeam:
Updated the PR, as I said in Discord:

```
Updated the variation tag PR, it should be a bit cleaner now. Also, made remove-tag remove both variation tags and normal tags, because otherwise it will not work (e.g you have item tagged only with variation tag but you try to remove it as regular tag, nothing happens)
```

deathbeam:
Updated the PR (unified edit-tags and edit-variation-tags) it works like this now:

peek 2018-10-30 09-04

PR #5987, opened 2018-10-13
Exclude boss drops from chat notification

SirGirion:
This adds an option to not be notified of boss drops while allowing notifications of your own name.


deathbeam:
Maybe adding option to skip game messages would be better than simply skipping something based on some text matching?

SirGirion:
The trade and duel notifications both utilize text matching as an additional check for skipping. I'm not opposed to skipping game messages but this feature was intended to only hide the drop notifications while retaining notifications from your name.

PR #5988, opened 2018-10-13
Bank tags: add ability to tag inventory/equipment

raiyni:


Kamielvf:
Any reason why the tag-equipment option is not on the deposit equipment button instead?

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -383,6 +445,60 @@ else if (activeTab != null
 			tagManager.removeTag(itemId, activeTab.getTag());
 			doSearch(InputType.SEARCH, TAG_SEARCH + activeTab.getTag());
 		}
+		else if (activeTab != null
```
deathbeam: you should probably merge these 2 ifs, and do the activeTab difference inside (also you could reduce amount of code like this, because you can just build the list, and use .addTags() in both cases

raiyni: One opens the chatbox, the other doesn't

deathbeam:
Cleaned this up

deathbeam:
Fixed some bugs and cleaned up the code

PR #6001, opened 2018-10-13
Log error when encountering an unmatched clue string

NathenSample:
Given the recent, and ongoing problems with clue text, this adds an error when we fail to match against any known clue.

This means in scenarios where strange formatting on jagex's side is to blame, the user can provide the needed information on how the string is formatted, rather than a developer having to run clues until they're able to reproduce it.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -477,6 +477,7 @@ public class ClueScrollPlugin extends Plugin
 		}
 
 		// We have unknown clue, reset
+		log.warn("Encountered unhandled clue text: {}", clueScrollText.getText());
 		resetClue(true);
 		return null;
 	}


deathbeam:
You should at least use log.warn

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -429,7 +431,7 @@ public static CrypticClue forText(String text)
 				return clue;
 			}
 		}
-
+		log.warn("Clue found with unmatchable text: {}", text);
```
deathbeam: Would this be just spam in log in case the clue is actually not cryptic clue but some other one?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/ThreeStepCrypticClue.java
```diff
@@ -177,10 +180,14 @@ public static ThreeStepCrypticClue forText(String plainText, String text)
 						// chop off the length of the clue text plus a space
 						threeStepText.delete(0, clue.getText().length() + 1);
 					}
-
+					clueMatched = true;
 					break;
 				}
 			}
+			if (!clueMatched)
```
deathbeam: you can probably just do this in steps.empty or even from ClueScrollPlugin, where resetClue is called in the widget check, that would also solve the other cryptic clue issue

NathenSample:
@deathbeam yeah, when not trying to do this at silly a'clock in the morning I looked into the actual logic the plugin uses, which would of had... irritating effects. 

I've removed the calls from the specific classes and inserted a generic message into `ClueScrollPlugin`

PR #6007, opened 2018-10-14
Remove switch case for swapQuick

forsco:
Remove the switch case for swapQuick as any swapping that comes later in the else if chain is broken


Adam-:
Maybe we should start adding some tests for this plugin?

deathbeam:
I would rather spent energy on making the plugin less bad

Nightfirecat:
Why not both?

PR #6030, opened 2018-10-15
Fix boss ENUM for Special Counter

alexbowers:


Alexsuperfly:
i would rather that be refactored to use IDs instead of names, to be more future proof

alexbowers:
How does one go about obtaining the IDs?

Alexsuperfly:
theres a NPCid file in the codebase

MagicfTail:
The boss timer plugin should have all the correct ones

alexbowers:
Okay, I'll update this PR to do it properly.

deathbeam:
I opened #6031 that handles the NPCIds properly (and fixes stuff like vetion and KQ special attack counters)

Issue #5956, opened 2018-10-12
Notifier is not calling .waitFor before .exitCode() check, causing .exitCode() to always fail


deathbeam:
```
2018-10-12 10:21:04 [pool-2-thread-1] WARN  n.r.c.util.RunnableExceptionLogger - Uncaught exception in runnable net.runelite.client.Notifier$$Lambda$418/1163862728@7733819c
java.lang.IllegalThreadStateException: process hasn't exited
        at java.lang.UNIXProcess.exitValue(UNIXProcess.java:423)
        at net.runelite.client.Notifier.lambda$null$0(Notifier.java:212)
        at java.util.Optional.map(Optional.java:215)
        at net.runelite.client.Notifier.lambda$sendLinuxNotification$1(Notifier.java:212)
        at net.runelite.client.util.RunnableExceptionLogger.run(RunnableExceptionLogger.java:41)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

marek-urban:
Greetings.
I would like to be involved in fixing this issue.

Adam-:
Fixed in 431713b9a4e90bfb4a2424b702ed7a09051ddc15

PR #6034, opened 2018-10-15
Fix of bug #5956 : Notification in linux OS cause an error

marek-urban:
Filter out alive processes before mapping in lambda exp.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -210,11 +210,17 @@ public class Notifier
 		{
 			try
 			{
-				sendCommand(commands, 500);
+				Process notificationProcess = sendCommand(commands);
+
+				boolean exited = notificationProcess.waitFor(500, TimeUnit.MILLISECONDS);
+				if (exited && notificationProcess.exitValue() == 0)
+				{
+					return;
+				}
 			}
 			catch (IOException | InterruptedException ex)
 			{
-				log.warn("Error sending Linux notification", ex);
+				log.warn("Error sending linux notification.", ex);
 			}
 
 			sendTrayNotification(title, message, type);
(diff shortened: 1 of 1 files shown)


Adam-:
Shouldn't this wait for alive processes?

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -209,6 +209,7 @@ private void sendLinuxNotification(
 		executorService.submit(() ->
 		{
 			final boolean success = sendCommand(commands)
+					.filter(process -> !process.isAlive())
```
deathbeam: Wouldn't this case success to be always false? As this will never wait for process to be terminated (e.g .waitFor() needs to be used before exitValue check)

marek-urban: Ok I've changed the code according to your instructions.
I've tested my first solution as well and it worked. But you are probably right and calling .waitFor() is more safe.

marek-urban:
> Shouldn't this wait for alive processes?
Code changed.

Adam-:
Probably part of the confusion here is `sendCommand` is over-engineered and using language features Because They Exist and not because they are particularly helpful. If you change it to:

```java
private Process sendCommand(final List<String> commands) throws IOException
{
	return new ProcessBuilder(commands.toArray(new String[commands.size()]))
		.redirectErrorStream(true)
		.start();
}
```

Then this code nicely becomes:

```java
try
{
	Process notificationProcess = sendCommand(commands);

	boolean exited = notificationProcess.waitFor(500, TimeUnit.MILLISECONDS);
	if (!exited || notificationProcess.exitValue() != 0)
	{
		sendTrayNotification(title, message, type);
	}
}
catch (IOException | InterruptedException ex)
{
	log.warn("error running notify-send", ex);
}
```

and then try/catch/log the `sendCommand` in `sendMacNotification`.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -208,14 +208,16 @@ private void sendLinuxNotification(
...
 		{
-			final boolean success = sendCommand(commands)
-					.map(process -> process.exitValue() == 0)
-					.orElse(false);
-
-			if (!success)
+			try
 			{
-				sendTrayNotification(title, message, type);
+				sendCommand(commands, 500);
```
Adam-: This is wrong because it will always fall back to the tray notif. regardless of it the command succeeds.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -248,23 +250,28 @@ private void sendMacNotification(
...
+	private void sendCommand(final List<String> commands, final int timeout) throws IOException, InterruptedException
+	{
+		Process notificationProcess = new ProcessBuilder(commands.toArray(new String[0]))
+				.redirectErrorStream(true)
+				.start();
+
+		if (!notificationProcess.waitFor(timeout, TimeUnit.MILLISECONDS)
+				|| notificationProcess.exitValue() != 0)
+		{
+			throw new InterruptedException();
```
Adam-: This is just abuse of `InterruptedException`

marek-urban:
Done.

Adam-:
Fixed in 431713b9a4e90bfb4a2424b702ed7a09051ddc15

PR #6046, opened 2018-10-16
Add external configuration change watcher

deathbeam:
Add watcher that will watch current configuration files for external
changes. This helps a lot when running multiple clients at once as they
can sync their changes between each other without completely overwriting
every change they made.


alexbowers:
Will this resolve issues I've been having using Runelite on a mac where occasionally all the settings and plugins get reverted back to their defaults (item prices, ground items, special attack) etc all deactivate for me. I tend to notice it after a new release, though not every release seems to do it.

Adam-:
No. That is unrelated and might be fixed by https://github.com/runelite/runelite/commit/33954ccf67f11c1da61ef45b21a7541b39c0a81a

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -175,24 +178,52 @@ public void load()
...
+
+		if (configWatcherTask != null)
+		{
+			configWatcherTask.cancel(true);
+		}
+
+		configWatcherTask = executor.scheduleAtFixedRate(new ConfigWatcher(propertiesFile, () ->
+		{
+			final Properties properties = new Properties();
+			loadProperties(properties);
```
Adam-: This should fail more immediately if it can't load config so it won't delete everything

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -175,24 +178,52 @@ public void load()
...
+
+			Map<String, String> copy = (Map) ImmutableMap.copyOf(properties);
+			copy.forEach((groupAndKey, value) ->
```
Adam-: This needs to be a map difference or something, this won't work at all for removed config items.

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -175,24 +178,52 @@ public void load()
...
+				if (split.length != 2)
+				{
+					log.debug("Properties key malformed!: {}", groupAndKey);
+					properties.remove(groupAndKey);
+					return;
+				}
+
+				final String groupName = split[0];
+				final String key = split[1];
+				final String oldValue = (String) this.properties.setProperty(groupName + "." + key, value);
```
Adam-: I guess our existing config changes are mostly on the AWT thread? Maybe not in this, but at some point we should lock around the config changes maybe or otherwise atomically apply them since they are accessed often by the client thread and assumed to not change.

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -175,24 +178,52 @@ public void load()
...
+			loadProperties(properties);
+
+			Map<String, String> copy = (Map) ImmutableMap.copyOf(properties);
+			copy.forEach((groupAndKey, value) ->
+			{
+				final String[] split = groupAndKey.split("\\.", 2);
+				if (split.length != 2)
+				{
+					log.debug("Properties key malformed!: {}", groupAndKey);
+					properties.remove(groupAndKey);
```
Adam-: This is not necessary

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
deathbeam: Done (this.properties.keySet().retainAll(copy.keySet());)

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
deathbeam: Hmm yeah a lot of the updating is from AWT thread (- like bank tags what is purely on client thread)

deathbeam:
Any update on this @Adam- ? The review comments are resolved

deathbeam:
Updated the PR (initially it had some issues, now it should work well when being logged in and also not being logged in + properly persists to remote when needed)

deathbeam:
There is nothing async in ConfigManager::load so there is nothing to race with (and runnign config watcher task is cancelled on new call to ::load). And I think we should not promote that use-case because it just causes data loss for users, idk.

deathbeam:
I can add cli param maybe, so people will have at least option if they really really want to, but it's really bad use-case that harms users.

Abextm:
A `--no-config-sync` flag would be good

Abextm:
the lambda is running on a separate thread than load, hence why it can race

deathbeam:
Ah i see, i need to cancel it at start of load, and not after. + added the cli param

deathbeam:
Hmm, that's dumb because cancel(true) should make sure that will not happen.. I can probably just set some boolean instead :man_shrugging:

Abextm:
`cancel(true)` just sets the thread's interrupted flag. It doesn't stop execution of user code. Preferably you would just use shared monitor on anything that will be editing the config storage.

Abextm:
> [12:44 AM] abex: so lets say you have a running rl, then you start another one and it crashes while writing it's newly downloaded config to disk. now your local config is mostly gone. The working rl sees its changed, updates it's loaded config to be mostly default, and the remote too
> [12:44 AM] abex: this is bad
> [12:46 AM] deathbeam: hmm
> [12:46 AM] deathbeam: i guess that would suck a lot
> [12:46 AM] abex: yes
> [12:48 AM] deathbeam: well ok. i guess that sucks enough for me to remove the remote sync

deathbeam:
Alright updated @Abextm , now it is only syncing local state but not updating remote state

deathbeam:
peek 2019-01-18 17-11

preview I guess

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -249,6 +273,71 @@ private synchronized void saveToFile() throws IOException
...
+			return;
+		}
+
+		log.info("Starting file change config watcher for file {}", propertiesFile);
+
+		configWatcherTask = executor.scheduleAtFixedRate(new ConfigWatcher(propertiesFile, () ->
+		{
+			try
+			{
+				syncSemaphore.acquireUninterruptibly();
```
Abextm: just enter `ConfigManager.this`'s monitor rather than using a semaphore

deathbeam:
Updated

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -126,9 +130,19 @@ private File getPropertiesFile()
 
 	public void load()
```
Abextm:
```suggestion
	public synchronized void load()
```

deathbeam:
Done

Abextm:
This code works and is ok, however the more I think about this the more I dislike it because it will make it impossible to have 2 RuneLites doing slayer without having an account for each of them. This should probably wait until we have profiles.

deathbeam:
That actually works, because slayer is not polled

deathbeam:
And I dont see any other scenario else where it "breaks" stuff.

Abextm:
That sounds like behavior we should not rely on

deathbeam:
Its way less bad that on what we rely on currently when running multiple clients, e.g nothing and everything breaks everything else.

deathbeam:
And after profiles will be eventually (maybe even this century) implemented, we would not need to rely on that minor slayer thing, and until then our users will stop losing their settings.

PR #6056, opened 2018-10-18
Added windertodt falling snow notifier

robinwithes:
Added a option in the windertodt plugin to notify you when you are standing in a spot where snow is about to fall when in windertodt.

Sorry for some of the variable names having "Avalanche" instead of "fallingsnow" (or similiar) in them. I did'n really know what to call it.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
@@ -59,7 +59,7 @@ public interface WintertodtConfig extends Config
 
 	@ConfigItem(
 		position = 3,
-		keyName = "notifyAvalanche",
+		keyName = "notifyFallingSnow",
 		name = "Notify when in a spot where snow is about to fall",
 		description = "Gives you a notification when in a falling snow spot"
 	)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
```diff
@@ -55,4 +56,15 @@ default Color damageNotificationColor()
 	{
 		return Color.CYAN;
 	}
+
+	@ConfigItem(
+		position = 3,
+		keyName = "notifyAvalanche",
```
deathbeam: Also keyname

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -205,6 +210,27 @@ private void checkActionTimeout()
...
+		{
+			return;
+		}
+
+		Player player = client.getLocalPlayer();
+		GameObject gameObject = event.getGameObject();
+
+		if (WINDERTODT_FALLING_SNOW == gameObject.getId())
+		{
+			if (gameObject.getLocalLocation().equals(player.getLocalLocation()))
```
deathbeam: Isnt the falling snow AOE? e.g, will this actually work?

robinwithes: I tested it and it worked.

Issue #6069, opened 2018-10-18
nbsp should be replaced with spaces in CC history to make it match properly


MagicfTail:

raiyni:
It shouldn't. 

`chats.removeIf(s::equalsIgnoreCase);`

MagicfTail:
![image](https://user-images.githubusercontent.com/12607046/47182732-79c56e00-d326-11e8-9251-a3ee027a2fb0.png)
🤷‍♀️

deathbeam:
So looks like this is not casing issue, but nbsp issue after all

PR #6074, opened 2018-10-18
Force clan chat names to a specific format

raiyni:
Closes https://github.com/runelite/runelite/issues/6069

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/Text.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/Text.java
@@ -25,6 +25,7 @@
  */
 package net.runelite.client.util;
 
+import com.google.common.base.CharMatcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/util/Text.java
```diff
@@ -56,6 +57,18 @@ public static String standardize(String str)
...
 	}
 
+	/**
+	 * Convert a string into Jagex username format
+	 * Remove all non-ascii characters, replace nbsp with space, replace _- with spaces, and trim
+	 *
+	 * @param str The string to standardize
+	 * @return The given `str` that is in Jagex name format
+	 */
+	public static String toJagexName(String str)
```
deathbeam: We already have method that replaces nbsp with space, 1 line above, is this one really necessary?

raiyni:
`standardize` lower cases the names (probably not desired for a history list) and it doesn't replace non-printables

PR #6082, opened 2018-10-19
XpGlobes: Use correct target XP

deathbeam:
Use correct target XP so actions left with set goal are not wrong (and
also to simply support in-game target goals).


deathbeam:
I actually made it from master, master was just force pushed. But, updated and rebased on actual master now

PR #6088, opened 2018-10-19
Barrows: Added dig area overlay

### LINKED ISSUE #4668: Barrows plugin crypt dig tiles

SebastiaanVanspauwen:
- Added barrows dig tiles overlays  (default on)

- Added Perspective function to draw outline of a poly with individual poly lines (so it takes height of tiles into account)
dig

fixes #4668

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Perspective.java
+++ b/runelite-api/src/main/java/net/runelite/api/Perspective.java
@@ -772,6 +772,14 @@ public class Perspective
 		return new Point(xOffset, yOffset);
 	}
 
+	/**
+	 * Calculates a polygon line from 2 LocalPoints
+	 *
+	 * @param client the game client
+	 * @param startLocation start location of the polygon
+	 * @param endLocation end location of the polygon
+	 * @return a {@link Polygon}
+	 */
 	private static Polygon linePoly(@Nonnull Client client, @Nonnull LocalPoint startLocation, @Nonnull LocalPoint endLocation)
 	{
 		LocalPoint startPoint = new LocalPoint(
@@ -793,7 +801,14 @@ public class Perspective
 		return null;
 	}
 
-
+	/**
+	 * Calculates a list of polygon lines with 2 Worldpoints
+	 * Start location is the South-West WorldPoint, end location is the North-East WorldPoint
+	 * @param client the game client
+	 * @param startLocation start location of the polygon
+	 * @param endLocation end location of the polygon
+	 * @return a {@link List} of {@link Polygon} 
+	 */
 	public static List<Polygon> getLinePolyList(@Nonnull Client client, @Nonnull WorldPoint startLocation, @Nonnull WorldPoint endLocation)
 	{
 		List<Polygon> pList = new ArrayList<>();


runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrothers.java
```diff
@@ -32,17 +32,26 @@
...
 public enum BarrowsBrothers
 {
-	AHRIM("Ahrim", new WorldPoint(3566, 3289, 0), Varbits.BARROWS_KILLED_AHRIM),
-	DHAROK("Dharok", new WorldPoint(3575, 3298, 0), Varbits.BARROWS_KILLED_DHAROK),
-	GUTHAN("Guthan", new WorldPoint(3577, 3283, 0), Varbits.BARROWS_KILLED_GUTHAN),
-	KARIL("Karil", new WorldPoint(3566, 3275, 0), Varbits.BARROWS_KILLED_KARIL),
-	TORAG("Torag", new WorldPoint(3553, 3283, 0), Varbits.BARROWS_KILLED_TORAG),
-	VERAC("Verac", new WorldPoint(3557, 3298, 0), Varbits.BARROWS_KILLED_VERAC);
-
+	AHRIM("Ahrim", new WorldPoint(3566, 3289, 0), Varbits.BARROWS_KILLED_AHRIM,
```
Nightfirecat: I feel like a `WorldArea` would be much better for representing these. That way, only a single area need be passed, rather than three separate points.

deathbeam:
I don't think adding this is good idea, it is very ugly

runelite-api/src/main/java/net/runelite/api/Perspective.java
```diff
@@ -771,4 +772,119 @@ public static Point getCanvasTextMiniMapLocation(
...
+
+	/**
+	 * Calculates a list of polygon lines with 2 Worldpoints
+	 * Start location is the South-West WorldPoint, end location is the North-East WorldPoint
+	 * @param client the game client
+	 * @param startLocation start location of the polygon
+	 * @param endLocation end location of the polygon
+	 * @return a {@link List} of {@link Polygon} 
+	 */
+	public static List<Polygon> getLinePolyList(@Nonnull Client client, @Nonnull WorldPoint startLocation, @Nonnull WorldPoint endLocation)
```
Adam-: Doesn't WorldArea already do this? We do mark some >1x1 areas already eg for npc respawns using it.

ollieatkinson:
Not sure I can get behind this, as deathbeam mentions - it's very ugly.

tyd3:
Ah this is what I was looking for; can this not be added, but with the Default option as off? It's similar to multicannon double hit spots being Defaulted to off; some people don't want to see such squares interfering with the game's graphics.

whartd:
> Ah this is what I was looking for; can this not be added, but with the Default option as off? It's similar to multicannon double hit spots being Defaulted to off; some people don't want to see such squares interfering with the game's graphics.

I support this. 

Also, perhaps another colour would look less "ugly"? I'm thinking a green that's close enough to the grass tone to be subtle, but different enough to be noticeable.

PR #6095, opened 2018-10-20
Kitten hungry idle notifiers

robinwithes:
Added a option in idle notifiers to warn you when your cat is hungry or wants attention.


deathbeam:
This should probably be in chat notifier, also, you can already use chat notifier for this

robinwithes:
> 
> 
> This should probably be in chat notifier, also, you can already use chat notifier for this

So should i just close it or implement it into chat notifier? I personally think this could be good as a seperate option.

deathbeam:
Probs just move it to chat notifier then

Trevor159:
This should be closed because it can already be done in chat notifications.

PR #5053, opened 2018-08-22
Multicombat borders 

mattec92:
Adds a plugin that when active will draw lines along the edges of multicombat zones.
Source of coordinates: https://github.com/runelite/runelite/issues/1764#issuecomment-384809773

bild

Fixes https://github.com/runelite/runelite/issues/1764

PR #6097, opened 2018-10-20
Add multicombat lines, pvp safe zone lines and deadman safe zone lines

### LINKED ISSUE #1764: multi area with outlines
Add multi area's outlines, great for wilderness and DMM

WooxSolo:
Adds zone indicators, which contains multicombat area outlines, pvp safe zone outlines and deadman safe zone outlines.

image

image

Fixes #1764


Adam-:
See also #5053. My opinion is this pushes the unfairness too far as it eliminates non multi to multi luring which is a common clan tactic[1]. I realize some smaller clients have this feature but they also contain other plugins we have been told to remove. 

https://www.reddit.com/r/2007scape/comments/9og6d7/fools_kills_abyss_for_max/

Kbman99:
@Adam- I can see this side of the argument, but at the same time tile markers can be used to achieve a similar outcome, albeit much more manual to perform. Most of the clans have custom clients as you mentioned and the guys luring are also more than likely to have multi lines marked in some fashion. Why not level the playing field?

devLotto:
Concerns from various content creators: https://twitter.com/ianspam/status/1044190116007366656

Abextm:
The way I see it is that we already have this feature via tile indicators, so logically there is no reason not to include this. However the PvP community probably won't realize that (similar to the drag thing), so __not__ adding this probably makes sense.

devOSRS:
full support

WooxSolo:
What part of fossil island are we talking about? Here's what is currently marked
image

Matt2408:
Hi Woox. I forgot to check the north side after I found this small part. I marked down the east - south area that's missing.

fossil

XrioBtw:
The unicorn pen in Ardougne is not a safe zone in deadman mode.

testerm0nkey:
> @Adam- I can see this side of the argument, but at the same time tile markers can be used to achieve a similar outcome, albeit much more manual to perform. Most of the clans have custom clients as you mentioned and the guys luring are also more than likely to have multi lines marked in some fashion. Why not level the playing field?

I am part of clan that uses the "Smaller Client" Exilent and we can all see the pvp lines and know what were doing. We also lure everyone we can into multi all the time. I like to use runelite when im not pking but would love it if we could get the necessary plugins to make it so that I dont have to switch all the time.

kgozi:
Is this added? I still use a different client for pking in the wilderness (as well as my clan).

This is the one feature we need holding us back from swapping fully. 

Love the client otherwise!

WooxSolo:
@kgozi No, it looks unlikely atm that it'll get added due to some people disliking the feature.

runelite-client/src/main/java/net/runelite/client/plugins/zoneindicators/ZoneIndicatorsPlugin.java
```diff
@@ -0,0 +1,296 @@
...
+				safeZonePath = Geometry.transformPath(safeZonePath, this::transformWorldToLocal);
+			}
+			pvpPathToDisplay[i] = safeZonePath;
+		}
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getKey().equals("collisionDetection") ||
```
osrs-music-map: "collisionDetection".equals(event.getKey()) would be better null handling

Nightfirecat: ConfigChanged is always posted with a key so it doesn't matter in this case.

osrs-music-map:
@WooxSolo does this plugin even work anymore?

WooxSolo:
@osrs-music-map No idea, I haven't touched it in the last few months as it looks like it might not get merged

osrs-music-map:
@WooxSolo ground lines don't show up unless I uncheck the collision filter option

what is the collision filter used for?

Adam-:
I'd rather not have this feature at this time.

whartd:
> See also #5053. My opinion is this pushes the unfairness too far as it eliminates non multi to multi luring which is a common clan tactic[1]. I realize some smaller clients have this feature but they also contain other plugins we have been told to remove.
> 
> https://www.reddit.com/r/2007scape/comments/9og6d7/fools_kills_abyss_for_max/

So this plugin would help prevent luring? How is that a bad thing? Jagex is removing 1 way safespots, another case where pkers have an advantage simply by knowing what tiles they should be on. You're saying that clans are luring legitimate players who wish to be in singles into multicombat - that *luring* is a "tactic" used by clans. If anything, preventing players from being scammed or lured sounds like a huge positive. Ignoring obviously overwhelming support for this plugin is disappointing @Adam-

Adam-:
I don't think this is comparable to 1 way safe spots. 1 way safe spots is not an intended game mechanic, which is evident by Jagex patching it out. Multi luring is and really has always been there. Also suggesting this prevents people getting "scammed" who are already *in the wilderness* is sort of disingenuous. I'm all for whatever we can do to prevent scams and maybe even the traditional non wildy -> wildy luring (see eg the hop-to world changes I did for CoX pvp world lures), but they are already in the wilderness so I think it is fair game.

jachymb:
I don't really see the big difference to object markers or tile markers. In theory, you could already use the tile marker to do the same, it would just be very tedious, but probably still easier than some in-game grinds, lol.

Issue #5075, opened 2018-08-23
Movable/ larger XP drops


Jmac191:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always
frustrated when [...]
N/A

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've
considered.

It'd be nice to be able to make the XP number drops larger than what we're currently able to. Would it also be possible to add an ALT+drag option somewhat like the XP globes to the XP drops? I'd like to be able to put the drops right beside my character or beside my inventory. 

**Additional context**
N/A

SirDylan:
Agreed.

CamiloJac:
Has anything happened with this yet?

Felanbird:
This is possible via the hub plugin `Customizable xp drops` 

![image](https://github.com/runelite/runelite/assets/41973452/9e257b23-cf59-4bab-b7e4-b3e2cf5edfda)

PR #6101, opened 2018-10-20
Moveable XP Drops

Hydrox6:
This has been a fun ride so far.

The widget will properly update its position when the Position value is changed in the XP Tracker, recalculating the widget's default position.

Multiple accounts are properly handled, including Deadman mode.

**Known Issues:**
* Due to the hierarchy of the widgets, the xp drops are not visible outside of the viewport.

Fulfills half of #5075


deathbeam:
This should use just WidgetOverlay, it has no relation to XP drops plugin at all

Hydrox6:
Updated it to work entirely within WidgetOverlay. Still feel it's messy, but ¯\\\_(ツ)\_/¯

PR #6111, opened 2018-10-21
Time tracking Cactus spine patch icon change

ltvill:
Potato cactus patches passed the poll, this is to avoid confusion.


devLotto:
This was changed in 72ca27edca65b579c5a4f30d45bf20d9d120418f.

PR #6119, opened 2018-10-21
Stretched Resizable Mode

### LINKED ISSUE #858: Add ability to scale resizable mode HUD
I'm not sure about the possibility of doing this, but a scaleable HUD would be a world of difference and open up playing on Resizeable Mode on pretty much most modern PC's. As it is, I really have trouble reading chat without leaning and clicking on menus would be far easier.

Cheers.

### LINKED ISSUE #5918: Stretched Resizable Mode
Just as fixed mode has a stretched mode (Stretched Fixed Mode in the Runelite options menu), where some horizontal & some vertical lines are doubled to give the game the effect of being larger, and if done on an integer scale, 2x or beyond scaled. Stretched resizable mode would do the same thing except with resizable mode.

Why? 

- This allows the UI to be "scaled" being more comfortable for some users. Eg. I'm used to stretched fixed mode's speed when turning the camera with middle mouse button which has a way slower sensitivity than regular resizeable mode
- This would also allow for less clipping of terrain which happens in fixed mode. 
- A more "clean" UI. More efficient use of game space 
- In effect, a "scaled UI" that Jagex was working on years ago and never got around to officially implementing. 

I'd also like to preface the rest of this with: I checked the to see if there were any completely similar suggestions. While #4769 and others were closed because they were recommended they use dpi scaling, this isn't a real solution to me because while it does scale the game, it also scales the Runelite UI, leading to a lessened experience than an official implementation. I could not find any other suggestions that were the exact same as mine.

fixed stretched
As you can see in this image, when using integer scale and using stretched fixed mode the game doesn't use all of the available screen. This is as intended and working perfectly if the user doesn't want to distort their game in any way. But this still leads to there being "wasted space" surrounding the game.

Here is a photo manipulated mock up of what this might look like when implemented:
resizeable stretched concept

Here is what resizable mode normally looks like:
08jrf5v

This mock up was just created by taking my [available screen real estate](https://i.imgur.com/rDXqlcn.png) (1642 x 1057) when accounting for the runelite panel (which I always have open) and dividing this number by half, (821x528.5) [then resizing my game to create an image this size](https://i.imgur.com/y8kvSQD.png). And finally just taking this image and scaling it 2x to fit the space.

So stretched resizable mode would work as so: Calculate the available screen real estate the user has (the area beside the panel), divide by 2, and set that as the client size, then scale that image 2x. Or if the user wants, 3x, 4x, etc.

devLotto:
Closes #5918 
Closes #858

Adds stretching of the game while in resizable mode to the ``Stretched Fixed Mode`` plugin. 
The plugin is renamed to ``Stretched Mode`` since it with this change also scales resizable mode.

As we all know, resizable likes to full up the whole game area. This feature works by limiting the area that it can take up, and then we resize the resulting game area to fill the whole client.

As a result, the UI is larger, and FPS is increased, with the downside that the game is slightly blurrier (since we're making an image larger). Still, with the scaling factor set to as high as 70%, the game looks OK, while the UI is enlargened by quite a bit.

**No scaling:**

The game is 1366x768 in all images.

**50% scaling:**

**70% scaling:** 

**100% scaling:**

Only works internally until merged.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -1103,9 +1103,9 @@ public interface Client extends GameEngine
 	RenderOverview getRenderOverview();
 
 	/**
-	 * Checked whether the client is in stretched mode.
+	 * Checks whether the client is in stretched mode.
 	 *
-	 * @return true if the client is in stretched, false otherwise
+	 * @return true if the client is in stretched mode, false otherwise
 	 */
 	boolean isStretchedEnabled();
 
@@ -1117,16 +1117,16 @@ public interface Client extends GameEngine
 	void setStretchedEnabled(boolean state);
 
 	/**
-	 * Checks whether the client is using fast rendering techniques when
-	 * stretching the client in fixed mode.
+	 * Checks whether the client is using fast
+	 * rendering techniques when stretching the canvas.
 	 *
-	 * @return true if client is fast rendering, false otherwise
+	 * @return true if stretching is fast rendering, false otherwise
 	 */
 	boolean isStretchedFast();
 
 	/**
-	 * Sets whether to use fast rendering techniques when in stretch
-	 * fixed mode.
+	 * Sets whether to use fast rendering techniques
+	 * when stretching the canvas.
 	 *
 	 * @param state new fast rendering state
 	 */
@@ -1134,18 +1134,35 @@ public interface Client extends GameEngine
 
 	/**
 	 * Sets whether to force integer scale factor by rounding scale
-	 * factors towards {@code zero} when stretching fixed mode.
+	 * factors towards {@code zero} when stretching.
 	 *
 	 * @param state new integer scaling state
 	*/
 	void setStretchedIntegerScaling(boolean state);
 
 	/**
-	 * Sets whether to keep aspect ratio when stretching fixed mode.
+	 * Sets whether to keep aspect ratio when stretching.
 	 *
 	 * @param state new keep aspect ratio state
 	 */
 	void setStretchedKeepAspectRatio(boolean state);
+
+	/**
+	 * Sets the scaling factor when scaling resizable mode.
+	 *
+	 * @param factor new scaling factor
+	 */
+	void setScalingFactor(int factor);
+
+	/**
+	 * Invalidates cached dimensions that are
+	 * used for stretching and scaling.
+	 *
+	 * @param resize true to tell the game to
+	 *               resize the canvas on the next frame,
+	 *               false otherwise.
+	 */
+	void invalidateStretching(boolean resize);
 
 	/**
 	 * Gets the current stretched dimensions of the client.
--- a/runelite-client/src/main/java/net/runelite/client/plugins/stretchedfixedmode/StretchedFixedModePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/stretchedfixedmode/StretchedFixedModePlugin.java
@@ -29,7 +29,9 @@ import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import javax.inject.Inject;
 import net.runelite.api.Client;
+import net.runelite.api.events.CanvasSizeChanged;
 import net.runelite.api.events.ConfigChanged;
+import net.runelite.api.events.ResizeableChanged;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.input.MouseManager;
 import net.runelite.client.plugins.Plugin;
@@ -84,6 +86,18 @@ public class StretchedFixedModePlugin extends Plugin
 	}
 
 	@Subscribe
+	public void onResizableChanged(ResizeableChanged event)
+	{
+		client.invalidateStretching(true);
+	}
+
+	@Subscribe
+	public void onCanvasSizeChanged(CanvasSizeChanged event)
+	{
+		client.invalidateStretching(false);
+	}
+
+	@Subscribe
 	public void onConfigChanged(ConfigChanged event)
 	{
 		if (!event.getGroup().equals("stretchedfixedmode"))
@@ -99,5 +113,8 @@ public class StretchedFixedModePlugin extends Plugin
 		client.setStretchedIntegerScaling(config.integerScaling());
 		client.setStretchedKeepAspectRatio(config.keepAspectRatio());
 		client.setStretchedFast(config.increasedPerformance());
+		client.setScalingFactor(config.scalingFactor());
+
+		client.invalidateStretching(true);
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/stretchedmode/StretchedModeConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/stretchedmode/StretchedModeConfig.java
@@ -23,14 +23,14 @@
  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package net.runelite.client.plugins.stretchedfixedmode;
+package net.runelite.client.plugins.stretchedmode;
 
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
-@ConfigGroup("stretchedfixedmode")
-public interface StretchedFixedModeConfig extends Config
+@ConfigGroup("stretchedmode")
+public interface StretchedModeConfig extends Config
 {
 	@ConfigItem(
 		keyName = "keepAspectRatio",
--- a/runelite-api/src/main/java/net/runelite/api/events/CanvasSizeChanged.java
+++ b/runelite-api/src/main/java/net/runelite/api/events/CanvasSizeChanged.java
@@ -1,16 +1,15 @@
 /*
- * Copyright (c) 2017. l2-
- *
+ * Copyright (c) 2018, Lotto <https://github.com/devLotto>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
  *
  * 1. Redistributions of source code must retain the above copyright notice, this
- *     list of conditions and the following disclaimer.
+ *    list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
- *     this list of conditions and the following disclaimer in the documentation
- *     and/or other materials provided with the distribution.
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
  *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
  * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
@@ -23,19 +22,11 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-
 package net.runelite.api.events;
 
-import lombok.Data;
-
 /**
- * An event where the game has changed from fixed to resizable mode or vice versa.
+ * An event posted when the canvas size might have changed.
  */
-@Data
-public class ResizeableChanged
+public class CanvasSizeChanged
 {
-	/**
-	 * Whether the game is in resizable mode.
-	 */
-	private boolean isResized;
 }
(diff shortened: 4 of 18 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java
```diff
@@ -36,18 +36,18 @@
 import net.runelite.client.plugins.PluginDescriptor;
 
 @PluginDescriptor(
-	name = "Stretched Fixed Mode",
-	description = "Resize the game while in fixed mode",
-	tags = {"resize"},
+	name = "Game Scaling",
```
deathbeam: I would probably name this canvas scaling, seems more appropriate

runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java
```diff
@@ -96,13 +105,58 @@ public void setStretchedKeepAspectRatio(boolean state)
 		cachedStretchedDimensions = null;
 	}
 
+	@Inject
+	@Override
+	public void setScalingFactor(int factor)
+	{
+		if (factor > 100)
```
deathbeam:
```
factor = Math.min(Math.max(factor, 0), 100)
```

runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java
```diff
@@ -96,13 +105,58 @@ public void setStretchedKeepAspectRatio(boolean state)
...
+		// Canvas needs to be resized so setMaxCanvasSize is called with the new realDimensions.
+		setResizeCanvasNextFrame(true);
+	}
+
 	@Inject
 	@Override
 	public Dimension getRealDimensions()
 	{
-		if (isStretchedEnabled() && !isResized())
+		if (isScalingEnabled())
```
deathbeam: I would flip this entire if statement and replace it with early return of getCanvas().getSize();

runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java
```diff
@@ -112,20 +166,26 @@ public Dimension getRealDimensions()
...
 	public Dimension getStretchedDimensions()
 	{
-		Canvas canvas = getCanvas();
+		gameScaling$checkUpToDate();
 
-		int width = canvas.getParent().getWidth();
-		int height = canvas.getParent().getHeight();
-
-		if (cachedStretchedDimensions == null || width != lastCanvasDimensions.width || height != lastCanvasDimensions.height)
+		if (cachedStretchedDimensions == null)
```
deathbeam: Same here

devLotto: I don't think that would work in this case

runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java
devLotto: I think 'canvas' is unknown to players so I picked 'game' instead. 
What do others who are reading this think?

runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java
deathbeam: if (cachedStetchedDimensions != null)
{
  return cachedStretchedDimensions;
}

Surely it will. Its just kind of hard to read these nested ifs, when basically the entire method is one big if block. But, its just cosmetical improvement so it does not really matter

devLotto: And then do 
```
cachedStretchedDimensions = new Dimension(width, height);
return cachedStretchedDimensions;
``` at the end?

File: runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java (hunk 1 above)
Comment (deathbeam) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: Imo it fits more because canvas usually refers to area where something is drawn, and scaling the drawing area describes what this plugin does perfectly.

File: runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java (hunk 4 above)
Comment (deathbeam) [reply runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java]: yeah

Comment (devLotto) [reply runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java]: I think that looks more confusing

Comment (deathbeam) [reply runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java]: Alright

File: runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java (hunk 1 above)
Comment (trimbe) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: While "Canvas Scaling" is obviously more accurate, I think many people wouldn't easily recognize its function. I agree with Lotto, "Game Scaling" seems much easier for most people to understand.

Comment (raiyni) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: Wouldn't `Interface Scaling` be more appropriate?

Comment (devLotto) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: If it only scaled resizable, ``Interface Scaling`` would be a good name, since that's what it visibly does in resizable mode. In fixed mode though, it visibly stretches out the whole canvas, which I wouldn't call interface scaling.

Comment (nickowz) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: Imo `Game Scaling` sounds the best and reflects what is really happening , since not only the interface is rescaled but also the game models.

Comment (deathbeam) [reply runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingPlugin.java]: I guess we have winner then, I disagree with Interface Scaling too

File: runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingConfig.java [hunk 5]
```diff
@@ -23,14 +23,14 @@
...
-package net.runelite.client.plugins.stretchedfixedmode;
+package net.runelite.client.plugins.gamescaling;
 
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
-@ConfigGroup("stretchedfixedmode")
-public interface StretchedFixedModeConfig extends Config
+@ConfigGroup("gamescaling")
```
Adam-: FWIW this will break existing stretched fixed mode config for users

runelite-client/src/main/java/net/runelite/client/events/AppletResized.java
```diff
@@ -0,0 +1,29 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.events;
+
+public class AppletResized
```
Adam-: Would it be better for plugins interested in this to just register a component listener on the client?

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -1147,6 +1147,15 @@
...
 	void setStretchedKeepAspectRatio(boolean state);
 
+	/**
+	 * Sets the scaling factor when scaling resizable mode.
+	 *
+	 * @param factor new scaling factor
+	 */
+	void setScalingFactor(int factor);
+	
+	void invalidateScaling();
```
Adam-: Add javadoc

runelite-client/src/main/java/net/runelite/client/events/AppletResized.java
devLotto: @deathbeam what do you reckon?

deathbeam: Hmm probably, even though I would keep this event and post it from e.g canvasWidth and canvasHeight properties

runelite-client/src/main/java/net/runelite/client/plugins/stretchedmode/StretchedModeConfig.java
```diff
@@ -55,10 +55,20 @@ default boolean increasedPerformance()
...
 	)
 	default boolean integerScaling()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "scalingFactor",
+		name = "Resizable Scaling (%)",
+		description = "In resizable mode, the game is reduced in size this much before it's stretched."
```
Adam-: We should probably add positions to these config items now

deathbeam: Well, as the options are now alphabetically sorted, the order would be
* Increased performance mode
* Integer scaling
* Keep aspect ratio
* Resizable scaling

What seems fine, I think

raiyni:
Closes https://github.com/runelite/runelite/issues/858

nickowz:
@devLotto @deathbeam Any possibility of this being merged, so there is more people testing it, and potentially finding any bug ? Without being merged we can't test this. Thanks.

runelite-client/src/main/java/net/runelite/client/plugins/stretchedmode/StretchedModeConfig.java
```diff
@@ -23,19 +23,20 @@
...
-public interface StretchedFixedModeConfig extends Config
+@ConfigGroup("stretchedmode")
+public interface StretchedModeConfig extends Config
 {
 	@ConfigItem(
 		keyName = "keepAspectRatio",
 		name = "Keep aspect ratio",
-		description = "Keeps the aspect ratio when stretching"
+		description = "Keeps the aspect ratio when stretching.",
+		position = 1
```
deathbeam: As I stated in my reply to @Adam- 's comment, these positions are useless as it is sorted by name already (and the position is exactly same as explicitly specified here)

devLotto: General stretching options at the top, resizable options at the bottom. 
Promise you'll catch anyone trying to add a non-resizable related config at the bottom and I'll remove it.

deathbeam: I dont think we will be adding any config here in future

devLotto: Remove positions.

runelite-client/src/main/java/net/runelite/client/events/AppletResized.java
devLotto: Changed to use fieldhooks on canvasWidth & canvasHeight, posting CanvasSizeChanged.

runelite-client/src/main/java/net/runelite/client/plugins/gamescaling/GameScalingConfig.java
devLotto: I think it's worth it.

deathbeam: And searching for "Stretched Fixed Mode" in config sidebar still returns this plugin, so it should also not confuse users v much

runelite-mixins/src/main/java/net/runelite/mixins/GameScalingMixin.java
devLotto: Changed to use ``Ints.constrainToRange(factor, 0, 100);``

devLotto:
Don't merge, looking into if this can be simplified quite a bit

runelite-client/src/main/java/net/runelite/client/plugins/stretchedfixedmode/StretchedFixedModePlugin.java
```diff
@@ -83,6 +85,18 @@ protected void shutDown() throws Exception
 		mouseManager.unregisterMouseWheelListener(mouseWheelListener);
 	}
 
+	@Subscribe
+	public void onResizableChanged(ResizeableChanged event)
```
deathbeam: the correct method name should be onResizeableChanged not onResizableChanged

raiyni: Resizable is the correct spelling in English

deathbeam: Doesn't matter, the event is named ResizeableChanged not ResizableChanged

deathbeam: The name is still wrong

PR #6129, opened 2018-10-22
Added option to swap "Cast Bloom" and "Wield" on Silver sickle (b)

Boredska:
Should make collecting mort myre fungi easier.


Boredska:
I mostly play with my hands off-keyboard, so shift-click is not an option.

deathbeam:
Use either sticky keys or the on screen keyboard and just toggle shift there

Issue #1765, opened 2018-04-21
Show monsters aggro area


Popedigoz:
Show monsters' aggro area, (Show the area of which you can not step out of, or the monsters become aggro again.)

eggbrow:
Yes this is something I wanted to ask also. It's interesting how some monsters have very different aggro area/attack range/aggro range inside that area.

Popedigoz:
Can be closed, as this is now in the "game". (NPC Aggression Timer -plugin)

Nightfirecat:
You should be able to close this issue yourself. But yes, this has been implemented.

PR #6133, opened 2018-10-22
Add NPC unaggression timer

WooxSolo:
Adds a timer and displays the area of which the player can move within without making NPCs aggressive again.

image

There's a config of which the user can edit to decide at which NPCs the plugin should be active at, and I suspect people might want a list of useful NPCs, so I've also put together one with some of the more commonly killed NPCs where this plugin would be useful at: https://pastebin.com/wYfVfu9S

Uses the same geometry methods as #6097 so if one PR gets merged, the other needs rebasing.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
+++ b/runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
@@ -1,0 +1,454 @@
+/*
+ * Copyright (c) 2018, Woox <https://github.com/wooxsolo>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.api.geometry;
+
+import java.awt.Shape;
+import java.awt.geom.AffineTransform;
+import java.awt.geom.GeneralPath;
+import java.awt.geom.PathIterator;
+import java.awt.geom.Point2D;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.BiPredicate;
+import java.util.function.Consumer;
+
+public class Geometry
+{
+	/**
+	 * Find the point where two lines intersect.
+	 *
+	 * @param x1 X coordinate of the first endpoint of the first line.
+	 * @param y1 Y coordinate of the first endpoint of the first line.
+	 * @param x2 X coordinate of the second endpoint of the first line.
+	 * @param y2 Y coordinate of the second endpoint of the first line.
+	 * @param x3 X coordinate of the first endpoint of the second line.
+	 * @param y3 Y coordinate of the first endpoint of the second line.
+	 * @param x4 X coordinate of the second endpoint of the second line.
+	 * @param y4 Y coordinate of the second endpoint of the second line.
+	 * @return The intersection point of the lines, or null if the lines don't intersect.
+	 */
+	public static Point2D.Float lineIntersectionPoint(
+		float x1, float y1, float x2, float y2,
+		float x3, float y3, float x4, float y4)
+	{
+		// https://stackoverflow.com/a/1968345
+
+		float p1x = x2 - x1;
+		float p1y = y2 - y1;
+		float p2x = x4 - x3;
+		float p2y = y4 - y3;
+
+		float s = (-p1y * (x1 - x3) + p1x * (y1 - y3)) / (-p2x * p1y + p1x * p2y);
+		float t = ( p2x * (y1 - y3) - p2y * (x1 - x3)) / (-p2x * p1y + p1x * p2y);
+
+		if (s >= 0 && s <= 1 && t >= 0 && t <= 1)
+		{
+			float x = x1 + (t * p1x);
+			float y = y1 + (t * p1y);
+			return new Point2D.Float(x, y);
+		}
+
+		// No intersection
+		return null;
+	}
+
+	/**
+	 * Find the intersection points between a Shape and a line.
+	 *
+	 * @param shape The shape.
+	 * @param x1 X coordinate of the first endpoint of the line.
+	 * @param y1 Y coordinate of the first endpoint of the line.
+	 * @param x2 X coordinate of the second endpoint of the line.
+	 * @param y2 Y coordinate of the second endpoint of the line.
+	 * @return A list with the intersection points.
+	 */
+	public static List<Point2D.Float> intersectionPoints(Shape shape, float x1, float y1, float x2, float y2)
+	{
+		List<Point2D.Float> intersections = new LinkedList<>();
+
+		PathIterator it = shape.getPathIterator(new AffineTransform());
+		float[] coords = new float[2];
+		float[] prevCoords = new float[2];
+		float[] start = new float[2];
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				start[0] = coords[0];
+				start[1] = coords[1];
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_LINETO)
+			{
+				Point2D.Float intersection = lineIntersectionPoint(
+					prevCoords[0], prevCoords[1], coords[0], coords[1], x1, y1, x2, y2);
+				if (intersection != null)
+				{
+					intersections.add(intersection);
+				}
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_CLOSE)
+			{
+				Point2D.Float intersection = lineIntersectionPoint(
+					coords[0], coords[1], start[0], start[1], x1, y1, x2, y2);
+				if (intersection != null)
+				{
+					intersections.add(intersection);
+				}
+			}
+			it.next();
+		}
+
+		return intersections;
+	}
+
+	/**
+	 * Transforms the points in a path according to a method.
+	 *
+	 * @param it The iterator of the path to change the points on.
+	 * @param method The method to use to transform the points. Takes a float[2] array with x and y coordinates as parameter.
+	 * @return The transformed path.
+	 */
+	public static GeneralPath transformPath(PathIterator it, Consumer<float[]> method)
+	{
+		GeneralPath path = new GeneralPath();
+		float[] coords = new float[2];
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				method.accept(coords);
+				path.moveTo(coords[0], coords[1]);
+			}
+			else if (type == PathIterator.SEG_LINETO)
+			{
+				method.accept(coords);
+				path.lineTo(coords[0], coords[1]);
+			}
+			else if (type == PathIterator.SEG_CLOSE)
+			{
+				path.closePath();
+			}
+			it.next();
+		}
+
+		return path;
+	}
+
+	/**
+	 * Transforms the points in a path according to a method.
+	 *
+	 * @param path The path to change the points on.
+	 * @param method The method to use to transform the points. Takes a float[2] array with x and y coordinates as parameter.
+	 * @return The transformed path.
+	 */
+	public static GeneralPath transformPath(GeneralPath path, Consumer<float[]> method)
+	{
+		return transformPath(path.getPathIterator(new AffineTransform()), method);
+	}
+
+	/**
+	 * Splits a line into smaller segments and appends the segments to a path.
+	 *
+	 * @param path The path to append lines to.
+	 * @param segmentLength The desired length to use for the segmented lines.
+	 * @param x1 X coordinate of the first endpoint of the line.
+	 * @param y1 Y coordinate of the first endpoint of the line.
+	 * @param x2 X coordinate of the second endpoint of the line.
+	 * @param y2 Y coordinate of the second endpoint of the line.
+	 */
+	private static void appendSegmentLines(GeneralPath path, float segmentLength,
+		float x1, float y1, float x2, float y2)
+	{
+		float x = x1;
+		float y = y1;
+		float angle = (float)Math.atan2(y2 - y1, x2 - x1);
+		float dx = (float)Math.cos(angle) * segmentLength;
+		float dy = (float)Math.sin(angle) * segmentLength;
+		float length = (float)Math.hypot(x2 - x1, y2 - y1);
+		int steps = (int)((length - 1e-4) / segmentLength);
+		for (int i = 0; i < steps; i++)
+		{
+			x += dx;
+			y += dy;
+			path.lineTo(x, y);
+		}
+	}
+
+	/**
+	 * Splits a path into smaller segments.
+	 * For example, calling this on a path with a line of length 6, with desired
+	 * segment length of 2, would split the path into 3 consecutive lines of length 2.
+	 *
+	 * @param it The iterator of the path to modify.
+	 * @param segmentLength The desired length to use for the segments.
+	 * @return The modified path.
+	 */
+	public static GeneralPath splitIntoSegments(PathIterator it, float segmentLength)
+	{
+		GeneralPath newPath = new GeneralPath();
+		float[] prevCoords = new float[2];
+		float[] coords = new float[2];
+		float[] startCoords = new float[2];
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				startCoords[0] = coords[0];
+				startCoords[1] = coords[1];
+				newPath.moveTo(coords[0], coords[1]);
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_LINETO)
+			{
+				appendSegmentLines(newPath, segmentLength, prevCoords[0], prevCoords[1], coords[0], coords[1]);
+				newPath.lineTo(coords[0], coords[1]);
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_CLOSE)
+			{
+				appendSegmentLines(newPath, segmentLength, coords[0], coords[1], startCoords[0], startCoords[1]);
+				newPath.closePath();
+			}
+			it.next();
+		}
+
+		return newPath;
+	}
+
+	/**
+	 * Splits a path into smaller segments.
+	 * For example, calling this on a path with a line of length 6, with desired
+	 * segment length of 2, would split the path into 3 consecutive lines of length 2.
+	 *
+	 * @param path The path to modify.
+	 * @param segmentLength The desired length to use for the segments.
+	 * @return The modified path.
+	 */
+	public static GeneralPath splitIntoSegments(GeneralPath path, float segmentLength)
+	{
+		return splitIntoSegments(path.getPathIterator(new AffineTransform()), segmentLength);
+	}
+
+	/**
+	 * Removes lines from a path according to a method.
+	 *
+	 * @param it The iterator of the path to filter.
+	 * @param method The method to use to decide which lines to remove. Takes two float[2] arrays with x and y coordinates of the endpoints of the line. Lines for which the predicate returns false are removed.
+	 * @return The filtered path.
+	 */
+	public static GeneralPath filterPath(PathIterator it, BiPredicate<float[], float[]> method)
+	{
+		GeneralPath newPath = new GeneralPath();
+		float[] prevCoords = new float[2];
+		float[] coords = new float[2];
+		float[] start = new float[2];
+		boolean shouldMoveNext = false;
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				start[0] = coords[0];
+				start[1] = coords[1];
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+				shouldMoveNext = true;
+			}
+			else if (type == PathIterator.SEG_LINETO)
+			{
+				if (method.test(prevCoords, coords))
+				{
+					if (shouldMoveNext)
+					{
+						newPath.moveTo(prevCoords[0], prevCoords[1]);
+						shouldMoveNext = false;
+					}
+					newPath.lineTo(coords[0], coords[1]);
+				}
+				else
+				{
+					shouldMoveNext = true;
+				}
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_CLOSE)
+			{
+				if (shouldMoveNext)
+				{
+					newPath.moveTo(prevCoords[0], prevCoords[1]);
+				}
+				if (method.test(prevCoords, start))
+				{
+					newPath.lineTo(start[0], start[1]);
+				}
+				shouldMoveNext = false;
+			}
+			it.next();
+		}
+
+		return newPath;
+	}
+
+	/**
+	 * Removes lines from a path according to a method.
+	 *
+	 * @param path The path to filter.
+	 * @param method The method to use to decide which lines to remove. Takes two float[2] arrays with x and y coordinates of the endpoints of the line. Lines for which the predicate returns false are removed.
+	 * @return The filtered path.
+	 */
+	public static GeneralPath filterPath(GeneralPath path, BiPredicate<float[], float[]> method)
+	{
+		return filterPath(path.getPathIterator(new AffineTransform()), method);
+	}
+
+	/**
+	 * Removes lines from a path that lie outside the clipping area and cuts
+	 * lines intersecting with the clipping area so the resulting lines
+	 * lie within the clipping area.
+	 *
+	 * @param it The iterator of the path to clip.
+	 * @param shape The clipping area to clip with.
+	 * @return The clipped path.
+	 */
+	public static GeneralPath clipPath(PathIterator it, Shape shape)
+	{
+		GeneralPath newPath = new GeneralPath();
+		float[] prevCoords = new float[2];
+		float[] coords = new float[2];
+		float[] start = new float[2];
+		float[] nextMove = new float[2];
+		boolean shouldMove = false;
+		boolean wasInside = false;
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				start[0] = coords[0];
+				start[1] = coords[1];
+				wasInside = shape.contains(coords[0], coords[1]);
+				if (wasInside)
+				{
+					nextMove[0] = coords[0];
+					nextMove[1] = coords[1];
+					shouldMove = true;
+				}
+				prevCoords[0] = coords[0];
+				prevCoords[1] = coords[1];
+			}
+			else if (type == PathIterator.SEG_LINETO || type == PathIterator.SEG_CLOSE)
+			{
+				if (type == PathIterator.SEG_CLOSE)
+				{
+					coords[0] = start[0];
+					coords[1] = start[1];
+				}
+
+				List<Point2D.Float> intersections = intersectionPoints(shape, prevCoords[0], prevCoords[1], coords[0], coords[1]);
+				intersections.sort((a, b) ->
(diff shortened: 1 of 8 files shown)


deathbeam:
Just quick observation based on screenshot, maybe the text color could change based on if you are inside aggro area or not, so you dont need to have the lines enabled in order to see whether you are inside or not

WooxSolo:
> Just quick observation based on screenshot, maybe the text color could change based on if you are inside aggro area or not, so you dont need to have the lines enabled in order to see whether you are inside or not

Not really sure what you mean. The player is always within the area. When they walk out of the area, the timer just resets to 10 mins and the area moves.

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -514,6 +516,18 @@ static Object stringToObject(String str, Class<?> type)
 			int mods = Integer.parseInt(splitStr[1]);
 			return new Keybind(code, mods);
 		}
+		if (type == WorldPoint.class)
+		{
+			String[] splitStr = str.split(",");
```
deathbeam: You can just use : as delimiter, like Point and Rectangle use

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,367 @@
...
+
+		return newPath;
+	}
+
+	public static GeneralPath unitifyPath(GeneralPath path, float unitSize)
+	{
+		return unitifyPath(path.getPathIterator(new AffineTransform()), unitSize);
+	}
+
+	public static GeneralPath filterPath(PathIterator it, BiFunction<float[], float[], Boolean> method)
```
Abextm: BiPredicate

deathbeam:
Oh, nevermind then

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,367 @@
...
+				start = null;
+				shouldMoveNext = false;
+			}
+			it.next();
+		}
+
+		return newPath;
+	}
+
+	public static GeneralPath filterPath(GeneralPath path, BiFunction<float[], float[], Boolean> method)
```
Abextm: BiPredicate

deathbeam:
For the NPC aggro timer, in addition to "highlighting" specific monsters, what about showing aggro timer for last interacted NPC type and make it default behaviour?

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
WooxSolo: Out of curiosity, is there a reason colons are preferred over commas for this?

deathbeam: I think someone just made it like that and then we just followed. So no reason really, but its better to stay consistent.

WooxSolo:
> For the NPC aggro timer, in addition to "highlighting" specific monsters, what about showing aggro timer for last interacted NPC type and make it default behaviour? (with config option)

Some NPCs are always aggressive, like dark beast and most bosses, and I don't think people would want it to show up at those places.

deathbeam:
Well, the plugin would be disabled by default and it would be config option, and maybe blacklist for those common ones would work. E.g hardcode exclusion for like Goraks and Dark Beasts.

WooxSolo:
I think a blacklist would be bigger than a whitelist because there's also a bunch of NPCs that are never aggressive, like nechryael and abyssal demons.

deathbeam:
Well, as I said we can have both, and hardcoded blacklist just for common stuff. Imo it is more user friendly when it just works when you want it to work without having to input the NPC to config.

WooxSolo:
What about instead of a blacklist we put my pastebin whitelist inside the plugin as a config checkbox? So enabling the checkbox enables all the common NPCs?

As for the plugin being disabled by default, the problem with that is that when the plugin is then enabled, the player needs to teleport (or enter a dungeon or similar) for the plugin to figure out where the unaggressive area is. So if the player goes to their favourite NPCs and then realise they want to use the plugin, they'll need to leave the place and come back for it to work.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+				recheckActive();
+				break;
+		}
+	}
+
+	private void loadConfig()
+	{
+		String configGroup = NpcAggroAreaConfig.CONFIG_GROUP + "." + username;
+
+		safeCenters[0] = configManager.getConfiguration(configGroup, NpcAggroAreaConfig.CONFIG_CENTER1, WorldPoint.class);
```
Adam-: All of this just for account specific config? Probably not worth it.

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
```diff
@@ -39,14 +39,19 @@
 	private final Duration duration;
 
 	public Timer(long period, ChronoUnit unit, BufferedImage image, Plugin plugin)
+	{
+		this(Instant.now(), period, unit, image, plugin);
+	}
+
+	public Timer(Instant startTime, long period, ChronoUnit unit, BufferedImage image, Plugin plugin)
```
Adam-: This constructor seems unnecessary? You should be able to calculate out the length yourself when creating the timer.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
WooxSolo: It's mostly there so the area and timer can be determined upon login. If we remove the config, the area will be unknown on login and the user would have to teleport/enter a dungeon before the plugin figures out where to outline the area.

Adam-: The config is fine - but it would be a lot simpler with using the normal config interface... the only reason to use config manager in this way is for login-specific config, which is sort of a hack.

WooxSolo: Well if the user logs onto another account, the config for the other account would need to be removed in that case, causing the same issue. As far as I know the only difference is adding the username to the config anyway, so it's not really more complicated?

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
```diff
@@ -0,0 +1,106 @@
...
+
+	@ConfigItem(
+		keyName = "npcUnaggroAlwaysActive",
+		name = "Always active",
+		description = "Always show this plugins overlays<br>Otherwise, they will only be shown when any NPC name matches the list",
+		position = 1
+	)
+	default boolean alwaysActive()
+	{
+		return false;
```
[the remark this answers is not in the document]
WooxSolo: If the plugin is disabled by default, the user would need to teleport/enter a dungeon after turning on the plugin for it to work. Is that preferable over having the plugin enabled by default?

Also, (as discussed with db) would it be preferred to instead have another checkbox enabled by default which makes the plugin work at a predefined list of NPCs instead of enabling ```alwaysActive```?

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
WooxSolo: It's there so the timer consistently becomes red at the same number every time (which I think is at 10%). If the timer starts at 5 mins (which it can do upon login if the user was already there for 5 min earlier) the text would be red at 30 sec instead of 60 sec.

Adam-: Oh. It might be better to make the warning time modifiable then (overriding a method?)

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
deathbeam: We already do this at other places and trying to migrate it from global config to account-specific in future would be more annoying than having it account specific in first place, even with "hacky" method imo.

deathbeam:
I still like the last NPC with blacklist a bit more. For whitelist with toggle, more preferred solution would be just pre-fill the text field. But, hardcoded blacklist is easily manageable as anyone can PR new entry to it, and also blacklist with last interacted would automatically work with new content, but whitelist will not.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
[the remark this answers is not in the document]
deathbeam: I dont think alwaysActive should ever be default as true, check the discussion going on PR @Adam- . I personally think that this option do not even need to exist

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
```diff
@@ -0,0 +1,106 @@
...
+		return true;
+	}
+
+	@ConfigItem(
+		keyName = "npcUnaggroCollisionDetection",
+		name = "Collision detection",
+		description = "Only show lines where they can be walked through",
+		position = 5
+	)
+	default boolean collisionDetection()
```
deathbeam: Does this even need config option?

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
```diff
@@ -0,0 +1,106 @@
...
+		return true;
+	}
+
+	@ConfigItem(
+		keyName = "npcUnaggroShowAreaLines",
+		name = "Show area lines",
+		description = "Display lines, when walked past, the unaggressive timer resets",
+		position = 4
+	)
+	default boolean showAreaLines()
```
deathbeam: This should be probably disabled by default

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,139 @@
...
+import net.runelite.api.coords.LocalPoint;
+import net.runelite.api.geometry.Geometry;
+import net.runelite.client.ui.overlay.Overlay;
+import net.runelite.client.ui.overlay.OverlayLayer;
+import net.runelite.client.ui.overlay.OverlayPosition;
+import net.runelite.client.ui.overlay.OverlayPriority;
+import net.runelite.client.ui.overlay.components.LineComponent;
+import net.runelite.client.ui.overlay.components.PanelComponent;
+
+public class NpcAggroAreaOverlay extends Overlay
```
deathbeam: remove public modifier

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,139 @@
...
+{
+	private final static int MAX_LOCAL_DRAW_LENGTH = 20 * Perspective.LOCAL_TILE_SIZE;
+
+	@Inject
+	private Client client;
```
deathbeam: Inject everything through the overlay constructor (and make it private and these fields final)

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,139 @@
...
+
+	@Inject
+	private NpcAggroAreaConfig config;
+
+	@Inject
+	private NpcAggroAreaPlugin plugin;
+
+	private PanelComponent panelComponent;
```
deathbeam: this can be final

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,139 @@
...
+	public NpcAggroAreaOverlay()
+	{
+		panelComponent = new PanelComponent();
+		panelComponent.setPreferredSize(new Dimension(150, 0));
+
+		setLayer(OverlayLayer.ABOVE_SCENE);
+		setPriority(OverlayPriority.LOW);
+	}
+
+	private void renderPath(Graphics2D graphics, GeneralPath path, Color color)
```
deathbeam: you should probably place this below overriden methods

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
deathbeam: I agree with @Adam- , everywhere else we simply override the methods when needed. Simply override getColor

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+import net.runelite.client.ui.overlay.OverlayManager;
+import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
+import net.runelite.client.ui.overlay.infobox.Timer;
+import net.runelite.client.util.WildcardMatcher;
+
+@Slf4j
+@PluginDescriptor(
+	name = "Unaggressive NPC timer",
+	description = "Highlights the unaggressive area of NPCs nearby and timer until it becomes active",
+	tags = {"highlight", "lines", "unaggro", "aggro", "aggressive", "npcs", "area", "timer", "slayer"}
```
deathbeam: Disable this plugin by default

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+	NpcAggroAreaConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(NpcAggroAreaConfig.class);
+	}
+
+	@Override
+	protected void startUp() throws Exception
+	{
+		overlayManager.add(overlay);
+		safeCenters = new WorldPoint[2];
```
deathbeam: just make safeCenters final and null content in shutdown instead of reinitializing the array (or make it list)

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+		linesToDisplay = new GeneralPath[Constants.MAX_Z];
```
deathbeam: same here

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+		npcNamePatterns = NAME_SPLITTER.splitToList(config.npcNamePatterns());
+
+		clientThread.invokeLater(() -> username = client.getUsername());
```
deathbeam: this should be in GameStateChanged or even better, should not exist at all (just call client.getUsername() when needed)

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+
+	@Subscribe
+	public void onNpcSpawned(NpcSpawned event)
+	{
+		if (config.alwaysActive())
+		{
+			return;
+		}
+
+		if (!active && isNpcMatch(event.getNpc()))
```
deathbeam: you should make this common method that accepts var amount of NPCs and do the reevaluate active and active true there (check ClueScrollPlugin what does basically the same thing)

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,579 @@
...
+			safeCenters[0] = null;
+			safeCenters[1] = null;
+			lastPlayerLocation = newLocation;
+		}
+	}
+
+	@Subscribe
+	public void onGameStateChanged(GameStateChanged event)
+	{
+		if (event.getGameState() == GameState.LOGGED_IN)
```
deathbeam: use switch(event.getGameState())

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
WooxSolo: What should we do if we remove the config option? Keep collision or not? I kinda like having both options, no collision detection makes it easier to see exactly where the area is, but adding collision detection fixes issues where the lines goes off the map like this
image

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
WooxSolo: It is there because client.getUsername() is null after the player logs out (when gamestate changes to LOGIN_SCREEN). Though, if we're not using account specific config, the plugin doesn't need to know the username at all.

MagicfTail:
A nice addition that's 100% going to be requested later is a notification when monsters stop agroing

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
deathbeam: Keep collision. If you cant walk to end of the area it is not important to know full area imo (as that is mostly just debug info and not actually useful info)

MagicfTail:
Bug: The area you turn on the plugin in will not work even after resetting the agro tiles by teleporting far away/going in a dungeon.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,138 @@
...
+	public Dimension render(Graphics2D graphics)
+	{
+		if (!plugin.isActive())
+		{
+			return null;
+		}
+
+		if (plugin.getSafeCenters()[1] == null)
+		{
+			setPosition(OverlayPosition.TOP_LEFT);
```
deathbeam: I dont think this will ever work, the positions are cached when overlay is added to the manager

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,554 @@
...
+				calculateLinesToDisplay();
+				break;
+			case "npcUnaggroNames":
+				npcNamePatterns = NAME_SPLITTER.splitToList(config.npcNamePatterns());
+				recheckActive();
+				break;
+		}
+	}
+
+	private void loadConfig()
```
deathbeam: Maybe after @Adam- requested removal of per-account settings this should not be persisted at all.

WooxSolo: Why? Removing it would cause the plugin to not work on login until the player teleports.

deathbeam: Because atm it would break also when using mutliple accounts on same PC.

Abextm: So rather than breaking on a subset of users it should break for everyone?

deathbeam: Or the change that @Adam- requested should be put back.

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -551,6 +565,15 @@ static String objectToString(Object object)
...
 			return k.getKeyCode() + ":" + k.getModifiers();
 		}
+		if (object instanceof WorldPoint)
+		{
+			WorldPoint wp = (WorldPoint)object;
+			return wp.getX() + ":" + wp.getY() + ":" + wp.getPlane();
+		}
+		if (object instanceof Duration)
+		{
+			return ((Duration)object).toMillis() + "";
```
Nightfirecat: `Long.toString(((Duration)object).toMillis())`

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -551,6 +565,15 @@ static String objectToString(Object object)
 			Keybind k = (Keybind) object;
 			return k.getKeyCode() + ":" + k.getModifiers();
 		}
+		if (object instanceof WorldPoint)
+		{
+			WorldPoint wp = (WorldPoint)object;
+			return wp.getX() + ":" + wp.getY() + ":" + wp.getPlane();
```
Nightfirecat: Using `':'` (char) versus `":"` (string) will be more performant.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -0,0 +1,555 @@
...
+		npcNamePatterns = NAME_SPLITTER.splitToList(config.npcNamePatterns());
+	}
+
+	@Override
+	protected void shutDown() throws Exception
+	{
+		removeTimer();
+		overlayManager.remove(overlay);
+		overlayManager.remove(notWorkingOverlay);
+		safeCenters[0] = null;
```
Nightfirecat: `Arrays.fill(safeCenters, null);`

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
WooxSolo: That would be inconsistent with how we've done it for other classes before though

DannyKong12:
can we add notifications to this plugin

WooxSolo:
Gonna post this here for people who want a more noob-friendly explanation of how the area is calculated: https://reddit.com/r/2007scape/comments/a9bqgl/how_monster_aggro_areas_and_resetting_works/

15987632:
https://streamable.com/31zm0 This probably shouldn't draw on black void, I don't know how you would check for this, maybe height?

Nightfirecat:
This will close #1765

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+		float[] prevCoords = new float[2];
+		float[] coords = new float[2];
+		float[] start = null;
+		boolean shouldMoveNext = false;
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				if (start == null)
```
Nightfirecat: Same as `unitifyPath()`, why is `start` initialized to null and used as a least-recent connector to SEG_CLOSE?

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+	public static GeneralPath unitifyPath(GeneralPath path, float unitSize)
+	{
+		return unitifyPath(path.getPathIterator(new AffineTransform()), unitSize);
+	}
+
+	/**
+	 * Removes lines from a path according to a method.
+	 *
+	 * @param it The iterator of the path to filter.
+	 * @param method The method to use to decide which lines to remove. Takes two float[2] arrays with x and y coordinates of the endpoints of the line.
```
Nightfirecat: This doesn't really make it clear whether a `true` or `false` test would indicate a line should be removed.

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+	 * @param y1 Y coordinate of the first endpoint of the line.
+	 * @param x2 X coordinate of the second endpoint of the line.
+	 * @param y2 Y coordinate of the second endpoint of the line.
+	 */
+	private static void appendUnitLines(GeneralPath path, float unitSize,
+		float x1, float y1, float x2, float y2)
+	{
+		float x = x1;
+		float y = y1;
+		float angle = (float)Math.atan2(y2 - y1, x2 - x1);
```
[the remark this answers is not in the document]
Abextm: `StrictMath` is not more accurate, and even if it was it wouldn't matter because we are in using 32 bit floats mostly, then casting to an integer.

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+		float dx = (float)Math.cos(angle) * unitSize;
+		float dy = (float)Math.sin(angle) * unitSize;
+		float length = (float)Math.hypot(x2 - x1, y2 - y1);
+		int steps = (int)((length - 1e-4) / unitSize);
+		for (int i = 0; i < steps; i++)
+		{
+			x += dx;
+			y += dy;
+			path.lineTo(Math.round(x), Math.round(y));
```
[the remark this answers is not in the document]
WooxSolo: You're right, the rounding was supposed to be in the collision filter.

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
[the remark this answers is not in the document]
WooxSolo: As far as I can tell, the `Math` methods seem to just be a wrapper to call the `StrictMath` versions anyway (at least on my machine).

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+	private static void appendUnitLines(GeneralPath path, float unitSize,
+		float x1, float y1, float x2, float y2)
+	{
+		float x = x1;
+		float y = y1;
+		float angle = (float)Math.atan2(y2 - y1, x2 - x1);
+		float dx = (float)Math.cos(angle) * unitSize;
+		float dy = (float)Math.sin(angle) * unitSize;
+		float length = (float)Math.hypot(x2 - x1, y2 - y1);
+		int steps = (int)((length - 1e-4) / unitSize);
```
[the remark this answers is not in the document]
WooxSolo: It's more like a Math.floor. If we round up the lines could go past the end. The int cast is there because if `steps` is something like 4.5, we'd go too far as we'd take 5 steps.
The idea behind the magic number is to always round down the result to remove negligible lines with length of something like 0.000000001 due to precision errors of floating-point values.

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+	/**
+	 * Splits a path into smaller segments.
+	 * For example, calling this on a path with a line of length 6, with desired
+	 * unit size of 2, would split the path into 3 consecutive lines of length 2.
+	 *
+	 * @param it The iterator of the path to modify.
+	 * @param unitSize The desired length to use for the segments.
+	 * @return The modified path.
+	 */
+	public static GeneralPath unitifyPath(PathIterator it, float unitSize)
```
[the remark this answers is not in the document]
WooxSolo: I always wanted a better name too but couldn't think of one. `splitIntoSegments` sounds good though. :)

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
```diff
@@ -0,0 +1,464 @@
...
+		GeneralPath newPath = new GeneralPath();
+		float[] prevCoords = new float[2];
+		float[] coords = new float[2];
+		float[] startCoords = null;
+		while (!it.isDone())
+		{
+			int type = it.currentSegment(coords);
+			if (type == PathIterator.SEG_MOVETO)
+			{
+				if (startCoords == null)
```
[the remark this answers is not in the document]
WooxSolo: The code just wasn't tested enough it seems. The plugin didn't call the method with any `SEG_CLOSE`-paths so the value of `startCoords` never mattered for the plugin to work.

runelite-api/src/main/java/net/runelite/api/geometry/Geometry.java
WooxSolo: Agreed, added some extra info.

WooxSolo:
> https://streamable.com/31zm0 This probably shouldn't draw on black void, I don't know how you would check for this, maybe height?

I've added a check to see whether the tile contains something to be drawn which seems to have fixed the issue.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaNotWorkingOverlay.java
```diff
@@ -0,0 +1,71 @@
...
+
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		if (!plugin.isActive())
+		{
+			return null;
+		}
+
+		if (plugin.getSafeCenters()[1] != null)
```
Nightfirecat: This should be combined with the check above to become

```
if (!plugin.isActive() || plugin.getSafeCenters()[1] != null)
{
    return null;
}
```

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -0,0 +1,125 @@
...
+
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		if (!plugin.isActive())
+		{
+			return null;
+		}
+
+		if (plugin.getSafeCenters()[1] == null)
```
Nightfirecat: Same as other comment, combine this with the guard clause above.

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
```diff
@@ -44,9 +46,9 @@ public Timer(long period, ChronoUnit unit, BufferedImage image, Plugin plugin)
 
 		Preconditions.checkArgument(period > 0, "negative period!");
 
-		startTime = Instant.now();
-		duration = Duration.of(period, unit);
-		endTime = startTime.plus(duration);
+		this.startTime = Instant.now();
```
Nightfirecat: These `this.` additions can be removed? I don't see any local variables in this scope that would hide the class fields.

WooxSolo: Tbh I like to prepend `this.` because it makes it clearer that we're not writing to local variables. But I don't care enough so I'll just change it

haakonrp:
Can this be added? Very useful on long slayer tasks etc.

Issue #6134, opened 2018-10-22
Toggling ignored on large amounts of items in grouped mode takes longer than expected.


MagicfTail:
Toggling ignored on 2000 bones in separated mode made the client unusable for around 10 seconds while toggling the same items in grouped mode took around 35 seconds.

https://i.imgur.com/YupZhtu.gifv

Edit: you also need to have a lot (20 ish) different items when hiding items for it to be really bad.

PR #6135, opened 2018-10-23
Revert "Loot tracker - Add ability to ignore items (#5483)" due to client freezing when hiding large amounts of items

MagicfTail:
https://i.imgur.com/YupZhtu.gifv
Having a feature that can lock up the client for this long is unacceptable

This reverts commit 31d8e88099588a6d4a8e7ab19e275a1f0fd73c13.


deathbeam:
What if you just comment out <https://github.com/runelite/runelite/pull/6135/files#diff-6a7f2b0ba476897b8e2a6a7c5288ddcfL114>? Did you even tried to debug this?

psikoi:
Did you try finding the reason for the freezing? What's the background, what circumstances does it crash in?

psikoi:
I found the origin issue, you should probably refer to it in this PR.

MagicfTail:
See issue #6134.

This pr might not be the cause of the issue, but it enables a way of freezing the client for many times longer than what was possible before, in a way that would not be unexpected for an actual user to experience (hiding an item with a large backlog of loot already tracked)

PR #6143, opened 2018-10-23
Remove barrows brother RuneLite overlay

### LINKED ISSUE #4366: Barrows brother tracker

deathbeam:
As positioning of vanilla barrows brothers overlay is decent and do not
obstructs view, remove ours (as it displays exactly same info anyway).

Closes #4366


deathbeam:
Looks like this will be still problem in resizable mode, so its probably better if we just keep ours

Issue #6150, opened 2018-10-23
Menu entry swapper plugin "Teleport item" option should have Ardougne Cloak removed from description


Nightfirecat:
Since the Jagex update to add individual menu entries for diary reward items with multiple teleports, the Ardougne Cloak and Morytania Legs no longer have their teleport options swapped to be left-click, as they are categorized as "Monastery Teleport", etc. rather than simply "Teleport". As such, we should update the description for this option to use some other item, perhaps Lumbridge Ring?

https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java#L208

PR #6157, opened 2018-10-24
Changed SwapTeleportItem Description

KasparRosin:
#6150 Ardougne Cloak removed from description


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -354,7 +354,7 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 			if (config.swapExchange())
 			{
 				swap("exchange", option, target, true);
-			}
+			}
```
forsco: you have unnecessary tabs here causing a diff

PR #6158, opened 2018-10-24
Removed Ardougne cloak from the menu entry swapper plugin

### LINKED ISSUE #6150: Menu entry swapper plugin "Teleport item" option should have Ardougne Cloak removed from description

KasparRosin:
Closes #6150


psikoi:
Please add "from the menu entry swapper plugin"

Issue #6146, opened 2018-10-23
Notification on dark core spawn


Trevor159:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always
frustrated when [...]

I die sometimes when afking corp too much when i spec him down when the core spawns in.

**Describe the solution you'd like**

An option in either the corp plugin or the npc indicator plugin that allows a notification when the NPC spawns

deathbeam:
You can just highlight dark core in NPC indicators plugin

15987632:
> You can just highlight dark core in NPC indicators plugin

Yes, I already have that done. I am asking for a notification when an npc (specifically the dark core) spawns, which to my knowledge can not be done through the NPC indicators plugin currently. Hence the request.

deathbeam:
I dont really see valid use case for that

Adam-:
I don't see anything wrong with making this an option of the corp plugin.

Twiglet1022:
I just had a go at trying to do this, but the dark core spawn even triggers every time it jumps so it ends up being a "The Dark Core has just jumped!" notifier.

15987632:
> I just had a go at trying to do this, but the dark core spawn event triggers every time it jumps so it ends up being a "The Dark Core has just jumped!" notifier.

If you looked closely you can see it is already solved https://i.imgur.com/lKOQ0iC.png

Twiglet1022:
> > I just had a go at trying to do this, but the dark core spawn event triggers every time it jumps so it ends up being a "The Dark Core has just jumped!" notifier.
> 
> If you looked closely you can see it is already solved https://i.imgur.com/lKOQ0iC.png
Ah thank you, I didn't realise that that meant it had already been solved.

abextm:
Jagex's [recent statement](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) prohibits `Adds additional visual or audio indicators of a boss mechanic`

PR #6161, opened 2018-10-24
Added config option in corp plugin for dark core spawn notification

robinwithes:
Closes https://github.com/runelite/runelite/issues/6146

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpConfig.java
@@ -52,4 +52,15 @@ public interface CorpConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "notifOnCore",
+		name = "Send notification on core spawn",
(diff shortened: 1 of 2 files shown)


Adam-:
If you kill the core does it respawn later?

MagicfTail:
Yes

Adam-:
You can probably just notify whenever the npc spawns in - maybe with a distance check so that it doesn't notify when walking into range.

robinwithes:
> You can probably just notify whenever the npc spawns in - maybe with a distance check so that it doesn't notify when walking into range.

The npc spawns in again each time it jumps.

15987632:
This probably isn't relevant to the way you coded this but if you kill the dark core in a way that the ashes do not drop(killing it with a cannon is the easiest way to do this) then it doesn't respawn for the rest of the kill.

abextm:
Jagex's [recent statement](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) prohibits `Adds additional visual or audio indicators of a boss mechanic`

PR #6164, opened 2018-10-24
Fix combat brace slayer task race condition

### LINKED ISSUE #6091: Combat bracelet Slayer Kill count is at -1

**Describe the bug**
Slayer bracelet is showing progress for the slayer task every 10 kills. But each time the message shoot on the the message box it sync the kill count of slayer task to number - 1.

**To Reproduce**
1. Equipe slayer bracelet
2. Do a slayer task
3. Wait for "You still need to kill xx monsters to complete your Slayer assignement.
4. See error on the slayer tracker top left.

**Expected behavior**
It should display the right count.

 - OS: MacOS
 - RuneLite version: 1.4.22
 - Launcher version: 1.6.1

deathbeam:
Apparently combat bracelet message is sent before XP drop is received,
so task updates to correct amount and then is substracted by 1. To
combat this, virtually increase amount by 1 so the xp drop will
substract to correct amount.

Closes #6091


Nightfirecat:
I'm pretty sure this is wrong. I believe the chat message and xp drop happen on the same tick, meaning that there's the possibility for the events to happen in either order (since that's not guaranteed by the event bus)

deathbeam:
Well based on the reports it seemed like pretty consistent breakage

PR #6165, opened 2018-10-24
Added missing falador rooftop course agility obstacle

robinwithes:


devLotto:
For any future reference: It's this one to the right that's being added.

image

robinwithes:
> 
> 
> For any future reference: It's this one to the right that's being added.
> 
> image

The right

PR #6168, opened 2018-10-25
Notifier: Add chat messages on game thread.

Abextm:


Adam-:
Can't this just use the chat message manager?

Abextm:
Possibly, but that is out of scope for this.

PR #6171, opened 2018-10-25
Add support for impostors to clue scroll plugin

deathbeam:
Add support for object impostors to clue scroll plugin.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -48,8 +48,10 @@ import net.runelite.api.Item;
 import net.runelite.api.ItemComposition;
 import net.runelite.api.ItemID;
 import net.runelite.api.NPC;
+import net.runelite.api.ObjectComposition;
 import net.runelite.api.Scene;
 import net.runelite.api.Tile;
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -588,10 +590,32 @@ private void highlightObjectsForLocation(final WorldPoint location, final int...
...
-			.filter(object -> object != null && ArrayUtils.contains(objectIds, object.getId()))
-			.toArray(GameObject[]::new);
+		for (GameObject object : tile.getGameObjects())
+		{
+			if (object == null)
+			{
+				continue;
+			}
+
+			for (int id : objectIds)
```
Adam-: Why not just `getImpostor()` and then check the id of that?

deathbeam: Right, I guess I can do that

deathbeam: Oh, that is because ObjectComposition does not have ID

deathbeam:
@Adam- added export for ObjectComposition#getId and used that instead of .getImpostorIds

PR #6172, opened 2018-10-25
cluescrolls: Add STASH unit object ids to emote clues

Abextm:
I would like to merge this before the rest of the feature so I don't have to rebase it 6800000 times.


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -46,6 +46,8 @@
 import static net.runelite.client.plugins.cluescrolls.clues.emote.Emote.BULL_ROARER;
 import net.runelite.client.plugins.cluescrolls.clues.emote.ItemRequirement;
 import net.runelite.client.plugins.cluescrolls.clues.emote.RangeItemRequirement;
+import static net.runelite.client.plugins.cluescrolls.clues.emote.STASHUnit.*;
+import static net.runelite.client.plugins.cluescrolls.clues.emote.STASHUnit.SHANTAY_PASS;
```
Nightfirecat: Don't need this if also using a star import.

Abextm: Its needed because there is another `SHANTAY_PASS` (from `ItemID`) in scope

PR #6173, opened 2018-10-25
Add support for objects to anagram clues

### LINKED ISSUE #6053: Awowogei anagram clue does not highlight him
(object ID should be 4771)

15987632:
Fixes #6053 
Depends on and closes #6171


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/AnagramClue.java
```diff
@@ -224,4 +258,9 @@ public static AnagramClue forText(String text)
 	{
 		return new String[] {npc};
 	}
-}
+
+	public int[] getObjectIds()
+	{
+		return new int[objectId];
```
TheStonedTurtle:
```suggestion
		return new int[] {objectId};
```

PR #6190, opened 2018-10-25
Regen meters: add animation smoothing

raiyni:
ezgif-1-553f737e40b9


MagicfTail:
Does this really need a config option?

raiyni:
There are some people that like the original aethstetic of runescape and turn off animation smoothing

Issue #6065, opened 2018-10-18
Examine tool doesn't work in worn equipment


raiyni:

deathbeam:
Can you update the description with that you showcased in Discord?

Nightfirecat:
That's covered by `MenuAction.EXAMINE_ITEM_BANK_EQ`, however the identifier of the `MenuEntryAdded` event is not correct for that type, as seen below: (the actual item ID is 1265)

```
018-10-25 11:45:42 [Thread-5] DEBUG n.r.c.p.devtools.DevToolsPlugin - MenuAction: EXAMINE_ITEM_BANK_EQ
2018-10-25 11:45:42 [Thread-5] DEBUG n.r.c.p.devtools.DevToolsPlugin - event: MenuEntryAdded(option=Remove, target=<col=ff9040>Bronze pickaxe</col>, type=57, identifier=1, actionParam0=-1, actionParam1=25362441)
```

PR #6207, opened 2018-10-26
Fix examine dev tool for equipment IDs

Nightfirecat:
This PR creates some generic utils for examining objects, and specifically getting item IDs and quantities from examined items of various types, and puts these utils to use in the Examine plugin and the examine ID dev tool.

Fixes runelite/runelite#6065
Closes runelite/runelite#6191
Fixes runelite/runelite#6192


runelite-client/src/main/java/net/runelite/client/plugins/devtools/DevToolsPlugin.java
```diff
@@ -283,35 +282,55 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 			return;
 		}
 
-		MenuAction action = MenuAction.of(event.getType());
+		final ExamineType examineType = ExamineType.of(MenuAction.of(event.getType()));
```
deathbeam: Maybe instead of doing what are you ding at line 297 you should instead check for last enu entry and not check for w/e is in MenuEntryAdded event

Nightfirecat: I've divested from the event as much as possible here. (It is now only used to guard against adding extra "ID: " suffixes to existing examine entries)

runelite-client/src/main/java/net/runelite/client/util/ExamineItem.java
```diff
@@ -0,0 +1,120 @@
...
+	 *                    {@link MenuEntryAdded#getActionParam0()})
+	 * @return An {@link ExamineItem} of appropriate {@code id} and {@code quantity}, or -1 for both fields if the item
+	 *         id could not be found.
+	 */
+	public static ExamineItem findExamineItem(final Widget widget, final ExamineType type, final int actionParam)
+	{
+		int foundItemId = -1;
+		int foundItemQuantity = -1;
+
+		if (widget == null)
```
deathbeam: This should return null and not empty examine item

runelite-client/src/main/java/net/runelite/client/util/ExamineItem.java
```diff
@@ -0,0 +1,120 @@
...
+
+import lombok.Value;
+import net.runelite.api.events.MenuEntryAdded;
+import net.runelite.api.events.MenuOptionClicked;
+import net.runelite.api.widgets.Widget;
+import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.api.widgets.WidgetItem;
+
+@Value
+public class ExamineItem
```
deathbeam: I was thinking about this a bit, and naming this ExamineItem is probably incorrect as it can be used at other places too (example, ItemStats plugin). Maybe MenuEntryItem?

deathbeam: Also, to be able to categorize it, ExamineItem should probably also hold WidgetInfo as type

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
```diff
@@ -22,12 +22,67 @@
...
 	OBJECT;
+
+	/**
+	 * Gets the {@link ExamineType} corresponding to the given {@link ChatMessageType}.
+	 *
+	 * @param chatMessageType The chat message type to get an appropriate examine type from
+	 * @return A corresponding examine type object
+	 */
+	@Nullable
+	public static ExamineType of(ChatMessageType chatMessageType)
```
deathbeam: This should probably stay in the ExaminePlugin as it will ever be useful only there

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
```diff
@@ -22,12 +22,67 @@
...
+	}
+
+	/**
+	 * Gets the {@link ExamineType} corresponding to the given {@link MenuAction}.
+	 *
+	 * @param menuAction The menu action type to get an appropriate examine type from
+	 * @return A corresponding examine type object
+	 */
+	@Nullable
+	public static ExamineType of(MenuAction menuAction)
```
deathbeam: Maybe naming this MenuActionItemType would be more approriate? After removing the chatbox stuff from here

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
Nightfirecat: That's only the case for now... It's possible that other future enhancements would make use of this for other purposes. (Examining npcs/monsters for instance) Given its relation to `ExamineType.of(MenuAction)`, I'm hesitant to split this up again.

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
Nightfirecat: See the comment above

Nightfirecat:
@deathbeam Implemented most of your feedback

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -543,4 +545,25 @@ public static int PACK(int groupId, int childId)
...
+	/**
+	 * Gets the {@link WidgetInfo} of given group and child IDs
+	 *
+	 * @param groupId The group ID to search for
+	 * @param childId The child ID to search for
+	 * @return The first matching {@link WidgetInfo} definition matching the passed IDs, or null if no such entry was
+	 *         found.
+	 */
+	@Nullable
+	public static WidgetInfo ofIds(final int groupId, final int childId)
```
deathbeam: I would probably name this method .from()

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
```diff
@@ -0,0 +1,94 @@
...
+	 */
+	ITEM,
+
+	/**
+	 * Item found in some other interface, such as bank, equipment, alternate inventory (like when using the Grand
+	 * Exchange or bank deposit box), shop, or items found when opening clue scrolls, barrows chests, etc.
+	 */
+	ITEM_BANK_EQ,
+
+	NPC,
```
deathbeam: When you documented ITEM, you should probably also document NPC and OBJECT

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
```diff
@@ -0,0 +1,94 @@
...
+
+public enum ExamineType
+{
+	/**
+	 * Inventory item
+	 */
+	ITEM,
+
+	/**
+	 * Item found in some other interface, such as bank, equipment, alternate inventory (like when using the Grand
```
deathbeam: You should probably document that this can sometimes (most of the times) be incorrect, e.g for chat message we thing that BANK_EQ is every server message

runelite-client/src/main/java/net/runelite/client/util/MenuEntryItem.java
```diff
@@ -0,0 +1,131 @@
...
+				if (actionParam < children.length)
+				{
+					Widget widgetItem = children[actionParam];
+					foundItemQuantity = widgetItem.getItemQuantity();
+					foundItemId = widgetItem.getItemId();
+				}
+			}
+		}
+
+		if (foundItemQuantity != -1)
```
deathbeam: checking for foundItemId would probably make more sense, also, I would just do early return null if foundItemId == -1.

runelite-client/src/main/java/net/runelite/client/util/ExamineType.java
Nightfirecat: Those were mostly self-explanatory. I only felt like `ITEM` and `ITEM_BANK_EQ` needed distinction, since the others were clear.

runelite-client/src/main/java/net/runelite/client/util/MenuEntryItem.java
```diff
@@ -0,0 +1,132 @@
...
+	 * @param widget      The examined widget
+	 * @param type        The examine type of the examination
+	 * @param actionParam The action param of the examination (see {@link MenuOptionClicked#getActionParam()} or
+	 *                    {@link MenuEntryAdded#getActionParam0()})
+	 * @return An {@link MenuEntryItem} of appropriate {@code id} and {@code quantity}. If the passed
+	 *         {@code widget} is null, or the item could not be found from the {@link WidgetItem}, null will be
+	 *         returned instead.
+	 */
+	@Nullable
+	public static MenuEntryItem findExamineItem(final Widget widget, final ExamineType type, final int actionParam)
```
deathbeam: You can probably just pass menu entry type here instead of ExamineType

Nightfirecat: That's the point of the Examinetype class though, that doesn't apply for chat messages...

deathbeam: Yes, but this method will not be called for chat messages but for menu entries (and the menu entry is actually already required for this method to work, or at least the Widget and action param from it)

Nightfirecat: What do you mean it is not called for chat messages? This method is called from `ExaminePlugin.findExamineItem()`, which is called from its `onChatMessage` subscriber.

deathbeam: that is just delayed MenuEntry event

Nightfirecat: I'll look into that tomorrow, I need to spend a little more time thinking about this to make sure I understand how everything's meant to be used.

Nightfirecat: I think I've implemented this how you've suggested it.

runelite-client/src/main/java/net/runelite/client/util/MenuEntryItem.java
```diff
@@ -0,0 +1,133 @@
...
+	 * @param widget      The examined widget
+	 * @param type        The menu action type of the examined entry
+	 * @param actionParam The action param of the examination (see {@link MenuOptionClicked#getActionParam()} or
+	 *                    {@link MenuEntryAdded#getActionParam0()})
+	 * @return An {@link MenuEntryItem} of appropriate {@code id} and {@code quantity}. If the passed
+	 *         {@code widget} is null, or the item could not be found from the {@link WidgetItem}, null will be
+	 *         returned instead.
+	 */
+	@Nullable
+	public static MenuEntryItem findExamineItem(final Widget widget, final MenuAction type, final int actionParam)
```
deathbeam: Now this can be named just .find, probably (or .from)

Nightfirecat: Renamed to `.find`

PR #6208, opened 2018-10-26
client: Add getWidget(int packedId)

Nightfirecat:


Nightfirecat:
Seems this isn't wanted.

PR #6209, opened 2018-10-26
Barb egg highlighter

### LINKED ISSUE #6026: A "Highlight egg colors" option in the Barbarian Assault plugin
Currently it is impossible to recolour 2> ground item names which have the same GP value. The only compromise is to highlight item1, default colour for item2, and then recolour a *menu* for item3 (requires a right-click). This is especially noticeable in Barbarian Assault, where the red, blue, and green eggs all have the same GP value. (there is also the yellow egg, on the last wave)

It would benefit a lot of players, especially those in the BA community, if in the "Barbarian Assault" plugin, there would be added a new toggle option such as "Highlight egg colors."

Hopefully that would isolate the need of otherwise having to make the ground item plugin even more complex and bloated.

Thanks for reading!

15987632:
Closes #6026


15987632:
This is stupid

PR #6216, opened 2018-10-27
Add enemy name to PVP kill screenshot

### LINKED ISSUE #6215: Add username of player killed in the filename on auto-screenshots of kills (PVP)
currently they are just saved as the date, but it would be really handy if it had the username in it so it could be searched. 

it's quite annoying trying to find a name you know you have a kill on, but being unable to do so out of thousands of generic pictures in a folder

i'm sure there's a way to read text printed in the chat box, so this should be a quick job :)

KasparRosin:
fixes #6215

Would this work?

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -357,15 +357,15 @@ public class ScreenshotPlugin extends Plugin
 				if (chatMessage.contains(KILL_MESSAGES(i)))
 				{
 					enemyName = chatMessage.replace((String.valueOf(KILL_MESSAGES(i))),"");
-					break
+					break;
 				}
 				else 
 				{
-					i++
+					i++;
 				}
 			}
 			if (enemyName)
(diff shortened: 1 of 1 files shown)


arthurtyukayev:
Could you use LocalPlayer.getInteracting?

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -351,7 +351,21 @@ public void onChatMessage(ChatMessage event)
...
+					enemyName = chatMessage.replace((String.valueOf(KILL_MESSAGES(i))),"");
+					break;
+				}
+				else 
+				{
+					i++;
+				}
+			}
+			if (enemyName)
+			String fileName = "Kill " + format(new Date()) + enemyName);
```
arthurtyukayev: This will be null if enemy name returns false. I would declare the file name where it was before and append if enemy name returns true.

raiyni: Are you even testing these? This will include the punctuation in the filename and it's the complete wrong way to do it. The KILL_MESSAGES should be converted to a regex with a group matcher for the name

PR #6220, opened 2018-10-27
Add back check for widget.isHidden() to TabInterface

deathbeam:
This check is still required, because in rare circumstances the bank
widget is simply not nulled.


raiyni:
Can this be taken care of in `onWidgetHiddenChanged`? This feels very ragtag.

deathbeam:
Idk how is this change related to onWidgetHiddenChanged

raiyni:
It stems from the fact that the bank widget doesn't get nulled but is just hidden.

PR #6241, opened 2018-10-29
Fix idle logout triggering when killing an enemy after 5 minutes of being idle

MagicfTail:
Stops killing an enemy resetting the combat countdown, triggering a notification


deathbeam:
If this is issue, I can reopen https://github.com/runelite/runelite/pull/5369/files as I am pretty sure I had it handled there correctly. The resetTimers() calls is still necessary to reset other idle timers

Issue #5133, opened 2018-08-26
Bags/Sacks overlays


DopeAnteater:
**Is your feature request related to a problem? Please describe.**
Request not related to problem

**Describe the solution you'd like**
Give a number over the item for each gem or coal, like jewelry has with it's appropriate charges

**Describe alternatives you've considered**
Make it a different overlay to look like boosts etc.

**Additional context**
Herb sack function as well

PR #6242, opened 2018-10-29
plugin/coalbag: add Coal Bag plugin

while-loop:
Add Coal Bag overlay to RuneLite.

This plugin keeps track of when a player Fills and Empties their coal bag as the Coal Bag does not use Varbits.

#5133
 
screen shot 2018-10-28 at 10 45 36 pm


Trevor159:
If you are still interested in this it would be a good addition to the plugin-hub

PR #6259, opened 2018-10-30
add functionality and config toggle for warning message in chat when …

Brodan:
…weapon change results in warned attack style

This PR addresses the TODO listed in AttackStylesPlugin.java.


raiyni:
You should probably add meaningful titles and descriptions to your PR vs letting your commit message be it by default.

Brodan:
just overhauled this from when i opened it last year, but it is not yet ready to be looked at. tests are still failing, and i can't seem to figure out why its breaking whenever a bow is equipped and switched between with weapons using the same attack style as the bow (i.e. equip a bow and use accurate, then switch to a sword also using accurate, the overlay will still say ranged even though it should update to accurate). either there's some deeper issue with varbits/varplayers that this PR can't solve, or this entire plugin needs to be overhauled

if anyone wants to checkout this branch and play around with it it would be greatly appreciated.

PR #6260, opened 2018-10-30
Make boost overlay to change to yellow at threshold 

### LINKED ISSUE #6243: The "STAT is getting low" notification plays one level before the stat shows "below threshold" color
It seems like it should only play once the stat would display in the yellow "below threshold" color.

Brodan:
Fixes #6243.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
@@ -81,7 +81,7 @@ public class BoostIndicator extends InfoBox
 			return new Color(238, 51, 51);
 		}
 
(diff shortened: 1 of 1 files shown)


deathbeam:
I think it would be better if the notification was fixed (and that is about what is the issue talking about). This just makes both events consistent and incorrect.

Brodan:
Ah okay, I misunderstood if the notification was supposed to be _at_ or _below_ the threshold. Will update the PR asap. Consistency is a good first step at least!

Brodan:
@deathbeam Updated.

deathbeam:
Hmm actually, i take what I said back, current behaviour is pretty confusing, just make the boost yellow indicator indicate at threshold

raiyni:
to be honest, I would like this without a notification

deathbeam:
Same, but, that is not related to this PR. the notification on boost seems annoying but the yellow indicator is nice

PR #6261, opened 2018-10-30
Plugin - Item Information Tooltips

BenDol:
Thought I'd submit this PR of something I put together over the weekend. I was finding it frustrating having to google search all the OSRS items during my adventures, so I figured since I couldn't find anything like this (I might have missed it) I'd create one.

**How it works:**
Currently this plugin is using osrsbox.com item database to retrieve all the item information via the REST API asynchronously and applies them to the item tooltip. The item search is standoff so it won't show information for an item thats name isnt an exact match. It might be worth while transitioning this database over to the runelite http service db and relying on that. For now this works fine.

**Demo:** https://youtu.be/_LbPpwzwN1U


deathbeam:
We already implemented this

BenDol:
Where can I find the existing implementation?

deathbeam:
Item Stats plugin

BenDol:
Item Stats Plugin doesn't accomplish what I have implemented in this PR, am I missing something? You get no item information like this plugin.

raiyni:
Then you have tooltips turned off?

BenDol:
It's a shame the current item stats plugin doesn't show item requirements like level requirement etc.

PR #6270, opened 2018-10-31
Add ability to set combat requirements for achievement diaries

Monsterxsync:
Added ability to add combat requirements to diarys also added all the current combat requirements.

ScreenShots:
When you meet the combat requirement.
when you have requirement
When you don't meet the combat requirement.
when you don't have requirement

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/DiaryRequirementsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/DiaryRequirementsPlugin.java
@@ -36,6 +36,7 @@ import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.ScriptID;
+import net.runelite.api.Skill;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetID;
@@ -242,7 +243,16 @@ public class DiaryRequirementsPlugin extends Plugin
 			{
 				RequirementStringBuilder requirementStringBuilder = new RequirementStringBuilder(i);
 
-				int realSkillLevel = client.getRealSkillLevel(i.getSkill());
+				Skill skill = i.getSkill();
+				int realSkillLevel;
+				if (skill == null && i.getCustomRequirement().equals("Combat"))
+				{
+					realSkillLevel = client.getLocalPlayer().getCombatLevel();
+				}
+				else
+				{
+					realSkillLevel = client.getRealSkillLevel(skill);
+				}
 				List<Integer> altRealSkillLevels = null;
 				if (i.getAltRequirements() != null)
 				{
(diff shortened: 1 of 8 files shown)


runelite-api/src/main/java/net/runelite/api/Skill.java
```diff
@@ -56,7 +56,8 @@
 	/**
 	 * The level of all skills added together.
 	 */
-	OVERALL("Overall");
+	OVERALL("Overall"),
+	COMBAT("Combat");
```
Nightfirecat: This is a very poor way to implement this feature. This will cause very bad side effects for any plugin which tries to iterate over `Skill.values()` such as the XP tracker plugin. This should instead be incorporated into `Requirement` somehow.

Monsterxsync: wouldn't overall cause issues with that too?, i checked if it interfered with the xp tracker and didnt see an issue.

Monsterxsync:
I've moved it out of the `Skill Enum` and added a 2nd constructor to `Requirement`, it just assumes that you're using the custom field for combat requirement if i can think of another use for the custom reqs i'll change it around.

runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/DiaryRequirementsPlugin.java
```diff
@@ -242,7 +244,25 @@ private GenericDiaryRequirement getRequirementsForTitle(String title)
 			{
 				RequirementStringBuilder requirementStringBuilder = new RequirementStringBuilder(i);
 
-				int realSkillLevel = client.getRealSkillLevel(i.getSkill());
+				Skill skill = i.getSkill();
+				int realSkillLevel;
+				//if skill is null assumed to be combat.
+				if (skill == null)
```
Nightfirecat: This should probably also check that the custom requirement is "Combat".

runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/DiaryRequirementsPlugin.java
```diff
@@ -242,7 +244,25 @@ private GenericDiaryRequirement getRequirementsForTitle(String title)
...
+				{
+					realSkillLevel =  Experience.getCombatLevel(
```
Nightfirecat: `realSkillLevel = client.getLocalPlayer().getCombatLevel();`

runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/Requirement.java
```diff
@@ -29,14 +29,22 @@
 @Getter
 public class Requirement
 {
-	private final Skill skill;
-	private final int levelRequirement;
-	private final Requirement[] altRequirements;
+	private Skill skill;
```
Nightfirecat: Make these final and `null` the requirement type which isn't set. (skill or customRequirement, that is)

runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/DiaryRequirementsPlugin.java
Monsterxsync: lol, i knew there had to be an easier way to grab the players combat level.

PR #6281, opened 2018-10-31
Fix hiscore api after jagex reordered it, and reorder lists after update

### LINKED ISSUE #6279: Hiscorepanel shows wrong values for non skill entries after Jagex reordered them

MagicfTail:
Fixes #6279


Adam-:
Theres at least one test in the http-service which also needs to be updated. Maybe also add to it a test for the things which moved.

MagicfTail:
Updated the existing test and added new tests for the fields that moved

PR #6300, opened 2018-11-01
Add NpcCompositionChanged event and fix NPC Indicators when NPC composition changes

### LINKED ISSUE #5161: NPC indicator highlighting after NPC transforming
Currently if a tagged (from the name in settings) NPC spawns it will be highlighted. If it then proceeds to transform into another NPC which is _not_ highlighted, it will stay tagged because of its initial spawn.

This is mainly an issue in the Theatre of Blood's Nylocas room.

Expected behavior: the NPC would no longer be highligted after transforming into another NPC.

WooxSolo:
Fixes #5161

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
@@ -35,6 +35,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -311,6 +312,29 @@ public void onNpcSpawned(NpcSpawned npcSpawned)
 		}
 	}
 
+	@Subscribe
+	public void onNpcCompositionChanged(NpcCompositionChanged event)
```
deathbeam: You should probably make common method for body of this event and use it in both npcspawned and here

runelite-mixins/src/main/java/net/runelite/mixins/RSNPCMixin.java
```diff
@@ -110,6 +111,13 @@ public void onCompositionChanged(RSNPCComposition composition)
 		}
 	}
 
+	@FieldHook("composition")
+	@Inject
+	public void afterCompositionChanged(int idx)
+	{
+		client.getCallbacks().post(new NpcCompositionChanged(this));
```
Adam-: new assignment can be null on npc despawn

Adam-:
FYI this depends on !34

runelite-mixins/src/main/java/net/runelite/mixins/RSNPCMixin.java
WooxSolo: I don't understand why this would be a problem

Adam-: It fires the event for comp change, when it is really a npc despawn. Seems bad?

WooxSolo: Well technically the composition does change when it despawns. I don't have a strong opinion about it though so I guess I'll change it

Adam-:
the event part of this was added in caaf98a8597a4be0a202c49256d083e8111a8b34

PR #6301, opened 2018-11-01
NMZ Plugin: Add option to show total points rather than round points

trimbe:


deathbeam:
Maybe instead of replacing current points it should show the total points below?

Issue #6256, opened 2018-10-29
Pin RuneLite "plugin" by default


Kamielvf:
This is to help people who start using the client find the general RuneLite settings more easily. As the list is quite long and a lot of people tend to think RuneLite settings is referring to the plugin sidebar and not to the RuneLite config entry in the list.

psikoi:
I think it should be called General Settings or something too btw

pavan-p:
Why not just make a [settings button](https://i.imgur.com/7a3hvTE.png) on the title bar? Instead of considering it as a plugin.

deathbeam:
We already have settings button

Hydrox6:
@pavan-p that just splits the settings up between two places, not a good idea.

Adam-:
I would prefer to see it permanently pinned at the top of the settings panel.

Hydrox6:
@Adam- Always visible at the top when scrolling, or just as the first item in the list?

tmk2003:
I also would prefer to see it permanently pinned at the top

Adam-:
Maybe the former unless it could be visibly differentiated some better way, since no one can find it currently.

deathbeam:
They cant find it because they dont understand what are we referring to because the name is confusing.

Hydrox6:
Yeah, RuneLite settings in the settings for RuneLite is a bit weird. Renaming it to `General` or `System` would probably work better

tmk2003:
@Adam- I have this started. If this will always be pinned, should I remove the pinning star for it to remove confusion of its purpose or no?

Kamielvf:
@tmk2003 I would remove  the pinning star for it yes, since it will always be at the top. Possibly make it so that its not part of the list so that when scrolling the general settings stay at the top.

deathbeam:
I would keep pinning star so you could unpin it if you dont want it, no reason to not have that option

tmk2003:
@deathbeam that differs from what I thought @Adam- was saying, where it is always pinned.

deathbeam:
Yes, this is my opinion not Adam's

Nightfirecat:
For what it's worth, I'm also of the opinion that pinning it by default would be enough, not permanently pinning it.

tmk2003:
Are we changing the name from RuneLite for the plugin, or keeping it RuneLife for now

PR #6304, opened 2018-11-01
Enable PlugIns to be Defaulted to Pinned

tmk2003:
Plugins have the option to be defaulted as pinned (within the source code). The RuneLite Plugin currently implements this feature as per the request of #6256


raiyni:
there are a lot of extraneous comments

PR #6311, opened 2018-11-02
Convert item variations to json and add support for trident, slayer helm and same-name different id items

deathbeam:


deathbeam:
@raiyni can you check this please?

Abextm:
Shayzien armour should probably be excluded from this

deathbeam:
Well, things like (t) dhides are included, and this sort of falls into same category. And as it is optional to use, simply supporting it sort of makes sense. And for shift-click config if for some reason you want to configure shift-click action for it, it makes sense too it will apply to all shayzien armour

Abextm:
I would expect that it doesn't alias completely different items.

deathbeam:
Well they are not different, just weaker variations of it (if you mean the one that is used for shamans)

Abextm:
Should it alias a normal defence potion to a super defence potion?

deathbeam:
Dumper is now here: <https://github.com/deathbeam/runelite/tree/item-variations-dumper>. I am not removing shayzien armour because 1. too annoying to filter, 2. it is already in master

PR #6312, opened 2018-11-02
Add full special attack notifier

robinwithes:
Added a option to idle notifier to notify you when your special attack energy hits a certain threshold.


deathbeam:
I dont think this needs threshold, just notify at full? And also not sure if this should be part of idle notifier

robinwithes:
> 
> 
> I dont think this needs threshold, just notify at full? And also not sure if this should be part of idle notifier

I only used the treshold to be more consistant with the other things in idle notifier, where would you prefer this if not in idle notifier? I think it fits well enough.

Hydrox6:
Could be a cool idea to get the current weapon's special attack % and notify when it reaches that % (maybe when it reaches multiples of that percent, so I can for example see when I have 2 DWH/BGS specs available), as it seems that's what the threshold was meant to be for.

forsco:
And what if you only want it for the spec weapon in your inventory? Thats why a manual threshold is nice

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -363,6 +365,11 @@ public void onGameTick(GameTick event)
 		{
 			notifier.notify("[" + local.getName() + "] has low oxygen!");
 		}
+
+		if (checkSpecialAttack())
+		{
+			notifier.notify("[" + local.getName() + "] has reached his special attack threshold!");
```
Adam-: `"[" + local.getName() + "] has regained special attack!"`

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -411,6 +418,26 @@ private boolean checkLowHitpoints()
 		return false;
 	}
 
+	private boolean checkSpecialAttack()
+	{
+		double specialAttack = client.getVar(VarPlayer.SPECIAL_ATTACK_PERCENT) * 0.1;
+		if (config.getSpecAttackThreshold() <= 0)
```
Adam-: This can be before accessing the varp

robinwithes:
Forgot this pr and i've seen someone else already did this.

PR #6315, opened 2018-11-02
stretchedmode: reduce wait time after changing scaling percentage

devLotto:
Now it takes max 1 frame for it to resize the canvas after changing the scaling percentage, instead of at most 500ms.

It now sets the max width and height right before the canvas is resized, so the canvas size uses those max bounds immediately, instead of setting it 500ms after the canvas is resized initially.


devLotto:
Fixed a bug in the PR which caused it to sometimes not update the sizes correctly when the window was resized. Works ok now.

PR #6316, opened 2018-11-02
Changed exception handling to log warning instead of swallowing

qbolt:
Changed exception handling to log warning instead of swallowing exception by returning false

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -316,7 +316,7 @@ public class Notifier
 			}
 			catch (IOException | InterruptedException e)
 			{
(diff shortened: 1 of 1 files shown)


Adam-:
Why? This is not really an error, terminal-notifier is allowed to not be available.

qbolt:
It's not an error - It's just seemingly inconsistent with the rest of the code base which seems to log errors instead of swallowing them - just a minor enhancement to adhere to better practices for exception handling.

qbolt:
Ok, great point! - I'll update PR.

Adam-:
Probably I would do debug + also pass the exception to slf4j.

qbolt:
Ok fair enough - I'll close.

PR #6326, opened 2018-11-03
Opponent Info - Added Show opponents opponent checkbox

### LINKED ISSUE #4361: Toggle for OpponentsOpponentName
Would like a toggle showing the player being targeted by the NPC. Even though it only shows in multi, there's no need for it for solo players which most pvm is nowdays or people don't find it useful. Differing sizes of the box also makes it awkward when wanting the opponent health box in a specific position around other plugin boxes (box enlargens to fit room for player name in multi from standard size when in singles). Owner of a previous thread requesting this closed his own thread but a label tag was added. https://github.com/runelite/runelite/issues/2570

TylerHarding:
Added new checkbox that if left unchecked, will not display the opponents opponent when in multi combat areas for the opponent info overlay, acting as it does in single combat areas.

Closes #4361

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
@@ -52,4 +52,15 @@ public interface OpponentInfoConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "showOpponentsOpponent",
+		name = "Show Opponents Opponent",
+		description = "Toggle showing opponents opponent if within a multi-combat area",
+		position = 2
+	)
+	default boolean showOpponentsOpponent()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -200,7 +200,8 @@ else if (opponent instanceof Player)
 		}
 
 		// Opponents opponent
-		if (opponentsOpponentName != null)
+		// Show if checkbox is marked within plugin panel
```
Nightfirecat: This comment is probably not needed

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
```diff
@@ -52,4 +52,15 @@ default boolean showPercent()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "showOpponentsOpponent",
+		name = "Show Opponents Opponent",
```
deathbeam: Maybe `Show enemy target` would be better? Also, title casing everything in here do not fits with naming of other config options

forsco:
Isn't this a duplicate of #5808 ?

PR #6338, opened 2018-11-04
Add killed player's names to screenshot filenames.

### LINKED ISSUE #6215: Add username of player killed in the filename on auto-screenshots of kills (PVP)

critzd:
Resolves #6215.

Feature adds the killed player's name to the filename, as is seen in examples below.
Also added the full set of death messages, found [here](https://twitter.com/JagexAsh/status/1056605406947143682) via Mod Ash's twitter, credit to @NathanQuayle for pushing for that information.

### Feature Demo
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -124,9 +124,16 @@ public class ScreenshotPlugin extends Plugin
 		"You have a funny feeling like you would have been followed");
 
 	private static final ImmutableList<String> KILL_MESSAGES = ImmutableList.of("into tiny pieces and sat on them", "you have obliterated",
-		"falls before your might", "A humiliating defeat for", "With a crushing blow you", "thinking challenging you",
-		"Can anyone defeat you? Certainly", "was no match for you", "You were clearly a better fighter than", "RIP",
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -643,6 +651,33 @@ public void onResponse(Call call, Response response) throws IOException
...
+	private String killedPlayerName(String chatMessage)
+	{
+		for (String killMessage:KILL_MESSAGES)
+		{
+			while (chatMessage.contains(killMessage))
+			{
+				chatMessage = chatMessage.replace(killMessage, "");
+			}
+		}
+		for (String killMessageTrim:KILL_MESSAGE_TRIM)
```
Abextm: this won't work if the player's name contains any of these strings

critzd: Ah yeah, jumped the gun on this one thinking I had a solid solution. I'll give it another look over in the morning to fix that.

PR #6339, opened 2018-11-04
Fix Werewolf agility XP given in skill calculator

jlmcdonnell:
~~Looks like the XP for the Penguin course was copied to the Werewolf course.~~

Edit: as waycooler pointed out, Jagex have doubled the stick bonus.


waycooler:
I just made an account to fix this myself. Jagex actually doubled the exp for handing in the stick at the end of the course back on July 12th. Prior to that, the course was worth 540 total. Now it's 730, as the stick doubled from 190exp > 380exp.

PR #6347, opened 2018-11-04
wasdplugin: Show Ironman Icons when chat is locked

Hydrox6:
Has a nice side effect of not requiring input to show the icon when the chat unlocks

Before:

After:

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/IconID.java
+++ b/runelite-api/src/main/java/net/runelite/api/IconID.java
@@ -24,9 +24,14 @@
  */
 package net.runelite.api;
 
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
 /**
  * Enum of all official icons that Jagex uses in chat.
  */
+@RequiredArgsConstructor
+@Getter
 public enum IconID
 {
 	PLAYER_MODERATOR(0),
@@ -43,12 +48,7 @@ public enum IconID
 	NO_ENTRY(11),
 	CHAIN_LINK(12);
 
-	int index;
-
-	IconID(int index)
-	{
-		this.index = index;
-	}
+	private final int index;
 
 	@Override
 	public String toString()
--- a/runelite-api/src/main/java/net/runelite/api/IconID.java
+++ b/runelite-api/src/main/java/net/runelite/api/IconID.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2018, Hydrox6 <ikada@protonmail.ch>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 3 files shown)


Hydrox6:
Updated `getPlayerNameWithIcon`

runelite-api/src/main/java/net/runelite/api/IconID.java
```diff
@@ -0,0 +1,51 @@
...
+	DMM_SKULL_1_KEYS,
+	SKULL,
+	HARDCORE_IRONMAN,
+	NO_ENTRY,
+	CHAIN_LINK;
+
+	@Override
+	public String toString()
+	{
+		return "<img=" + String.valueOf(this.ordinal()) + ">";
```
deathbeam: Are you sure you want to use ordinals here? That seems like bad idea, mainly if we will be missing any icons in middle or want to add new icons at the end without adding any icons before that

runelite-api/src/main/java/net/runelite/api/IconID.java
```diff
@@ -0,0 +1,58 @@
...
+	DMM_SKULL_4_KEYS(5),
+	DMM_SKULL_3_KEYS(6),
+	DMM_SKULL_2_KEYS(7),
+	DMM_SKULL_1_KEYS(8),
+	SKULL(9),
+	HARDCORE_IRONMAN(10),
+	NO_ENTRY(11),
+	CHAIN_LINK(12);
+
+	int index;
```
deathbeam: Change this to `@Getter`

runelite-api/src/main/java/net/runelite/api/IconID.java
```diff
@@ -0,0 +1,58 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.api;
+
+/**
+ * Enum of all official icons that Jagex uses in chat.
+ */
+public enum IconID
```
deathbeam: Add `@RequiredArgsConstructor` and remove what you made

runelite-api/src/main/java/net/runelite/api/IconID.java
```diff
@@ -43,12 +48,8 @@
 	NO_ENTRY(11),
 	CHAIN_LINK(12);
 
-	int index;
-
-	IconID(int index)
-	{
-		this.index = index;
-	}
+	@Getter
```
Nightfirecat: You shouldn't need this annotation if `@Getter` is already on the class.

Hydrox6: Ah, true. Removed

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
```diff
@@ -181,9 +182,24 @@ void unlockChat()
...
-					chatboxInput.setText(client.getLocalPlayer().getName() + ": " + ColorUtil.wrapWithColorTag(client.getVar(VarClientStr.CHATBOX_TYPED_TEXT) + "*", textColor));
+					chatboxInput.setText(getPlayerNameWithIcon() + ": " + ColorUtil.wrapWithColorTag(client.getVar(VarClientStr.CHATBOX_TYPED_TEXT) + "*", textColor));
 				}
 			}
 		}
 	}
+
+	private String getPlayerNameWithIcon()
+	{
+		switch (client.getAccountType())
```
Nightfirecat: Can this be changed to determine the prefix for the player name? e.g.

```
String playerPrefix = '';
switch (client.getAccountType())
{
    case IRONMAN:
        playerPrefix = IconID.IRONMAN;
        break;
    case ULTIMATE_IRONMAN:
        playerPrefix = IconID.ULTIMATE_IRONMAN;
        break;
    case HARDCORE_IRONMAN:
        playerPrefix = IconID.HARDCORE_IRONMAN;
        break;
}
return playerPrefix + client.getLocalPlayer().getName();
```

That way, refactoring the way the prefix is gotten, or refactoring the player name call would require less modification to this method.

Hydrox6: I did this before, and DB didn't want me using `toString` in here, to just let the jvm do it. Removing the `toString` calls causes the compiler to fail, the types are incompatible and it doesn't try to `toString` unless told to.

Nightfirecat: Hm... I suppose it's fine as-is then.

Hydrox6:
For `TextTest` it doesn't really matter whether the icons are icons or not, just that they are tags that can be removed. Adding an `IconID` reference to them would look really weird whilst providing no positive effect.

runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraPlugin.java
deathbeam: No, you didnt @Hydrox6 . Just save the playerPrefix as enum and at the end you can do playerPredix + getName and it will work. And that is better that what you have now with duplicated code

PR #6361, opened 2018-11-06
zoom: Adjust inner zoom limit to safe value

### LINKED ISSUE #6358: Camera zoom plugin is causing game crash when zooming too much
As reported in Discord, when zooming in with camera zoom enabled near Chambers Of Xeric, the game crashes (but without anything in log). The sidebar is still interactable so it is just applet exploding.

[game-crash](https://gyazo.com/c619a6eac76a0b4f410daa6ef488adba)
[sidebar-working](https://gyazo.com/a4778dce062f5a0e26b55f3561a08b35)
[application.log](https://pastebin.com/ShZ8XBsL)

Abextm:
Any value greater than 1004 can cause an overflow with a large height, causing 3dzoom to be zero, crashing the renderer

Fixes #6358

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
@@ -47,6 +47,13 @@ import net.runelite.client.plugins.PluginDescriptor;
 )
 public class ZoomPlugin extends Plugin implements KeyListener
 {
+	/**
+	 * The largest (most zoomed in) value that can be used without the client crashing.
+	 *
+	 * Larger values trigger an overflow in the engine's fov to scale code.
+	 */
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -84,7 +84,7 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
 
 		if ("innerZoomLimit".equals(event.getEventName()) && zoomConfig.innerLimit())
 		{
-			intStack[intStackSize - 1] = 1200;
+			intStack[intStackSize - 1] = 1004;
```
Adam-: I would move this to a static final int with javadoc explaining why it is what it is

PR #6363, opened 2018-11-06
 runelite-client: Use Runescape themed closer for sidebar when overlaying 

Abextm:
Left: after Right: before

We need to use a different icon anyway because the one with transparency pulses on the login screen because the whole login screen doesn't get repainted every frame. Also fixes several bugs with having a customChrome config value that is different from what the window is actually.


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -180,7 +181,7 @@ public void onNavigationButtonAdded(final NavigationButtonAdded event)
 			final NavigationButton navigationButton = event.getButton();
 			final PluginPanel pluginPanel = navigationButton.getPanel();
 			final boolean inTitle = !event.getButton().isTab() &&
```
deathbeam: If you are using the withTitlebar var you created then SwingUtil.isCustomTitlePanePresent is no longer necessary

PR #6375, opened 2018-11-06
Additional flash notification options

Twiglet1022:
I wanted an even more noticeable notification so added some options to the flash notification:

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -46,11 +46,13 @@ import javax.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
+import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.client.chat.ChatColorType;
 import net.runelite.client.chat.ChatMessageBuilder;
 import net.runelite.client.chat.ChatMessageManager;
 import net.runelite.client.chat.QueuedMessage;
+import net.runelite.client.config.FlashNotification;
 import net.runelite.client.config.RuneLiteConfig;
 import net.runelite.client.ui.ClientUI;
 import net.runelite.client.util.OSType;
@@ -68,7 +70,8 @@ public class Notifier
 
 	// Notifier properties
 	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
-	private static final int FLASH_DURATION = 2000;
+	private static final int MINIMUM_FLASH_DURATION_MILLIS = 2000;
+	private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / Constants.CLIENT_TICK_LENGTH;
 
 	private final Client client;
 	private final String appName;
@@ -79,6 +82,7 @@ public class Notifier
 	private final Path notifyIconPath;
 	private final boolean terminalNotifierAvailable;
 	private Instant flashStart;
+	private long mouseLastPressedMillis;
 
 	@Inject
 	private Notifier(
@@ -146,9 +150,10 @@ public class Notifier
 				.build());
 		}
 
-		if (runeLiteConfig.enableFlashNotification())
+		if (runeLiteConfig.flashNotification() != FlashNotification.DISABLED)
 		{
 			flashStart = Instant.now();
+			mouseLastPressedMillis = client.getMouseLastPressedMillis();
 		}
 
 		log.debug(message);
@@ -156,13 +161,19 @@ public class Notifier
 
 	public void processFlash(final Graphics2D graphics)
 	{
-		if (flashStart == null || client.getGameCycle() % 40 >= 20)
-		{
-			return;
-		}
-		else if (client.getGameState() != GameState.LOGGED_IN)
+		if (flashStart == null || client.getGameState() != GameState.LOGGED_IN)
 		{
 			flashStart = null;
+			return;
+		}
+
+		FlashNotification flashNotification = runeLiteConfig.flashNotification();
+
+		if (client.getGameCycle() % 40 >= 20
+			// For solid colour, fall through every time.
+			&& (flashNotification == FlashNotification.FLASH_TWO_SECONDS
+			|| flashNotification == FlashNotification.FLASH_UNTIL_CANCELLED))
+		{
 			return;
 		}
 
@@ -171,9 +182,27 @@ public class Notifier
 		graphics.fill(new Rectangle(client.getCanvas().getSize()));
 		graphics.setColor(color);
 
-		if (Instant.now().minusMillis(FLASH_DURATION).isAfter(flashStart))
+		if (!Instant.now().minusMillis(MINIMUM_FLASH_DURATION_MILLIS).isAfter(flashStart))
 		{
-			flashStart = null;
+			return;
+		}
+
+		switch (flashNotification)
+		{
+			case FLASH_TWO_SECONDS:
+			case SOLID_TWO_SECONDS:
+				flashStart = null;
+				break;
+			case SOLID_UNTIL_CANCELLED:
+			case FLASH_UNTIL_CANCELLED:
+				// Any interaction with the client since the notification started will cancel it after the minimum duration
+				if (client.getMouseIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
+					|| client.getKeyboardIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
+					|| client.getMouseLastPressedMillis() > mouseLastPressedMillis)
+				{
+					flashStart = null;
+				}
+				break;
 		}
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/config/FlashNotification.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/FlashNotification.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2019, Twiglet1022 <https://github.com/Twiglet1022>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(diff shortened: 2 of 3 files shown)


Twiglet1022:
I've redone this PR in a better way. Instead of supplying an input box for customising the duration of the flash notification I've just added an option to have it flash until you notice it which is what I was going for originally anyway.

Also added an option to have to have it stay a solid red instead of flashing.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -68,7 +69,9 @@
 
 	// Notifier properties
 	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
-	private static final int FLASH_DURATION = 2000;
+	private static final int MINIMUM_FLASH_DURATION_MILLIS = 2000;
+	private static final int MINIMUM_FLASH_DURATION_TICKS = 33; // 33 = 2000 / 60 rounded down
```
Nightfirecat: This doesn't look correct. Per documentation for `client.getMouseIdleTicks()` and `client.getKeyboardIdleTicks()`, those values are measured in client ticks, not game ticks, which are defined in Constants to be 20ms in length. Additionally, this should be `private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / 60;` (or with 60 replaced with the appropriate constant reference) so that the comment isn't needed.

Twiglet1022: Ah... good spot, yes you’re right. I’ll fix that later thanks!

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -68,7 +69,9 @@
 
 	// Notifier properties
 	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
-	private static final int FLASH_DURATION = 2000;
+	private static final int MINIMUM_FLASH_DURATION_MILLIS = 2000;
+	private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / 20;
```
Nightfirecat: This should use `Constants.CLIENT_TICK_LENGTH`.
```suggestion
	private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / Constants.CLIENT_TICK_LENGTH;
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -169,10 +178,27 @@ else if (client.getGameState() != GameState.LOGGED_IN)
 		graphics.fill(new Rectangle(client.getCanvas().getSize()));
 		graphics.setColor(color);
 
-		if (Instant.now().minusMillis(FLASH_DURATION).isAfter(flashStart))
+		// Enforce a minimum flash duration of 2 seconds so as not to cancel it too quickly
+		if (!Instant.now().minusMillis(MINIMUM_FLASH_DURATION_MILLIS).isAfter(flashStart))
```
Nightfirecat: This whole block is pretty confusing and not very straightforward. I would recommend rewriting it to be more like the following so it is easier to differentiate between the guard clause and the code which is actually updating the flash timer.

```java
if (!Instant.now().minusMillis(MINIMUM_FLASH_DURATION_MILLIS).isAfter(flashStart))
{
    return;
}

switch (notificationFlashType)
{
	case FLASH_TWO_SECONDS:
	case SOLID_TWO_SECONDS:
		flashStart = null;
		break;
	case SOLID_UNTIL_CANCELLED:
	case FLASH_UNTIL_CANCELLED:
	{
		if (client.getMouseIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
			|| client.getKeyboardIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
			|| client.getMouseLastPressedMillis() > mouseLastPressedMillis)
		{
            flashStart = null;
		}
		break;
	}
}
```

runelite-client/src/main/java/net/runelite/client/config/NotificationFlashType.java
```diff
@@ -0,0 +1,47 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
Nightfirecat: No newline between license and package definition
```suggestion
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -169,10 +178,27 @@ else if (client.getGameState() != GameState.LOGGED_IN)
 		graphics.fill(new Rectangle(client.getCanvas().getSize()));
 		graphics.setColor(color);
 
-		if (Instant.now().minusMillis(FLASH_DURATION).isAfter(flashStart))
+		// Enforce a minimum flash duration of 2 seconds so as not to cancel it too quickly
```
Nightfirecat: I would recommend against explicitly calling out the duration as 2 seconds, as that would become incorrect if `MINIMUM_FLASH_DURATION_MILLIS` were ever to change in the future.

runelite-client/src/main/java/net/runelite/client/Notifier.java
Twiglet1022: Yea this is much a clearer way of doing it, thanks.

runelite-client/src/main/java/net/runelite/client/Notifier.java
Twiglet1022: I also probably shouldn't have made the "Flash for 2 seconds" and "Solid for 2 seconds" options rely on MINIMUM_FLASH_DURATION_MILLIS to be 2000 so I've also reordered things a little.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -169,9 +180,30 @@ else if (client.getGameState() != GameState.LOGGED_IN)
 		graphics.fill(new Rectangle(client.getCanvas().getSize()));
 		graphics.setColor(color);
 
-		if (Instant.now().minusMillis(FLASH_DURATION).isAfter(flashStart))
+		switch (notificationFlashType)
 		{
-			flashStart = null;
+			case FLASH_TWO_SECONDS:
+			case SOLID_TWO_SECONDS:
+				if (Instant.now().minusMillis(2000).isAfter(flashStart))
```
Nightfirecat:
```suggestion
				if (Instant.now().minusMillis(MINIMUM_FLASH_DURATION_BEFORE_CANCELING_MILLIS).isAfter(flashStart))
```

Nightfirecat: Also, this check is used in all branches of this switch statement. Is there any reason it's not a guard clause before the switch starts?

Twiglet1022: The config options are "Flash/Solid for two seconds" so I figured I probably shouldn't use the minimum duration and assume it will be 2 seconds because that constant could be updated.

That being said, I don't think there's much reason to ever change that so perhaps I could just call the constant `TWO_SECOND_MINIMUM_FLASH_DURATION_MILLIS` and make a guard clause out of it?

Nightfirecat: I think it's most likely that if it changes, it would change to become some value we look up from user config, so it's probably fine to reference the constant there.

runelite-client/src/main/java/net/runelite/client/config/NotificationFlashType.java
```diff
@@ -0,0 +1,46 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.config;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum NotificationFlashType
```
Twiglet1022: Would it be preferable to condense "Enable flash notification" and "Flash notification type" into a single config? I think it would be cleaner, especially since the runelite config is already a bit long. Downside though is that as one-off, everyone's flash notification will be turned off until an option is selected so idk.

Twiglet1022:
The config changed from a boolean to an enum so I've added a method to migrate the old setting to the new one such that it preserves whatever behaviour the user had previously.

Possibly a bit overkill for something so minor? But it's a separate commit so can be split out if unwanted.

I tested all 3 cases with the previous config (true, false, unset) to check that it works.

Adam-:
This is a rarely used feature I'd be okay just ditching this migrating code to make it less complex.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -68,7 +76,9 @@
 
 	// Notifier properties
 	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
-	private static final int FLASH_DURATION = 2000;
+	private static final int MINIMUM_FLASH_DURATION_MILLIS = 2000;
+	private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / Constants.CLIENT_TICK_LENGTH;
+	private FlashNotification flashNotification;
```
Adam-: You should just be able to access the config directly from `processFlash` instead of storing this.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -146,23 +161,29 @@ public void notify(String message, TrayIcon.MessageType type)
...
 	{
-		if (flashStart == null || client.getGameCycle() % 40 >= 20)
+		if (flashStart == null || client.getGameState() != GameState.LOGGED_IN)
 		{
+			flashStart = null;
 			return;
 		}
-		else if (client.getGameState() != GameState.LOGGED_IN)
+
+		if (client.getGameCycle() % 40 >= 20
```
Adam-: This code is what adds the flash effect, so it makes more sense to check `flashNotification` is one of the flash types instead of not all of the non flash types.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -171,9 +192,27 @@ else if (client.getGameState() != GameState.LOGGED_IN)
...
 		{
-			flashStart = null;
+			return;
+		}
+
+		switch (flashNotification)
+		{
+			case FLASH_TWO_SECONDS:
+			case SOLID_TWO_SECONDS:
+					flashStart = null;
```
Adam-:
```suggestion
				flashStart = null;
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
Twiglet1022: I stored it because otherwise it would be reading it from config every frame. So I think this way is better. What do you think?

Adam-: It looks to me like it would only be reading from config each frame when it was flashing, but not otherwise.

Twiglet1022: Yes that's right only when it's flashing, so I guess it's not that bad. There is also a slight (and this may be the epitome of slightness) advantage in that you can update the setting while it's flashing and have it take affect immediately.

I went ahead and changed it as you suggested.

PR #6383, opened 2018-11-08
Correct some explanatory comments in runelite api

Twiglet1022:
While working with getMouseIdleTicks(), getKeyboardIdleTicks() and getMouseLastPressedMillis() I noticed that the comments were incorrect or misleading so I've clarified them.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -971,19 +971,20 @@ public interface Client extends GameEngine
 	BufferProvider getBufferProvider();
 
 	/**
-	 * Gets the amount of ticks since the last mouse movement occurred.
+	 * Gets the amount of client ticks {@link ClientTickLength} since the last mouse movement occurred.
 	 *
 	 * @return amount of idle mouse ticks
 	 */
 	int getMouseIdleTicks();
 
 	/**
-	 * Gets the number of milliseconds since the last mouse press occurred.
+	 * Gets the time at which the last mouse press occurred in milliseconds since
+	 * the UNIX epoch.
 	 */
 	long getMouseLastPressedMillis();
 
 	/**
-	 * Gets the amount of ticks since the last keyboard press occurred.
+	 * Gets the amount of client ticks {@link ClientTickLength} since the last keyboard press occurred.
 	 *
 	 * @return amount of idle keyboard ticks
 	 */
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -971,9 +971,10 @@ public interface Client extends GameEngine
 	BufferProvider getBufferProvider();
 
 	/**
-	 * Gets the amount of client ticks {@link ClientTickLength} since the last mouse movement occurred.
+	 * Gets the amount of client ticks since the last mouse movement occurred.
 	 *
 	 * @return amount of idle mouse ticks
+	 * @see Constants#CLIENT_TICK_LENGTH
 	 */
 	int getMouseIdleTicks();
 
@@ -984,9 +985,10 @@ public interface Client extends GameEngine
 	long getMouseLastPressedMillis();
 
 	/**
-	 * Gets the amount of client ticks {@link ClientTickLength} since the last keyboard press occurred.
+	 * Gets the amount of client ticks since the last keyboard press occurred.
 	 *
 	 * @return amount of idle keyboard ticks
+	 * @see Constants#CLIENT_TICK_LENGTH
 	 */
 	int getKeyboardIdleTicks();
 
--- a/runelite-api/src/main/java/net/runelite/api/Constants.java
+++ b/runelite-api/src/main/java/net/runelite/api/Constants.java
@@ -36,30 +36,37 @@ public class Constants
 	 * The original width of the game when running in fixed mode.
 	 */
 	public static final int GAME_FIXED_WIDTH = 765;
+
 	/**
 	 * The original height of the game when running in fixed mode.
 	 */
 	public static final int GAME_FIXED_HEIGHT = 503;
+
 	/**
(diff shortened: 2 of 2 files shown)


Abextm:
`client ticks` is the term we use for this

Twiglet1022:
> `client ticks` is the term we use for this

Now that you mention it I see both game cycles and client ticks being used to refer to the same thing. I feel I've clarified any potential confusion by including 20ms in the comment but can update it to say client ticks instead if that's preferred.

Abextm:
I would prefer `client ticks` because some people use `game cycles` for server ticks too.

Nightfirecat:
Would it hurt to establish a `CLIENT_TICK_LENGTH` static var somewhere with a javadoc comment explaining it is roughly 20ms in length? Seems that would help in this case, where a `{@link CLIENT_TICK_LENGTH}` annotation could be inserted.

Twiglet1022:
> Would it hurt to establish a `CLIENT_TICK_LENGTH` static var somewhere with a javadoc comment explaining it is roughly 20ms in length? Seems that would help in this case, where a `{@link CLIENT_TICK_LENGTH}` annotation could be inserted.

Is this what you meant? I'm not really sure what I'm doing. This looks like it works, but feels a bit wrong.

I tried adding it to constants first though because that seemed like a reasonable place to put it, but I couldn't get `{@link CLIENT_TICK_LENGTH}` to work because it seems I can't link directly to a variable within a class.

devLotto:
Client ticks are only 20ms at 50fps. At lower fps they are longer.

Twiglet1022:
So a client tick is basically the length of a frame?

Nightfirecat:
@Twiglet1022 I'd be fine seeing this move to an existing constants file (if there's an appropriate one) and simply referencing the var name within `{@link ClassName}`.

Twiglet1022:
@Nightfirecat I went with your suggestion. The existing Constants class seemed appropriate to me given the description is ``A utility class containing constant values.`` and the rest of the constants in there refer to the client in some way too.

I got in a bit of a muddle with trying to revert to an earlier commit but I think it's okay now looking at the files changed.

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -940,19 +940,22 @@
...
+	 * since the last mouse movement occurred.
 	 *
 	 * @return amount of idle mouse ticks
 	 */
 	int getMouseIdleTicks();
 
 	/**
-	 * Gets the number of milliseconds since the last mouse press occurred.
+	 * Gets the time at which the last mouse press occurred in milliseconds since
+	 * 1st January 1970 00:00
```
Nightfirecat: I'd be fine with simply referring to this as the UNIX epoch, since that's fairly well-understood by people who deal with time in that format.

runelite-api/src/main/java/net/runelite/api/Constants.java
```diff
@@ -67,4 +67,8 @@
 	 * the plane value to 0-3.
 	 */
 	public static final int MAX_Z = 4;
+	/**
+	 * The number of milliseconds in a client tick.
```
Nightfirecat: Might be worth mentioning that this is the length of one frame when the client is running at its max framerate of 50fps.

Twiglet1022:
Should be clearer now

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -948,8 +948,8 @@
 	int getMouseIdleTicks();
 
 	/**
-	 * Gets the time at which the last mouse press occurred in milliseconds since
-	 * 1st January 1970 00:00
+	 * Gets the time at which the last mouse press occurred in UNIX time measured in
```
Nightfirecat: I would've called it "milliseconds since the UNIX epoch".

Twiglet1022: There we go

Twiglet1022:
Oops must have misclicked on close

deathbeam:
Needs rebase

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -940,19 +940,22 @@
...
 	/**
-	 * Gets the number of milliseconds since the last mouse press occurred.
+	 * Gets the time at which the last mouse press occurred in milliseconds since
+	 * the UNIX epoch.
 	 */
 	long getMouseLastPressedMillis();
 
 	/**
-	 * Gets the amount of ticks since the last keyboard press occurred.
+	 * Gets the amount of client ticks (see CLIENT_TICK_LENGTH in {@link Constants})
```
deathbeam: You can just use @see Constants#CLIENT_TICK_LENGTH

runelite-api/src/main/java/net/runelite/api/Constants.java
```diff
@@ -67,4 +67,11 @@
 	 * the plane value to 0-3.
 	 */
 	public static final int MAX_Z = 4;
+	/**
```
deathbeam: Add new line above

runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -940,19 +940,22 @@
 	BufferProvider getBufferProvider();
 
 	/**
-	 * Gets the amount of ticks since the last mouse movement occurred.
+	 * Gets the amount of client ticks (see CLIENT_TICK_LENGTH in {@link Constants})
```
deathbeam: You can just use @see Constants#CLIENT_TICK_LENGTH

runelite-api/src/main/java/net/runelite/api/Constants.java
Twiglet1022: The rest of the Constants.java file was also formatted without new lines between each java doc comment so I added them into the whole file. I presume that's preferred?

Twiglet1022:
Rebased and updated.

PR #6393, opened 2018-11-08
Add AMMO(13) to EquipmentInventorySlot

deathbeam:


arthurtyukayev:
It would be a good idea to also add 13 to KitType. I've done it in my ammo plugin but I need to fix a few things and this maybe merged in sooner.

deathbeam:
I have no clue what KitType is or what is the difference so i dont want to mess with that one

PR #6399, opened 2018-11-08
Fixed maniacal monkey hunting trap bug

### LINKED ISSUE #6269: Edge "Large boulder"s at Maniacal Monkeys Don't Trigger Hunter Plugin
The hunter plugin (trap indicator) seems to work fine on some rocks but not others.

Working

Not Working

MaxBartlett:
Closes #6269 

As far as I can tell, this bug occurred because when the player's distance to the trap was greater than 1 tile, it didn't register the trap as being set by the player. As the boulders are 2x2 objects, the threshold needs to be set to 2 tiles.


deathbeam:
This will cause that when you are standing next to one boulder the other near one will be falsely trigerred

MaxBartlett:
> This will cause that when you are standing next to one boulder the other near one will be falsely trigerred

I just tested this for all the boulders and this issue never came up.

PR #6400, opened 2018-11-08
Add notification option for valuable and untradeable items

Reasel:
This adds an option to notify player when they receive a valuable drop or untradeable drop as configured by the in game options.

Also this will fix the issue where chat notifications would send game messages as well when the option was selected resulting in double game messages.

Settings

[Demo](https://i.imgur.com/hogEsYD.gif)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -105,12 +105,22 @@ public class Notifier
 		storeIcon();
 	}
 
+	public  void notify(String message, boolean shouldPrint)
+	{
+		notify(message, TrayIcon.MessageType.NONE, shouldPrint);
+	}
+
 	public void notify(String message)
 	{
-		notify(message, TrayIcon.MessageType.NONE);
+		notify(message, TrayIcon.MessageType.NONE, true);
 	}
 
 	public void notify(String message, TrayIcon.MessageType type)
+	{
+		notify(message, type, true);
+	}
+
+	public void notify(String message, TrayIcon.MessageType type, boolean shouldPrint)
 	{
 		if (!runeLiteConfig.sendNotificationsWhenFocused() && clientUI.isFocused())
 		{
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -105,12 +105,12 @@ private Notifier(
 		storeIcon();
 	}
 
-	public void notify(String message)
+	public void notify(String message, boolean... shouldPrint)
```
deathbeam: This is not how you do optional arguments....

runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
```diff
@@ -137,13 +138,22 @@ public void onSetMessage(SetMessage event)
...
 				break;
 			case DUEL:
 				if (event.getValue().contains("wishes to duel with you.") && config.notifyOnDuel())
 				{
-					notifier.notify(event.getValue());
+					notifier.notify(event.getValue(), false);
+				}
+				break;
+			case SERVER:
+				if ((config.notifyOnValuable() && event.getValue().contains("Valuable drop:")) || (config.notifyOnUntradeable() && event.getValue().contains("Untradeable drop:")))
```
deathbeam: Why contains? It should start with that message.

runelite-client/src/main/java/net/runelite/client/Notifier.java
deathbeam: Also I really dislike how you decided to special case chat notifier

Reasel: It is a sort of workaround for the lack of optional parameters in Java. 

As for the special case what do you mean? Currently if I have chat notifier set up and also have "Enable game message notification" enabled all messages that trigger a notification  duplicated in chat.

Reasel: Alright fixed it.

PR #6409, opened 2018-11-09
Unlimit stretched resizable mode and properly limit scaling factor

deathbeam:
- The limits are no longer necessary as it is limited anyway to fixed mode
base size, and with new method of scaling the 100% max is quite
limiting.
- Instead of limiting width and height when stretching to fixed mode size,
limit scaling factor to fixed mode size to prevent improportional
stretching in resizable mode.


devLotto:
100% default scaling is a little too much on 1080p

deathbeam:
Now that I actually tested it, yeah you are right, 100% default is too much, changed it back to 50

PR #6411, opened 2018-11-09
Theater of Blood Damage Recount

akarhi:
This plugin will only count the damage you have done to (example) Bloat for that fight 500 damage. There is only two message when you die or survive the fight. The information will be lost if you disconnect/xlog the client during combat. The damage calculation is an estimate not 100% accurate most likely around 95% accurate.

two picture of examples how it should look like after a boss fight/death
If you survived*
javaw_2018-11-11_02-14-47
If you died*
javaw_2018-11-11_02-44-35


raiyni:
what's the value in this

akarhi:
Lets you see how much damage you done in ToB, could also help improve person to not miss a game tick when attack for example like bloat or be a little niche update

akarhi:
added a damage taken counter, going to "fix" the grammar issues more make it like user friendly but not too nooby for pro players etc
picture of the damage taken, given and death
javaw_2018-11-11_01-49-41

PR #6416, opened 2018-11-10
Improve transparent chatbox performance

WooxSolo:
Currently when transparent chatbox is enabled, Jagex renders the dark background at 2 alpha at a time and the white lines at 5-6 alpha at a time. My fix will make it draw each pixel with an appropriate alpha value to not have to render the same pixel multiple times. This should increase overall performance of the game by 5-10% (rough numbers, depends on client size and other factors) when transparent chatbox is enabled. However this will also change the shade of the chatbox slightly, but will probably be unnoticeable for most people.


Adam-:
This was taken in e603a2a15729e4c633a4f745b2b7f7963ecb89d3

PR #6426, opened 2018-11-11
Add real-time clock plugin

c01dc0ffee:
Adds a convenient clock overlay to the client. Supports milliseconds, 12 and 24 hour formats, and 3 font selections.

clock image


deathbeam:
Or report button plugin.

PR #6433, opened 2018-11-11
adds clock tag to reportbutton plugin so people find this setting easier and a 24Hour option

Malfuryent:
adds a "clock" tag to the reportbutton plugin to make the clock easier to find/adjust in the plugin list, i guess i wasnt the only one trying to find this/adjust it and didnt find the option easily.

Also adds the option to show time in a 24Hour time format


Malfuryent:
removed the plugin, instead this PR adds "clock" to the tags in the reportbutton plugin so it's easier to find.

Malfuryent:
added military time format to the reportbutton plugin as a toggle for people who grew up with that kind of time format like me. makes sense to have this option for a clock.

PR #6436, opened 2018-11-12
Make Grand Exchange plugin panel background color consistent

Arrexel:
This PR changes the background color of all panels for the GE plugin to match the rest of RuneLite's plugins. It is the only one using medium_gray rather than darker_gray.


raiyni:
Have a pre-post picture?

Arrexel:
Original
before1
before2

Fixed
after1
after2

PR #6443, opened 2018-11-12
Fix size of RuneLite minimap dots

deathbeam:
The dots are supposed to be size of 4 and not 5.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayUtil.java
@@ -47,6 +47,8 @@ import net.runelite.api.coords.LocalPoint;
  */
 public class OverlayUtil
 {
+	private static final int MINIMAP_DOT_RADIUS = 4;
+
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayUtil.java
```diff
@@ -61,9 +61,9 @@ public static void renderPolygon(Graphics2D graphics, Polygon poly, Color color)
 	public static void renderMinimapLocation(Graphics2D graphics, Point mini, Color color)
 	{
 		graphics.setColor(Color.BLACK);
-		graphics.fillOval(mini.getX() - 2, mini.getY() - 2 + 1, 5, 5);
```
Nightfirecat: Let's make this a constant.

Nightfirecat:
This had bothered me for a while now. I'll test this to see that it exactly overlays current minimap dots, since that's what I'd worry about with this change.

Nightfirecat:
Looks like it's not perfect (since arc drawing is dithered), but I don't think that's noteworthy. LGTM, pending magic numbers becoming a constant.

PR #6444, opened 2018-11-12
ArdougneCloak Menu Entry Swapper

KasparRosin:


deathbeam:
For stuff like this we have shift-click configuration

KasparRosin:
> 
> 
> For stuff like this we have shift-click configuration
Shift click is something i don't like or/and i'm not used to it for some reason. I'd much rather enjoy just clicking it without using my keyboard and I thought maybe someone else would like it too.

deathbeam:
Use on-screen keyboard and just press shift there then and it will stay pressed

KasparRosin:
There are other items that would be affected by if shift stays pressed.
I mean, you can close this PR, but since there is an off option i don't see how it could hurt, maybe i am missing something. I can do that **off** is default.

deathbeam:
Doesnt matter if it is on or off by default, it is not necessary when you can already achieve it with shift-click config. And if we would start adding swaps like this for specific items to menu entry swapper just because people are uncomfortable of using shift-click then the PRs would never stuff and there would be hundreds of toggles inside the plugin.

Nightfirecat:
Given that our goal is (as far as I'm aware, anyway) to remove the left-click default swaps eventually, I feel we shouldn't be adding more.

KasparRosin:
Closed reason:
not necessary

Issue #6427, opened 2018-11-11
Add a clear all option to the inventory tags plugin


casper20051:
![inventorytags](https://user-images.githubusercontent.com/3000350/48310498-46a97f80-e55e-11e8-8462-d87347e030c1.jpg)
Instead of having to remove all inventory tags one item at time, add an option to remove all of them.

Nightfirecat:
@casper20051 Why did you close this issue? A fix for this request hasn't been merged yet...

PR #6450, opened 2018-11-13
Add clear all option for inventory tags

nhuerta53:
Add a way to clear all inventory tags per issue#6427. I attached a gif showing the process of adding multiple tags and removing them with the clear all option. 
inventory_tag_clear

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Provides;
 import java.awt.Color;
+import java.util.ArrayList;
 import java.util.List;
 import javax.inject.Inject;
 import lombok.AccessLevel;
@@ -73,6 +74,7 @@ public class InventoryTagsPlugin extends Plugin
 	private static final String MENU_TARGET = "Inventory Tags";
 	private static final String MENU_SET = "Mark";
 	private static final String MENU_REMOVE = "Remove";
+	private static final String MENU_CLEAR = "Clear";
 
 	private static final WidgetMenuOption FIXED_INVENTORY_TAB_CONFIGURE = new WidgetMenuOption(CONFIGURE,
 		MENU_TARGET, WidgetInfo.FIXED_VIEWPORT_INVENTORY_TAB);
@@ -88,6 +90,7 @@ public class InventoryTagsPlugin extends Plugin
 		MENU_TARGET, WidgetInfo.RESIZABLE_VIEWPORT_BOTTOM_LINE_INVENTORY_TAB);
 
 	private static final List<String> GROUPS = ImmutableList.of(SETNAME_GROUP_4, SETNAME_GROUP_3, SETNAME_GROUP_2, SETNAME_GROUP_1);
+    private static List<String> itemIdsList = new ArrayList<>();
 
 	@Inject
 	private Client client;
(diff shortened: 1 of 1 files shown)


deathbeam:
This closes #6427

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
```diff
@@ -131,14 +134,24 @@ String getTag(int itemId)
...
 		configManager.setConfiguration(InventoryTagsConfig.GROUP, ITEM_KEY_PREFIX + itemId, tag);
 	}
 
 	private void unsetTag(int itemId)
 	{
 		configManager.unsetConfiguration(InventoryTagsConfig.GROUP, ITEM_KEY_PREFIX + itemId);
+		itemIdsList.remove(ITEM_KEY_PREFIX + itemId);
 	}
 
+	private void clearTag()
```
deathbeam: this should be probably named clearTags()

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
```diff
@@ -131,14 +134,24 @@ String getTag(int itemId)
 
 	private void setTag(int itemId, String tag)
 	{
+		itemIdsList.add(ITEM_KEY_PREFIX + itemId);
```
deathbeam: You should maybe store just itemId and then prepend prefix in clearTag method

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
```diff
@@ -131,14 +134,24 @@ String getTag(int itemId)
...
 		configManager.setConfiguration(InventoryTagsConfig.GROUP, ITEM_KEY_PREFIX + itemId, tag);
 	}
 
 	private void unsetTag(int itemId)
 	{
 		configManager.unsetConfiguration(InventoryTagsConfig.GROUP, ITEM_KEY_PREFIX + itemId);
+		itemIdsList.remove(ITEM_KEY_PREFIX + itemId);
```
deathbeam: Same here

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
```diff
@@ -212,23 +231,49 @@ public void onMenuOpened(final MenuOpened event)
...
 			int num = 0;
 
 			for (final String groupName : GROUPS)
 			{
 				final String group = getTag(itemId);
 				final MenuEntry newMenu = new MenuEntry();
 				final Color color = getGroupNameColor(groupName);
-				newMenu.setOption(groupName.equals(group) ? MENU_REMOVE : MENU_SET);
-				newMenu.setTarget(ColorUtil.prependColorTag(groupName, MoreObjects.firstNonNull(color, Color.WHITE)));
+				if (groupName.equals(group))
```
deathbeam: You should probably add the clear to the inventory button and not to each item in inventory

nhuerta53: When I tried adding the clear button to the inventory button, the clear button was always above the save button because it displays the options in sorted order. I asked the devchat on discord and it was suggested that I add the clear tags option when shift clicking on a tagged item. I'm going to work on that if theres no better way to get it on the inventory button.

deathbeam: Imo it is fine when it is sorted. Adding clear tags option on each tagged item is worse.

nhuerta53: Alright, I'll get to work on getting that working. I'm just concerned there might be some people mis-clicking clear instead of save.

forsco: You could make it so you have to confirm the clear by pressing clear again with a prompt message in the chatbox.

Just a suggestion, not sure how people would feel about that.

nhuerta53: So, after doing additional testing it appears that the menu options are not displayed in sorted order but instead they tend to switch places randomly. This will only occur when the client is restarted though.  They will remain in the same place while the client remains open but once its closed and reopened there is a chance that the positions will change. Not sure where to go from here.

PR #6457, opened 2018-11-13
Improve responsiveness of fishing idle notification

Twiglet1022:
Using interactions instead of animations detects idling while fishing much sooner because fishing animations are incredibly long and can persists for up to 10 seconds beyond the point when the character has stopped fishing.

Here is a video demonstration of the new notification while barehanded fishing sharks. You can see how the  animation continues long after fishing has stopped.

https://www.youtube.com/watch?v=g5IkN4N8GtM

Example with cage fishing for lobsters too:

https://www.youtube.com/watch?v=Ju5m9RSAYMQ

One further change I made was to improve the check that prevents idle notifications from being sent while you're actually playing the game.

At the start of the onGameTick method in there is a check that includes checking for whether the user has clicked recently. If the user has clicked recently then the method is exited and no notification is sent because obviously they are not idle.

The problem is that it checks if the user has clicked within the last 200 ms but game ticks only run every 600 ms. While in debug mode I found that there can be gaps of up to 800 ms between onGameTick running during the normal course of play.

This video demonstrates how idle fishing notifications are sometimes sent when clicking away from the fishing spot, but that extending the window from 200 ms to 1000 ms solves this.

https://www.youtube.com/watch?v=sQC_PRidlo0

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -43,12 +43,12 @@ public interface IdleNotifierConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "combatidle",
-		name = "Combat Idle Notifications",
-		description = "Configures if out of combat notifications are enabled",
+		keyName = "interactionidle",
+		name = "Idle Interaction Notifications",
+		description = "Configures if idle interaction notifications are enabled e.g. combat, fishing",
 		position = 2
 	)
-	default boolean combatIdle()
+	default boolean interactionIdle()
 	{
 		return true;
 	}
@@ -107,16 +107,4 @@ public interface IdleNotifierConfig extends Config
 	{
 		return 0;
 	}
-
-	@ConfigItem(
-		keyName = "fishingIdle",
-		name = "Fishing Idle Notifications",
-		position = 8,
-		description = "Configures if fishing idle notifications are enabled."
-	)
-	default boolean fishingIdle()
-	{
-		return true;
-	}
-
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -108,4 +108,15 @@ public interface IdleNotifierConfig extends Config
 		return 0;
 	}
 
+	@ConfigItem(
+		keyName = "fishingIdle",
+		name = "Fishing Idle Notifications",
+		position = 8,
+		description = "Configures if fishing idle notifications are enabled."
+	)
+	default boolean fishingIdle()
+	{
+		return true;
+	}
+
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -92,7 +92,7 @@ public class IdleNotifierPlugin extends Plugin
 	private int lastCombatCountdown = 0;
 	private Instant sixHourWarningTime;
 	private boolean ready;
-	private Instant lastInteractingFishing;
+	private boolean lastInteractWasCombat;
 
 	@Provides
 	IdleNotifierConfig provideConfig(ConfigManager configManager)
@@ -265,12 +265,15 @@ public class IdleNotifierPlugin extends Plugin
 			resetTimers();
 			lastInteract = target;
 			lastInteracting = Instant.now();
+			lastInteractWasCombat = true;
 		}
-
-		if (target.getName() != null && target.getName().contains(FISHING_SPOT))
+		else if (target.getName() != null && target.getName().contains(FISHING_SPOT))
 		{
+			// Player is fishing
 			resetTimers();
-			lastInteractingFishing = Instant.now();
+			lastInteract = target;
+			lastInteracting = Instant.now();
+			lastInteractWasCombat = false;
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


Twiglet1022:
Is it possible that the problem is actually with the unit tests? I've worked out how to resolve all the problems that are causing the unit tests to fail, but I don't think these "fixes" should be implemented.

1. A NullPointerException on line 270 should be impossible because line 253 prevents the variable target from being null. If I stick a try catch around it the unit test passes.
2. The NullPointerException on line 617 should also impossible because of the null check on line 616 which will prevent line 617 from being evaluated if there is a null. Again, sticking a try catch around it makes the unit test pass.
3. Finally the unit test specifies that client.getKeyboardIdleTicks will be 42, so setting the check on line 334 to 50 as I have done causes the unit test to failed due to a "Wanted but not invoked". Setting this value to 42 or less, and removing the check on client.getMouseIdleTicks will make the unit test pass. The 42 just seems completely arbitrary though.

Twiglet1022:
Fixed it... by editing the test cases. Is that allowed?

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -331,8 +329,10 @@ public void onGameTick(GameTick event)
 
 		if (client.getGameState() != GameState.LOGGED_IN
 			|| local == null
-			|| System.currentTimeMillis() - client.getMouseLastPressedMillis() < 200
-			|| client.getKeyboardIdleTicks() < 10)
+			// If user has interacted with game in last second then they're not idle so don't send idle notification
+			|| System.currentTimeMillis() - client.getMouseLastPressedMillis() < 1000
+			|| client.getKeyboardIdleTicks() < 50
+			|| client.getMouseIdleTicks() < 50)
```
Adam-: Are you sure about this mouse idle ticks? I seem to recall @deathbeam trying this before and it breaking everything.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -331,8 +329,10 @@ public void onGameTick(GameTick event)
 
 		if (client.getGameState() != GameState.LOGGED_IN
 			|| local == null
-			|| System.currentTimeMillis() - client.getMouseLastPressedMillis() < 200
-			|| client.getKeyboardIdleTicks() < 10)
+			// If user has interacted with game in last second then they're not idle so don't send idle notification
+			|| System.currentTimeMillis() - client.getMouseLastPressedMillis() < 1000
```
Adam-: Is it necessary to change this and the keyboard timeout?

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
Twiglet1022: I have been fishing sacred eels with this version for a few days with no issues, however I haven’t done any other activities. I could remove this to be on the safe side. It’s probably a little overkill in all honesty with the check for clicks and keyboard presses already in there.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
Twiglet1022: The reason I changed it is that they were both too short before to make any difference about 2/3rds of the time. Game ticks are every 600ms (and it seems the onGameTick method actually only runs every 600-800ms unless debug mode slows it down) so most of the time it would notify you even if you were the one to click away from the fishing spot.

Although keyboard isn’t really used while fishing it made sense that it would have the same issue as mouse clicks so I did both.

Twiglet1022: I've put the keyboard timeout and mouse idle ticks timeout back (along with the unit test) and tidied up the commit history.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -108,4 +108,15 @@ default int getOxygenThreshold()
 		return 0;
 	}
 
+	@ConfigItem(
```
deathbeam: I dont think this needs separate toggle

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -274,6 +266,12 @@ public void onInteractingChanged(InteractingChanged event)
 			lastInteract = target;
 			lastInteracting = Instant.now();
 		}
+
+		if (target.getName() != null && target.getName().contains(FISHING_SPOT))
```
deathbeam: use else if

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -372,6 +371,11 @@ public void onGameTick(GameTick event)
 		{
 			notifier.notify("[" + local.getName() + "] has low oxygen!");
 		}
+
+		if (config.fishingIdle() && checkFishingIdle(waitDuration, local))
```
deathbeam: This should probably run right after animation idle, or inside it

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -372,6 +371,11 @@ public void onGameTick(GameTick event)
...
+		{
+			notifier.notify("[" + local.getName() + "] has stopped fishing!");
```
deathbeam: You should probably still use same message as animation idle

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -90,6 +92,7 @@
 	private int lastCombatCountdown = 0;
 	private Instant sixHourWarningTime;
 	private boolean ready;
+	private Instant lastInteractingFishing;
```
deathbeam: Maybe rename this to lastInteractingNonCombat to generalize this a bit for future use? Or even better, simply reuse lastInteracting and add like boolean inCombat to differentiate between out of combat and non-combat notification, as otherwise the logic is just duplicated

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -577,5 +608,15 @@ private void resetTimers()
 		{
 			lastInteract = null;
 		}
+
+		// Reset fishing idle timer
```
deathbeam: You could've avoided this if you just unified combat idle with your new functionality

Twiglet1022:
Thanks for the feedback. I agree I was unnecessarily duplicated some code so I've now unified the combat and fishing idle notifications as you suggested. How does this look now?

Issue #6449, opened 2018-11-13
Add a cannon tile marker to Myth's Guild


ieneoj:
I did some testing and I believe that the optimal location for placing a cannon for blue dragons in the Myth's guild is at [this spot](https://media.discordapp.net/attachments/396147184237805569/511730103080976385/unknown.png), and I believe a cannon tile marker should be placed there.

deathbeam:
Fixed in https://github.com/runelite/runelite/pull/6460

PR #6460, opened 2018-11-14
Adds a cannon tile marker to the blue dragon area in the Myth's Guild…

Luk4ro:
… (#6449)


Luk4ro:
Yeah, here you go: https://imgur.com/a/IHZn7GU

PR #6467, opened 2018-11-14
 NMZ Overload Timed Notifier

awbasham:
Added overload timed threshold for notifications in nightmare zone. Allows notifications before overload runs out. The timer is slightly redundant, as there is one already used for the overload UI timer, but it couldn't be accessed conveniently.


awbasham:
I think I screwed up trying to fork the project from an originally cloned version.. I only wanted to pull request commit 831a7de.

PR #6470, opened 2018-11-15
Make Enum dump output nicer

Joshua-F:
Adds `ScriptVarType` which includes a way to convert a `char` to a `ScriptVarType`. This enum contains (most) types used in OSRS.

### DIFF
--- a/cache/src/main/java/net/runelite/cache/util/ScriptVarType.java
+++ b/cache/src/main/java/net/runelite/cache/util/ScriptVarType.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2018, Joshua Filby <joshua@filby.me>
+ * All rights reserved.
+ *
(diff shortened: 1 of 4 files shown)


cache/src/main/java/net/runelite/cache/util/ScriptVarType.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public enum ScriptVarType
+{
+	INTEGER('i', "integer"),
+	BOOLEAN('l', "boolean"),
+	SEQ('A', "seq"),
+	COLOUR('C', "colour"),
+	COMPONENT('I', "component"),
```
Abextm: Perhaps comment this with RuneLite names

Abextm:
Merged in ec5737a1d92705ad82a7449962aa59900387a38c

PR #6472, opened 2018-11-15
Add sound when taking a screenshot

akarhi:
I have my notification set to off because I do not like the windows notification system. Sometimes when I take a screenshot I might forget that I already did. I just added a sound to the plugin the failing splashing sound. The config is to set off for default because the notification is set on by default the user can change whenever he/she feels like it. I know you can check your screenshot by pasting the link this change allows you to remember if you taken a screen shot or not.


deathbeam:
You can just enable send sound notifications and disable tray notifications in "RuneLite" settings, I dont see why this change is ever necessary

akarhi:
I tried testing it out. The sound doesn't seem to play on my PC maybe I turned off like everything that doesn't look like windows 7.

PR #6485, opened 2018-11-15
Null-check GL and window on plugin shutdown 

deathbeam:
If GL is not even created due to missing library error or unsupported
 version, this throws a lot of unnecessary exceptions to log and fails
 on shutDown resulting in possibly improperly feed objects and displays.


Adam-:
Does this still happen?

deathbeam:
Yes, it can still happen if the .getGl4 fails (what will happen if you have e.g Gl3 support only)

PR #6512, opened 2018-11-16
Config: support integer ranges 

raiyni:


jplsek:
I think the config manager should also return the appropriate values as well to avoid checks like [this ](https://github.com/runelite/runelite/blob/2cfdde3a076306e3fa23044e4371ff00880b54a7/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java#L534)in the plugins. In the current state of this PR, people making new plugins would assume that the config manager does the validation already, so they wouldn't put in validation code.

This is also in case people manually change the settings file. Or in the current live case, where people have their GPU draw distance greater than 90.

raiyni:
ConfigManager isn't aware of the ConfigItem when configurations are fetched. When the panel is opened again, it will be constrained.

runelite-client/src/main/java/net/runelite/client/config/ConfigItem.java
```diff
@@ -46,4 +46,8 @@
 	String warning() default "";
 
 	boolean secret() default false;
+
+	int maxInt() default Integer.MAX_VALUE;
```
deathbeam: I think you should instead use custom type as return value (e.g something like Range) instead of having maxInt and minInt available for all fields (e.g even for string fields and boolean fields where it makes no sense)

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -337,6 +339,33 @@ void openGroupConfigPanel(PluginListItem listItem, Config config, ConfigDescript
...
+				Range range = configManager.getConfiguration(cd.getGroup().value(), cid.getItem().keyName(), Range.class);
+
+				if (range == null)
+				{
+					try
+					{
+						range = (Range) ConfigInvocationHandler.callDefaultMethod(config, cid.getMethod(), null);
+					}
+					catch (Throwable throwable)
+					{
```
deathbeam: Maybe you should do `continue` here? As something must be broken when this happens

runelite-client/src/main/java/net/runelite/client/config/ConfigInvocationHandler.java
```diff
@@ -81,7 +81,13 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
 			
 			try
 			{
-				return ConfigManager.stringToObject(value, returnType);
+				Object returnValue = ConfigManager.stringToObject(value, returnType);
+				if (returnValue == null && method.isDefault())
```
Adam-: I don't think this can return null?

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -337,6 +339,35 @@ void openGroupConfigPanel(PluginListItem listItem, Config config, ConfigDescript
...
 
+			if (cid.getType() == Range.class)
+			{
+				Range range = configManager.getConfiguration(cd.getGroup().value(), cid.getItem().keyName(), Range.class);
+
+				if (range == null)
+				{
+					try
+					{
+						range = (Range) ConfigInvocationHandler.callDefaultMethod(config, cid.getMethod(), null);
```
Adam-: Why are you doing this at all?

runelite-client/src/main/java/net/runelite/client/config/ConfigInvocationHandler.java
```diff
@@ -32,7 +32,7 @@
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-class ConfigInvocationHandler implements InvocationHandler
+public class ConfigInvocationHandler implements InvocationHandler
```
Adam-: Why is this public?

15987632:
I haven't looked into how the config is rendered much so I don't know how easy this is to implement but, I think a better way to show this in the UI is a JSlider rather than a JSpinner

PR #6513, opened 2018-11-16
Add toggle for color banding

### LINKED ISSUE #6505: GPU plugin 'smooths out' gradiants
Before/after of a farming patch:
[Before](https://i.imgur.com/YixGssZ.png) / [After](https://i.imgur.com/GpT3vPn.png)

Before/after of random ground:
[Before](https://i.imgur.com/KVtbuqC.png) / [After](https://i.imgur.com/kfrkDZb.png)

Toocanzs:
Adds a toggle to the GPU plugin to reintroduce the color banding that exists in the CPU renderer. 

The GPU renderer makes some dark places look really smooth compared to the CPU renderer.

Banding off:
https://i.imgur.com/RGqwVwD.png

Banding on: 
https://i.imgur.com/mfdvokF.png

CPU renderer:
https://i.imgur.com/SQUbuiW.png

Closes #6505


Kamielvf:
I think this should be inverted, to have the CPU behaviour as result by default, and an option to smoothen them out.

Also closes #6505

Toocanzs:
Good point, I've inverted the toggle to instead be "Remove Color Banding"

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -119,7 +46,16 @@ void main()
 
   vec3 rgb = hslToRgb(hsl);
 
+  int h = (hsl >> 16) & 0xff;
```
Adam-: this is always 0

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/frag.glsl
```diff
@@ -27,16 +27,26 @@
...
+#include hslToRgb.glsl
+
 void main() {
   float n = fUv.x;
   float u = fUv.y;
   float v = fUv.z;
 
+  int hsl = (int(fHsl.x * 255.f) << 8) | int(fHsl.y * 255.f);
+
+  vec3 rgb = (hslToRgb(hsl) * smoothBanding) + (Color.rgb * (1f - smoothBanding));
```
devLotto: ``1f`` does not compile on AMD graphics cards. Use ``1.f`` or ``1.0f``.

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -121,5 +49,6 @@ void main()
 
   vPosition = vertex;
   vColor = vec4(rgb, 1.f - a);
+  vHsl = vec2((hsl >> 8)/255.f, (hsl & 0xff)/255.f);
```
Adam-: Does this work the same way if you just use an ivec2? or an int?

Toocanzs: You can't have ivec2 as a vertex attribute because then the attribute needs to be flat(the entire triangle shares one value). If it shares one value then there's the entire triangle is going to share the same rgb value too.

PR #6522, opened 2018-11-16
Add ability to remap hotkeys

pratik-p:
Allows user to remap F-keys(1-12) to a new Keybind.


raiyni:
you can use AHK with vastly less code and is better.

Ashe:
Saying 'Just use AHK' doesn't help linux users or people who use the same runelite account on multiple machines. This game really needs keybindings and people who don't have F keys / don't want to use them are incredibly at a disadvantage when it comes to bossing. I really hope people come back to this and reconsider as it saddens me that there isn't any other effort to change things.

PR #6534, opened 2018-11-17
Switch to core profile in GPU plugin

deathbeam:
This makes GPU plugin available to more people and workarounds bug in
JOGL on Linux where it is not detecting compatible profile on some
systems.


pacf531:
I can confirm that this in conjunction with the other pull request fixes Linux rendering for Intel integrated graphics. Performance is okay, not amazing. You can hit the 50 FPS frame limit at default view distance if you zoom in halfway between max zoom and the next level. Max zoom out with a birds eye view at default view distance and you can get into the teens for FPS. Probably still some optimizations that could be made but it's a good start.

whitehooder:
This is probably the way to go to support the most devices unless it breaks something that relies on removed deprecated features of earlier versions (such as quad rendering).

MP2E:
Tested this on my Linux distro with an nvidia GTX 660 Ti on binary driver 410.73, works perfectly as far as I can tell. Experienced no issues over a few hours of gameplay.

I'm thinking that this is the way to go unless other issues are found

PR #6545, opened 2018-11-17
Add puzzle solver for monkey madness 1 puzzle box

### LINKED ISSUE #2063: puzzlesolver: add support for the Monkey Madness puzzle box
Our current algorithm for solving puzzle boxes (IDA* with Manhattan distance) depends on the puzzle having been randomized only a small amount of times (<30). Clue puzzles always seem to have less than 30 steps.

The Monkey Madness puzzle box appears to be using around 65 steps, which leaves the algorithm to run for a solid 2 minutes or more before the solution is displayed, which is unacceptable.

Maybe a special case for the MM puzzle can be added, where it uses some other algorithm. One that doesn't give the optimal solution but still helps the user.

Also, the MM puzzle uses inventory ID 221 instead of 140. The item IDs in the container consist of even numbers, and can as such not be passed into the current ``convertToSolverFormat`` method. This differs from the clue puzzle container 140, which has no gaps.

steffenhauge:
Solves the monkey madness 1 puzzle box by first solving the top row, then the left most column and then using the IDA* algorithm to solve the remaining part of the puzzle.
This approach gets a solution to the puzzle in a few seconds, at the cost of some additional steps.

picture

Gif:
gif

Closes #2063


steffenhauge:
Closes #2063

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/monkeymadness/PuzzleStateMM.java
```diff
@@ -0,0 +1,855 @@
...
+		validRowNumbers.add(Arrays.asList(6, 7, 8, 9));
+		validRowNumbers.add(Arrays.asList(12, 13, 14));
+		validRowNumbers.add(Arrays.asList(18, 19));
+		validColumnNumbers.add(Arrays.asList(5, 10, 15, 20));
+		validColumnNumbers.add(Arrays.asList(11, 16, 21));
+		validColumnNumbers.add(Arrays.asList(17, 22));
+		validColumnNumbers.add(Arrays.asList(23));
+	}
+
+	PuzzleState solvePathfinder(PuzzleState preSolved, Pathfinder pathfinder)
```
devLotto: I don't think this is needed because it will always fail for MM puzzles?

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/monkeymadness/PuzzleStateMM.java
```diff
@@ -0,0 +1,855 @@
...
+
+		if (solver.hasFailed())
+		{
+			return null;
+		}
+
+		return solver.getStep(solver.getStepCount() - 1);
+	}
+
+	void solveRow(int row)
```
devLotto: Don't solve the puzzle in the puzzle state classes. Solving should be in the puzzle solver.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/monkeymadness/PuzzleStateMM.java
steffenhauge: The IDA* algo?

steffenhauge:
Rebased, removed overlayMM and stateMM.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/monkeymadness/PuzzleStateMM.java
steffenhauge: Removed unneeded valid lists.

devLotto: solvePathfinder

runelite-client/src/test/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverTest.java
```diff
@@ -85,6 +86,20 @@
 
 	private static final int[] FINISHED_STATE = new int[]{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, -1};
 
+	@Test
+	public void testSolverMM()
```
devLotto: Test with MM puzzle states. The START_STATES are clue puzzlebox states.

steffenhauge: Should i test with the translated states or the raw states?

devLotto: I'd say translated states

steffenhauge: Some test results:

4x4 then IDA*: https://pastebin.com/xXeuiJ8u
3x4 then IDA*: https://pastebin.com/4SSiU9qK

Doing 4x4 then IDA* takes 131 steps on average but can exceed the 5 second timer on the puzzlesolver.
Doing 3x4 then IDA* takes 166 steps on average but always solves in a few 100ms.

steffenhauge:
Some test results:

4x4 then IDA*: https://pastebin.com/xXeuiJ8u
3x4 then IDA*: https://pastebin.com/4SSiU9qK

Doing 4x4 then IDA* takes 131 steps on average but can exceed the 5 second timer on the puzzlesolver.
Doing 3x4 then IDA* takes 166 steps on average but always solves in a few 100ms.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,838 @@
...
+					//Left
+					swap(locBlank, locVal);
+				}
+			}
+		}
+	}
+
+	private void swap(Point p1, Point p2)
+	{
+		int[] pieces = currentState.getPuzzle().clone();
```
devLotto: Please make a method in PuzzleState that creates a copy of the state, does the swap and then returns the new PuzzleState. Then remove the getPuzzle method. This way we don't expose the puzzle array, as we want to keep it immutable.

runelite-client/src/test/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverTest.java
```diff
@@ -83,8 +84,72 @@
...
+	};
+
 	private static final int[] FINISHED_STATE = new int[]{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, -1};
 
+	@Test
+	public void testSolverMM()
+	{
+		for (PuzzleState state : START_STATES_MM)
+		{
+			PuzzleSolver solver = new PuzzleSolver(new IDAStarMM(new ManhattanDistance()), new PuzzleState(state.getPuzzle()));
```
devLotto: You should not have to copy the start state before passing it into the PuzzleSolver. It should not look like the solver edits the passed in state (I don't think your code does that either?).

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,838 @@
...
+
+			if (valCurrent != valTarget)
+			{
+				Point locTarget = new Point(column, i);
+				moveTowardsVal(valTarget, locTarget,  false);
+			}
+		}
+	}
+
+	private void validateInput(int val)
```
devLotto: This method is only ever used when ``val = 0``, it doesn't seem to be needed?

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,838 @@
...
+		{
+			return true;
+		}
+
+		return false;
+	}
+
+	private int getPiece(Point p)
+	{
+		return currentState.getPiece(p.getX(), p.getY());
```
devLotto: There's liberal usage of Point all over. I think you can just pass in x and y as ints instead of packing it into Point everywhere.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
steffenhauge: It was just there to ensure that when solveRow or solveColumn was parameterized, then number entered as parameter was valid.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/PuzzleState.java
```diff
@@ -179,4 +180,45 @@ public int getHeuristicValue(Heuristic heuristic)
 
 		return h;
 	}
+
+	public PuzzleState swap(Point p1, Point p2)
+	{
+		int[] pieces = this.pieces.clone();
```
devLotto: This cloning should be after the check. Also use this for cloning instead of cloning the array itself.
```java
PuzzleState newState = new PuzzleState(this);

newState.pieces[p1.getY() * DIMENSION + p1.getX()] = val2;
newState.pieces[p2.getY() * DIMENSION + p2.getX()] = val1;

return newState;
```

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -79,6 +80,7 @@
 	private PuzzleSolver solver;
 	private Future<?> solverFuture;
 	private int[] cachedItems;
+	private boolean useNormalSolver;
```
devLotto: This variable should be a local variable in render(..) since it's overwritten everytime that's run, and it's only in there it is set.
Pass the boolean into ``convertToSolverFormat`` and ``solve`` and you'll be okay.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -393,7 +400,7 @@ else if (yDelta > 0)
 		{
 			if (items[i] != BLANK_TILE_VALUE)
 			{
-				convertedItems[i] = items[i] - lowestId;
+				convertedItems[i] = (items[i] - lowestId) / ratio;
```
devLotto:
```java
int value = items[i] - lowestId;

// The MM puzzle has gaps
if (!useNormalSolver)
{
    value /= 2;
}

convertedItems[i] = value;
```

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,791 @@
...
+					swapLeftColumn(valTarget, locTarget);
+				}
+			}
+		}
+	}
+
+	private void alignTargetX(int valTarget, Point locTarget)
+	{
+		Point locVal = findPiece(valTarget);
+		Point locBlank;
```
devLotto: This isn't C, consider declaring ``locBlank`` the first time it's used.
That makes it easier to see in which scope the variable is used.
It's overwritten every iteration so in this case it shouldn't be a problem?

steffenhauge: Yeah, moved it

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/PuzzleState.java
steffenhauge: Had to set the empty piece also.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/PuzzleState.java
```diff
@@ -53,26 +54,32 @@ public PuzzleState(int[] pieces)
...
 
-	private PuzzleState(PuzzleState state)
+	public PuzzleState(PuzzleState state)
 	{
 		this.pieces = Arrays.copyOf(state.pieces, state.pieces.length);
-		this.emptyPiece = state.emptyPiece;
+		this.emptyPiece = state.findEmptyPiece();
+	}
+
+	private int findEmptyPiece()
```
devLotto: I would make this method not return anything and instead set the ``this.emptyPiece`` inside.
And then ``throw new IllegalStateException("Incorrect empty piece passed in!");`` after the loop.
Then remove the exception throwing for emptyPiece in the constructor.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/PuzzleState.java
```diff
@@ -179,4 +186,47 @@ public int getHeuristicValue(Heuristic heuristic)
...
+
+		newState.pieces[p1.getY() * DIMENSION + p1.getX()] = val2;
+		newState.pieces[p2.getY() * DIMENSION + p2.getX()] = val1;
+
+		newState.emptyPiece = newState.findEmptyPiece();
+
+		return newState;
+	}
+
+	private boolean checkSwap(Point p1, Point p2)
```
devLotto: isValidSwap

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,789 @@
...
+		for (int i = row; i < DIMENSION; i++)
+		{
+			int valTarget = row * DIMENSION + i;
+
+			int valCurrent = currentState.getPiece(i, row);
+
+			if (valCurrent != valTarget)
+			{
+				Point locTarget = new Point(i, row);
+				moveTowardsVal(valTarget, locTarget,  true);
```
devLotto: I don't think there is any reason to use Point all over.

Consider just passing in ``valTarget``, ``i`` and ``row`` and remove the Point usage.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,789 @@
...
+				else if (diff == -1)
+				{
+					//Left
+					swap(locBlank, locVal);
+				}
+			}
+		}
+	}
+
+	private void swap(Point p1, Point p2)
```
devLotto: No reason to pass in Points, consider passing in ``int x1, int y1, int x2, int y2`` instead.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,789 @@
...
+	{
+		PuzzleState newState = currentState.swap(p1, p2);
+
+		currentState = newState;
+		stateList.add(newState);
+	}
+
+	private Point findPiece(int val)
```
devLotto: This usage of Point is probably okay, so keep that.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
steffenhauge: Should i change the usage of points in PuzzleState aswell?

steffenhauge: For swap and isValidSwap

devLotto: Yeah

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,787 @@
...
+		}
+	}
+
+	private void moveTowardsVal(int valTarget, int x, int y, boolean rowMode)
+	{
+		//Not in place
+		boolean reached = false;
+
+		int row = y;
+		int column = x;
```
devLotto: I think these are forever the same as ``x`` and ``y``? Consider using x and y where you use row and column.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,787 @@
...
+
+		while (currentState.getPiece(x, y) != valTarget)
+		{
+			//Find piece location
+			Point locVal = findPiece(valTarget);
+			Point locBlank = findPiece(BLANK_TILE_VALUE);
+
+			if (!reached)
```
devLotto: Consider flipping this condition so the small code block comes first, then ``else``, and then the large block.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,787 @@
...
+						{
+							Point locSwap = new Point(locBlank.getX(), locBlank.getY() - 1);
+							swap(locBlank.getX(), locBlank.getY(), locSwap.getX(), locSwap.getY());
+						}
+					}
+					else
+					{
+						//Check if already correct to the left
+						if (locBlank.getX() - 1 == column &&
+							validColumnNumbers.get(column).contains(currentState.getPiece(locBlank.getX() - 1, locBlank.getY())) &&
```
devLotto: The ``&&`` should come before the condition. Look at for example https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksPlugin.java#L113.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,787 @@
...
+						else
+						{
+							//Space below, use lower rotate
+							Point loc1 = new Point(locVal.getX() + 1, locVal.getY() + 1);
+							Point loc2 = new Point(locVal.getX(), locVal.getY() + 1);
+							Point loc3 = new Point(locVal.getX() - 1, locVal.getY() + 1);
+							Point loc4 = new Point(locVal.getX() - 1, locVal.getY());
+
+							swap(locBlank.getX(), locBlank.getY(), loc1.getX(), loc1.getY());
+							swap(loc1.getX(), loc1.getY(), loc2.getX(), loc2.getY());
```
devLotto: Oh, I see now. The points are used in more than one swap.. Didn't see that, sorry.
I think the best thing in that case would be to have the swap in this class take in two Points like you had before, and have the PuzzleState swap take in integers like it is now.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,787 @@
...
+			for (int y = 0; y < DIMENSION; y++)
+			{
+				if (currentState.getPiece(x, y) == val)
+				{
+					return new Point(x, y);
+				}
+			}
+		}
+
+		return loc;
```
devLotto: This return is never reached, because if findPiece returns null your code crashes with an NPE.
It should never reach it, so maybe an exception should be thrown if the code for whatever reason gets there.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
steffenhauge: I think i spaced it one tab to much?

devLotto: Yeah it should only be 1 tab.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,781 @@
...
+		{
+			for (int y = 0; y < DIMENSION; y++)
+			{
+				if (currentState.getPiece(x, y) == val)
+				{
+					return new Point(x, y);
+				}
+			}
+		}
+		throw new IllegalStateException("Empty piece wasn't found!");
```
devLotto: ``val`` isnt always the empty piece? Also probably add a comment "This should never happen"

steffenhauge: Derp, yeah ofc.

steffenhauge:
Added gif to first post

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/solver/pathfinding/IDAStarMM.java
```diff
@@ -0,0 +1,722 @@
...
+				offsets = ROTATE_UP_LEFT.getPoints();
+				break;
+			case LAST_PIECE_ROW:
+				offsets = pattern.getPoints();
+				break;
+			case LAST_PIECE_COLUMN:
+				offsets = pattern.getPoints();
+				break;
+			default:
+				offsets = pattern.getPoints();
```
devLotto: These last 3 cases do the same thing?

steffenhauge: Actually the two above can be merged into default aswell.

PR #6547, opened 2018-11-17
Move trig calculations to CPU

Toocanzs:
All but two trig functions calls in the shaders can be done on the CPU instead


Adam-:
This is superseded by #6561

PR #6553, opened 2018-11-18
add change tab name feature

svermeulen1:
Adds a rename tab option when you right click a tag tab. It also changes the tag on every object in the tab.


raiyni:
You can just export the tab, paste in notepad, change the name, and import it.

Nightfirecat:
This was added in #7652.

PR #6564, opened 2018-11-18
Inventory idle notification

Twiglet1022:
The idle notifier now tracks changes to your inventory to detect when you are performing an action that automatically consumes items in your inventory. When you run out of enough items for another action it sends the following notification.

image

Benefits of this approach over checking animations:
- Triggers more quickly than waiting for the animation to end
- Works even if there is no animation e.g. cutting fish from Lake Molch
- Is generic so doesn't need a new cases added for new or obscure activities
- Some animations like stringing longbows go idle for a game tick or two between actions and so trigger animation idle notification if the delay is set too low. This approach doesn't have that issue.

Here's a demonstration of using it at Lake Molch while cutting fish, which does not have an animation:

OoO Aerial Fishing

Some things that I had to consider when making this:
- The activities that this applies to overlap with those that animation idle checks for. So I prevent animation idle notifications being sent after an out of items notification.
- Some automated actions use multiple different items so I have included support for creating item groups for use with this check. Currently this only applies to Aerial Fishing at Lake Molch. I didn't add the case for baking multiple pies with lunar magic because it's an unreasonable edge case, but this could cope with that if added.
- The notification does not trigger if doing things like banking items or dropping them because the variables that track inventory changes are reset if the user clicked recently.

<Details><summary>Expand for more GIF examples</summary>
The simplest possible case, using one item per action:

OoO Magic Long Cutting

Using multiple different items in the same tick:

OoO Magic Long Stringing

Using multiple of the same item in one tick:

OoO Green Dragonhide

</Details>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -43,10 +43,21 @@ public interface IdleNotifierConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "outOfItemsIdle",
+		name = "Out of Items Idle Notifications",
+		position = 2,
+		description = "Configures if notifications for running out of items for another action are enabled."
+	)
+	default boolean outOfItemsIdle()
+	{
+		return true;
+	}
+
+	@ConfigItem(
 		keyName = "interactionidle",
 		name = "Idle Interaction Notifications",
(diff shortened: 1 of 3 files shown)


Twiglet1022:
Updated the approach slightly per initial feedback from discord. 

Rather than replacing any animation idle notifications this is now just an enhancement allowing many notifications to trigger sooner than they would if you had to wait for the animation to end. The animation idle notification are still there as a backup for edge cases where there is another interruption besides running out of items like levelling up.

Twiglet1022:
Made some tweaks and added support for item grouping to account for things like Lake Molch fish cutting which consumes multiple different items.

PR #6569, opened 2018-11-18
Add compost bins to time tracking

### LINKED ISSUE #2960: Add rotting compost in farming tab
Feature request.
In the farming tab, please add a tab that keeps track of where compost is rotting.

steffenhauge:
Closes #2960.

Adds compost bin tracking:
picture

Additional information for varbits: https://pastebin.com/fZ1DvpxB

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -375,6 +375,7 @@ public enum Varbits
 	FARMING_4772(4772),
 	FARMING_4773(4773),
 	FARMING_4774(4774),
+	FARMING_4775(4775), //Compost bin
 
 	/**
 	 * Transmog controllers for grapes
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/Tab.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/Tab.java
@@ -43,9 +43,10 @@ public enum Tab
 	HOPS("Hops Patches", ItemID.BARLEY),
 	BUSH("Bush Patches", ItemID.POISON_IVY_BERRIES),
 	GRAPE("Grape Patches", ItemID.GRAPES),
-	SPECIAL("Special Patches", ItemID.MUSHROOM);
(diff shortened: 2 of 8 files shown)


steffenhauge:
Fixed, works now.

steffenhauge:
Remove compost bins from testing since compost and super compost timers run on a 40 minute tick but the ticks happens roughly 10 to 19 minutes after a 40 minute tick. The code in PatchImplementation accounts for this behaviour.

steffenhauge:
Added the last seen state for compost bins to the config key (only if varbit == 4775) and made stage for compost bins use the cached value, such that it doesn't reset, everytime you visit a compost bin.

lavarekira:
When can this PR get merged?

raftor:
> 
> 
> When can this PR get merged?

Sorry if it's bad form to comment on these, but I would like to second this question. If all checks have passed and the commits have been made, when can the branch get merged?

rejectedpromise:
Last night I started looking into writing this myself.
Looks like it has gone stale, so I will try to pick up from here, resolve conflicts, test, and get the PR ready again for merge. Please post an update if you begin to work on again!

steffenhauge:
Hi rejectedpromise. I kinda stopped working on this once i realised that farming timers are messy and the farming "ticks" are different from each user. I found that the compost bins takes two 40 minute ticks to complete but the farming "tick" happens 10-19 minutes into a 40 minute tick (tested on 3-4 different accounts) and therefore it is often off by 20 minutes. So this current solution would be infeasible, however i could change it back to just use a two 40 minute tick period. Also it seems that PatchImplementation is generated from a script by @Abextm so maybe the best way to add compost bin timers, is to add them in his script?

rejectedpromise:
@steffenhauge thank you very much for replying and giving so much useful information!
I think I am going to still take a look into this, as I think it will be a good opportunity to become more familiar with ticks and varbits, even if I am unable to make any progress.
I will also take a look into the PathImplementation and scripts to see if I can make sense of it!

steffenhauge:
This is the script im talking about https://github.com/Abextm/nipple/blob/master/scripts/farming.js you might have to contact Abex for help :)

PR #6578, opened 2018-11-19
Try to enable compute shaders on GL 4.2

deathbeam:
Compute shaders should be supported on GL4.2 with ARB_compute_shader
extension so try to load the GLSL version with 420 support.

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/comp.glsl
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/comp.glsl
@@ -22,7 +22,10 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-#version 430 core
+
+#version 420
+#extension GL_ARB_compute_shader : require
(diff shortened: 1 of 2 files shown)


deathbeam:
For context: https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_compute_shader.txt

deathbeam:
And for reference for mesa support for GPUs with 4.2 and required extensions <https://people.freedesktop.org/~imirkin/glxinfo/#diff&v=Mesa%2017.2.0>

cernodile:
Confirmed working on Arch Linux on a Ivy Bridge CPU using Mesa 19.0-devel, core profile 4.2, GLSL 4.20

deathbeam:
Merged this because:
1. I want to get this in very early after release so if any problems occur, we can always revert
2. I tested this on 3 different PCs, + some other people tested this for me. Tested scenarios:
  1. Windows PC with intel iGPU
  2. Windows PC with Nvidia GPU
  2. Windows PC with AMD GPU
  4. Linux PC with Mesa that supports GLSL 4.20 with all required extensions (without this patch GPU plugin was not working)
  5. Linux PC with Mesa that supports GL 4.5 and newest GLSL (without this patch, the GPU plugin froze the screen, probably because of buggy drivers)

Both extensions used in this patch are core extensions (so 1 step before actual addition to next GL version), so all vendors that ship GL 4.3 should ship also 4.2 with these extensions

PR #6580, opened 2018-11-19
Unlock surface only with X11JAWTWindow

deathbeam:
Unlocking it on Windows with some iGPUs apparently causes problems and
this is relevant only for X11, so unlock only there.


Adam-:
@MESLewis @whitehooder can you both confirm this fixes both of your systems?

PR #6585, opened 2018-11-19
Farming profit plugin

### LINKED ISSUE #5088: Farming runs profit
When doing some of my farming runs, especially herb runs I would like to know how much I have profited after a completed run. However having to price check the herbs every time I complete a run can be quite annoying.

So what I would really like to see is a plugin that will track the herbs that you have harvested during the current run. This would then be hooked up with the GE prices of the herbs and seeds to calculate the profit made during the farm run.
The runs could be separated using some kind of time interval to check whether the current harvest is a new run or should contribute to the current run. This way profit per run can be tracked.

I think the best option to visualize this would be with a plugin panel just like the loot tracker which can keep track of the herb profit during the whole day. In addition to this an overlay could be drawn of the profit of the current farm run.

_I started working on this on my own but haven't had much success yet so I would be happy if someone else picked this up as well ;)_

MKuijpers:
## Description
This PR adds a farming profit plugin, this is used to track your harvests during your farm runs. After a harvest the corresponding seed is retrieved and the profit is calculated using the seed and the amount harvest.

### Supported patch types:
* Herbs
* Allotments
* Hops
* Bushes
* Special (cactus + seaweed)

### Herb run example

Fixes #5088


Nightfirecat:
We're not likely to add a new panel for this plugin, but it's clear this feature is of interest to the community. I can see a couple options for having this feature added to the client:

1. Submit this as an external plugin [to the plugin hub](/runelite/plugin-hub) where users can install it
2. Move this code into the loot tracker so farming "loot" is tracked there. I can't say for sure how the team would feel about this approach, so I would recommend discussing that approach thoroughly with maintainers before going ahead with that change.

MKuijpers:
I am currently not in the position to start working on it again but I agree with you. The feature is not large enough for a separate panel and that it is better to somehow incorporate it in another panel or find a different solution.

I was not aware that the plugin hub existed, this might be a good alternative to just get it out there if people really want it. Finding a better solution would be better in my opinion though.

I think the most elegant solution would be to incorporate it into some other panel, with the loot tracker as the best candidate. However, I think it will lead to some problems, mostly UI wise. Just to note three problems:
1. It can fill up the loot tracker very quickly. The harvests can be grouped but there are a lot of options here. Would the harvests be grouped by run, by day, by crop type? I'm not sure what would be the best option and I can imagine some people will want a different group method anyway, no matter which method is chosen. I'm leaning towards grouping by crop type though.
2. Somehow the cost of seeds should be incorporated and shown intuitively. This is not done anywhere in the loot tracker yet, as far as I know. It is both a UI problem, showing that seeds decrease value, but storing the loot will also be challenging I think. I have not looked much at the loot tracker code but I assume items IDs are used to store the loot. Making seeds decrease the value would need some elegant solution that can also be synced with the website. (Maybe quantity set to -1? But would not be too elegant)

I think I will just keep it on hold for now until some other, more elegant, options arise. I am happy to pick it up again, but for now, I don't see a clearly best solution. I'm open to other opinions or ideas!

russel204:
Bump this. It would be nice if it could track profits but also loot from each herb similar to how the loot tracker can group or separate individual monsters. I hope someone can pick this up possibly.

Nightfirecat:
As mentioned earlier, this would be a good candidate for a hub plugin. Should you (or someone else) feel they want to submit it there, we can review it for submission as a hub plugin instead.

PR #6589, opened 2018-11-19
Statusbars : fixing number format exceptions for food with random values

Malfuryent:
This is done by ignoring strings that are longer than 3 because it cant parse strings with 2 numbers and a ~ in between and these are the only ones that have longer strings.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
@@ -190,6 +190,11 @@ class StatusBarsOverlay extends Overlay
 
 				for (final StatChange c : statsChanges.getStatChanges())
 				{
+					if (c.getTheoretical().length() > 3)
+					{
+						break;
+					}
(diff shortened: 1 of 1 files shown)


deathbeam:
I would rather use Integer.tryParse as raiyni suggested

Malfuryent:
implemented the method @Nightfirecat suggested so it still previews the highest possible heal from those random items.

runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
```diff
@@ -190,6 +191,13 @@ else if (poisonState >= 1000000)
 
 				for (final StatChange c : statsChanges.getStatChanges())
 				{
+					Integer string = Ints.tryParse(c.getTheoretical().substring(1));
+					if (string == null)
```
deathbeam: this is wrong, because if the number starts with - or nothing it just deletes that char. Also, in case this matches you do nothing with `string` var

deathbeam:
Any update on this?

PR #6591, opened 2018-11-19
 camera plugin: add expand outer zoom limit option

MagicfTail:


runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomConfig.java
```diff
@@ -42,11 +42,22 @@ default boolean innerLimit()
 		return false;
 	}
 
+	@ConfigItem(
+		keyName = "outerLimit",
+		name = "Expand outer zoom limit",
+		description = "Configures whether or not the outer zoom limit is increased",
+		position = 2
+	)
+	default int outerLimit()
```
Nightfirecat: I'm assuming this is not meant to be an int? Either that, or the description is woefully unhelpful.

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -81,6 +83,7 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
 			return;
 		}
 
+		int outerZoomLimit = Math.min(zoomConfig.outerLimit(), OUTER_ZOOM_LIMIT);
```
Adam-: This would probably break horribly if too negative.

MagicfTail: The zoom resets at around -2500, so it never becomes a problem for me at least. Even then it's limited at -1k

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -81,6 +83,7 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
 			return;
 		}
 
+		int outerZoomLimit = Math.max(-zoomConfig.outerLimit(), OUTER_ZOOM_LIMIT);
```
Abextm: Maybe this should be `128 - zoomConfig.outerLimit()` so setting the config to 1 doesn't change the limit massively.

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -95,6 +98,12 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
 			return;
 		}
 
+		if ("outerZoomLimit".equals(event.getEventName()) && outerZoomLimit != 128)
```
Nightfirecat: Given that setting intStack values is cheap, is it important for us to have that `outerZoomLimit != 128` check? I'd like to see that check removed, or have `128` defined as a constant somewhere and not just be a magic number.

MagicfTail: Removed the check

Issue #6567, opened 2018-11-18
Show combat level in player indicators overlay


cannon:
**Describe the solution you'd like**
The player indicator (floating names) could easily be extended to show the combat level. This would be very helpful in wilderness for determining who to attack. Additionally, it should be able to be color coded whether or not you are able to attack the player in the current wilderness level.

pwatts6060:
Is all of these features allowed? If so I can implement it, someone asked me today. 
1. Is combat level next to name allowed?
2. Can it only be visible when you're in the wilderness/pvp areas
3. Can it be color coded green/red for whether you can attack them or not

RobynTheDevil:
If approval is denied, could the info be put next to mouse cursor during hover?

PR #6593, opened 2018-11-19
add toggle for drawing player combat levels in player indicators plugin

PinedaJ:
Add the option to either draw a player's combat level next to their name or draw the player's combat only.

Addresses issue #6567, without the ability to see which players you can attack in the wilderness since I believe that would be rejected.


Nightfirecat:
It would be a bit strange to draw the combat level as `(123)` instead of `123` if the player name drawing was disabled. Could that be handled in this PR?

PinedaJ:
Resolved suggested changes.

deathbeam:
I dont think so @Adam-

PR #6595, opened 2018-11-19
Untradable item price support for ItemManager

### LINKED ISSUE #6418: Mark of grace examine giving the price of 10 amylase crystals
I would like to see a feature added that when you examine a stack of marks of grace it will show you the price of 10 amylase crystals instead of showing just the alch price of a mark.

https://i.gyazo.com/38d9db78ed81df48df15b4bc04e596fe.png

MKuijpers:
Add support to set prices of untradable items. The price of the untradable item is an amount of a tradable item (for example a mark of grace is worth 10 amylase crystals). These prices will be used in the getItemPrice method of the ItemManager. 
As a result, these prices will be displayed as GE prices in examine messages and the bank value. 

The following untradable items are already added:
* Marks of grace + Full graceful set
* Golden nuggets + Full prospector outfit

Fixes #6418

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
@@ -26,9 +26,11 @@ package net.runelite.client.game;
 
 import com.google.common.collect.ImmutableMap;
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import net.runelite.api.ItemID;
 
 @Getter
+@RequiredArgsConstructor
 public enum UntradeableItemMapping
 {
 	MARK_OF_GRACE(ItemID.MARK_OF_GRACE, 10, ItemID.AMYLASE_CRYSTAL),
@@ -46,11 +48,12 @@ public enum UntradeableItemMapping
 	PROSPECTOR_LEGS(ItemID.PROSPECTOR_LEGS, 40, ItemID.GOLDEN_NUGGET),
 	PROSPECTOR_BOOTS(ItemID.PROSPECTOR_BOOTS, 24, ItemID.GOLDEN_NUGGET);
 
+	private static final ImmutableMap<Integer, UntradeableItemMapping> UNTRADEABLE_RECLAIM_MAP;
+
 	private final int itemID;
 	private final int quantity;
 	private final int priceID;
 
-	private static final ImmutableMap<Integer, UntradeableItemMapping> UNTRADEABLE_RECLAIM_MAP;
 	static
 	{
 		ImmutableMap.Builder<Integer, UntradeableItemMapping> map = ImmutableMap.builder();
@@ -59,13 +62,6 @@ public enum UntradeableItemMapping
 			map.put(p.getItemID(), p);
 		}
 		UNTRADEABLE_RECLAIM_MAP = map.build();
-	}
-
-	UntradeableItemMapping(int itemID, int quantity, int priceID)
-	{
-		this.itemID = itemID;
-		this.quantity = quantity;
-		this.priceID = priceID;
 	}
 
 	public static UntradeableItemMapping map(int itemId)
--- a/runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
@@ -29,7 +29,7 @@ import lombok.Getter;
 import net.runelite.api.ItemID;
 
 @Getter
-public enum UntradeableItemSellPrice
+public enum UntradeableItemMapping
 {
 	MARK_OF_GRACE(ItemID.MARK_OF_GRACE, 10, ItemID.AMYLASE_CRYSTAL),
 	GRACEFUL_HOOD(ItemID.GRACEFUL_HOOD, 28, ItemID.MARK_OF_GRACE),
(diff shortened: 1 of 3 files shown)


forsco:
I personally don't like this. If you want that value, trade them for amylase crystals.

MagicfTail:
You can't sell graceful, why does it have a value?

MKuijpers:
> You can't sell graceful, why does it have a value?

You can convert it back to marks of grace for a reduced amount of marks of grace. This amount of marks of grace are used as the value of the graceful pieces.

MagicfTail:
The wiki says otherwise 🤷‍♀️

MKuijpers:
The amount of marks of grace in the column 'Sells for' is used:

MagicfTail:
https://i.imgur.com/2RPy5Wn.png

forsco:
> The wiki says otherwise 🤷‍♀️

Selling the whole set back to Grace will yield 208 Marks

Taken from: https://oldschool.runescape.wiki/w/Graceful_outfit

However I still feel that these items shouldn't be given a value.

MKuijpers:
We decided to give these items a value for the same reasoning that for example a slayer helm is given a value. The value of the item is the possible amount of GP you can get back from this item. And since you can turn the graceful pieces into marks of grace, into amylase crystals and then into GP we thought that this would be the best option.

forsco:
I still think it's wrong. Im not going to go converting my only graceful set on an ironman into gp am i? Nor do I think any main account would either.

forsco:
I wouldn't be so opposed to this if it could be toggled.

deathbeam:
Well what we agreed on for bank value pricing is that the prices should be liquify value, and this fits into that category. As I said in Discord too, I will also not go and convert my slayer helm to black mask and then unimbue it just to sell it, but it still has (and should have) value.

TheStonedTurtle:
@Adam- The JSON loaded via `ItemVariation` seems to already include the worn item IDs for graceful.

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemSellPrice.java
```diff
@@ -0,0 +1,76 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.game;
+
+import com.google.common.collect.ImmutableMap;
+import lombok.Getter;
+import net.runelite.api.ItemID;
+
+@Getter
+public enum UntradeableItemSellPrice
```
deathbeam: Maybe rename this to UntradeableItemMapping?

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
```diff
@@ -0,0 +1,76 @@
...
+	PROSPECTOR_HELMET(ItemID.PROSPECTOR_HELMET, 32, ItemID.GOLDEN_NUGGET),
+	PROSPECTOR_JACKET(ItemID.PROSPECTOR_JACKET, 48, ItemID.GOLDEN_NUGGET),
+	PROSPECTOR_LEGS(ItemID.PROSPECTOR_LEGS, 40, ItemID.GOLDEN_NUGGET),
+	PROSPECTOR_BOOTS(ItemID.PROSPECTOR_BOOTS, 24, ItemID.GOLDEN_NUGGET);
+
+	private final int itemID;
+	private final int quantity;
+	private final int priceID;
+
+	private static final ImmutableMap<Integer, UntradeableItemMapping> UNTRADEABLE_RECLAIM_MAP;
```
deathbeam: Move static fields above instance fields

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
```diff
@@ -0,0 +1,76 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.game;
+
+import com.google.common.collect.ImmutableMap;
+import lombok.Getter;
+import net.runelite.api.ItemID;
+
+@Getter
+public enum UntradeableItemMapping
```
deathbeam: Use `@RequiredArgsConstructor` and remove the one you created

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
MKuijpers: What about the static code block? Should it stay behind the static field or should it stay after the instance fields?

deathbeam: static field block is probably fine where it is

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
MKuijpers: Did not know about that annotation, very nice

runelite-client/src/main/java/net/runelite/client/game/UntradeableItemMapping.java
MKuijpers: Alright, I'll just leave it then

traddcodes:
Is it not possible to add this toggle? I understand that Graceful can be sold back for MOG, and used to purchase Amylase crystals, but I don't think anyone would typically do that thus contributing to an inaccurate "useable" Bank Value calculation.

PR #6550, opened 2018-11-17
gpu: add scene anti-aliasing

devLotto:
**Off:**

no_aa

**MSAA x8:**

msaa_x8

PR #6597, opened 2018-11-20
 gpu: add stretched mode support 

### LINKED ISSUE #6487: Stretched mode plugin does not work with gpu plugin enabled
**Describe the bug**
When having both the gpu plugin and stretched mode plugin enabled in either fixed or stretched mode it will glitch out and malform your game window.

**To Reproduce**
1. Enable both the gpu plugin and stretched mode plugin
2. See what happens

**Expected behavior**
I would expect them to work at the same time

**Screenshots**
https://gyazo.com/b6f309eaafda7a7bb06eee7ca8ee6971

 - OS: [Windows]

devLotto:
Draws the scene + ui on a frame buffer that is then stretched when it's drawn on the default framebuffer.

Maybe there is a more clever way but this seems like it works, on my system anyway.

Fixes #6487

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -39,6 +39,7 @@ import com.jogamp.opengl.GLDrawableFactory;
 import com.jogamp.opengl.GLException;
 import com.jogamp.opengl.GLProfile;
 import java.awt.Canvas;
+import java.awt.Dimension;
 import java.awt.Image;
 import java.awt.image.BufferedImage;
 import java.awt.image.DataBufferInt;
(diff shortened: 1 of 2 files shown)


formatme:
https://i.imgur.com/J4ktrBL.png
everything stops rendering on screen but works

devLotto:
@formatme does #6550 work for you?

MP2E:
This works well on my system. The UI did disappear for me when I tried to merge both the anti aliasing PR and this one together, but that was fixed by simply disabling antialiasing if stretched mode is turned on by changing

```
if (aaEnabled)
```

to

```
if(aaEnabled && !client.isStretchedEnabled())
```

in both locations where aaEnabled is checked. Not sure if that's the best solution in general

devLotto:
Yeah you can't use both at the same time without changing some code. Were you doing that? @formatme

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -679,6 +736,29 @@ public void draw()
...
+			if (lastStretchedDimensions == null || !lastStretchedDimensions.equals(stretchedDimensions))
+			{
+				shutdownStretchedFbo();
+				initStretchedFbo(stretchedDimensions.width, stretchedDimensions.height);
+				lastStretchedDimensions = stretchedDimensions;
+			}
+
+			gl.glBindFramebuffer(gl.GL_DRAW_FRAMEBUFFER, fboStretchedHandle);
+		}
+		else
```
Adam-: `else if (lastStretchedDimensions != null)`

formatme:
ya i was trying to use both.

formatme:
thanks MP2E

Issue #6590, opened 2018-11-19
GPU plugin throwing warning every frame


MagicfTail:
```GLDebugMessageHandler: GLDebugEvent[ id 0x20071
	type Warning: generic
	severity Unknown (0x826b)
	source GL API
	msg Buffer detailed info: Buffer object 7 (bound to GL_SHADER_STORAGE_BUFFER (0), usage hint is GL_STREAM_DRAW) will use VIDEO memory as the source for buffer object operations.
	when 1542660293797
	source 4.5 (Core profile, arb, debug, compat[ES2, ES3, ES31, ES32], FBO, hardware) - 4.5.0 NVIDIA 416.94 - hash 0x14641a6f]```

Specs:
Win 10
I7 4770k
GeForce 780 Ti

deathbeam:
This do not happens on every GPU, e.g I am not seeing this on intel igpu

PR #6599, opened 2018-11-20
Suppress debug message for generic nvidia warning with severity Unknown.

l2-:
should fix https://github.com/runelite/runelite/issues/6590


MagicfTail:
Confirmed fixes 6590 for me

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -247,6 +249,10 @@ protected void startUp()
 				{
 					// Debug config on context needs to be set before .makeCurrent call
 					glContext.enableGLDebugMessage(true);
+
+					// 0x826b = severity Unknown
+					glContext.glDebugMessageControl(GL_DEBUG_SOURCE_API, GL_DEBUG_TYPE_OTHER,
```
deathbeam: Isnt there constant for severity Unknown in the GL api somewhere? Also, you can access the constants from gl. object directly, no need to import them

deathbeam: https://jogamp.org/wiki/index.php/GL_Debug_Output based on this there should be constants for severity

l2-: No. https://i.imgur.com/WFoGVr7.png

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -267,6 +270,10 @@ protected void startUp()
 				if (log.isDebugEnabled())
 				{
 					gl.glEnable(GL_DEBUG_OUTPUT);
+
+					// Need to control the debug messages here too.
+					gl.getContext().glDebugMessageControl(gl.GL_DEBUG_SOURCE_API, gl.GL_DEBUG_TYPE_OTHER,
```
deathbeam: Are you sure it is necessary on both places? Because gl.getContext() == glContext, and so it should have the setting already set

l2-: From testing it, it seems gl.getContext().glDebugMessageControl works but glContext.glDebugMessageControl doesn't so i removed it as it is unnecessary.

PR #6602, opened 2018-11-20
World Map Plugin: optimize clipping for world map points

trimbe:
Changes the world map overlay to only clip the graphics if it's necessary. `setClip` is fairly expensive and re-setting the clip for every map point could almost halve FPS when the world map was open.


oabdellatif:
Can confirm this fixes the FPS issue with the world map for me. The drop is even more noticeable with the introduction of the GPU plugin and the higher average frame rate.

PR #6603, opened 2018-11-20
Fix glsl errors

deathbeam:
- Rename common.glsl to comp_common.glsl -Otherwise IntelliJ complains about reserved word
- Add default return statement to count_prio_offset - Otherwise intellij complains
- Move return statement to end of priority_map -IntelliJ complains about no return statement when it is in switch { default } case.


Adam-:
I don't think any of these fixes are actual problems.

deathbeam:
That is why I labeled them as enhancement and not bug, its just so intellij stops complaining. Even though, honestly, just looking at `count_prio_offset` it makes no sense, unless GLSL has implicit returns.

Adam-:
Priority must be 0-17

deathbeam:
Yeah, but the function returns nothing unless priority is 0

Adam-:
As far as I can tell, none of these are actual errors.

deathbeam:
As I stated before, IntelliJ is complaining about it

PR #6604, opened 2018-11-20
WIP - Add option to use native window with JOGL

deathbeam:

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuNativeWindow.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuNativeWindow.java
@@ -195,7 +195,7 @@ public class GpuNativeWindow
 				final long now = System.currentTimeMillis();
 				final java.awt.event.KeyEvent event = new java.awt.event.KeyEvent(
 					canvas, java.awt.event.KeyEvent.KEY_RELEASED,
-					now, 0, e.getKeyCode(), e.getKeyChar());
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuNativeWindow.java
```diff
@@ -0,0 +1,216 @@
...
+				}
+			}
+
+			@Override
+			public void keyReleased(KeyEvent e)
+			{
+				final long now = System.currentTimeMillis();
+				final java.awt.event.KeyEvent event = new java.awt.event.KeyEvent(
+					canvas, java.awt.event.KeyEvent.KEY_RELEASED,
+					now, 0, e.getKeyCode(), e.getKeyChar());
```
pacf531:
```suggestion
					now, 0, AWTNewtEventFactory.newtKeyCode2AWTKeyCode(e.getKeyCode()), e.getKeyChar());
```

This fixes an issue where keyRelease events never worked and testing this, I would get bad things happening like the camera forever rotating and not recognizing I only pressed the arrow key once and released already.

pacf531:
I did some testing and it does function on a basic level. I commented on a bug I found but besides that, there's one big downside to this when I tested this, but it seems like this in its current state breaks the Ctrl-V pasting LoginScreenPlugin most noticeably for me and would also probably break any plugins that would implement their own KeyListener and MouseListener, which would probably be quite a few plugins.

deathbeam:
I think that is simply because I am not propagating modifiers @pacf531. But thank you very much for testing it, I plan to look at this a bit more in future, I just got distracted a bit

PR #6606, opened 2018-11-20
runelite-client: Make Skotizo widget layoutable

Abextm:
I haven't tested this


deathbeam:
I will test it in few hours when I will be home

PR #6608, opened 2018-11-20
changes prayer to light up with any prayer activated instead of quick-prayers

### LINKED ISSUE #6605: Status bars light up when any prayer is on
The prayer side of the status bars currently lights up when you turn your quick prayers on. Would help if it lit up if you turned prayers on manually as well in case you forgot one on.

Side note: potentially adding this to the prayer orb as well?

Malfuryent:
Changes prayer bar to light up when any prayer is activated instead of only quick-prayer activation.

This makes more sense and will alert people that they have prayers activated.
people can still check if quick-prayers are active on the orb if they want to know that.

closes #6605


Malfuryent:
This should have the requested behaviour now, im just not sure if my execution of the requested change is correct.

PR #6615, opened 2018-11-21
Fix incorrect slayer kill remaining count in xp tracker

### LINKED ISSUE #6614: XP Tracker Slayer Kill Tracker is Off for Normal Tasks
**Describe the bug**
When doing slayer tasks the XP Tracker plugin gets the kills remaining for level up wrong for slayer xp.

**To Reproduce**
1. Go to a slayer master and start a task
2. Kill 1 monster for that task - it will drop some amount of slayer XP
3. Note that kills left counter under the XP tracker for slayer xp
4. It will be off by a lot

**Expected behavior**
Kills counter would take the xp per kill of the current slayer task from the table of xp drops per monster type. Then it would divide the remaining xp by the xp per kill and take the ceil of that value to get the remaining kills.

**Screenshots**
6_kills_left
5_kills_left
Notice how in the first screen cap there is 60 xp gained of slayer (the amount per moss giant in this specific example) but with over 1000 xp remaining the XP tracker still estimates only 6 kills to level up. Then the kills to level up goes down to 5 but this is only after 2 additional kills (which still doesn't really make sense).

 - OS: Windows
 - RuneLite version: [e.g 1.4.6]
 - Launcher version: [e.g 1.5.2]

**Additional context**

daviscook477:
this fixes #6614 and could fix some other potential issues by making sure that the new combat xp per kill system from (https://github.com/runelite/runelite/commit/a97b4384f6ab960055ead3dc3fd213d2d0bad041) only is applied to the specified list of combat skills and then in order to make slayer still use the word kills to describe its xp actions it has to add in a text label check specific to slayer.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpState.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpState.java
@@ -170,7 +170,7 @@ class XpState
 
 			action.setActionsHistoryInitialized(true);
 		}
-		
+
 		interactedNPC = npc;
 		state.setActionType(XpActionType.ACTOR_HEALTH);
 	}


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -197,7 +197,11 @@ private void rebuildAsync(boolean updated, boolean skillPaused, XpSnapshotSingle
 			// Update information labels
 			expGained.setText(htmlLabel("XP Gained: ", xpSnapshotSingle.getXpGainedInSession()));
 			expLeft.setText(htmlLabel("XP Left: ", xpSnapshotSingle.getXpRemainingToGoal()));
-			actionsLeft.setText(htmlLabel(xpSnapshotSingle.getActionType().getLabel() + ": ", xpSnapshotSingle.getActionsRemainingToGoal()));
+			// Note that slayer is a special case since it works like an EXPERIENCE skill but should be labeled as
```
deathbeam: I would probably not special case slayer because 1. it is ugly, 2. you need to special case it everywhere where the label is used

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -263,10 +264,11 @@ public void onExperienceChanged(ExperienceChanged event)
 		state.setActionType(XpActionType.EXPERIENCE);
 
 		final Actor interacting = client.getLocalPlayer().getInteracting();
-		if (interacting instanceof NPC)
+		if (interacting instanceof NPC && COMBAT.contains(skill))
 		{
 			final NPC npc = (NPC) interacting;
 			xpState.updateNpcExperience(skill, npc, npcManager.getHealth(npc.getName(), npc.getCombatLevel()));
+			state.setActionType(XpActionType.ACTOR_HEALTH);
```
deathbeam: Why is this line necessary? I dont think it is

daviscook477: it defaults all skills to the experience type. this line was previously inside the updateNpcExperience method but I felt that it would be easier to understand what the program was doing if the actionType was set in one place (i.e. only in the onExperienceChanged method) so I refactored it to here.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
daviscook477: should be addressed by d18bb493b2443d381984f40440cb466c94d78936 - refactored the action type into two parts the label and the task (so now label can be either actions or kills and the task can be either experience or actor health). now it still tracks actions for experience vs actor health separately but the label for those actions can be changed to actions or kills whenever needed. So this makes it possible to set the label for slayer's experience task type to kills in the onExperienceChanged method.

daviscook477:
also this pr fixes a bug that we don't have a report for where if a player performs an action and then interacts with an npc it will treat that skill as a combat skill. Meaning that burying bones could become kills.
bones_are_kills

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -197,7 +197,9 @@ private void rebuildAsync(boolean updated, boolean skillPaused, XpSnapshotSingle
 			// Update information labels
 			expGained.setText(htmlLabel("XP Gained: ", xpSnapshotSingle.getXpGainedInSession()));
 			expLeft.setText(htmlLabel("XP Left: ", xpSnapshotSingle.getXpRemainingToGoal()));
-			actionsLeft.setText(htmlLabel(xpSnapshotSingle.getActionType().getLabel() + ": ", xpSnapshotSingle.getActionsRemainingToGoal()));
```
Adam-: Don't reformat code in otherwise untouched files.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -36,6 +36,7 @@
 import java.util.List;
 import java.util.Objects;
 import javax.inject.Inject;
+
```
Adam-: remove this line.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
deathbeam: I liked it more in the updateNpcExperience

deathbeam:
Any update on this?

daviscook477:
Yeah I'll clean up the code with those changes in like an hour

Issue #6633, opened 2018-11-22
[GPU] White squares shows on pillars, but not in CPU rendering


JorgenRe1:
Location is the temple where you kill the dog in the Priest in Peril quest. 
In that temple every wooden pillar on the wall has a white square on its bottom. Something that is not shown when not using GPU rendering.
Lighting seems to be very different too.

**To Reproduce**
1. Go to Priest in Peril temple undeground with the dog.
2. Turn on GPU rendering 
3. See white squares on the wooden pillars on the walls.

**Expected behavior**
CPU and GPU rendering should be equal.

**Environment (please complete the following information):**
 - OS: Windows
 - RuneLite version: 1.5.1
 - Launcher version: ??

Adam-:
Fixed in #6634

PR #6634, opened 2018-11-22
GPU: Fix Untextured Faces

DevDennis:
Before:
image
After:
image


Adam-:
Why does this fix that? are the uvs negative?

DevDennis:
Yes apparently uvs can be negative

Nightfirecat:
Fixes #6633 maybe?

Issue #6540, opened 2018-11-17
Volcanic Abyssal Whip missing textures with GPU plugin


2Milky:
**Describe the bug**
Equipping the Abyssal Whip with the Volcanic Mix applied to it causes the volcanic texture/animation to disappear from the whip. I'd assume this is also the case with the Frozen Mix as well (though I don't have one to confirm my assumption).

**To Reproduce**
1. Equip an Abyssal Whip that has the Volcanic Mix added to it
2. In the RuneLite plugin list, turn the GPU plugin on and off to notice the difference

**Expected behavior**
Expected the Volcanic Abyssal Whip to have its regular textures

**Screenshots**
https://imgur.com/a/jPDcrVu

**Environment (please complete the following information):**
 - OS: Windows 10 Pro Version 10.0.17134
 - RuneLite version: 1.5.0
 - Launcher version: Unknown
 - CPU: Intel i7 4790K
 - GPU: GTX 1080

raiyni:
it does that on the mobile renderer too

PR #6635, opened 2018-11-22
GPU: Fix Textured Faces With Alpha

DevDennis:
The software renderer ignores the alpha when drawing textured faces

Fixes: https://github.com/runelite/runelite/issues/6540

Before:
image
image

After:
image
image


Nightfirecat:
Does this fix the canoe interface as well? I remember the water texture background looks a lot like that on GPU currently.

PR #6636, opened 2018-11-22
Persistent boss drop log

Poppro:
Adds a plugin to track popular in-game boss drops.
Tracks bosses via the plugin config to provide a persistent, account-based experience.

**Features**
- Choose from list of bosses to track
- Automatically rescales UI to account for large drop-tables
- Persistent tracking independent of local storage
- Tracks loot for seperate in-game accounts automatically
- Ability to reset boss log via GUI
- Displays boss uniques to provide completionist log
- Displays tracked KC
- Displays total profit of tracked boss.

**Images**
screen shot 2018-11-22 at 2 47 14 pm

screen shot 2018-11-22 at 2 45 53 pm

screen shot 2018-11-22 at 2 47 48 pm

screen shot 2018-11-22 at 2 48 14 pm


Hydrox6:
Why is this a separate plugin?

Poppro:
Just seemed to me at the time I started development that the idea of the plugin was departed enough from a general non-persistent loot tracker that it would be fine to have it separate. If it's controversial as a separate plugin I think that it would certainly make sense as a "boss log" sub-menu of the current loot tracker plugin though.

raiyni:
Having the option for every boss isn't necessary imo

deathbeam:
I am pretty sure jagex is implementing this for their collection log (or w/e it is named).

Adam-:
I think Jagex is implementing this "soon", so it is probably not worth adding a plugin for.

PR #6639, opened 2018-11-22
Add UI scaling when using GPU with stretched mode.

dbolya:
This changes the previous behavior, which instead of just scaling the UI texture, rendered the entire game into a smaller framebuffer and stretched that. This is much closer to actual UI scaling, in that we render the world view at its native resolution and only scale up the UI.

This should support all the options in the stretched mode menu as well as both fixed and resizeable mode.

Before:
image

After:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -72,14 +72,12 @@ import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.plugins.PluginInstantiationException;
 import net.runelite.client.plugins.PluginManager;
 import static net.runelite.client.plugins.gpu.GLUtil.glDeleteBuffer;
-import static net.runelite.client.plugins.gpu.GLUtil.glDeleteFrameBuffer;
-import static net.runelite.client.plugins.gpu.GLUtil.glDeleteRenderbuffers;
 import static net.runelite.client.plugins.gpu.GLUtil.glDeleteTexture;
 import static net.runelite.client.plugins.gpu.GLUtil.glDeleteVertexArrays;
+import static net.runelite.client.plugins.gpu.GLUtil.glDeleteFrameBuffer;
 import static net.runelite.client.plugins.gpu.GLUtil.glGenBuffers;
-import static net.runelite.client.plugins.gpu.GLUtil.glGenFrameBuffer;
-import static net.runelite.client.plugins.gpu.GLUtil.glGenRenderbuffer;
 import static net.runelite.client.plugins.gpu.GLUtil.glGenTexture;
+import static net.runelite.client.plugins.gpu.GLUtil.glGenFrameBuffer;
 import static net.runelite.client.plugins.gpu.GLUtil.glGenVertexArrays;
 import static net.runelite.client.plugins.gpu.GLUtil.inputStreamToString;
 import net.runelite.client.plugins.gpu.template.Template;
@@ -148,17 +146,20 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 	private int glUiVertexShader;
 	private int glUiFragmentShader;
 
+	private int glUiPremulProgram;
+	private int glUiPremulVertexShader;
+	private int glUiPremulFragmentShader;
+
 	private int vaoUiHandle;
 	private int vboUiHandle;
+
+	private int fboUiHandle;
+	private int texUiHandle;
 
 	// scene vertex buffer id
 	private int bufferId;
 	// scene uv buffer id
 	private int uvBufferId;
-
-	private int fboStretchedHandle;
-	private int texStretchedHandle;
-	private int rboStretchedHandle;
 
 	private int textureArrayId;
 
@@ -204,7 +205,6 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 	private int lastViewportHeight;
 	private int lastCanvasWidth;
 	private int lastCanvasHeight;
-	private Dimension lastStretchedDimensions;
 
 	private int centerX;
 	private int centerY;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -879,10 +807,27 @@ else if (lastStretchedDimensions != null)
...
+			int renderViewportHeight = viewportHeight;
+			int renderViewportWidth = viewportWidth;
 
-			gl.glViewport(widthOff, canvasHeight - viewportHeight - heightOff, viewportWidth, viewportHeight);
+			if (client.isStretchedEnabled())
+			{
+				Dimension dim = client.getStretchedDimensions();
+				renderCanvasHeight = dim.height;
+
+				float scaleFactorY = dim.height / (float) canvasHeight;
```
devLotto: Use dim.getHeight() instead of dim.height and you won't have to cast canvasHeight to float.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -991,7 +921,18 @@ private void drawUi(final int canvasHeight, final int canvasWidth)
...
-		gl.glViewport(0, 0, canvasWidth, canvasHeight);
+
+		int renderCanvasWidth = canvasWidth;
+		int renderCanvasHeight = canvasHeight;
+
+		if (client.isStretchedEnabled())
+		{
+			Dimension dim = client.getStretchedDimensions();
+			renderCanvasWidth = dim.width;
+			renderCanvasHeight = dim.height;
```
devLotto: Maybe just do
```java
if (client.isStretchedEnabled())
{
	Dimension dim = client.getStretchedDimensions();
	gl.glViewport(0, 0, dim.width, dim.height);
}
else
{
	gl.glViewport(0, 0, canvasWidth, canvasHeight);
}
```

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1021,6 +962,14 @@ private void drawUi(final int canvasHeight, final int canvasWidth)
 		gl.glBindTexture(gl.GL_TEXTURE_2D, interfaceTexture);
 		gl.glUniform1i(uniTex, 0);
 
+		// If necessary, change the sampling function. Honestly, this should be using OpenGL
```
devLotto: I guess "change the sampling function" is what the code does, and what follows in the comment is potential improvement. Maybe those two parts can be split up with a newline to make it less confusing what the comment is trying to convey.

What the code does can probably be clarified, maybe "Sets the sampling function used when stretching the UI".

dbolya: Yeah comment there is a bit cryptic. Fixed that (and the line break). Also I was talking about [sampler objects](https://www.khronos.org/opengl/wiki/Sampler_Object). That's the _correct_ way to do it, but meh. Should I even leave that part of the comment in?

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
dbolya: Good idea, changed.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
dbolya: Fixed.

oabdellatif:
If this is merged, I think the name of the plugin should be changed to "Interface Scaling" to better reflect its new behavior. "Stretched Mode" could be misinterpreted to still refer to upscaling the contents of the game window.

devLotto:
@oabdellatif It still stretches the scene on CPU so the name should be kept.

Adam-:
I think the text looking weird is due to the stretching method since if you turn on increased perf mode it affects it.

dbolya:
It looks like it might not be possible to _perfectly_ position the viewport. For instance, if you had a scale factor of 1.5 and an x offset in fixed mode of 1 pixel, then the viewport would need to start at x=1.5. Since we can only give integer values to glViewport, I have it start at x=1.

Now the that means that objects in the UI are up to 1 pixel off of objects in the world view. I'm thinking that's an acceptable amount of error?

I also added some padding to get rid of the black borders in fixed mode, so that should be fixed:
image

dbolya:
Pushed a change that should make that one pixel wide text more legible by always using nearest neighbors for the alpha channel.

Before:
image

After:
image

This is kind of an imperfect solution because all the text is still darker (since the pixels around the text are black with alpha 0 and opengl blends that black in anyway), but I guess it's better.

dbolya:
So I did some searching and it looks like the exact problem we were having is best solved by pre-multiplying the alpha into the color and then using `glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA)`.

So I implemented that and reverted the previous shader change. Looks better now:
image

It's blurrier, but that's because it's properly doing bilinear sampling now (as in now it looks the same as stretched cpu mode).

Also I had to sort of unpack and repack the pixels array to achieve this. Let me know if there is a better way / if it's not a good idea to modify the pixels array directly.

outpoints:
This is exactly what I've wanted!

Good luck! If you need any help feel free to post some code and I'll take a look! :)

devLotto:
How is spending milliseconds iterating a pixel array better than just using an FBO?

dbolya:
It's not. Actually in the process of implementing a second pass with an FBO right now.

dbolya:
Ok I added back the FBO and moved the alpha pre-multiplication step to the GPU. I had to make this a new shader instead of using the old one and a boolean uniform because rendering to a framebuffer texture is a little different. I also only do the first pass when scaling is on and the scaling is linear.

Man, that was a lot of code for such a simple thing, but we do two passes now if we want to use the first pass for anything extra in the future.

devLotto:
Works well.

Adam-:
Thanks!

Issue #6576, opened 2018-11-19
"Kills Left" reports incorrectly for split-exp attack types


P-Squiddy:
it looks like the "Kills Left" is taking the exp you'd get for a specific single-attack type and using that to figure out how many kills it should have, instead of the actual exp amounts.

I'm attacking Bandits in the Desert. According to my "Kills Left" each enemy should give me 200 exp per kill, which is correct since i'm getting 85 exp spread accross Attack, Strength and Defense, but it tells me it's going to take 54 kills to get the 10.6k exp to get from level 50 -> 51 at 85 exp per enemy...which isn't right at all.

I'm rather certain of this as switching from Lunge -> Block with a Brine Sabre is now giving roughly 200 exp per enemy, making the "Kills Left" report correctly.

I'm noticing that the HP "Kills Left" is correct in either case because exp isn't split for it.

daviscook477:
Okay I confirmed that this is happening. It is in fact taking the exp for a single-attack type and using that to calculate the xp you get for a combat skill while attacking an npc. Yeah the HP kills left is calculated separately from the other combat skills right now but it looks like we have to check what combat style the player is using when performing the estimated xp drop per kill calculation in the xp tracker plugin.

travisclare:
Still having this issue on the newest version

PR #6643, opened 2018-11-23
Fix controlled combat styles estimating kill counts wrong

daviscook477:
Fixes (https://github.com/runelite/runelite/issues/6576).

Duplicates a lot of code from the attackstyles plugin. Basically just need to extend what attackstyles does in tracking the players current combat style to also provide the xp per hp damage dealt info from the wiki (https://oldschool.runescape.wiki/w/Combat_Options) on a per skill basis for each combat style. Probably should be refactored but does work.


Nightfirecat:
This should definitely be refactored if you're going to copy hundreds of lines of code from another plugin.

daviscook477:
so the way I would want to refactor the attackstyle stuff is to have an AttackStylesManager in the client.game package. But the main issue I see with this is that the attackstyle plugin needs to have it's warnings be fired only after the attack style is changed (which is seen in the onVarbitChanged event. So a possible race condition if the attackstyle tracking is refactored out of the attackstyles plugin itself is that the onVarbitChanged event fires for the attackstylesplugin before the AttackStylesManager gets the onVarbitChanged event to update the attack style.

A potential solution is to have a publisher/subscriber system where things can subscribe to an attackstyle changed event list in the AttackStylesManager that would then fire only after the attackstyle was changed within the AttackStylesManager.

Thoughts?

daviscook477:
@Nightfirecat I made the requested changes to refactor out the attack style and equipped weapon tracking. Now that tracking is done in its own manager to which both the xp tracker plugin and attack styles plugin subscribe to weaponchanged and attackstylechanged events to get the info they need.

daviscook477:
Realized I was missing copyright headers for a few of the added files. Should be fixed now.

Nightfirecat:
No longer relevant as of 6f956995bb90192e15835bb2831a3dd57e31b5a9

PR #6652, opened 2018-11-23
Truncate username and password string before pasting if necessary

oabdellatif:
Trying to log in after pasting a string that was too long into the username or password field would cause an ArrayIndexOutOfBoundsException beforehand.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
@@ -178,11 +178,13 @@ public class LoginScreenPlugin extends Plugin implements KeyListener
 				// 0 is username, 1 is password
 				if (client.getCurrentLoginField() == 0)
 				{
-					client.setUsername(data);
+					// Truncate data to maximum email length if necessary
+					client.setUsername(data.substring(0, Math.min(data.length(), 254)));
 				}
 				else
 				{
-					client.setPassword(data);
+					// Truncate data to maximum password length if necessary
+					client.setPassword(data.substring(0, Math.min(data.length(), 20)));
 				}
 			}
 			catch (UnsupportedFlavorException | IOException ex)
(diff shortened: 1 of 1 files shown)


Adam-:
Maybe set these limits as constants?

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -174,15 +174,18 @@ public void keyPressed(KeyEvent e)
 					.getData(DataFlavor.stringFlavor)
 					.toString()
 					.trim();
+				final int maxEmailLength = 254, maxPasswordLength = 20;
```
deathbeam: constant in java = private/public static final, this is just named variable

oabdellatif: Alright, think I got that right this time in the latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -62,6 +62,8 @@
 
 	private String usernameCache;
 
+	private static final int MAX_EMAIL_LENGTH = 254, MAX_PASSWORD_LENGTH = 20;
```
deathbeam: And constants go at top of the file, before instance fields

deathbeam: Oh, and also, MAX_EMAIL_LENGTH is wrong, because its username/email (for older accounts), so I would just name this username. And, declare the constants on separate lines

Joshua-F: I think `MAX_LOGIN_LENGTH` would probably cover email and username.

PR #6655, opened 2018-11-24
Improve mining status detection at motherlode mine

### LINKED ISSUE #1377: MLM returns false positives on idle
The top left corner mining plugin works great, but when mining in Motherlode Mine it shows I'm idle between animations but while the vein is not depleted yet.

### LINKED ISSUE #9052: Idle Notifier - Notify immediately if someone else depletes the same vein in MLM
Would it be possible for idle notifier to notify as soon as the ore vein is depleted in MLM if multiple people are on the same vein? (Mostly helps with upstairs mining)

Currently if you deplete a vein yourself, idle notifier goes off instantly, but if there are multiple players on the same vein and someone else depletes it, your animation continues and the notification is delayed until it finishes...

Twiglet1022:
Fixes #1377
Fixes #9052

I've split the commits up into three parts:

1. Fix the mining status detection so it actually works.
2. Move the idle notification into the motherlode plugin to use the new logic.
3. Leverage the improved detection to add an option to recolour the icon on the vein currently being mined. I think it looks nice and it's more noticeable than the mining/not mining overlay because it's in the centre of the screen. Default off.

image

## How it works:
When you click the mine option on a vein the plugin registers the location of that vein.

The idle notification will then trigger if any of these conditions are met:

1. The vein at that location depletes i.e. a depleted vein spawns there
2. Check widgets for full inventory or level up
3. As a fallback option for edge cases, check the animation with a 1800ms tolerance for going idle

Additionally there is one other thing that will register as not mining, but not send a notification:

1. When the player moves away from the target vein

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
@@ -182,6 +182,7 @@ public class MotherlodePlugin extends Plugin
 
 	@Getter(AccessLevel.PACKAGE)
 	private boolean isMining;
+	@Getter(AccessLevel.PACKAGE)
 	private WorldPoint targetVeinLocation = null;
 	private boolean playerHasReachedTargetVein;
 	private int lastAnimation = AnimationID.IDLE;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeRocksOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeRocksOverlay.java
@@ -40,11 +40,13 @@ import net.runelite.api.Point;
 import net.runelite.api.Skill;
 import net.runelite.api.WallObject;
 import net.runelite.api.coords.LocalPoint;
+import net.runelite.api.coords.WorldPoint;
 import net.runelite.client.game.SkillIconManager;
 import net.runelite.client.ui.overlay.Overlay;
 import net.runelite.client.ui.overlay.OverlayLayer;
 import net.runelite.client.ui.overlay.OverlayPosition;
 import net.runelite.client.ui.overlay.OverlayUtil;
+import net.runelite.client.util.ImageUtil;
 
 class MotherlodeRocksOverlay extends Overlay
 {
@@ -55,6 +57,9 @@ class MotherlodeRocksOverlay extends Overlay
 	private final MotherlodeConfig config;
 
 	private final BufferedImage miningIcon;
+	private final BufferedImage targetMiningIcon;
+	private static final Color miningIconOldColor = new Color(117, 123, 124);
+	private static final Color miningIconNewColor = new Color(0, 150, 0);
 
 	@Inject
 	MotherlodeRocksOverlay(Client client, MotherlodePlugin plugin, MotherlodeConfig config, SkillIconManager iconManager)
@@ -66,6 +71,7 @@ class MotherlodeRocksOverlay extends Overlay
 		this.config = config;
 
 		miningIcon = iconManager.getSkillImage(Skill.MINING);
+		targetMiningIcon = ImageUtil.recolorImage(miningIcon, miningIconNewColor, Color -> Color.getRGB() == miningIconOldColor.getRGB());
 	}
 
 	@Override
@@ -86,7 +92,6 @@ class MotherlodeRocksOverlay extends Overlay
 	private void renderTiles(Graphics2D graphics, Player local)
 	{
 		LocalPoint localLocation = local.getLocalLocation();
-
 		if (config.showVeins())
 		{
 			for (WallObject vein : plugin.getVeins())
@@ -97,7 +102,16 @@ class MotherlodeRocksOverlay extends Overlay
 					// Only draw veins on the same level
 					if (plugin.isUpstairs(localLocation) == plugin.isUpstairs(vein.getLocalLocation()))
 					{
-						renderVein(graphics, vein);
+						if (WorldPoint.fromLocal(client, location).equals(plugin.getTargetVeinLocation())
+							&& plugin.isMining()
+							&& config.showTargetVein())
+						{
+							renderVein(graphics, vein, true);
+						}
+						else
+						{
+							renderVein(graphics, vein, false);
+						}
 					}
 				}
 			}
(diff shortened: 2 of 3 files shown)


runelite-api/src/main/java/net/runelite/api/coords/WorldPoint.java
```diff
@@ -261,6 +261,40 @@ public int distanceTo2D(WorldPoint other)
...
+	/**
+	 * Find the straight-line distance from this point to another point.
+	 * <p>
+	 * This method disregards the plane value of the two tiles and returns
+	 * the simple distance between the X-Z coordinate pairs.
+	 *
+	 * @param other other point
+	 * @return the straight-line distance
+	 */
+	public float distanceTo2DHypotenuse(WorldPoint other)
```
deathbeam: LocalPoint already contains distanceTo, so you should probably use that

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeOverlay.java
```diff
@@ -90,7 +80,7 @@ public Dimension render(Graphics2D graphics)
 
 		if (config.showMiningState())
 		{
-			if (MINING_ANIMATION_IDS.contains(client.getLocalPlayer().getAnimation()))
+			if (!plugin.isIdleMotherlode)
```
deathbeam: Use getters and not direct field access

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -134,6 +140,17 @@
...
+	private WorldPoint targetVeinLocation = null;
+	@Getter(AccessLevel.PACKAGE)
+	private float distanceFromTargetVein;
+
+	private boolean isIdleInventoryFull;
+	private boolean isIdleLevelUp;
+	private boolean isIdleTargetVeinDepleted;
+	private boolean reachedTargetVein;
+	private boolean isIdleNotNextToTargetVein;
+	boolean isIdleMotherlode;
```
deathbeam: getter

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -257,6 +274,89 @@ public void checkMining()
 		}
 	}
 
+	@Schedule(
+		period = 200,
+		unit = ChronoUnit.MILLIS
+	)
+	public void checkVeinDepleted()
```
deathbeam: You can just use GameTick

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -257,6 +274,89 @@ public void checkMining()
...
+			isIdleTargetVeinDepleted = true;
+			targetVeinLocation = null;
+		}
+	}
+
+	@Schedule(
+		period = 200,
+		unit = ChronoUnit.MILLIS
+	)
+	public void checkDistanceToTargetVein()
```
deathbeam: Same here

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -257,6 +274,89 @@ public void checkMining()
...
+		{
+			isIdleNotNextToTargetVein = true;
+		}
+	}
+
+	@Schedule(
+		period = 200,
+		unit = ChronoUnit.MILLIS
+	)
+	public void checkMotherlodeIdle()
```
deathbeam: Same here

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -368,6 +468,54 @@ else if (event.getGameState() == GameState.LOGIN_SCREEN)
 		}
 	}
 
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked menu)
+	{
+		int id = menu.getId();
```
deathbeam: cant you get w/e you need from MenuOptionClicked event? You should not need to iterate menu entries

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeRocksOverlay.java
```diff
@@ -66,6 +69,7 @@
 		this.config = config;
 
 		miningIcon = iconManager.getSkillImage(Skill.MINING);
+		targetMiningIcon = ImageUtil.getResourceStreamFromClass(getClass(), "targetMining.png");
```
deathbeam: You do not need to load new recolored image, just recolor it here from Java

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
Twiglet1022: I need a MenuEntry to get param0 and param1 which correspond to the x and y coordinates of LocalPoint representing the location of the object that was clicked. I don't see any way to get that directly from MenuOptionClicked. Am I missing something?

trimbe: `actionParam` is x and `widgetId` is y for clicks on game objects.

trimbe: It's not documented anywhere, I only knew it from looking at the Kourend Library plugin.

runelite-api/src/main/java/net/runelite/api/coords/WorldPoint.java
Twiglet1022: I couldn't get LocalPoint to work (without tweaking the fromScene method in LocalPoint) after trying for quite a while so I've reverted to using WorldPoint and it works great. I need that method to calculate a proper diagonal distance and detect if the player is one tile away to the side.

If you're okay with that then I think I'm done, unless you'd like any other changes.

Twiglet1022:
My ultimate goal with this PR was to hook it up the idle notifier to make it work properly without needing a long delay. I've got it working in another branch. Would it be better to wait for this to be merged first or can I just add a commit for it to this?

Twiglet1022:
As I've been spending a lot of time at MLM and this was one of my oldest PRs I've given it a much needed rewrite. It's not as trivial to detect when you're idle while mining as it is with fishing, but hopefully this isn't considered too complex.

Twiglet1022:
Just discovered that all the animations called MINING_MOTHERLODE_XXX are actually used at Amethyst too (probably any wall mining?)

So moving the motherlode idle notification into the motherlode plugin actually breaks the idle animation notification for those things. So I'll need to try to fix that.

PR #6665, opened 2018-11-24
Add a skull timer

### LINKED ISSUE #6980: Suggestion: Skull timer
frustrated when [...]

I forget for how long im skulled

show an icon left top for how long im skulled

A clear and concise description of any alternative solutions or features you've
considered.

**Additional context**

I hope this isnt too much trouble to implement.

Trevor159:
WIP PR
EDIT: no longer WIP and closes #6980

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -549,8 +549,16 @@ public class TimersPlugin extends Plugin
 		{
 			SkullIcon currentSkullIcon = player.getSkullIcon();
 			int currentRegionID = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
-			Item[] equippedItems = client.getItemContainer(InventoryID.EQUIPMENT).getItems();
-			boolean avariceThisTick = (equippedItems[2].getId() == ItemID.AMULET_OF_AVARICE);
+			Item[] equippedItems;
+			try
+			{
+				equippedItems = client.getItemContainer(InventoryID.EQUIPMENT).getItems();
+			}
+			catch (NullPointerException exception)
+			{
+				equippedItems = null;
+			}
+			boolean avariceThisTick = equippedItems != null && equippedItems[2].getId() == ItemID.AMULET_OF_AVARICE;
 
 			//Enter the abyss
 			if (currentRegionID != lastRegionID && currentRegionID == abyssRegionID && currentSkullIcon == SkullIcon.SKULL)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -29,22 +29,7 @@ import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.Actor;
-import net.runelite.api.AnimationID;
-import net.runelite.api.ChatMessageType;
-import net.runelite.api.Client;
-import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.GameState;
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemContainer;
-import net.runelite.api.ItemID;
-import net.runelite.api.NPC;
-import net.runelite.api.NpcID;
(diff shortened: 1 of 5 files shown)


Nightfirecat:
A 20-minute timer isn't universally correct; [see the skull wiki page](https://oldschool.runescape.wiki/w/Skull_(status)):

> The "skull" will disappear automatically after 10 or 20 minutes depending on how the player got it; it disappears after 10 minutes after using the Abyss, and 20 after attacking a player. If a player re-enters the Abyss or attack another player that has not attacked them yet, the skull timer will reset. Having the amulet of avarice worn will cause the player to be indefinitely skulled up until 20 minutes after unequipping it.

15987632:
> A 20-minute timer isn't universally correct; [see the skull wiki page](https://oldschool.runescape.wiki/w/Skull_(status)):
> 
> > The "skull" will disappear automatically after 10 or 20 minutes depending on how the player got it; it disappears after 10 minutes after using the Abyss, and 20 after attacking a player. If a player re-enters the Abyss or attack another player that has not attacked them yet, the skull timer will reset. Having the amulet of avarice worn will cause the player to be indefinitely skulled up until 20 minutes after unequipping it.

Ya, we are talking about edge cases in discord. It's going to be more of a pain than I thought. There is also the amulet of avarice. Also what happens if you enter pvp with the abyss skull timer on.

15987632:
This doesn't check if somebody had an abyss skull and went into pvp. I don't know of a way to do that intuitively, I doubt it even changes the behavior of the skull, and I don't think it is a needed thing to be included. The abyssal bracelet does not proc when you have a skull up so it isn't needed to be checked at all.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -518,6 +529,46 @@ public void onGameTick(GameTick event)
 		Player player = client.getLocalPlayer();
 		WorldPoint currentWorldPoint = player.getWorldLocation();
 
+		SkullIcon currentSkullIcon = player.getSkullIcon();
+		int currentRegionID = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
+		final int abyssRegionID = 12107;
```
daviscook477: should probably be a private static final int at the start of the file so it's more clearly a constant definition

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -518,6 +529,46 @@ public void onGameTick(GameTick event)
...
+		if (equippedItems[2].getId() == ItemID.AMULET_OF_AVARICE)
+		{
+			avariceThisTick = true;
+		}
+		else
+		{
+			avariceThisTick = false;
+		}
+
+		if (currentRegionID != lastRegionID && currentRegionID == abyssRegionID && currentSkullIcon == SkullIcon.SKULL)
```
daviscook477: not sure if you want to but for this series of if/else clauses it might be a good idea to comment each one with each of the wilderness skull interactions it is checking for - that way it's easier for maintainers to know what skull check each is performing. I.e. something like for this line `else if (avariceLastTick == true && avariceThisTick == false)` could be "only start the skull timer for the amulet of avarice once it is removed since the timer won't start ticking down until the amulet is taken off"

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
15987632: I will use your first suggestion, as to my knowledge there isn't a regionID api of any sort yet.

15987632:
Just thought I should leave a comment as this isn't exactly finished. I need to figure out how to refresh the timer in pvp and see if hopping worlds changes it at all.

15987632:
I think this is as good as it will get functionality wise. Reviews are welcomed. The known edge cases are that this wont work if you are refreshing your skull on somebody who has previously attacked you but relogged recently. There is no way for me to check if another player has relogged as far as I can tell. The other known edge case is if somebody's animation changed while interacting with the lcoal player but not attacking them it will count as if they were attacking the local player (for example if someone throws a snowball at the player) and wont include them in the cantSkullOn map.

Spedwards:
@15987632 "There is no way for me to check if another player has relogged as far as I can tell." - There is but it might be out of the question. That's using their total exp. When a player logs out, their experience is updated on the hiscores. Though it probably is not worth storing every player you come into contact with's exp.

Trevor159:
1. This is terrible
2. There isn't a good way to do this and there always will be edge cases

PR #6669, opened 2018-11-24
Add Kept on Death plugin

TheStonedTurtle:
Modifies the Items Kept on Death widget to interactive and slightly more accurate. If the player is inside a safe zone the normal interface will be shown.

### Although this is an improvement on the original interface there may be situations where the new interface is wrong. Please report these cases so we can improve the interface further

It will automatically detect if the player is Skulled, Protecting Item, and the Wilderness safety (Low Lvl/High). It also support PvP world Safe/Non Safe areas.

## Death Mechanics

<details>
<p>

### Known Death Mechanics:
These mechanics should never change.

1) Every account protects 3 items.
2) Skulled account loses all Items
3) Protect item keeps an additional item based on value
4) UIM accounts lose all items, like they were Skulled.

### Assumed Death Mechanics:
These mechanics were implemented due to server instability and are still not fully understood.

1) **Outside wildy (Non-PvP death)**: Most non-tradeable items are kept (ie, Looting bag is lost)
2) **Lvl 1-20 wildy**: Non-tradeable items are lost on death unless they are one that turns into a broken variant.
3) **Above 20 wildy**: Non-tradeable items are lost on death and turned into coins.

The assumed death mechanics apply after the Known death mechanics which means You can keep untradeable items above 20 wildy if they are kept by the Known death mechanics.

</p>
</details>


runelite-client/src/main/java/net/runelite/client/plugins/keptondeath/ActuallyTradeableItem.java
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	ActuallyTradeableItem(int... ids)
+	{
+		this.ids = ids;
+	}
+
+	public static Boolean check(int id)
+	{
+		return ID_MAP.get(id) != null;
```
raiyni: use a hash set

runelite-client/src/main/java/net/runelite/client/plugins/keptondeath/BrokenOnDeathItem.java
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	BrokenOnDeathItem(int itemID)
+	{
+		this.itemID = itemID;
+	}
+
+	public static boolean check(int itemID)
+	{
+		return ID_MAP.get(itemID) != null;
```
raiyni: use a hash set

deathbeam:
Imo this could be placed inside the death indicator plugin, not sure

daviscook477:
The Updated Skulled reminder text doesn't seem too helpful. Why remove the bit about "the normal amount of items you keep is 3. You're marked with a PK skull ....". And the bit that says "Protecting 0 items by default PK skull -3" isn't really all that clear.

daviscook477:
Also I'm not too familiar with the wilderness mechanics but why would the bag looking thing with the white outline be lost in <20 wilderness but kept when skulled. Is that right?

TheStonedTurtle:
@daviscook477 The Rune Pouch has a white outline by default but is protected on deaths in non-pvp scenarios. The skulled pictures were in a safe zone so it was added to the top part without the white outline. I guess I could remove the white outline in both cases.

As for why I updated the text, its extremely cluttered with the default text and I added two new sentences which would not have fit otherwise. This photo doesn't even have the bond text.

daviscook477:
Maybe removing the white outline would be helpful. It's a bit misleading for the game to say "items with a white outline will always be lost" and then it not be true. I still think the "Protecting 0 items by default PK skull -3" is a bit confusing. What is that text trying to communicate?

TheStonedTurtle:
The interface has been reworked to be interactive with toggles for the different effects. See gif:


A few use cases were also addressed:
* Added PvP world support
* Show default interface inside safe zones (PoH/Clan Wars/etc)
* Fixed incorrect interface being shown when widget is rebuilt

TheStonedTurtle:
Value of items doesn't work correctly for noted items
Max items kept on death doesn't account for stacked items



Looks like rune pouch doesn't count towards total but does currently? Not sure about the other AlwaysLost items



Extra blank row for kept items? (Barrows is degraded btw)

TheStonedTurtle:
I fixed a few minor things, but I'm unsure of how the game handles ornate items in the wilderness. I.E. occult ornate is 4.8m (3.9m kit, 900k occult). What value does the game use?

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/AlwaysLostItem.java
```diff
@@ -0,0 +1,65 @@
...
+@AllArgsConstructor
+@Getter
+enum AlwaysLostItem
+{
+	RUNE_POUCH(ItemID.RUNE_POUCH, true),
+	LOOTING_BAG(ItemID.LOOTING_BAG, false),
+	CLUE_BOX(ItemID.CLUE_BOX, false);
+
+	private final int itemID;
+	private final boolean kept;
```
deathbeam: This should be named keptOutsideOfWilderness, or at least documented like that, otherwise it is very confusing to have enum named "AlwaysLostItem" and then field inside named "kept". I had to actually go through the plugin implementation to understand what this field means.

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/AlwaysLostItem.java
```diff
@@ -0,0 +1,65 @@
...
+	{
+		final ImmutableMap.Builder<Integer, AlwaysLostItem> map = ImmutableMap.builder();
+		for (final AlwaysLostItem p : values())
+		{
+			map.put(p.itemID, p);
+		}
+		ID_MAP = map.build();
+	}
+
+	static AlwaysLostItem getByItemID(final int itemID)
```
deathbeam: We should keep some consistent naming for these methods, e.g clue scroll plugin is using forItemId

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/BrokenOnDeathItem.java
```diff
@@ -0,0 +1,111 @@
...
+	{
+		final ImmutableSet.Builder<Integer> set = new ImmutableSet.Builder<>();
+		for (final BrokenOnDeathItem p : values())
+		{
+			set.add(p.itemID);
+		}
+		ID_SET = set.build();
+	}
+
+	static boolean check(final int itemID)
```
deathbeam: check?

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -0,0 +1,582 @@
...
+	}
+
+	/**
+	 * Creates an Item Widget for use inside the Kept on Death Interface
+	 *
+	 * @param qty Amount of item
+	 * @param c   Items Composition
+	 * @return
+	 */
+	private Widget createItemWidget(final Widget parent, final int qty, final ItemComposition c)
```
deathbeam: static

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -0,0 +1,582 @@
...
+		max.setText(String.format("<col=ffcc33>Max items kept on death:<br><br><col=ffcc33>~ %d ~", keptQty));
+	}
+
+	/**
+	 * Check if an item is tradeable to another player
+	 *
+	 * @param c The item
+	 * @return
+	 */
+	private boolean isTradeable(final ItemComposition c)
```
deathbeam: static

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -0,0 +1,582 @@
...
+		positionWidgetItems(kept, keptItems);
+		positionWidgetItems(lost, lostItems);
+
+		updateKeptWidgetInfoText(hasAlwaysLost, keptItems, lostItems);
+	}
+
+	/**
+	 * Position a list of widget items in the parent container
+	 */
+	private void positionWidgetItems(final Widget parent, final List<Widget> widgets)
```
deathbeam: static

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -0,0 +1,582 @@
...
+
+	private WidgetButton deepWildyButton;
+	private WidgetButton lowWildyButton;
+
+	private boolean isSkulled;
+	private boolean protectingItem;
+	private int wildyLevel;
+
+	@Subscribe
+	protected void onScriptCallbackEvent(ScriptCallbackEvent event)
```
deathbeam: We use private everywhere else atm

Adam-: I think you mean public?

deathbeam: right, yes :d

PR #6671, opened 2018-11-24
Add fog and skybox

AdamBallinger:
Adds dynamic fog and skybox to GPU plugin that scales with the draw distance.
ccb6af54769134b49c830ad082e0da01

Credits to Psikoi for the skybox texture and kamielvf for the original fog/skybox code.


Adam-:
We've been talking about this a lot internally lately (since this PR appears to be mostly other peoples code). The reasons its not been PRd yet is

1) we don't like the idea of storing a png image this large in git, or at least in the main repository
2) the scene is not circular (it is a square) and so applying a circular fog will hide some parts of the scene that we otherwise could render
3) it is also possible to stand closer to the edge of the scene than your draw distance, which means you would need to move the fog up potentially closer

There are some other issues with the PR I could review, but first we need to figure out these 3 problems (maybe 2 and 3 are the same problem?).

AustinGarrettRichards:
This is amazing and I'd kill for this, even in its current state.

logarrhythmic:
I made some changes to this on my own, but I don't think I have any way to add them to this PR or anything. Namely, I made the fog rectangular and made separate config options for the skybox color and fog rendering. It still doesn't solve problems 1 and 3, but I could do 3 if I knew how to find the edges of whatever part of the world the server is currently allowing the client to render.

On a lesser note, the ratcatcher sewer hole in Port Sarim has the skybox color at the bottom when using this code. That doesn't seem like too much of a problem to me, but it's the first thing I've encountered that feels like someone might have an issue with it.

raiyni:
Because he took the image and made this PR before it was complete.

Issue #1046, opened 2018-03-21
Monsters left on slayer task completely useless when barraging


MagicfTail:
When bursting/barraging slayer monsters you stack them and kill multiple at the same time. This results in multiple monsters dying in the same tick, giving only one slayer xp drop, and the counter only counting one monster dying. This throws it off, making i unusable when bursting slayer tasks.

Adam-:
When more than one person contributes to a slayer monster kill, the xp drop is halved, so you cannot tell solely based on xp drops how many npcs have died that tick. You would have to be able decide who damaged what NPC as well, which is either not possible or not feasable, so I do not think we will be fixing this anytime soon.

rune3132:
Yeah, all the other clients have the same problem, so I imagined it wouldn't be an easy fix. Also, I'm sorry if my wording was a bit harsh, that was in no way my intent

CasvanDongen:
I can look into this, there are some ways to determine the single kill xp drop but they aren't 100% failsafe.

ShaunDreclin:
It may be safe to assume that most of the time you will not be sharing kills, so we could just ignore that and let the counter get slightly out of sync when people do interfere. More important to make multikills work with the counter imo

Adam-:
I think this is probably a good assumption.

lund8619:
Just finished a bursting task and was thinking about this. What do people think of the logic of something along the lines of if xp % hp == 0, kills remaining = kills remaining - xp/hp?

One potential issue I see there is superiors. Thoughts?

lund8619:
It's not a 100% solution, but it'd be more often correct than currently. Are "less wrong" logic fixes ever implemented, or do things like this usually wait for a absolute fix? It looks like Davis Cook's got a much more sophisticated PR that would resolve this, so just curious about the approach to issues like this now.

raiyni:
there's also the fact some monsters have bonus xp

lund8619:
I acknowledged that issue in my first comment. That's why the "if xp % hp == 0" part. The else would just have to default back to the current decrement by 1. So for the philosophical question, I take it the answer is keep the most simplistic logic until somebody comes up with a complete solution and don't bother with the might be better might not type of solution I'm posing here?

Adam-:
fixed in 67030e38a44f355a0131cff5250ed45c7b01dc53

PR #6673, opened 2018-11-25
Wip slayer plugin kill count fix

daviscook477:
Attempts to accurately count slayer kills for the slayer plugin by rather than assuming 1 xp drop = 1 kill, it instead takes the slayer xp drops possible by each monster killed on the tick the xp dropped and then determines which of those monsters most likely died to produce that total xp drop.

Reasons for WIP:

1. Need death animation ids for the following npcs [DUSK_7888, DUSK_7889, GIANT_ROCKSLUG, SMALL_LIZARD, DESERT_LIZARD, ANCIENT_ZYGOMITE]
2. Need to add more tests of the kill estimation
3. Need to do in game test of death animation check for the npcs that die at hp > 0
4. like to have - make partial xp drops factored in

Attemps to for the most part fix (https://github.com/runelite/runelite/issues/1046).
Will likely also fix cannoning on slayer tasks.
Both need to be tested. I cannot personally test either due to a lack of osrs levels/stuff.


daviscook477:
All data was grabbed using (https://github.com/daviscook477/osrs-data/blob/scrape-monster-data/src/main/java/net/runelite/data/dump/wiki/SlayerXpDumper.java)

runelite-client/src/main/resources/slayer_xp.json
```diff
@@ -0,0 +1,4166 @@
...
+    -1.0,
+    172.0,
+    -1.0,
+    257.0
+  ],
+  "Shadow": [
+    -1.0,
+    73.0
+  ],
+  "Me": [
```
MagicfTail: Monsters that don't giver slayer xp shouldn't be on the list

daviscook477:
-1 indicates no Slayer xp is earned from that creature. Format of the list is slayerxp, combat level so those monsters you highlighted are indicated to give no Slayer xp.

I could run the list through a quick bit of code to remove all monsters for which the Slayer xp is -1. Is that what you're recommending?

runelite-client/src/main/resources/slayer_xp.json
MagicfTail: You should probably reply her :p But yeah, if i understand correctly, the monsters that give 0 slayer xp is never used, so there's no reason to have them on the list. And i would edit the scraper to not include them instead of running another program

daviscook477: I'll edit the scraper but I'm going to run this list through a quick different program because running the scraper takes at least an hour and a quick program to delete the extra -1 xp creatures takes way less than that.

daviscook477: And yeah you're understanding correctly. I agree that it does make more sense for the slayer xp json file to only include NPCs that actually can give Slayer xp

daviscook477:
cleaned up proposed code for this plugin by:

1.  refactoring functionality (namely knapsack solver and determining xp drop for an npc) outside of the slayerplugin class into own classes
2. adding some null checks and additional comments
3. tracking monster death animations by name rather than id so it's less brittle

daviscook477:
so this bug fix isn't exactly working. I've tried testing this a few times and continually run into the problem of not realistically being able to determine what tick an npc died on. Basically in order for this to work I have to be able to figure out what xp drops are obtained on each tick **and** what npcs die on that tick too. If that information could be determined then the idea of using a knapsack solver on the xp drops that each npc could give works pretty well (as seen with the unit tests I added for this). Unfortunately in testing, I can't correlate the xp drops and npc deaths accurately. For the most part a slayer xp drop is given the same tick that the npc starts its death animation. Using this gives fairly accurate results but every few minutes of slayer some death animation/xp drop doesn't line up and results in the kill tracker getting confused.

If anyone has a different idea on how to determine which npcs have died when an xp drop comes I'd really appreciate the ideas since I do want to get this pr working.

My current idea on how to make this work is to cut the idea of tracking npc deaths completely from this pr. Instead I would integrate (https://github.com/runelite/runelite/pull/6746) into this pr and then any time an xp drop is given rather than trying to compute the kills from the actual npcs that died I could compute kills by just looking at all the npcs that could be killed on task. So rather than solving bounded knapsack problem on the actual npcs that died, you just do unbounded knapsack problem on the npcs that *could* have died. This would definitely be less accurate than a working version of my current solution but it wouldn't require having to time the xp drops to the npc deaths. The only main issue I see with this is that without tracking npc deaths filtering out xp drops from xp lamps and stuff like that will really mess up the counter. Since the current counter would just treat those drops as 1 kill it wouldn't be too messed up but with this one trying to say figure out how many blue dragons make up a 670xp lamp drop, it would miss by maybe 5 kills or something. This could mostly be avoided by checking to see if the player killed any of the slayer npcs recently but again this gets into the problem of trying to track kills.

Another potential solution would simply be to track all the npc deaths within some rolling window of time and rather than trying to actually pair those deaths with the xp drops simply using those npcs (that are on the list of actually killable npcs for some task) to determine the number of kills. So this way we can see that if we can't detect that say a fire giant was killed the tick a 111xp drop comes in we could see that hey over the last 20/30 ticks lots of fire giants have despawned on a fire giants task. Then we can just solve unbounded knapsack problem using that fire giant npc. This would mostly avoid quest/lamp xp drops because we could see that no npcs had been killed prior to an xp drop coming in. The only main issue I see with this result is that it might be slightly less accurate than actually knowing which npcs died on which tick (which I'm starting to think is impossible) and that for the first kill of a task there wouldn't be any npcs in this death log so we couldn't figure out what npcs might have given the xp drop. This could probably be avoided by simply delaying processing each xp drop for like a few ticks (maybe 6 ticks that's like 3 seconds) so that the despawn log could be filled up with some npcs.

So the basic proposal would be this:
1. keep track of which npcs that have despawned that are on task (by using https://github.com/runelite/runelite/pull/6746) in a window of ~20 ticks.
2. when a slayer xp drop comes in wait 6 ticks and then solve unbounded knapsack problem on the npc types that are in the despawn log (the 20 tick window basically giving like 14 ticks before the xp drop and 6 ticks after the xp drop)
3. whatever the result of that is is the actual amount of kills that caused that xp drop.
4. So if we see a 222xp drop on a fire giants task and the player in fact has been killing fire giants this will say 2 kills were made - similar idea for if they were bursting dust devils

The only trouble with this is I already tried a rolling window of kills and it did not work sufficiently accurately enough. Sometimes kills just wouldn't get registered in time. One way to avoid this would be to use a hybrid of kills and currently loaded npcs. Basically this means we would try to see what xp drops out of those from killed monsters and currently loaded npcs that could make up the slayer xp drop. We can't rely solely on the loaded npcs since the player realistically could kill all the npcs of one type in a multicombat zone with bursting. The only problematic thing for this type of solution is something like desert lizards. The 3 different cb levels of lizards give 15, 25 and 40 xp respectively. A 40 xp drop has no way to differentiate between being a 40 lizard or a 15 and a 25 lizard when we aren't tracking the kills.

PR #6675, opened 2018-11-25
Recharge Prayer Notification

dylanshine:
This implementation includes a notification when a player needs to recharge their prayer points.


forsco:
You can already do this within the Idle Notifier plugin.

https://lambda.sx/g0Z.png

deathbeam:
As @forsco said (also it can be done with chat notifier too)

Issue #5945, opened 2018-10-11
First assigned slayer task is not parsed correctly


Nightfirecat:
When being assigned your first slayer task, the message from the slayer master is different than future tasks, and is not parsed properly. This can be tested by making a new deadman mode account and traveling to Turael to receive a task.

PR #6677, opened 2018-11-25
First Slayer Task

Mikefaheysd:
Add first slayer task dialog #5945


Mikefaheysd:
Example of first slayer task dialog

[https://imgur.com/TSp1A76](url)

PR #6689, opened 2018-11-26
Show error message on macOS when trying to enable GPU plugin

oabdellatif:
It's confusing for users who haven't read the FAQ when the plugin just turns itself off again.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -47,6 +47,8 @@ import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.util.function.Function;
 import javax.inject.Inject;
+import javax.swing.JOptionPane;
+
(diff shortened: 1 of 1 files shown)


deathbeam:
I dont see a reason why there should be error message just for mac when atm we do not support a lot of other systems/hardware

oabdellatif:
Other than if you're running macOS, not having a GPU that supports OpenGL 4.3 or greater would be the other main factor resulting in the GPU plugin not being able to run, at least as far as I can tell. Would either a check for a compatible OpenGL version or just a generic error message be desirable?

devLotto:
I mean, a message for macOS users would be a good start. You could add an OpenGL version check too, later.

melkypie:
MacOS gpu support was added in 0e6f9f9aeca8ebb46b40780dbfeca24a5682f457. Think this should be closed.

makingcomment:
The proposed error message could now be for trying to enable compute shaders.

Nightfirecat:
An error message for compute shaders would be a better approach, but this PR is no longer needed as written.

Issue #6440, opened 2018-11-12
toggle for sidebar switcher image?


Runeliteisgud:
Is there a toggle to turn this off? I prefer not using the custom chrome, but I dislike this little arrow that now appears on the screen.

If there's no toggle, can it be added in for next update?

![image](https://user-images.githubusercontent.com/44961825/48331128-baa65f00-e614-11e8-9b07-deff22111682.png)

deathbeam:
Maybe instead of toggle we can just not show it if your mouse is not in corner

Runeliteisgud:
> Maybe instead of toggle we can just not show it if your mouse is not in corner

that would work too I guess, although it really is kinda pointless when you can just hide it by clicking any icon on the sidebar anyway

Berry-95:
Think I've got a solution to this using @deathbeam's auto-hide when mouse is not in the corner idea, just doing some final testing and will probably submit a pr tomorrow.

PR #6694, opened 2018-11-26
Add auto-hide sidebar toggle button…

Berry-95:
…when mouse is not close to right-side edge of client window.
Potential fix for: #6440

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -118,6 +118,8 @@ public class ClientUI
 	private final Provider<ClientThread> clientThreadProvider;
 	private final CardLayout cardLayout = new CardLayout();
 	private final Rectangle sidebarButtonPosition = new Rectangle();
+	private final Rectangle sidebarButtonRange = new Rectangle();
+	private boolean mouseInSidebarButtonRange = false;
 	private boolean withTitleBar;
 	private BufferedImage sidebarOpenIcon;
 	private BufferedImage sidebarClosedIcon;
@@ -368,6 +370,32 @@ public class ClientUI
 						mouseEvent.consume();
 					}
 
+					return mouseEvent;
+				}
+
+				@Override
+				public MouseEvent mouseMoved(MouseEvent mouseEvent)
+				{
+					if (!(ClientUI.this.client instanceof Client) || withTitleBar)
+					{
+						return mouseEvent;
+					}
+
+					if (sidebarButtonRange.contains(mouseEvent.getPoint()))
+					{
+						mouseInSidebarButtonRange = true;
+					}
+					else if (mouseInSidebarButtonRange)
+					{
+						mouseInSidebarButtonRange = false;
+
+						final Client client = (Client) ClientUI.this.client;
+						//Repaint of client canvas required to remove image artifact in black bars when on login screen with stretched enabled.
+						if (client.getGameState() == GameState.LOGIN_SCREEN)
+						{
+							client.getCanvas().repaint();
+						}
+					}
 					return mouseEvent;
 				}
 			};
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -370,6 +372,32 @@ public MouseEvent mousePressed(MouseEvent mouseEvent)
...
+					}
+					else if (mouseInSidebarButtonRange)
+					{
+						mouseInSidebarButtonRange = false;
+
+						final Client client = (Client) ClientUI.this.client;
+						//Repaint of client canvas required to remove image artifact in black bars when on login screen with stretched enabled.
+						if (client.getGameState() == GameState.LOGIN_SCREEN)
+						{
+							client.getCanvas().repaint();
```
deathbeam: Is this really necessary? I think this is general stretched mode bug and we should not be solving it case-per-case

Berry-95: If I removed it the sidebar toggle button won't disappear on the login screen when the mouse exits the "area", but I guess that's not too big an issue as the main desired effect is hiding it when in game, so I can remove this if you want.

deathbeam: Yeah, that would greatly simplify the code (as the entire mouse moved and boolean would not be necessary)

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -118,6 +118,7 @@
 	private final Provider<ClientThread> clientThreadProvider;
 	private final CardLayout cardLayout = new CardLayout();
 	private final Rectangle sidebarButtonPosition = new Rectangle();
+	private final Rectangle sidebarButtonRange = new Rectangle();
```
deathbeam: This field is no longer necessary too

Berry-95: That depends on whether we want to check if the mouse is in the area of the button (pretty small if you don't know where it is) or if the mouse is in a larger "sidebarButtonRange" area, which is currently set to the right-side of the client, which is easily readable and configurable as a Rectangle object...

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -630,11 +630,17 @@ public void paintOverlays(final Graphics2D graphics)
...
+		Point mousePosition = client.getMouseCanvasPosition();
+		if (sidebarButtonRange.contains(mousePosition.getX(), mousePosition.getY()))
+		{
+			graphics.drawImage(image, x, y, null);
+		}
 
 		// Update button dimensions
 		sidebarButtonPosition.setBounds(x, y, image.getWidth(), image.getHeight());
-	}
+}
```
deathbeam: Fix indentation here

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -630,11 +630,17 @@ public void paintOverlays(final Graphics2D graphics)
 			: 5;
 
 		final BufferedImage image = sidebarOpen ? sidebarClosedIcon : sidebarOpenIcon;
-		graphics.drawImage(image, x, y, null);
+
+		Rectangle sidebarButtonRange = new Rectangle(x - 15, 0, image.getWidth() + 5 + 15, client.getRealDimensions().height);
```
deathbeam: Mark this final

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -630,11 +630,17 @@ public void paintOverlays(final Graphics2D graphics)
...
+		Point mousePosition = client.getMouseCanvasPosition();
```
deathbeam: same here

PR #6696, opened 2018-11-26
Properly parse stat change value in status bars

deathbeam:
This prevents all kinds of unsupported formats and number format
exceptions.

Closes #6589


Nightfirecat:
For the record, I feel implementing #6699 is the most correct way of correcting this issue, but this will do for now.

Issue #6672, opened 2018-11-25
Pets have an NPC dot on the minimap in barrows tunnels.


SomeZer0:
**Describe the bug**
Pets don't normally have a dot on the minimap.
But they do while in barrows tunnels on the minimap plugin

**To Reproduce**
1. have pet
2. go into barrows tunnel

PR #6698, opened 2018-11-26
Remove pet from minimap in barrows

Berry-95:
Fixes: #6672
Also adds a helper function in NPC called isPet, which returns true if an NPC is a pet.
I could be modify this to simply hide NPCs with undefined combat level if my NPC isPet method is not desirable.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/NPC.java
+++ b/runelite-api/src/main/java/net/runelite/api/NPC.java
@@ -75,4 +75,6 @@ public interface NPC extends Actor
 	 * @return
 	 */
 	boolean isDead();
+
+	boolean isPet();
 }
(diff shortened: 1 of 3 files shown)


Adam-:
There is a flag for whether or not npcs show on minimap, maybe we could use that instead?

Berry-95:
> There is a flag for whether or not npcs show on minimap, maybe we could use that instead?

I was thinking about that, I'd have to test that they're not all set to false in barrows though, as you're not supposed to see them on the vanilla minimap.

Adam-:
It is already exported as `isMinimapVisable` so its worth checking

Berry-95:
> It is already exported as `isMinimapVisable` so its worth checking

Not got membership at the moment, so might be a while before I can confirm it myself.

Berry-95:
I'm going to do some testing now that I've got some time, however my only pet is a cat...
Thinking of fiddling with isMinimapVisible or confirming all NPCs in Barrows have a combat level, so that a pet (with no combat level can be hidden).
Any other potential solution suggests/ideas worth testing feel free to let me know.
I'll also be investigating #6918 further.

PR #2204, opened 2018-05-01
Add double tap hotkey(ALT) toggles hide all ground items.

Berry-95:
Potential solution for: "Add toggle for hiding ground items #1781"

Issue #1781, opened 2018-04-22
Add toggle for hiding ground items


deathbeam:
Add new toggle for quickly enabling/disabling ground items overlays. As requested on Discord.

PR #6712, opened 2018-11-27
 Add ability to hide all ground items on double tap of hotkey(ALT)

Berry-95:
Add listener for double tap of hotkey
Add ability to show all ground items again with single tap

Re-implementation of my old attempt that became outdated: #2204
Fix for: #1781


raiyni:
Should this really be on double-tap and not a combination key? It's very easy to double tap alt and I think plenty of users will do that and not realize it.

deathbeam:
I like double that, it usually achieves similar effect in other apps and is simple to use

raiyni:
I also use the alt toggle a lot to see items that are hidden.

Berry-95:
> I also use the alt toggle a lot to see items that are hidden.

The timing I have setup is so that a key slip shouldn't trigger as a double tap and your double tap has to be pretty closely time (intentional) to hide the items and if you hide it by accident, one more tap of alt brings it all back

raiyni:
I triggered it easily enough.

Nightfirecat:
The double-tap duration should either be (a) made into a configuration, which would help people who trigger this too often, or (b) made into a constant and referenced so it's not just a magic number.

deathbeam:
Instead of making the delay configurable I would simply add toggle to disable it

Nightfirecat:
In addition to configurable delay imo

deathbeam:
If we want to keep delay, then no toggle (maybe 0 = disable as we do on other places0

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -313,4 +312,16 @@ default int insaneValuePrice()
 	{
 		return 10000000;
 	}
+
+	@ConfigItem(
+		keyName = "doubleTapDelay",
+		name = "Delay for double-tap ALT to hide",
+		description = "Decrease this number if you accidentally hide ground items often.",
```
Nightfirecat: This should be updated to indicate a value of `0` will disable this behavior.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemInputListener.java
```diff
@@ -50,7 +57,22 @@ public void keyPressed(KeyEvent e)
...
 		if (e.getKeyCode() == HOTKEY)
 		{
-			plugin.setHotKeyPressed(true);
+			if (plugin.isHideAll())
+			{
+				plugin.setHideAll(false);
+				plugin.setHotKeyPressed(true);
+				lastPress = null;
+			}
+			else if (lastPress != null && !plugin.isHotKeyPressed() && config.doubleTapDelay() > 0 && Duration.between(lastPress, Instant.now()).compareTo(Duration.ofMillis(config.doubleTapDelay())) < 0)
```
Nightfirecat: Alternatively, the explicit check for `> 0` could be removed, since there's no way to register two presses in that time.

Berry-95: > Alternatively, the explicit check for `> 0` could be removed, since there's no way to register two presses in that time.

Yeah not sure whether it's necessary, but thought it might be good practice to explicitly check.

Berry-95:
Resolve merge conflict in groundItemConfig, Any further change requests?

deathbeam:
why does this also contain configurable ground items hotkey

Berry-95:
> why does this also contain configurable ground items hotkey

I thought it could be a useful addition, but if it's undesired I can remove the configurable hotkey.

deathbeam:
Yeah, remove that

Berry-95:
@deathbeam Removed configurable hotkey 👍

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -302,7 +302,6 @@ default Color insaneValueColor()
 	{
 		return Color.decode("#FF66B2");
 	}
-
```
Adam-: Please stop changing formatting in unrelated parts of files in your prs.

Berry-95: Sorry, that one must of snuck through when rebasing.

PR #6717, opened 2018-11-27
xptracker: Add support for target/goal levels

while-loop:
This PR allows users to set a custom target level from 1 - 126 in XP Tracker.
TTL and xp to level is also reflected in this PR.

screen shot 2018-11-27 at 5 16 09 pm


Nightfirecat:
Goals are already supported via the ingame goal settings.

deathbeam:
We already have this feature: https://github.com/runelite/runelite/wiki/XP-Tracker#setting-level-goals

while-loop:
Thanks for the quick responses. That's one feature I did not know existed in osrs. lol

PR #6724, opened 2018-11-28
Remove the fight cave plugin

MagicfTail:
Completing the fight caves is supposed to be the first real challenge you do, and defeating Jad for the first time, an achievement. This plugin completely trivialises the fight against Jad, going beyond what 3rd party clients should be allowed to do (IMO).


robinwithes:
Completely agree, third party clients should offer QOL features. This went way across the line.

james-munson:
no

yoghurtis:
yes

az0977776:
There are many other plugins that makes many features of the game drastically easier. I think it should be left in and we should let players choose if they want to ruin their sense of accomplishment.

robinwithes:
> 
> 
> There are many other plugins that makes many features of the game drastically easier. I think it should be left in and we should let players choose if they want to ruin their sense of accomplishment.

Since other things like zulrah helpers and stuff is blacklisted, i dont understand why this is not.
And the point you made ("we should let players choose if they want to ruin their sense of accomplishment") is really not a attitude i think the runelite team should be going for. You might as well defend botting clients with this as well since they are just ruining their sense of accomplishment.

Kberrio:
If you **can** do it without the helper, good for you! You should feel proud of your achievement.
If you **can't**, good for you! Use the helper to guide you and teach you how to do it in the future.

That's just how I feel about it.

Psyda:
I'm not sure I'd agree fully. 
I found it to be more distracting when using it. 

I was coming back to rs after years away and then doing Jad with runelite. 

It served as a bit of a reminder on which sounds to pray what against, but relying on it caused me to die. Knowing what to pray to stay alive isn't enough to beat Jad. Its a pretty big handhold but I don't think it's anything too over the top.

It doesn't quite fit with what Runelite has been aiming to be lately but I'd say its far from the most concerning plugin. The Worldhopper plugin with PVP escape quickhop hotkeys is quite a bit more OP than a prayer reminder when in sight of Jad.

EchoZeroTwo:
Why don't you stop trying to control everything and let people decide for themselves

EchoZeroTwo:
Selfish stupid PR @MagicfTail

MagicfTail:
I do not like plugins whose sole purpose is making a specific boss/NPC easier to kill. That is in no way QOL, but making the gamer easier for the sake of making it easier. The fight cave plugin is especially bad since it targets an activity that is only a challenge for new players.

Furthermore, I don't believe having something that clearly shows what to pray against the next attack doesn't make it easier to use the correct prayer, and even then, knowing that the plugin will show you will probably remove some nerves.

Btw, I don't think pointing out that the client has way more op plugins is a good way of showing another op plugin should stay.

EchoZeroTwo:
ok disable jad plugin on your client then you brainlet. Unless I'm mistaken you don't speak for the whole Runelite community

robinwithes:
> 
> 
> ok disable jad plugin on your client then you brainlet. Unless I'm mistaken you don't speak for the whole Runelite community

No need to get insulting, this is a discussion for a reason.

WillThomas:
It's difficult to remove existing plugins for being too powerful and we will never have consensus on these topics unless Jagex explicitly enforces their removal.

Jagex have proven before they do not like some of the plugins that make pvm easier and have enforced their removal before citing game integrity as the major reason, additionally they have requested that 3rd party clients do not create new plugins for parts of the game such as Theatre of Blood and The Inferno[1]. The Jad plugin falls into a similar category, it is in _no_ way a QOL plugin and is purely designed to aid players in defeating a difficult boss in a way that the official client does not; ergo it is breaking the games integrity.

It is upto the Runelite devs how to proceed with this plugin removal request and future ones where Jagex does not have an official stance. It should not matter whether other clients have an equivalent plugin or not for this discussion, only the plugins features. Does Runelite want to keep toeing the line and getting overpowered features removed later or stop these powerful plugins from making it into the main client in the first place?

Ideally there would be a way to quickly get approval from the Oldschool team as to whether a plugin is acceptable before releasing but it is unlikely this will ever happen. Possibly when the Oldschool team has expanded and there are a few dedicated game integrity/balance Jmods this may be an option.

[1] https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

Adam-:
I think that this is too far and the fact that it is here in the first place was an error in judgement on my part.

I could ask the oldschool team about it, but you have to pick your battles with them, and I would rather pick something more important than this.

In the past we have had three separate instances of being directed to not impact bossing, specifically Zulrah, (Vorkath?) AOE's, and preemptively ToB. All of this direction was completely one-sided and I have to try and interpret the direction I have the best I can. For example we still don't know why Konduit is allowed to show Raids AOE - only that what we had was "not allowed". I would guess it is because theirs is raids-specific, but I really *don't know*.

Given this, my guess is they would want us to remove it. And regardless, I haven't liked it for awhile but did not want to start heated discussions over it. We also do not have to keep it to keep parity with OSB/Konduit since neither have it. So, I think that this should be merged.

Jacoblairm:
Good decision, always thought this was an unnecessary plugin.

TheIrishCoder:
Pretty upset about this - since there was no official complaint from Jagex, I'm much more in favour of allowing players who want to use it, to use it. I'm pretty awful at prayer flicking and have tried a few times using the plugin and still found it extremely hard. I will still feel incredible accomplishment even using the plugin, and believe this decision should be reverted. Is there any chance of that or is this a final decision?

BattousaiTMS:
Please bring this back, this plugin was really the only thing that gave me hope I would ever actually get the fire cape. I have tried 4 times already with the plugin and have failed every time but when I heard this was being removed I figured there is no point in trying anymore. I am pretty casual at this game and I am bad a play flicking but this plugin was just the confidence boost i needed to keep trying.

troypressley1:
This will be the thing that keeps OSBuddy premium alive now.

TheIrishCoder:
Hey Pressley, is this feature available on osbuddy premium?

robinwithes:
No its not..

Soxs:
What a shame :-)

AlligatorFarts:
I'm sorry, but the Fight Caves are not very challenging, and if you keep failing, just keep trying it and "get gud" cause that plugin will destroy any sense of accomplishment you would've had by beating it the right way. You need to learn jad if you even want to consider doing any late game PvM

troypressley1:
They're probably going to add it now as a premium feature. I got my Fire Cape with the plug-in. Doesn't bother my sense of accomplishment. If it bothers you, then you should've just turned it off.

Soxs:
Good job RuneLite developers! further enabling those who abuse their knowledge of programming, now every dick dom and harry will be selling versions of runelite with the jad script enabled. :-)

Issue #6619, opened 2018-11-21
Show level up message/widget when getting virtual levels


MagicfTail:
Also, have the screenshot plugin take a screenshot when the level occurs

PR #6726, opened 2018-11-28
Chat message on virtual level up + Level up event

Haashi:
First look at #6619 feature-request.

First commit is about adding an event on level up (any level, even real level) and xptracker posting the event.

Second one is adding the message in the chatbox using the already existing virtual level plugin.

Any suggestions is welcome.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/events/SkillLeveledUp.java
+++ b/runelite-api/src/main/java/net/runelite/api/events/SkillLeveledUp.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2018, Haashi <https://github.com/Haashi>
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


Haashi:
any idea what can post the event then ? because the leveling up is quite related to xp tracking, don't really know how we can trigger it without tracking XP.
Could track XP inside a non plugin class, but that would be duplicating code/task right ?

deathbeam:
Well it is not really related to XP tracking. You can either copy-paste the code, or just look how it works, and use parts of it (we already do that at bunch of places, it is pretty simple)

Haashi:
Updated. Did it my own way to check after every ExperienceChanged event.

melkypie:
This has been added in runelite/plugin-hub#171

PR #6730, opened 2018-11-28
Add ability to hide status bars when out of combat

jochim-vd:
Hide the status bars when the specified amount of seconds have passed since you last got hit.


raiyni:
Added in https://github.com/runelite/runelite/commit/9b5c87e9e06211b27895cba135612fadd39c37eb

PR #6465, opened 2018-11-14
Add "support" for npcids to the slayer plugin and add dwarf guards to the dwarves task.

15987632:
closes #6458
This probably wont get merged because its pretty hacky and I did it in a stupid way. The problem was that the dwarf guard has the name "Guard" and many non dwarf guards share the same name and the way the slayer plugin currently handles the slayer monsters that get highlighted for a task it wasn't possible to distinguish between the types of guards.

edit: It's not stupid anymore

Issue #6458, opened 2018-11-13
Highlight dwarven guards for slayer dwarf task


englishlegion:
https://i.gyazo.com/143e5904eb5126b763dc3b16677a1cc6.png

something minor but, guards aren't selected while "highlighting targets" on the slayer tab, with dwarfs as the task

Issue #6285, opened 2018-10-31
"Pirate" highlighted during Rat slayer task


Nickenator8:
**Describe the bug**
When doing a Rat slayer task on my obby mauler i noticed that a nearby Pirate was highlighted

**To Reproduce**
1. Get Rat slayer task
2. Go to a Pirate

**Expected behavior**
A pirate is not a rat

**Screenshots**
https://i.gyazo.com/52348c4088e4de6aa9994a551cdaadd3.png

**Environment (please complete the following information):**
 - OS: Windows
 - RuneLite version: 1.4.23
 - Launcher version: 1.6.0

KasparRosin:
Could the rat have been under him and that's why he was highlighted?

Hydrox6:
@KasparRosin this happens because the Slayer Plugin uses strings to find what to kill, as opposed to NPC IDs. The check that's used is a simple `String#contains`, and since `Pirate` contains `Rat`, it highlights them.

Nickenator8:
@KasparRosin exactly what @Hydrox6 said. it has it's own highlight box (idk what you call it)

Thought it might just be a simple fix so I reported it. It's not a big deal at all really

Issue #1800, opened 2018-04-22
Certain Slayer Targets Not Being Highlighted


CharlesLove:
All monsters featured below do not automatically highlight when assigned on slayer tasks.

Flesh Crawlers,
Small Lizards (Desert Lizards alternative), and
Baby Dragons (Have only seen blue, but this likely affects all color variants)

Kamielvf:
The issue with baby dragons is that they're all named "baby dragon", and I think the plugin is matching names.

CharlesLove:
For baby dragons, maybe levels could also be taken into consideration to differentiate them.

rherms:
Hey I'm a new dev to RuneLite and this fix sounds relatively easy so I'm going to give it a shot!

rherms:
Baby dragons are all the same level unfortunately.

titegtnodI:
Sulphur Lizards also don't highlight. They also don't decrease the slayer counter (I assume it's the same issue, so I'm only mentioning it here).

Alexsuperfly:
with the update today jagex have said `Baby dragons now have their colour in their name in order to help colourblind players determine which is which.`  So solving the baby dragon portion of this issue is now possible.

Narozzy:
Is this still an issue, looking through the `Task.java` and the `SlayerPlugin.java` code, seems like we are just checking if the name of a monster contains the target name, which are not set for any of the dragons as an example.

Nightfirecat:
Yes, this is still an issue for the specific cases outlined in the original post.

PR #6741, opened 2018-11-29
[WIP] Change slayer tasks to use NpcIDs instead of strings

Berry-95:
Fixes a lot of highlighting issues:
#6458, #6285, #1800 and can easily be expanded if any task alternatives are missing.
Continues on from:
#6465

Currently making sure I've got as many alternatives included as possible.


ShaunDreclin:
I feel like this would be much more annoying to maintain than doing it by name

Issue #3801, opened 2018-06-14
Ground Items Plugin - Hide items that don't belong to the player


Gubzs:
**Is your feature request related to a problem? Please describe.**
When playing ironman mode, it is a hassle to sort through loot on the ground and determine what you can and cannot pick up. This is especially obnoxious during slayer at areas like Kurasks where many players will leave valuable herbs on the ground that they cannot carry.

**Describe the solution you'd like**
I would like a toggled option within the ground items plugin to hide text for items that do not belong to the player. I would also like to remove them from the left click option if possible, so that the user can walk over other player's items without using the right click menu.

**Describe alternatives you've considered**
Just hiding the item name or "greying it out" as the "hide items below x value" option does would be great by itself if that could be done.

**Additional context**
The game knows who a ground item belongs to, if it's dropped by them or from a monster - hence the delay of dropped items appearing for others, and ironman's inability to loot other player's things. I'm hoping the client has that information as well, this would be a huge QoL feature for ironman players. Especially when doing slayer-like activities in crowded areas.

An alternate use case might include any account type turning this on to hide other player's dropped skilling items such as players mass dropping fish or logs, while still keeping those items visible on the ground items list for other purposes. Nonetheless, most ironman players would want to keep this option on at all times.

BryanHaley:
For what it's worth, the game is aware of what items are owned by whom, but I don't think the _client_ is. I can't say with certainty as I've only skimmed the source code, but I believe the client simply sends a request to pick something up and the server tells them whether or not they can, which makes it difficult or impossible to implement a feature like this.

Karl456:
> For what it's worth, the game is aware of what items are owned by whom, but I don't think the _client_ is. I can't say with certainty as I've only skimmed the source code, but I believe the client simply sends a request to pick something up and the server tells them whether or not they can, which makes it difficult or impossible to implement a feature like this.

Could you not do this in a different way, the loot tracker plugin knows which loot you have earned from kills, etc. So is it not possible to have **everything** hidden, then only show things which the loot tracker would have tracked? I'm not sure of the inner workings of the client, but is this possible?

daviscook477:
If I recall correctly the loot tracker isn't perfectly accurate and it would kind of suck to not get a cool drop because some loot hiding plugin malfunctioned.

Karl456:
[...]

PR #6756, opened 2018-11-30
Added Feature for Hiding Other Player's Loot

BryanHaley:
This is an implementation of the feature request at #3801 

**What does it do?**
This adds an option to the Ground Items plugin that prevents the overlay from being shown on loot that drops from NPCs killed by other players.
screen shot 2018-11-29 at 10 01 00 pm

Here it is in action:
screen shot 2018-11-29 at 10 40 58 pm

**What is the purpose of this feature?**
This is a quality-of-life feature for players in Ironman mode. It makes it much easier to differentiate from NPC drops that can and cannot be picked up.

**How does it work?**
The Ground Items plugin now subscribes to onNpcLootReceived and onPlayerLootReceived events, and keeps track of items dropped by NPCs killed by the player, and the location at which they drop. Then, when it's time for the overlay to be rendered, the item that is being checked for an overlay is compared to the list of loot items, and skipped if it is not present in that list.

Items are removed from the loot item list when they despawn. To avoid memory leaks, the loot item list is also cleared when the game enters the "LOADING" state, as this implies the player is moving away from the area (despawn events are not sent once the player has left the area the item is in, afaik).

**Are there any downsides?**
To my knowledge, there is no way to track specific items on the ground; rather, the best I can do is to track the item type and location. For this reason, this option cannot guarentee 100% accuracy for a variety of reasons, including but not limited to: non-pickup-able items dropping in the same tile as pickup-able items of the same type, picking up and re-dropping items, et cetera.

In addition, since the overlay checks the loot item list every frame, there can be a significant performance penalty while the option is enabled. My MBP 2015 i5 handles it without any noticeable drop in framerate, but results may vary on weaker machines.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/overlay/OverlayIndex.java
+++ b/runelite-api/src/main/java/net/runelite/api/overlay/OverlayIndex.java
@@ -52,6 +52,10 @@ public class OverlayIndex
 		{
 			log.warn("unable to load overlay index", ex);
 		}
+		catch (NullPointerException ex)
+		{
+			log.warn("unable to load overlays", ex);
+		}
 	}
 
 	public static boolean hasOverlay(int indexId, int archiveId)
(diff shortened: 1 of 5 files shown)


BryanHaley:
Fixing style check violations.

Nightfirecat:
FYI in the future you can add commits or force-push to update a PR branch instead of opening a new PR. :)

raiyni:
If you use the style xml file from the wiki and configure your imports correctly, you can just use the intellij code formatter to make your life easier.

Issue #6558, opened 2018-11-18
Add option to screenshot loot received over a given amount.


MrH4mm3r:
Include a box in the screenshot settings where you can enable screenshotting loot over a given amount of gp. This could also be added for untradeables.

Adam-:
Added in #6762

PR #6762, opened 2018-11-30
Screenshot on valuable/untradeable item drop

tanlines:
#6558

Uses ingame drop notifier to set the loot value and whether to include untradeables.
Triggers on chat message.

Tested on the following strings. Will not work if there's an item that contains '<' or '>' in their name.
`public static void main(String[] args) {
        testPattern("<col=ef1020>Valuable drop: 6 x Bronze arrow (42 coins)`</col>`");
        testPattern("<col=ef1020>Valuable drop: 2222 x Ranging potion(3) (987 coins)`</col>`");
        testPattern("<col=ef1020>Untradeable drop: Rusty sword");
        testPattern("<col=ef1020>Untradeable drop: 111111 x Ranging potion(3)");
        testPattern("<col=ef1020>Untradeable drop: 3 x Bronze arrow");
    }

    public static void testPattern(String message) {

        Pattern VALUABLE_DROP_PATTERN = Pattern.compile("<col=ef1020>(Valuable|Untradeable) drop: ([^<>]+)(?:</col>)?");
        Matcher m = VALUABLE_DROP_PATTERN.matcher(message);
        if (m.matches())
        {
            String dropType = m.group(1);
            String dropName = m.group(2);
            String fileName = dropType + " drop " + dropName + " " + LocalDate.now();
            System.out.println(fileName);
        }
    }
`


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -119,6 +119,7 @@
 	private static final Pattern NUMBER_PATTERN = Pattern.compile("([0-9]+)");
 	private static final Pattern LEVEL_UP_PATTERN = Pattern.compile(".*Your ([a-zA-Z]+) (?:level is|are)? now (\\d+)\\.");
 	private static final Pattern BOSSKILL_MESSAGE_PATTERN = Pattern.compile("Your (.+) kill count is: <col=ff0000>(\\d+)</col>.");
+	private static final Pattern VALUABLE_DROP_PATTERN = Pattern.compile("<col=ef1020>(Valuable|Untradeable) drop: ([^<>]+)(?:</col>)?");
```
Nightfirecat: What's the reasoning behind this? Can't this be `<col=ef1020>(Valuable|Untradeable) drop: (.+)(?:</col>)?`

raiyni: I would drop the color specific matching, too

tanlines: > What's the reasoning behind this? Can't this be `<col=ef1020>(Valuable|Untradeable) drop: (.+)(?:</col>)?`

The Valuable drop text will match the </col> text then. Valuable drop has </col>, Untradeable drop doesn't.

Nightfirecat: You should add tests for this, by the way.

Nightfirecat: For this plugin, https://github.com/runelite/runelite/blob/master/runelite-client/src/test/java/net/runelite/client/plugins/screenshot/ScreenshotPluginTest.java

tanlines: Thanks, I looked at some other PRs and didn't see tests, didn't realise not all plugins have tests.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -366,6 +370,18 @@ public void onChatMessage(ChatMessage event)
...
 		}
+
+		if (config.screenshotValuableDrop())
+		{
+			Matcher m = VALUABLE_DROP_PATTERN.matcher(chatMessage);
+			if (m.matches())
+			{
+				valuableDropType = m.group(1);
+				valuableDropName = m.group(2);
+				String fileName = valuableDropType + " drop " + valuableDropName + " " + LocalDate.now();
```
Adam-: Use `format(new Date())` like everything else

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -655,6 +671,18 @@ String getClueType()
 		return clueType;
 	}
 
+	@VisibleForTesting
+	String getValuableDropType()
+	{
+		return valuableDropType;
```
Adam-: You can just test that a frame was requested with `verify(drawManager).requestNextFrameListener(Matchers.any(Consumer.class));`. If we really want to start testing filenames then we should have the screenshot-taker be put in a different class which we can mock and then use `verify`.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
Nightfirecat: This pattern still seems like it is overcomplicated. You do not need to match `[^<>]+` since the pattern matcher will attempt to match the trailing `</col>` before the second group is captured. I would recommend trying to run the tests with the pattern I wrote above.

tanlines: I did try your method, was the first pattern I tried but it doesn't seem to match the trailing part first.

Nightfirecat: https://regexr.com/44bbr ?

Nightfirecat: Ah, I see. Matching to the end of the string and adding a lazy modifier to the `(.+)` group fixes that issue: https://regexr.com/44bcp

tanlines: In any case, I think it's still fine since `<>` are invalid file name characters. Right now there doesn't seem to be any file name checking being done in any of the tests.

Nightfirecat: I suppose that's fair. :+1:

PR #6772, opened 2018-12-01
Add XP per action to XP Tracker

### LINKED ISSUE #6727: XP per action.
I would like if I could find out my xp per action.

I would like to be able to hover over the xp bar in the XP Tracker tab and be able to see my xp per action. Make it toggleable if you want too.

Manually opening up the calculator and dividing my xp over how many actions I did.

**Additional context**
For example, say you're Barbarian Fishing, you're going to catch more than one type of fish which is going to give different XP amounts. It'd be cool to see how many feathers you'd need to get to the next level (roughly) by simply looking at the xp per action.

Atynine:
Closes #6727 
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
@@ -61,6 +61,7 @@ class XpInfoBox extends JPanel
 	private static final String HTML_TOOL_TIP_TEMPLATE =
 		"<html>%s %s done<br/>"
 			+ "%s %s/hr<br/>"
+			+ "%s XP/%s<br/>"
 			+ "%s till goal lvl</html>";
 	private static final String HTML_LABEL_TEMPLATE =
 		"<html><body style='color:%s'>%s<span style='color:white'>%s</span></body></html>";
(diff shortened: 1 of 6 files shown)


daviscook477:
How does it work with the kills tracking for the combat stats?

raiyni:
how is this information even useful

Atynine:
Works the same with combat skills by tracking xp/kill instead of xp/action
image

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -170,6 +170,13 @@ int getXpHr()
 		return toHourly(xpGained);
 	}
 
+	int getXpAction()
+	{
+		int actions = getXpAction(actionType).getActions();
+		if (actions == 0) return 0;
```
Adam-: Fix formatting

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -193,11 +194,11 @@ private void rebuildAsync(boolean updated, boolean skillPaused, XpSnapshotSingle
 			}
 
 			paused = skillPaused;
-
```
Adam-: Probably don't remove this line since it separates the state updating from the label building

xdflames:
Can this say "XP Last Action" instead of "XP/Action"?

Would make more sense this way and accurately describe what the value is, since for some skills you can be training in a way where you're getting differing exp per action. Agility is an easy example, or for Fishing where some fishing spots have a chance for multiple fish.

xdflames:
Then shouldn't the wording be Average XP / Action?

Hoten:
> Then shouldn't the wording be Average XP / Action?

+1 to this. Or even `XP / Action (Average)`. @Atynine what do you think?

PR #6773, opened 2018-12-01
Adding feature to show the Looting Bag contents on your screen.

ErinTales:
This plugin works in a similar way to InventoryViewer, by bringing up the contents of your Looting Bag in an identical interface. Updates whenever you use the "Check" option on your bag.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/InventoryID.java
+++ b/runelite-api/src/main/java/net/runelite/api/InventoryID.java
@@ -56,7 +56,12 @@ public enum InventoryID
 	/**
 	 * Theater of Blood reward chest inventory (Raids 2)
 	 */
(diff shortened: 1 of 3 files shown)


raiyni:
This will get complete wrong as soon as you deposit something or auto pick up something.

ErinTales:
That's true, but it's still useful to have an approximation of your current loot if you're clan pking, and it's very useful for UIM to use.

Nightfirecat:
Most pkers nowadays will open the bag and pick stuff up rather than using items on looting bags, so I don't feel this would be helpful enough, given that it can't support that.

Adam-:
I think there are way too many corner and not-so-corner cases for this to possibly work at a level the average user would consider satisfactory.

PR #6774, opened 2018-12-01
don't check chatbox status on last tick for fairyring search

### LINKED ISSUE #6759: configuring the fairy ring same tick as teleporting doesn't remove the input box
https://i.imgur.com/Sed6Dm8.gifv

daviscook477:
fixes #6759

fairyring plugin was only closing the chatbox if the fairyring menu was gone and the chatbox had been open on the last tick (which wasn't the case when opening the chatbox and then teleporting on the same tick as in #6759) so now we only check if the fairyring menu is gone and the chatbox is currently open.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
@@ -91,7 +91,6 @@ public class FairyRingPlugin extends Plugin
 
 	private ChatboxTextInput searchInput = null;
 	private Widget searchBtn;
-	private boolean chatboxOpenLastTick = false;
 	private Collection<CodeWidgets> codes = null;
 
 	@Data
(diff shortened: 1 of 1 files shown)


daviscook477:
@Abextm was there any reason why in your original commit there was a need to check for the chatbox having been open on the last tick (not just the current tick)? In testing this code that only checks the current tick it worked when:

1. configuring the fairy ring, waiting for the search to popup, waiting a while, and then exiting out (by walking away)
2. configuring the fairy ring, waiting for the search to popup, then immediately exiting out (by walking away)
3. configuring the fairy ring, and then teleporting away with last-destination immediately once the search pops up (the bug from #6759)
4. configuring the fairy ring, waiting for the search to popup, waiting a while, and then exiting out (by clicking the x)  - in this case the ui immediately closes and the search window only closes on the following tick (but this behavior exists even within the current implementation)

Abextm:
I'm about 70% sure it was so it wouldn't close other chatbox inputs with the old system. I think that the check is no longer needed since I ported it to the ChatboxPanelManager.

daviscook477:
on_fairyring_fix_branch_no_issues_with_closing_other_textentries

Abextm:
Yeah, that check was only needed with `net.runelite.client.game.ChatboxInputManager`. Now it uses `net.runelite.client.game.chatbox.ChatboxPanelManager` which keeps track of which input is open.

PR #6775, opened 2018-12-01
 slayer plugin: show initial assignment amount in infobox tooltip

Adam-:


daviscook477:
getting the assignment properly has the remaining kill count set and the initial kill count set for the tooltip
getting_assignment_start
after a kill is made the initial kill count remains unchanged but the remaining kill count is decreased as expected
after_a_kill
initial kill count is saved so it persists across restarts
after_a_restart

I think it works

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
```diff
@@ -129,6 +129,23 @@ default int amount()
...
+		keyName = "initialAmount",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default int initialAmount()
+	{
+		return -1;
+	}
+	@ConfigItem(
```
deathbeam: Newline above

PR #6787, opened 2018-12-02
Add blast mine collection exp

jky-dev:
Adds an overlay that displays the total exp that would be recieved from the operator at blast mine. There is also an option to display the exp if collecting with full prospectors outfit on.

bm exp counter

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
@@ -44,11 +44,11 @@ class BlastMineCollectExp extends Overlay
 	private final BlastMinePluginConfig config;
 	private final PanelComponent panelComponent = new PanelComponent();
 
-	public static final int COAL_EXP = 30;
-	public static final int GOLD_EXP = 60;
-	public static final int MITHRIL_EXP = 110;
-	public static final int ADAMANT_EXP = 170;
-	public static final int RUNITE_EXP = 240;
+	private static final int COAL_EXP = 30;
+	private static final int GOLD_EXP = 60;
+	private static final int MITHRIL_EXP = 110;
+	private static final int ADAMANT_EXP = 170;
+	private static final int RUNITE_EXP = 240;
 
 	@Inject
 	public BlastMineCollectExp(Client client, BlastMinePluginConfig config)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2017, Jacky Liang <liangj97@gmail.com>
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


Alexsuperfly:
why not auto detect if they are wearing full prospector or not and just display that value, in order to not have the bloat with a toggle

jkybtw:
@Alexsuperfly the design decision was made because usually when doing blast mine, you're equipping full graceful, and only equipped prospectors when collecting the ore at the end

runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
```diff
@@ -0,0 +1,98 @@
...
+import net.runelite.client.ui.overlay.components.LineComponent;
+import net.runelite.client.ui.overlay.components.PanelComponent;
+
+class BlastMineCollectExp extends Overlay
+{
+	private final Client client;
+	private final BlastMinePluginConfig config;
+	private final PanelComponent panelComponent = new PanelComponent();
+
+	public static final int COAL_EXP = 30;
```
Adam-: These should be private

runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMinePluginConfig.java
```diff
@@ -98,4 +98,26 @@ default Color getWarningColor()
...
+		return false;
+	}
+
+	@ConfigItem(
+		position = 7,
+		keyName = "isWearingFullProspector",
+		name = "Prospector's outfit equipped",
+		description = "Configures whether or not the collection exp should be multiplied by bonus"
+	)
+	default boolean isWearingFullProspector()
```
Adam-: You could eg check for prospector once a tick if in the mine.

runelite-client/src/main/java/net/runelite/client/plugins/blastmine/BlastMineCollectExp.java
```diff
@@ -0,0 +1,98 @@
...
+import net.runelite.api.Client;
+import net.runelite.api.Varbits;
+import net.runelite.api.widgets.Widget;
+import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.client.ui.overlay.Overlay;
+import net.runelite.client.ui.overlay.OverlayPosition;
+import net.runelite.client.ui.overlay.components.LineComponent;
+import net.runelite.client.ui.overlay.components.PanelComponent;
+
+class BlastMineCollectExp extends Overlay
```
Adam-: Is this really worth its own overlay? It might be able to just go into the ore overlay

mccore (commented):
A friend of mine requested this same thing and I stumbled across this PR. I'll admit I am not very Git savvy and its been a while since I wrote Java. I modified BlastMIneOreCountOverlay.java and had the same BlastMinePluginConfig.java as in this PR. I am not sure how to include the code properly in this PR and would appreciate your guidance. I took your comments into consideration @Adam-. I couldn't figure out how to properly subscribe on game tick to checking if the user is wearing prospectors so I left the config in there to manually set it at least as a version one. I did, however, remove the XP as a separate panel and instead added it as another line component of the Ore Count Overlay.

PR #6789, opened 2018-12-02
Static-colored fog

logarrhythmic:
[YouTube link](https://www.youtube.com/watch?v=jVNt9BcU_KI)
Adds a skybox and fog with configurable color and fog thickness. The fog is rectangular and aligned to the bounds of the currently drawn area.

Credit to Kamielvf for the original fog/skybox code and AdamBallinger for the fixes, original config options and the original PR #6671.


logarrhythmic:
Currently this code has the fog at the edge of the currently drawn chunk of map, or the draw distance if that's closer. Your code snippet is just for finding the bounds for the latter, right? I'll look at it tomorrow when I'm more awake. 

It does look strange underground if the map ends at the wall of the area, but that seems to be rare. Before the last commit, the one with "more robust scene bounds calculation" I got fog inside the HAM hideout's storage, and it looked pretty weird. There is stuff drawn beyond the wall, though, and moving the min/max coordinate storing to the scene uploader fixed that. 

A quick fix for any potential buggy fog in caves would be to just disable fog entirely outside the overworld, which seems to be detectable by the player position having a Y-coordinate above 4200. I'm using that to set the sky color to black in caves already, and it works in the few caves I've tried. Still needs more testing, though.

logarrhythmic:
Basically, I don't know how to leave #3 unfogged since I don't know how to get the scene edge without looking at each tile that gets drawn.
Can I get the scene edges by just looking at the dimensions for the scene tile array or something? I'm only getting the bounds by looking at each tile that gets drawn because I wasn't sure if the tile array was ordered, if I can get scene positions from tiles that don't exist, etc.

Adam-:
The scene edges are always at `x=0, y=0, x=13312,and y=13312`

logarrhythmic:
Ah, I see, that's simple enough.

Adam-:
Can you remove the skybox stuff so we can focus on the fog?

logarrhythmic:
Sure.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -99,6 +99,9 @@
 	private static final int SMALL_TRIANGLE_COUNT = 512;
 	private static final int FLAG_SCENE_BUFFER = Integer.MIN_VALUE;
 	private static final int MAX_DISTANCE = 90;
+	private static final int SCENE_TILES = 104; // in tiles
```
Adam-: There is a constant in `Constants` for this, which you can just use inline.

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -50,4 +62,14 @@ void main()
   vColor = vec4(rgb, 1.f - a);
   vHsl = float(hsl);
   vUv = uv;
+
+  if (useFog == 1)
+  {
+    float fogDistance = sqrtMultEdgeDistance(vPosition, sceneBounds);
+    vFogAmount = fogFactorLinear(fogDistance, 0, fogDepth);
```
Adam-:
```glsl
vFogAmount = fogFactorLinear(fogDistance, 0, fogDepth) * useFog;
```

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -50,4 +62,14 @@ void main()
   vColor = vec4(rgb, 1.f - a);
   vHsl = float(hsl);
   vUv = uv;
+
+  if (useFog == 1)
+  {
+    float fogDistance = sqrtMultEdgeDistance(vPosition, sceneBounds);
```
Adam-: What is this doing?

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
logarrhythmic: Thanks!

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/frag.glsl
```diff
@@ -58,4 +60,10 @@ void main() {
   } else {
     FragColor = smoothColor;
   }
+
+  // only apply fog to pixels that are not transparent. (Trees would be colored black where the texture is transparent)
```
Adam-:
```glsl
     vec3 mixedColor = mix(FragColor.rgb, fogColor.rgb, fogAmount);
     FragColor = vec4(mixedColor, FragColor.a);
```

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1051,6 +1063,23 @@ public void draw()
...
+			gl.glUniform1i(uniUseFog, config.enableFog() ? 1 : 0);
+			gl.glUniform4f(uniFogColor, config.fogColor().getRed() / 255f, config.fogColor().getGreen() / 255f, config.fogColor().getBlue() / 255f, 1f);
+			gl.glUniform1i(uniFogDepth, config.fogDepth() * Perspective.LOCAL_TILE_SIZE * Math.min(SCENE_TILES / 2, drawDistance) / 100);
+
+			// currently this doesn't care if the camera is in the fog
+			int fogWest = client.getCameraX2() - distance;
+			int fogEast = client.getCameraX2() + distance - Perspective.LOCAL_TILE_SIZE;
+			int fogSouth = client.getCameraZ2() - distance;
+			int fogNorth = client.getCameraZ2() + distance - Perspective.LOCAL_TILE_SIZE;
+			gl.glUniform4i(uniSceneBounds,
```
Adam-: Can all of this + the calculations be moved to the shader?

logarrhythmic: Possibly. I'm not sure if that's necessary, though, since this only needs to be calculated once per frame.

logarrhythmic:
[YouTube link](https://www.youtube.com/watch?v=9-q23UMflxM)
Here's a video of the current state of the fog.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
Adam-: I think it should be tried.

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/frag.glsl
```diff
@@ -58,4 +60,7 @@ void main() {
   } else {
     FragColor = smoothColor;
   }
+
+  vec3 mixedColor = mix(FragColor.rgb, fogColor.rgb, fogAmount);
```
Adam-: I would avoid trying to reading from 'out' variables since I'm not sure how well that is supported.

logarrhythmic: Yeah, fixed

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -23,20 +23,61 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#version 330
+#version 400
```
deathbeam: any reason why this was changed?

runelite-client/src/main/java/net/runelite/client/plugins/devtools/CameraSceneOverlay.java
```diff
@@ -0,0 +1,47 @@
+package net.runelite.client.plugins.devtools;
```
deathbeam: License header

runelite-client/src/main/java/net/runelite/client/plugins/devtools/CameraSceneOverlay.java
```diff
@@ -0,0 +1,47 @@
...
+import net.runelite.client.ui.overlay.OverlayLayer;
+import net.runelite.client.ui.overlay.OverlayPosition;
+
+class CameraSceneOverlay extends Overlay
+{
+	private final Client client;
+	private final DevToolsPlugin plugin;
+
+	@Inject
+	CameraSceneOverlay(Client client, DevToolsPlugin plugin)
```
deathbeam: private

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/frag.glsl
```diff
@@ -28,10 +28,12 @@ uniform sampler2DArray textures;
 uniform vec2 textureOffsets[64];
 uniform float brightness;
 uniform float smoothBanding;
+uniform vec4 fogColor;
```
DevDennis: This can be changed to vec3 because alpha will always be 1.0

logarrhythmic: Done

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
logarrhythmic: at any rate it's no longer necessary, so I reverted it

deathbeam:
what is `fogColor`? or is that customizable skybox color?

Roqua99:
Man this looks awesome good job!
 Will the fog colour change per location? In the video 6ou we're in Canifis, so black obviously suits.

What if I telepprt to the G.E, would I require to manually change from black or will this have a dynamic colour from location?

logarrhythmic:
`fogColor` in the config was a leftover from earlier, I removed it for the time being. `fogColor` is used in the shader instead of hardcoded 0 values to make it easier to implement colored fog once the sky map is done. `skyColor` would also work as a name for that one, especially if the shaders end up handling the skybox drawing. 

Rogua99, the dynamic skybox color is still far from done, so this fog will be black everywhere to match the non-colored skybox. The two features are obviously related, but mostly separate in terms of code, so implementing this while the sky color is still being worked on made sense to me.

Adam-:
I think we had decided to hold this until the skybox/color stuff is done, and release it at one.

logarrhythmic:
Yeah, that was also considered. I still personally think this enhances the look even in places where the sky would be blue later, but it's not as impressive without a colored sky.

logarrhythmic:
> 
> 
> I just want to quickly state this doesn't work in instances like pest control, not sure if this is intented but just letting you know.

Huh, that's interesting. Will have to investigate how this behaves with other instanced areas that are smaller than the normal scene size.

dginovker:
How is this coming along?

Adam-:
I decided to go with a stripped down version of this in 85460f16c336e7362767ec96bf253fbdea9a069b, Reopen a rebased version if you want to propose additional features.

PR #6794, opened 2018-12-02
Revert "Add support for kills left to XpTrackerPlugin"

### LINKED ISSUE #6479: The remaining-kills-until-level does not adjust base on defensive combat style
**Describe the bug**
When I train with long-ranged mode to gain both ranged and defense exp at the same time, I get 2 exp in each skill per damage done. But, the remaining kill count assumes 4 exp per damage done. The causes it to think I need half of the kills that I actually need.

**To Reproduce**
1) Equip ranged equipment
2) Switch to the long-ranged attack style
3) Fight some NPCs
4) See the kills remaining until next level is half of what it should be for the ranged and defense levels in the progress tracker sidebar

**Expected behavior**
The remaining kill count should show the number of kills needed based on 2 exp her damage done.

 - OS: Linux Mint 19
 - RuneLite version: 1.4.25

### LINKED ISSUE #6576: "Kills Left" reports incorrectly for split-exp attack types

abextm:
This feature doesn't work at all if you are:
 - Using an attack style that gives you multiple types of XP
 - Using ranged at range
 - Using a cannon
 - Barraging/Bursting
 - Attacking multiple monsters
 - Attacking monsters in series with differing HP values

As most of these are either difficult to solve or unsolvable and common cases this feature should be removed.

Fixes #6576
Fixes #6479
Closes #6643

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpActionType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpActionType.java
@@ -31,8 +31,7 @@ import lombok.Getter;
 @AllArgsConstructor
 public enum XpActionType
 {
-	EXPERIENCE("Actions"),
-	ACTOR_HEALTH("Kills");
+	EXPERIENCE("Actions");
 
 	private final String label;
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpState.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpState.java
@@ -27,7 +27,6 @@ package net.runelite.client.plugins.xptracker;
 import java.util.EnumMap;
 import java.util.Map;
 import lombok.NonNull;
-import net.runelite.api.NPC;
 import net.runelite.api.Skill;
 
 /**
@@ -38,11 +37,8 @@ import net.runelite.api.Skill;
  */
 class XpState
 {
-	private static final double DEFAULT_XP_MODIFIER = 4.0;
-	private static final double SHARED_XP_MODIFIER = DEFAULT_XP_MODIFIER / 3.0;
 	private final XpStateTotal xpTotal = new XpStateTotal();
 	private final Map<Skill, XpStateSingle> xpSkills = new EnumMap<>(Skill.class);
-	private NPC interactedNPC;
 
 	/**
 	 * Destroys all internal state, however any XpSnapshotSingle or XpSnapshotTotal remain unaffected.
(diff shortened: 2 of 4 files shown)


deathbeam:
1st and 2nd point can be easily fixed, 6th point is not an issue, because it just averages the kills. Cannon is probably easily fixable too. Maybe we can just add toggle for the kills left, because in common cases it is still useful, and it was heavily requested feature. Removing some feature entirely just because it do not works all the time is super lazy and half-assed solution. Should we remove slayer plugin just because it do not works in scenario 3, 4 and 5 too?

Abextm:
I have never seen this feature work correctly. Your "common case" is not very common. We can always re add this when it functions in any capacity. It is a lazy solution, but we haven't been able to make any others and having something that is more often than not broken is not a good way to make a quality product. The slayer plugin should probably detect when its going to be obviously wrong and disable itself, but that is unrelated to this issue.

deathbeam:
Well, the actions left for combat are very useless information regardless. So, by reverting this instead of feature that is broken in some scenarios, you have info that is completely useless all the time.

Abextm:
I would rather display useless information than misleading information

daviscook477:
How about we revert this and then do a change to make it so combat skills don't show anything - either kills left or actions left.

deathbeam:
Without kills left the information has no value all the time, with it it has some value often. As I said, if you do not like it, add toggle, and disable it, it is 3 lines of code. But imo it is still more valuable to have this than dont.

Abextm:
Actions left is mildly useful at barb fishing and other non-combat-combat-skill activities

Abextm:
If you want to keep the feature we can replace with a static string that says `Kills Left: 4` as that will be right atleast once, unlike the current solution.

deathbeam:
Well, that would be still more useful than actions left

Abextm:
Its still there, I only reverted the commit for Kills, the npc health refactor and the action type are still there so we can do agi laps.

deathbeam:
Oh, I actually did not checked the contents of the PR :D

Nightfirecat:
Per https://github.com/runelite/runelite/issues/18771#issuecomment-2667053315, kills/actions left are no longer displayed for combat skills.

PR #6811, opened 2018-12-04
Cluescroll Widget - Update width to scale up some.

Demivrga:
This will let more text be on same line for the clue scrolls, as it looks a bit weird when listing currently.

From:
d6251c99cc1152a4d7ab6ad759da5a9b

To:
ae5539001af178145b57c73a805f31b6


Demivrga:
Updated pull.

PR #6812, opened 2018-12-04
Add Skull/unskull notifications

Trevor159:
Useful at revs, idk where else.


15987632:
I have a skull timer PR also, I don't have the knowledge to make it 100% accurate when reseting your skull in pvp which is why that one is on hold for now. Yes I can move this to idle notifications but personally I didn't feel this was fitting.

15987632:
I'm pretty sure i'll keep the skull timer as a separate PR, this is in a finished state and the skull timer is not. And the skull timer would do into a completely separate plugin anyways.

Adam-:
I guess it is convenient to put this in the idle notifier but I'm not sure how many people skull when idle.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -64,11 +64,33 @@ default boolean logoutIdle()
 		return true;
 	}
 
+	@ConfigItem(
+			position = 4,
```
Adam-: Fix indenting

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -328,6 +333,29 @@ public void onHitsplatApplied(HitsplatApplied event)
 	public void onGameTick(GameTick event)
 	{
 		final Player local = client.getLocalPlayer();
+		SkullIcon currentTickSkull = local.getSkullIcon();
```
Adam-: Probably all of this logic should be in its own method.

Trevor159:
I think I was the only person who ever wanted this, probably useless.

PR #6816, opened 2018-12-04
grounditems: add option to only show loot

Adam-:


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -313,4 +313,15 @@ default int insaneValuePrice()
 	{
 		return 10000000;
 	}
+
+	@ConfigItem(
+		keyName = "onlyShowLoot",
+		name = "Only show loot",
+		description = "Only shows drops from NPCs and players",
```
peanubnutter, line 320: I would highly suggest changing this description. My friends and I all read this item and disregarded it entirely because it is inaccurate. It's not that the plugin only "shows drops from NPCs and players (this is ambiguous - I think it means PVP kills but could be interpreted to mean other players' loot)", but it doesn't actually mention that it HIDES other players' drops.

PR #6818, opened 2018-12-05
Add Xeric's aid and revitalisation potions to itemstats plugin

acrinklaw:
I altered the SaradominBrew and SuperRestore classes to better support the (-), normal, and (+) variants of Xeric's Aid / Revitalization potions, as well as adding the (+) variants to the itemstats plugin. I cannot find the information for (-) or normal modifiers so I did not add them, but I took the (+) information from the wiki and tested in game. The wiki improperly suggests that the standard potions are equivalent to sara brews/super restores but a friend who tested said the numbers were off. Adding these should be easy once I or someone else has the data


laagone:
The values are a bit off. At 99 prayer, revitalisation+ potions restore 40 prayer for me while the tooltip says +35. I tested these about 6 weeks ago on the tournament worlds with varying prayer levels and put my findings [here](https://oldschool.runescape.wiki/w/Revitalisation_(%2B)). I don't have any experience with lower level potions though, and Xeric's aid+ appears to be correct.

acrinklaw:
Interesting, I took what the wiki had said at the time and it worked for me but I'm not 99 prayer, it might have been coincidence. I'll change it and make a new pull request thanks for pointing it out

laagone:
@acrinklaw I got an email about you tagging me, I'm not sure if this is the thread and the comment is invisible or if I'm missing something, I'm not too familiar with Github. Anyway, the issue persists on a freshly launched client, version 1.5.6.1, it says +35 prayer when it restores 40.

acrinklaw:
Yeah sorry about that I typed up a reply but spoke too soon. I wanted to reuse as much code as possible but it seems that revitalisation potions are too different from super restores so I am working on a way around that right now and just need to test it out

acrinklaw:
> @acrinklaw I got an email about you tagging me, I'm not sure if this is the thread and the comment is invisible or if I'm missing something, I'm not too familiar with Github. Anyway, the issue persists on a freshly launched client, version 1.5.6.1, it says +35 prayer when it restores 40.

I completely forgot about this but I was never able to recreate the +35 / +40 issue. Does it still happen? 
image
This is what I saw in a video

PR #6819, opened 2018-12-05
Bank Tags: remove tab separators in tag tabs

trimbe:
Removes the tab separators and corrects the layout of items so that tag tabs mimic regular tabs.

Also disables drag reordering while in tag tabs as people may now mistake them as regular tabs - this may not be necessary or desirable.


deathbeam:
Maybe make this toggleable? Moving stuff around is slightly useful, depends on your bank categorization (I actually use it for pots as they come from my main tab where they are nicely organized, and so when i move them in tag tab it works nicely as well). Also, the separators are useful too, if your bank is nicely organized (e.g, I have consumables in tab 1, then armour/weapons, special stuff, and so in tag tab I have consumables and gear nicely separated)

trimbe:
Sure. I figured nobody would want the separators, but my bank is very messy, if your bank is organized I can see the value. 

Would a default on config for the drag disable be useful?

deathbeam:
Removing the separators and drag disabling by default sounds fine, with toggle to disable (probably just 1 toggle for both is fine)

Lucwousin:
If you close your bank with a bank tag open and reopen bank (with remember last tag tab enabled) you'll get the same tag tab but with divider lines.

The other way around, if a tag tab closes when not closing it by clicking the "real tab" you want (e.g. messaging someone, searching for items or simply pressing the same tag tab again), you'll get your whole bank without seperators.

trimbe:
Fixed both of those issues.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -157,6 +158,12 @@ public void shutDown()
 		shiftPressed = false;
 	}
 
+	private boolean isSearching()
+	{
+		return client.getVar(VarClientInt.INPUT_TYPE) == 11
```
Nightfirecat: The next line is more self-explanatory, but what does `INPUT_TYPE == 11` represent? That should probably be a constant.

trimbe: True, turns out we even have an enum for it already.

Lucwousin:
`2019-01-30 04:25:16 [Thread-5] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
	at net.runelite.client.plugins.banktags.BankTagsPlugin.isSearching(BankTagsPlugin.java:164)
	at net.runelite.client.plugins.banktags.BankTagsPlugin.onScriptCallbackEvent(BankTagsPlugin.java:218)
	at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:72)
	at net.runelite.client.eventbus.EventBus.post(EventBus.java:218)
	at net.runelite.client.callback.Hooks.post(Hooks.java:130)
	at client.vmExecuteOpcode(client.java:104)
	at client.copy$runScript(client.java:153)
	at client.r(client.java:135)
	at client.copy$runWidgetOnLoadListener(client.java:3642)
	at o.ao(o.java:843)
	at f.jl(f.java:11370)
	at client.hc(client.java:6839)
	at client.fn(client.java:3281)
	at client.ae(client.java:1552)
	at ah.u(ah.java:330)
	at ah.run(ah.java:309)
	at java.lang.Thread.run(Thread.java:748)`

I get this exception multiple times before opening the bank after logging in.

The tag tabs also seem to be bugged, moving the items every time you reopen a tab.
image
image

Finally, when opening a tag tab when search is open the items dance around every tick.
[gif](https://i.imgur.com/lMmaZp3.gifv)

Jacoblairm:
Can this please be considered? I always misclick on the lines and the tab looks a lot nicer without the them.

Jilms:
would love to see this get worked on again and added to the client in the future, my friends and I have been wanting for something like this for the tag tabs for a while now.

Jacoblairm:
Keen for this!

trimbe:
This is now updated to re-layout the tag tabs in Java instead of relying on multiple script hooks.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -457,6 +461,88 @@ public void onConfigChanged(ConfigChanged configChanged)
...
+			if (itemContainer == null)
+			{
+				return;
+			}
+
+			int items = 0;
+			final int itemsPerRow = 8;
+			final int itemVerticalSpacing = 36;
+			final int itemHorizontalSpacing = 48;
+			final int itemRowStart = 51;
```
Nightfirecat, line 497: Define these as `private static` plugin constants.

deathbeam:
Needs rebase again

Nightfirecat:
I've rebased this to resolve the WidgetInfo conflict here.

Nightfirecat:
Rebased again, and updated bank widget stuff. Still looks and works fine ingame.

raiyni:
This has merge conflicts

Nightfirecat:
Rebased to fix conflicts again.

Nightfirecat:
I found a bug:

1. Open any bank tag tab
2. Right-click an item and select "Withdraw X"
3. Click on the active bank tag tab

The above causes the bank to display the "all items" tab briefly, then return to the tag tab display without laying out items as this PR should adjust them. (tab lines still appear)

Nightfirecat:
Another minor issue is that tag tabs with few enough items to not fill one bank screen end up having a scrollbar which does not behave the same as bank tabs which fit the same criteria--it is slightly larger than the full scroll height and behaves a bit odd if you try to click and drag it. Doesn't seem to cause any functional issues but is a bit visually odd.

trimbe:
I believe I've fixed the scroll issue you described. The Withdraw-X bug seems to be caused by a bug within bank tags rather than this PR, though. The bank tag should close if it's clicked again while active. I will take a look at it but I think it's out of the scope of this PR.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -458,6 +467,94 @@ public void onConfigChanged(ConfigChanged configChanged)
...
+					child.setHidden(true);
+				}
+			}
+
+			int itemContainerHeight = client.getWidget(WidgetInfo.BANK_ITEM_CONTAINER).getHeight();
+			// add a second row of height here to allow users to scroll down when the last row is partially visible
+			int adjustedScrollHeight = (items / ITEMS_PER_ROW) * ITEM_VERTICAL_SPACING + ITEM_VERTICAL_SPACING;
+			if (adjustedScrollHeight > itemContainerHeight)
+			{
+
```
Nightfirecat, line 543: One extra line too many
```suggestion
```

Nightfirecat:
Between your last change and #11654, I haven't found any other bugs. :+1:

raiyni:
If you go to a tag tab without first searching the bank and toggle the menu option, nothing happens.

trimbe:
This will now layout the bank when the config option is toggled. @raiyni I think that's what you were referring to?

raiyni:
Yes, it was only an issue before you used normal bank search.

PR #6820, opened 2018-12-05
Move core configs outside Scroll Pane

TheStonedTurtle:
Just an idea. Also updated the scroll increment (with mouse wheel) to be exactly the height of one config entry regardless of OS scroll amount.


deathbeam:
What was wrong with the way it was done here? https://github.com/runelite/runelite/pull/6304/files . Just making plugin favorited is better imo, as what you have looks quite bad.Also, why move chat colors? Noone cares about them and people did not had trouble finding that config.

PR #6826, opened 2018-12-05
Revert "don't check chatbox status on last tick for fairyring search"

raiyni:
Reverts runelite/runelite#6774

This causes the chatbox to close everywhere.


deathbeam:
If we will not have fix until tomorrow's release this needs to be reverted.

Abextm:
Fixed in 05a17e00654f6d006a06746b29456924cc9c5ed7

PR #6853, opened 2018-12-07
[WIP] Loot Tracker persistent data

TheStonedTurtle:
Need to make minor adjustments to the UI. Also not sure if I should change this so the config key includes the NPC name instead of date so we can clear data specific to NPCs.

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/loottracker/GameItem.java
+++ b/http-api/src/main/java/net/runelite/http/api/loottracker/GameItem.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2018, TheStonedTurtle <https://github.com/TheStonedTurtle>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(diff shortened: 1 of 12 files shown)


deathbeam:
1. This will be A LOT of data, so it should definitely not use config system (if not because of misuse then because of data overhead by storing everything as json string). You could optimize the storage a lot more if you created separate database structure (it will be at least 2 tables)
2. This should definitely not load all the data to sidebar as again, it will be A LOT of data. You should maybe limit it to current day if anything.

TheStonedTurtle:
> Adam: I may be ok with stuffing it into json and storing it into the runelite config. but there might be some length limits there

I made this mostly as a prototype to see if this was the way Adam wanted to move forward. Whatever method of storing the data is decided on let me know and I'll work on it. 

Once the persistent data functionality is finalized I'll work on limiting data sent to the UI and adding a clear data button.

deathbeam:
Well, as I said, atm we are barely handling XP tracker data, that are optimized properly. So I am telling you now that using config system for this is very bad idea, otherwise at some point we will be forced to aggressively clean loot tracker data (and cleanup will happen sooner and be more aggressive based on how bad is the implementation).

deathbeam:
What I would like to see on remote as DB structure

```
LootRecord | id | userId | type (enum, or separate table with types) | npcId |
LootItem   | id | lootRecordId | itemId | itemQuantity |
```

All this needs to be behind RL authentication.

Local storage might as well be giant json, it doesnt matter.

Abextm:
If we do local storage I would like to see csv or something that is as trivial as possible to parse

deathbeam:
CSV could work, then you would be able to also import it to excel or w/e

TheStonedTurtle:
I can output to a .csv file which will open in excel I just need to know how often to write the file. Pretty sure appending to the file on every loot received wouldn't be viable.

forsco:
I personally prefer having the data on my local machine vs's a remote. If the remote is ever purged I lose all my data which I would of thought would be `persistent`.

I don't want to lose the original chunks of my log due to having to purge a database and considering that the user base is forever growing this should be considered.

deathbeam:
We can add option to export the remote data at any point. But, it will work like config works, e.g you are logged in -> stores to remote, you are not logged in -> stores to local.

Also if you read the previous discussions about this on Discord, the remote part is more important, as then we can do cool stuff with it on website.

forsco:
And if a purge happens and the user is unable to export?

steffenhauge:
> 
> 
> We can add option to export the remote data at any point. But, it will work like config works, e.g you are logged in -> stores to remote, you are not logged in -> stores to local.
> 
> Also if you read the previous discussions about this on Discord, the remote part is more important, as then we can do cool stuff with it on website.

Would still prefer to have the option to also save to local, even if logged in then.

deathbeam:
Did you read my reply? Mainly the second part

deathbeam:
Well, you can view data for today in client, otherwise you can use website or excel, I dont think anything else is needed

Adam-:
The *best* solution is what @deathbeam said which is to store the loot in the database server side. I don't know what impact this will have on the server load, it could be unsustainable given the current architecture. I think we would have to do at least 1 release with it not visibly enabled/don't tell anyone to see if it explodes or not, if we did that.

TheStonedTurtle:
Alright, I'll get the API (and CSV output for non-logged in users?) working in this PR and we'll re-evaluate then.

PR #6868, opened 2018-12-07
Separate the combat idle notification from the fishing idle notification

Twiglet1022:
I think that it was a bad idea to combine the fishing and combat idle notifications under a single config option. I'd like to be able to turn off combat idle notifications without having to turn them back every time I go fishing.

I've seen a few complaints about this in discord too since my PR that improved the fishing notification and simultaneously grouped it with the combat idle notification under "Idle Interaction Notifications". Having now played with this a bit more I agree with the complaints and would like to propose it be changed.

I've kept the fishing notification generic enough that further non-combat interaction idle notifications can be added to it easily in future if ever it becomes relevant.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -45,7 +45,7 @@ public interface IdleNotifierConfig extends Config
 	@ConfigItem(
 		keyName = "interactionidle",
 		name = "Idle Interaction Notifications",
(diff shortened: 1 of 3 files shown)


deathbeam:
I think master functionality is fine and do not needs any changes

Twiglet1022:
With fishing you can use a delay of 0 and it's fine, but if you play with that delay while doing any combat the notification triggers after pretty much every kill so it's a bit annoying. 

I just turn off the idle notifier (easier than adjusting the delay) now when doing combat, but it's not ideal and decoupling them is pretty simple change.

Twiglet1022:
Nevermind 🧹

PR #6869, opened 2018-12-08
Numerous performance optimizations

KGoodale13:
I ran a bunch of profiles on the client and optimized expensive functions as they showed up. Most of these are in overlay render hooks.


runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -75,9 +75,12 @@ protected void shutDown() throws Exception
 	@Subscribe
 	public void onVarbitChanged(VarbitChanged event)
 	{
-		favor = client.getVar(Varbits.KINGDOM_FAVOR);
-		coffer = client.getVar(Varbits.KINGDOM_COFFER);
-		processInfobox();
+		if(isInKingdom())
```
Abextm: this is
1) wrong
2) saving zero time

KGoodale13: Hmm yea it is blatantly wrong and calls IsInKingdom twice. I think profiling via instrumentation inflated the cost of getting varbits and made it appear like that was superexpensive. My bad

Nightfirecat:
If you don't mind me asking, how did you execute this profiling, and what profiles did you use?

runelite-api/src/main/java/net/runelite/api/GameState.java
```diff
@@ -62,6 +66,10 @@
 	 */
 	HOPPING(45);
 
+	private static final Map<Integer, GameState> stateValueMap =
+			Arrays.stream(GameState.values())
```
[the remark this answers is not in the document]
deathbeam: its static final value, it will be called exactly once so it doesnt matter if it is stream or loop

PR #6870, opened 2018-12-08
Allow lowering of outer zoom limit in the camera zoom plugin

### LINKED ISSUE #6832: Allow lowering of maximum outer zoom in the Camera Zoom plugin
Duplicate of #5930, which was closed but I don't think the issue they had was addressed fully, the new zoom cap is too high for me and I'd like to _constrict_ rather than _expand_ it. The allowing of putting negative values for the "Expand outer zoom limit" option seems like the simplest way of doing this

Twiglet1022:
Closes #6832 

Added support for negative values on the int spinner and applied to the camera zoom plugin.

<Details><Summary>Expand for GIF demonstration</Summary>
decrease outer zoom limit
</Details>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigItem.java
@@ -46,4 +46,6 @@ public @interface ConfigItem
 	String warning() default "";
 
 	boolean secret() default false;
+
+	int minimumInt() default 0;
 }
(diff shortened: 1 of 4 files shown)


deathbeam:
As I already commented on original version #6512, having integer range related stuff on every single config item is not ideal, you should check how is it implemented there

Twiglet1022:
Ah I wasn't aware of that PR. That is a much better approach. I'll wait for that change to be merged.

Adam-:
This has been added in c5bb84a57b745a689c4c621d7ac57c901e289641 if you want to rebase this PR

PR #6877, opened 2018-12-08
Notify on Full Special Attack

### LINKED ISSUE #1727: Notify on full special attack

jkufro:
This is made under the Idle Notifier plugin as requested by @Adam- in the #3922 thread. Users should  be notified when special attack energy naturally regenerates to a set value (and if set to 0 it will be disabled just like other options in the plugin).

Fixes #1727
Closes #3922

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -107,4 +107,15 @@ public interface IdleNotifierConfig extends Config
 	{
 		return 0;
 	}
+
+	@ConfigItem(
+		keyName = "spec",
+		name = "Special Attack Energy Notification Threshold",
+		position = 8,
+		description = "The amount of spec energy reached to send a notification at. A value of 0 will disable notification."
+	)
+	default int getSpecEnergyThreshold()
+	{
+		return 0;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -42,6 +42,7 @@ import net.runelite.api.NPC;
 import net.runelite.api.NPCComposition;
 import net.runelite.api.Player;
 import net.runelite.api.Skill;
+import net.runelite.api.VarPlayer;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.AnimationChanged;
 import net.runelite.api.events.GameStateChanged;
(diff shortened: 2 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -382,6 +386,41 @@ public void onGameTick(GameTick event)
...
+		// having this call here avoids repeating the line inside the checkFullSpecEnergy function
+		lastSpecEnergy = client.getVar(VarPlayer.SPECIAL_ATTACK_PERCENT);
+	}
+
+	private boolean checkFullSpecEnergy()
+	{
+		if (config.getSpecEnergyThreshold() == 0) {
+			return false;
+		}
+		if (client.getVar(VarPlayer.SPECIAL_ATTACK_PERCENT) != lastSpecEnergy && lastSpecEnergy < (config.getSpecEnergyThreshold() * 10))
```
Adam-: You should just be calling `client.getVar(VarPlayer.SPECIAL_ATTACK_PERCENT)` once per tick, not four times.

jkufro: Very true. I made changes to call it once per tick.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -107,4 +108,15 @@ default int getOxygenThreshold()
 	{
 		return 0;
 	}
+
+	@ConfigItem(
+			keyName = "spec",
+			name = "Special Attack Energy Notification Threshold",
+			position = 7,
+			description = "The amount of spec energy naturally regenerated to send a notification at. A value of 0 will disable notification."
```
raiyni: You should shorten this up or add line breaks.

jkufro: Would dropping the word 'naturally' be sufficient?

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -382,6 +386,43 @@ public void onGameTick(GameTick event)
...
+		{
+			return false;
+		}
+		if (currentSpecEnergy != lastSpecEnergy && lastSpecEnergy < (config.getSpecEnergyThreshold() * 10))
+		{
+			if ((config.getSpecEnergyThreshold() * 10) <= currentSpecEnergy)
+			{
+				// heuristic for checking if energy was naturally regained based on if the difference between the
+				// last spec energy and current spec energy is small enough AND energy is increasing
+				if ((currentSpecEnergy - lastSpecEnergy) > 0 && (currentSpecEnergy - lastSpecEnergy) <= 100)
```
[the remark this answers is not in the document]
jkufro: You're right. I removed the redundant check

Adam-:
I've refactored this to 1ad41c76fd40c57dcb90d4ef5bca6e8031bd59f0 - can you test this commit still behaves correctly?

PR #6878, opened 2018-12-08
Allow lowering of outer zoom limit in the camera zoom plugin

### LINKED ISSUE #6832: Allow lowering of maximum outer zoom in the Camera Zoom plugin

Twiglet1022:
Closes #6832

Now uses the recently added Range annotation to allow reducing the outer zoom limit by up to 400.

<Details><Summary>GIF</Summary>

decrease outer zoom limit
</Details>


Twiglet1022:
There was a problem with this. I tested what would happen if someone already had their config set to a value outside the allowed range of [-400,400] e.g. 600 and it causes the config panel to break and show only:

image

Instead of:

image

So to fix that I've caught the exception in the JSpinner and forced it into the acceptable range when the exception is thrown. I also had to forcefully update the config when this happens or else the plugin would continue to use the old value until it was updated again manually.

Should be good now.

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
```diff
@@ -100,7 +97,7 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
 
 		if ("outerZoomLimit".equals(event.getEventName()))
 		{
-			int outerLimit = Math.max(OUTER_CONFIG_ZOOM_LIMIT_MIN, Math.min(OUTER_CONFIG_ZOOM_LIMIT_MAX, zoomConfig.outerLimit()));
+			int outerLimit = zoomConfig.outerLimit();
```
Adam-: The `@Range` is really supposed to be config-panel only enforced, you should not assume the value from the config is in a valid range probably

deathbeam: Well as there is no good way to get the actual range, and the only way to make this wrong is from config, i think it is probably fine? Even though with config from older release this can be a problem

Twiglet1022: The other change I made should solve the issue of old configs being out of range

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -350,7 +350,28 @@ void openGroupConfigPanel(PluginListItem listItem, Config config, ConfigDescript
 					max = range.max();
 				}
 
-				SpinnerModel model = new SpinnerNumberModel(value, min, max, 1);
+				SpinnerModel model;
```
raiyni: This should be simplified to a constraint before setting model rather than a try/catch after .

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
Adam-: If you remove the resetting of the config, the other change is actually only +1 line which is `constrainToRange` on value. I would do that and then readd the min/max checks here (also constrainToRange actually).

Twiglet1022: Thanks I didn't know about Ints.constrainToRange()

Forcing an update of the config was flawed anyway because that only runs when the config panel is opened. It would be possible for a user to carry over an invalid config for as long as they didn't open the config options.

Therefore I've re-added the min/max checks to ZoomPlugin.

Is the way I've done this okay? I exposed the min and max values in the config to guarantee that they're consistent between the config and the plugin.

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomConfig.java
```diff
@@ -42,10 +43,16 @@ default boolean innerLimit()
 		return false;
 	}
 
+	int OUTER_LIMIT_MIN = -400;
```
Adam-: We generally put (implicit)  static/final fields at the top of classes/interfaces

runelite-client/src/main/java/net/runelite/client/plugins/zoom/ZoomPlugin.java
Adam-: This looks much better.

PR #6885, opened 2018-12-09
Add instance map for clues

MedPeppa:
Is nothing more frustrating when you have to click on world map for every single step and inconter that laggy map.Add instance map that open automatically


Kamielvf:
I don't know how you managed to open this as a PR, but this should be closed and re-opened as an issue/feature request.
[Click here](https://github.com/runelite/runelite/issues/new/choose)

PR #6886, opened 2018-12-09
Change movable bounty hunter widget

Kamielvf:
This changes the movable bounty hunter widget ID to use the root widget instead of the stat child widget. As it currently is, it only allows moving of the stat widget _inside_ of its parent, so it is not possible to move the whole widget around.

Before:
image

After:
image


Kamielvf:
Small thing to note: Jagex added their own options to dock this widget on the left, middle or right. Due to our overlay system overriding widget locations, this will move the widget to it's default top-right position (until moved around using our movable feature) for people who have it docked somewhere else trough Jagex's docking options.

PR #6890, opened 2018-12-09
Move cerberus plugin overlay above ghosts

MagicfTail:
The plugin will now show the ghosts' attack style above their head, instead of in an overlay. This makes it feel more integrated into the game and makes it so you need to pay attention when killing cerb, while still helping colourblind people know what to pray.

https://i.imgur.com/0YxtOfC.gif


runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
-import net.runelite.client.ui.overlay.components.PanelComponent;
 
-@Singleton
 public class CerberusOverlay extends Overlay
 {
-	private final CerberusPlugin plugin;
-	private final SkillIconManager iconManager;
-	private final PanelComponent panelComponent = new PanelComponent();
+	private static final Color COLOR_ICON_BACKGROUND = new Color(0, 0, 0, 128);
+	private static final Color COLOR_ICON_BORDER = new Color(0, 0, 0, 255);
```
Nightfirecat: This can just be `Color.BLACK`.

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
+	private static final int OVERLAY_ICON_DISTANCE = 50;
+	private static final int OVERLAY_ICON_MARGIN = 8;
+	private static final int OVERLAY_Z_OFFSET = 100;
+
+	private Client client;
+	private CerberusPlugin plugin;
+
+	@Inject
+	private SkillIconManager iconManager;
```
Nightfirecat:
```suggestion
	private final SkillIconManager iconManager;
```

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
 
 	@Inject
-	CerberusOverlay(final CerberusPlugin plugin, final SkillIconManager iconManager)
+	public CerberusOverlay(Client client, CerberusPlugin plugin)
```
Nightfirecat: Constructor should be `private`.

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
-	private final CerberusPlugin plugin;
-	private final SkillIconManager iconManager;
-	private final PanelComponent panelComponent = new PanelComponent();
+	private static final Color COLOR_ICON_BACKGROUND = new Color(0, 0, 0, 128);
+	private static final Color COLOR_ICON_BORDER = new Color(0, 0, 0, 255);
+	private static final int OVERLAY_ICON_DISTANCE = 50;
+	private static final int OVERLAY_ICON_MARGIN = 8;
+	private static final int OVERLAY_Z_OFFSET = 100;
+
+	private Client client;
```
Nightfirecat:
```suggestion
	private final Client client;
```

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
+	private CerberusPlugin plugin;
```
Nightfirecat:
```suggestion
	private final CerberusPlugin plugin;
```

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
+
+	@Inject
```
Nightfirecat: Inject this in ctor.

```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
-		plugin.getGhosts().stream()
-			// Iterate only through the correct amount of ghosts
-			.limit(CerberusGhost.values().length)
-			.forEach(npc -> CerberusGhost
-				.fromNPC(npc)
-				.ifPresent(ghost -> panelComponent
-					.getChildren()
-					.add(new ImageComponent(iconManager.getSkillImage(ghost.getType())))));
+					int bgPadding = 4;
+					int currentPosX = 0;
```
Nightfirecat: This value isn't ever changed? Seems not needed.

runelite-client/src/main/java/net/runelite/client/plugins/cerberus/CerberusOverlay.java
```diff
@@ -24,53 +25,108 @@
...
+
+					BufferedImage icon = getIcon(ghost.getId());
+
+					if (icon == null)
+					{
+						continue;
+					}
 
-		panelComponent.getChildren().clear();
+					int width = OVERLAY_ICON_MARGIN + icon.getWidth();
```
Nightfirecat: This part of the render loop might be easier to read if it were moved into its own function, perhaps `private static void renderPrayerIndicator(Graphics2D graphics, BufferedImage icon, Point point)`?

runelite-client/src/test/java/net/runelite/client/plugins/cerberus/CerberusPluginTest.java
```diff
@@ -1,92 +0,0 @@
...
-import org.junit.Before;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-import org.mockito.runners.MockitoJUnitRunner;
-
-@RunWith(MockitoJUnitRunner.class)
-public class CerberusPluginTest
```
Nightfirecat: imo the test should be rewritten, not removed.

MagicfTail:
I made it the same as the gorilla plugin since that's what people are used to, I can try to get it looking similar to the ingame prayer overheads.

Not really, I just didn't see any reason for it to be there. I'll change it back if that's prefered.

Abextm:
We should not use the vanilla prayer style as we are not indicating what the npc will not take damage from, we are indicating what attack style the npc is using, which is very different.

Abextm:
We should use the same one as the gorilla plugin for the same reason as in the gorrila plugin. It should not look like the vanilla one because it does not do the same thing as the vanilla one.

deathbeam:
Well I am just saying that it do not looks very good and I dont like it, and if magic wants to mess with plugins he should spend more effort than just copy/pasting code from other plugin.

Abextm:
I agree it does not look very good, but making it look like vanilla will create a much worse problem, so it should not do that. This and the gorilla plugin should look the same because they convey the same information.

MagicfTail:
I think the gorilla overlays look fine, and I wanted it to be the same as them. If anyone has any ideas on how to make it look better, feel free to chime in.

deathbeam:
Well if you want to stick to use same overlay for both, make common method in OverlayUtil like we do for every single other common overlay we have and use it in both plugins

abextm:
Jagex's [recent statement](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) prohibits `Indicates what prayers to use in what order, for example in the Cerberus boss fight`

Issue #6778, opened 2018-12-01
Add ability to hide items in bank that are under X value


kingjasko:
Add the ability to hide/show items in bank that are under/above X value. 

Being able to do so will save people time to look for items manually when they need to liquidate their bank and sell big ticket items. It seems pretty self-explanatory and quite easy to do considering bank tags are so advanced nowadays.

PR #6901, opened 2018-12-11
Add bank filter by price range

tanlines:
#6778 

Option to filter by GE or HA price
Ability to set price range
Hides items in bank not within price range when doing any bank filtering, including with tags
When active, if no filter is being applied, automatically applies a "pricefilter" search to begin filtering items
    If user opens search box, this filter is disabled (bug: requires spam clicking to open search box if on All items tab)

Mostly due to my lack of knowledge of how the bank filtering works, I've spent the last couple days working on this plugin, hoping for some feedback especially on where I should be calling banksearch.search() so the bug doesn't occur.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
@@ -87,10 +87,10 @@ public interface BankTagsConfig extends Config
 	}
 
 	@ConfigItem(
-			keyName = "minimumValue",
-			name = "Minimum price",
-			description = "Configures the minimum value of the bank item to show",
-			position = 6
+		keyName = "minimumValue",
+		name = "Minimum price",
+		description = "Configures the minimum value of the bank item to show",
+		position = 6
 	)
 	default int lowValuePrice()
 	{
@@ -98,10 +98,10 @@ public interface BankTagsConfig extends Config
 	}
 
 	@ConfigItem(
-			keyName = "maximumValue",
-			name = "Maximum price",
-			description = "Configures the maximum value of the bank item to show",
-			position = 7
+		keyName = "maximumValue",
+		name = "Maximum price",
+		description = "Configures the maximum value of the bank item to show",
+		position = 7
 	)
 	default int highValuePrice()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
@@ -54,6 +54,61 @@ public interface BankTagsConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "priceFilter",
+		name = "Filter by price",
+		description = "Enable hiding of items not within given price range.",
+		position = 3
+	)
+	default boolean priceFilter()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "highAlchemyPrice",
+		name = "Filter by HA value",
+		description = "Enable filtering by HA value.",
+		position = 4
+	)
+	default boolean filterHAValue()
+	{
+		return true;
+	}
+
+	@ConfigItem(
+		keyName = "rememberTab",
+		name = "Filter by GE value",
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
```diff
@@ -53,6 +53,61 @@ default boolean rememberTab()
 		return true;
 	}
 
+	@ConfigItem(
```
raiyni: This should just work if bank tags plugin is enabled

runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
```diff
@@ -169,6 +178,35 @@ public void removeTag(int itemId, String tag)
 		}
 	}
 
+	public boolean itemWithinPriceRange(Item item)
```
raiyni: Nothing is done with tag search, so why is this here?

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -205,6 +206,21 @@ else if (!tagSearch)
 				{
 					intStack[intStackSize - 2] = itemName.contains(search) ? 1 : 0;
 				}
+
+				if (config.priceFilter()) {
+					if (search.equals(PRICE_FILTER)) {
+						intStack[intStackSize - 2] = 1;
+					}
+					if (intStack[intStackSize - 2] == 1) {
+						ItemContainer container = client.getItemContainer(InventoryID.BANK);
```
raiyni: This should be cached on itemcontainerchanged

runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
tanlines: I wasn't sure where to put it.
I considered the price of the item as a sort of "tag".

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
tanlines: I don't think it should be on by default if bank tags is enabled, as it forces a filter on all items and all tags.

raiyni: It shouldn't always be searching price value.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
tanlines: Are you saying I should cache it or it's already cached?
How does this change what I'm doing? I still need to check the items individually and filter them.

raiyni: You are querying the bank item container for every item in the bank and then looping over it.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
tanlines: The original issue didn't specify how to do the filtering.

How do you think it should be activated then?
If you have to type the keyword in every time the search resets it's very cumbersome in my opinion. Rather, I let the user toggle it on and off in the settings.

raiyni: Through search tags of `ge` or `ha` would make more sense rather than always having a clunky filter on. 

Plus it's faster to enable it that way.

tanlines: How about a "price:" tag, and let the user toggle HA or GE values in the sidebar? I don't think these would be changed very much, and easier to just remember 1 price tag.

It'll use a similar method of checking if the user is searching by tag and also let the user filter by item name as well.

raiyni: why would you want that much complexity

Issue #6836, opened 2018-12-06
Add laps/hr to agility display


Feenoh:
title

jackbrown1993:
I think @Feenoh is asking for a display showing an estimated amount of laps per hour based on the time you are taking to complete the current course.

deathbeam:
The preferred way to go about implementing this is move the laps data to XpTrackerPlugin (or if used in both agi and xp tracker, to shared place) and then integrate laps done to XpTrackerPlugin in similar way as kills done and kills/hr are integrated atm.

LlemonDuck:
This was added in a626e97176a872425a9209ca03eaa0325e277c30

PR #6903, opened 2018-12-11
Add laps per hour for agility courses

tanlines:
#6836

Adds laps per hour for agility courses using the XPTrackerPlugin
When doing laps, the XPTracker will change "Actions" to "Laps"


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -68,4 +71,19 @@ void resetLapCount()
 		totalLaps = 0;
 		lapsTillLevel = 0;
 	}
+
+	int getLapsHr()
```
JAnken123: I feel like there's no need for three separate methods and reduces readability in this case.

Adam-:
Instead I would probably just put this on the existing agility overlay which shows laps until level.

PR #6908, opened 2018-12-11
Bank Value: Use ItemContainer instead of BankItemQuery

trimbe:
Since we can easily calculate tabs using a few varbits, there's no reason for this to be using a `BankItemQuery`.


Adam-:
Why is the title change mixed in with the commit to remove the query?

trimbe:
Because I forgot to do that, apparently. Commits split now.

runelite-client/src/main/java/net/runelite/client/plugins/bankvalue/BankValuePlugin.java
```diff
@@ -54,45 +60,101 @@
...
 		return configManager.getConfig(BankValueConfig.class);
 	}
 
+	private int lastTab = -1;
+
 	@Override
 	protected void shutDown()
 	{
-		clientThread.invokeLater(bankTitle::reset);
+		clientThread.invokeLater(() -> bankSearch.reset(false));
```
Adam-: What's with the behavior change here?

trimbe: This just forces a rebuild of the bank, which resets the title since I'm doing it in script now.

PR #6909, opened 2018-12-11
Fix freeze on disabling GPU plugin

### LINKED ISSUE #6524: Screen freezes when turning off GPU plugin
**Describe the bug**
Turning off the GPU plugin freezes the screen. Happens even when not logged in. The game will still accept inputs so I can log in and interact with the game, but the image stays frozen from the moment I turned the plugin off.

Resizing the game window or closing/collapsing the sidebar unfreezes the image, as does turning the GPU plugin back on.

This only happens when RuneLite runs on my dedicated graphics card: NVIDIA GeForce GTX 960M. I don't get this freeze when running on integrated graphics: Intel HD Graphics 530 (but then that has other issues).

**To Reproduce**
Steps to reproduce the behaviour:
1. Load up the client (Optional: log in)
2. Turn on the GPU plugin if it isn't already turned on.
3. Turn it off.
4. Screen will be frozen.

 - OS: Windows 10 Home
 - RuneLite version: 1.5.0 and 1.5.1-SNAPHSHOT both have this issue
 - Launcher version: Unknown
 - Graphics card: NVidia GeForce GTX 960M

**Additional context**
I ran 1.5.1-SNAPSHOT in debug mode and grabbed the red text from the debug console if it helps:
[Debug Extract.txt](https://github.com/runelite/runelite/files/2591297/Debug.Extract.txt)
Also part of the dxdiag output relating to graphics:
[DxDiag Extract.txt](https://github.com/runelite/runelite/files/2591303/DxDiag.Extract.txt)

Twiglet1022:
Fixes #6524 

Not gonna lie, I don't really know what I'm doing here because there doesn't seem to be a way to see exactly what setReplaceCanvasNextFrame() does other than inferring from its name, but this fixes the issue I was having by calling setReplaceCanvasNextFrame(true) on shutdown of the GPU plugin.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/GameEngine.java
+++ b/runelite-api/src/main/java/net/runelite/api/GameEngine.java
@@ -53,4 +53,6 @@ public interface GameEngine
 	boolean isClientThread();
 
 	void resizeCanvas();
+
+	void setReplaceCanvasNextFrame(boolean replace);
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -420,6 +420,8 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 
 			// force main buffer provider rebuild to turn off alpha channel
 			client.resizeCanvas();
+
+			client.setReplaceCanvasNextFrame(true);
 		});
 	}
 


Adam-:
This causes the gpu plugin to get stuck in a loop enabling/disabling if the canvas is invalidated for any reason.

Twiglet1022:
Is there any way to update this PR now that it's been merged? Pushing to the same branch doesn't seem to update this PR anymore.

Anyway, I've made a change that solves the issue with the client getting stuck in an endless loop when the canvas is invalidated and still solves my initial issue. All I did was move the `client.setReplaceCanvasNextFrame(true)` to the end of the `shutDown()` method, outside of `clientThread.invoke()`

I don't quite understand what that code is doing, but I observed while debugging that although `shutDown()` was only called once after the canvas was initially invalidated, any code inside `clientThread.invoke()` was called twice and it was the second time that `client.setReplaceCanvasNextFrame(true)` ran which seemed to cause the problem.

By moving `client.setReplaceCanvasNextFrame(true)` to the very end my issue is still fixed, but no endless looping.

Nightfirecat:
Just answering the first question: you need to open a new PR.

PR #6913, opened 2018-12-12
Add cache control to more API endpoints

KGoodale13:
Adding cache control to a few more API endpoints. I realize there's already some caching built into some of these endpoints but this will allow CloudFlare to cache it on the edge layer.

### DIFF
--- a/http-service/src/main/java/net/runelite/http/service/CacheControlFilter.java
+++ b/http-service/src/main/java/net/runelite/http/service/CacheControlFilter.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2017, Adam <Adam@sigterm.info>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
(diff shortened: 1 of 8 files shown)


Adam-:
Cloudflare will not cache things unless the path ends in specific extensions - so this does not actually help us much as they are the only things providing cache for us.

Adam-:
I have looked into this fairly extensively and what you say is not how it works. You should try it and see.

KGoodale13:
I use Cloudflare on a different project I work for and it works exactly like that. We don't need extensions or page rules on the public API endpoints. Just cache control headers. 

It is currently working for you as well. For example, on the /worlds endpoint you set the cache control header to 10 minutes and public.
`cacheControl(CacheControl.maxAge(10, TimeUnit.MINUTES).cachePublic())`

When I make a request to /worlds (not /worlds.js) I hit the cache and receive the expected max-age of 600 seconds.
https://gyazo.com/6b6a0f1a4f1a88644ac66d1245ddaeea

Adam-:
I have a page rule to cache /worlds. I just disabled it, and you can see now it is no longer caching it.

KGoodale13:
@Adam- I added a controller advisor now that will set the cache-control to no-cache on any response that does not already have a cache-control header set. With this, you can use a single page rule for `api.runelite.net/*` with `CacheEverything` and `Origin Cache-Control` on and it will respect any cache control headers you set

PR #6919, opened 2018-12-13
Fix Item Stats overlay with null equipement container

TheStonedTurtle:
Fix the Item Stats overlay null pointer exception when item container is null.

Before:

After:


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -188,7 +189,8 @@ private String buildStatBonusString(ItemStats s)
 
 		if (s.isEquipable() && currentEquipment != null)
 		{
-			final Item[] items = client.getItemContainer(InventoryID.EQUIPMENT).getItems();
+			ItemContainer c = client.getItemContainer(InventoryID.EQUIPMENT);
+			final Item[] items = c == null ? new Item[0] : c.getItems();
```
deathbeam: Just include this in if? making empty array is diry hack

TheStonedTurtle:
Changed as requested

PR #6925, opened 2018-12-13
Have the equipment stats be off by default

MagicfTail:
This overlay most likely won't be used by most people and should, therefore, be off by default.


deathbeam:
I disagree, this is very very helpful for new players, and people can just disable it when they want (and we had food stats enabled by default too before, this is basically same thing)

PR #6927, opened 2018-12-13
Change  the vengeance indicators to use varbits

MagicfTail:
This pr

- Makes the indicators way more reliable, especially veng other
- Make veng and veng other use the same indicator since it's the same cooldown
- Make it so the indicator is removed if the config is switched off


runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -421,7 +421,17 @@
...
+
+	/**
+	 * Vengeance is active
+	 */
+	VENGEANCE_ACTIVE(2450),
+
+	/**
+	 * Spell cooldowns
+	 */
+	VENGEANCE_VENGEANCE_OTHER_COOLDOWN(2451);
```
Adam-: If the cooldowns are shared why is this called `VENGEANCE_VENGEANCE_OTHER_COOLDOWN`?

Nightfirecat: Because this affects both Vengeance and Vengeance Other? What are you expecting it to be called?

MagicfTail: As Nightfire said. I could name it `VENGEANCE_AND_VENGEANCE_OTHER_COOLDOWN`to make it even more clear, but that feels a bit excessive?

MagicfTail:
To clarify why the previous implementation wasn't very consistent.
Getting venge other'ed didn't trigger the venge active indicator as normal veng and veng other uses different graphics, but even then, the graphic would often be overwritten.
Using veng other also often doesn't trigger the veng cooldown timer as you aren't neccesarily interacting with the other player if you are moving while using the psell.

PR #6941, opened 2018-12-14
Fix party hat buy limits

### LINKED ISSUE #6940: Incorrect GE limit for partyhats
All partyhats have a trade limit of 5, not 15. The client shows 15 for every color.

TheStonedTurtle:
Fixes #6940


deathbeam:
Can you also update them on wiki? Or, do that first

deathbeam:
https://github.com/deathbeam/osrs-data because I have this for dumping the GE data and afaik all data on wiki are already updated to what we have in client, so we should keep doing that

TheStonedTurtle:
So if I edit the wiki that repo will automatically update the files for next release?

deathbeam:
No, but I would like to have wiki in sync so we could use that in future

TheStonedTurtle:
Looks like the wiki pages for the individual hats already has the correct buy limits.

https://oldschool.runescape.wiki/w/Partyhat

deathbeam:
Awesome

PR #6944, opened 2018-12-14
Stop chat notifications when a player logs in or hops

15987632:
Currently you receive notifications for any message that would give you a notification that is in your chat history each time you hop. This makes it not do that.


15987632:
This is WIP currently

15987632:
I thought I already closed this for being stupid

PR #6961, opened 2018-12-15
Dismiss Option

Dava96:
Added to the menu entry swapper. The Dismiss option is now the top option when the Dismiss option is checked (off by default). Applies to every random event other than the Dunce and Genie (xp giving events).

### DIFF
--- a/runelite-client/pom.xml
+++ b/runelite-client/pom.xml
@@ -334,7 +334,7 @@
 							</filters>
 							<transformers>
 								<transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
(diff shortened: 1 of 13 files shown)


TheStonedTurtle:
I went ahead and made the PR for you with a cleaner commit history since it seems like you are having some issues with this.

#6962 

For future reference you should have just updated your original PR instead of closing and creating new ones.

Dava96:
> I went ahead and made the PR for you with a cleaner commit history since it seems like you are having some issues with this.
> 
> #6962
> 
> For future reference you should have just updated your original PR instead of closing and creating new ones.

Apologies, I'm new to using github. Thank you though!

15987632:
this should be closed

PR #6962, opened 2018-12-15
Add dismiss swap to menu entry swapper

TheStonedTurtle:
ignores Dunce & Genie since they reward xp

Co-Authored-By: Dava <dava96@users.noreply.github.com>

Added to the menu entry swapper. The Dismiss option is now the top option when the Dismiss option is checked (off by default). Applies to every random event other than the Dunce and Genie (xp giving events).


Dava96:
Thanks for committing this for me!

formatme:
!!

PR #6963, opened 2018-12-15
Add missing clue requirement "Salute in warriors guild"

Monsterxsync:
Add missing ammo requirement for the clue step "Salute in the Warriors' guild bank."


Adam-:
There is at least 1 other instance of this same issue in this file.

Adam-:
The reason I pointed this out was I was hoping it would be addressed in this PR @deathbeam

deathbeam:
Well we do not need to do everything at once.

PR #6970, opened 2018-12-15
Fix the description of the wintertodt plugin and add boss tag

MagicfTail:
🙂 

https://oldschool.runescape.wiki/w/Wintertodt


JackCopland:
My mistake. You should also remove the "minigame" tag and add a "boss" tag to the tags list in that case.

MagicfTail:
Nice catch, I won't remove the minigame tag as other people might also consider it a minigame.

PR #6991, opened 2018-12-17
Instance reload helper

lyzrds:
This plugin is used for reloading chambers of xeric instances to find out more info before starting the raid.
Example of use, https://gfycat.com/PinkSoggyAmericangoldfinch


Abextm:
Jagex has said on the Q&A in the past that they would not add a way to hop to your current world, and has put measures in place to prevent it. Adding it is not something we want to add ourselves.

MagicfTail:
This should be marked as `wontfix`

PR #6997, opened 2018-12-17
Add dropped items to your ground item loot

15987632:
Would probably be more accurate with a DroppedItem event with the dropped item sounds.


15987632:
This is probably done in a stupid way, I don't think there is a good way other than listening for the drop sound which isn't possible currently (I think).

PR #7000, opened 2018-12-17
Cooking plugin

Cronick:
As talked about on discord. I have tried to create a basic cooking plugin.
Basic overlay, and there is also burn procentage for "Burnt" food.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingOverlay.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2018, Joris K <kjorisje@gmail.com>
+ * Copyright (c) 2018, Lasse <cronick@zytex.dk>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingOverlay.java
```diff
@@ -0,0 +1,111 @@
...
+import net.runelite.client.ui.overlay.components.LineComponent;
+import net.runelite.client.ui.overlay.components.PanelComponent;
+import net.runelite.client.ui.overlay.components.TitleComponent;
+
+import static net.runelite.api.AnimationID.COOKING_FIRE;
+import static net.runelite.api.AnimationID.COOKING_RANGE;
+
+class CookingOverlay extends Overlay
+{
+    private static final int[] animationIds =
```
raiyni: This should be a hashset

runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingOverlay.java
```diff
@@ -0,0 +1,111 @@
...
+		{
+			panelComponent.getChildren().add(TitleComponent.builder()
+				.text("NOT cooking")
+				.color(Color.RED)
+				.build());
+		}
+
+		panelComponent.getChildren().add(LineComponent.builder()
+			.left("Cooked:")
+			.right(session.getCookAmount() + (session.getCookAmount() >= 1 ? " (" + xpTrackerService.getActionsHr(Skill.COOKING) + "/hr)" : ""))
```
Adam-: This would throw whenever you aren't cooking.

Cronick: Yeah. I forgot about that part, and I can see that you have provided a better solution for a good amount of the code, so should have been fixed 👍

Adam-:
This is still broken when cooking on fires since the animation isn't long enough it alternates between cooking and not cooking.

Cronick:
> This is still broken when cooking on fires since the animation isn't long enough it alternates between cooking and not cooking.

Added to the if statement for isCooking() a delay checking, so if there is over 3 seconds between each animation it will say "Not cooking"

Took a look at client.getLocalPlayer().getAnimation(), and it keeps saying going to IDLE (-1) between the animation for some weird reason randomly, and it varies in length. So I know it's not the most beautiful solution on the issue.

Issue #6089, opened 2018-10-19
Wilderness K/D widget takes up space even when hidden


Nightfirecat:
**Describe the bug**
When entering the wilderness, the (now layoutable) kill-death ratio widget takes up space on snap points even when it is not set to display ingame.

**To Reproduce**
1. Turn off k/d display at Edgeville
2. Enter the wilderness
3. Observe issue

**Expected behavior**
This widget should not consume space on snap points when not shown.

**Screenshots**

![2018-10-19_12-28-57](https://user-images.githubusercontent.com/2199511/47243777-b5be0900-d3e2-11e8-866a-d068d6e30984.png)

PR #7003, opened 2018-12-18
widget overlay: Fix wilderness KD empty box overlay

Nightfirecat:
The wilderness KD widget, when disabling the display, will contain only
a single hidden static widget child. By checking for that case, we can
hide the parent widget if that case is met. For other movable widgets
which may have a single static child, that widget would not be hidden,
and would thusly not erroneously hide the parent widget.

Fixes runelite/runelite#6089


Adam-:
What about changing the layoutable widget to be the one that is hidden?

Nightfirecat:
This has been updated to use an optional array of dependent widgets (at least one of which must be visible) to properly hide the K/D widget when the child text widget is not visible.

Nightfirecat:
Well, this approach doesn't work anymore since they've changed how this widget works. Previously, the widget was a container (which was never hidden) which conditionally contained the static child widget which actually had the K/D text. When you disabled the K/D display, the child widget was not added, but the parent remained and was not hidden.

Now, the K/D widget container always contains a single dynamic widget, which has either the K/D text, or an empty string for text. I'll look at writing a different solution for this, but this PR is effectively dead in the water since it only applied to this widget, and doesn't work for it at all anymore.

Issue #3927, opened 2018-06-21
If using OSB prices, have all GE price sources use them


Jalopyy:
Currently I think OSB price only shows if you're actually at the GE. Having it so you could see the actively traded price at all places if the option is enabled (!price lookup, GE plugin, examine values, etc) would be nice.

deathbeam:
I disagree, that would make everything too tied to OSB prices when enabled. Also, the only important place where exact price is needed is in GE interface for flipping, everywhere else it is irrelevant.

LeviSchuck:
I second dearhbeam’s opinion.

Jalopyy:
On those terms I'd argue the GE lookup portion could equally use it, but fair enough.

damnnitbobby:
I agree with Jalopyy, if you hear of a sudden steep price rise, you need to tele to the ge to get an accurate price

deathbeam:
Well, for the GE sidebar it could be fine I guess. But also, you would need to visit GE anyway if you would want to buy/sell the item, so you would end up there one way or another.

Jalopyy:
That's true, but if you're waiting for an item to drop in value before picking it up, or rise to sell, being able to track it in realtime via the GE plugin would definitely be useful. 99% of people who look up items in the sidebar just want to see the price of it, which the OSB data giving them a more accurate value would definitely be better.

jgkennedy:
There are a number of items where the GE price is quite different than the price it actually trades for. Also, prices tend to swing a good bit during the day.

deathbeam:
As I stated in my original comment

PR #7009, opened 2018-12-18
Add OSBuddy Active Price to GE Item Panel (in Search Panel)

Berry-95:
Add lookup of OSBuddy price in Search Panel
Utilises current 'Enable OSB Prices' config

Solves not being able to look up OSBuddy Prices on the fly in the GE Plugin Panel (currently have to be in the GE in-game interface), as mentioned in: #3927


raiyni:
Wiki prices are now used for active trading

PR #7010, opened 2018-12-18
Send config updates on close

raiyni:
Since the change to batch post config updates was merged, config changes that happen between last post and client closed aren't synced to the server.


Adam-:
The config client actually enqueues the requests on the okhttp thread so I'm not sure this guarantees they are submitted. Does this work?

raiyni:
On graceful swing exits, yes

PR #7023, opened 2018-12-20
Add Flax and Bonemeal to Daily Task plugin

Hydrox6:
Please do let me know if either of these should be off by default, or if anyone has suggestions on the message wording.


runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksPlugin.java
```diff
@@ -220,6 +234,48 @@ else if (dailyReset)
...
+			}
+		}
+	}
+
+	private void checkBonemeal(boolean dailyReset)
+	{
+		if (client.getVar(Varbits.DIARY_MORYTANIA_MEDIUM) == 1)
+		{
+			int collected = client.getVar(Varbits.DAILY_BONEMEAL_STATE);
+			//The varbit tracks how much bonemeal has been redeemed from Robin
```
Nightfirecat: Comments on how the varbit work should probably be on the definition, not on the usage of it.

PR #7032, opened 2018-12-20
Make slayer weakness overlay work for the npc you are interacting with

15987632:
Title, it used to work for any npc on your screen that your slayer task included, now it only works for the npc you are interacting with which allows it to work offtask.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -663,6 +663,25 @@ public class SlayerPlugin extends Plugin
 		counter = null;
 	}
 
+	public boolean isInTask(Task task, NPC npc)
+	{
+		String npcName = npc.getName().toLowerCase();
+
+		if (npcName.contains(task.getName().toLowerCase().replaceAll("s$", "")))
+		{
+			return true;
+		}
+
+		for (String taskName : task.getTargetNames())
+		{
+			if (npcName.contains(taskName.toLowerCase()))
+			{
+				return true;
+			}
+		}
+		return false;
+	}
+
 	//Utils
 	private String capsString(String str)
 	{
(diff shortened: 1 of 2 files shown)


deathbeam:
If this is supposed to work off-task, then why should it be in slayer plugin? Also, looping all tasks, all the time, and all target names inside seems suboptimal.

15987632:
I can move the logic into the Interacting changed event which seems a lot better. And it should stay in the slayer plugin because its for slayer specific monsters? I could make it it's own plugin or move it if you have a better idea.

deathbeam:
Interactingchanged sounds good. And that makes sense, they are all slayer monsters

15987632:
I moved the main of the logic into InteractingChanged, now the logic left in the overlay is less than it was when it was merged.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -173,6 +180,9 @@
 	@Getter(AccessLevel.PACKAGE)
 	private int points;
 
+	@Getter(AccessLevel.PACKAGE)
+	private Task weaknessTask = null;
```
Adam-: Probably the slayer plugin should always hold a `Task` for the current task instead of specific to the weakness one.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -491,6 +501,29 @@ public void onExperienceChanged(ExperienceChanged event)
...
+		{
+			weaknessTask = null;
+			return;
+		}
+
+		final NPC npc = (NPC) interacting;
+
+		for (Task task : weaknessTasks)
+		{
+			if (isInTask(task, npc))
```
Adam-: You should just be able to `isTarget` for this?

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
15987632: This variable is what task the npc the players is interacting with belongs to and if it doesnt belong to one of the tasks that show a weakness overlay it returns null.

PR #7033, opened 2018-12-20
World-Specific Highscores lookup

### LINKED ISSUE #7001: World-Specific Hi-Score Lookup
frustrated when [...]

When on SDMM worlds, I right click lookup a player, and it shows me their regular account, and I have to click manually each time to switch to SDMM.

Either

a. Have it determine which hi-score lookup to use depending on your current world.

or 

b. Have it remember your previous selection, in other words, if I currently have the SDMM section clicked in the client, when I right click lookup it will lookup in that section.

Digbigpig:
Closes #7001 
The right-click -> lookup option checks if the current world is a DMM world and will display the corresponding high scores tab instead of defaulting to the normal high scores.

3e9629fb3d214a3ba9ffd8ca04a0822e

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
@@ -38,6 +38,7 @@ import java.awt.event.MouseEvent;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import javax.annotation.Nullable;
@@ -50,6 +51,7 @@ import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.Experience;
 import net.runelite.api.Player;
+import net.runelite.api.WorldType;
 import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.ui.PluginPanel;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -634,5 +636,31 @@ private void resetEndpoints()
 	{
 		// Select the first tab (NORMAL hiscores)
 		tabGroup.select(tabGroup.getTab(0));
```
rbbi: This logic could be moved inside your new method

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -634,5 +636,31 @@ private void resetEndpoints()
...
+
+	private void selectWorldEndpoint()
+	{
+		if (client != null)
+		{
+			EnumSet<WorldType> wTypes = client.getWorldType();
+
+			if (wTypes.contains(WorldType.DEADMAN_TOURNAMENT))
+			{
+				tabGroup.select(tabGroup.getTab(6));
```
rbbi: Using constants instead of arbitrary numbers would improve readability
`tabGroup.getTab(6)`

Nightfirecat: I think this can use the enum ordinals? eg. `WorldType.DEADMAN_TOURNAMENT.ordinal()`

rbbi: I don't know about WorldType ordinals.. perhaps HiscoreEndpoint ordinals since they're added that way via HiscoreEndpoint.values()

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -634,5 +636,29 @@ private void resetEndpoints()
 	{
 		// Select the first tab (NORMAL hiscores)
 		tabGroup.select(tabGroup.getTab(0));
+
+		// Selects the DMM tab if the current world is DMM.
+		selectWorldEndpoint();
+	}
+
+	private void selectWorldEndpoint()
```
Adam-: This should return a `HiscoreEndpoint` and not have side effects.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -634,5 +636,29 @@ private void resetEndpoints()
 	{
 		// Select the first tab (NORMAL hiscores)
 		tabGroup.select(tabGroup.getTab(0));
```
Adam-: `tabGroup.select(selectWorldEndPoint());`

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
Digbigpig: Changes have been commited

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -632,7 +634,29 @@ private static String sanitize(String lookup)
 
 	private void resetEndpoints()
 	{
-		// Select the first tab (NORMAL hiscores)
-		tabGroup.select(tabGroup.getTab(0));
+		// Selects the DMM tab if the current world is DMM.
```
Adam-: "Select the correct tab based on the world type"

Digbigpig:
Changes have been made. All the forced pushes were to correct a weird issue i'm having where commits from the upstream master are being added to my branch.

PR #7045, opened 2018-12-21
Ground Items: Retrieve ItemLayer height in render rather than on spawn

### LINKED ISSUE #5330: Item tooltip appears 1 block too high after killing unicorn and going through a fairy ring and back
image
I killed 2 unicorns, then went through the fairy ring and bcak and the left unicorn horn tooltip was 1 block higher than it should be. I tried several times, and it only appears to be happening when I kill a unicorn that drops loot on that specific spot. Just to be clear, ,before I went though the fairy ring, the tooltips were at a correct height.

### LINKED ISSUE #6849: Ground Item Label shows 1 to high on Z axis
If you have ground items on and you kill a monster and then climb up a level before the drop. When you go back down the label will be one level up like it was trying to label it up one on the Z cord. This isn't that big of an issue just thought I would post. I did this specifically at greater demons in catacombs. Using the rope to climb up before the demon death animation completed.

trimbe:
After a scene load, the `ItemLayer`'s height isn't updated at the time of the item spawn event. This causes the height to be incorrect if the height for that scene tile was different in the previous scene. 

Closes #5330, Closes #6849

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -41,6 +41,7 @@ import net.runelite.api.Client;
 import net.runelite.api.Perspective;
 import net.runelite.api.Player;
 import net.runelite.api.Point;
+import net.runelite.api.Tile;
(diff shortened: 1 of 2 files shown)


Abextm:
`ItemLayer::height` is only set once, immediately before it is assigned and the FieldHook is called.

PR #7054, opened 2018-12-22
barrows plugin: add a toggle for overlay

### LINKED ISSUE #7046: [Suggestion] Add a toggle for the Barrows info box (vanilla one is better imo)
https://i.gyazo.com/75eca0740e8305839ecb520049d43db0.png

jwngo:
Closes #7046

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -44,6 +44,9 @@ public class BarrowsBrotherSlainOverlay extends Overlay
 	private final PanelComponent panelComponent = new PanelComponent();
 
 	@Inject
+	private BarrowsConfig config;
+
+	@Inject
 	private BarrowsBrotherSlainOverlay(Client client)
 	{
 		setPosition(OverlayPosition.TOP_LEFT);
@@ -60,7 +63,6 @@ public class BarrowsBrotherSlainOverlay extends Overlay
 		{
 			return null;
 		}
-
 		// Hide original overlay
 		final Widget barrowsBrothers = client.getWidget(WidgetInfo.BARROWS_BROTHERS);
 		if (barrowsBrothers != null)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
```diff
@@ -54,41 +57,44 @@ private BarrowsBrotherSlainOverlay(Client client)
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		// Do not display overlay if potential is null/hidden
-		final Widget potential = client.getWidget(WidgetInfo.BARROWS_POTENTIAL);
-		if (potential == null || potential.isHidden())
+		if (config.showBarrowsOverlay())
```
Nightfirecat: This should be a guard clause, not a large nested if statement.

jwngo: Changed it to guard clause.
Not sure whether the default barrows overlay can become movable, just adding toggle as requested by #7046.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
```diff
@@ -89,6 +91,10 @@ public Dimension render(Graphics2D graphics)
 				.rightColor(rewardPercent >= 73.0f && rewardPercent <= 88.0f ? Color.GREEN : rewardPercent < 65.6f ? Color.WHITE : Color.YELLOW)
 				.build());
 
-		return panelComponent.render(graphics);
+		// Toggle for barrows overlay
+		if (config.showBarrowsOverlay())
```
Nightfirecat: This isn't the right way to do a guard clause; we shouldn't even be running the code that updates `panelComponent` if `config.showBarrowsOverlay()` is false.

Adam-:
I think we were discussing just using the in-game interface and making it movable, which is preferable.

PR #7065, opened 2018-12-23
Update skill calculator fields automatically

### LINKED ISSUE #6667: When filter applied to stats calculator and skill button is clicked, keep filter applied
## Overview

When skilling, you can see how many actions to the next level by clicking the calculator icon, then a skill icon. From within a skill icon's menu, you can filter the actions you can do by searching, eg "alch" in the Magic calculator. This number does not update live, so to update it, you click the same skill icon again. This keeps the filter text, but removes the actual filtering, so all the items come back into the list of actions. You have to click the filter text and hit enter to reapply the filter. I believe this is a bug

## Version information

- RuneLite version 1.5.1
- OSRS Rev177
- Windows 10 Version 1803 (Build 17134.407)
- Java 8 Update 131

## Location

I don't exactly know where the filter events happen, but it seems to be related to here: https://github.com/runelite/runelite/tree/master/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator

Specifically this commit: https://github.com/runelite/runelite/commit/8d8c8e475421e8528c32d6ac8fd6451372999f6d

rbbi:
Skill calculator fields (total XP, level) aren't kept up-to-date after they have been set.

~~Instead, use the Scheduler to schedule an update job at a defined interval (2sec)~~
Instead, use the EventBus to listen for ExperienceChanged events and update where appropriate

Fixes #6667

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
@@ -66,6 +66,8 @@ class PrayerDoseOverlay extends Overlay
 	@Setter(AccessLevel.PACKAGE)
 	private boolean hasPrayerRestore;
 	@Setter(AccessLevel.PACKAGE)
+	private int bonusPrayer;
+	@Setter(AccessLevel.PACKAGE)
 	private boolean hasHolyWrench;
 
 	@Inject
@@ -135,14 +137,9 @@ class PrayerDoseOverlay extends Overlay
 		final double dosePercentage = hasHolyWrench ? .27 : .25;
 		final int basePointsRestored = (int) Math.floor(maxPrayer * dosePercentage);
 
-		// how many points a prayer and super restore will heal
-		final int prayerPotionPointsRestored = basePointsRestored + 7;
-		final int superRestorePointsRestored = basePointsRestored + 8;
+		final int pointsRestored = basePointsRestored + 7 + bonusPrayer;
 
-		final boolean usePrayerPotion = prayerPointsMissing >= prayerPotionPointsRestored;
-		final boolean useSuperRestore = prayerPointsMissing >= superRestorePointsRestored;
-
-		if (!usePrayerPotion && !useSuperRestore)
+		if (prayerPointsMissing < pointsRestored)
 		{
 			return null;
 		}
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -406,4 +415,23 @@ private boolean slotContainsText(UIActionSlot slot, String text)
...
+	 * Updates the current skill calculator (if present) by looking up
+	 * the player's current XP and level for that skill
+	 * <p>
+	 * This method is invoked at a defined internal,
+	 * see: {@link SkillCalculatorPlugin#updateSkillCalculator()}
+	 */
+	void updateSkillCalculator()
+	{
+		// If the user has selected a calculator, update its fields
+		Optional.ofNullable(calculatorType).ifPresent(calc ->
```
Adam-: this is just `if (calc != null) {`

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculatorPlugin.java
```diff
@@ -82,4 +85,10 @@ protected void shutDown() throws Exception
 	{
 		clientToolbar.removeNavigation(uiNavigationButton);
 	}
+
+	@Schedule(period = 2, unit = ChronoUnit.SECONDS)
+	public void updateSkillCalculator()
```
Adam-: There is an xp changed event which might be better here.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
rbbi: I liked the readability of ifPresent

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculatorPlugin.java
rbbi: Thanks - i'll take a look

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
deathbeam: Still it should be changed to null check, because you are not doing anything with the Optional after

rbbi: Done

deathbeam:
Looks like your rebase went horribly wrong

rbbi:
@deathbeam @Adam- any update on this please

PR #7068, opened 2018-12-23
Display persistent data inside Loot Tracker UI

TheStonedTurtle:
Record quantity is limited to the default return value of the API endpoint

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
@@ -48,4 +48,14 @@ public interface LootTrackerConfig extends Config
 		description = ""
 	)
 	void setIgnoredItems(String key);
+
+	@ConfigItem(
+		keyName = "saveLoot",
+		name = "Save loot",
+		description = "Save loot between client sessions (requires being logged in)"
+	)
+	default boolean saveLoot()
+	{
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -29,17 +29,21 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.swing.SwingUtilities;
+import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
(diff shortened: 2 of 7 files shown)


TheStonedTurtle:
#7071 fixed the API data issue

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -32,10 +32,33 @@
 @ConfigGroup("loottracker")
 public interface LootTrackerConfig extends Config
 {
+	@ConfigItem(
```
deathbeam: I dont think this needs a toggle?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -32,10 +32,33 @@
...
+		name = "Persistent data",
+		description = "Configures whether loot tracker data should persist between sessions.<br/> Must be Logged in.",
+		position = 0
+	)
+	default boolean persistentData()
+	{
+		return false;
+	}
+
+	@ConfigItem(
```
deathbeam: This should be enabled by default, and it should mention that it requires login in name (e.g Historical data (Requires login!)

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -209,6 +213,34 @@ protected void startUp() throws Exception
...
+				{
+					switch (client.getGameState())
+					{
+						case STARTING:
+						case UNKNOWN:
+							return false;
+					}
+
+					// Run blocking web requests on new thread
+					new Thread(() ->
```
deathbeam: This should use executor service, and the entire thing should be limited to maybe 100-1000 loot records

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
TheStonedTurtle: Some people may not want to persist data between sessions so i figured why store their data for no reason? I can revert this config change if you'd like.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
TheStonedTurtle: The get API route limit defaults to `1024` already so should be fine as is.

I've also updated the `new Thread()` to use `ScheduledExecutorService.submit()` instead

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
deathbeam: Well, we don't have toggle in master, so why add it now. And for people that doont want to see the persisted data, they will just disable the other toggle

TheStonedTurtle: The changes in master were meant to just test the API performance so we didn't want to announce it when we were still testing the feasibility.

Either way I've removed that config option as requested.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -392,4 +427,31 @@ boolean isIgnored(String name)
...
+
+	private Collection<LootTrackerRecord> convertToLootTrackerRecord(final Collection<LootRecord> records)
+	{
+		Collection<LootTrackerRecord> trackerRecords = new ArrayList<>();
+		for (LootRecord record : records)
+		{
+			// Convert GameItem drops into LootTrackerItems
+			LootTrackerItem[] drops = record.getDrops().stream().map(itemStack ->
+			{
+				final ItemComposition itemComposition = itemManager.getItemComposition(itemStack.getId());
```
Adam-: Why is all of this logic duplicated from above?

TheStonedTurtle: No idea, should have separated it into a single function that is used in both places. I'll do so now unless you are working on this.

PR #7075, opened 2018-12-24
Add NPC list to keep menu swapping swapping even if they have a hint arrow on them

### LINKED ISSUE #7074: The Port Sarim monks shouldn't have the menu entry swapped on the master clue
If you're on the "anger the priests" step, it switches the switched menu entry swapperooing back to its original state, but the clue specifically wants you to travel, not talk.

Could this one be added as an edge case?

15987632:
Closes #7074


15987632:
This should probably not be done this way

PR #7102, opened 2018-12-29
ChatMessageManager: Don't set timestamp if it's default

MagicfTail:
Fixes the twitch plugin showing wrong timestamps


Nightfirecat:
Wouldn't the better solution be to set the message's timestamp from the twitch plugin?

MagicfTail:
Then every plugin that creates quedMessages would have to set the timestamp instead of the default just being the current time.

Issue #2513, opened 2018-05-09
Add chompy bird undertile color


Fox4148:
Visual Improvement while killing chompy birds.
Like the slayer under tile color, but for chompy.

Coollegolas:
This appears to be addressed in https://github.com/runelite/runelite/pull/2309

Nightfirecat:
Can't this be accomplished by just using the NPC Indicators plugin?

Spedwards:
So that PR was closed, but it seems to have addressed this. Came here hoping something was in the works.

@Nightfirecat It can to a degree. There's no way to differentiate between alive and dead chompy birds so when you have a lot, it can be rather confusing.

Here's just one example of that: ![Example](https://i.imgur.com/fG5FeOv.png)

SJacobs501:
@Spedwards I was literally just hunting Chompy birds an hour ago and fixed this issue while I was at it. I made a pull request but you can check my commit on how I fixed it.

PR #7118, opened 2018-12-31
Chompy bird npchighlight fix

SJacobs501:
This fixes issue #2513

Dead Chompy birds won't be detected by the npc highlight plugin anymore.


runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -41,12 +41,7 @@
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.Client;
-import net.runelite.api.GameState;
-import net.runelite.api.GraphicID;
-import net.runelite.api.GraphicsObject;
-import net.runelite.api.MenuAction;
-import net.runelite.api.NPC;
+import net.runelite.api.*;
```
Spedwards: Use more specific imports here.

SJacobs501: Except you're making it less specific. Lol.

SJacobs501: Nevermind, I was on phone and I didn't realize this was my commit. I couldn't recall this change but I just found out IntelliJ turns imports into wildcard automatically once they reach a certain number (I know because after adding all the imports specifically it removed them all and turned them into a wildcard again). I pushed the change.
There are 16 specific imports now, is it really bad practice to import wildcard in this case? Should wildcard never be used?

Spedwards: There isn't any problem with using it in this project, but it's a good practice not to use wildcard imports in general.

A brief article on why it's better to use explicit imports: http://www.javadude.com/articles/importondemandisevil.html

Nightfirecat:
Rather than add a special case, it would make more sense for live chompy birds to be highlighted by a hub plugin or via some other method. Anyway, this PR is very old and stale, so I am closing it.

Issue #1412, opened 2018-04-12
Shift click walk here


kulershk:
Would be nice if you could use shift click to walk under npc

SoyChai:
Clever idea, I would've never thought of this.

We could make this even more universal by allowing shift-clicks to interact with objects behind NPCs too.

kulershk:
For bossing it would be really helpful

Fryslan:
I would realy like this feature

beshj:
please add this it is literally the most needed feature not in the game....

deathbeam:
Based on discord discussion I don't think we want to implement this

PR #7127, opened 2018-12-31
Feature/shift walker plugin

plinko60:
Feature requested in https://github.com/runelite/runelite/issues/1412 .

This plugin was built based off the work in the MenuEntrySwapperPlugin . This works very similar to how that plugin works, if this is more suited to extend the MenuEntrySwapperPlugin that can be done, however more extensive changes are needed to accommodate multiple entries being tied to the shift hotkey (MenuEntry will currently only handle 1).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/shiftwalker/ShiftWalkerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/shiftwalker/ShiftWalkerPlugin.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright (c) 2018, Plinko60
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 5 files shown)


plinko60:
Current features:
**Walk Under Everything** - Enable this option when you do not want to interact with anything while Shift is pressed. If Walk Here is an option it will be the action taken.

**Walk Under Box Traps** - Press "Shift" to be able to walk under instead of picking up a Box Trap.

**Walk Under Attack Options** - Press "Shift" to be able to walk instead of attacking. Make sure Left Click Attack is on.

plinko60:
So there was conversation on discord about this being potentially too strong for PvP situations. (Such as barrage and walk under).

If requested, I can look into making this not work for Pvp. The change would be to disable this walk under when you have an attack option on the player.

runelite-client/src/main/java/net/runelite/client/plugins/shiftwalker/ShiftWalkerPlugin.java
```diff
@@ -0,0 +1,184 @@
...
+	 * Finds the index of the menu that contains the verbiage we are looking for.
+	 * @param pMenuEntries The list of {@link MenuEntry}s.
+	 * @param pMenuEntryToSearchFor The Option in the menu to search for.
+	 * @return The index location or null if it was not found.
+	 */
+	private Integer searchIndex(MenuEntry[] pMenuEntries, String pMenuEntryToSearchFor)
+	{
+		Integer indexLocation = 0;
+
+		for (MenuEntry menuEntry : pMenuEntries)
```
plinko60: This needs to be a search in the reverse order than this. Searching like this only does the top level entry. But multiple entries in the menu won't get touched, meaning this won't work then.

deathbeam:
See reply in #1412

PR #7129, opened 2018-12-31
Bank Overlay for most valuable items

Jake-Watkins:
Built off the bankvalue plugin.
Highlights the most valuable items in the bank, can be configured in the bankvalue config settings. 
Also reduces function calls in the bankvalue plugin.
Added 3 config options
**Show Valuable Items**
**High Value Colour**
**Number of Items to Highlight**

I don't think this is something that resolves any feature requests but it's something that I really wanted for the purposes of clearing out loot tabs.


rbbi:
Hello,

Something strange happened when I opened a tag tab, see below:

image

Jake-Watkins:
Oh interesting, I'm pretty sure this is because the calculations are done at the same time as the bank value calcs, I'll try to update it to check on search queries as well. Nice catch! Thanks!

Jake-Watkins:
It should be fixed now, we now check if the item is hidden, since the search window creates extra copies of items. also, we now update during the search window if the option is enabled in the config

raiyni:
This was added as an external plugin `Bank Heatmap`

PR #7131, opened 2018-12-31
Add missing sanfew serums to the prayer dose indicator.

Monsterxsync:
just adds sanfew serums to prayer dose indicator.


deathbeam:
Now that I am looking at this, why exactly there are 3 separate booleans for same thing

Monsterxsync:
god knows i was thinking the same thing lol

PR #7132, opened 2019-01-01
Add config item nesting and hiding/unhiding of config items

Craftiii4:
Allow config items to have a parent, causing the child to be under the parent and slightly indexed to the right. Can be stacked to allow children to have children.

Allow config items to be hidden by a boolean config item by setting the unhide option;

Example:
https://i.gyazo.com/98b4a5a78689d88be21dc9f568511d8d.mp4

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigPanelItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigPanelItem.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, Adam <Adam@sigterm.info>
+ * Copyright (c) 2018, Craftiii4 <craftiii4@gmail.com>
  * All rights reserved.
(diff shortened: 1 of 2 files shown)


deathbeam:
I don't see why is this better than simply adding headers and groups. And we don't really have use-case for having toggles for showing "more" config, but we have use-case for grouping stuff, what is kind of similar to this but also not really.

Craftiii4:
Ah groups already exist? I was using https://github.com/runelite/runelite/wiki/Creating-plugin-config-panels - guess I should have looked a bit more for them;

The toggling was because I would like the user to be able to disable/enable every ore type, but not clutter up the config with them

deathbeam:
No, we don't have groups yet, I am saying that groups will be better than this.

Craftiii4:
Ah okay, I'll close this then

PR #7145, opened 2019-01-01
gpu: Fix stretched mode with global premul alpha

Abextm:
This seems to work, but I can't figure out what the code I removed was supposed to do.


Abextm:
@dbolya If you would look at this that would be very helpful.

PR #7148, opened 2019-01-02
make the "Ctrl-V paste" option enabled by default

15987632:
:shirtrl:


deathbeam:
It is intentionally disabled by default in order to prevent people who are not aware of the feature to not accidentally paste their password to username field on stream etc

PR #7152, opened 2019-01-02
Add menu swapper for Altar of the Occult to swap to magic book of choice

Spedwards:
Many players use hosted occult altars to swap their magic book. Most of the time, it's back-and-forth between the standard book and one of the other books. With this extension, you'll be able to set which magic book they want to switch between.

This extension is disabled by default.

Example: https://gfycat.com/ThatDisastrousBobwhite


Spedwards:
I've added a gif to explain what this does since apparently I'm horrible at explaining with my words.

PR #7158, opened 2019-01-03
itemprices: Add "tooltips" plugin tag

Nightfirecat:


deathbeam:
On side-note, you mentioned in Discord that "tooltip" is not picking up stuff with "tooltips" tags correctly right? So I guess the tag search is not using contains or something?

PR #7161, opened 2019-01-03
Fix world map surface selector ID

Nightfirecat:


Adam-:
`eca3dd3e03ee9a793417bff7f071b6dd14b216be` seems unnecessary?

Nightfirecat:
It's kind of an artifact of me going down a rabbit hole of not realizing that `		WidgetInfo.WORLD_MAP_SURFACE_SELECTOR` had the wrong child ID and adding a third entry. It can be dropped if not wanted.

Nightfirecat:
Dropped that commit.

PR #7182, opened 2019-01-04
Add all known Kourend & Kebos diary requirements

Monsterxsync:
Adds all the currently known diary requirements will update this when the levels i'm missing are known to us, was also thinking about adding the house favour requirements to the plugin as there are many that require them.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/diaries/KourendDiaryRequirement.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/diaries/KourendDiaryRequirement.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2019 William <https://github.com/monsterxsync>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Unless we know the exact text ahead of time (I don't think we will?) I think this should probably wait to be released the week after, or sometime shortly after the game update as a point release perhaps.

Abextm:
They showed the complete list in game on stream, so I doubt the wording will change too much

Monsterxsync:
diary reqs image Heres an image with the requirements i doubt they would change the wording around. You never know though

runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/diaries/KourendDiaryRequirement.java
```diff
@@ -0,0 +1,119 @@
...
+		add("Create a Strength potion in the Lovakengi Pub.",
+			new SkillRequirement(Skill.HERBLORE, 12));
+		add("Fish a Trout from the River Molch.",
+			new SkillRequirement(Skill.FISHING, 20));
+
+		//MEDIUM
+		add("Travel to the Fairy Ring south of Mount Karuulm.",
+			new QuestRequirement(Quest.FAIRYTALE_II__CURE_A_QUEEN, true));
+		add("Use Kharedst's memoirs to teleport to all five cities in Great Kourend.",
+			new QuestRequirement(Quest.THE_DEPTHS_OF_DESPAIR, false),
```
Spedwards: Adding `false` to the quest requirements is not required and is redundant as it already defaults that way.

Abextm:
Added in e0534ebb22f3bf9120c89ba4d2fe86916b68a62a

PR #7187, opened 2019-01-04
 Loot Tracker - Add optional automatic local data output 

TheStonedTurtle:
Since data is only stored in the database for 30 days, and I personally want to keep my loot for excel spreadsheet fun, I added a config option that will automatically append `LootRecord`s as JSON to `.log` files located in `.runelite/loots/<ingame-username>/`

Output Format:
```JSON
{"eventId":"Man","type":"NPC","drops":[{"id":526,"qty":1},{"id":877,"qty":2}]}
```


TheStonedTurtle:
> Adam: if were going to provide another way to export loottracker data it will be from the website

PR #7190, opened 2019-01-04
Allow npcs to be excluded from the NPC Indicator plugin

### LINKED ISSUE #7147: Have a hidden NPC list for the NPC Indicator Plugin
When questing I like using the NPC Indicator plugin to have the names of the NPCs appear above their heads, so I use the "*" to highlight all the NPCs instead of typing in each NPC that I want highlighted. But it is kind of annoying that EVERYTHING is highlighted; things like butterflies, bankers, rats and so on.

If we could have an NPC Hidden list, similar to the one used in the Ground Items plugin, where you can put items on a list that you don't want to show up. This would be nice if it could be don't for NPCs to filter out the NPCs that you don't want highlighted.

An alternative is to just type in the names of the NPCs that you want highlighted, but I'd rather have all NPCs highlighted and then filter to hide NPCs that I don't want highlighted.

Craftiii4:
Resolves #7147

Allows NPCs to be excluded from the NPC indicator using a '-' at the start of their name. Wildcards are accepted. 
Also fixed a concurrent modification exception found when modifying the NPCs to highlight text field. 

image

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
@@ -116,7 +116,7 @@ public class NpcIndicatorsPlugin extends Plugin
 	 * NPCs to highlight
 	 */
 	@Getter(AccessLevel.PACKAGE)
-	private final Set<NPC> highlightedNpcs = new HashSet<>();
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
```diff
@@ -55,7 +55,18 @@ default String getNpcToHighlight()
 	}
 
 	@ConfigItem(
-		position = 2,
+			position = 2,
+			keyName = "hiddenNPCs",
+			name = "Excluded NPCs",
+			description = "List of NPC names to hide"
```
Spedwards: Indentation needs to be fixed here.

PR #7191, opened 2019-01-04
Gargoyle Hammer Toss Icon Fix

ftwfabian:
The Hammer throw for Gargoyles currently does not show an icon when the gargoyle is on 9 hp, which is the hp when you are able to throw the hammer to successfully kill the gargoyle


deathbeam:
Needs rebase

PR #7192, opened 2019-01-04
Add Quest Points chat command

rbbi:
Adds the QUEST_POINTS local player variable (id: 101) and !qp chat command to print the local player's quest points

qp-command


Hydrox6:
What's the point of this? It only gets the local player's quest points, so everyone will appear to have the same quest points as you do.

rbbi:
> What's the point of this? It only gets the local player's quest points, so everyone will appear to have the same quest points as you do.

Ah, thanks for letting me know - this shouldn't be the case! :)

Adam-:
Added in ecd0f7dce3180df996bdf1a3e11374d0f746f848

rbbi:
Thanks @Adam-

PR #1570, opened 2018-04-15
Show tooltips based on Runescape's advanced settings

jplsek:
If the advanced setting has the tooltip set to disabled, also disable RuneLite tooltips.

Closes #1143

~If preferred, I could also add a toggle to allow RuneLite tooltips anyways.~ Edit: Did that.

PR #7195, opened 2019-01-05
tooltips: show tooltip based on advanced settings

### LINKED ISSUE #1143: Advance Options (from in-game settings) Tooltips
If the option to have tooltips on certain options in advance options in game is turned off, runelites tooltips will not overlay on top of that. (aka so if i have prayer tooltip turned off in advanced options I dont get runelites tooltips if I have that setting turned on. 

Alternative have multiple tooltip options for different things. Couldn't tell if that part was posted before or not

jplsek:
If the advanced setting has the tooltip set to disabled, then also disable the RuneLite tooltips.

Closes #1143

(Revive of PR #1570)

Runescape and Runelite default:
image

Runescape disable tooltip and Runelite default:
image

Runescape disable tooltip and Runelite follow advanced tooltip settings:
image
(imagine there is a mouse over the bar ;) )

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
+++ b/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
@@ -39,6 +39,12 @@ public enum VarPlayer
 
 	BANK_TAB(115),
 
+	/**
+	 * Use prayer tooltip bit setting: 30
+	 * Use special attack tooltip bit setting: 31
+	 */
+	TOOLTIPS(284),
+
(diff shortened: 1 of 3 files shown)


Adam-:
Shouldn't this type of logic be in the tooltip overlay?

deathbeam:
It's correct to have it in Mouse Tooltips plugin @Adam- because it's related only to in-game mouse tooltips we showvand not stuff like item stats etc

PR #7200, opened 2019-01-05
Add Skybox plugin

### LINKED ISSUE #3941: Request: Skyboxes
Would it be possible to add a toggle that allows a skybox? Not asking for anything crazy, a simple shade of light blue would work for me.

Thanks,

Abextm:
Closes #3941
sky1
sky2
sky3

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
+++ b/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
@@ -46,6 +46,7 @@ import net.runelite.api.MainBufferProvider;
 import net.runelite.api.RenderOverview;
 import net.runelite.api.Renderable;
 import net.runelite.api.WorldMapManager;
+import net.runelite.api.events.BeforeRender;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.hooks.Callbacks;
 import net.runelite.api.hooks.DrawCallbacks;
@@ -81,6 +82,7 @@ public class Hooks implements Callbacks
 	private static final OverlayRenderer renderer = injector.getInstance(OverlayRenderer.class);
 
 	private static final GameTick GAME_TICK = new GameTick();
+	private static final BeforeRender BEFORE_RENDER = new BeforeRender();
 
 	@Inject
 	private EventBus eventBus;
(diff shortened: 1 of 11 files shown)


runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -149,6 +151,8 @@ public void clientMainLoop()
 			client.setTickCount(tick + 1);
 		}
 
+		eventBus.post(BEFORE_RENDER);
```
Adam-: Maybe this should be a 'main loop' event? Also should possibly be in a separate commit

Abextm: I feel like 'main loop` could be confused with a client tick event

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
```diff
@@ -0,0 +1,115 @@
...
+	{
+		cx -= client.getBaseX() / 8;
+		cy -= client.getBaseY() / 8;
+		try
+		{
+			return client.getInstanceTemplateChunks()[plane][cx][cy];
+		}
+		catch (ArrayIndexOutOfBoundsException e)
+		{
+			return -1;
```
Adam-: When does this happen? Generally I avoid having exceptions in the normal control flow. The eventbus should log it anyway.

devLotto:
Can we have a sand color skybox while in the desert? #c2b280

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+			px = p.getX();
+			py = p.getY();
+		}
+
+		// Inverse of camera location / 2
+		int spx = px - ((client.getCameraX() - px) >> 1);
+		int spy = py - ((client.getCameraY() - py) >> 1);
+
+		client.setSkyboxColor(skybox.getColorForPoint(
+			client.getBaseX() + (spx / 128.f),
```
Adam-: Should this be off of the player location? Otherwise the sky changes if you move the camera which is weird

devLotto: The sky currently changes when moving about using the Oculus orb, which I find nice.

PR #7212, opened 2019-01-07
Add anisotropic filtering

### LINKED ISSUE #7062: Mipmapping for GPU Plugin
Textures in RS look very poor when far away. Mip mapping would solve this issue. It's also very low overhead, as you're rendering smaller textures for farther away textures.
Example of the effects of mip mapping textures. 

image

pacf531:
For the most part, this was structured very similar to how anti-aliasing is handled inside the GPU plugin. Only thing that might be objectionable is the glGetFloatv and not putting a utility function for it inside the GLUtils class, and not checking for the extension explicitly itself since for all intents and purposes for desktop and laptop gaming, AMD, intel, and Nvidia have supported anisotropic filtering for a long time, predating the OpenGL 4.2 minimum support needed for the plugin itself. I also added bilinear and trilinear filtering in case for some lower end integrated graphics processors, bilinear with just enabling mipmapping and trilinear attempting to interpolate and blend the mipmaps but at a slight performance cost.

The main problem is that it doesn't seem to give a very appreciable image quality boost, although you can still see it slightly, most noticeably with the foliage. I am running an AMD Vega Frontier Edition for these images.

Disabled:

Anisotropic Filtering x16:

Fixes #7062, I would assume, with my bilinear option.


runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -990,6 +992,37 @@ public void draw()
...
+				{
+					case BILINEAR:
+						gl.glTexParameteri(gl.GL_TEXTURE_2D_ARRAY, gl.GL_TEXTURE_MIN_FILTER, gl.GL_LINEAR_MIPMAP_NEAREST);
+						break;
+					case TRILINEAR:
+						gl.glTexParameteri(gl.GL_TEXTURE_2D_ARRAY, gl.GL_TEXTURE_MIN_FILTER, gl.GL_LINEAR_MIPMAP_LINEAR);
+						break;
+					default:
+						final float maxSamples[] = new float[1];
+						gl.glGetFloatv(gl.GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, maxSamples, 0);
```
devLotto: Make a glGetFloat in GLUtil

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -990,6 +992,37 @@ public void draw()
...
+						final float samples = Math.min(anisotropicFilteringMode.getSamples(), maxSamples[0]);
+						gl.glTexParameteri(gl.GL_TEXTURE_2D_ARRAY, gl.GL_TEXTURE_MIN_FILTER, gl.GL_LINEAR_MIPMAP_LINEAR);
+						gl.glTexParameterf(gl.GL_TEXTURE_2D_ARRAY, gl.GL_TEXTURE_MAX_ANISOTROPY_EXT, samples);
+						break;
+				}
+				gl.glGenerateMipmap(gl.GL_TEXTURE_2D);
+			}
+			else if (!afEnabled && lastAnisotropicFilteringMode != anisotropicFilteringMode)
```
devLotto: Both if statements contain ``lastAnisotropicFilteringMode != anisotropicFilteringMode``, maybe do

```java
if (lastAnisotropicFilteringMode != anisotropicFilteringMode)
{
	if (afEnabled)
	{
		...
	}
	else
	{
		...
	}
}
```

pacf531:
So I have been looking at why the image improvement is so small with the filtering on. I took a look at the rooftops on several agility courses and it doesn't seem like anisotropic filtering is applying to the roof tile textures where it might be expected. I would have thought that that would be part of the texture array provided by the TextureProvider. Are the roof tile textures not part of the array and stored separately?

pacf531:
So I finally got time to revisit this and fixed the mipmapping not actually going past one level, and it does seem like that fixed the one remaining issue I had which was moire patterns on rooftops that were bothering me. Unfortunately, there is unlikely much I can do here after fixing that issue as far as increasing the effectiveness of the feature on visual quality. I tried to see which textures were affected by what I did and I get the following, as a test near Draynor Village. All the objects that are black are affected.



As can be seen, it apparently seems like there is a good portion of things that are affected but character models, accessible terrain, and character models are unaffected, along with some weird things like the bank windows. I'm not that experienced with how the rendering works in RuneLite but what I am assuming from what I can read now that I have more experience is that the rest of the objects that aren't in black are part of the compute shaders that are drawing them and thus I can't really touch them.

But other than that, I do have another concern, which is now that the TexStorage has to allocate 7 more levels of mipmapping, this will increase memory usage and there isn't much way I can really avoid it since the code creates the TexStorage at initialization and if the mipmaps aren't created, the fix I did doesn't work. No good way really to fix this. It can be mitigated by runtime texture compression but not sure if that is acceptable here or not.

But yeah, this is pretty much feature complete, I think, unless someone can point to something I did wrong. Not sure about the CI failing issue though unfortunately. If this isn't worth it to integrate in the client or something, feel free to toss it and close the issue. I did learn a good deal from writing this. More testing by other people would be appreciated to see if there are other areas that adding ansiotropic filtering.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1007,21 +1007,27 @@ public void draw()
...
-							if (anisotropicFilteringMode != anisotropicFilteringMode.TRILINEAR && gl.isExtensionAvailable("GL_EXT_texture_filter_anisotropic"))
+							if (gl.isExtensionAvailable("GL_EXT_texture_filter_anisotropic"))
 							{
 								final float maxSamples = glGetFloat(gl, gl.GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT);
 								final float samples = Math.min(anisotropicFilteringMode.getSamples(), maxSamples);
 								gl.glTexParameterf(gl.GL_TEXTURE_2D_ARRAY, gl.GL_TEXTURE_MAX_ANISOTROPY_EXT, samples);
 							}
+							else
+							{
+								log.warn("Ansiotropic filtering specified but no support detected. Defaulting to trilinear filtering.");
```
Nightfirecat: If reached, would this log on each `draw()` call? That is, 50x per second? 😬

pacf531: See line 1006 in the version you commented. The warning log only happens in this situation if the check for lastAnisotropicFilteringMode is not equal to anisotropicFilteringMode, so this warning should only happen when one changes the ansiotropic setting in Runelite and doesn't support ansiotropic filtering properly.

On a sidenote, the check is a bit pointless given practically all video cards made within the last decade has the extension and OpenGL 4.6 mandates ansiotropic support, but it's just a little bit of defensive programming for the truly esoteric GPU/driver.

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPluginConfig.java
```diff
@@ -74,11 +75,22 @@ default AntiAliasingMode antiAliasingMode()
 	@Range(
```
Toocanzs, line 75: This range annotation should be on the fog depth not on the anisotropic filtering mode

pacf531: Ah okay, this looks like a minor thing. However, as I am a bit too busy to revisit this and you have a separate pull request going, I will probably only fix this if it is completely necessary to do so.

pacf531:
Yeah, sorry, I have been busy so I haven't revisited this patch in quite a while. I should've added an image comparison again after I fixed things, correct, but yeah, I was a bit more down on things. My initial disappointment was more or less the fact that moire and shimmering could still be present due to the way the game renders and how much of it wasn't textured. I was initially only testing near Draynor and couple of other places around the area which may or may not have been the right choice, so seeing it in another area would've probably changed my mine on things.

In any case, since you have a more recent patch in your pull request which improves on what I did, I will close this pull request.

PR #7094, opened 2018-12-28
'Collapse Menu Entries' Plugin

Sypanite:
Collapses duplicate context menu entries on ground items, reducing clutter. Ten occurrences of "Take Spade" becomes one occurrence of "Take Spade (x10)".

It [sort of blew up](https://www.reddit.com/r/2007scape/comments/a9nosb/suggestion_clean_this_up/eckwaq2) on /r/2007scape!

PR #3570, opened 2018-06-02
Cleaner menus plugin

ittwit:
Added plugin that merges duplicates (i.e. same target and action) and displays the number of times it showed up next to the entry
Example: 
Picture demonstrating it

PR #7225, opened 2019-01-07
Condense duplicate items to item [quantity] in right click menu

RickyLoader:
Supports '[quantity]', '{quantity}', '(quantity)', and 'quantity', all with optional 'x' before the quantity. Quantity can be placed either before or after the item name.  

e.g: 

example image


Adam-:
This is duplicate of functionality of of #7094 and #3570. The reason I've pulled none of them is I had dropped a DM to Jagex trying to figure out their stance on this matter, since it is crossing some lines we've not crossed before (removal of items from grounditems menu). If I do not hear back in some timeframe I will probably accept some form of this since we've discussed it at length on Discord and I've not yet seen any reason why it might be considered OP.

Adam-:
As a follow up to this, I've not heard back anything and it looks like OSB has fully released this feature already, so we will take this, but probably a different PR.

RickyLoader:
Sounds good

PR #7227, opened 2019-01-07
timetracking: add default time in minutes to config

### LINKED ISSUE #7224: Setting the default timer time.
I use timers to track runite rock respawns on different worlds. I start the timer right after the rock gets depleted and label it with the world number. This way i know exactly when to hop to a certain world.

Now the problem is that each new timer is set to 5 minutes, and I want it to be 6 or 12 minutes by default so I wouldn't have to change it each time when I add it. So could you please add a feature, to change the default time of a new timer?

That shouldn't be too difficult to implement and it would help a lot of people out.

Thanks!

jakewilson:
The default time for timers in the timetracking plugin is now configurable.

Closes #7224

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
@@ -76,7 +76,8 @@ public interface TimeTrackingConfig extends Config
 			description = "The default time for the timer in minutes",
 			position = 4
 	)
-	default int defaultTimerMinutes() {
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -70,6 +70,17 @@ default boolean birdHouseNotification()
 		return false;
 	}
 
+	@ConfigItem(
+			keyName = "defaultTimerMinutes",
```
deathbeam: Fix your indentation

PR #7228, opened 2019-01-08
Add on-screen color for objects worth high-alching

the1satanfears:
Possible to indicate which items that are dropped on ground are worth high alching?


Spedwards:
I think you were supposed to create an issue, not a pull request...

the1satanfears:
Sorry didn't know the difference!

PR #7232, opened 2019-01-08
Combine RegenMeter & RunEnergy into a StatusOrbs plugin

TheStonedTurtle:
The config migration will want to be removed eventually.

Gif Showing it working:


deathbeam:
As poison indicators were added in separate plugin too I think this is fine and do not needs to be changed

PR #7236, opened 2019-01-08
Add option for configurable level marker width

Spedwards:
When the intermediate level markers were added, the markers were originally thicker. These changes would add an option to increase the width of the markers. Default remains 1px and the max is currently 5px because anything more would be a little ridiculous.


raiyni:
why not just set them at a reasonable width?

Spedwards:
deathbeam asked for them to be their current width for them to be merged. Being configurable, everyone gets what they want.

raiyni:
I'll leave what I said in discord here.

I think this is a dumb thing to have a config for. It should have a more reasonable width like 2 or 3.

PR #7247, opened 2019-01-09
Add priority to the eventbus

Trevor159:


runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
```diff
@@ -105,6 +107,9 @@ public synchronized void register(@Nonnull final Object object)
 			builder.putAll(subscribers);
 		}
 
+		builder.orderValuesBy(Comparator.comparing(Subscriber::getPriority)
```
deathbeam: This should not sort by name, because then it breaks registration order what should be preserved without priority

runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
```diff
@@ -62,6 +63,7 @@
 	{
 		private final Object object;
 		private final Method method;
+		private final float priority;
```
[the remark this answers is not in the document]
15987632: It was abex's idea which I agree with to be made a float, this way you don't have to refactor code if two side by side ints are taken and you can just put it in the middle.

runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
15987632: This can be done easily, also abex's idea.

PR #7258, opened 2019-01-10
Login Profiles

Spedwards:
Add and remove profiles to RuneLite so when you click one, the username field of the login screen is automatically filled out.


runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilesPanel.java
```diff
@@ -0,0 +1,222 @@
...
+			txtAccountLogin.setEchoChar((char) 0);
+			txtAccountLogin.setForeground(ColorScheme.LIGHT_GRAY_COLOR);
+		});
+
+		add(btnAddAccount, c);
+
+		addAccounts(config.profilesData());
+	}
+
+	void remove(ProfilePanel panel)
```
TheStonedTurtle: This is only called via `redrawProfiles()` and can be replaced with just the super call

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilesPanel.java
```diff
@@ -0,0 +1,222 @@
...
+		c.gridy++;
+		c.insets = new Insets(0, 0, 15, 0);
+
+		JButton btnAddAccount = new JButton("Add Account");
+		btnAddAccount.setPreferredSize(PREFERRED_SIZE);
+		btnAddAccount.setBackground(ColorScheme.DARKER_GRAY_COLOR);
+		btnAddAccount.setMinimumSize(MINIMUM_SIZE);
+		btnAddAccount.addActionListener(e ->
+		{
+			String data = txtAccountLabel.getText() + ":" + String.valueOf(txtAccountLogin.getPassword());
```
TheStonedTurtle: May want to prevent them from adding the default values and duplicate login names?

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilePanel.java
```diff
@@ -0,0 +1,167 @@
...
+		bottomContainer.setBackground(ColorScheme.DARKER_GRAY_COLOR);
+		bottomContainer.addMouseListener(new MouseAdapter()
+		{
+			@Override
+			public void mousePressed(MouseEvent e)
+			{
+				log.info("profilePanel", e);
+				if (SwingUtilities.isLeftMouseButton(e))
+				{
+					if (client.getGameState() == GameState.LOGIN_SCREEN)
```
TheStonedTurtle: See above comment

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilePanel.java
```diff
@@ -0,0 +1,167 @@
...
+		labelWrapper.add(panelActions, BorderLayout.EAST);
+		label.addMouseListener(new MouseAdapter()
+		{
+			@Override
+			public void mousePressed(MouseEvent e)
+			{
+				log.info("label", e);
+				if (SwingUtilities.isLeftMouseButton(e))
+				{
+					if (client.getGameState() == GameState.LOGIN_SCREEN)
```
TheStonedTurtle: Add to above if statement and use `.equals()` instead of `==`

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilePanel.java
```diff
@@ -0,0 +1,167 @@
...
+		JLabel delete = new JLabel();
+		delete.setIcon(DELETE_ICON);
+		delete.setToolTipText("Delete account profile");
+		delete.addMouseListener(new MouseAdapter()
+		{
+			@Override
+			public void mousePressed(MouseEvent e)
+			{
+				panel.getParent().remove(panel);
+				ProfilesPanel.removeProfile(String.join(":", parts));
```
TheStonedTurtle: You can just use `data` here right?

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilesPanel.java
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	private void addAccount(String data)
+	{
+		c.gridy++;
+		c.insets = new Insets(0, 0, 5, 0);
+		log.info(data);
+		ProfilePanel profile = new ProfilePanel(client, data, profilesConfig);
+		add(profile, c);
+		profiles.add(profile);
```
TheStonedTurtle: You never clear `profiles` and thus whenever call `redrawProfiles()` it creates new panels and adds them to this list.

You can just move all the profiles into a JPanel and remove this entire list.

runelite-client/src/main/java/net/runelite/client/plugins/profiles/ProfilePanel.java
Spedwards: In the case of enums, `.equals()` just uses `==` so it makes absolutely no difference.

Danny1071:
Added this with the latest changes and still doesn't work?

Trevor159:
this should be closed because it is an external plugin now

PR #7274, opened 2019-01-10
Added New Bones to skill_prayer.json

wonnetz:
Added Wyrm Bones, Drake Bones, and Hydra Bones


wonnetz:
I'm new to GitHub and how it works. I saw that other people were comparing branches to runelite:master and I figured that I shouldn't have been comparing wonnetz:master to it. The information should be correct, but I was trying to figure out the correct way to go about pushing it. Should I have not closed it?

Nightfirecat:
For now, this is fine, but you should generally use feature branches and not PR from your master branch.

Issue #7259, opened 2019-01-10
World hop history


datashe12:
Could a world hop history be added? Just something showing which worlds I have just been in, I would prefer it to be ordered so that the world I just hopped to is at the top and the first world I joined at the bottom, or the other way around. I would find it useful for when hopping around in wild looking for people, or hopping looking for a spot for bursting, sometimes I will hop back to a world I already hopped to, with the world hop limit u don't get that many world hops before getting timed out so I would find it useful.

zlshames:
I can work on this. I already have a prototype going. Just need to get some kinks out.

zlshames:
Here is what I ended up with: [World History Screenshot](https://i.imgur.com/NcdnRXQ.png)

zlshames:
I was told that having 2 tabs is unfavorable. I need to do some more work to just show recents at the top of the list, below the favorites

zlshames:
Ok, here is what I have now without the tabs: [Recent worlds at top of list](https://i.imgur.com/M4e1wN5.png)

zlshames:
I am no longer working on this. After I spent a bunch of time doing the history tab and other stuff, I realized I did not understand the feature request correctly. @datashe12 if you want the history tab, they can merge the code, but if you want it as a list at the top of the worlds list, then I deleted my code and frankly I don't want to go and redo it.

PR #7276, opened 2019-01-10
Adds world history to world hopper plugin (Issue #7259)

zlshames:
Features:
- World Hopper Plugin now is tabbed (Worlds and History)
- Clearing history
- Creating favorites from history tab
- Hoping using history tab
- Sorts by most recent world


Abextm:
This doesn't actually solve the problem listed in #7259

PR #7288, opened 2019-01-11
Add Slayer Task Locations (Konar)

### LINKED ISSUE #7267: Tasks from konar don't trigger the slayer counter to show.

ChaoticConundrum:
Fix some pattern matching for new slayer game text, and record task locations. Locations are also stored in config, and displayed on the task counter:

capture2
(ignore the zero points/streak, the config was somehow wiped out, unrelated to this PR)

This does **NOT** add the new slayer monsters to the plugin, I will leave it to someone else to pick the icons for that.

The game text currently understood:
- Gem 'Check" progress (same as slayer helmet?)
- Konar dialogue for new task
- Konar dialogue with current Konar task
- Konar dialogue with current non-Konar task
- Slayer master dialogue with current Konar task

I don't know what the game text is for Konar tasks with:
- Receiving a task from a slayer partner

Testing this is slow, so help testing would be appreciated.

This would resolve #7267.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
@@ -165,6 +165,24 @@ public interface SlayerConfig extends Config
 	void initialAmount(int initialAmount);
 
 	@ConfigItem(
+		keyName = "taskLocation",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default String taskLocation()
+	{
+		return "";
+	}
+
+	@ConfigItem(
+		keyName = "taskLocation",
+		name = "",
+		description = ""
+	)
+	void taskLocation(String key);
+
+	@ConfigItem(
 		keyName = "streak",
 		name = "",
 		description = "",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
@@ -141,6 +141,13 @@ public interface SlayerConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "amount",
+		name = "",
+		description = ""
+	)
+	void amount(int amt);
+
+	@ConfigItem(
 		keyName = "initialAmount",
 		name = "",
 		description = "",
@@ -156,13 +163,6 @@ public interface SlayerConfig extends Config
 		description = ""
 	)
 	void initialAmount(int initialAmount);
-
-	@ConfigItem(
-		keyName = "amount",
-		name = "",
-		description = ""
-	)
-	void amount(int amt);
 
 	@ConfigItem(
 		keyName = "streak",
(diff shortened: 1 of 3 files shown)


ChaoticConundrum:
Apparently tests are broken, working on it.

Adam-:
This should be extending the test with the newer messages too.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -604,10 +614,20 @@ private void rebuildTargetList()
...
+	}
+
+	private void setTask(String name, int amt, int initAmt, String location)
 	{
 		taskName = name;
 		amount = amt;
 		initialAmount = initAmt;
+		if (location == null)
+		{
+			location = "";
```
Adam-: Why?

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
```diff
@@ -140,12 +140,20 @@ default int amount()
 		return -1;
 	}
 
+	@ConfigItem(
```
Adam-: Why was this moved?

ChaoticConundrum: Just to reorganize each config item into getter/setter pairs, at some point another item was added between the getter/setter for that item. Of course it's not essential, but I was editing right next to it anyway.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
ChaoticConundrum: To simplify the logic where Matcher.group(String) is used, for the Pattern group that selects the task location, which is not always present. Matcher.group(String) returns null if the group is not present, so I sanitize null into the empty string here.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -98,10 +98,10 @@
 	private static final Pattern COMBAT_BRACELET_TASK_UPDATE_MESSAGE = Pattern.compile("^You still need to kill (\\d+) monsters to complete your current Slayer assignment");
 
 	//NPC messages
-	private static final Pattern NPC_ASSIGN_MESSAGE = Pattern.compile(".*Your new task is to kill\\s*(\\d*) (.*)\\.");
+	private static final Pattern NPC_ASSIGN_MESSAGE = Pattern.compile(".*(?:Your new task is to kill|You are to bring balance to)\\s*(?<amount>\\d+) (?<name>.+?)(?: in the (?<location>.+))?\\.");
```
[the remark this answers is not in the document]
ChaoticConundrum: You are right, good catch.

ChaoticConundrum:
From the examples I have seen, the Krystilia tasks should be correctly captured. I don't do wilderness slayer, so my experience is limited.

ChaoticConundrum:
I still don't have the text for receiving a Konar task from a slayer partner, in case anyone comes across it.

ChaoticConundrum:
I thought PR commits were squashed?

deathbeam:
They are squashed if they are doing only 1 thing

PR #7295, opened 2019-01-11
Don't send projectileMoved event if the projectile id isn't used

MagicfTail:
Meant to break custom plugins deemed not suitable for the client such as Vorkath and Alchemical Hydra plugins.

### DIFF
--- a/runelite-mixins/src/main/java/net/runelite/mixins/RSProjectileMixin.java
+++ b/runelite-mixins/src/main/java/net/runelite/mixins/RSProjectileMixin.java
@@ -24,7 +24,13 @@
  */
 package net.runelite.mixins;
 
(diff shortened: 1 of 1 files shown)


andrewterra:
Won't this prevent people from using plugins that use projectile IDs in the future? Why not just make it so that the only IDs that don't come through are the one's from Hydra and Vorkath?

RSNinja:
They already do this but at a lower level inside the injected client. Besides wouldn't doing it in the mixin like this make it easy for someone to edit?

andrewterra:
I'm pretty sure you can't personally change mixins and rebuild using Maven to make it so those projectile IDs will indeed come through if this change goes through. I think they are injected with the client as well.

PR #7312, opened 2019-01-12
Add hydra to boss respawn timers

### LINKED ISSUE #7305: Hydra boss doesn't have a boss respawn timer

steffenhauge:
Title. Closes #7305


steffenhauge:
#7313 Does the same, just use that one.

PR #7315, opened 2019-01-12
MLM - Add overlay for nuggets in bank, nuggets found in session and nuggets per hour

ConradMelcher:
Let me know if anything should be changed


deathbeam:
This was implemented i think

PR #7324, opened 2019-01-13
Add missing Forsaken Tower requirement to kourend & kebos diary plugin.

Monsterxsync:
Title.


EvanWalden:
You don't have to finish the quest to get the achievement, I myself did it during the quest. You *might* have to start the quest, I'm not sure and don't have an account I can test it with.

Monsterxsync:
Hence why its started the quest not finished. Need to complete upto the part where the furnace is lit.

Monsterxsync:
Removed the misspelled variant of diary requirement, didn't think it was worth making a separate pr for this.

PR #7331, opened 2019-01-13
Allow WASD Camera movement while shift is pressed.

### LINKED ISSUE #7165: Allow WASD Camera movement while shift is enabled
The WASD feature is amazing, but when using sticky keys to make it easier to drop items/just having it be more useful to have enabled in general, I am unable to use the awesome WASD camera movement and have to rely on the mouse-wheel instead

I have no idea how coding works but this would be very useful as I wouldn't have to keep turning on and off sticky keys which is somewhat rough considering I can't remap keys on my keyboard.

So yeah, being able to move camera with WASD while shift is enabled, if anyone can make this happen I'd be forever grateful

Keith-Cancel:
Fixes #7165 and allows camera movement while modifier keys are pressed. Also add an option to disable this option if anyone modified the default WASD keys and key combination like CTRL+W.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wasdcamera/WASDCameraConfig.java
@@ -76,4 +76,14 @@ public interface WASDCameraConfig extends Config
 	{
 		return new Keybind(KeyEvent.VK_D, 0);
 	}
+	@ConfigItem(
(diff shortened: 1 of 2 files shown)


Adam-:
Why does this currently not allow movement with modifiers pressed? It sees the input as a shift+w and not just w?

deathbeam:
yeah. its mostly because we have the keybindings configurable and so it takes exact input.

Adam-:
Can we make our config keybind ignore modifiers/

PR #7344, opened 2019-01-13
Issue #6026: Highlight egg colors in context menu

### LINKED ISSUE #6026: A "Highlight egg colors" option in the Barbarian Assault plugin

### LINKED ISSUE #6206: Add "Requires Battered key" to Elemental Workshop clues
image

Nava2:
* Highlights eggs with colour of egg

Color highlights

Closes #6026.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
@@ -25,16 +25,15 @@
  */
 package net.runelite.client.plugins.barbarianassault;
 
-import com.google.common.collect.ImmutableMap;
-import com.google.inject.Provides;
 import java.awt.Color;
 import java.awt.Font;
 import java.awt.Image;
 import java.util.Map;
 import javax.inject.Inject;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.Provides;
 import lombok.AccessLevel;
-import lombok.Data;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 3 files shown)


Spedwards:
Closes #6026, not 6206.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -205,6 +254,11 @@ private void announceTime(String preText, String time)
 			.build());
 	}
 
+	private boolean isInGame()
```
deathbeam: What is the point of this method?

Nava2:
> Closes #6026, not 6206.

Thanks! Fixed

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
Nava2: Adjusted it a bit now and reused it within the plugin. Just checks that the current player is in game. The logic within the plugin is a bit confusing -- going to try and improve it and update the logic more.

PR #7345, opened 2019-01-13
Add link and cheat-sheet for commits to guidelines

deathbeam:


Adam-:
I think most of the content of this file should just be moved to the wiki. The "how to git" section I think is maybe out of scope entirely.

PR #7353, opened 2019-01-14
Add heal for Dr Ford at Clan Wars to menu entry swapper

### LINKED ISSUE #7352: MenuEntrySwapper Suggestion: Dr. Ford @ Clan Wars
If you hit the clan wars portal, it resets your stats that are boosted. 

Swapping the option to "Heal" with "Talk" on Dr. Ford at the Clan Wars bank area.

**Additional context**
Doing this would allow for much quicker healing @ Dr. Ford, then hitting the prayer alter that is there as well to regenerate your health and prayer at clan wars, without resetting boosted stats.

Not really sure how many other people do this, but I know when I just hit a stat boosting potion I much prefer to heal with Dr. Ford and use alter rather than using the portal there.

gvandycke:
Add heal for Dr Ford at Clan Wars to menu entry swapper
Closes #7352


Bjdufre1:
closes #7352

PR #7354, opened 2019-01-14
Add Performance Tracker plugin

TheStonedTurtle:
Adds the Performance Tracker plugin which tracks your performance at certain activities. The tracker is toggled on/off based on region ID. You can right-click the overlay to pause or reset it. The tracker starts paused and will automatically start as soon as you interact with an NPC.

### Supported Activities
- Most PvM Bosses (Slayer, GWD, etc)
- Theatre of Blood

### Stats Tracked
- Damage Dealt (& Max hit)
- Damage Taken (& Max hit)
- Time Spent
- Damage per Second

Overlay

Recap message upon leaving area:

Gif Example:
https://gfycat.com/InexperiencedIllustriousBorderterrier


james-munson:
PerformanceTrackerOverlay.java:[54,17] cannot find symbol
  symbol:   method getMenuOptions()

TheStonedTurtle:
@james-munson looks like the PR for right-click menu entries was reverted, I'm waiting for it, or something similar to be remerged.

PR #7358, opened 2019-01-14
Retain previous ping UI until updated

### LINKED ISSUE #7144: World switcher panel deletes pings before updating them
This is obviously intended but I think it shouldn't be this way. A better behavior would be to change the ping of the world when it is updated rather than deleting the ping of every world and then updating them all individually. This way the panel doesn't do this every time the pings get updated if you have the panel open and sorted by ping https://streamable.com/26i9n

Digbigpig:
I'm looking for feedback on this. Closes #7144 

The world hopper plugin clears all rows to the panel before updating the ping.
That would result in the previous ping being removed from the UI before it could 
be updated. This commit allows the row to retain the previous ping value until it 
gets updated by the ping service.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
@@ -217,12 +217,9 @@ class WorldSwitcherPanel extends PluginPanel
 	{
 		Map<Integer, Integer> pingHistory = new HashMap<>();
 
-		if (!rows.isEmpty())
+		for (WorldTableRow row : rows)
 		{
-			for (WorldTableRow row : rows)
-			{
-				pingHistory.put(row.getWorld().getId(), row.getPing());
-			}
+			pingHistory.put(row.getWorld().getId(), row.getPing());
 		}
 
 		rows.clear();
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
@@ -64,7 +64,6 @@ class WorldSwitcherPanel extends PluginPanel
 	private ArrayList<WorldTableRow> rows = new ArrayList<>();
 	private WorldHopperPlugin plugin;
 
-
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
```diff
@@ -63,6 +64,8 @@
 	private ArrayList<WorldTableRow> rows = new ArrayList<>();
 	private WorldHopperPlugin plugin;
 
+	private Map<World, Integer> pingHistory = new HashMap<>();
```
Digbigpig: I'm not familiar with all the datatypes in java. Is this the best implementation of what I'm trying to achieve here?

deathbeam: Probably

stcktrce:
Wouldn't it be better to put the HashMap inside the `populate();` function itself? Considering the data is essentially stale and useless after a new ping has been made?

```
void populate(List<World> worlds)
{
    for (WorldTableRow row : rows)
    {
        // grab ping, shove into hashmap
    }
    for (int i = 0; i < worlds.size(); i++)
    {
        // grab data from hashmap, insert into constructor
    }
}
```

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
```diff
@@ -214,12 +217,33 @@ void resetAllFavoriteMenus()
...
 
 		for (int i = 0; i < worlds.size(); i++)
 		{
 			World world = worlds.get(i);
-			rows.add(buildRow(world, i % 2 == 0, world.getId() == plugin.getCurrentWorld() && plugin.getLastWorld() != 0, plugin.isFavorite(world)));
+			Integer ping;
+
+			if (!pingHistory.containsKey(world))
+			{
+				ping = 0;
```
raiyni: just initialize ping to 0

Digbigpig: Went with .getOrDefault() 
Seemed to be the same result with less code.

Trevor159:
This issue is fixed in release

PR #7362, opened 2019-01-15
Add a filter for player indicator based on your level. feature-request

ar664:
Useful for things like BH and the wildy when you have to keep right clicking people to find out their combat level. This will add an option to limit the player names seen based on the your level and an option you configure. So it could be your_level +-5 names are displayed along with their combat lvl.


raiyni:
Pk scouters are rejected content

1337umad:
runeliteplus accepts rejected features ;)
https://github.com/runelite-extended/runelite/
https://runelitepl.us

PR #7384, opened 2019-01-16
farming: Support Kebos Lowlands update

Abextm:


runelite-client/pom.xml
```diff
@@ -220,6 +220,12 @@
 			<version>4.12</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
Adam-: This is already a dependency of junit so I wouldn't expect you to have to pull it in?

Abextm: JUnit only has a dependency on `hamcrest-core` which does not contain any `Matcher` implementations.

MagicfTail:
It would make more sense if the celestrus tree is under fruit tree patches since it's a tree that has produce. It also has almost the same grow time and will most likely be farmed in the same farming run, making the user not have to follow 2 different tabs.

deathbeam:
If redwood works like other regular trees then that one imo should be in respective category tab too.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/Produce.java
```diff
@@ -102,17 +105,31 @@
...
-	TEAK("Teak", ItemID.TEAK_LOGS, 560, 8),
 	GRAPE("Grape", ItemID.GRAPES, 5, 8, 0, 5),
 	MUSHROOM("Mushroom", ItemID.MUSHROOM, 40, 7, 0, 7),
-	MAHOGANY("Mahogany", ItemID.MAHOGANY_LOGS, 640, 9),
-	CACTUS("Cactus", ItemID.POTATO_CACTUS, 80, 8, 20, 4),
 	BELLADONNA("Belladonna", ItemID.CAVE_NIGHTSHADE, 80, 5),
 	CALQUAT("Calquat", ItemID.CALQUAT_FRUIT, 160, 9, 0, 7),
-	SPIRIT_TREE("Spirit Tree", ItemID.SPIRIT_TREE, 320, 13);
+	SPIRIT_TREE("Spirit tree", ItemID.SPIRIT_TREE, 320, 13),
+	CELASTRUS("Celastrus", ItemID.BATTLESTAFF, 160, 6, 0, 4),
```
MagicfTail:
```suggestion
	CELASTRUS("Celastrus", ItemID.CELASTRUS_BARK, 160, 6, 0, 4),
```

This should be the bark

runelite-client/pom.xml
deathbeam: https://junit.org/junit4/javadoc/4.12/org/junit/rules/ErrorCollector.html based on this it should be already in JUnit

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/Produce.java
MagicfTail: That's what the tree produces, and it looks better imo

Abextm: A battlestaff is more recognizable. Note that this is just the icon, it does nothing else.

PR #7385, opened 2019-01-16
Add Rare Tree map points

### LINKED ISSUE #5223: Rare tree indicators on the map should show what type of tree is there in the tooltip
This one is pretty self-explanatory. I would like to be able to hover over the rare tree indicators to see what kind of tree is in that location( I.E. "Rare Trees - Yew"). 
Pardon me if this is already a feature as I am relatively new to Runelite, but I couldn't find this option in the menu.

Spedwards:
Closes #5223


deathbeam:
We had issue for this if I remember correctly, mind linking it?

PR #7395, opened 2019-01-16
Ignore modifiers in WASD plugin input

### LINKED ISSUE #7165: Allow WASD Camera movement while shift is enabled

Abextm:
Supersedes/Closes #7369
Supersedes/Closes #7331
Closes #7165

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
@@ -514,11 +514,15 @@ public class ConfigManager
 		{
 			return Instant.parse(str);
 		}
-		if (type == Keybind.class)
+		if (type == Keybind.class || type == ModifierlessKeybind.class)
 		{
 			String[] splitStr = str.split(":");
 			int code = Integer.parseInt(splitStr[0]);
 			int mods = Integer.parseInt(splitStr[1]);
+			if (type == ModifierlessKeybind.class)
+			{
+				return new ModifierlessKeybind(code, mods);
+			}
 			return new Keybind(code, mods);
 		}
 		return str;
--- a/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
@@ -60,7 +60,7 @@ public class Keybind
 	private final int keyCode;
 	private final int modifiers;
 
-	public Keybind(int keyCode, int modifiers)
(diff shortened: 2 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/config/Keybind.java
```diff
@@ -60,7 +60,7 @@
 	private final int keyCode;
 	private final int modifiers;
 
-	public Keybind(int keyCode, int modifiers)
+	protected Keybind(int keyCode, int modifiers, boolean ignoreModifiers)
```
deathbeam: maybe this constructor should be just keyCode, ignoreModifiers? as passing modifiers when you want to ignore them do not makes much sense. Or even just keyCode, even though that can maybe be a bit confusing idk

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -514,11 +514,15 @@ static Object stringToObject(String str, Class<?> type)
...
 		}
-		if (type == Keybind.class)
+		if (type == Keybind.class || type == ModifierlessKeybind.class)
 		{
 			String[] splitStr = str.split(":");
 			int code = Integer.parseInt(splitStr[0]);
 			int mods = Integer.parseInt(splitStr[1]);
+			if (type == ModifierlessKeybind.class)
+			{
+				return new ModifierlessKeybind(code, mods);
```
deathbeam: Again here, storing mods and then using them for modifierLessKeybind do not makes much sense

Abextm: same thing as above

runelite-client/src/main/java/net/runelite/client/config/Keybind.java
deathbeam: But shift is also KeyCode i believe, and so you dont need to do that

Abextm: internally (and therefore through this constructor too) modifier only keys are stored as `VK_UNDEFINED` with modifiers set

PR #7397, opened 2019-01-17
boss timers: add notifications on respawn

jplsek:
This adds a notification when the boss has respawned. This can be useful if you're soloing certain bosses that have long respawn times and you are not paying attention while waiting for them to respawn.


runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/RespawnTimer.java
```diff
@@ -25,22 +25,39 @@
...
 		return boss;
 	}
+
+	@Override
+	public boolean render()
+	{
+		Duration timeLeft = Duration.between(Instant.now(), endTime);
+		if (timeLeft.isNegative())
+		{
+			bossPlugin.notifyRespawn(name);
```
jplsek: Whoops, just went to test it more by enabling the chat notifications and realized that it sends notifications in multiple frames. For some reason I thought once render returns false that the infobox would be removed in the same frame. Will have to change how this works.

Kamielvf:
Wouldn't this be more useful _before_ the boss respawns? e.g. 5 seconds left?

jplsek:
Yea, that would make sense. I could then remove it from the Timer and just keep it all in the actual plugin by using a executor or some other timer.

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersPlugin.java
```diff
@@ -77,10 +96,24 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+		if (config.enableNotifications())
+		{
+			notificationTimer.schedule(new TimerTask()
+			{
+				@Override
+				public void run()
+				{
+					notifier.notify(name + " is about to respawn");
+				}
+			}, boss.getSpawnTime().toMillis() - 5_000);
```
Nava2: Could this timer value be configurable?

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersPlugin.java
```diff
@@ -77,10 +96,24 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
 		infoBoxManager.addInfoBox(timer);
+
+		if (config.enableNotifications())
+		{
+			notificationTimer.schedule(new TimerTask()
+			{
+				@Override
+				public void run()
+				{
+					notifier.notify(name + " is about to respawn");
```
Nava2: Consider: `name + " will respawn in {DURATION}.`

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersPlugin.java
jplsek: PR Updated 👍

deathbeam:
I dont like #7439 because it will never be useful for regular NPCs

andreykaipov:
> 
> 
> True, it's only useful in a low number of situations. It can be useful when soloing only Rex at DKs (which is typical for low leveled Ironman and inexperienced bossers).
> 
> Maybe something like #7439 would be a better fit, since it can be customized per NPC?
> 
> What should be used instead of a `java.util.Timer`? Would some variation of [this ](https://github.com/runelite/runelite/blob/90711882e00b8f3996b8633dee6e8d855dbc79f4/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java#L397)be better?

Heh - Rex is exactly why I searched for this feature. It's unfortunate there's pushback regarding this.

PR #7408, opened 2019-01-17
Inverse collapse menu entry loop

deathbeam:
This makes sure that menu collapses to first entry because menu is built
in inverse. Before setting the entries, inverse them back to original
order.


deathbeam:
s/inverse/reverse

Issue #7404, opened 2019-01-17
Farming Guild redwoods don't have the indicator that Woodcutting Guild ones have


Enriath:
Top: Farming Guild
Bottom: Woodcutting Guild
![](https://i.imgur.com/4rn2l0e.jpg)

Reported by `Kaan#3962` in the Discord

russel204:
Bump. Farming guild redwood still doesn't have the overlays

PR #7412, opened 2019-01-17
Added Missing Indicator to on of the Redwood Trees

ermalsh:
Redwood Tree Indicator for REDWOOD_29671 was missing. Added to Close: #7404

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
@@ -29,11 +29,12 @@ import java.util.Map;
 import lombok.Getter;
 import static net.runelite.api.ObjectID.REDWOOD;
 import static net.runelite.api.ObjectID.REDWOOD_29670;
+import static net.runelite.api.ObjectID.REDWOOD_29671;
 
 @Getter
 enum Tree
 {
-	REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670);
+	REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670, REDWOOD_29671);
 
 	private final int[] treeIds;
 


deathbeam:
Are you sure this is correct? <https://github.com/runelite/runelite/pull/7260/commits/2e0c7bfbfac05bbecf8b5067d0d8cfa8c0a96491> based on this, this ObjectID was there before farming guild update

ermalsh:
> Are you sure this is correct? [2e0c7bf](https://github.com/runelite/runelite/commit/2e0c7bfbfac05bbecf8b5067d0d8cfa8c0a96491) based on this, this ObjectID was there before farming guild update

I will double check, I reckoned it was this that was missing. Is there anyway to view object ID's even if you don't have the specified level to spawn them? For the Redwoods, you need 90 farming - if you don't have 90, how can we get the objectID?

deathbeam:
If you cant get near to the object nope, but I am sure you can just ask some contributor etc to get that data for you (i unfortunately dont have 90 farming as well, i think)

Abextm:
`::setvarb 7907 41` will fully grow a redwood. The redwood is split into atleast 18 different multiloc objects, and maybe more - I think devtools doesn't show them all

deathbeam:
We need the actual chop areas for the redwood I think, can you get them without actually climbing up?

Abextm:
You have to climb up, but you can do that without a grown redwood.

ermalsh:
First level
https://gyazo.com/6722a4160def8c8e1491f658aef1b45a
Second level
https://gyazo.com/bac8643a6eff142a75de6a63d4de0464
Third level
https://gyazo.com/1c67da5561085219f4e57d87a8487687

deathbeam:
*668, *669 and *670 looks like the ids that might work, did you tried adding those?

ermalsh:
> *668, *669 and *670 looks like the ids that might work, did you tried adding those?

These are the three that are already in the code, both in [ObjectID.java](https://github.com/runelite/runelite/blob/fef01ed2a07365159703d1c5e34ae5da3b2913ce/runelite-api/src/main/java/net/runelite/api/ObjectID.java#L15861) and [Tree.java](https://github.com/runelite/runelite/blob/fef01ed2a07365159703d1c5e34ae5da3b2913ce/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java#L37). That's why I figured it would work, may someone can with 90 farming can give it a shot.

deathbeam:
Can someone verify if this actually works?

Adam-:
I too would like someone to verify this.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
```diff
@@ -28,12 +28,14 @@
...
 import static net.runelite.api.ObjectID.REDWOOD;
+import static net.runelite.api.ObjectID.REDWOOD_29669;
 import static net.runelite.api.ObjectID.REDWOOD_29670;
+import static net.runelite.api.ObjectID.REDWOOD_29671;
 
 @Getter
 enum Tree
 {
-	REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670);
+	REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29669, REDWOOD_29670, REDWOOD_29671);
```
SebastiaanVanspauwen: This should be changed to:  
REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670, NULL_34633, NULL_34635, NULL_34637, NULL_34639);

REDWOOD_29669 and REDWOOD_29671 are the depleted versions.

SebastiaanVanspauwen: This should be changed to :
REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670, NULL_34633, NULL_34635, NULL_34637, NULL_34639);

Adam-:
Are you sure about this? If they are named `NULL` it means they have no names ingame, which seems unlikely.

15987632:
The icons currently don't go away when the wood is depleted

PR #7450, opened 2019-01-19
Add Swap Climb-up/Climb-down for Climb - Alternate with Hotkey

### LINKED ISSUE #7447: Menu Entry Swapper - Stairs option
Would be cool to have the option to set a default option for climbing stairs that go up and down. Could be configured by shift clicking the option you want default or changed trough the settings, witch ever is easier to do. This would make using the spinning wheel in Lumbridge slightly less annoying.

ermalsh:
Closes #7447 
- Switch between up/down with hotkey
- Preview: https://gyazo.com/07fafada1d90e0a586fc7e67b2f574a3

**Note: My line endings for Config files are still acting up... but it's only on Config files. I've tried multiple attempts to fix it with no luck. I even used a config file deathbeam gave me, anyone know how to fix?**


Nava2:
You likely need to reset the file's line endings. `dos2unix ./path/to/file` should fix the line endings. Re-add the file and commit.

PR #7453, opened 2019-01-19
Add correct time for anti vemon potion

gregg1494:
Changed Anti-venom timer overlay time from 12 minutes to 1 minutes. Which is now consistent with Anti-vemon plus.


15987632:
This should probably start two different timers, one for anti-poison and one for anti-venom

PR #7464, opened 2019-01-20
Cleanup overlay menu entry usage

deathbeam:
- Remove OverlayManager#addMenu/removeMenu - removeMenu is completely unnecessary and addMenu does nothing.
- Remove overlayManager#overlayId - It is not used outside of OverlayManager so just swap it with direct use
of indexOf.
- Move overlay menu entries to Overlay class
- Pass plugins to overlay constructor where it was missing


deathbeam:
It's added to those where it was missing. Its required for configure menu to work and I just added it as I was going through the overlays and moving menu entries there, that is why it is in same commit as the menu entry moving

PR #7467, opened 2019-01-20
world hopper: allow disabling chat hop message

jplsek:
This adds an option to disable the quick hop chat message to the world hopper. This can be useful to streamers that are world hopping and don't want to show their world to the stream.

Even though streamers could just create an overlay over the in game world hopper, this is just a quality of life change so they don't need to, and instead can utilize the sidebar or quick-hop features of the plugin.

If there is a game integrity reason to not add this (even for the point above), let me know, and I'll either modify or close this PR.


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperConfig.java
```diff
@@ -90,4 +90,16 @@ default boolean ping()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "showMessage",
+		name = "Show world hop message in chat",
+		description = "Shows what world is being hopped to in the chat. " +
+			"This may be useful to disable when world hopping as a streamer.",
```
deathbeam: I dont think this part of description is useful or needed, remove it.

PR #7480, opened 2019-01-21
Add Combat Indicator Plugin

MrMagaw:
Adds an indicator on NPCs interacting with the player


nexio123:
not working

MrMagaw:
I don't suppose I could get a little more information than that, @nexio123.
Does it not show up in the config?
Does it show up (and is enabled) but the indicators don't appear?

nexio123:
oh Sorry, yeah Once enabled it will not show the indicators, https://gyazo.com/b38322c51d67c7ccbeb5c76dd2b18991

deathbeam:
What is the point of this?

MrMagaw:
@nexio123
Currently it's set up to show the indicator when someone has you as a target, not when you have them as a target.
Do you think we should have it so there are multiple colour options to indicate the three possible states (1. You're targeting them, 2. They're targeting you, and 3. Both 1 and 2)?

@deathbeam
It has a few uses including being able to pick out which NPC is attacking you when they're stacked,  or seeing if they're attacking you when you're stacked with other players. It's something I got used to with RS3, and I like it, so I wanted to bring it to OSRS

nexio123:
I'd say make it customizeable so when you target someone it will also highlight them. could come of use

MrMagaw:
@nexio123 It now is configurable to show the indicator on entities that you're targeting, as well as entities that are targetting you

abextm:
Jagex's [recent statement](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) prohibits `Indicates which player an NPC is focused on`

PR #7500, opened 2019-01-22
Add GE Fuzzy Search Option

dennisdev:
Depends on internal PR 43

image
image


Adam-:
Unless this only performs fuzzy searching when nothing else matches I think this would annoy people more than anything.

nexio123:
What Adam said, make it only show the "fuzzy options" if nothing what was searched has a match

DevDennis:
image
Added some more options
You can now choose to only use fuzzy matching when no match is found

dennisdev:
Updated

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
```diff
@@ -96,4 +97,38 @@ default boolean showExact()
 	{
 		return false;
 	}
+
```
Nightfirecat: One newline too many here.
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeSearchMode.java
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
Nightfirecat: Another extra newline too many
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -156,6 +171,67 @@
...
+
+				if (queryChar == termChar)
+				{
+					indices.add(termIndex);
+
+					// we can leave the nested loop. Every character in the
+					// query can match at most one character in the term.
+					termCharacterMatchFound = true;
+				}
+			}
```
Nightfirecat: `termCharacterMatchFound` can be removed and be replaced by a `break;` statement.
```suggestion
			while (termIndex < termLowerCase.length())
			{
				final char termChar = termLowerCase.charAt(termIndex);

				if (queryChar == termChar)
				{
					indices.add(termIndex);

					// we can leave the nested loop. Every character in the
					// query can match at most one character in the term.
					break;
				}
				
				termIndex++;
			}
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +517,138 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+		{
+			highlightSearchMatches();
+		}
+	}
+
+	private void highlightSearchMatches()
+	{
+		String input = client.getVar(VarClientStr.INPUT_TEXT);
+
+		String highlightColorHex = String.format("%06x", config.highlightMatchColor().getRGB() & 0xFFFFFF);
```
Nightfirecat:
```suggestion
		String highlightColorHex = ColorUtil.colorToHexCode(config.highlightMatchColor());
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +517,138 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+
+		event.consume();
+
+		int resultCount = 0;
+		if (searchMode == GrandExchangeSearchMode.FUZZY_FALLBACK)
+		{
+			List<Integer> ids = IntStream.range(0, client.getItemCount())
+					.mapToObj(itemManager::getItemComposition)
+					.filter(item -> item.isTradeable() && item.getNote() == -1
+							&& item.getName().toLowerCase().contains(input))
```
Nightfirecat: One indent less here
```suggestion
						&& item.getName().toLowerCase().contains(input))
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +517,138 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+					.limit(MAX_RESULT_COUNT + 1)
```
Nightfirecat: Why are we limiting this to 1 above the max result count?

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +517,138 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+					.mapToObj(itemManager::getItemComposition)
+					.filter(item -> item.isTradeable() && item.getNote() == -1)
+					.filter(item -> getScore.applyAsInt(item) > 0)
+					.sorted(Comparator.comparingInt(getScore).reversed()
+							.thenComparing(ItemComposition::getName))
+					.limit(MAX_RESULT_COUNT)
+					.map(ItemComposition::getId)
+					.collect(Collectors.toList());
+
+			client.setGeSearchResultIndex(0);
```
Nightfirecat: This looks like it's common between the `if (searchMode == GrandExchangeSearchMode.FUZZY_FALLBACK)` block and this block. Can it be extracted from each of them?

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeSearchMode.java
dennisdev: This newline is also in the plugin class and some other classes. Should it be removed from the plugin class as well?

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
dennisdev: That is for checking whether we've exceeded the max result count so the client can display the too many matches message
```java
if (ids.size() > MAX_RESULT_COUNT)
```

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeSearchMode.java
Nightfirecat: It's not too important to remove it from the plugin, but since this is newly added it shouldn't be there.

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +516,137 @@ public void onWidgetLoaded(WidgetLoaded event)
...
 	{
 		// GE offers setup init
-		if (event.getScriptId() != ScriptID.GE_OFFERS_SETUP_BUILD)
+		if (event.getScriptId() == ScriptID.GE_OFFERS_SETUP_BUILD)
+		{
+			rebuildGeText();
+		}
+		else if (event.getScriptId() == ScriptID.GE_ITEM_SEARCH && config.highlightSearchMatch())
+		{
+			highlightSearchMatches();
```
Adam-: What is the point of this without fuzzy matches? This seems like it would not be useful since the match would always contain the literal search

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
```diff
@@ -96,4 +97,37 @@ default boolean showExact()
...
+	@ConfigItem(
+			position = 8,
+			keyName = "highlightMatchColor",
+			name = "Highlight Match Color",
+			description = "Color of the search match highlight"
+	)
+	default Color highlightMatchColor()
+	{
+		return new Color(0x800000);
```
Adam-, line 120: This seems like overkill for config I would just make it a static final field

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
```diff
@@ -441,11 +502,134 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+		Widget[] children = results.getDynamicChildren();
+		int resultCount = children.length / 3;
+
+		for (int i = 0; i < resultCount; i++)
+		{
+			Widget itemNameWidget = children[i * 3 + 1];
+			String itemName = itemNameWidget.getText();
+
+			List<Integer> indices;
+			FuzzyScore fuzzy = new FuzzyScore(Locale.ENGLISH);
```
Adam-, line 535: Can't this just be used each loop?

dennisdev: Yes, I've moved it to a constant

PR #7509, opened 2019-01-23
Wilderness level swap

lund8619:
Add swap for wilderness lever to Edgeville.


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -421,6 +421,10 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 				swap("quick-travel", option, target, true);
 			}
 		}
+		else if (config.swapWildernessLever() && option.equals("ardougne") && target.equals("lever"))
```
Spedwards:
```suggestion
		else if (config.swapWildernessLever() && target.equals("lever"))
```

Checking the option isn't required here. Best to keep it simple.

PR #7512, opened 2019-01-23
Ancient staff prayer bonus

lund8619:
Add ancient staff to prayer items list


lund8619:
Could one of the admins please unblock me from the dev channel? I'm having some git trouble I'd really like to ask about in there.

lund8619:
Discord name is IronMosiah.

lund8619:
This is worse than what I imagine playing as a UIM would be. So... permban from the dev page?

Nightfirecat:
It looks like you just need to rebase and drop the first three commits. Something like `git rebase -i HEAD~4` and marking the first three commits as `drop` should do the trick.

lund8619:
Awesome. I was able to rebase and then had to do a force-push. When reviewing the code, everything looks the way I want it. Now I can't reopen because I did a force-push I guess? Do I have to make a new PR since I did a force-push?

Nightfirecat:
You can do that, or reset to the previous state, force-push, re-open the PR, then repeat the steps you followed. At this point, opening a new PR is probably easier.

PR #7514, opened 2019-01-23
Updated dependencies

ESchouten:


Bartvhelvert:
This shouldn't include alpha and beta builds.

ESchouten:
Removed alpha and beta versions

ESchouten:
Downgraded alpha and beta versions to the latest stable release

Nightfirecat:
Per [Adam's comment](https://github.com/runelite/runelite/pull/8710#issuecomment-488401185):

> I probably don't want to upgrade anything without having a reason to (security, stability, bug fixes, new features?). I also certainly wouldn't want to upgrade everything at once because when it inevitably breaks its harder to pinpoint. What we have now works without any issuses that I know of though.

PR #7519, opened 2019-01-23
Limit anim idle duration to minimum of 600ms

deathbeam:
Animation idle notifications can only work correctly if the delay is at
least 1 game tick.


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -70,6 +72,7 @@ default boolean logoutIdle()
 		description = "The notification delay after the player is idle",
 		position = 4
 	)
+	@Range(min = ANIMATION_IDLE_MINIMUM)
```
Twiglet1022: This will also affect the interaction idle notification. Isn't it enough to just do what you've done in `IdleNotififerPlugin.java` so that it does not respect any configs below 601 for animations?

I need my instant fishing interaction notifications. :P

PR #7520, opened 2019-01-23
Added Cast Bloom to Menu Entry Swapper

StanleyMole:
Added Cast Bloom for Silver sickle (b) and Ivandis flail to Menu Entry Swapper


deathbeam:
Use shift click configuration

PR #7528, opened 2019-01-24
 combat level plugin: Fix level calculation, move logic to plugin 

### LINKED ISSUE #7411: Levels to next combat are not correct for most skills
Related to #7103 and its fix.

https://i.imgur.com/nGLPrL6.png

This is with the following stats, and should be the same result with 2 prayer levels to level, not 0:

* 70 attack
* 74 hp
* 66 defense
* 69 magic
* 60 prayer
* 70 range
* 76 strength

Nightfirecat:
This commit addresses a number of issues previously evident in the
Combat Level plugin:

- Level calculation logic was run per-frame in the render loop
- Tests for level calculation were not exhaustive enough to cover
  several cases which calculated levels to next combat incorrectly
- Levels to next combat logic, while attempting to be smart and avoid
  excess calculations were not correct

This commit attempts to address these issues by updating the levels to
next combat logic to follow more naive logic--that is, incrementing
levels until calculating a higher combat level--and by using the
experience changed event to trigger recalculations. It also adds two new
test cases which failed on the previous level calculation code in an
attempt to better ensure its correctness.

Fixes #7411


lund8619:
Yeah... I had no chance. Looks awesome. I really like the fact that it's not calculating every frame. Is there a way to see other things that are calculating every frame? Seems like it wouldn't be difficult to greatly increase fps if other code could be improved similarly to this, even if the existing code is functioning properly.

Nightfirecat:
Most overlays are already doing as little calculation per frame as possible. (generally only containing render logic in the `render()` function) I don't know why this one wasn't written to do level calculation separately from rendering.

deathbeam:
It was doing the calculations only when hovering the widget and they were not really expensive so I dont think it matters that much

deathbeam:
It might even be slightly worse now tbh, because its calculating the next level even when not hovering right

Nightfirecat:
Updated to null tooltip on combat exp change and rebuild on hover only if null.

runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelPlugin.java
```diff
@@ -112,4 +136,127 @@ public void onGameTick(GameTick event)
...
+
+		final int level = Math.min(Experience.MAX_REAL_LEVEL, Experience.getLevelForXp(client.getSkillExperience(skill)));
+		cmbSkillLevels.put(skill, level);
+	}
+
+	Tooltip getLevelsUntilTooltip()
+	{
+		if (levelsUntilTooltip != null)
+		{
+			return levelsUntilTooltip;
```
TheStonedTurtle: check for == null and build here and always return `levelsUntilTooltip` under the if statement

Nightfirecat: Good point, fixed

Nightfirecat:
Rebased after wilderness level range changes.

Nightfirecat:
Rebased after 8823.

@Adam- This PR adds two test cases--one of which failed before 8823 (but works fine on master currently), the other fails even on master.

Nightfirecat:
I've split the main bulk of this PR into two commits: one to drop the logic into the plugin, and then the other to fix the calculation bugs.

PR #7534, opened 2019-01-24
Add highlight to fairy rings

daviscook477:
This plugin highlights the clickbox of fairy rings when the player is close enough to use them.
fairy ring highlight

With the GPU plugin allowing for the users to see more of the world (and #6539 preventing them from interacting with it) it's easy to try to interact with fairy rings too early while moving around OSRS. This modification to the fairy ring plugin helps to eliminate that issue by highlighting the fairy rings.
too_far

Additionally the fairy ring hitbox is a square when clearly the mushrooms are in a circle on the ground so this clarifies that to the player.
square_clickbox

In zanaris.
in_zanaris

At the lighthouse.
at_the_lighthouse


raiyni:
You can just mark the fairy ring with the Object Markers plugin.

daviscook477:
Marked objects stay highlighted outside of the valid interaction range.
marked_objects_highlighted_outside_interaction_range

PR #7536, opened 2019-01-24
Add option to require shift for overlay menus

deathbeam:
Enabled by default. This was on original PR review and was discussed in
Discord. It is better to just use what we already use for adding our
custom entries at other places, and there is possibility that overlays
will have a lot of right click menus in future, so making it require
shift is most sane option.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -75,6 +75,7 @@ public class OverlayRenderer extends MouseAdapter implements KeyListener
 	private final Point mousePosition = new Point();
 	private Overlay movedOverlay;
 	private boolean inOverlayDraggingMode;
+	private boolean inMenuEntryMode;
 	private MenuEntry[] menuEntries;
 
 	// Overlay state validation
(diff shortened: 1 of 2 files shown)


Adam-:
Did we give up trying to find a smarter way to determine when this is being drawn over?

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -406,6 +418,12 @@ public void keyReleased(KeyEvent e)
 		{
 			inOverlayDraggingMode = false;
 		}
+
+		if (!e.isShiftDown())
+		{
+			inMenuEntryMode = false;
+			menuEntries = null;
```
Adam-: This is weird since it would clear the menu just from releasing a key? Also this is on a different thread from the render thread so it might race with onClientTick

deathbeam: Yeah i dont think its necessary, I will remove it

deathbeam:
I mostly gave up on that because I don't think there is a way to completely prevent left-clicks @Adam- as you can also move overlays on top of of widgets where you will have only cancel option.

PR #7543, opened 2019-01-24
Add anti-poison timers to anti-venom potions

deathbeam:
Preview:
peek 2019-01-24 21-43


Adam-:
I would try adding both and overriding `render` to make it not render for the first minute.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -49,6 +49,7 @@
 	DMM_FULLTB(SpriteID.SPELL_TELE_BLOCK, GameTimerImageType.SPRITE, "Deadman Mode Full Teleblock", 150, ChronoUnit.SECONDS, true),
 	DMM_HALFTB(SpriteID.SPELL_TELE_BLOCK, GameTimerImageType.SPRITE, "Deadman Mode Half Teleblock", 75, ChronoUnit.SECONDS, true),
 	ANTIVENOMPLUS(ItemID.ANTIVENOM4_12913, GameTimerImageType.ITEM, "Anti-venom+", 3, ChronoUnit.MINUTES, true),
+	ANTIVENOMPLUS_ANTIPOSION(ItemID.SUPERANTIPOISON4, GameTimerImageType.ITEM, "Anti-venom+ Antipoison", 12, ChronoUnit.MINUTES),
```
Nightfirecat: Shouldn't this also be for 11 minutes, not 12?

deathbeam: no, because its 15 - 3

deathbeam:
Updated based on @Adam- 's suggestion because my way would not work (because of list modification while iterating it, unfortunate)

PR #7548, opened 2019-01-25
Switch xp bar intermediate markers to int

### LINKED ISSUE #7479: XP Tracker - Intermediate levels are off
The intermediate levels displayed on the experience tracker are not matching the experience bar progress.

notes marker

notes marker

As displayed with the above images I am 81 slayer and the progress bar is yet to meet the intermediate level line for 81 slayer.

paulbcook:
Resolves #7479 

This bug was occurring because the bar progress was stored as an int: [0, 100], while the intermediate line positions were stored as a double: [0, 1]. As the bar size is over 100 pixels, even if the progress is higher (but not by more than 1%), the line will render to the right.

This also has the benefit of simplifying the ProgressBar fields - both the value and the positions are now in the range of [0, maximumValue].

(This is also my first PR ever so be easy)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBoxManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBoxManager.java
@@ -77,17 +77,19 @@ public class InfoBoxManager
 	public void removeInfoBox(InfoBox infoBox)
 	{
 		log.debug("Removing InfoBox {}", infoBox);
-		infoBoxes.remove(infoBox);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
That explanation doesn't make sense, since I have reproduced the issue on Windows without changing any DPI settings.

forsco:
I noticed the issue originally on windows 10.

paulbcook:
You’re right, I was working on retina and misread the sizing. It’s always going to be over 100 pixels wide so the rounding errors should kick in on any client.

PR #7549, opened 2019-01-25
Slayer kill count fix

### LINKED ISSUE #1046: Monsters left on slayer task completely useless when barraging

daviscook477:
Fixes slayer count being wrong when bursting or using a canon.

Resolves #1046
Incorporates #6746

Example of the new code in action:

Starting a dust devil task.
start_of_dust_devils

Here is what the slayer plugin predicts is left at the halfway point.
half_way_through_predicted

After using the slayer helm to check task progress, the prediction turns out to be spot on.
half_way_through_actual

The task is completed without any error in count (the reason the debug says 1 error in 77 is because the completion message shows up before the last 1 xp drop. The 77 is because checking the slayer helm at 77 kc reset the error check count).
completed_accurately

Here is some debug information about what the slayer plugin now does:

It handles partial xp drops (where you get the kill or 50% damage but not both) by ensuring that if the player is nearby their current task any xp drop is treated as at least 1 kc. (however if the player isn't nearby their task an xp drop will not be counted as a kc - like a quest reward or lamp).
handles_partial_xp_drop

Debug log shows that when an xp drop is received it's compared to the xp drops that can be obtained from the on-task npcs that are currently nearby the player or were nearby the player recently (a few ticks).
debug_from_devils


daviscook477:
Explanation of changes for code review:

Main idea of the changes:
* we track which npcs can be killed for each task in the Task.java enum (runelite already did this) so we use this information to know which npcs near the player can kill for slayer xp.
* if we know which npcs may be dying to give the player slayer xp we can run the bounded knapsack problem over these potential slayer xp drops to try and determine which npcs died to produce the slayer xp drop that the client receives.
* decrease the task kc by the result of the knapsack solver rather than 1 each time we see a slayer xp drop
* this necessitated using #6746 to more accurately list out potential npcs that can be killed for each slayer task by changing task to associate each slayer task with the names of the npcs that can be killed and some specific outlier npcids.
* in order to determine which npcs are dying that could be giving the player slayer xp we run the list of npcs available to the client through the `isTarget` check to see which are actually npcs that can be killed on this task. These npcs are the potential xp drops. However if the player were for example killing some npc like the Alchemical Hydra and the npc despawned before the slayer xp drop came in to the client then this method would fail so we also must keep track of the npcs that despawned for some number of ticks after their despawning.

*KnapsackSolver.java*
New class that simply solves the bounded knapsack problem. Uses fairly straightforward implementation from wikipedia pseudocode.

*NPCPresence*
New class that is used to remember which npcs the player was nearby recently. Used as a part of tracking npcs for some number of ticks after despawning so we don't accidentally miss a potential xp source.

*SlayerConfig.java*
Added lastCertainAmount var to config that tracks the last amount of kc that we are certain the player needed to do. The things that make us certain is when the player is assigned a new task or they use a slayer gem or combat bracelet check. This was necessary as part of debugging/verifying the validity of these changes. By knowing what the last certain amount is the amount that are left over at the end of the task can be compared to see what the error rate is. I.e. for knowing the player had 77 dust devils left if the task  was completed when we thought there were 15 remaining we would have erred at 15/77.

*SlayerPlugin.java*
Lines related to the config have been changed to add the lastCertainAmount var. Code related to the lingeringPresences var is for tracking npcs after death to avoid missing a potential xp source. SlayerXpDropLookup is a utility class for looking up how much xp drops from a specific npc using the slayer xp drop file. The estimateKillCount method does the estimation using the KnapsackSolver. Two main things happen here. The xp drop is increased by 1 because xp drops are actually rational numbers not integers so we overestimate the xp drop so that a 133.3 xp giving npc can fit into a 133 xp drop. The numbers for all of the solver have to be multiplied by 10 because it works on integers and xp drops are floats with 1 decimal point. Main other change to SlayerPlugin is change to isTarget method. isTarget makes three checks. It can do a simple substring check to see if the npc name contains one of the target names. It can also do a check to see if the npc name exactly matches the target names. Third it can check if the npc id matches one of the target names.

*SlayerXpDropLookup.java*
Utility class for looking up how much xp drops from a specific npc using the file with that info.

*Task.java*
Task is changed to have npcids in addition to names.

*NPCHelper.java* and *SlayerPluginTests*
Helper for tests and tests for the changes.

Adam-:
fixed in 67030e38a44f355a0131cff5250ed45c7b01dc53

PR #7559, opened 2019-01-25
Convert game thread events to singletons

deathbeam:
This greatly reduces the amount of objects that need to be created literally all the time thanks to our event publishing.


Adam-:
I would probably rather not do this for any events with mutable fields unless we can prove noticeable GC related performance improvements.

deathbeam:
I did it only for events on game thread. I don't see why events with mutable fields should be an issue as long as we do not change them in process, and we dont atm.

deathbeam:
ok

PR #7567, opened 2019-01-26
Fix Desert Lizards not getting highlighted by the slayer plugin

### LINKED ISSUE #7565: Slayer target highlighting no longer works on Desert Lizards
This was caused by the Desert Lizards task being changed to just Lizards with the addition of Sulphur Lizards.

ShaunDreclin:
Resolves #7565


deathbeam:
You can just add "Desert Lizard" and "Sulphur Lizard" to list in the enum, why is the pattern change necessary?

ShaunDreclin:
The pattern change was because the task itself was renamed to lizards, so it would start highlighting lizardmen as well.

PR #7574, opened 2019-01-26
Force RUNELITE_OVERLAY menuops to not be left click

### LINKED ISSUE #7539: With only 'Cancel' option overlay right-click menu is left-click when below f.e bank interface
Title

Abextm:
Depends !45
Closes #7539

This makes it so RUNELITE_OVERLAY menu entries open the right click menu when left clicked on, similar to the `Remove all placeholder` button in the bank settings interface.


runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightOverlay.java
```diff
@@ -72,6 +73,13 @@ public Dimension render(Graphics2D graphics)
 		MenuEntry menuEntry = menuEntries[last];
 		String target = menuEntry.getTarget();
 		String option = menuEntry.getOption();
+		int type = menuEntry.getType();
+
+		if (type == MenuAction.RUNELITE_OVERLAY.getId() || type == MenuAction.EXAMINE_ITEM_BANK_EQ.getId())
```
Adam-: Can this just be a call to `shouldLeftClickOpenMenu`?

Abextm: We could, but that would make it never show the tooltip if you have single mouse button on. I'm not entirely sure what the behavior should be in that case.

PR #7591, opened 2019-01-28
Make the LootTrackerClient log when submitting isn't successful

Hydrox6:
Currently the LootTrackerClient will claim it has successfully submitted loot as long as a response came back.

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/loottracker/LootTrackerClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/loottracker/LootTrackerClient.java
@@ -76,7 +76,14 @@ public class LootTrackerClient
 			@Override
 			public void onResponse(Call call, Response response)
 			{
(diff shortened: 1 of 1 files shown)


Adam-:
I'd rather not do this since it might get loud if something breaks. Also you should not pass a format identifier `{}` for exceptions

Hydrox6:
Considering the 400 issue wasn't noticed for 3 days partially because the logs didn't notify the user that the loot submission was unsuccessful, it should get loud if something breaks.

Also I'm not using an exception, just the HTTP code and message, which for the 400 issue was `Bad Request`

Adam-:
This is something that is my issue due to lack of monitoring on the server logs.

Hydrox6:
Still, the client saying that the loot has been submitted when it's actually been rejected doesn't seem right

PR #7612, opened 2019-01-29
Use localized time format in time tracking panel

DropsOfSerenity:
12 hour clocks is a more common format and easier to parse at a glance. This pull requests makes them default for the time tracking tab and ads a config option for those who wants 24 hour time instead.

* Added a 12 hour clock option instead of military time as this is a more common clock format.
* Added a config option to the time tracking plugin for the 12 hour clock for those that want 24 hour time.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -77,6 +77,7 @@ public class RuneLite
 	public static final File RUNELITE_DIR = new File(System.getProperty("user.home"), ".runelite");
 	public static final File PROFILES_DIR = new File(RUNELITE_DIR, "profiles");
 	public static final File SCREENSHOT_DIR = new File(RUNELITE_DIR, "screenshots");
+	public static final Locale SYSTEM_LOCALE = Locale.getDefault();
 
 	@Getter
 	private static Injector injector;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -86,14 +86,7 @@ public abstract class TabContentPanel extends JPanel
 				sb.append(endTime.getDayOfWeek().getDisplayName(TextStyle.FULL, Locale.getDefault())).append(" ");
 			}
 
-			switch (OSType.getOSType()) {
-				case Windows:
-					sb.append("at ").append(WinApi.getTimeFormatString(endTime));
-					break;
-				default:
-					sb.append("at ").append(StackFormatter.getLocalizedDateTimeFormatter(FormatStyle.SHORT).format(endTime.toLocalTime()));
-					break;
-			}
+			sb.append("at ").append(StackFormatter.getPlatformFormattedDateTime(endTime));
 			return sb.toString();
 		}
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -26,15 +26,11 @@ package net.runelite.client.plugins.timetracking;
 
 import java.awt.Dimension;
 import java.time.LocalDateTime;
-import java.time.format.FormatStyle;
 import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.util.Locale;
 import javax.swing.JPanel;
-
-import net.runelite.client.util.OSType;
 import net.runelite.client.util.StackFormatter;
-import net.runelite.client.util.WinApi;
 
 public abstract class TabContentPanel extends JPanel
 {
@@ -86,7 +82,7 @@ public abstract class TabContentPanel extends JPanel
 				sb.append(endTime.getDayOfWeek().getDisplayName(TextStyle.FULL, Locale.getDefault())).append(" ");
 			}
 
-			sb.append("at ").append(StackFormatter.getPlatformFormattedDateTime(endTime));
+			sb.append("at ").append(StackFormatter.getPlatformTimeStringFromLocalDateTime(endTime));
 			return sb.toString();
 		}
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -31,6 +31,8 @@ import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.util.Locale;
 import javax.swing.JPanel;
+
+import net.runelite.client.util.OSType;
 import net.runelite.client.util.StackFormatter;
 
 public abstract class TabContentPanel extends JPanel
@@ -83,7 +85,14 @@ public abstract class TabContentPanel extends JPanel
 				sb.append(endTime.getDayOfWeek().getDisplayName(TextStyle.FULL, Locale.getDefault())).append(" ");
 			}
 
-			sb.append("at ").append(StackFormatter.getLocalizedDateTimeFormatter(FormatStyle.SHORT).format(endTime.toLocalTime()));
+			switch (OSType.getOSType()) {
+				case Windows:
+
+					break;
+				default:
+					sb.append("at ").append(StackFormatter.getLocalizedDateTimeFormatter(FormatStyle.SHORT).format(endTime.toLocalTime()));
+					break;
+			}
 			return sb.toString();
 		}
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -34,6 +34,7 @@ import javax.swing.JPanel;
 
 import net.runelite.client.util.OSType;
 import net.runelite.client.util.StackFormatter;
+import net.runelite.client.util.WinApi;
 
 public abstract class TabContentPanel extends JPanel
 {
@@ -87,7 +88,7 @@ public abstract class TabContentPanel extends JPanel
 
 			switch (OSType.getOSType()) {
 				case Windows:
-
+					sb.append("at ").append(WinApi.getTimeFormatString(endTime));
 					break;
 				default:
 					sb.append("at ").append(StackFormatter.getLocalizedDateTimeFormatter(FormatStyle.SHORT).format(endTime.toLocalTime()));
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -24,12 +24,12 @@
  */
 package net.runelite.client.plugins.timetracking;
 
-import java.util.Locale;
 import java.awt.Dimension;
 import java.time.LocalDateTime;
 import java.time.format.FormatStyle;
 import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
+import java.util.Locale;
 import javax.swing.JPanel;
 import net.runelite.client.util.StackFormatter;
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
@@ -26,6 +26,7 @@ package net.runelite.client.plugins.timetracking;
 
 import java.awt.Dimension;
 import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.util.Locale;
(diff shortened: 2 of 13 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -37,11 +37,22 @@
 	String TIMERS = "timers";
 	String STOPWATCHES = "stopwatches";
 
+	@ConfigItem(
+		keyName = "amPmTime",
+		name = "12 hour clock",
```
Nightfirecat:
```suggestion
		name = "Use 12-hour time format",
```

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -37,11 +37,22 @@
...
+		description = "Use 12 hour clock instead of 24 hour clock",
```
Nightfirecat:
```suggestion
		description = "Show remaining time in 12-hour time format instead of 24-hour time format",
```

Nightfirecat:
Per some discussion on discord, maybe this should be an app-wide config (under Runelite settings) instead of a plugin-specific config? There are multiple plugins which use different time formats, so this could apply to all of them.

DropsOfSerenity:
That sounds like a better option, I will work on this and have a PR soon.

DropsOfSerenity:
Alright @Nightfirecat I injected runeLiteConfig where needed and branched on that. Added a config option to the RuneLite config. Do I need to use this anywhere else? How does this look?

DropsOfSerenity:
@deathbeam Can you point in the right direction of an example of where you this functionality exists.

deathbeam:
We don't use it atm, but first result in google: http://tutorials.jenkov.com/java-internationalization/dateformat.html . But we set locale in main method atm so you would need to explicitely get system one

deathbeam:
See commit `96ce9898e413808ab41253c0d9a963a618445d75`. We previously had locale-based time in time tracking but we changed it due to bug elsewhere

deathbeam:
There is also this: https://stackoverflow.com/a/4594941 what is probably most useful in this case

deathbeam:
I did this locally and I think something like that could work: http://ix.io/1zui/diff . But can't test it atm

DropsOfSerenity:
Understood, looking at it now.

DropsOfSerenity:
@deathbeam How about this. This works and displays something like this.

image

deathbeam:
Yeah, that looks good. Except that you should not do this: https://github.com/runelite/runelite/pull/7612/files#diff-ecb629cba997669bc1265be9eecfd5f6R84 because we want day of week only in english. + you should import codestyle from here: https://github.com/runelite/runelite/wiki/Code-Conventions to intellij and update your PR with that

DropsOfSerenity:
@deathbeam I imported the codestyle and fixed the import *. Also what was there before was using system locale already not english day name.

deathbeam:
Yeah but it was intentionally changed to not do that and we dont want that part back

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
```diff
@@ -24,11 +24,14 @@
  */
 package net.runelite.client.plugins.timetracking;
 
+import net.runelite.client.RuneLite;
+import net.runelite.client.util.StackFormatter;
```
deathbeam: Fix your imports

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
```diff
@@ -78,9 +81,10 @@ protected static String getFormattedEstimate(long remainingSeconds, boolean useR
 			LocalDateTime currentTime = LocalDateTime.now();
 			if (endTime.getDayOfWeek() != currentTime.getDayOfWeek())
 			{
-				sb.append(endTime.getDayOfWeek().getDisplayName(TextStyle.FULL, Locale.getDefault())).append(" ");
+				sb.append(endTime.getDayOfWeek().getDisplayName(TextStyle.FULL, RuneLite.SYSTEM_LOCALE)).append(" ");
```
deathbeam: Dont use system locale for day of the week name

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -24,10 +24,14 @@
  */
 package net.runelite.client.util;
 
+import net.runelite.client.RuneLite;
```
deathbeam: Fix your imports

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
DropsOfSerenity: Replaced with `Locale.ENGLISH`

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
DropsOfSerenity: Moved to bottom of imports which is I assume what you mean.

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
DropsOfSerenity: Removed since no longer needed here.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TabContentPanel.java
DropsOfSerenity: Replaced with `Local.getDefault()` as that was what was there before.

DropsOfSerenity:
@Abextm Ok I've written a windows specific method and abstracted things a bit. I've tested on windows and linux, I do not have a mac to test.

It now respects the windows `Short Time Format` as specified here:

image

On linux and mac it uses the locale default time formats (Which is better than what we had there previously)

runelite-client/src/main/java/net/runelite/client/util/WinApi.java
```diff
@@ -0,0 +1,38 @@
...
+		WinDef.DWORD dwFlags = new WinDef.DWORD(Kernel32.TIME_NOSECONDS);
+
+		WinBase.SYSTEMTIME time = new WinBase.SYSTEMTIME();
+		time.wHour = (short) localDateTime.getHour();
+		time.wMinute = (short) localDateTime.getMinute();
+		time.wSecond = (short) localDateTime.getSecond();
+		time.wMilliseconds = 0;
+
+		kernel32.GetTimeFormatEx(null, dwFlags, time, null, lpTimeStr, 80);
+		return new String(lpTimeStr);
```
Abextm: The docs say nothing about this function working will null terminated strings

runelite-client/src/main/java/net/runelite/client/util/WinApi.java
```diff
@@ -0,0 +1,38 @@
...
+		char[] lpTimeStr = new char[80];
+		WinDef.DWORD dwFlags = new WinDef.DWORD(Kernel32.TIME_NOSECONDS);
+
+		WinBase.SYSTEMTIME time = new WinBase.SYSTEMTIME();
+		time.wHour = (short) localDateTime.getHour();
+		time.wMinute = (short) localDateTime.getMinute();
+		time.wSecond = (short) localDateTime.getSecond();
+		time.wMilliseconds = 0;
+
+		kernel32.GetTimeFormatEx(null, dwFlags, time, null, lpTimeStr, 80);
```
Abextm:
```suggestion
		kernel32.GetTimeFormatEx(null, dwFlags, time, null, lpTimeStr, lpTimeStr.length);
```

runelite-client/src/main/java/net/runelite/client/util/WinApi.java
```diff
@@ -0,0 +1,38 @@
...
+import com.sun.jna.platform.win32.WinDef;
+import com.sun.jna.win32.StdCallLibrary;
+import java.time.LocalDateTime;
+
+interface Kernel32 extends StdCallLibrary
+{
+	int TIME_NOSECONDS = 0x00000002;
+
+	int GetTimeFormatEx(
+			WString lpLocaleName, WinDef.DWORD dwFlags, WinBase.SYSTEMTIME lpTime,
```
Abextm: you can use `int` instead of `WinDef.DWORD`, which should save some allocations and make the code slightly simpler.

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -61,6 +65,36 @@
...
+		}
+	}
+
+	/**
+	 * Get a localized DateTimeFormatter for use.
+	 *
+	 * @param formatStyle The format style to use for the formatter
+	 * @return The localized DateTimeFormatter
+	 */
+	public static DateTimeFormatter getLocalizedDateTimeFormatter(FormatStyle formatStyle)
```
Abextm: private

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -61,6 +65,36 @@
...
 
+	/**
+	 * Attempts to call the platform to get a localized time string based on
+	 * the users preferences. Falls back on using locale default if it is on a
+	 * platform that has no consistent way of obtaining this information.
+	 *
+	 * @param localDateTime The LocalDateTime object to format as a string
+	 * @return The formatted string.
+	 */
+	public static String getPlatformTimeStringFromLocalDateTime(LocalDateTime localDateTime)
```
Abextm: This has nothing to do with stack formatting, it should be somewhere else. The method name is too long; don't include types in it, that is what the method signature is for.

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -77,6 +77,7 @@
 	public static final File RUNELITE_DIR = new File(System.getProperty("user.home"), ".runelite");
 	public static final File PROFILES_DIR = new File(RUNELITE_DIR, "profiles");
 	public static final File SCREENSHOT_DIR = new File(RUNELITE_DIR, "screenshots");
+	public static final Locale SYSTEM_LOCALE = Locale.getDefault();
```
Abextm: This is unsused

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
deathbeam: I would maybe rename the class to formatter. Or make TimeUtil with 1 method? But I like just formatter

runelite-client/src/main/java/net/runelite/client/RuneLite.java
deathbeam: Its still used for non-windows formatter

Abextm: oh.
This shouldn't exist at all; any users of it should just call Locale.getDefault().

deathbeam: Well, then it would be always US because of our Main method "hack" where we force set default locale

Abextm: then the place where we set the locale to us should set it to english, not us

deathbeam: Oops, its actually set to ENGLISH not us
```
		Locale.setDefault(Locale.ENGLISH);
```

but point of this change is so the formatter respects system locale, and on linux switching between locales is trivial (e.g I want AM/PM, I just change locale to US, I want 24hr, I change to UK).

Abextm: Most OSes have proper locale support, so that isn't reasonable for most users because that changes many other things, such as decimal formatting, spell checking/autocorrect, etc.

deathbeam: Well, that is my use-case. But still, it should respect system locale, as there is nothing like AM/PM switch on Linux (idk on Mac). And most people just prefer their system locale date format, so I dont think this is an issue. You have your windows switch, and I want my Linux switch.

Abextm: the default locale should be affected by the switch. We shouldn't have to maintain 2 locales for different things

deathbeam: It is the only way to get AM/PM setting on Linux (and as I already said, that is the point of this PR), so yes we do. and using static variable is less hacky than using WinAPI on Windows, so I think it is fine?

Abextm: having 2 static variables is worse than 1. we can just have 1 and be done with it. you said that it should already work without this

deathbeam: Without this it will always use Locale.ENGLISH instead of user-set locale that will determine AM/PM. ENGLISH is UK afaik, so it would be always 24hr if we would use Locale.getDefault() instead of this in StackFormatter

Abextm:
```
Locale.setDefault(new Locale.Builder()
	.setLocale(Locale.getDefault())
	.setLanguageTag("en")
	.build());
```

deathbeam: Hmmm, will that actually work? I guess tag is first part of this

```
export LANG="en_US.UTF-8"
export LC_ALL="en_US.UTF-8"
```

so it might.

PR #7620, opened 2019-01-30
Update slayer points in tooltip while in rewards

### LINKED ISSUE #7611: Buying slayer rewards doesn't decrement the points counter in the task infobox
The counter doesn't update until you finish your next task

daiya7:
Slayer points are updated while in the rewards menu, including before
and after a purchase. Previously, the points variable was updated on
each tick to whatever was displayed in the "Slayer points: "
widget, but that change was never reflected in the tooltip until the
next tooltip update. The tooltip now updates whenever a change in
slayer points is detected while in the rewards menu.

Closes #7611

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -372,7 +372,15 @@ public class SlayerPlugin extends Plugin
 				Matcher mPoints = REWARD_POINTS.matcher(w.getText());
 				if (mPoints.find())
 				{
+					int prevPoints = points;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -372,7 +372,15 @@ else if (braceletText.contains("expeditious bracelet"))
 				Matcher mPoints = REWARD_POINTS.matcher(w.getText());
 				if (mPoints.find())
 				{
+					int prevPoints = points;
```
Nightfirecat: This can be `final` since it's only assigned once.
```suggestion
					final int prevPoints = points;
```

PR #7623, opened 2019-01-30
Function key remapping

Keirathi:
I modified the WASD Camera plugin to also let you remap the function keys.

I realize that there have been some PRs for this in the past that were closed and never merged in, but the main argument I saw was "Just use AHK".  However, there are a couple of problems with that, especially if you want to use the 1-0 number keys.

1)  You have to be using the WASD Camera plugin already to use an AHK script unless you want it to be typing in the chatbox all day.

2) You either have to have a keybind to manually pause the AHK script when you actually want to chat, or you have to write all the logic to deal with actually wanting to chat.  WASD Camera already has that logic fleshed out, for the most part.

3)  If you want to use number keys, then when you have NPC  or crafting dialogs where you can use the number keys to progress, you again need to manually pause the script.  There's no real workaround for it that I can think of.  But if I do it in RuneLite, i can check if a DIALOG_NPC or DIALOG_OPTION widget is open and automatically disable my custom remaps.

So, while AHK is a 'workable' solution, it isn't ideal.  Hopefully something like this can be added into runelite.


Adam-:
Most of the hard parts of this was added in c227694ff0bf24f87902230e7d1b8e8534e81ec7. The keys aren't configurable though.

melkypie:
This was added in 3125b74f4fb4cbfb05e90d78a61c9ad216b1b08e

PR #7624, opened 2019-01-30
Edited Gargoyle task to include Marble Gargoyles + Comments

ftwfabian:
Just quick added Marble Gargoyles as part of the Gargoyle task

Also added some commenting to the constructors for the task plugin that the rest of the slayer plugins feed off of, would love some feedback if my style of commenting is preferred or if there is a different way of doing it that is better.


ftwfabian:
> Shouldn't marble gargoyles already be considered part of the task since they contain `gargoyle` in their name, so explicitly adding them to task enum doesn't actually do anything.

Ar

PR #7642, opened 2019-01-31
Fix color picker opening partly off screen when client is maximized

Keirathi:
Not a huge deal, but fixes the color picker dialogs opened by screen marker and chat color configs and keeps them from opening partially off screen.  Also opens them in slightly more logical place (centered on the button you clicked, instead of at the top of the window) all the time.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ui/ScreenMarkerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ui/ScreenMarkerPanel.java
@@ -492,6 +492,7 @@ class ScreenMarkerPanel extends JPanel
 		RuneliteColorPicker colorPicker = new RuneliteColorPicker(SwingUtilities.windowForComponent(this),
 			marker.getMarker().getFill(), marker.getMarker().getName() + " Fill", false);
 		colorPicker.setLocationRelativeTo(this);
+		colorPicker.fitInWindowBounds();
 		colorPicker.setOnColorChange(c ->
 		{
 			marker.getMarker().setFill(c);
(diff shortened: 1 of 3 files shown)


Keirathi:
Is that a problem?

Hmm, I'm not sure how to get the exact dimensions of a window in Java, but my estimation the window is 400x390 instead of 400x400.

Anyways, don't need the pack() in there to fix left or right off screen, but setPositionRelativeTo() made buttons very low on the screen open slightly off the bottom of the screen.  Not so far down that you couldn't still use all of the components, but it didn't look great.  pack() fixed that issue, but if it caused another then I wasn't able to find it.

raiyni:
Yes

runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ui/ScreenMarkerPanel.java
```diff
@@ -27,6 +27,7 @@
 
 import java.awt.BorderLayout;
 import java.awt.Color;
+import java.awt.Component;
```
raiyni: unused import

runelite-client/src/main/java/net/runelite/client/ui/components/colorpicker/RuneliteColorPicker.java
```diff
@@ -342,6 +344,15 @@ private void updateHex()
 		updatePanels();
 	}
 
+	public void fitInWindowBounds()
```
raiyni: this doesn't work on monitors that have a different resolution/scaling than main monitor.

raiyni:
added in https://github.com/runelite/runelite/commit/707ab78cdb6da090032c59f72e4ebfbc245126b1

PR #7645, opened 2019-02-01
Implement ItemUtil for plugins

TheStonedTurtle:
Seems like a few plugins do the same thing on `Item[]`s, and sometimes slightly differently so i created a Utility file for these use cases.


TheStonedTurtle:
Fixed merge conflicts

PR #6855, opened 2018-12-07
Add Right-Click rename feature to Bank Tags

dog-of-wisdom:
This is a first contribution (to runelite and an open source project in general) so I'm not too familiar with the runelite codebase. ~~Since this is the case, I wasn't too familiar with chatbox notifications and so when renaming a bank tag to a name that's already in use, the chatbox notification that shows is probably not optimal (see image below).~~ Edit: With the revision to how already-existing tags are handled in commit 5347d76, the chatbox notification is no longer an issue.

renaming

Closes #6486

PR #7652, opened 2019-02-01
Add right-click rename to Tag Tabs

### LINKED ISSUE #6486: Ability to right-click rename bank tags
Short version below.

I use bank tags for everything I do. I have tags for every slayer assignment, every daily... Pretty much, if I ever need to have the same gear at some point in the future, I'll make a tag for it. I've just ran into the following issue. I have a fire_giants tag for Fire Giants. These are usually pretty AFK-able, so I use ranged in the Waterfall dungeon. I now want to slay them using melee in the Catacombs of Kourend. Now, optimally, I'd rename fire_giants to fire_giants_ranged and make a new tag fire_giants_melee, but unfortunately, I have to take out each item from the fire_giants tag, make a new tag fire_giants_ranged, then mass-add them all there by right-clicking the Deposit Inventory button and clicking Tag-inventory fire_giants_ranged. I'd like to see the option to rename a tag on the right-click menu of a tag. This would allow the user to change the name, and all items with that tag would have the old tag removed and the new tag applied to them.

Short version: Right-click rename bank tag tabs.

raiyni:
Cleaned up some of the widget code, search code, and rename code from #6855.

Supersedes/closes #6855
Closes #6486

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
@@ -167,6 +167,20 @@ public class TagManager
 		}
 	}
 
+	public void renameTag(String oldTag, String newTag)
+	{
+		List<Integer> items = getItemsForTag(Text.standardize(oldTag));
+		items.forEach(id ->
+		{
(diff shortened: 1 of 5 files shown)


deathbeam:
ci is failing

PR #7653, opened 2019-02-01
Use Mark/Unmark for Ground Markers and Object Markers and Tag/Untag for NPC Indicators

### LINKED ISSUE #2184: Tile and NPC Marker tag/untag mark/unmark
Could NPCs and Tiles you've already marked have the "Tag" right-click option be changed to say "Un-tag"?
When there are multiple NPCs in close proximity, it can make it impossible to tell which NPC you're trying to remove a tag from. It can also be difficult to tell if you've right-clicked the correct tile, and you may end up marking *another* tile accidentally, when all you wanted was to remove the tag from it.

deathbeam:
Closes #2184


Nightfirecat:
LGTM at a glance. Add this change for object indicators as well?

deathbeam:
Updated for object indicators too

PR #7660, opened 2019-02-02
client: Make instance methods static where possible

Nightfirecat:
The methods modified are not synchronized, only reference static methods and fields, and are not overridden in subclasses.


deathbeam:
Provider methods for configuration being static look really odd to me.

Nightfirecat:
I agree it's a bit strange looking, but it should function correctly.

PR #7668, opened 2019-02-02
Improve Menu Entry Swapper Plugin

Spedwards:


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -63,11 +63,11 @@ default boolean swapAssignment()
 	}
 
 	@ConfigItem(
-		keyName = "swapBanker",
-		name = "Bank",
-		description = "Swap Talk-to with Bank on Bank NPC<br>Example: Banker"
+		keyName = "swapBankExchange",
```
15987632: changing keynames for no reason is probably bad, it would change peoples current config options

Puskunalis:
They do get swapped, if I recall correctly. It doesn't just get pushed up the list.

Spedwards:
@Puskunalis Could you please elaborate on what you mean? The current way is that your desired option is swapped with the current top item. This means if you're swapping the 1st option (say "Talk-to") with the 4th option (again, say "Examine"), the 1st option will be "Examine" and the 4th will be "Talk-to".

This PR will move the desired option instead of swapping, so using the same examples, "Examine" will be the 1st option, and "Talk-to" gets moved down to 2nd.

Puskunalis:
@Spedwards Sorry for the confusion, I thought it was just the name change!

Spedwards:
In case the commit message on 26d1fa7 was unclear, I hope these images helps.

How it currently works:


How this changes it:

Spedwards:
I would like to consolidate `claimSlime`, `swapDarkMage`, and `swapDecant` into another option, probably `swapTrade`, but I'm hesitant as none of them are trades.

My reasoning is because it seems pointless having these three options for NPCs that you don't talk to (outside of clues) when there could just be one.

I'll await feedback as to whether to consolidate into trade, or a new option for the three (and any future that meet this case).

Spedwards:
I have a branch that depends on this that would close 9 of the Menu Swap PRs while only adding 3 new config items.

PR #7675, opened 2019-02-03
Add Fish until next level to Fishing Overlay

fairyfoam48:
fishlvl


deathbeam:
I already mentioned this before, just use XP tracker sidebar

PR #7676, opened 2019-02-03
mlm plugin: fix duplicate ore counting

SirGirion:
This commit aims to fix the duplication of ores as described in issue
7641. There were 2 bugs leading to duplication. First, the flag for
determining if the inventory should be checked for ores was being set
any time the ore count varbit was changed, regardless of if it increased
or decreased. Second, the plugin was checking the entire inventory
instead of computing a delta of the inventory changes causing ores to be
counted twice if you withdrew from the sack before banking.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
@@ -26,7 +26,10 @@
  */
 package net.runelite.client.plugins.motherlode;
 
+import com.google.common.collect.HashMultiset;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multiset;
+import com.google.common.collect.Multisets;
 import com.google.inject.Provides;
 import java.time.Duration;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -199,7 +203,7 @@ public void onVarbitChanged(VarbitChanged event)
 		{
 			int lastSackValue = curSackSize;
 			refreshSackValues();
-			shouldUpdateOres = curSackSize != lastSackValue;
+			shouldUpdateOres = curSackSize < lastSackValue;
```
Adam-: Wouldn't you just have to create the snapshot here? And all it really has to do is keep track of each ore type and not take a snapshot of the whole inventory.

SirGirion: I suppose it might be a bit more efficient to take the snapshot here but the snapshot I take only counts the ores anyways, it doesn't track any non ore/nugget items.

PR #7677, opened 2019-02-03
Remove discord tag from party member name

TheStonedTurtle:


forsco:
That`s all good and well if you need the unique identifier for the plugin to function, but that portion doesn't need to be displayed on the overlays. Especially for longer names that span outside of the overlay box and rather than truncating the string removing the `#0000` seems better suited.

deathbeam:
It does, because you can have 2 different people with same name in party.

forsco:
Then we go with the in game name PR and update it to not remove the required data for the plugin to function and just change the rendered string on the overlay.

deathbeam:
As I said, this is bad idea

PR #7678, opened 2019-02-03
add logs until next level

fairyfoam48:
wc


deathbeam:
I already mentioned this before, just use XP tracker sidebar

PR #7682, opened 2019-02-03
World Map: Fix Zeah Magic tree and add missing ones

Puskunalis:


TheStonedTurtle:
You can rename pull requests and you have 2 commits with the same name, just squash the 2nd into the first.

PR #7705, opened 2019-02-04
Add .getBounds() to LayoutableRenderableEntity

deathbeam:


deathbeam:
@TheStonedTurtle

deathbeam:
Or actually, that is fine, whoever is using the API from overlay can just manually use overlay bounds as base.

Issue #7691, opened 2019-02-03
Add High Gamble Count for Barbarian Assault


rlw0014:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to show off my high gamble count in public chat.

**Describe the solution you'd like**
!gc (for gamble count) => Barbarian Assault High-level gambles: 9001

So the only places we could grab this information is from:
1) Adventurer Log in PoH(in counters)
2) Collection Log Book Minigames tab
3) Doing a High-level gamble
**Additional context**
![55c16c7d4a 1](https://user-images.githubusercontent.com/38933102/52183404-a61dd100-27d5-11e9-992c-ce010210f208.gif)
![9ff5ddce5a 1](https://user-images.githubusercontent.com/38933102/52183428-debdaa80-27d5-11e9-9353-85d3d2cf17c7.png)
![cf9e24981b 1](https://user-images.githubusercontent.com/38933102/52183520-8dfa8180-27d6-11e9-9758-711722636d81.png)

Aukk:
Why wasnt this added ?

TheFuries:
Wow! This feature would be a prodigious help! Why wasnt it added yet!

kschmanski:
I could have sworn I saw this in a PR months ago.  I'd be happy to work on it

Adam-:
Things get lost. There's no reason it can't be added. I'll look at #7706 this morning.

Adam-:
Added in #7706

PR #7706, opened 2019-02-04
Add BA high gamble count chat command

Jacoblairm:
Aimed at fixing issue #7691

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
@@ -216,4 +216,46 @@ public class ChatClient
 			return Integer.parseInt(response.body().string());
 		}
 	}
+
+	public boolean submitGc(String username, int gc) throws IOException
+	{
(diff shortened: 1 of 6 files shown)


deathbeam:
I think this should be part of kill count command

Adam-:
At the very least, it could internally use the `/kc` endpoint to save some code.

Adam-:
If `!kc penance queen` is already a thing I would just pick some other "boss name" like `ba high gamble` or something.

jmm140330:
We do not track penance queen kills with !kc, for whatever reason. Gamble count is commonly used by the ba community as well, I think the separate command like this would be a good fit?

Jacoblairm:
!kc chat command is registered to use killCountLookup() then killCountSubmit(), how could killCountSubmit() be used to submit the Gamble count considering it's just a varbit (similar to quest points)?

Also, a high gamble isn't directly related to killing a monster, "BA high gamble kill count: 355" would seem a little unusual.

Adam-:
@Jacoblairm have you confirmed this works outside of ba?

Adam-:
Alex on Discord checked it so we should be good.

PR #7713, opened 2019-02-05
Hide auto retaliate from combat options

### LINKED ISSUE #7686: Hide auto retaliate as addititon to "Attack Styles" plugin (for skillers)
Skillers currently have no option to hide the auto retaliate button, which could potentially ruin their account if they are attacked by a NPC and the retliate button is turned on by accident. 
The solution could be to create an "warn for auto retaliate" and add it to the "removed warned styles" option. In that way, skillers could never accidently turn on auto retaliate!

Mabel5:
Closes #7686


Tiadeche:
Hi, I noticed that hiding auto retaliate button causes the special attack bar to freeze. The special attack bar is low, but still shows it as full: 

https://i.imgur.com/p2deOkD.png

PR #7717, opened 2019-02-05
[WIP] add !hosts chat command

Spedwards:
Adds a `!hosts` chat command that will output players that are hosting in the area you've picked in your config.
The players output will have the items you've selected in your config, else none will be listed.

Hosts will be automatically added to the list upon entering their house at either Rimmington or Yanille provided their private chat is set to All and they are not an ironman. They must also have a Construction level at 67 or higher and either a Guilded Altar or Altar of the Occult.

To do:
- [x] Determine if house is player's and get owner's name
- [ ] Get Spirit Tree ID(s) for player owned houses
- [ ] Get Fairy Ring ID(s) for player owned houses


Lucwousin:
POH Fairy ring is `FAIRY_RING_29392`

PR #7719, opened 2019-02-05
Merge SetMessage and ChatMessage event into one

Adam-:


runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -1245,8 +1248,13 @@ public static void onAddChatMessage(int type, String name, String message, Strin
 			logger.debug("Chat message type {}: {}", ChatMessageType.of(type), message);
 		}
 
+		// Get the message node which was added
```
deathbeam: This methodHook needs to be run after the original method otherwise this will not work (or at least based on the addChatMessage code, it is added to buffer from inside of it and the method also assumes that buffer can be null and creates it when needed)

runelite-api/src/main/java/net/runelite/api/events/ChatMessage.java
```diff
@@ -39,6 +40,7 @@
  */
 @Data
 @AllArgsConstructor
+@NoArgsConstructor
```
deathbeam: As future improvement, this should be immutable, e.g `@Value`, with builder maybe? Because changing stuff on this is not working and can be confusing for people.

PR #7735, opened 2019-02-07
Add decorations to object marker plugin

### LINKED ISSUE #7711: Object marker doesn't work on mounted glory.
**Describe the bug**
When using the object marker on a mounted glory, it doesn't highlight.

**Screenshots**
alt-text

 - OS: Windows
 - RuneLite version: 1.5.10 (?)

Mabel5:
Fixes #7711

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
@@ -29,6 +29,7 @@ import java.awt.Graphics2D;
 import java.awt.Polygon;
 import javax.inject.Inject;
 import net.runelite.api.Client;
+import net.runelite.api.DecorativeObject;
 import net.runelite.api.GameObject;
 import net.runelite.api.TileObject;
 import net.runelite.client.ui.overlay.Overlay;
@@ -65,22 +66,31 @@ class ObjectIndicatorsOverlay extends Overlay
 			}
 
 			final Polygon polygon;
+			Polygon polygon2 = null;
 
 			if (object instanceof GameObject)
 			{
 				polygon = ((GameObject) object).getConvexHull();
+			}
+			else if (object instanceof DecorativeObject)
+			{
+				polygon = ((DecorativeObject) object).getConvexHull();
+				polygon2 = ((DecorativeObject) object).getConvexHull2();
 			}
 			else
 			{
 				polygon = object.getCanvasTilePoly();
 			}
 
-			if (polygon == null)
+			if (polygon != null)
 			{
-				continue;
+				OverlayUtil.renderPolygon(graphics, polygon, config.markerColor());
 			}
 
-			OverlayUtil.renderPolygon(graphics, polygon, config.markerColor());
+			if (polygon2 != null)
+			{
+				OverlayUtil.renderPolygon(graphics, polygon2, config.markerColor());
+			}
 		}
 
 		return null;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -180,12 +183,42 @@ public void onGameObjectSpawned(GameObjectSpawned event)
 		}
 	}
 
+	@Subscribe
+	public void onDecorativeObjectSpawned(DecorativeObjectSpawned event) {
+		final WorldPoint worldPoint = WorldPoint.fromLocalInstance(client, event.getDecorativeObject().getLocalLocation());
```
deathbeam: I am pretty sure you can just reuse the logic in gameObjectSpawned

Mabel5: that is the same logic from the GameObjectSpawned, should I put it in a private method or something?

deathbeam: yes

Mabel5: The changes have been made

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
```diff
@@ -70,6 +71,10 @@ public Dimension render(Graphics2D graphics)
 			{
 				polygon = ((GameObject) object).getConvexHull();
 			}
+			else if (object instanceof DecorativeObject)
+			{
+				polygon = ((DecorativeObject) object).getConvexHull();
```
Rheon-D: This will fail on diagonal walls with back to back decorative objects.
If you want to see this behaviour yourself check Juliets house, there is a painting on a diagonal wall on the ground floor.

Abex was looking into this.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -271,6 +293,12 @@ private TileObject findTileObject(Tile tile, int id)
 		}
 
 		final GameObject[] tileGameObjects = tile.getGameObjects();
+		final DecorativeObject tileDecorativeObject = tile.getDecorativeObject();
+
+		if (tileDecorativeObject != null)
```
Rheon-D: You need to check if the id of the object is the same as what was clicked otherwise if a GameObject exists on the same tile as a DecorativeObject you will tag the decorative instead of the GameObject.

An example of this would be a crate and a window on a wall occupying the same tile.

Mabel5: Ah right, I'll get this sorted in the morning.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
Adam-: Is this fixed now from #7824 ?

Abextm: with #7824 this just needs to render both convex hulls

PR #7739, opened 2019-02-07
Use HSB instead of RGBA to generate color from object

deathbeam:

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
@@ -231,16 +231,12 @@ public class ColorUtil
 	/**
 	 * Creates color from passed object hash code
 	 * @param object object with hashCode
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -231,16 +231,12 @@ public static Color fromHex(String hex)
...
-	public static Color fromObject(@Nonnull final Object object, boolean skipAlpha)
+	public static Color fromObject(@Nonnull final Object object)
 	{
 		int i = object.hashCode();
-		int r = (i >> 24) & 0xFF;
-		int g = (i >> 16) & 0xFF;
-		int b = (i >> 8) & 0xFF;
-		int a = i & 0xFF;
-		return new Color(r, g, b, skipAlpha ? 255 : a);
+		float h = (i % 359) / 360f;
```
Abextm: this should be %360

PR #7748, opened 2019-02-07
Add "true" tile tagging to the NPC highlight plugin

jesse1412:
Indicates the true tiles that tagged NPCs are currently on. Useful for anywhere animations are staggered or generally vague.

[In action](https://streamable.com/seamx)


jesse1412:
Re-opening since it's been over a year since the last discussion. I remember people talking about how the plugin was too powerful at verzik specifically, but her animations have been fixed for a long time now, so maybe this plugin is acceptable now.

Nightfirecat:
I've pushed a cleanup commit. Can you confirm this still works as you intended? @jesse1412

jesse1412:
> I've pushed a cleanup commit. Can you confirm this still works as you intended? @jesse1412

Seems to work correctly, tested on multiple different sized NPCs.

Adam-:
We've removed this in the past, see 79dda8b7d0bc773039e780708908aa2673d1b250 and https://twitter.com/JagexRoq/status/1189634926234230794. So I'd probably vote no.

PR #7754, opened 2019-02-08
Make sure WorldHopperPlugin's executor is active on startup

### LINKED ISSUE #6923: World hopper sidebar widget doesn't function if the plugin isn't enabled before startup.
If the world hopper plugin is disabled on startup, it will not function even when enabled unless it was on during launch.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to settings.
2. Disable "world hopper" plugin.
3. Enable "world hopper" plugin.
4. All "world hopper" widget functionality will be broken.
5. Restart client.
6. All "world hopper" widget functionality will work.

The world hopper should function when enabled and not require a client restart.

 - OS: Windows 10
 - RuneLite version: 1.5.5

henryzhangsta:
Fixes #6923

We can't submit tasks to shutdown Executors. These need to be recreated
on startup

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -131,7 +131,7 @@ public class WorldHopperPlugin extends Plugin
 	@Inject
 	private WorldHopperConfig config;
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -440,7 +436,7 @@ private void tick()
 		}
 	}
 
-	void refresh()
+	void refreshIfNeeded()
```
deathbeam: You should probably put this to separate commit

PR #7755, opened 2019-02-08
Random event notifier

dylanvoll:
This plugin will send a notification if any of the random event NPC's spawn near the player. Currently I do not think there is a way to tell who the random event is targeting until after a player interacts with it, so any other players random events will also trigger a notification.

The configuration for this plugin allows players to toggle which random events they would like to be notified about.


runelite-client/src/main/java/net/runelite/client/plugins/randomeventnotifier/RandomEventNotifier.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	@Subscribe
+	public void onNpcSpawned(NpcSpawned event)
+	{
+		final NPC npc = event.getNpc();
+		List<String> npcActions = Arrays.asList(npc.getComposition().getActions());
+		if (isRandomEventNPC(npc.getId()) && npcActions.contains("Dismiss"))
+		{
+			lastRandomEventNpc = npc;
```
robinwithes: I would probably remove
`
	@Subscribe
	public void onGameTick(GameTick event)
	{
		if (randomEventSpawned())
		{
			notifier.notify("[" + lastRandomEventNpc.getName() + "] has spawned!");
			lastRandomEventNpc = null;
		}
	}

	public boolean randomEventSpawned()
	{
		if (lastRandomEventNpc != null) return true;
		return false;
	}
`

And just handle everything in the OnNpcSpawned event.

robinwithes: I also think this plugin would not meet quality standards if it sends a notification even when a randomevent spawns for another player then the one using the plugin.

deathbeam:
Yeah it should definitely work only for random event npcs that spawned for you, otherwise it will not be accepted

Adam-:
This should probably be in the idle notifier. Did you try checking for random events spawning which are interacting with the local player? This would be the most sensible thing for most events. Except for a few things like strange plants, I think they always interact with the player theyre spawned for?

robinwithes:
> This should probably be in the idle notifier. Did you try checking for random events spawning which are interacting with the local player? This would be the most sensible thing for most events. Except for a few things like strange plants, I think they always interact with the player theyre spawned for?

Sometimes they randomly do not interact with the player. That was my experience at least when trying to make the same plugin a few months ago. (Even events like dr jekkyl)

dylanvoll:
I made some changes on determining the target player. Right now im using the overhead text of the NPC when they spawn. I believe this will work for every random except for evil bob/strange plant. I did some more experimenting, and I dont think there is one reliable way to determine the player. Maybe a special case can be found for strange plant/evil bob.

runelite-client/src/main/java/net/runelite/client/plugins/randomeventnotifier/RandomEventNotifier.java
dylanvoll: @robinwithes With the new overhead text implementation, I think keeping the on game tick check is necessary as the overhead text might not happen on the same game tick as the spawn event.

devLotto:
Your code is not passing checkstyle.

PR #7781, opened 2019-02-10
dulling removed to prevent difficult to see markers

### LINKED ISSUE #7170: xptracker: hunter immediate markers are almost invisible
https://twitter.com/Bearistdk/status/1080853514115375106

JAnken123:
closes #7170


Spedwards:
Would you like to post screenshots of what this now looks like?

JAnken123:
Markers are the same colors as foreground (the skill colors). Instead of 30% more dull as it were previously. I dont think a picture is necessary as it is straight forward to imagine what it looks like, but I can add one tomorrow.

Nightfirecat:
A picture of the difficult-to-see skill colors in question (hunter in this case) would be good.

JAnken123:
Hunter looks like this.
image

PR #7790, opened 2019-02-10
Add true current tile to tile indicator plugin

jesse1412:
Same concept as from [this close PR](https://github.com/runelite/runelite/pull/2764).

In action:

https://streamable.com/4uir6

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
@@ -74,4 +74,25 @@ public interface TileIndicatorsConfig extends Config
 	{
 		return false;
 	}
+
+	@Alpha
+	@ConfigItem(
+		keyName = "highlightCurrentColor",
+		name = "Color of current true tile highlighting",
+		description = "Configures the highlight color of current true tile"
+	)
+	default Color highlightCurrentColor()
+	{
+		return Color.CYAN;
+	}
+
+	@ConfigItem(
+		keyName = "highlightCurrentTile",
+		name = "Highlight current true tile",
+		description = "Highlights true tile player is on as seen by server"
+	)
+	default boolean highlightCurrentTile()
+	{
+		return false;
+	}
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
```diff
@@ -54,6 +54,27 @@ default boolean highlightDestinationTile()
...
+	}
+
+	@ConfigItem(
+			keyName = "highlightCurrentTile",
+			name = "Highlight current tile",
+			description = "Highlights tile player is on"
+	)
+	default boolean highlightCurrentTile()
+	{
+		return true;
```
deathbeam: This should be off by default

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
```diff
@@ -54,6 +54,27 @@ default boolean highlightDestinationTile()
...
 
+	@Alpha
+	@ConfigItem(
+			keyName = "highlightCurrentColor",
+			name = "Color of current tile highlighting",
+			description = "Configures the highlight color of current destination"
+	)
+	default Color highlightCurrentColor()
+	{
+		return Color.GRAY;
```
deathbeam: This should probably not be same color as destination color by default

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
```diff
@@ -54,6 +54,27 @@ default boolean highlightDestinationTile()
...
 
+	@Alpha
+	@ConfigItem(
+		keyName = "highlightCurrentColor",
+		name = "Color of current tile highlighting",
+		description = "Configures the highlight color of current tile position"
+	)
+	default Color highlightCurrentColor()
+	{
+		return Color.cyan;
```
deathbeam: use Color.CYAN

MagicfTail:
This would make a lot of content easier. A few obvious examples are Solo raids, ToB maze, and Vorkath.

jesse1412:
> This would make a lot of content easier. A few obvious examples are Solo raids, ToB maze, and Vorkath.

I would agree if the tile indicator plugin didn't already exist. This is probably less useful over all (the tile indicator plugin can provide the same information as this plugin where it counts and additionally shows you if you've misclicked the wrong tile).

MagicfTail:
Yeah, that is already bad enough imo. No need to make it even more obvious.

deathbeam:
I disagree with it making Vorkath easier, but idk about rest of points you mentioned.

melkypie:
This was added in runelite/plugin-hub#215

deathbeam:
Imo this should be merged to master and the external plugin removed

raiyni:
Does someone want to fix this up then?

PR #7798, opened 2019-02-11
ClueScrollPlugin - STASH unit indicator and highlighting

hex-agon:
This PR adds an indicator if the relevant STASH unit is built or not for the current emote clue. It also adds the ability for highlighting where the STASH unit is (configurable), here's two images of the changes in action:

image
image

I've data mined all of the locations & varbits from the cache so there shouldn't be any mistakes, I've done a few clues to test and everything seemed ok. I'm still looking for some feedback in the following parts:

* Varbit names
* Where i placed the stash unit indicator/the text I've used
* The color of the highlighted stash unit location


Abextm:
I would prefer if you would pull the varbit of the object composition, rather than storing it in an enum

Abextm:
Also you can just just search for the object in the scene to highlight it, which also takes care of the case of the lumbridge swamp caves stash, which is in 2 locations.

HexRealm:
Ok, here's what i'm thinking about doing:

* Revert the `STASHUnit` enum back to it's original state
* Change the `STASHUnit` class to an interface as it only holds constants
* Swap to using object composition to get the varbit
* Try finding a way to use the object spawn callback instead of scanning for the specific object

Thoughts?

Abextm:
All of our other `public static final`-only classes are just `public final class`, so just leave it like that

HexRealm:
I've ran into a problem, currently `ObjectComposition` does not exposes the varbit. Also the getter on the `Client` interface is called `getObjectDefinition` and not `getObjectComposition`.

HexRealm:
I've updated the PR with a simple change to support STASH units that appear at multiple locations.

HexRealm:
Any updates on this? @Abextm @deathbeam

PR #7810, opened 2019-02-11
Add config option to always show Discord party

deathbeam:


Adam-:
IMO this seems like a bad solution to not-yet-in-a-party invites not working problem

deathbeam:
Well there is no other solution. You either have counter shown and party invites work or you dont.

PR #7813, opened 2019-02-12
Add Guildmaster Jane to MenuEntrySwapper

rejectedpromise:
Allows for contract option on left-click, rather than talk-to.


Hydrox6:
#7789

rejectedpromise:
Implemented in #7789

PR #7823, opened 2019-02-12
Remove mustBeLoggedIn from client#isFriended

deathbeam:
Always use mustBeLoggedIn true when checking if passed name is friend.


deathbeam:
Mixins are :cloud:

PR #7832, opened 2019-02-12
Remove teleport and rub menu entry swaps

deathbeam:
These were replaced with shift-click configuration and variation
shift-click long time ago and are not necessary.


Adam-:
I don't think this would go over very well.

deathbeam:
Maybe, but it's for greater good 🤷‍♂️

PR #7842, opened 2019-02-13
metronome plugin: allow metronome to play sound effects while volume is muted

### LINKED ISSUE #530: Make Metronome work while game sounds are muted.
The metronome isn't very helpful for prayer flicking when it's completely drowned out by the sounds of all your prayers turning on and off!

Possible solutions:
Add a system for playing external wav/mp3 files
Add a 4th audio category to put RuneLite sounds in
Change the Metronome to use area sounds to allow sound effects to be muted.

trimbe:
Closes #530 
Depends on !47

Manually queues sound effects, bypassing the volume check the vanilla client normally does. 

Currently uses a medium high volume (2nd to last pip on the slider) but it may be more desirable to have this configurable or a different volume, not sure.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -983,6 +983,15 @@ public interface Client extends GameEngine
 	void playSoundEffect(int id, int x, int y, int range, int delay);
 
 	/**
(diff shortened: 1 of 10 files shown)


Reasel:
Could we get a bump on this? Would be nice to be able to work with the metronome without having to have game sounds on...

PR #7846, opened 2019-02-13
Add Swagger and Swagger UI support

deathbeam:
The swagger ui with api docs should be at <url>/swagger-ui.html (untested, I don't have http service set-up locally).

EDIT: Now tested:
screenie


deathbeam:
Closing in favour of #7847 as I got it to decent state

PR #7847, opened 2019-02-13
Add Swagger static document generation

deathbeam:
This is alternative method to #7846 

Preview: http://ix.io/1B11
(download and run in browser)


deathbeam:
I will update this later with static serving of this file from API instead of having to deploy it with releases (because its useful also for snapshots).

deathbeam:
Moved to 1.5.16 milestone

PR #7850, opened 2019-02-14
Use xdg-open for link browsing on Linux

Desetude:
Desktop#browse only supports GNOME environments on Linux. This PR adds functionality to use xdg-open, which should work for nearly every Linux environment.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
@@ -43,6 +43,8 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class LinkBrowser
 {
+	private static boolean shouldAttemptXdg = OSType.getOSType() == OSType.Linux;
+
 	/**
 	 * Tries to navigate to specified URL in browser. In case operation fails, displays message box with message
 	 * and copies link to clipboard to navigate to.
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -56,30 +56,67 @@ public static boolean browse(final String url)
...
+		{
+			log.debug("Opened browser through Desktop#browse to {}", url);
+			return true;
+		}
+
+		showMessageBox("Unable to open link. Press 'OK' and link will be copied to your clipboard.", url);
+		return false;
+	}
+
+	private static boolean attemptCommandBrowse(String command)
```
deathbeam: you should store if xdg-open exists in static variable or something instead of trying it all the time on linux

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -56,30 +56,67 @@ public static boolean browse(final String url)
 			return false;
 		}
 
+		if (OSType.getOSType() == OSType.Linux && attemptCommandBrowse("xdg-open " + url))
+		{
+				log.debug("Opened browser through xdg-open to {}", url);
+				return true;
+		}
+
+		if (attemptDesktopBrowse(url))
```
deathbeam: You should try desktop browse first, then try xdg-open and if nothing works show the message box

PR #7863, opened 2019-02-15
Prevent screenshot overwriting

### LINKED ISSUE #7840: Screenshots from Unrestricted/Beta Worlds are put into the screenshot folder for Normal Accounts
Notably this happens with boss kill screenshots.

Similar to #2050

NPittinger:
Closes #7840
The current datetime was not being appended to every screenshot filename, resulting in overwritten screenshots. This commit adds missing datetime values to filenames and switches to the local datetime, which should be more meaningful to the user than UTC.
 - Use LocalDateTime API and DateTimeFormatter
 - Append missing formatted date to file name strings
 - Remove duplicate call to takeScreenshot

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -41,10 +41,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.InvocationTargetException;
-import java.text.DateFormat;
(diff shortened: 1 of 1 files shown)


raiyni:
I'm going to leave my comments from Discord here.

1. Screenshots should be saved to a login name folder because you can transfer the name to a lower level account and overwrite all of your screenshots that way too.

2. Check for tournament world type.

raiyni:
`* If someone wants to change the organization of the screenshot folders at a later date, nothing will stop them from doing that. The question of the best screenshot folder structure is unrelated to the immediate problem.`

It's not about organization. It's about the screenshots being in the right place in the first place.

PR #7867, opened 2019-02-15
Change screenshot folder structure

### LINKED ISSUE #7840: Screenshots from Unrestricted/Beta Worlds are put into the screenshot folder for Normal Accounts

NPittinger:
Closes #7840
Screenshots from different world types but with the same name were being saved to the same directory, resulting in overwritten screenshots. This commit adds world type folders to the screenshot directory structure and switches to using the local datetime, which should be more meaningful to the user than UTC.

 - Use username and world type as subfolders of the default screenshot directory
 - Check if screenshot file already exists, and append a number to filename in that instance
 - Use LocalDateTime and DateTimeFormatter APIs
 - Remove duplicate call to takeScreenshot
 - Fix camelCase spelling of test method

(Sorry, accidentally closed my previous pull request while merging.)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -41,10 +41,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.InvocationTargetException;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.time.LocalDate;
-import java.util.Date;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.EnumSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
@@ -120,7 +118,7 @@ public class ScreenshotPlugin extends Plugin
 	private static final HttpUrl IMGUR_IMAGE_UPLOAD_URL = HttpUrl.parse("https://api.imgur.com/3/image");
 	private static final MediaType JSON = MediaType.parse("application/json");
 
-	private static final DateFormat TIME_FORMAT = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss");
+	private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
 
 	private static final Pattern NUMBER_PATTERN = Pattern.compile("([0-9]+)");
 	private static final Pattern LEVEL_UP_PATTERN = Pattern.compile(".*Your ([a-zA-Z]+) (?:level is|are)? now (\\d+)\\.");
@@ -132,12 +130,9 @@ public class ScreenshotPlugin extends Plugin
 		"You feel something weird sneaking into your backpack",
 		"You have a funny feeling like you would have been followed");
 
-	static String format(Date date)
+	static String format(LocalDateTime date)
 	{
-		synchronized (TIME_FORMAT)
-		{
-			return TIME_FORMAT.format(date);
-		}
+		return date.format(TIME_FORMAT);
 	}
 
 	private String clueType;
@@ -196,7 +191,7 @@ public class ScreenshotPlugin extends Plugin
 		@Override
 		public void hotkeyPressed()
 		{
-			takeScreenshot(format(new Date()));
+			takeScreenshot(format(LocalDateTime.now()));
 		}
 	};
 
@@ -219,7 +214,7 @@ public class ScreenshotPlugin extends Plugin
 			.tab(false)
 			.tooltip("Take screenshot")
 			.icon(iconImage)
-			.onClick(() -> takeScreenshot(format(new Date())))
+			.onClick(() -> takeScreenshot(format(LocalDateTime.now())))
 			.popup(ImmutableMap
 				.<String, Runnable>builder()
 				.put("Open screenshot folder...", () ->
@@ -294,7 +289,7 @@ public class ScreenshotPlugin extends Plugin
 	{
 		if (config.screenshotPlayerDeath())
 		{
-			takeScreenshot("Death " + format(new Date()));
+			takeScreenshot("Death " + format(LocalDateTime.now()));
 		}
 	}
 
@@ -305,7 +300,7 @@ public class ScreenshotPlugin extends Plugin
 		{
 			final Player player = playerLootReceived.getPlayer();
 			final String name = player.getName();
-			String fileName = "Kill " + name + " " + format(new Date());
+			String fileName = "Kill " + name + " " + format(LocalDateTime.now());
 			takeScreenshot(fileName);
 		}
 	}
@@ -373,7 +368,7 @@ public class ScreenshotPlugin extends Plugin
 
 		if (config.screenshotPet() && PET_MESSAGES.stream().anyMatch(chatMessage::contains))
 		{
-			String fileName = "Pet " + format(new Date());
+			String fileName = "Pet " + format(LocalDateTime.now());
 			takeScreenshot(fileName);
 		}
 
@@ -395,7 +390,7 @@ public class ScreenshotPlugin extends Plugin
 			if (m.matches())
 			{
 				String valuableDropName = m.group(1);
-				String fileName = "Valuable drop " + valuableDropName + " " + format(new Date());
+				String fileName = "Valuable drop " + valuableDropName + " " + format(LocalDateTime.now());
 				takeScreenshot(fileName);
 			}
 		}
@@ -406,7 +401,7 @@ public class ScreenshotPlugin extends Plugin
 			if (m.matches())
 			{
 				String untradeableDropName = m.group(1);
-				String fileName = "Untradeable drop " + untradeableDropName + " " + format(new Date());
+				String fileName = "Untradeable drop " + untradeableDropName + " " + format(LocalDateTime.now());
 				takeScreenshot(fileName);
 			}
 		}
@@ -461,7 +456,7 @@ public class ScreenshotPlugin extends Plugin
 		{
 			case KINGDOM_GROUP_ID:
 			{
-				fileName = "Kingdom " + LocalDate.now();
+				fileName = "Kingdom " + format(LocalDateTime.now());
 				break;
 			}
 			case CHAMBERS_OF_XERIC_REWARD_GROUP_ID:
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -53,6 +53,8 @@ import java.util.regex.Pattern;
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
 import javax.swing.SwingUtilities;
+
+import joptsimple.internal.Strings;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


raiyni:
You don't need to open a new pr. Just push commits and edit your post if you need to.

NPittinger:
While dealing with a conflicting merge while rebasing, I accidentally closed the previous pull request without a way to reopen it. Now I know how to avoid that in the future. Lesson learned. 😅

deathbeam:
Change looks good, but, any reason for the dateTime changes? what was wrong with existing API, and what is the difference in new api and previous.

NPittinger:
The biggest reason to change is that the newer APIs use immutable classes that provide thread safety. It also makes it much easier and clearer to to use local time, which should be more meaningful to the user.

deathbeam:
Hmm I guess that makes sense. But you should put that at least to separate commit.

NPittinger:
Do you want me to move it to a separate commit, or should I just keep that in mind moving forward?

deathbeam:
Yeah, move the date change to separate commit and title it accordingly (+ explain why in the commit description)

NPittinger:
I believe I've successfully separated the date API commits! 😄

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -621,25 +620,16 @@ private void takeScreenshot(String fileName, Image image)
 		// Draw the game onto the screenshot
 		graphics.drawImage(image, gameOffsetX, gameOffsetY, null);
 
-		File playerFolder;
-		if (client.getLocalPlayer() != null && client.getLocalPlayer().getName() != null)
+		File playerFolder = SCREENSHOT_DIR;
+		if (client.getUsername() != null)
```
deathbeam: You should check for Strings.isNullOrEmpty (i dont think the username can ever be null anyway)

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -621,25 +620,16 @@ private void takeScreenshot(String fileName, Image image)
...
 		{
-			final EnumSet<WorldType> worldTypes = client.getWorldType();
-			final boolean dmm = worldTypes.contains(WorldType.DEADMAN);
-			final boolean sdmm = worldTypes.contains(WorldType.SEASONAL_DEADMAN);
-			final boolean dmmt = worldTypes.contains(WorldType.DEADMAN_TOURNAMENT);
-			final boolean isDmmWorld = dmm || sdmm || dmmt;
-
-			String playerDir = client.getLocalPlayer().getName();
-			if (isDmmWorld)
+			String sanitizedUsername = client.getUsername().replaceAll("[\\W]", "_");
```
deathbeam: Just use Text.standardize() and then .replace(' ', '_')

deathbeam: mark this final

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -621,25 +620,16 @@ private void takeScreenshot(String fileName, Image image)
...
+			playerFolder = new File(playerFolder, sanitizedUsername);
+			String worldType = getWorldFolderType();
```
deathbeam: mark this final

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -716,6 +711,49 @@ public void onResponse(Call call, Response response) throws IOException
...
+	 * Determines the world type folder, if any, to which the
+	 * screenshot should be saved.
+	 *
+	 * @return the world type folder name as a string,
+	 * or {@code null} if the world type is a normal world
+	 * (Free, Members, PVP, High-Risk PVP)
+	 */
+	private String getWorldFolderType()
+	{
+		EnumSet<WorldType> worldTypes = client.getWorldType();
```
deathbeam: mark this final

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
NPittinger: ✔️

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
NPittinger: ✔️

NPittinger: ✔️

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
NPittinger: ✔️

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
NPittinger: ✔️

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -53,6 +53,8 @@
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
 import javax.swing.SwingUtilities;
+
+import joptsimple.internal.Strings;
```
deathbeam: Wrong import, and also newline

chivesrs: com.google.common.base.Strings?

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -621,25 +617,16 @@ private void takeScreenshot(String fileName, Image image)
 		// Draw the game onto the screenshot
 		graphics.drawImage(image, gameOffsetX, gameOffsetY, null);
 
-		File playerFolder;
-		if (client.getLocalPlayer() != null && client.getLocalPlayer().getName() != null)
+		File playerFolder = SCREENSHOT_DIR;
+		if (!Strings.isNullOrEmpty(client.getUsername()))
```
chivesrs: Could you check the behavior of `getUsername()`? From what I read in the docs, if the scenario was:

username `bob@bobmcbobface.com` (post 2011?) or `bobbyplayer` (pre 2011?)
password `redacted`
in game display name `Bob McBobFace`

Then you'd get back `bob@bobmcbobface.com` or `bobbyplayer`, assuming I understand the docs correctly.

It looks like this would inadvertently expose an email address or a username. I know I don't want my username floating around my file system given it's sensitive information, especially when it comes to shared computers. Prefer reverting back to old behavior for this conditional?

chivesrs: Be sure to change line 623 below as necessary.

chivesrs:
I have some reservations on using username instead of display name, I'll post them on the review thread.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -716,6 +708,49 @@ public void onResponse(Call call, Response response) throws IOException
...
 
+	/**
+	 * Determines the world type folder, if any, to which the
+	 * screenshot should be saved.
+	 *
+	 * @return the world type folder name as a string,
+	 * or {@code null} if the world type is a normal world
+	 * (Free, Members, PVP, High-Risk PVP)
+	 */
+	private String getWorldFolderType()
```
chivesrs: Consider Optional<String> getSubdirectoryName()?

Maybe change Javadoc to something like:
/** Determines the subdirectory name (if any) that the screen should be saved. This is determined by the type of the world (Deadman, Last man standing, etc) that the player is on. */

And update the caller accordingly.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -716,6 +708,49 @@ public void onResponse(Call call, Response response) throws IOException
...
+			return "last-man-standing";
+		}
+		else if (worldTypes.contains(WorldType.SKILL_TOTAL))
+		{
+			return "skill-total";
+		}
+		else if (worldTypes.contains(WorldType.BOUNTY))
+		{
+			return "bounty";
+		}
```
chivesrs: Should `SKILL_TOTAL` and `BOUNTY` be their own subdirectories? Are they considered normal worlds in terms of stats?

Same with `LAST_MAN_STANDING`, but I'm not sure how LMS works because if I remember correctly, Jagex does some stat faking, so that block might be needed.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
chivesrs: Looks like markdown ate my comment, updated.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
deathbeam: Well, the issue with current approach is that if you change your display name in-game, then screenshot will start saving in different folder, and if you for some reason pick display name that you were using on other account before, screenshot plugin will start overwriting those, effectively deleting your data.

chivesrs: Yeah, they both sound like hairy situations. But, how will the file get overwritten? Time will be part of the file name, and if (somehow, because dumb luck with time and/or clock skew) time was identical, a _1 or _2 would be appended?

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -648,6 +635,11 @@ private void takeScreenshot(String fileName, Image image)
 		{
 			File screenshotFile = new File(playerFolder, fileName + ".png");
 
+			for (int offset = 1; screenshotFile.exists(); offset++)
```
chivesrs: Consider adding a comment above this block to explain the appending. I naively thought that time would be unique (it's not). Maybe like:

// While time is part of the file name, events like Daylight Savings Time can cause non unique LocalDateTimes to occur. To avoid overwriting files, append _1, _2, etc to the end of the file name.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
deathbeam: Well, with this patch they will not be overwriten, on master they are. But still that do not solves the issue of different accounts being mixed in 1 screenshot folder, that is supposed to be unique per-account.

chivesrs: I feel you, I just don't like exposing username. Another scenario I had in mind is, you know how all the Twitch streamers hide their username and have the client auto remember so they never show it on stream? Maybe one day if they want to show a screenshot from their screenshot archive to the stream, the username will be visible in all its glory. That would not be ideal.

PR #7869, opened 2019-02-15
Use default PanelComponent size in Cooking plugin

deathbeam:
So it is consistent with rest of plugins, mainly session ones.


deathbeam:
For context:

screenie

This looks like :hankey:

PR #7884, opened 2019-02-16
Update deselected wiki icon, add hover state

Nightfirecat:
Fixed viewport default and hovered icons:
fixed-default fixed-hover

Resizable viewport default and hovered icons:
resizable-default resizable-hover


Abextm:
`Lotto: that looks worse`
`Hydrox: no outline makes it harder to read`
`deathbeam: no outline looks better. its not fixing the core problem, but at least its less contrasting with everything else around it`

Nightfirecat:
@devLotto @Hydrox6 Would updating this PR to use an icon with a more subtle outline be an acceptable solution? (that is, white-tinted, but not 100% white like the image currently uses)

Nightfirecat:
I've updated the images to have a soft white outline, which roughly matches the design of nearby UI elements such as the minimap and minimap orbs.

Nightfirecat:
@devLotto Does the updated image address your concern adequately?

devLotto:
The updated icon looks nice, maybe the hovered one should be a bit less lit up so the beveled details on the Wiki text are still there.

devLotto:
I still think the hover image has too much detail lost, is this better?

wiki_hover

Nightfirecat:
I don't feel very strongly on this, but I would think that having a more vibrant hover state would be helpful so people know that there are available actions when it is hovered.

Nightfirecat:
Seems like nobody cares if this gets added, and since we've removed our sprite (since there's an ingame sprite now) I don't really want to keep maintaining this.

PR #7895, opened 2019-02-17
Replace deprecated calls

NPittinger:
These commits replace deprecated method calls with the recommended replacements as denoted by their Javadocs.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
@@ -25,7 +25,6 @@
 package net.runelite.client.plugins.wiki;
 
 import com.google.common.primitives.Ints;
-import java.net.URLEncoder;
 import java.util.Arrays;
 import javax.inject.Inject;
 import javax.inject.Provider;
@@ -70,12 +69,25 @@ public class WikiPlugin extends Plugin
 			WidgetInfo.QUESTLIST_MINIQUEST_CONTAINER.getId(),
 		};
 
-	static final String WIKI_BASE = "https://oldschool.runescape.wiki";
-	static final HttpUrl WIKI_RSLOOKUP = HttpUrl.parse(WIKI_BASE + "/w/Special:Lookup");
-	static final HttpUrl WIKI_API = HttpUrl.parse(WIKI_BASE + "/api.php");
+	static final String WIKI_SCHEME = "https";
+	static final String WIKI_HOST = "oldschool.runescape.wiki";
+	static final HttpUrl WIKI_BASE = new HttpUrl.Builder()
+			.scheme(WIKI_SCHEME)
+			.host(WIKI_HOST)
+			.addPathSegment("w")
+			.build();
+	static final HttpUrl WIKI_SEARCH = WIKI_BASE.newBuilder()
+			.addPathSegment("Special:Search")
+			.build();
+	static final HttpUrl WIKI_RSLOOKUP = WIKI_BASE.newBuilder()
+			.addPathSegment("Special:Lookup")
+			.build();
+	static final HttpUrl WIKI_API = WIKI_BASE.newBuilder()
+			.removePathSegment(0)
+			.addPathSegment("api.php")
+			.build();
 	static final String UTM_SORUCE_KEY = "utm_source";
 	static final String UTM_SORUCE_VALUE = "runelite";
-	static final String UTM_PARAMS = UTM_SORUCE_KEY + "=" + UTM_SORUCE_VALUE;
 
 	private static final String MENUOP_GUIDE = "Guide";
 	private static final String MENUOP_QUICKGUIDE = "Quick Guide";
@@ -224,7 +236,7 @@ public class WikiPlugin extends Plugin
 
 			HttpUrl url = WIKI_RSLOOKUP.newBuilder()
 				.addQueryParameter("type", type)
-				.addQueryParameter("id", "" + id)
+				.addQueryParameter("id", String.valueOf(id))
 				.addQueryParameter("name", name)
 				.addQueryParameter(UTM_SORUCE_KEY, UTM_SORUCE_VALUE)
 				.build();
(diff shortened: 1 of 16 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiSearchChatboxTextInput.java
```diff
@@ -297,7 +299,14 @@ public void keyPressed(KeyEvent ev)
 
 	private void search(String search)
 	{
-		LinkBrowser.browse(WikiPlugin.WIKI_BASE + "?search=" + URLEncoder.encode(search) + "&" + WikiPlugin.UTM_PARAMS);
+		try
+		{
+			LinkBrowser.browse(WikiPlugin.WIKI_BASE + "?search=" + URLEncoder.encode(search, Charset.defaultCharset().name()) + "&" + WikiPlugin.UTM_PARAMS);
```
Abextm: you have just inlined the deprecated method. Try reading the deprecation notice as to why this is wrong.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiSearchChatboxTextInput.java
```diff
@@ -297,7 +299,14 @@ public void keyPressed(KeyEvent ev)
...
+		}
+		catch (UnsupportedEncodingException e)
+		{
+			log.warn("error searching wiki", e);
+		}
 		chatboxPanelManager.close();
```
Spedwards: This should be in a `finally` block.
```suggestion
		finally
		{
			chatboxPanelManager.close();
		}
```

Abextm: no it should not

NPittinger: Agreed, there is no reason to have a finally block here.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiSearchChatboxTextInput.java
NPittinger: I have not inlined the deprecated method. The deprecated method used here was `encode(String s)`. The change I committed uses the method in the deprecation notice, `encode(String s, String enc`.

Abextm: the way you have used `encode(String, String)` is exactly what `encode(String)` does. You have inlined it. It is incorrect for the same reason that `encode(String)` is deprecated.

deathbeam: Use HttpUrl.build or HttpUrl.parse

NPittinger: Yes, they do the same thing, but `encode(String, String)` is not deprecated, whereas `encode(String)` is. That's the point. However, if you would prefer I change the code to use the HttpUrl class, I will do so.

Abextm: Things are not deprecated arbitrarily. calling `encode(String, Charset.defaultCharset().name())` is the same as `encode(String)`, but without a warning. The bug that the warning warns you about still exists. IMO your way is worse because it hides a valid bug. That being said, the correct thing to do is use HttpUrl.Builder, so do that

NPittinger: Commit e46a69d now uses `HttpUrl.parse`.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiSearchChatboxTextInput.java
```diff
@@ -297,7 +296,7 @@ public void keyPressed(KeyEvent ev)
 
 	private void search(String search)
 	{
-		LinkBrowser.browse(WikiPlugin.WIKI_BASE + "?search=" + URLEncoder.encode(search) + "&" + WikiPlugin.UTM_PARAMS);
+		LinkBrowser.browse(HttpUrl.parse(WikiPlugin.WIKI_BASE + "?search=" + search + "&" + WikiPlugin.UTM_PARAMS).toString());
```
Abextm: 🤦‍♂️

deathbeam:
```
		new HttpUrl.Builder()
			.addQueryParameter("blah")
			.addQueryParameter("blaah")
			.build()
			.toString()
```

NPittinger: `HttpUrl.parse` uses the `HttpUrl.Builder` under the hood, but if you really want it to be used directly, I can do that too.

Abextm: Maybe you should do some reading on URL encoding and why it is required before attempting this change.

NPittinger: I already knew what URL encoding is, but upon review, I see why using `HttpUrl.parse` in that way would not work properly - my mistake. Commit 5b3708d now uses `HttpUrl.Builder`.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -224,7 +236,7 @@ private void onMenuOptionClicked(MenuOptionClicked ev)
 
 			HttpUrl url = WIKI_RSLOOKUP.newBuilder()
 				.addQueryParameter("type", type)
-				.addQueryParameter("id", "" + id)
+				.addQueryParameter("id", String.valueOf(id))
```
raiyni: why change this line

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -70,12 +69,25 @@
 			WidgetInfo.QUESTLIST_MINIQUEST_CONTAINER.getId(),
 		};
 
-	static final String WIKI_BASE = "https://oldschool.runescape.wiki";
-	static final HttpUrl WIKI_RSLOOKUP = HttpUrl.parse(WIKI_BASE + "/w/Special:Lookup");
-	static final HttpUrl WIKI_API = HttpUrl.parse(WIKI_BASE + "/api.php");
+	static final String WIKI_SCHEME = "https";
```
Abextm: don't change this

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -235,16 +247,24 @@ private void onMenuOptionClicked(MenuOptionClicked ev)
...
-					quickguide = "/Quick_guide";
+					quickGuide = true;
 					//fallthrough;
 				case MENUOP_GUIDE:
 					ev.consume();
-					String quest = Text.removeTags(ev.getMenuTarget());
-					LinkBrowser.browse(WIKI_BASE + "/w/" + URLEncoder.encode(quest.replace(' ', '_')) + quickguide + "?" + UTM_PARAMS);
+					String quest = Text.removeTags(ev.getMenuTarget())
+							.replace(' ', '_');
+					quest = (quickGuide) ? quest + "/Quick_guide" : quest;
```
raiyni: why change the logic on this

Issue #1392, opened 2018-04-11
Duplicate fishing sprites


Bartvhelvert:
![image](https://user-images.githubusercontent.com/9639072/38645873-f72a5134-3de5-11e8-98c1-c5ea9a2b37e5.png)
![image](https://user-images.githubusercontent.com/9639072/38645880-fdcfcb18-3de5-11e8-9e5e-4e1c9e8cfb74.png)

Happens when there is more than 1 fishing spot in the same tile.

ShaunDreclin:
Was just about to report this myself haha

![screenshot](https://user-images.githubusercontent.com/5851433/38720186-405f2df6-3ec3-11e8-8e71-5f5e4ced7263.png)

SomeoneWithAnInternetConnection:
I want to invoke "it's a feature! not a bug" here. As long as the second icon isn't 3 miles up in the sky (which sometimes happened in the past), showing two when there are two fishing spots in the same tile seems fairly reasonable, since there are two fishing spots there after all.

Nightfirecat:
It seems redundant to show two spots of the same type. If they're for different fish, it makes sense to show both.

MESLewis:
But it is useful to know that 2 spots are occupying the same tile. I don't know about anyone else but usually when I'm fishing I know how many spots there are that I care about and I keep track of where all of them are to an extent so it's nice to know if in trying to find where all 4 spots are that 2 are in the same tile.

Bartvhelvert:
Maybe it can be shown in another way which looks less bugged. Currently it's also impossible to see how many spots there are in one place.

SomeoneWithAnInternetConnection:
The current behaviour doesn't look particularly buggy to my eyes, and also shows you how many spots there are in one place?

Bartvhelvert:
As shown in the picture on my initial post there are 3 fishing spots in one place with only 2 icons.

ypperlig:
What about instead of it floating in the air they could be shown side by side? Example:
![fishing](https://user-images.githubusercontent.com/4660715/38755196-20f7181e-3f65-11e8-8a00-0b4523b94266.png)

Bartvhelvert:
That would be confusing I think because it wouldn't match the clickbox. I suggest just having a text overlay if the amount of spots is bigger than one.

ShaunDreclin:
Having the icons stack closer together would also work, the main thing that looks wrong with them is how high the second icon floats above the first one

[...]

PR #7897, opened 2019-02-17
Fishing Plugin: Duplicate fishing icons fix

nrodriguesdev:
This bug fix addresses #1392.

**Problem:** The fishing plugin doesn't keep track of the fishing spots near the player. As a result, it tries to draw multiple icons in the same spot whenever fishing spots are stacked. 

**Fix:** I implemented 2 HashMaps to keep track of core information regarding each fishing spot. The logical height is no longer used to determine the height of text/icons. Instead, a fixed value (derived from the logical height of a fishing spot) is used to avoid odd logical heights caused by stacked spots.

image

**Concerns**: I'm not confident that using two HashMaps is the right approach. Any feedback on this will be appreciated.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingSpotOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingSpotOverlay.java
@@ -61,7 +61,8 @@ class FishingSpotOverlay extends Overlay
 	private final Client client;
 	private final ItemManager itemManager;
 
-	private static HashMap<WorldPoint, HashMap<NPC, Integer>> SPOTS_LIST = new HashMap<>();
+	private static HashMap<WorldPoint, HashMap<FishingSpot, NPC>> SPOTS_LIST = new HashMap<>();
+	private static HashMap<WorldPoint, HashMap<FishingSpot, Integer>> SPOTS_COUNT = new HashMap<>();
 
 	@Setter(AccessLevel.PACKAGE)
 	private boolean hidden;
@@ -86,6 +87,7 @@ class FishingSpotOverlay extends Overlay
 		}
 
 		SPOTS_LIST.clear();
+		SPOTS_COUNT.clear();
 
 		for (NPC npc : plugin.getFishingSpots())
 		{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingSpotOverlay.java
```diff
@@ -57,6 +61,9 @@
 	private final Client client;
 	private final ItemManager itemManager;
 
+	private static HashMap<WorldPoint, HashMap<FishingSpot, NPC>> SPOTS_LIST = new HashMap<>();
```
deathbeam: Just make new type and use 1 map

nrodriguesdev: I've gone ahead and created a new inner class to take care of that. Each instance of the class FishingSpotInfo will store 3 ArrayLists, containing a list of FishingSpots, NPCs, and counts. The getters were implemented using indexes to maintain the same lookup speed as before.

I've tested these new changes and they're functioning as expected.

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingSpotOverlay.java
```diff
@@ -123,37 +158,139 @@ public Dimension render(Graphics2D graphics)
...
 					}
 				}
 			}
+		}
+
+		return null;
+	}
 
-			if (config.showSpotNames())
+	private class FishingSpotInfo
```
Digbigpig: Move this to it's own file

nrodriguesdev: I can certainly do that. In regards to design, I assume you recommend moving the class to its own file because it contains more than basic getters/setters?

Digbigpig: If you can, get on the RuneLite discord and talk with the guys in the #development chat channel. It sounds like they would like different/ smaller solution to this, and I wouldn't want you to waste your time.

As far as the separate file, I think it is just convention with this project to keep everything separate. The purpose for that might be to enforce modularity and make things obvious for other developers, but don't quote me on it.

nrodriguesdev:
Thanks for the feedback everyone. I've followed the advice of others on Discord. 

I've removed the new class and replaced it with a Multiset. This has reduced the number of lines of code and made it easier to read. 

I added this nasty bit to convert the Multiset(NPC) to Multiset(FishingSpot)
```
			Multiset<FishingSpot> spotList = HashMultiset.create();
			for (NPC npc : npcArray)
			{
				FishingSpot spot = FishingSpot.getSPOTS().get(npc.getId());
				spotList.add(spot);
			}
```
If there's a better way to do this, then please let me know. Thanks for checking this PR out and having the patience to deal with me lol.

xDemoN:
@Adam- you superseded this with 9fc04bf

PR #7899, opened 2019-02-17
Hop-To left click as rank in Clan Chat

### LINKED ISSUE #7868: Allow users to move the kick option from the clan chat
Currently, as soon as your get to a rank that you can kick people from a CC the kick option replaces the hop-to option, this means that lots of new admins in chats end up kicking people they want to hop to, would it be possible to allow users to move this to the bottom option still allowing people to kick from the clan chat, but also removes the risk of kicking people from the cc.

Mabel5:
Replaces Kick User with Hop-To closes #7868


Hydrox6:
Current behaviour is to have the menu forced open on when kick is the top option, so this isn't needed anymore

PR #7900, opened 2019-02-17
Add option to hide traps placed by other players

Lucwousin:
Add an option to make "Walk here" the standard left-click option for hunter traps placed by other players. This uses the hunter plugin to add locations to a set of worldpoints, and every trap not on a 'whitelisted' tile will have their options moved swapped. The positions are reset on client restart.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterPlugin.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.hunter;
 import com.google.inject.Provides;
 import java.time.Instant;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


Adam-:
I do not think our method of determining which traps are yours is reliable enough to do this.

Lucwousin:
Attempting to put down a hunter trap sends out a chat message. You could add all traps which spawn within 4 ticks after getting that message, and remove the distance checks. Do you think that would be reliable enough? It could of course see other people's traps as your own, but I personally don't think that's a huge issue as you would have to be putting down traps yourself for that to happen, in which case having someone else's traps left clickable wouldn't be a problem.

Thanks github for turning tabs into spaces by the way, last time I'm using the browser editor.

Adam-:
The chat message seems okay. Maybe we could do a "hunter session" like we do with woodcutting/fishing and then incorporate this into that.

Nightfirecat:
You'd need both the distance check and the time check in order to be more accurate, wouldn't you? Removing the distance check would make traps which are spawned far away from your character to be "owned" by you.

Lucwousin:
It would be more accurate with both checks, but at the same time have a bigger risk of not adding one of your own traps to the list. Not adding your own traps is worse than adding someone else's traps, in my opinion; especially when the traps only get added if they spawn in a (at most) 3 tick window after putting down your own trap.

That said, readding those checks back in is easily done, of course.

Lucwousin:
Trap ownership should be 99,9% accurate now for box traps, bird snares and magic boxes (the 3 kinds of trap worth hiding).

I've also made the recognition changes to #7910 but I'm not sure if that feature is useful in any way (yet).

deathbeam:
What is the point of adding the option to menu entry swapper when its using everything from hunter plugin, that makes no sense

Lucwousin:
Because it's not something which has any impact on you training hunter, but rather on when someone else is placing traps. It just doesn't make sense to go in the hunter settings in my opinion. The only reason I made it use the hunterplugin was the fact the hunterplugin already was collecting which traps are yours, and I didn't want to have that run 2 times.

deathbeam:
If it already depends on hunter plugin, just put it there. Your code atm do not makes sense because you are putting it to menu entry swapper just because you decided to, but thanks to that it is mess.

Lucwousin:
I'll change it to all be in the hunter plugin then, but I don't think it should be a hunter plugin setting.

deathbeam:
Jagex solved this in latest OSRS update so I dont think this is necessary anymore. or at least the hiding traps part.

PR #7902, opened 2019-02-17
Slayer Plugin: Add slaughter/expeditious bracelet charge infobox

### LINKED ISSUE #2407: Add Slaughter/Exped Brace to Indicators
To be able to see charges left on either while slaying just be handy for when afking.

ShaunDreclin:
Resolves #2407

screenshot

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -46,7 +46,10 @@ import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
+import net.runelite.api.EquipmentInventorySlot;
 import net.runelite.api.GameState;
+import net.runelite.api.InventoryID;
+import net.runelite.api.Item;
 import net.runelite.api.ItemID;
 import net.runelite.api.MessageNode;
 import net.runelite.api.NPC;
@@ -58,6 +61,7 @@ import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.ExperienceChanged;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
+import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.NpcDespawned;
 import net.runelite.api.events.NpcSpawned;
 import net.runelite.api.vars.SlayerUnlock;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -723,6 +793,35 @@ private void removeCounter()
...
+		{
+			return;
+		}
+
+		int braceletID = equippedSlaughter ? ItemID.BRACELET_OF_SLAUGHTER : ItemID.EXPEDITIOUS_BRACELET;
+		int braceletCount = equippedSlaughter ? slaughterChargeCount : expeditiousChargeCount;
+		String braceletTooltip = equippedSlaughter ? "Bracelet of slaughter" : "Expeditious bracelet";
+
+		braceletCounter = new Counter(itemManager.getImage(braceletID), this, String.valueOf(braceletCount));
+		braceletCounter.setPriority(InfoBoxPriority.MED);
```
deathbeam: This do not needs priority? (idk since when we have this anyway)

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -555,6 +594,37 @@ private void onConfigChanged(ConfigChanged event)
...
+		int gloveSlotID = EquipmentInventorySlot.GLOVES.getSlotIdx();
+
+		if (equipment.length <= gloveSlotID)
+		{
+			return;
+		}
+
+		equippedSlaughter = equipment[gloveSlotID].getId() == ItemID.BRACELET_OF_SLAUGHTER;
+		equippedExpeditious = equipment[gloveSlotID].getId() == ItemID.EXPEDITIOUS_BRACELET;
+		removeBraceletCounter();
```
deathbeam: just update the infobox instead of removing and readding it. it makes no sense to do it here

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
ShaunDreclin: The priority is set so the task counter and bracelet infoboxes always show up in the same order. Without it they'll be ordered based on whichever one was added more recently

I can remove it if you prefer but I think it's good to keep them consistent

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
ShaunDreclin: The logic looks pretty ugly if I do it that way https://pastebin.com/DhQu6QVy

Also, #7620 and `setTask()` both handle it by removing and readding the infobox

devLotto:
Should the infoboxes for the slayer bracelets really be in the slayer plugin? Or be put in the item charges plugin like #8090? What's best?

deathbeam:
Imo ideal scenario would be item charges plugin, but then it would need dependency on Slayer plugin.

Allemaalbaas:
Hi, I was wondering if you guys could make look back into this. Personally I totally agree with deathbeam that the ideal spot would be in the item charges plugin. but if it doesn't work without the slayer plugin then that wouldn't really make sense and the slayer plugin would be the best spot. I'd rather see it in a little misplaced spot than not. If there really isn't any way to fit this plugin into the existing plugins, I think plugin hub would be a really nice substitute.

PR #7904, opened 2019-02-17
Properly use Spring profiles in http-service tests and main implementation

deathbeam:


deathbeam:
Updated, now nothing except tomcat is in main application.yaml and this is simply migrating existing stuff to profiles. I will improve the main config in Docker PR instead of this PR.

deathbeam:
Added https://github.com/runelite/runelite/pull/7904/commits/e648460099bccf324abc4c4e32533cf4c0e7bebd that makes intellij integration and maven build warnings work great with our spring configuration (will properly throw warnings, provide completion etc)

deathbeam:
Removed the property part, as I know how to make it in better way, but I will move that part to docker stuff PR or separate PR (docker stuff PR is mostly pre-requisite to that, i think)

deathbeam:
And this PR should be very low intrusive so it makes more sense for that to be separate

PR #7910, opened 2019-02-18
(WIP) Add hunter session, refactor hunter plugin

Lucwousin:
WIP of a hunter session tracker, with amount of traps up, amount of catches and catches per hour.


Lucwousin:
The more I think about it the less useful a hunter session becomes. If you're catching chins you'll keep them in your inventory, so you know how many you've caught. If you're hunting other things you won't care how many you've caught as you probably drop them/get nothing. Xp/hour would be more useful and big surprise, that's already a feature.

Hunter being the energy draining, high attention, mentally draining skill it is probably also doesn't benefit from notifications for (or even just displaying) things like "amount traps active". You've got 5/6 traps in a not-changing pattern needing constant attention - there's no way you're gonna lose track of them.

If someone has a different opinion, feel free to enlighten me.

PR #7912, opened 2019-02-18
Item goals plugin

halworsen:
Lets you add item goals (i.e. 10k gp or 1 drakan's medallion) and track your progress towards those goals.

Pictures:

Known issues:
* The plugin uses the item variation map to keep the goal items consistent and to reduce clutter in the search results. The downside to this is that all variations of an item will count towards the goal. For example, if you set an air rune goal, NMZ air runes will count towards the goal since it's a variation of the air rune item.
* Some item variation maps will map items to a base item which is a members-only variation of an f2p items. The best example for this is coins. While itemID 995 is what's actually used for coins, the variation map maps all coins to itemID 667, which is a members only item. This results in F2Ps not being able to add coins as a goal. This is easily fixed by editing the item_variations.json file to map items to the proper itemID but it's so huge that I don't know if it's reasonably doable.
* Dropping items and then picking them back up will increase your progress. As far as I know it's not possible to check who dropped what item on the ground.


raiyni:
I don't think this is anymore useful than using notes.

yanikore:
Technical wise I think it would be better to recalculate counts on bank visits and inventory updates rather than while picking up items. Essentially so it always corrects itself.

> I don't think this is anymore useful than using notes.

I think this would add value if it were to also allow experience, quests and other ingame milestones and achievements. That way it could become an all-in-one tracker. Draggable orders could make it an auto updating TODO list.

raiyni:
This would be useful if it was able to work in a sensible

halworsen:
> Technical wise I think it would be better to recalculate counts on bank visits and inventory updates rather than while picking up items. Essentially so it always corrects itself.

Progress is always and only updated on inventory changes. The only exception is when you're banking, so that goals aren't progressed when you take stuff you already have out of the bank.

> I think this would add value if it were to also allow experience, quests and other ingame milestones and achievements.

This sounds like a good idea but also like a lot of work for a single PR. I think it'd be better to look into extending the plugin later.

> If the infoboxes were draggable in order this could make it a good TODO list.
I was considering implementing some sort of priority but wasn't sure how to do it, but this sounds great. I'll try to see what I can do.

halworsen:
After some thinking and tinkering, I think it might actually be easier than I thought to extend this to a more general milestones plugin. I'm going to open a new draft PR for it since I want to add some more kinds of milestones.

PR #7933, opened 2019-02-19
Add a config option to select Wiki left click action

yanikore:
image


runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,22 @@
+
```
Nightfirecat: License header

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,22 @@
...
+		keyName = "leftClickOption",
+		name = "Left click option",
+		description = "Configures the default left click option of the wiki button",
+		position = 1
+	)
+	default WikiLeftClickOption leftClickOption()
+	{
+		return WikiLeftClickOption.LOOKUP;
+	}
+}
```
Nightfirecat: Newline at EOF

runelite-client/src/main/java/net/runelite/client/plugins/wiki/config/WikiLeftClickOption.java
```diff
@@ -0,0 +1,20 @@
+package net.runelite.client.plugins.wiki.config;
```
Nightfirecat: License header.

Also, this does not need to be in a config directory, it can be in the same dir as the rest of the plugin files.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -129,6 +142,16 @@ public void shutDown()
 		});
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals("wiki"))
+		{
+			shutDown();
```
Nightfirecat: Don't invoke shutDown and startUp methods directly. Move the code called from there into other methods and call those.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/config/WikiLeftClickOption.java
yanikore: I copied the structure from `grounditems` and `wintertodt`. Should I change it?

Nightfirecat: I suppose it can stay where it is. It doesn't really make much of a difference, but being in a separate package requires import statements in other plugin files to access it.

PR #7944, opened 2019-02-20
Kourend Library Plugin: Add config option to hide duplicate books

Koekkruimels:
Added a config in the Kourend Library setting to hide book you already have in your inventory.
Since you can't have any duplicate books in your inventory.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import lombok.AccessLevel;
 import lombok.Setter;
 import net.runelite.api.Client;
@@ -57,15 +58,19 @@ class KourendLibraryOverlay extends Overlay
 	private final static int MAXIMUM_DISTANCE = 24;
 	private final Library library;
 	private final Client client;
+	private final KourendLibraryConfig config;
+	private final KourendLibraryPlugin plugin;
 
 	@Setter(AccessLevel.PACKAGE)
 	private boolean hidden;
 
 	@Inject
-	private KourendLibraryOverlay(Library library, Client client)
+	private KourendLibraryOverlay(Library library, Client client, KourendLibraryConfig config, KourendLibraryPlugin plugin)
 	{
 		this.library = library;
 		this.client = client;
+		this.config = config;
+		this.plugin = plugin;
 
 		setPosition(OverlayPosition.DYNAMIC);
 		setLayer(OverlayLayer.ABOVE_SCENE);
@@ -136,7 +141,7 @@ class KourendLibraryOverlay extends Overlay
 				Color color = bookIsKnown ? Color.ORANGE : Color.WHITE;
 
 				// Render the poly on the floor
-				if (!(bookIsKnown && book == null) && (library.getState() == SolvedState.NO_DATA || book != null || !possible.isEmpty()))
+				if (!(bookIsKnown && book == null) && (library.getState() == SolvedState.NO_DATA || book != null || !possible.isEmpty()) && !shouldHideOverlayIfDuplicateBook(book))
 				{
 					Polygon poly = getCanvasTilePoly(client, localBookcase);
 					if (poly != null)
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
```diff
@@ -229,4 +242,15 @@ public Dimension render(Graphics2D g)
 
 		return null;
 	}
+
+	private boolean shouldHideOverlayIfDuplicateBook(@Nullable Book book)
+	{
+		ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
```
Abextm: You don't need to load the item container here anymore

Koekkruimels:
> I would like for the plugin to keep a `EnumSet<Book>` that updates on `ItemContainerChanged`, so the overlay doesn't have to iterate the whole inventory multiple times a frame

I've addressed the problem., but something came to my attention, the merged [pull request](https://github.com/runelite/runelite/pull/7951) also contain this problem. When shall I pick this up and how?

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
```diff
@@ -148,20 +158,23 @@ public Dimension render(Graphics2D g)
...
-							screenBookcase.getY() - (height / 2) + (int) bounds.getHeight()
-						);
-						OverlayUtil.renderTextLocation(g, textLoc, book.getShortName(), color);
-						g.drawImage(
-							book.getIcon(),
-							screenBookcase.getX() - (book.getIcon().getWidth() / 2),
-							screenBookcase.getY() + (height / 2) - book.getIcon().getHeight(),
-							null
-						);
+						if (!shouldHideOverlayIfDuplicateBook(book))
```
Abextm: combine this with the enclosing if

Abextm:
I didn't worry about that because that was running at most once per frame. If you want to make it use the enumset, you can just add a commit to this pr, and smash everything else together.

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
Koekkruimels: Could you elaborate? We had this discussion last time, that INVENTORY couldn't be null, but the method doesn't state specifically that -> Nullable.

Koekkruimels:
> I didn't worry about that because that was running at most once per frame. If you want to make it use the enumset, you can just add a commit to this pr, and smash everything else together.

By smash everything together, you mean squash all commits into one?

Abextm:
yes, combine the commits, except for the one changing the merged pr

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
Abextm: You don't use the inventory in this anymore because you load it in the Plugin. The `INVENTORY` Inventory is a special case in that it cannot be null, all other Inventories are nullable, hence why that method is marked `@Nullable`

Koekkruimels: Ah, I looked at the wrong location, my bad.

Koekkruimels:
> I didn't worry about that because that was running at most once per frame. If you want to make it use the enumset, you can just add a commit to this pr, and smash everything else together.

Done

Koekkruimels:
> yes, combine the commits, except for the one changing the merged pr

Done

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
Koekkruimels: Done

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
Koekkruimels: Removed

runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
```diff
@@ -136,20 +141,20 @@ public Dimension render(Graphics2D g)
...
 				// Render the poly on the floor
-				if (!(bookIsKnown && book == null) && (library.getState() == SolvedState.NO_DATA || book != null || !possible.isEmpty()))
+				if (!(bookIsKnown && book == null) && (library.getState() == SolvedState.NO_DATA || book != null || !possible.isEmpty()) && !shouldHideOverlayIfDuplicateBook(book))
 				{
-					Polygon poly = getCanvasTilePoly(client, localBookcase);
-					if (poly != null)
-					{
-						OverlayUtil.renderPolygon(g, poly, color);
-					}
+						Polygon poly = getCanvasTilePoly(client, localBookcase);
```
Abextm: whitespace

PR #7949, opened 2019-02-20
Hide Dark Manuscript with Config option

Koekkruimels:
Added a config in the Kourend Library setting to hide dark manuscripts, since they become irrelevant once you have 100% favor in the Arceuus faction.


Beagon:
Dark Manuscripts can still be used for Magic XP so I disagree that they become irrelevant. Still always good to be able to config things off whenever you like.

Koekkruimels:
> Dark Manuscripts can still be used for Magic XP so I disagree that they become irrelevant. Still always good to be able to config things off whenever you like.

I didn't knew that, the [wiki](https://oldschoolrunescape.fandom.com/wiki/Dark_manuscript) doesn't states that. As well the plugin doesn't support the NPC Horphis to handle dark manuscripts.

But I will update the previous comment slightly, thanks for the information.

yanikore:
Maybe hiding them under 20% favour as well?

Koekkruimels:
> Maybe hiding them under 20% favour as well?

It will be a setting and users can decide for themselves if they want to hide these dark manuscripts.  

Removing it for them is unwanted, I can imagine, besides the whole dark manuscript aspect is quite unsupported in the plugin, if I'm correct.

yanikore:
> It will be a setting and users can decide for themselves if they want to hide these dark manuscripts.
> 
> Removing it for them is unwanted, I can imagine, besides the whole dark manuscript aspect is quite unsupported in the plugin, if I'm correct.

It's not accurate indeed. It's just that until 20% favour you can't even get any dark manuscripts.

PR #7951, opened 2019-02-20
Added color to the library customer based on player inventory

Koekkruimels:
It is sometimes hard to tell if you already have the book in your inventory that the library customer asks for, to simply change the color of the tile under the NPC from WHITE to GREEN it would solve this problem.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
@@ -220,7 +220,9 @@ class KourendLibraryOverlay extends Overlay
 				{
 					Book b = library.getCustomerBook();
 					ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
-					boolean hasBookInInventory = itemContainer != null && b != null && Arrays.stream(itemContainer.getItems()).anyMatch(item -> item.getId() == b.getItem());
+					boolean hasBookInInventory = itemContainer != null
+						&& b != null
+						&& Arrays.stream(itemContainer.getItems()).anyMatch(item -> item.getId() == b.getItem());
 					LocalPoint local = n.getLocalLocation();
 					Polygon poly = getCanvasTilePoly(client, local);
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryOverlay.java
```diff
@@ -216,9 +219,11 @@ public Dimension render(Graphics2D g)
 				.forEach(n ->
 				{
 					Book b = library.getCustomerBook();
+					ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
+					boolean hasBookInInventory = itemContainer != null && b != null && Arrays.stream(itemContainer.getItems()).anyMatch(item -> item.getId() == b.getItem());
```
Abextm: neither the book nor the inventory can ever be null

Koekkruimels: Client.java, line 1003
```
	@Nullable
	ItemContainer getItemContainer(InventoryID inventory);
```

ItemContainer can be null from the context given.

Book can't be null indeed in this case, since LibraryCustomer is not null, but I find it a bit unsafe assumption that could cause problems in the future.

Abextm: Yes, every other ItemContainer can be null, but not the inventory.
If you want to keep the null checks, atleast split this onto multiple lines

Koekkruimels: Split them up into multiple lines, like the following?

```
boolean hasBookInInventory = itemContainer != null
    && b != null
    && Arrays.stream(itemContainer.getItems()).anyMatch(item -> item.getId() == b.getItem());
```

Abextm: yes

Koekkruimels: Requested changes are done.

PR #7956, opened 2019-02-21
Timers: Make antipoison/venom timers work off the poison varplayer

### LINKED ISSUE #1471: Poison resistance overlay trigger
The poison resistance overlay does not trigger if you click on the hp orb after being poisoned to cure the poison.

### LINKED ISSUE #10724: drinking antidote++ in bank doesnt trigger timer
Drinking an antidote++ from the bank inventory does not create the timer infobox, drinking outside the bank works correctly.

Oddly enough i have tried to reproduce this with other antipoison style potions and all the ones ive tried have correctly created the timer

inspecting the bank inventory widget containing the antipoison potion displays the same itemID as outside the bank aswell

Lucwousin:
This commit allows the poison timer to be much more accurate as it
receives periodic updates from the server, and covers edge cases where
poison may be delayed or cured from sources besides potion clicks.
(banked potion drinks, HP orb "Cure" menu option, etc.)

This commit removes the `initialDelay` field from GameTimer as the field
was used only to chain antivenom timers into their follow-up antipoison
timers.

Fixes #1471
Fixes #10724

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
@@ -2,6 +2,7 @@
  * Copyright (c) 2017, Seth <Sethtroll3@gmail.com>
  * Copyright (c) 2017, Adam <Adam@sigterm.info>
  * Copyright (c) 2018, Jordan Atwood <jordan.atwood423@gmail.com>
+ * Copyright (c) 2019, Lucas <https://github.com/Lucwousin>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


deathbeam:
So what is the point of this change? You could've just improved Timers plugin then, instead of adding duplicate functionality

Lucwousin:
I could've (and probably should've) added it to the timers plugin, but the poison plugin is nothing more than a timer plugin right now as well. The fact there was a poison plugin which already did timers made me add it onto that. The fact it's duplicate came from me forgetting to remove the old antipoison timers.

Other question would be why is the poison plugin a seperate plugin when it's basically a timer plugin?

deathbeam:
Timers contains all potion timers, so this should be there

jbaker5011:
What about just deleting the poison plugin (really why does that even exist?) and then moving this all to timers

Lucwousin:
Updated PR title to actually explain what it does, and added the timers to the timers plugin again.

Adam-:
Why does this show ~18 seconds for antivenom when the wiki says that antivenom lasts 1 minute?

Lucwousin:
Wiki is wrong, or antidote++ also protects against venom. Drinking antivenom sets the poison var to 41, and antidote++ to 40.

The randomness comes from the fact the poison cycle stays the same, think of it like stat boosts: if you boost your stats with only 5 seconds left on the counter they'll tick down after those 5 seconds. Antipoison works the same way.

Adam-:
I don't understand your reply. If I drink an anti venom it will show anti-venom timer for maybe a short time (0 seconds, or 3 seconds, or more) and then after that it will change to anti poison for ~11 minutes or so.

Lucwousin:
If you drink it while poisoned/venomed/with antipoison active there'll be up to 30 ticks less protection than if you drink it without having anything active. This also applies to normal antipoison and poison/venom itself. Sometimes if you get venomed it'll hit almost instantly, that's the same thing happening here.

After the (up to) 18 seconds of antivenom it should change into 12 minutes of antipoison (poison value of -40, like antidote++). I don't really think there's a way to test venom protection other than to ask a jmod for confirmation.

Lucwousin:
[This](https://giant.gfycat.com/MaleIllfatedKingfisher.webm) maybe clarifies what I mean. Quick change to display ticks instead of seconds. As you can see drinking antipoison doesn't change when the poison value changes.

Alexsuperfly:
this would likely solve #1471

Nightfirecat:
I've rebased this PR and updated the original post to reflect the (multiple) issues it will close if merged.

Adam-:
This still needs rebase. It may be better if you just open your own PR and take it since the original author is no longer with us.

PR #7959, opened 2019-02-21
Add Temple Trekking plugin with bog overlay

### LINKED ISSUE #7957: Bog path highlighter for Temple Trekking
The bog in Temple Trekking has a path, where hard tiles have ID 13838 and soft tiles have ID 13839. 

Add a feature in which 13838 tiles are highlighted (for Swampletics reasons)

ChaoticConundrum:
I did this a long time ago, it's super simple. No one wanted it at the time (because it's a cheating/botting grey area), and I haven't really used it after testing. But Settled let the cat out of the bag, so I may as well post it so no one rewrites it.

bog

Closes #7957


Jehu920:
If this is turned down due to being a cheating grey area, could it be changed to only highlight tiles once the player steps/clicks on them?

deathbeam:
See reply here: https://github.com/runelite/runelite/issues/7957

PR #7962, opened 2019-02-21
Barrows plugin: add configuration to hide or show widget.

Malfuryent:
The current barrows plugin forces the player to have 2 GUI elements to enable all functionality of the plugin. 

This PR allows the player to hide the widget to declutter the screen without needing to disable the whole plugin.


Grabbin:
@Malfuryent would it instead be possible to remove the osrs gui, rather than the runelite one?

PR #7965, opened 2019-02-21
Do not try to draw infobox caption if text is null or empty

### LINKED ISSUE #7964: Prayer boost indicator breaks other indicators
Just updated to latest client release, and the prayer boost indicator just overwrites the slayer infobox and breaks the other indicators so they can't move and sometimes are misplaced.

deathbeam:
Closes #7964


runelite-client/src/main/java/net/runelite/client/ui/overlay/components/InfoBoxComponent.java
```diff
@@ -31,6 +31,7 @@
 import java.awt.Image;
 import java.awt.Point;
 import java.awt.Rectangle;
+import joptsimple.internal.Strings;
```
Adam-: This seems wrong.

PR #7967, opened 2019-02-21
Poison plugin: Add check whether or not the poison cycle should change

Lucwousin:


Adam-:
I don't know what this means.

Lucwousin:
Every 30 ticks after a player gets poisoned the poison var value decrements and the player gets damaged. This plugin was creating a timer for 18 seconds every time the poison var changes but there's cases where the poison var changes but the cycle of when the poison hits doesn't. This fixes the plugin to check whether or not the cycle should change.

Opening a interface stalls the poison damage. This fixes this plugin to still show the infoboxes of poison about to hit, and stops decrementing the time left until the poison wears off when a expected cycle doesn't happen.

Adam-:
This sounds like a bug, then, if the plugin is giving incorrect times.

Trevor159:
I think this can be closed now that we use varbs for the poison timers.

PR #7969, opened 2019-02-21
Move Binding necklace tracking from Runecrafting to Item Charge plugin

TheStonedTurtle:
Moves the Bind necklace charge tracking into the Item Charge plugin since it seems like a better fit. Also adds a new config option for a notification when the necklace is destroyed.

**Adds support for manually destroying jewelry. Includes dodgy necklace as well.**
The runecrafting plugin had a `destroy` handler for the binding necklace that didn't work with the key binds, `1` and `y`, so i implemented a fix using a script callback

### End-user Changes
* Adds binding necklace infobox
* Adds notification for broken binding necklace
* Binding necklace now tracks between client sessions
* Destroying dodgy necklace now updates the counter


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -275,6 +308,23 @@ private void updateBindingNecklaceCharges(final int value)
...
+		Widget widgetDestroyItemName = client.getWidget(WidgetInfo.DESTROY_ITEM_NAME);
+		if (widgetDestroyItemName == null)
+		{
+			return;
+		}
+
+		switch (widgetDestroyItemName.getText())
+		{
+			case "Binding necklace":
+				updateBindingNecklaceCharges(MAX_BINDING_CHARGES);
```
deathbeam: break;

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -275,6 +308,23 @@ private void updateBindingNecklaceCharges(final int value)
...
+			case "Dodgy necklace":
+				updateDodgyNecklaceCharges(MAX_DODGY_CHARGES);
```
deathbeam: break;

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -275,6 +308,23 @@ private void updateBindingNecklaceCharges(final int value)
 		}
 	}
 
+	private void checkDestroyWidget()
+	{
+		Widget widgetDestroyItemName = client.getWidget(WidgetInfo.DESTROY_ITEM_NAME);
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -104,13 +123,15 @@ ItemChargeConfig getConfig(ConfigManager configManager)
...
 		overlayManager.add(overlay);
+		keyManager.registerKeyListener(this);
 	}
 
 	@Override
 	protected void shutDown() throws Exception
 	{
 		overlayManager.remove(overlay);
 		infoBoxManager.removeIf(ItemChargeInfobox.class::isInstance);
+		keyManager.unregisterKeyListener(this);
```
deathbeam: also reset lastKeyTick here

TheStonedTurtle:
The requested changed have been made.

Adam-:
I don't see anything wrong with this staying in the rc plugin since it is specific to runecraft.

TheStonedTurtle:
I moved it here since it doesn't have an infobox or notification inside the RC plugin. It's also using queries inside its overlay so it seemed like moving it here would be cleaner than "fixing" those issues inside the RC plugin.

Adam-:
The item charges plugin also does queries so I don't see why that is any better? I suppose for the infobox stuff it may be worth moving though.

deathbeam:
Its better because in order to support infoboxes it is like 3 lines of code instead of like 100 because we would need to copy all infobox code to RC plugin.

Also imo it always made more sense to just put that counter to item charges instead of RC plugin so all logic would be grouped together because its like 99% same.

runeliteop:
This change would be absolutely huge for runecrafters, would be a great addition!!! Hopefully this gets added.

TheStonedTurtle:
The destroy functionality has been updated to use a script callback instead of a key listener.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -64,8 +67,14 @@
 	private static final Pattern DODGY_BREAK_PATTERN = Pattern.compile(
 		"Your dodgy necklace protects you\\..*It then crumbles to dust\\.");
 	private static final String RING_OF_RECOIL_BREAK_MESSAGE = "<col=7f007f>Your Ring of Recoil has shattered.</col>";
+	private static Pattern BINDING_CHECK_PATTERN = Pattern.compile(
+		"You have ([0-9]+|one) charges? left before your Binding necklace disintegrates.");
```
Adam-: `\\.`

Adam-:
Needs rebase, and probably doublecheck the script hash is still correct.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
TheStonedTurtle: Also made this variable final.

TheStonedTurtle:
The PR has been rebased and the script hash is still correct

PR #7972, opened 2019-02-21
Made Alchemy Protect

MarcelBostelaar:
Alchemy Protect is a plugin that protects selected items against accidentally being alched.
It uses the same CSV based item list with default wildcard functionality as the Ground Item plugin.

Use case:
You accidentally equip the rune dagger you wanted to alch and now you alched your dragon claws. Not anymore with this plugin (if you add them)!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/alchemyprotect/AlchemyProtectPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/alchemyprotect/AlchemyProtectPlugin.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2019, Marcel Bostelaar <https://github.com/MarcelBostelaar>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *	list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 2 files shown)


Alexsuperfly:
pretty much already exists, just right click the spell

MarcelBostelaar:
> 
> 
> pretty much already exists, just right click the spell

The in game feature only has untradable protection and price-based protection. It does not allow you to specify specific items. With this plugin you can protect certain items without losing the ability to alch more other items with higher prices (Very usefull to ironmen).

renewablezip01:
seems like a good idea :D

runelite-client/src/main/java/net/runelite/client/plugins/alchemyprotect/AlchemyProtectPlugin.java
```diff
@@ -0,0 +1,171 @@
...
+	{
+		reset();
+	}
+
+	private void reset()
+	{
+		protectedItems = Text.fromCSV(config.protecteditems());
+	}
+
+	private Boolean ItemIsMatch(String itemname)
```
GSchuringa: Method names should always start with lower case letters.

Nightfirecat: This method name is not descriptive or clear. `itemIsFiltered` would be a substantially better name.

PinkyOSRS:
was this plugin removed or renamed? because I can't find it in the hub

Nightfirecat:
Looks like this would do it: https://runelite.net/plugin-hub/show/alch-blocker

PR #7973, opened 2019-02-21
FarmingTracker: send notification when a crop becomes diseased

### LINKED ISSUE #7970: Send a notification when a farming patch is updated to a diseased state
It would be helpful to have a notification sent when coming within range of a farming patch and having its state change to a diseased one. This would be helpful for players to know to address the diseased patch before it dies during its next growth stage.

RyBo:
Fixes  #7970 

Tried to keep this as close to the way the bird house tracker sends notifications.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
@@ -50,8 +50,6 @@ public class FarmingTracker
 	private final ConfigManager configManager;
 	private final TimeTrackingConfig config;
 	private final FarmingWorld farmingWorld;
-	private static final String POUCH_DECAYED_NOTIFICATION_MESSAGE = "Your rune pouch has decayed.";
-	private static final String POUCH_DECAYED_MESSAGE = "Your pouch has decayed through use.";
 	private final Notifier notifier;
 
 	private final Map<Tab, SummaryState> summaries = new EnumMap<>(Tab.class);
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -187,10 +190,16 @@ public PatchPrediction predictPatch(FarmingPatch patch)
 
 		PatchState state = patch.getImplementation().forVarbitValue(value);
 
+
```
Nightfirecat: Remove this extra newline.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -187,10 +190,16 @@ public PatchPrediction predictPatch(FarmingPatch patch)
...
 		int stage = state.getStage();
 		int stages = state.getStages();
 		int tickrate = state.getTickRate() * 60;
 
+		if (config.diseasedCropNotification() && state.getCropState() == CropState.DISEASED)
+		{
+			notifier.notify("You have a diseased crop, cure it quick!");
```
Nightfirecat: This should probably have more information, like crop type and location.

RyBo: Added location and crop type.

Abextm:
I don't think this is a good idea because it will only know about things becoming diseased when you enter the region with the crop in it, which seems moderately useless and confusing to those who do not understand how the farming system works.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -191,6 +194,14 @@ public PatchPrediction predictPatch(FarmingPatch patch)
 		int stages = state.getStages();
 		int tickrate = state.getTickRate() * 60;
 
+		if (config.diseasedCropNotification() && state.getCropState() == CropState.DISEASED)
```
Abextm: This method does not run when a crop changes state. It runs when the ui updates, so this will notify you every 30 seconds when you have a diseased plant an the ui is open

RyBo: Thanks for pointing that out, I agree with your comment that this feature (even if corrected) would be confusing.

PR #7980, opened 2019-02-22
Catch JVM incompatible library JNA error when initializing Discord

deathbeam:


deathbeam:
```
There is an incompatible JNA native library installed on this system
Expected: 5.2.0
Found:    4.0.1
To resolve this issue you may do one of the following:
 - remove or uninstall the offending library
 - set the system property jna.nosys=true
 - set jna.boot.library.path to include the path to the version of the
```

PR #7981, opened 2019-02-22
raids plugin: update scouter on chunk load and let players relog without breaking

Trevor159:
Scouter: https://streamable.com/sjvq3

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -304,6 +304,9 @@ public enum Varbits
 	PERSONAL_POINTS(5422),
 	RAID_PARTY_SIZE(5424),
 
+	// 0 = raid not started, >0 = raid started
+	RAID_STATE(5425),
+
 	/**
 	 * Making Friends with My Arm fire pits
 	 *
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
@@ -62,7 +62,10 @@ import static net.runelite.api.SpriteID.TAB_QUESTS_BROWN_RAIDING_PARTY;
 import net.runelite.api.Tile;
 import net.runelite.api.VarPlayer;
 import net.runelite.api.Varbits;
+import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
+import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.WidgetID;
@@ -108,6 +111,9 @@ import net.runelite.http.api.ws.messages.party.PartyChatMessage;
 public class RaidsPlugin extends Plugin
 {
 	private static final int LOBBY_PLANE = 3;
+	private static final int SECOND_FLOOR_PLANE = 2;
+	private static final int ROOMS_PER_PLANE = 8;
+	private static final int AMOUNT_OF_ROOMS_PER_X_AXIS_PER_PLANE = 4;
 	private static final String RAID_START_MESSAGE = "The raid has begun!";
 	private static final String LEVEL_COMPLETE_MESSAGE = "level complete!";
 	private static final String RAID_COMPLETE_MESSAGE = "Congratulations - your raid is complete!";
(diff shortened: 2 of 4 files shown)


MagicfTail:
After completing a raid. When making a new party, the raid overlay from the previous raid will show outside of the raid.

15987632:
Rebased and fixed the known bug.

15987632:
Updated PR to also remove the raid timer infobox and put the tooltip information onto the points widget.

15987632:
nvm

m-yk:
"Show scout overlay outside lobby" is broken because it seems to null the scout when you leave the instance even if someone's holding / still in party

Trevor159:
fixed

deathbeam:
@Kamielvf is this fine? You should probably look at this

deathbeam:
Any update on this?

Trevor159:
> Any update on this?

I just rebased and fixed the part I thought was bad. It should be good for review.

deathbeam:
Nice. @Kamielvf can you check it?

runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -296,6 +296,7 @@
 	TOTAL_POINTS(5431),
 	PERSONAL_POINTS(5422),
 	RAID_PARTY_SIZE(5424),
+	RAID_STATE(5425),
```
Nightfirecat: Doc for this var would be helpful, even if it's something as simple as `0 = raid not started, >0 = raid started`

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -807,4 +907,50 @@ private void screenshotScoutOverlay()
...
+				{
+					break;
+				}
+			}
+		}
+
+		return x + (y * 4);
+	}
+
+	private void reset()
```
Nightfirecat: This should also set `timer = null`.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -807,4 +907,50 @@ private void screenshotScoutOverlay()
...
+			{
+				int sceneX = gridBase.getX() - 1 - RaidRoom.ROOM_MAX_SIZE * x;
+				if (sceneX < 0 || tiles[sceneX][gridBase.getY()] == null)
+				{
+					break;
+				}
+			}
+		}
+
+		return x + (y * 4);
```
Nightfirecat: Since this is a bitpacked int, we should use bitwise operations.
```suggestion
		return y << 2 | x;
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -807,4 +907,50 @@ private void screenshotScoutOverlay()
 		imageCapture.takeScreenshot(overlayImage, "CoX_scout-", false, config.uploadScreenshot());
 		graphic.dispose();
 	}
+
+	private int findLobbyIndex(Point gridBase)
```
Nightfirecat: It would be lovely to have some documentation on this, or at least constants relating to the otherwise magic numbers in this method. (why is `y` only 0 or 1? why is `x`'s max value 3?)

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -549,75 +626,98 @@ private Point findLobbyBase()
...
+			// The x position of the current room relative to other rooms
+			// The west most room will have a value of 0 and the east most will have a value of 3
+			int x = i % 4;
+
+			// The y position of the current room relative to other rooms
+			// The north rooms will have a y value of 0 and the south will have a value of 1
+			int y = i % 8 > 3 ? 1 : 0;
+
+			// The plane of the current room
+			int plane = i > 7 ? 2 : 3;
```
Nightfirecat: The comments are helpful, but these are still pretty magical numbers without any meaning directly associated with them.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
Trevor159: this is covered by `updateInfoBoxState()`

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
Trevor159: Let me know if the new comments make it easier to understand. I understand it very well so I am not a good judge.

PR #7983, opened 2019-02-22
Fix the string encoding for building scripts with the maven plugin on windows

l2-:
java Strings are utf-16 and the current implementation implicitly converts from utf-16 to iso88591.
This seems to go alright on unix but results in weird characters such as the Â behind the name in the chatbox, privatechat and clanchat.
https://stackoverflow.com/a/655948
decoding to utf-8 first and then encoding it to iso88591 fixes this problem for me and i no longer see the Â.

### DIFF
--- a/cache/src/main/java/net/runelite/cache/io/OutputStream.java
+++ b/cache/src/main/java/net/runelite/cache/io/OutputStream.java
@@ -26,7 +26,6 @@ package net.runelite.cache.io;
 
 import com.google.common.base.Preconditions;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
@@ -184,23 +183,13 @@ public final class OutputStream extends java.io.OutputStream
 	public void writeString(String str)
 	{
 		Charset utf8charset = Charset.forName("UTF-8");
-		Charset iso88591charset = Charset.forName("ISO-8859-1");
+		Charset cp1252charset = Charset.forName("Cp1252");
 
-		byte[] b;
-		try
-		{
-			b = str.getBytes("ISO-8859-1");
-		}
-		catch (UnsupportedEncodingException ex)
-		{
-			throw new RuntimeException(ex);
-		}
-
-		ByteBuffer inputBuffer = ByteBuffer.wrap(b);
+		ByteBuffer inputBuffer = ByteBuffer.wrap(str.getBytes());
 
 		CharBuffer data = utf8charset.decode(inputBuffer);
 
-		ByteBuffer outputBuffer = iso88591charset.encode(data);
+		ByteBuffer outputBuffer = cp1252charset.encode(data);
 		byte[] outputData = outputBuffer.array();
 
 		writeBytes(outputData);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I encountered the same issue on Windows, and this patch resolved the issue for me.

According to Abex, the proper charset is `cp1252`

Nightfirecat:
This still fixes the issue after latest commit and doesn't fail builds anymore! :tada:

Adam-:
Why isnt this just `getBytes("cp1252")`?

l2-:
```
public void writeString(String str)
{
	byte[] b;
	try
	{
		b = str.getBytes("cp1252");
	}
	catch (UnsupportedEncodingException ex)
	{
		throw new RuntimeException(ex);
	}
	writeBytes(b);
	writeByte(0);
}
```
doesnt work. just tried. 
i think for some reason you need to decode with utf-8 yourself because the system default (in my case utf-16?) results in the Â

chivesrs:
https://github.com/google/error-prone/blob/master/docs/bugpattern/DefaultCharset.md

cache/src/main/java/net/runelite/cache/io/OutputStream.java
```diff
@@ -181,16 +182,17 @@ public void writeLengthFromMark(int var1)
...
-		{
-			throw new RuntimeException(ex);
-		}
-		writeBytes(b);
+		Charset utf8charset = Charset.forName("UTF-8");
+		Charset cp1252charset = Charset.forName("Cp1252");
+
+		ByteBuffer inputBuffer = ByteBuffer.wrap(str.getBytes());
+
+		CharBuffer data = utf8charset.decode(inputBuffer);
```
Abextm: Why are you converting from utf16 to utf 8 to utf 16 to cp1252? Just go from utf16 to cp1252.

l2-: Is that not what adam suggested and what I tried with the snippet 2 comments back or do you have another way in mind? the snippet above didn't have work for me.

Abextm: It is what Adam suggested; you should figure out why it does not work because what you have submitted looks very wrong and you don't seem to understand why it works.

Nightfirecat: The following patch, while following your recommendation of encoding bytes directly as utf-16 -> cp1252, doesn't seem to fix it for me.

```diff
diff --git a/cache/src/main/java/net/runelite/cache/io/OutputStream.java b/cache/src/main/java/net/runelite/cache/io/OutputStream.java
index 77853598b..c1591bbdb 100644
--- a/cache/src/main/java/net/runelite/cache/io/OutputStream.java
+++ b/cache/src/main/java/net/runelite/cache/io/OutputStream.java
@@ -27,7 +27,7 @@ package net.runelite.cache.io;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
+import java.nio.charset.Charset;
 
 public final class OutputStream extends java.io.OutputStream
 {
@@ -181,7 +181,7 @@ public final class OutputStream extends java.io.OutputStream
 
 	public void writeString(String str)
 	{
-		writeBytes(str.getBytes(StandardCharsets.ISO_8859_1));
+		writeBytes(str.getBytes(Charset.forName("Cp1252")));
 		writeByte(0);
 	}
```

PR #7991, opened 2019-02-24
Move Barrows chat value message to LootTracker Plugin and add support for Raids

farhan1666:
First time contributing to the client, this is a feature that I have wanted in the client for a while as each time I complete a raid I either price check my loot or look at my loot tracker as I know many others do, this enhancement will eliminate that step 😄 , this may be appropriate somewhere other than loottracker and originally I had it in RaidsPlugin but as I wanted it in ToB as well and there is no base ToB plugin loottracker was the plugin that made the most sense

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -45,7 +45,6 @@ import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.*;
-import net.runelite.api.ChatMessageType;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.ConfigChanged;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -44,14 +44,7 @@
...
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.ChatMessageType;
-import net.runelite.api.Client;
-import net.runelite.api.InventoryID;
-import net.runelite.api.ItemComposition;
-import net.runelite.api.ItemContainer;
-import net.runelite.api.NPC;
-import net.runelite.api.Player;
-import net.runelite.api.SpriteID;
+import net.runelite.api.*;
```
Alexsuperfly: revert this

farhan1666: My IntelliJ snuck that under my nose, fixing now

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -334,6 +343,29 @@ public void onWidgetLoaded(WidgetLoaded event)
 			return;
 		}
 
+		if (!(event.getGroupId() == WidgetID.CLUE_SCROLL_REWARD_GROUP_ID) && config.chestLootChat())
```
Adam-: Is this here because the game already shows you clue scroll price? This should be written as `event.getGroupid() != WidgetID.CLUE_SCROLL_REWARD_GROUP_ID`

TheStonedTurtle:
Separating all the price message logic into a new function and utilizing the `LootTrackerItem[] entries` would probably be better. This would allow you to implement it for the `PlayerLootReceived` and doesn't add new usages of itemManager

PR #7997, opened 2019-02-24
api: add RS Enum API

### LINKED ISSUE #6862: Add cache enum API
This would be useful for things like quest completion states and also working with the collection log.

Relevant script opcodes are 3400 and 3408

Adam-:
Closes #6862

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -1579,4 +1579,6 @@ public interface Client extends GameEngine
      * Returns client item composition cache
 	 */
 	NodeCache getItemCompositionCache();
+
+	EnumComposition getEnum(int id);
 }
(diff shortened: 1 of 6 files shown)


runelite-api/src/main/java/net/runelite/api/Enum.java
```diff
@@ -0,0 +1,32 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.api;
+
+public interface Enum
```
Joshua-F: I would maybe only suggest not using just `Enum` due to `java.lang.Enum`. I'd suggest `EnumType` or `EnumComposition` (following convention of other config classes)

PR #7998, opened 2019-02-24
Loot tracker: Add grouping by day.

mattec92:
Adds an additional filter to the loot tracker, filtering by day.
When enabled loots are split per day. Days are separated by a date label.

loot_tracker_grouped

Works together with other filters like separate drops:
loot_tracker_separate

And ignored items:
loot_tracker_hidden
If all drops are ignored for a day the header would also be removed.


deathbeam:
This should be on website and not in client. In client it do not even makes sense because client only loads last 1000 entries.

mattec92:
The use case is when I do the same PVM multiple days in a row without resetting. Then I want to know how much was from today vs yesterday. 

The intention was never to create some complete drop log that is searchable by date. Also, I believe most people are not logged to get their loot saved?

Adam-:
I agree this should be on the site since the UI can be done much better there. It will also let you export the raw data so you can organize it in whatever way you want to.

PR #7999, opened 2019-02-24
Split loading and uploading loot data into separate configs

MagicfTail:
Seeing your old loot in client shouldn't be a requirement to use the online loot tracker


deathbeam:
Without uploading data there is nothing to load so I dont think this makes sense?

deathbeam:
Oh, I see, it is to not see data in client

15987632:
This should probably also change the reset button to not delete your stored loot, only delete loot from your tracker.

MagicfTail:
I think that it, although needed, should be a separate pr

deathbeam:
I would maybe do

- Submit loot tracker data
- Synchronize panel with remote (what would disable both loading of the data and deleting kills on panel reset/delete)

MagicfTail:
Updated

PR #8002, opened 2019-02-24
Add ability to swap Wilderness Obelisk options

J3sven:
For those that have completed the Wilderness hard diaries you might want to be able to left-click "Set destination" or "Teleport to destination".


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -171,6 +171,16 @@ default HouseMode swapHomePortal()
 	{
 		return HouseMode.HOME;
 	}
+	
+	@ConfigItem(
+			keyName = "swapObelisk",
```
Adam-: Use one tab for indenting

raiyni:
Added in external plugin `Menu Swapper Extended`

PR #8006, opened 2019-02-25
Wiki Plugin - Added option for quest rewards and video guide 

zzznz27:
Added option for quest rewards and video guide 

Video guide links to youtube search for Slayermusiq1 guides


runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -258,6 +261,17 @@ private void onMenuOptionClicked(MenuOptionClicked ev)
 					{
 						LinkBrowser.browse(WIKI_BASE + "/w/" + URLEncoder.encode(skillRegex.group(1)) + "?" + UTM_PARAMS);
 					}
+				case MENUOP_REWARDS:
```
raiyni: We already link to the quest wiki

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -258,6 +261,17 @@ private void onMenuOptionClicked(MenuOptionClicked ev)
...
+					ev.consume();
+					String questNA = Text.removeTags(ev.getMenuTarget());
+					LinkBrowser.browse(WIKI_BASE + "/w/" + URLEncoder.encode(questNA.replace(' ', '_')) + quickguide + "?" + UTM_PARAMS + "#Rewards");
+					break;
+
+                case MENUOP_WELCOMETOMYGUIDE:
+                    ev.consume();
+                    String questSM = Text.removeTags(ev.getMenuTarget());
+                    LinkBrowser.browse("https://www.youtube.com/results?search_query=Slayermusiq1+" + URLEncoder.encode(questSM.replace(' ', '+')) );
```
raiyni: We probably don't want to be linking to arbitrary youtubers

Abextm:
This is not a feature we are interested in

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
ShaunDreclin: I wouldn't call slayermusiq arbitrary, if we linked to *any* video guides it would make sense to use his. Still not a fan of the idea though

deathbeam: Soup or bust

b3ntly:
A large number of users would benefit from this feature, what about this proposal:

* Feature is opt-out by default
* No youtuber is selected by default
* Feature settings allow you to pick between 1/2/3 most popular quest guide youtubers
* The code is written such that it is very easy to support new youtubers (there-by nullifying favoritism - any youtuber could make a small update or request to include themselves as an option)

PR #8017, opened 2019-02-26
Add account prompt to the OAuth URL

Hydrox6:
This will allow users to select what Google Account to use, since Chrome will just use the currently logged in account without a prompt.

The user will now see this upon logging in
vivaldi_2019-02-26_13-11-32


deathbeam:
Tested this locally, works perfectly

PR #7707, opened 2019-02-04
Clanchat: Show amount of members near you

SebastiaanVanspauwen:
Shows amount of clan members near you (default disabled). Can be useful for when trying to know how many clan members turned up to an event, so you don't have to manually count them.
image

PR #8018, opened 2019-02-26
Remove our raids points overlay and instead make vanilla one moveable

deathbeam:
There is no need to maintain our own infobox when vanilla one is already moveable. Only thing that our overlay adds is party size, what is easily viewable from vanilla interface already, and there is #7707 what will make this feature even more obsolete.


bjornenalfa:
Would it be possible to extended the vanilla infobox with partysize? #7707 is not an adequate replacement since it only counts visible players. Being able to see the partysize is important in order to tell if someone has disconnected or left the raid. Now we have to tediously switch tabs to check every couple of minutes. Also it is very helpful just to be able to quickly check if you have enough secondaries/pots without having to close the bank and switch from the inventory.

PR #8024, opened 2019-02-27
Null-check remote configuration in config manager

deathbeam:
This field can be null if the config service do not find absolutely
anything in config table for the user.


deathbeam:
To be fair, I managed to reproduce it only few times on stack running on local machine, and I am unsure why it returned null instead of empty list.

raiyni:
This happens to me on local API too

PR #8027, opened 2019-02-27
Change logged-in redir to redirect to HTTPS

deathbeam:


Nightfirecat:
Is there a reason we don't redirect all pages to https?

deathbeam:
We do. But this should still point directly to https version.

deathbeam:
@Adam- do you have https redirect handled in RL oauth app?

Adam-:
Seems so https://github.com/runelite/runelite-oauth/blob/master/src/main/java/net/runelite/oauth/OAuthServlet.java#L49?

deathbeam:
screenie

I meant on google apps

Issue #4847, opened 2018-08-11
If client fails to start, show the error in a popup in addition to the logs


Gamer1120:
I've seen multiple issues get opened about clients not starting up. Usually there is more information required to find the cause (logs etc.), which is usually not given by the OP. MSPaint mock-up:
![image](https://user-images.githubusercontent.com/9074448/43991426-e00a93f6-9d6c-11e8-8bb9-65ac80c3e090.png)

dekvall:
Launcher 2.0 has this.

Alexsuperfly:
I believe #9629 added this

PR #8035, opened 2019-02-28
Add error dialog before exiting client loader

### LINKED ISSUE #8034: Add an error message / dialog box when Runelite can't connect to the servers
I tried to launch Runelite just now, but it wouldn't open. It went through the entire loading bar, but then the game never launched. I went through the troubleshooting steps, but nothing seemed to help. There were no errors in my launcher.log file either.

I tried uninstalling it and reinstalling it, but that didn't fix the issue either.

Then I checked Reddit and discovered that the servers were down due to a bug. It would have been nice if Runelite had told me that at the very start instead of silently disappearing.

Would it be possible to display an error message when the client can't connect to the server?

TheStonedTurtle:
Adds a confirmation dialog window letting the user know the client failed to load before exiting. Currently the client will just silently fail (it has logging but that isn't very helpful for the average end user)

Fixes ~~#8034~~ #4847

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
+++ b/runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
@@ -37,6 +37,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
@@ -45,6 +46,7 @@ import java.util.jar.JarInputStream;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import javax.swing.JOptionPane;
 import lombok.extern.slf4j.Slf4j;
 import static net.runelite.client.rs.ClientUpdateCheckMode.*;
 import net.runelite.http.api.RuneLiteAPI;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
```diff
@@ -203,14 +205,25 @@ public Applet load()
...
 		{
+			String message = "Error loading the Old School RuneScape client";
 			if (e instanceof ClassNotFoundException)
 			{
 				log.error("Unable to load client - class not found. This means you"
 					+ " are not running RuneLite with Maven as the client patch"
 					+ " is not in your classpath.");
 			}
 
+			if (e instanceof SocketTimeoutException)
```
deathbeam: else-if

TheStonedTurtle:
The requested change has been implemented

runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
```diff
@@ -203,14 +205,24 @@ public Applet load()
...
+			String message = "Error loading the Old School RuneScape client";
 			if (e instanceof ClassNotFoundException)
 			{
 				log.error("Unable to load client - class not found. This means you"
 					+ " are not running RuneLite with Maven as the client patch"
 					+ " is not in your classpath.");
 			}
+			else if (e instanceof SocketTimeoutException)
+			{
+				log.error("Unable to load client - connection timed out. This is most"
```
Adam-: This is already being logged below so this makes no sense?

runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
```diff
@@ -203,14 +205,24 @@ public Applet load()
...
+					+ " likely caused by a Jagex server issue.");
+				message = "Error connecting to the Old School RuneScape servers.";
```
Adam-: This error message is not useful. Maybe "Timeout when connecting to Old School RuneScape servers". Also I think there are probably more errors than just socket timeout?

runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
TheStonedTurtle: I followed the existing format of a more in-depth log based on the error instance, you'll see that the `ClassNotFoundException` has a similar message right above this. Should I remove this or leave it in?

runelite-client/src/main/java/net/runelite/client/rs/ClientLoader.java
TheStonedTurtle: All other errors will default to `Error loading the Old School RuneScape client`. I'm not sure what error messages to use in the other cases. IOException is caused by multiple things but the others aren't

Line 172: `Patch.patch(file.getValue(), bytes, patchOs);` throws the `InvalidHeaderException` & `CompressorException`
Line 201: `Applet rs = (Applet) clientClass.newInstance()` throws the `InstantiationException` & `IllegalAccessException`

I could make generic error messages like `Failed to apply patch` and `Failed to initialize new client instance`?

PR #8051, opened 2019-03-01
Daily Task Indicator: Fix plugin not checking tasks when first enabled

ShaunDreclin:


runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksPlugin.java
```diff
@@ -86,10 +86,15 @@ DailyTasksConfig provideConfig(ConfigManager configManager)
...
+	public void startUp()
+	{
+		loggingIn = true;
+	}
+
+	@Override
+	public void shutDown()
 	{
 		lastReset = 0L;
-		loggingIn = false;
```
deathbeam: Why? Do not remove this

ShaunDreclin: It's redundant since it's set to true on startup now, you'd be setting it to false to just set it back to true when you turn it on again

PR #8062, opened 2019-03-01
Mass digger

Doctor13J:
digging time


deathbeam:
This will not get merged into main runelite repo, so no point in having this open

PR #8064, opened 2019-03-02
clan chat: add temporary join/leave messages

### LINKED ISSUE #726: Add temporary chat messages for when someone joins/leaves a CC
Similar to how konduit does it, a brief chat message that stays for a couple of seconds then deletes itself notifying when someone joins/leaves a CC

trimbe:
Depends on !51

Closes #726

### DIFF
--- a/runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
+++ b/runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2016-2017, Adam <Adam@sigterm.info>
+ * Copyright (c) 2018, trimbe <github.com/trimbe>
+ * Copyright (c) 2018, Adam <Adam@sigterm.info>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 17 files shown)


runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
```diff
@@ -0,0 +1,70 @@
...
+	@Inject
+	@Override
+	public void removeMessageNode(MessageNode node)
+	{
+		MessageNode[] lines = getLines();
+		int found = -1;
+
+		for (int i = 0; i < lines.length; i++)
+		{
+			if (lines[i] != null && lines[i].getId() == node.getId())
```
Adam-: This should just be a reference comparison?

runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
```diff
@@ -0,0 +1,70 @@
...
+	private static RSClient client;
+
+	@Inject
+	@Override
+	public void removeMessageNode(MessageNode node)
+	{
+		MessageNode[] lines = getLines();
+		int found = -1;
+
+		for (int i = 0; i < lines.length; i++)
```
Adam-: There is a separate `getLength` you should use instead

runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
```diff
@@ -0,0 +1,70 @@
...
+
+		// the client appears to stop building messages on the first null it encounters
+		// shift all entries, starting at the one we're removing, to the left by 1 place
+		if (found >= 0)
+		{
+			for (int i = found; i < lines.length - 1; i++)
+			{
+				lines[i] = lines[i + 1];
+			}
+			lines[lines.length - 1] = null;
```
Adam-: This also needs to update `length` since it is updating the number of lines?

runelite-mixins/src/main/java/net/runelite/mixins/RSChatLineBufferMixin.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		node.unlink();
```
Adam-: This should not unlink anything if the node isn't found

runelite-mixins/src/main/java/net/runelite/mixins/RSClanMemberManagerMixin.java
```diff
@@ -0,0 +1,64 @@
...
+import net.runelite.api.mixins.Mixin;
+import net.runelite.api.mixins.Shadow;
+import net.runelite.rs.api.RSClanMemberManager;
+import net.runelite.rs.api.RSClient;
+import net.runelite.rs.api.RSName;
+import net.runelite.rs.api.RSNameable;
+import net.runelite.rs.api.RSNameableContainer;
+
+@Mixin(RSClanMemberManager.class)
+public abstract class RSClanMemberManagerMixin extends RSNameableContainerMixin
```
Adam-: this should be implementing `RSClanMemberManager`

PR #8066, opened 2019-03-02
Clanchat: count players already in scene

SebastiaanVanspauwen:
Currently the clanchat counter change #7707 didn't take into account that players could already be in scene thus giving wrong numbers. 
This change makes it count players already in scene when joining a clanchat, or removes the counter when leaving a clanchat


deathbeam:
Is the clientThread.invokeLater intentional? e.g you intentionally want to delay it 1 tick?

SebastiaanVanspauwen:
Yes otherwise player.isClanMember() would return false for clan members

PR #8068, opened 2019-03-02
Clanchat: clear counter on login/connection lost instead of loading

SebastiaanVanspauwen:
Clear counter on logging_in state instead of loading. Sometimes when you run into new regions (revenant caves pking for example) you get the loading state which clears the counter.


deathbeam:
You should maybe do what NPC-related clearing does then, e.g connection lost etc etc

deathbeam:
I meant like this:
```
		if (gameState == GameState.LOGIN_SCREEN || gameState == GameState.HOPPING || gameState == GameState.CONNECTION_LOST)
```

Issue #6317, opened 2018-11-02
Agility PlugIn Not Accurate if level up during lap


tmk2003:
**Describe the bug**
The agility PlugIn "Laps till Level" does not update until after the lap is complete if you level up in the middle of your lap.

**To Reproduce**
1. Go to an agility course
2. Level up agility on an action that does NOT finish the lap
3. See that the Agility PlugIn does not update laps till level until lap is complete

**Expected behavior**
Once level up occurs, the agility plugin should update "laps till level" accurately

PR #8072, opened 2019-03-02
Changed agility plugin functionality to correctly account for xp on i…

tchar989:
**Notes**
- Changed agility plugin functionality to correctly account for xp on intermediate obstacles. 

  - This resolves a state where leveling up on intermediate obstacles does not recalculate laps left. 

- This fixes the bug outlined in issue [runelite#6317](https://github.com/runelite/runelite/issues/6317).
- Changed List.transform function into JDK native stream function for HotColdClue.java.

**Testing**
- Tested agility changes on my RS account by intentionally getting an agility level through intermediate obstacles. Confirmed that lap count is now updated.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -195,25 +195,20 @@ public class AgilityPlugin extends Plugin
 
 		// Get course
 		Courses course = Courses.getCourse(client.getLocalPlayer().getWorldLocation().getRegionID());
-		if (course == null
-			|| (course.getCourseEndWorldPoints().length == 0
-			? Math.abs(course.getLastObstacleXp() - skillGained) > 1
-			: Arrays.stream(course.getCourseEndWorldPoints()).noneMatch(wp -> wp.equals(client.getLocalPlayer().getWorldLocation()))))
+		if (course == null)
 		{
 			return;
 		}
 
-		if (session != null && session.getCourse() == course)
-		{
-			session.incrementLapCount(client);
-		}
-		else
+		if (session == null || session.getCourse() != course)
 		{
 			session = new AgilitySession(course);
 			// New course found, reset lap count and set new course
 			session.resetLapCount();
-			session.incrementLapCount(client);
+			session.calculateLapsTillLevel(client, skillGained);
 		}
+
+		session.calculateLapsTillLevel(client, skillGained);
 	}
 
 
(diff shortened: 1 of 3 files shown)


tchar989:
The cause of the bug was that lapsTillNextLevel in AgilitySession.java was being updated only when experience gained was equal to the lap completion bonus xp: https://tinyurl.com/y4n8k72h. If the amount of xp gained was less than the last obstacle XP, the method would return and lapsTillNextLevel in AgilitySession.java would never be recalculated.

The changes in AgilitySession.java were minor style tweaks. The IncrementLapCount method was actually calculated laps till next level and incrementing lap count, so I separated them out for more clarity.

tchar989:
Looks like 2 checkstyle errors were thrown due to curly bracket placement and an unused import. Working on making those tweaks.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -45,22 +46,34 @@
...
 
 		int remainingXp;
-		do
+		do 
 		{
 			remainingXp = nextLevel <= Experience.MAX_VIRT_LEVEL ? Experience.getXpForLevel(nextLevel) - currentExp : 0;
 			nextLevel++;
 		} while (remainingXp < 0);
 
 		lapsTillLevel = remainingXp > 0 ? (int) Math.ceil(remainingXp / course.getTotalXp()) : 0;
```
Adam-: This value becomes wrong if it is computed at each obstacle in the course

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/HotColdClue.java
```diff
@@ -86,7 +87,7 @@ public static HotColdClue forText(String text)
 	@Override
 	public WorldPoint[] getLocations()
 	{
-		return Lists.transform(digLocations, HotColdLocation::getWorldPoint).toArray(new WorldPoint[0]);
+		return digLocations.stream().map(hotColdLocation -> hotColdLocation.getWorldPoint()).collect(Collectors.toList()).toArray(new WorldPoint[0]);
```
Adam-: What?

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -195,25 +194,20 @@ public void onExperienceChanged(ExperienceChanged event)
...
-			session.incrementLapCount(client);
-		}
-		else
+		if (session == null || session.getCourse() != course)
 		{
 			session = new AgilitySession(course);
 			// New course found, reset lap count and set new course
 			session.resetLapCount();
-			session.incrementLapCount(client);
+			session.calculateLapsTillLevel(client, skillGained);
```
Adam-: Why is this being calculated twice?

PR #8079, opened 2019-03-03
Update CheckStyle XML link to one that exists

chivesrs:
https://checkstyle.org/dtds/configuration_1_3.dtd is a link that exists (see https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml for an example). The old puppycrawl links 404 (both HTTP and HTTPS).


deathbeam:
Updated the PR to also remove puppycrawl description and replaced it with actual description as suggested on http://checkstyle.sourceforge.net/config.html

PR #8081, opened 2019-03-03
Replace Java asserts in test code with JUnit asserts.

chivesrs:
Replace assert keyword with JUnit assertX functions.


runelite-client/src/test/java/net/runelite/client/plugins/hiscore/HiscorePanelTest.java
```diff
@@ -31,7 +31,6 @@
 	@Test
 	public void testConstructor()
 	{
-		new HiscorePanel(new HiscoreConfig() {});
+		new HiscorePanel(new HiscoreConfig() { });
```
Adam-: This change should not be in this commit/pr

chivesrs: Reverted in second commit.

PR #8088, opened 2019-03-03
xptracker: correct total xp per hour calculation

### LINKED ISSUE #8086: XpTrackerPanel per hour label doesnt use session time
### Steps to Reproduce
1) train a skill at 100k/hr for 1 hour 
2) train a different skill at 50k/hr for 1hour

### Expected behavior
Some people are expecting the xp/hr to show 75k since you spent 2 hours and gained 150k xp. Currently it shows 100k since the rates are technically 50k/hr in both skill.

### Possible Solutions
Add a new label below the existing one showing the total xp/hr for the entire session or update the existing one

Adam-:
This refactors the total xp tracker to be the same as how skills are tracked, and fixes it to handle >2.1b total xp.

Fixes #8086


runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -504,14 +504,8 @@ public int getSkillExperience(Skill skill)
...
 		{
-			int totalExperience = 0;
-
-			for (int experience : experiences)
-			{
-				totalExperience += experience;
-			}
-
-			return totalExperience;
+			logger.debug("getSkillExperience called for {}!", skill);
```
deathbeam: This debug log is probably just spam

runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -504,14 +504,8 @@ public int getSkillExperience(Skill skill)
...
+			return (int) getOverallExperience();
```
deathbeam: Why not just change return value to long? This basically intentionally returns wrong value, and also then you need to do stuff like you do in unpause method

Adam-: I thought about this but it is used a lot of places, most of which aren't getting overall xp.

runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
Adam-: This would indicate incorrect usage, so it should be logged.

deathbeam: Oh i see, with your other reply it makes sense. Maybe I would then make this method not accept Overall, e.g do assert

Adam-: I'm not that confident that it won't break something currently if we did that.

deathbeam: Well ok

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -302,10 +301,18 @@ public void onExperienceChanged(ExperienceChanged event)
 		}
 
 		final XpUpdateResult updateResult = xpState.updateSkill(skill, currentXp, startGoalXp, endGoalXp);
-		final boolean updated = XpUpdateResult.UPDATED.equals(updateResult);
-		xpPanel.updateSkillExperience(updated, xpPauseState.isPaused(skill), skill, xpState.getSkillSnapshot(skill));
-		xpState.recalculateTotal();
-		xpPanel.updateTotal(xpState.getTotalSnapshot());
+		xpPanel.updateSkillExperience(updateResult == XpUpdateResult.UPDATED, xpPauseState.isPaused(skill), skill, xpState.getSkillSnapshot(skill));
+
+		if (skill == Skill.CONSTRUCTION && updateResult == XpUpdateResult.INITIALIZED)
```
deathbeam: Hmm, if you reset construction wouldnt this cause reset of overall too? I would maybe flip this, e.g check for isInitialized(OVERALL) before checking for skill being construction

deathbeam: Actually, if you reset skill this would never return initialized again until resetState method call, so I guess this is fine

PR #8091, opened 2019-03-04
NMZ High HP Notifier

robrichardson13:
- NMZ ability to set a max HP level to receive a notification at
- Useful for when using DH and want to be notified when your HP goes from 1 to 2


robrichardson13:
Looks like when overloads wear off, it will cause this notification to fire as well, which isn't ideal. I'm going to see if I can implement a simple fix to avoid this.

PR #8094, opened 2019-03-04
NMZ 1 HP Notifier

robrichardson13:
- Adds a toggle in the NMZ plugin to get notified when your HP goes above 1
- Checks if HP went from 1hp -> 2hp (wont fire if overloads is the reason HP went up)


robrichardson13:
Created a bigger feature in #8123. Closing this PR

PR #8096, opened 2019-03-04
Pest control

yanikore:
Big Pest Control update.

One of the main features of this plugin is the hint arrow, which even points to the next portal if it isn't active yet. Spinners & the portal rotation overlay are the same as before, except I fixed I little bug in the overlay. 

Config:
image

Points Infobox, always visible on island and in the minigame
image

Highlighted gangplanks + combat indicator
image
image

Highlight portals (active/shielded/both) and a hint arrow that points to it
image

Highlights repairables
image

Highlights Brawlers and the spot they occupy (as they block your path)
image

Show portal weakness
image

Show time till next portal
image

Also made the contribution bar show you that you failed when you've hit 0% (it just showed 6% before):
image

The NPC id's for the Veteran portals still need to be tested/added as I believe they are different.

I also got full void and all helmets while coding this which is pretty nice :P


james-munson:
gz on full void, this will be a big help with i go for void myself c:

LShirohana:
This would be very nice and I don't see why it hasn't gained traction, all of it is stuff I'd like to do myself but it's already done.

One thing I question is why you added your name to the copyright text box. Adam put the copyright notice over all the files because the project was created by him, not individually all code.

yanikore:
> One thing I question is why you added your name to the copyright text box. Adam put the copyright notice over all the files because the project was created by him, not individually all code.

It's the right way of doing it. Have a look at the other plugins.

LShirohana:
I stand corrected, thanks for pointing this out.

yanikore:
> I stand corrected, thanks for pointing this out.

No problem. Besides adding a file you can also add your name if you edit one. You can't remove somebody else's name though.

yani:
Would this be better as a plugin for the plugin hub?

runelite-client/src/main/java/net/runelite/client/plugins/pestcontrol/WidgetOverlay.java
```diff
@@ -199,6 +156,45 @@ private void renderWidgetOverlay(Graphics2D graphics, Portal portal, String text
...
+		if ((int) (prgs.getBounds().getX()) - bounds.getX() != 2)
+		{
+			percentage = 0;
+			text = "FAILED";
+		}
+		else
+		{
+			percentage = (int) ((prgs.getBounds().getWidth() / bounds.getWidth()) * 100);
+			text = String.valueOf(percentage) + "%";
+		}
```
hawkins, line 179: Really love this change to replace the failed `6%` with `FAILED`, feels like a bugfix more than a feature addition like the rest of this PR. Would love to see this piece merged someday, even without the other changes this PR brings

KWall1202:
These features would be awesome to have, either merged in here or made into a separate plugin on the plugin hub. Using the current plugin and not being able to toggle the spinner highlights is killing me

PR #8104, opened 2019-03-04
clanchat: update counter when player in scene joins / leaves clan chat

SebastiaanVanspauwen:
Currently when a player in already scene joins the clan chat, the counter does not get updated.
This fix checks on the ClanMemberJoined / ClanMemberLeft event, gets the clanmember name, checks all names in scene and adds / removes the player from counter if there is a match.


deathbeam:
This can be done in better way. we dont need player instance, it was used just because it was easy to remove from list based on reference. You can just use name directly

PR #8107, opened 2019-03-04
Do not check for cannonPlaced when cannon is spawned

deathbeam:
This prevents cannon being "stuck" if you do not pick it up and it
despawns and goes back to dwarf.


deathbeam:
Closing in favor of #8108 as it is way better solution

PR #8111, opened 2019-03-05
Support specifying API endpoins via java props

deathbeam:
- Add support fo specifying runelite.session.url java prop to be used
instead of RuneLite session base
- Add support fo specifying runelite.http-service.url java prop to be used
instead of RuneLite api base url
- Add support fo specifying runelite.ws.url java prop to be used
instead of RuneLite websocket url
- Add support for specifying runelite.static.url java prop to be used
instead of RuneLite s.r.n url

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/RuneLiteAPI.java
+++ b/http-api/src/main/java/net/runelite/http/api/RuneLiteAPI.java
@@ -95,9 +95,9 @@ public class RuneLiteAPI
 			.build();
 	}
 
-	public static HttpUrl getApiRoot()
+	public static HttpUrl getSessionBase()
 	{
-		return HttpUrl.parse(BASE);
+		return HttpUrl.parse(BASE + "/session");
 	}
 
 	public static HttpUrl getApiBase()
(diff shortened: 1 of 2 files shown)


http-api/src/main/java/net/runelite/http/api/RuneLiteAPI.java
```diff
@@ -102,6 +102,13 @@ public static HttpUrl getApiRoot()
 
 	public static HttpUrl getApiBase()
 	{
+		final String prop = System.getProperty("runelite.api.url");
+
+		if (prop != null && !prop.isEmpty())
```
chivesrs: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)

deathbeam: I am aware of that, we don't use Guava in http-api

Adam-:
nothing for getApiRoot()? Or the srn base?

deathbeam:
srn base is not important at all, and API root is just for session counter, and also is not very flexible (because its not possible to change the session suffix) and I do not wanted to refactor that.

deathbeam:
Updated with configurable s.r.n url

Issue #8047, opened 2019-02-28
ChatCommandsPlugin: "!price" without an item name throws an exception


deathbeam:
I was randomly standing around while somebody in public chat wrote `!Price` and it threw an error:

```
2019-02-28 23:36:37 [Thread-3] DEBUG net.runelite.client.RuneLite - Chat message type PUBLIC: !Price
2019-02-28 23:36:37 [Thread-3] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.base/java.lang.String.substring(String.java:1841)
	at net.runelite.client.plugins.chatcommands.ChatCommandsPlugin.itemPriceLookup(ChatCommandsPlugin.java:578)
```

Moved from https://github.com/runelite/runelite.net/issues/216

tchar989:
This would be a very simple fix [here](http://tinyurl.com/y35vq5nk).

Change the following:
`		if (!config.price())
		{
			return;
		}
`

To:
		if (!config.price() || message.length() == 0)
		{
			return;
		}

I don't currently have time to push this code change but that's the solution if someone wants to take a crack at it. This bug isn't really of high impact, but it does bloat the logs unnecessarily and we should try to prevent that wherever possible.

Hydrox6:
That wouldn't work @tchar989, since the message is always at the very least `!price`

tchar989:
@Hydrox6 I see, I also see there's already a pull request that fixes this bug awaiting a revision over a checkstyle error. Once that gets pushed this issue should be resolved.

chivesrs:
This should also be done for other chat commands (and `!task` which is in the slayer plugin).

Here's the stack trace for `!task`:
```
java.lang.NullPointerException: null
	at net.runelite.client.plugins.slayer.SlayerPlugin.taskLookup(SlayerPlugin.java:757)
	at net.runelite.client.chat.ChatCommandManager.lambda$onChatMessage$0(ChatCommandManager.java:123)
	at net.runelite.client.util.RunnableExceptionLogger.run(RunnableExceptionLogger.java:41)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
[...]

PR #8125, opened 2019-03-06
Send message to user if they do not have a Slayer task

chivesrs:
The RuneLite API returns empty string if no task is set, and serializing that into a Task object returns null.

Example of the chat message: https://i.imgur.com/6MNRqTI.png

Resolves `!task` issue from #8047.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -756,9 +756,17 @@ public class SlayerPlugin extends Plugin
 
 		if (task == null)
 		{
+			StringBuilder sb = new StringBuilder();
+			sb.append("Unable to find current Slayer task");
+			if (chatMessage.getName().equals(client.getLocalPlayer().getName()))
+			{
+				sb.append(".");
+				sb.append(" ");
+				sb.append("Do you have one? Check your Enchanted gem");
+			}
 			String response = new ChatMessageBuilder()
 				.append(ChatColorType.NORMAL)
-				.append("Unable to find current Slayer task. Do you have one? Check your Enchanted gem/Slayer ring/Slayer helmet")
+				.append(sb.toString())
 				.build();
 
 			final MessageNode messageNode = chatMessage.getMessageNode();
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -587,7 +587,7 @@ public class SlayerPlugin extends Plugin
 	private boolean doubleTroubleExtraKill()
 	{
 		return WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID() == GROTESQUE_GUARDIANS_REGION &&
-				SlayerUnlock.GROTESQUE_GARDIAN_DOUBLE_COUNT.isEnabled(client);
+			SlayerUnlock.GROTESQUE_GARDIAN_DOUBLE_COUNT.isEnabled(client);
 	}
 
 	private boolean isTarget(NPC npc)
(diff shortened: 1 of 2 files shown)


chivesrs:
Added in a basic test for the functionality.

deathbeam:
I don't think this is correct, because this will tell everyone that see the message to check their slayer helm/gem, what do not makes much sense.

chivesrs:
Fair, I was considering the scenario of both 1: you have no task and 2: RuneLite doesn't detect a task.

I'll just do something like `Current slayer task not detected` and you can comment on the wording.

Nightfirecat:
The issue is that it will display that error message to everyone, and not just to the player who sent `!task`. See #5327

chivesrs:
Oh I see, you are saying chat message to all, vs a non chat message to only the user. Let me find out what API to use for that.

Nightfirecat:
You shouldn't need to use a different API, just don't set the message if the sender is not the local player.

chivesrs:
Looks like I have the behavior we are going for:
https://i.imgur.com/YRIOSDM.png

(left client is my code, right is head)

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -754,6 +754,28 @@ void taskLookup(ChatMessage chatMessage, String message)
 			return;
 		}
 
+		if (task == null)
```
Adam-: I don't think this can happen

chivesrs: It can. See stack trace https://github.com/runelite/runelite/issues/8047#issuecomment-469944438. It's when the HTTP API returns back an empty string for the task, and trying to deserialize "" into a task object.

chivesrs: To repro:
1. have an account without slayer task
2. check enchanted gem so the RuneLite HTTP API knows you have no slayer task
3. perform `!task`

chivesrs:
Here's how the task can be null btw,

```
2019-03-18 00:32:34 [pool-2-thread-1] WARN  n.r.c.util.RunnableExceptionLogger - Uncaught exception in runnable net.runelite.client.chat.ChatCommandManager$$Lambda$821/765398357@3ff6ee08
	at net.runelite.client.plugins.slayer.SlayerPlugin.taskLookup(SlayerPlugin.java:757)
	at net.runelite.client.chat.ChatCommandManager.lambda$onChatMessage$0(ChatCommandManager.java:123)
	at net.runelite.client.util.RunnableExceptionLogger.run(RunnableExceptionLogger.java:41)
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)
2019-03-18 00:32:43 [Thread-5] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
	at net.runelite.client.plugins.timetracking.farming.FarmingTracker.predictPatch(FarmingTracker.java:190)
	at net.runelite.client.plugins.timetracking.farming.FarmingTracker.updateCompletionTime(FarmingTracker.java:265)
	at net.runelite.client.plugins.timetracking.farming.FarmingTracker.updateData(FarmingTracker.java:139)
	at net.runelite.client.plugins.timetracking.TimeTrackingPlugin.onGameTick(TimeTrackingPlugin.java:195)
	at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:72)
	at net.runelite.client.eventbus.EventBus.post(EventBus.java:218)
	at net.runelite.client.callback.Hooks.clientMainLoop(Hooks.java:148)
	at client.methodDraw(client.java:1339)
	at client.ak(client.java)
	at bf.ag(bf.java:397)
	at bf.run(bf.java:351)
	at java.lang.Thread.run(Unknown Source)
```

image

chivesrs:
@deathbeam or @Adam- could you take a look at this?

PR #8126, opened 2019-03-06
changed response.message() to call toString() instead

### LINKED ISSUE #8042: Improve 404 error messages (WorldClient, FeedClient...)
My branch was outdated and the WorldClient error message didn't show that I was getting a 404.

```
2019-02-28 18:44:14 [Thread-3] DEBUG n.r.http.api.worlds.WorldClient - Error looking up worlds: 
```

A quick fix is changing the `response.message()` in `WorldClient.java:60` to `response.toString()` so the full error is shown:
```
2019-02-28 18:44:14 [Thread-3] DEBUG n.r.http.api.worlds.WorldClient - Error looking up worlds: Response{protocol=h2, code=404, message=, url=https://api.runelite.net/runelite-1.5.14-SNAPSHOT/worlds.js}
```

However, other classes also throw empty error messages, like FeedClient. A different approach might be to populate the message when it's empty and a specific response code is found.

ndngo:
Closes #8042

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/feed/FeedClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/feed/FeedClient.java
@@ -55,7 +55,7 @@ public class FeedClient
 		{
 			if (!response.isSuccessful())
 			{
-				logger.debug("Error looking up feed: {}", response.toString());
+				logger.debug("Error looking up feed: {}", response);
 				return null;
 			}
 
--- a/http-api/src/main/java/net/runelite/http/api/feed/FeedClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/feed/FeedClient.java
@@ -55,7 +55,7 @@ public class FeedClient
 		{
 			if (!response.isSuccessful())
 			{
-				logger.debug("Error looking up feed: {}", response.message());
+				logger.debug("Error looking up feed: {}", response.toString());
 				return null;
 			}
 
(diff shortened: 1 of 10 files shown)


http-api/src/main/java/net/runelite/http/api/feed/FeedClient.java
```diff
@@ -55,7 +55,7 @@ public FeedResult lookupFeed() throws IOException
 		{
 			if (!response.isSuccessful())
 			{
-				logger.debug("Error looking up feed: {}", response.message());
+				logger.debug("Error looking up feed: {}", response.toString());
```
chivesrs: Doesn't the logger call toString() natively on the object passed in? It seems redundant to have a toString() here if that's the case.

Ref: https://github.com/runelite/runelite/issues/8042#issuecomment-468382802

deathbeam: It does

ndngo:
Ah, I see. That makes sense.

http-service/src/main/java/net/runelite/http/service/feed/twitter/TwitterService.java
```diff
@@ -146,7 +146,7 @@ private void updateToken() throws IOException
 		{
 			if (!response.isSuccessful())
 			{
-				throw new IOException("Error authing to Twitter: " + response.message());
+				throw new IOException("Error authing to Twitter: " + response.toString());
```
deathbeam: This still has toString call

Issue #8124, opened 2019-03-06
Ground markers in instanced areas still broken after #8046


ShaunDreclin:
**Screenshots**
![screenshot](https://i.imgur.com/NhebXyO.gif)

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.5.15-SNAPSHOT

**Additional context**
Issues: #7985, #8045
PR: #8046

Nightfirecat:
This appears to have been introduced in #5890, specifically by commit ad65a9a. I'm looking into the source of this issue and a fix now.

Nightfirecat:
The bug introduced by the original multi-colored ground markers PR was reverted in 33988a72e1ece243ef24d5441b7d6f80420a950d so this is no longer present on master.

ShaunDreclin:
Should I close this now or wait for #8128 to be merged

deathbeam:
Well, its not happening anymore. Just double-check it on new nightfirecat's PR.

PR #8127, opened 2019-03-06
ground markers: Fix marker saving in instances

Nightfirecat:
Fixes runelite/runelite#8124


Nightfirecat:
Will be fixed in the colored ground markers re-do PR

Issue #1848, opened 2018-04-23
Re-add Silver Sickle (b) in Menu Entry Swapper


vanzoest96:
This was in here a few days ago and now it's gone? Can it be added again?

Adam-:
This was removed after we gained the ability to configure custom shift clicks, since you can just do that now. We are thinking of adding a permanent-shift click mode in #1556 which would address this.

vanzoest96:
Just figured this out. Thank you.

PR #8132, opened 2019-03-07
Added menuentryswapper entry to swap wield and cast bloom on silver sickle(b)

Crollt:


Hydrox6:
Use Shift Click config instead. See the discussion on #1848

PR #8139, opened 2019-03-07
wiki: Update the Fixed Minimap Clickbox to account for the spec orb

### LINKED ISSUE #8137: Special attack orb clicks on minimap
**Describe the bug**
If you click on the spec attack orb in fixed mode then it will click on the minimap and move the character. I tested this on vanilla and it doesnt appear to do that there

**To Reproduce**
Fixed mode, click on orb

**Expected behavior**
moves the character instead of only clicking the button

Hydrox6:
Closes #8137


raiyni:
It looks like you just overlayed the old cache image on top of what's in master. This is what should be used. 

fixed_mode_minimap_clickmask

photo_xy4zxvtanv

PR #8147, opened 2019-03-08
use RuneStar cs2 opcode names

Abextm:
With exception of the opcodes with jvm analogues, which follow jvm style


Adam-:
As another fun project, it would be neat to automatically replace the names of things with the RuneStar names.

Joshua-F:
LGTM

PR #8154, opened 2019-03-09
Check message length before use.

### LINKED ISSUE #8047: ChatCommandsPlugin: "!price" without an item name throws an exception

pwendelboe:
If we don't check the length of the message before we access it with substring(command.length() + 1) we will get a string index out of range error.
This fixes issue #8047 and #8057.

(Also as a note it's my first pr here so if I did something wrong please let me know thanks!)

Closes #8047

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -366,6 +366,11 @@ public class ChatCommandsPlugin extends Plugin
 			return;
 		}
 
+		if (message.length() <= KILLCOUNT_COMMAND_STRING.length())
+		{
+			return;
+		}
+
 		ChatMessageType type = chatMessage.getType();
 		String search = message.substring(KILLCOUNT_COMMAND_STRING.length() + 1);
 
@@ -483,6 +488,11 @@ public class ChatCommandsPlugin extends Plugin
 			return;
 		}
 
+		if (message.length() <= PB_COMMAND.length())
+		{
+			return;
+		}
+
 		ChatMessageType type = chatMessage.getType();
 		String search = message.substring(PB_COMMAND.length() + 1);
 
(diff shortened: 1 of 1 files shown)


pwendelboe:
As a side note I would've preferred to do this more dynamically so that it would work with future commands, but I just couldn't figure it out.
~P

deathbeam:
You can just use < everywhere and not add 1. also put newlines around the ifs

pwendelboe:
Ok will do. Thanks!

pwendelboe:
Ok changed. Still need the +1 as otherwise it seems to generate the same error. Since the message of "!price".length() == PRICE_COMMAND_STRING.length() it would fail after changing to just a <. Also added comments to just be explicit about the reason we are testing that.

If there's anything else just let me know.

msosa:
If `x < x + 1` is working than you should be able to just do `x <= x`

pwendelboe:
True. Changed.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -629,6 +645,11 @@ private void playerSkillLookup(ChatMessage chatMessage, String message)
 		{
 			return;
 		}
+		
+		if (message.length() <= LEVEL_COMMAND_STRING.length())
```
Adam-: Even though I think this works, this should be moved into the case for message starting with `LEVEL_COMMAND_STRING`

pwendelboe: Oh because it is also called when you do !total. I see, missed that initially. Thanks.

PR #8155, opened 2019-03-09
Add watch list config option to clan chat plugin

TheStonedTurtle:
You may want to add people, such as friends, who don't meet the rank threshold to the join/leave messages. This facilitates that.


runelite-client/src/main/java/net/runelite/client/plugins/clanchat/ClanChatPlugin.java
```diff
@@ -113,6 +113,7 @@
 
 	private List<String> chats = new ArrayList<>();
 	private List<Player> clanMembers = new ArrayList<>();
+	private List<String> watchList = new ArrayList<>();
```
raiyni: This should be a set

PR #8160, opened 2019-03-09
 clanchat: Add world info to join/leave messages

TheStonedTurtle:
Adds the world information to the join/leave messages. Has a config option that is on by default.


TheStonedTurtle:
Added config option to ignore rank requirement on join/leave messages based on world type.

runelite-client/src/main/java/net/runelite/client/plugins/clanchat/ClanChatPlugin.java
```diff
@@ -121,6 +131,8 @@
 	private Map<String, ClanMemberActivity> activityBuffer = new HashMap<>();
 	private int clanJoinedTick;
 
+	private final Map<Integer, EnumSet<WorldType>> worldTypeMap = new HashMap<>();
```
Nightfirecat: I believe this can be a SetMultiMap

Issue #8019, opened 2019-02-27
change chaos fanatic slayer task icon to match boss timers icon


Alexsuperfly:

nateisonfire:
Not sure how to assign this to myself... but I created a PR for it here: https://github.com/runelite/runelite/pull/8162

PR #8162, opened 2019-03-10
Changed Chaos Fanatic's slayer task info box sprite from pet to ancient staff to match the respawn timer

nateisonfire:
`\runelite\runelite-client\src\main\java\net\runelite\client\plugins\bosstimer\Boss.java` contains `ItemID.ANCIENT_STAFF` as it's sprite as well 👍 

Closes: #8019


nateisonfire:
Closes: #8019

raiyni:
Boss timer should be changed to pet.

deathbeam:
Nope, because then both chaos fanatic and chaos elemental would have same icons

nateisonfire:
The Chaos elemental has the pet (cloud looking thing) as it's sprite for both Boss Timer and Task Info Box. 

Now, the chaos fanatic has the ancient staff as both the boss timer and task info box sprites.

Boss monsters drop the same pet, but the pet is a direct render of the chaos elemental. So the ancient staff looks more like the chaos fanatic.

PR #8169, opened 2019-03-10
Revert "Remove raids points overlay"

### LINKED ISSUE #8164: Bring back RuneLite CoX points overlay
The official overlay looks really bad and inconsistent with other RL overlays like timers.
Can't put it over your chat box either though that is a small issue.

At least let you choose between them it really looks terrible. Ty.

MagicfTail:
Our old overlay:

1. Looked better
2. Had more functionality
3. Was able to be moved more freely

closes #8164 

This reverts commit fbd3ea6202a07b4dc4d3aab4eeebf62afa793c7c.


Trevor159:
The RL points overlay was added to the plugin hub so you can use it from there if it is wanted.

Issue #8173, opened 2019-03-10
[BUG] Friend Notes Don't save due to spaces


pancake4me:
**Describe the bug**
I love using the friend notes to remember different people on my friends list. It helps alot because not all of them are active 24/7 so i might forget them if they leave for a few months. However certain individuals on my friends list dont keep the note I setup for them. I noticed it was olny on the freinds that have a space in their name (ex. "Mrs BlahBlah"    or "Dr BlahBlah"    or "Blah Blah")

**To Reproduce**
1. Go to your friends list
2. Right click and add notes to certain people who have a space and those that dont
3. Log out and Relaunch the client. 
4. Those that didnt have a space in their name keep their notes while those that did dont keep their notes. their just gone. 

**Expected behavior**
Normally they would save like any other friend notes. 

**Environment (please complete the following information):**
Not sure if the problems are just for my PC and OS or universally wide 
- OS: Windows 10 64 bit
- version 64 bit .exe launcher.

Adam-:
I can't reproduce this using the given steps.

Seven-Ate:
> I can't reproduce this using the given steps.

It was easily reproducible for me, any time my client closed and re-opened I lost notes

tehjagjr:
Can confirm, reproduced using the given steps by @pancake4me. I did notice that it does not affect all users with spaces in their name only a subset of users. 
![Issue8173](https://user-images.githubusercontent.com/7780322/54157720-cda13280-4440-11e9-9e4a-e94377caefc4.gif)

Adam-:
Do you have the name of a person it doesn't work on?

pancake4me:
> Do you have the name of a person it doesn't work on?

These are the only friends for me that the friend notes don't save and each one has a space. 

Alfano alt
A Jug
Take Jug
Mrs Theodora
Dr Craig

Adam-:
This was due to the way the config database is encoding non breaking spaces. I've fixed it in 6371a7dd0b0d49259e2a3326c62e1f3e2c0f28ac just by using normal spaces.

PR #8174, opened 2019-03-11
Fix friend notes not saving on display names with spaces

critzd:
Resolves issue #8173 

The display names were not being formatted correctly when utilizing the name as a key, as however the key is saved resulted in a format looking like "LordÂ Vionas" When it should be retrieving/saving a key of "Lord Vionas". Using Text.toJagexName resolved the formatting issue with the characters. I assume that would have been an issue with unicode or something, but I wasn't sure.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
@@ -29,9 +29,16 @@ package net.runelite.client.plugins.friendnotes;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.ObjectArrays;
+import java.awt.Color;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.*;
+import net.runelite.api.Client;
+import net.runelite.api.Friend;
+import net.runelite.api.MenuAction;
+import net.runelite.api.MenuEntry;
+import net.runelite.api.Nameable;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.MenuOptionClicked;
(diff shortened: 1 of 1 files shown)


Adam-:
I'm not seeing where the original bug is in the code. Also won't this break a bunch of existing notes?

critzd:
**Note:** I still need to add comments to explain the code change above via another commit, so I'll be doing that right now.

So this wouldn't actually break existing notes as notes made for names with spaces such as "Lord Vionas" wouldn't be saving properly anyway. Those without spaces and just numbers and letters are also unaffected as their characters aren't being modified when we use the .toJagexName method.

Just as a way to reproduce this with the current build to be more clear:
1. Log into the Runelite client with your runelite account/google account
2. Log into Runescape via the client
3. Find two friends, one with a name without spaces, e.g. "CluelessNoob", and then one with a space, e.g "Experienced Player"
4. Add a note for each player. Note that at this time, you **will** see the note for both players. You must log out for this bug to happen.
5. Log out of runescape, and close the runelite client.
6. Log back in (Follow steps 1 and 2 once again)
7. Hover over both names, find that only the non space name was working.

The way I noticed the issue was when the config value was being loaded upon login to runescape. You can see the non functioning and functioning config values in the image below:
image

Note how the ones with spaces were adding an extra character "Â" before the space, and that is the _current_ result when using the current build. I wasn't 100% on it other than when you go and look at the char array after we format the display name with .toJagexName was utilizing a normal ASCII space, where prior it was using a non breaking space (Unicode?).

You can see how the key is correct when we utilize the .toJagexName now as well, avoiding that "Â" appended to the beginning, loading the correct config values.

Now what I was unsure about was the reason it was adding that specific character. Is it something with how it saves key values? I just know that the way that it is implemented in the fix it corrects the issue without breaking anything, since names with spaces don't save anyway.

Adam-:
This was fixed in d617d1960ffa9bc50aab478b3c4bef5ce9d35e7c

PR #8181, opened 2019-03-12
Reverse order of loot tracker grouped view

Seven-Ate:
This puts monsters that were recently killed near the top of the list, instead of having to scroll down to find loot from your current session. Thought a config was superfluous, can add if it's wanted.


deathbeam:
Never make pull requests from master

Hydrox6:
At some point between now and this PR's creation, functionality was added that does exactly this.

PR #8186, opened 2019-03-12
Transparent text support for TextComponent

CalebWhiting:
I changed how TextComponent is rendered, so that the text can be transparent, without being affected by the border color.

Originally the text was simply rendered on top of the border, which meant that text couldn't be transparent, it would simply result in a darker variant of the text.

With the new technique the text can be completely transparent, showing only the border.

Opaque Text
image

Half Opaque
image

Transparent
image

I've also added a couple of features to the ground items plugin
* Font style customisation
* Alpha channel customisation

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -26,9 +26,11 @@
 package net.runelite.client.plugins.grounditems;
 
 import java.awt.Color;
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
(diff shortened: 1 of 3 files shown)


psikoi:
Why is this useful? Am I missing something here?

CalebWhiting:
> Why is this useful? Am I missing something here?
It's useful in the same way that being able to choose the color is, customization.

CalebWhiting:
Here's a example use case, while it's an extreme one you get the point.
image
image

raiyni:
That's what the ignore list is for?

CalebWhiting:
> That's what the ignore list is for?

I said it was an extreme example, I don't understand the opposition to this, it's something the current implementation cannot do, therefore it's better.

CalebWhiting:
Another place that this is useful is the wilderness, if there's a lot of loot around it can hide your character, leaving you unable to see if you're being attacked.

PR #8191, opened 2019-03-13
Unidentified herbs plugin

hex-agon:
This is a fun plugin that modifies grimy herbs item compositions so they look like the old unidentified herbs that were present in the game at the launch of oldschool. 
It also intercepts chat messages when the player fails to clean a herb due to not having the required level and replaces it with a generic message that doesn't give it away which herb it is.

Currently this plugin **cannot be merged** because it is relying on reflection to set some item composition fields that are not exposed by the runelite's api, the following fields are being reflected:

* name (opcode 2)
* modelId (opcode 1)
* sourceColors (opcode 40)
* targetColors (opcode 40)
* inventoryActions (opcodes 35-39)

It would also need access to the `itemSpriteCache` to reset it when enabling/disabling the plugin.

##### Known issues

* After enabling/disabling the item sprite will not change imediately (fixed by reseting the itemSpriteCache)
* This plugin affects the grand exchange search results

##### Images of the plugin
Plugin image 1
Plugin image 2
Plugin image 3

If this plugin is something that is not wanted on runelite or if it is not possible to expose those item compositions fields please close this PR.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/unids/UnidHerbsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/unids/UnidHerbsPlugin.java
@@ -1,0 +1,104 @@
+package net.runelite.client.plugins.unids;
+
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import javax.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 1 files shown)


Adam-:
This is neat but it seems pretty niche, I can't imagine anyone really wanting such a thing?

PR #8198, opened 2019-03-13
Remote Bank Item Quantity Viewer Plugin

Jake-Arnold:
Currently fairly basic but does the job.

Populates a hash map with the current bank contents and quantities which are then displayed in a tooltip when hovering over an inventory item or when the item is examined (can be changed in plugin options).

Useful for gathering things and knowing how many are left to collect etc.

Below is an example of the tooltip.

image 
nJDCXrw

BXiuhNC


khunnaball:
Looks good to me, very useful.

Jake-Arnold:
I need to update the formatting still, apparently. Currently working so I will update it soon. It shouldn't affect functionality as the only code that I've edited at all is the code within my four files.

Jake-Arnold:
Updated the main post with two more screenshots of examples of how this currently displays.

Jake-Arnold:
Next update will be along these lines:

image

Beagon:
The ChatMessageTypes have been refactored which makes this module not compile. I've fixed this for you in the following PR: https://github.com/Jake-Arnold/runelite/pull/2

Jake-Arnold:
> The ChatMessageTypes have been refactored which makes this module not compile. I've fixed this for you in the following PR: [Jake-Arnold#2](https://github.com/Jake-Arnold/runelite/pull/2)

Thanks for this. I've merged your PR. I've been really busy finishing off other work so I haven't been able to keep up with the commits/further development of this for about 10 days.

Nightfirecat:
This would be a good candidate for a plugin hub submission.

Nightfirecat:
Actually, the Bank Value plugin which was added to the plugin hub basically implements this functionality, so I would recommend any users wanting this feature to consider using that plugin.

PR #8200, opened 2019-03-13
Window snap feature

CoreyForsyth:
I added a window-snap feature to the default runelite client.

With this plugin, you can drag the client to the side or the corner of your screen, and it will snap to side or the corner. I also added some keyboard shortcuts, ctrl + arrow keys to do the same thing. It is very much like the default Windows snap feature, however this does work with the custom chrome that RuneLite uses.

Here is a screenshot of what the settings look like:
image

The custom size will be used if enabled, rather than snapping to half of the screen size. This is so if people like their client a particular size, they can enable it, and then still snap it so that it goes to the corner of the screen.

I have only tested this on windows, so it would be awesome if someone could test on mac (and possibly linux as well).

I posted what this looks like on the subreddit [here](https://www.reddit.com/r/2007scape/comments/b0he2m/would_anyone_be_interested_in_runelite_having_the/)

Let me know you think

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/windowsnap/WindowSnapConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/windowsnap/WindowSnapConfig.java
@@ -12,7 +12,7 @@ public interface WindowSnapConfig extends Config
 			position = 1,
 			keyName = "enableMouseSnap",
 			name = "Mouse Drag Snaps Window",
-			description = "The window will snap to the side of the screen with the mouse")
+			description = "The window will snap to the side/corner of the screen with the mouse")
 	default boolean mouseEnabled()
 	{
 		return true;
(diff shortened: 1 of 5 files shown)


raiyni:
This plugin will fail on screens with different resolutions and scaling from what I can tell.

deathbeam:
Imo just disabling custom chrome is enough? And then simply let your OS to manage your window. This is pointless because afaik with windows layouting it also relayouts windows around, but this doesnt. Basically it will be always way worse than provided OS functionality.

ShaunDreclin:
The main complaint with disabling custom chrome is the default window border being pretty ugly, but I don't think that warrants the effort it would take to properly integrate this with native window snapping on all operating systems.

CoreyForsyth:
@raiyni I was under the impression that java awt pixels are standardized even with scaling, but I could be wrong. Also I realize now that I need to set the bounds to be from the system Insets instead of assuming it is 0, that might help what you are talking about.

@deathbeam @ShaunDreclin This would definitely have to be disabled by default, but it does seem like a lot of people like it. I realize that most linux WMs are able to manage the window well enough, but it is not great on windows as it stands. 
Another thing to keep in mind is the fact that you can set the custom resolution instead of forcing the client to be half of the screen. I find it useful because I usually keep the client on one side of the screen with the width set to minimum, and stretching the height to maximum. For single-monitor setups, it definitely useful

raiyni:
Swing/Java has sizing issues between versions when using multiple monitors.

Adam-:
I'd rather leave this up to the WM as I don't want to make RL aware of all possible WMs it is running in and how they behave.

PR #8203, opened 2019-03-13
Add support for item quanities for the !price command

DarkSlaayer:
example !price cannonball 5000

image


Adam-:
This looks very complex, I do not think it is that difficult. What is the syntax this uses?

DarkSlaayer:
> This looks very complex, I do not think it is that difficult. What is the syntax this uses?

Just add the quantity to the end like !price cannonball 5000

Yes and I thought about it after you said that and I think it could be done a better way. 
Maybe check if the last argument of the command is a number if so do a substring and remove the last argument and do a item look up to see if it returned anything? if so use the last argument as an amount?

Adam-:
Something like `!price 5000 cannonball` seems much more natural.

DarkSlaayer:
> Something like `!price 5000 cannonball` seems much more natural.

Alright when I get a chance ill do that and check the arguments a better way.

PR #8238, opened 2019-03-16
Add menu swap for magic boxes

Lucwousin:
Swaps take with activate on [magic box](os.rs.wiki/w/Magic_box) ground items, like take and lay are swapped on box traps.


deathbeam:
You should also update Take description

Lucwousin:
It's under swapBoxTrap key, make that into swapHunter and make people have a dead config key in their config or change it's name to hunter while keeping the key name?

PR #8239, opened 2019-03-17
Add party support to special attack counter plugin

Trevor159:
Currently only takes party member specs if the last npc interacted with of the speccer and the local player are the same, improvements to this would be nice.

Still works without the party plugin.

Thanks turtle for help testing :d


runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterUpdate.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.Value;
+import net.runelite.http.api.ws.messages.party.PartyMemberMessage;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class SpecialCounterUpdate extends PartyMemberMessage
+{
+	private final int npcId;
+	private final SpecialWeapon weapon;
+	private final int experience;
```
Adam-: Wouldn't it be better to just send either successful hits or total dmg hit?

15987632: The only reason I decided to go with experience is because that is what was already sent to the update counter method so I had to modify less of the current implementation.

15987632: I have implemented this change

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -42,18 +42,24 @@
...
+import net.runelite.client.ws.PartyService;
+import net.runelite.client.ws.WSClient;
 
 @PluginDescriptor(
 	name = "Special Attack Counter",
 	description = "Track DWH, Arclight, Darklight, and BGS special attacks used on NPCs",
 	tags = {"combat", "npcs", "overlay"},
 	enabledByDefault = false
 )
+@PluginDependency(PartyPlugin.class)
```
Adam-: Is this required?

15987632:
I have made this slightly less complex, now the tooltip looks like this

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounter.java
```diff
@@ -24,43 +24,79 @@
...
+				}
+				else
+				{
+					tooltip += name + ": " + hitValue + " " + weapon.getName() + " specs.";
+				}
 			}
 			else
 			{
-				return weapon.getName() + " special has hit " + hitValue + " times.";
+				tooltip += tooltip += name + ": " + hitValue + " " + weapon.getName() + " spec damage.";
```
chivesrs: Is the extra `tooltip +=` intentional? If so, please move to own line

PR #8240, opened 2019-03-17
Add option to highlight object clickbox

bjornenalfa:
Sometimes an objects hull can be very different from the clickbox and it can be nice to see the clickbox instead.

Also adds some basic framework for future customization of object highlighting.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
@@ -25,6 +25,7 @@
 package net.runelite.client.plugins.objectindicators;
 
 import java.awt.*;
+import java.awt.geom.Area;
 import javax.inject.Inject;
 
 import net.runelite.api.Client;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
```diff
@@ -24,10 +24,10 @@
  */
 package net.runelite.client.plugins.objectindicators;
 
-import java.awt.Dimension;
-import java.awt.Graphics2D;
-import java.awt.Polygon;
+import java.awt.*;
```
l2-: Don't let the mods see a wildcard import

Adam-:
This seems to be for object indicators? I would rather have a config opt for it probably than a Mark vs Mark clickbox for each object, if it is something we want.

bjornenalfa:
Do you mean a config option to set if you want to highlight the object or the clickbox, or do you mean an option to switch the way it is rendered?

I am not sure why you would want to highlight something other than the clickbox of an object or objects that you can't click, but in those cases it would seem strange to highlight the clickbox instead of the hull.

I think the most useful thing would be to not have an option and simply highlight the clickbox when the object is clickable, but there does not seem to be anything in the API for checking if it can be clicked or not.

Adam-:
I meant a config option to decide which is rendered. The hull stuff really only exists because it predates our ability to determine clickboxes, but also it is significantly faster. I think you can determine whether or not an object has actions by looking at the object composition.

PR #8247, opened 2019-03-18
added various bosses & npcs to the boss enum in special counter

lordzuku:
added various bosses & npcs from https://docs.google.com/spreadsheets/d/1wzy1VxNWEAAc0FQyDAdpiFggAfn5U6RGPp2CisAHZW8/edit?ouid=118202333630091586149&usp=sheets_home&ths=true

to the specialattack boss enumation


Nightfirecat:
PR is stale and needs rebasing. Also, I think we'd probably want to dynamically calculate the EXP modifier for monsters ala #8463 instead of maintaining an ever-growing list.

PR #8249, opened 2019-03-18
Use consistent rounding for XP bar markers

### LINKED ISSUE #7479: XP Tracker - Intermediate levels are off

paulbcook:
Closes #7479

XP bar progress is stored as an integer of percentage, while the intermediate markers are stored as a double from 0 to 1. This means that even if the progress is higher than the intermediate level (but not by more than 1%), the intermediate line will render to the right.

This PR makes both as a percentage integer, which means both are rounded in the same fashion, preventing any discrepancy in rendering.

(This is a duplicate PR because I messed up my branches. The old one has been closed.)


Adam-:
@forsco can you test this?

forsco:
@Adam- I’ll test it when I get home from work.

forsco:
After testing this, I could not reproduce the same scenario as per the screenshot supplied in issue #7479.

PR #8252, opened 2019-03-19
WIP: Add raid data storer to raids plugin

Trevor159:
Hopefully this isn't completely wrong, but I wouldn't be surprised if it is.

The goal is to have something like this except not terrible on the website behind oauth https://streamable.com/bwklw 

Right now this stores:
username
kc
whether the raid was challenge mode or not (this can probably be removed as you know if it was cm or not based on if the raid had a middle floor or not)
the downtimes of each floor and the raid time
personal points
team points
team size
the rotation of the raid
and whether you preped in the raid or not (this can be removed if it isn't wanted, as a soloer its nice to have)

These are obviously subject to change and I want feedback for them, right now I am storing the rotation as the long string it is. This is obviously not optimal.

Someone will probably have to touch on the SQL, I am not very knowledgeable on it. I did get rid of the 30 day deletion on the entries as that seemed a lot less needed as very few people even have thousands of raids kc. This is the reason I store the drops even though they might be stored in the loot tracker also.

This was mostly copied from the loot tracker plugin.

I want #7981 merged first so the rotations never show unknown rooms and so raid never gets set to null before the raid is over (current it does if a player logs out or DCs)
Relies on #8122 so the times will use regex and be accurate

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/raidtracker/RaidRecord.java
+++ b/http-api/src/main/java/net/runelite/http/api/raidtracker/RaidRecord.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2018, TheStonedTurtle <https://github.com/TheStonedTurtle>
+ * Copyright (c) 2019, Trevor <https://github.com/15987632>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


Trevor159:
Superseded by the adv log PR

PR #8255, opened 2019-03-19
fixed typos in Hooks

andrewghaly:


runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -210,7 +210,7 @@ private void checkWorldMap()
 
 		if (manager != null && manager.isLoaded())
 		{
-			log.debug("World map was closed, reinitializing");
```
chivesrs: I wouldn't bother with this one due to https://www.google.com/search?q=reinitializing
(and on line 189 remove the hyphen)

andrewghaly: Wasn't sure about the hyphens either. I removed them but kept the correct spelling.

PR #8263, opened 2019-03-19
 Add plugin to display Hitpoints and Prayer in the system tray

### LINKED ISSUE #1340: Health and prayer indicators in tray
Well there are some afk killing spots and would be nice to see your HP while game is minimized, for example : 
https://imgur.com/a/hXc6v
Maybe even show your prayer also. 👍 
P.S. I'm not really into clients etc. so I don't even know if it's possible (Smiling Face With Open Mouth And Smiling Eyes, but with a single drop of sweat on one side of the face).

paulbcook:
Closes #1340

This is a new plugin that sets and updates the system tray icon to an hp bar, prayer bar, or both. It also can display the current value for either. I chose the blocky over anything fancier version for 2 reasons: it's easy to dynamically render, and when it's small it's most readable. This icon includes an offset for mac; windows centers the icon but mac relies on transparent padding (the current icon has this problem if you compare to the default icons).

### Screenshots
**With hitpoints**
Screen Shot 2019-03-19 at 10 46 26 AM

**With hitpoints over 99**
Screen Shot 2019-03-19 at 10 46 40 AM

**With prayer (and config panel)**
Screen Shot 2019-03-19 at 10 46 54 AM

**Showing both bars and no text (and config panel)**
Screen Shot 2019-03-19 at 10 45 51 AM

### Possible enhancements
- Make it nicer looking
  - I chose the colors based on what I thought looked nice, but this can use actual HP bar colors if desired
  - Change the font, I just picked the RS font because it looks decent low-res
  - Text alignments are arbitrary and can be changed as needed
- Add tooltip
  - Might be unnecessary for a quick afk tabbed-out view (which is the idea behind this in general)
  - Can show things like prayer time left, poison, etc
- Color differently if prayers are active
- Color differently if poisoned/venomed
- Re-render only on changed values
  - It's not at all intensive to create/set this icon per-tick (no flickering on mac at least), so checking previous tick values may be unnecessary 
- Unify with another plugin
  - This is similar to the status bars on inventory but didn't have enough in common to justify


runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+
+	private void resetTrayIcon()
+	{
+		clientUI.getTrayIcon().setImage(ClientUI.ICON);
+
+	}
+
+	private int getYOffset()
+	{
+		return OSType.getOSType() == OSType.MacOS ? MAC_HEIGHT_OFFSET : 0;
```
Adam-: Why?

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+
+		g.dispose();
+
+		clientUI.getTrayIcon().setImage(hpImage);
+	}
+
+	@Subscribe
+	public void onGameStateChanged(final GameStateChanged event)
+	{
+		if (event.getGameState() != GameState.LOGGED_IN)
```
Adam-: This should reset at login screen probably instead.

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+			g.setColor(MISSING_PRAYER_COLOR);
+			g.fillRect(prayerX, getYOffset(), prayerWidth, scaledMissingPrayer);
+		}
+
+		if (config.numberToDisplay() !=  StatusNumberMode.NONE)
+		{
+			int numberToDisplay = config.numberToDisplay() == StatusNumberMode.HITPOINTS ? currentHealth : currentPrayer;
+			g.setFont(numberToDisplay < 100 ? STATUS_FONT : STATUS_FONT_SMALL);
+			g.setColor(TEXT_COLOR);
+			int numberDrawY = (numberToDisplay < 100 ? 34 : 30) +  getYOffset();
```
Adam-: These should be constants with a descriptive name so I can tell what they are doing.

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+	public void onGameTick(GameTick event)
+	{
+		if (client.getGameState() != GameState.LOGGED_IN || !(config.showHitpoints() || config.showPrayer()))
+		{
+			return;
+		}
+
+		final int maxHealth = client.getRealSkillLevel(Skill.HITPOINTS);
+		final int currentHealth = client.getBoostedSkillLevel(Skill.HITPOINTS);
+		final int scaledMissingHealth = (int) Math.ceil((Math.max(maxHealth - currentHealth, 0) * HEIGHT * 1.0) / maxHealth);
```
Adam-: You can just cast one side of the mul to a float instead of `* 1.0`

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+	@Inject
+	private ClientUI clientUI;
+
+	@Inject
+	private StatusTrayIconConfig config;
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		if (client.getGameState() != GameState.LOGGED_IN || !(config.showHitpoints() || config.showPrayer()))
```
Adam-: Game tick only ever fires when gamestate is logged in.

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
```diff
@@ -0,0 +1,164 @@
...
+
+		final int maxHealth = client.getRealSkillLevel(Skill.HITPOINTS);
+		final int currentHealth = client.getBoostedSkillLevel(Skill.HITPOINTS);
+		final int scaledMissingHealth = (int) Math.ceil((Math.max(maxHealth - currentHealth, 0) * HEIGHT * 1.0) / maxHealth);
+
+		final int maxPrayer = client.getRealSkillLevel(Skill.PRAYER);
+		final int currentPrayer = client.getBoostedSkillLevel(Skill.PRAYER);
+		final int scaledMissingPrayer = (int) Math.ceil((Math.max(maxPrayer - currentPrayer, 0) * HEIGHT * 1.0) / maxPrayer);
+
+		final BufferedImage hpImage = new BufferedImage(WIDTH, HEIGHT + getYOffset(), BufferedImage.TYPE_4BYTE_ABGR);
```
Adam-: Maybe cache this image?

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
paulbcook: On mac, the way that `SystemTray` sets up the icon pushes it up against the top of the screen instead of centering it as the other icons are. Windows doesn't have this issue so I added a Mac specific offset.

paulbcook:
Seeing a little weirdness with text alignment so I'll need to look at that.

New mac/windows screenshots coming.

runelite-client/src/main/java/net/runelite/client/plugins/statustrayicon/StatusTrayIconPlugin.java
Adam-: This is the type of information which should be in a comment either on the method here or on the field `MAC_HEIGHT_OFFSET`

Adam-:
Can someone on Windows check this looks ok?

paulbcook:
Due to legal requirements at my new employment I’m no longer able to contribute.

If someone still thinks this is still worth adding or improving they can take this over (it looks a lot nicer than the old screenshots now, on Mac at least).

PR #8284, opened 2019-03-22
Master farmer timer

ConradMelcher:
Shows a timer above any master farmer that counts down from 300 seconds (5 minutes) and resets back to 300 whenever the farmer moves. The config allows you to hide the timer until it is less than "x" seconds. The space between the farmer's head and the timer can also be configured, along with the color of the timer.

Very simple; I didn't implement as many features as I wanted to because I feel the current result is what most people will want.
image
image


runelite-client/src/main/java/net/runelite/client/plugins/masterfarmer/MasterFarmerOverlay.java
```diff
@@ -0,0 +1,78 @@
...
+			for (NPC npc : client.getNpcs())
+			{
+				final String npcName = npc.getName();
+
+				if (npcName == null)
+				{
+					continue;
+				}
+
+				if (npc.getId() == 3257 || npc.getId() == 3258)
```
daviscook477: reference the var MASTER_FARMER from NpcID class instead of 3257 as a magic number. same for 3258

runelite-client/src/main/java/net/runelite/client/plugins/masterfarmer/MasterFarmerOverlay.java
```diff
@@ -0,0 +1,78 @@
...
+					}
+				}
+			}
+		}
+		return null;
+	}
+
+	private void renderTimer(final MasterFarmerNPC npc, final Graphics2D graphics)
+	{
+		double timeLeft = 300 - npc.getTimeWithoutMoving();
```
daviscook477: make the 300 into a constant var like static final int FREE_MOVEMENT_TIMER = 300, something that makes it more clear why it's 300 there

runelite-client/src/main/java/net/runelite/client/plugins/masterfarmer/MasterFarmerPlugin.java
```diff
@@ -0,0 +1,124 @@
...
+				else
+				{
+					mf.setTimeWithoutMoving(lastTickUpdate.getEpochSecond() - mf.getStoppedMovingTick().getEpochSecond());
+				}
+				continue;
+			}
+		}
+	}
+
+	private void MasterFarmer(NPC npc)
```
daviscook477: java methods should be camelCase

ConradMelcher: Fixed

runelite-client/src/main/java/net/runelite/client/plugins/masterfarmer/MasterFarmerOverlay.java
ConradMelcher: Fixed

runelite-client/src/main/java/net/runelite/client/plugins/masterfarmer/MasterFarmerOverlay.java
ConradMelcher: Fixed

ShaunDreclin:
Master farmers aren't the only NPCs people trap; maybe this could be changed to a general thieving plugin that works on farmers, ardy knights, guards, etc?

iirc there was a PR for a pickpocket stun cooldown overlay that would go well with it too

PR #7761, opened 2019-02-08
SupplyTrackerV3

Dyldozer:
Automatically keeps track of all consumables, ammo, trident casts, teleport tablets, and teleport scrolls used during the session and allows user to add, edit, or delete the amount.(Additions are right click submenu, Amount edit is pulls up a new gui element)
Works for all consumables, ammo, throwing weapons(including chins), trident casts(seas, swamp, sang, and their (e) versions), teleport scrolls, and teleport tablets.
Works with tick eating.
Built in protection against spam clicking adding too many of one item.
Arrow tracking works even if ammo type is switched or more of the same type is added.

[Main Features](https://i.gyazo.com/bd1d02ac3522de673ce0a373be9f52e0.mp4)
[Longer session test](https://i.gyazo.com/0d2e3a5dce47a7dedc6193f08ab3929f.png)
[Arrow tracking](https://i.gyazo.com/a9ae578f414005268af5fbd264bb2fd7.mp4)
[Throwing weapon Tracking](https://i.gyazo.com/1cd2ac25dd03bb1cf0d19694b6ff37f4.mp4)
[Add items from sub-menu](https://i.gyazo.com/f955f90cada806ff54a0341b1b3febc2.mp4)
[10 trip Zulrah test](https://i.gyazo.com/31492212d278f356502ceb7bf031e0a8.png)

Took a break from Runescape that's why I never finished it last time but I believe I made the changes requested(using ItemID instead of hardcoding ids and generating the menu from an enum) and added the new items and moved the calculations off swing thread.

PR #8286, opened 2019-03-22
Supplies tracker v4?

daviscook477:
taken feedback from https://github.com/runelite/runelite/pull/7761 to improve upon supply tracker. this pr is additional commits on top of commits by @Dyldozer.

see linked pr for images/gifs of supply tracking.

this pr redoes the ui for the supply tracker to match with the loot tracker.
ex: 
suggested_ui

in this ui you can right click on a category or individual supply type to reset the tracker for it.

additionally it implements some additional supply tracking. in addition to the ammo, consumables, teleport tabs/scrolls, and trident casts - this implementation adds runes on spell casts, darts/scales for the blowpipe, and cannonballs. 

rune_usage

blowpipe_usage

track_cannonball_usage_too

since blowpipe does not make usage available in varbits it is estimated and requires user to chose what dart type they use in the config.

chose_dart_choice

some notable limitations of this current implementation:
* runes in the rune pouch are not checked for spell casting supplies
* enchanted jewelry charges are not tracked b/c the chatbox that is used to confirm teleportation is difficult to hook into - a potential workaround is to figure out what rectangle of world points each jewelry can teleport the player into and then use the player entering those world points as an indication of teleportation happening
* there is small bug where sometimes w/ lag tick-eating doesn't go through and only the main food (not the karambwan/potion) will go through. this bug exists in the pr here too https://github.com/runelite/runelite/pull/7761.
* cannonball tracking only works when cannon plugin is enabled - this could be fixed by either duplicating cannon tracking logic or refactoring it out into some CannonManager utility.

looking for feedback on code clarity/sensibility. thank you. I know I need to add unit tests and fix checkstyle at some point - I should be able to handle that


Beagon:
First of all, thanks for taking my suggestions! It looks great, although I've integrated this into my clan client and I found an issue:
- For some reason Runes are sometimes tracked that are "gained". I don't know if that's intended but I sometimes seem to find runes that have a -price (-234) added which gets removed from my total usage.

Also what would be an improvement:
- Charges are not being counted from my Trident of the Swamp/Seas.

I'll update this post if I find anything else that's wrong or might be good to add!

Thanks again!

daviscook477:
Hey @Beagon I have some quick questions about your bug report. Charges from the tridents are supposed to be tracked and was implemented by @Dyldozer. However it tracks the charges by adding the runes/coins to the supply cost rather than a charge. Are you not seeing the runes/coins cost of the trident appearing or are you recommending tracking trident charges in particular rather than the supplies used per charge?

Do you have any examples of activities that would cause runes to be "gained" in the supply tracker? I believe I could mostly fix this by ensuring that any supplies added to the tracker is always positive rather than accidentally adding negative values - but knowing what behavior is causing the rune tracking to be invalid would also be helpful.

Thanks for the follow up. I'll try and fix some of these issues that you've brought up.

daviscook477:
@Beagon here is what I see when using the trident of the swamp:

swamp_trident_usage

If there's a specific scenario in which the casts don't seem to be counted please let me know. I've made the assumption that the trident of the seas/swamp only uses one specific animation but that may be wrong if it uses a different animation in some situation I haven't thought of.

Beagon:
> Hey @Beagon I have some quick questions about your bug report. Charges from the tridents are supposed to be tracked and was implemented by @Dyldozer. However it tracks the charges by adding the runes/coins to the supply cost rather than a charge. Are you not seeing the runes/coins cost of the trident appearing or are you recommending tracking trident charges in particular rather than the supplies used per charge?
> 
> Do you have any examples of activities that would cause runes to be "gained" in the supply tracker? I believe I could mostly fix this by ensuring that any supplies added to the tracker is always positive rather than accidentally adding negative values - but knowing what behavior is causing the rune tracking to be invalid would also be helpful.
> 
> Thanks for the follow up. I'll try and fix some of these issues that you've brought up.

Sorry for the late response!

Both of these bugs/issues happened while I was slaying Black Demons in the Chasm of Fire while also having a cannon out. To be sure that was the scenario I will be updating the Clan Client with the current version and let everyone keep an eye out for these issues or any other issues that may occur.

james-munson:
Added this to https://runelitepl.us/

Nightfirecat:
This has long since been added as a plugin on the hub: https://runelite.net/plugin-hub/show/supplies-tracker

Issue #7968, opened 2019-02-21
[Suggestion] Marked Tiles Option: To appear on the Minimap as well


robthedrunk:

eX-C0n:
can you give a use case for this? you are aware they'd be a single dot, similar to an npc on the minimap

robthedrunk:
My apologizes, for the late reply. The usage could be not only to map out areas for example infernal safe locations on the game screen, but to also have them visible on the minimap to know where to click as well, without having to turn the game camera to see the marked tile locations.

Nightfirecat:
This could use the `[good-for-first-contribution]` tag

Adam-:
Added in #8411

Issue #8243, opened 2019-03-17
Show Ground Markers on the minimap


Andrek25:
Is there a way to do it?

xDemoN:
It's being worked on, #7906

Nightfirecat:
I'm pretty sure that is not what this request is asking for.

Andrek25:
I forget explain that i would like ground markers on the minimap to be easy to click, in this way to be able to move without seeing the ground markers. Thanks!

xDemoN:
My bad, I totally misinterpreted this. Although, it does look like a duplicate of #7968

PR #8288, opened 2019-03-22
Add option to show marked ground tiles on the minimap

Deamon5550:
Adds an option (default off) to overlay marked tiles onto the minimap. 

Closes: #7968  #8243

Example:
image

I did my best to make them rotate properly, let me know if anyone has a better idea for doing this:
rotated

This is my first contribution here so please let me know if I've done anything wrong.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerConfig.java
@@ -54,4 +54,24 @@ public interface GroundMarkerConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "showMinimap",
+		name = "Show on minimap",
+		description = "Shows marked tiles on the minimap"
+	)
+	default boolean showMinimap()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+			keyName = "minimapOpacity",
+			name = "Minimap Overlay Opacity",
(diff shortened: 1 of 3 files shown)


Andrek25:
Can you do that when you put the cursor on the mark grow 50% or shine to know you are going to click there? of course you can disable this feature. Great work!

Matsyir:
> Can you do that when you put the cursor on the mark grow 50% or shine to know you are going to click there? of course you can disable this feature. Great work!

I think growing wouldn't be allowed as it's basically falsifying inputs, because without the grow you might click another tile but it would correct it to your marked tile. It's unfair to anyone not using RL. The shine/light up is a good idea though

Deamon5550:
I made it change the color to white when hovered with the mouse, certainly makes it easier to click on a specific tile now.

runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerConfig.java
```diff
@@ -54,4 +54,24 @@ default boolean rememberTileColors()
...
+	{
+		return false;
+	}
+
+	@ConfigItem(
+			keyName = "minimapOpacity",
+			name = "Minimap Overlay Opacity",
+			description = "The opacity of the minimap markers [0, 255]"
+	)
+	default int minimapOverlayOpacity()
```
Adam-: Is this really required? I would just pick something which looks good and use that.

Adam-:
This was added in #8411

PR #8292, opened 2019-03-22
item prices: Show alch price while alching

Hydrox6:
I always found it weird that you couldn't see the value of the item you were about to alch, so here it is

Also works with the Explorer's Ring interface!


deathbeam:
This should be configurable. Also needs rebase

Hydrox6:
Rebased, added configuration

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -96,6 +97,11 @@ public Dimension render(Graphics2D graphics)
 		// Tooltip action type handling
 		switch (action)
 		{
+			case ITEM_USE_ON_WIDGET:
+				if (!menuEntry.getOption().equals("Cast") || !menuEntry.getTarget().contains("High Level Alchemy") || !config.showWhileAlching())
```
deathbeam: move the boolean check at start of if

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesConfig.java
```diff
@@ -86,4 +86,14 @@ default boolean showAlchProfit()
 		return false;
 	}
 
+	@ConfigItem(
+			keyName = "showWhileAlching",
```
deathbeam: Fix your indentation

Hydrox6:
Rebased and updated to work with #9070

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -95,9 +95,9 @@ public Dimension render(Graphics2D graphics)
 		switch (action)
 		{
 			case ITEM_USE_ON_WIDGET:
-				if (!menuEntry.getTarget().contains("High Level Alchemy") || !config.showAlchProfit())
+				if (!config.showWhileAlching() || !menuEntry.getOption().equals("Cast") || !menuEntry.getTarget().contains("High Level Alchemy"))
 				{
-					break;
+					return null;
```
deathbeam: There is no reason to change this to return null when you can just use break like it is atm on master

PR #2933, opened 2018-05-19
 Slayer Plugin: use varbit for retrieving slayer reward points

leguaan:
Currently slayer plugin uses widgets to retrieve the slayer reward points, this PR makes that process utilise varbits. The updateTick variable method isn't nice, but it's the easiest way for me to delay the update process by just enough to give varbits enough time to initialize, I'm up for any suggestions how make it nicer.

PR #8293, opened 2019-03-22
Slayer Plugin: Retrieve Reward Points via Varbits

xDemoN:
This is a recommit of #2933 

@deathbeam showed interest, but the author abandoned it


Adam-:
Does this work? I wouldn't think the varbit would be always available.

deathbeam:
As long as its used only for getting points from slayer rewards interface then yeah (as its available only there afaik)

xDemoN:
I pulled this from the previous PR, rebased and updated it but whatever I've done doesn't work. I don't have enough knowledge to fix it either so I'm closing it.

PR #8294, opened 2019-03-22
runelite: log client and launcher versions

Hydrox6:
Having the version number in the log allows readers to link it to that client version, instead of having to guess what version of the client it is based on when it was receive.

Launcher option is there for convenience, but that could be in the Launcher itself.


Adam-:
You could place this even before building the injector which would let it show up in the logs if the injector building fails.

Issue #8282, opened 2019-03-21
Chat filter should not filter messages from yourself, your friends and clan members


deathbeam:
I don't think this needs any explanation, it just makes sense.

Hydrox6:
This should be an option, since streamers would love to be able to block everyone from saying certain things.

dmeredith96:
Should something like this always be on, or should there be an option for it? Expanding on the option idea, should users be able to only allow friends or only allow clan members to bypass the filter (e.g. messages from friends are not filtered, but clan member messages are, or vice versa), or should the setting apply to both groups?

PR #8305, opened 2019-03-24
Do not filter messages from yourself, friends, or clan members when using chat filter plugin.

dmeredith96:
Fixes issue #8282. No longer filters messages from yourself, clan members, or friends. 

Not a huge fan of how the unit testing is mocked, maybe I should make a separate method for checking local player? Tests are passing, though.


Hydrox6:
There should be an option to let the filter still affect yourself, friends, and clanmates

MagicfTail:
This won't work for the chat, only the overhead

deathbeam:
Also this works only for overheads and not for chat messages. it should work for both

psikoi:
@Nightfirecat #8775 was merged and references this (and apparently does the same thing), I assume this is a dupe and should be closed?

Issue #5042, opened 2018-08-21
Quest helper


Gamer1120:
The Quest helper would work similar to the Clue Helper, but for quests.

SirDylan:
I've thought of this before, but it seems like a super large project.

Gamer1120:
This is available under the plugin hub now: https://runelite.net/plugin-hub/show/quest-helper

Issue #3749, opened 2018-06-12
NPC Quest Start Location


supermxn:
I was wondering if we could add a plugin it shows where you can start the quest by adding a tile or arrow in the image.
![image](https://user-images.githubusercontent.com/10204414/41263093-f5d3c20c-6da1-11e8-95f6-d9db946cb72f.png)

I believe this will help new players that want to start quests, most of the time the new player have a struggle finding their quest start locations. To activate the where the quest will start, we could hold SHIFT + Click on any quest then the world map will appear showing where it is.

Felanbird:
[Quest Helper](https://runelite.net/plugin-hub/show/quest-helper) was added as a plugin from the hub about 4 years ago

PR #8312, opened 2019-03-25
Add quest helper

Trevor159:
Example of what I currently have in action, this was recorded before I changed a few things and the Veos highlight at the end not working correct seems to be fixed: https://youtu.be/75QmQZKVA8k

Hopefully the structure isn't completely wrong, I tried to keep it small initially for better chance of review.

I tried to make it as simple as possible to add quests, basically how it works is each quests acts as its own "Plugin" as in it can Inject and subscribe to events like a plugin can. Each step can do this also. This was done to keep the plugin class as clean as possible once more steps get added and to increase simplicity when adding more.

To make a new quest all you do is need to make a map of a step for each value the quest var can be. If a quest isn't that simple they can override methods, create new steps etc to get what they want.

Right now there is no UI for starting or ending a quest helper which is something I want to do with an in game widget but adam/abex have been talking about an interface editor which would be nice :d Right now to start a quest helper you just click the quest in the current ingame quest menu. To stop it you restart the plugin.

I am waiting for Hydrox's arrow hint thing for those for no particular reason.

Something I definitely want feedback on is the binder/injector stuff, I am not 100% I did it right but I removed/added stuff and when I have seems to not be excessive and it works.

I suck with words and a lot of this is probably going to be wrong, please bear with me :d

Closes: #3749
Closes: #5042

EDIT: idk if I will ever have time to continue this. Feel free to take over if you want. The ernest the chicken stuff is very WIP and terrible.


Coa7:
Is there anything I can help with? Don't know anything about code but with some proper guidance I can help map out quest maybe?? Would love to see this in action! Just wish it would've came out before I got 250 quest points lol

Nightfirecat:
Generally, probably not. You have to have at least some programmatic understanding of how RuneScape advances quest states.

james-munson:
Would be nice if you added a panel so people could help add quest, like they could add/remove/edit steps

Reasel:
I am interested in helping. Hit me up on Discord: Reasel#0925.

Zoinkwiz:
@Trevor159 Hi, I've been experimenting with a quest helper, and since finding this I've tried basing my core structure on it (https://github.com/Zoinkwiz/runelite/tree/quest-helper). Now that I'm starting to feel like I've got the hang of things, I'd like to work on trying to make this plugin a reality and get all quests done. Just wanted to check with you you're okay with that, or happen to know if there's any other work happening on a quest helper?

Trevor159:
This is now on the plugin-hub

PR #8322, opened 2019-03-27
Fix plugin list layout when a plugin dynamically adds / removes buttons.

romanalexander:
Specifically this fixes the Kourend Library plugin panel button. Plugins that dynamically add/remove buttons to the plugin panel based on game state would cause the layout to break. Behavior isn't observed when enabling/disabling plugins entirely since it looks like the config invalidates all plugin buttons when it saves.

Steps to reproduce:
1. Turn on Kourend Library plugin, and ensure Hide Button config option selected.
2. Enter Kourend Library region
3. Button appears at top-left of plugin panel, rather than correctly listed with rest of plugins.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -224,10 +224,12 @@ public class ClientUI
 			if (inTitle)
 			{
 				titleToolbar.addComponent(event.getButton(), button);
+				titleToolbar.revalidate();
 			}
 			else
 			{
 				pluginToolbar.addComponent(event.getButton(), button);
+				pluginToolbar.revalidate();
 			}
 		});
 	}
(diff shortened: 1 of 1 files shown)


deathbeam:
I am pretty sure this will make startup time way longer

Adam-:
Can you test and see if this does make startup take longer?

romanalexander:
I don't think this code would block the client startup since this runs on the AWT-EventQueue thread.

Did a quick check on a cold JVM with Stopwatch surrounding the code block. From boot to login screen it ran for:

Without patch
`Stopwatch: 7.899 ms`
With patch
`Stopwatch: 8.598 ms`

Adam-:
That sounds somewhat significant.. is there a way we can avoid that?

romanalexander:
Let me clarify that it is a 1 _millisecond_ difference, noting that to the right of the decimal point indicates _microseconds_. This was measured like so:
```
			Stopwatch stopwatch = Stopwatch.createStarted();
			if (inTitle)
			{
				titleToolbar.addComponent(event.getButton(), button);
				titleToolbar.revalidate();
			}
			else
			{
				pluginToolbar.addComponent(event.getButton(), button);
				pluginToolbar.revalidate();
			}
			log.info("Stopwatch: " + stopwatch.toString());
```

Obviously not a conclusive test, but this PR doesn't seem to have a significant effect on boot time.

Adam-:
Ah, okay. @deathbeam you ok with this then?

deathbeam:
Sure. I think its different now because previous it was called when navbar was being rearranged for each nav button added for w/e reaon

PR #8325, opened 2019-03-27
Add !kc herbiboar

jfnilsen:
Reads from server chat messages and stores herbiboar kc, so herbiboar kc can be shared using !kc herbiboar.

This is my first time contributing to an open source project, so be sure to tell me if there are things I should improve.


Adam-:
Hi! See https://github.com/runelite/runelite/wiki/Code-Conventions for instructions on setting up intellij for code formatting.

Also, it would be good for this to have a test added to `ChatCommandsPluginTest.java`

Alexsuperfly:
I feel like this entire pr could just me making the kill count pattern accept `kill|harvest`

jfnilsen:
> I feel like this entire pr could just me making the kill count pattern accept `kill|harvest`

Hey, that's a good idea, I could try that out.

PR #8335, opened 2019-03-29
Add seed vault price calculation to bank plugin

### LINKED ISSUE #8332: Farming Guild Seed Vault Seed Value
I would like to be able to see the G.E. Value of my seeds in the new seed vault. If it's possible be able to see vault value within any bank interface aswell.

TheStonedTurtle:
Title. Doesn't calculate for individual categories as i didn't see an easy way to accomplish this. This config option defaults to on and checks for GE/HA config options, matching the bank title format.

This last commit in this PR also updates the Bank Calculation to leverage the newly created `ContainerCalculation` file.

Closes #8332 

Seeds inside vault:

Updated vault title:


EVOuser94:
Was there a possible way to add this to like the regular bank client, so you can see how much your seed vault value is worth without technically going to the seed vault? Looks great though!

TheStonedTurtle:
We only know the contents when the interface is opened, so it'd either require visiting it every client session or storing it in a config option. Both of those seems like a bad idea for such a small feature imo.

EVOuser94:
Makes sense! Thank you for quick reply and for getting this going regardless!

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -171,4 +181,57 @@ public void onScriptCallbackEvent(ScriptCallbackEvent event)
...
+		if (event.getGroupId() != WidgetID.SEED_VAULT_GROUP_ID || !config.seedVaultValue())
+		{
+			return;
+		}
+
+		handleSeedVaultTotal();
+	}
+
+	@Subscribe
+	public void onItemContainerChanged(ItemContainerChanged event)
```
Adam-: Is there a script to set the vault title like there is for banks?

runelite-client/src/main/java/net/runelite/client/plugins/bank/VaultCalculation.java
```diff
@@ -0,0 +1,85 @@
...
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.Client;
+import net.runelite.api.InventoryID;
+import net.runelite.api.Item;
+import net.runelite.api.ItemContainer;
+import net.runelite.client.game.ItemManager;
+
+@Slf4j
+class VaultCalculation extends BankCalculation
```
Adam-: This should not be vault specific.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
TheStonedTurtle: Per my understanding of a discussion with Polar and Abex this is created using a generic script that builds the layout of interfaces. 

> [5:56 PM] abex: just edit 631:1[1] whenever the inventory changes
[5:57 PM] abex: maybe onWidgetLoaded 631 too,

So i guess no?

runelite-client/src/main/java/net/runelite/client/plugins/bank/VaultCalculation.java
TheStonedTurtle: I'm not sure i understand what you want me to change here, do you want me to just move the `calculate()` inside the plugin file? I'm extending BankCalculation to leverage the `isBankDifferent()` code, i thought this was the simplest way.

TheStonedTurtle:
Merge conflicts have been fixed

ieneoj:
Hello, no rush, but I'm very interested in this plugin and I'm wondering if this will ever find its way into the master branch?

ryanleonard1:
Keen for this, nice work.

runelite-client/src/main/java/net/runelite/client/plugins/bank/ContainerCalculation.java
```diff
@@ -0,0 +1,121 @@
...
+
+class ContainerCalculation
+{
+	private final ItemManager itemManager;
+
+	private int hash;
+	private ContainerPrices containerPrices;
+
+	@Inject
+	ContainerCalculation(ItemManager itemManager)
```
deathbeam: private

runelite-client/src/main/java/net/runelite/client/plugins/bank/ContainerPrices.java
```diff
@@ -0,0 +1,34 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.bank;
+
+import lombok.Data;
+
+@Data
```
deathbeam: This should be Value

runelite-client/src/main/java/net/runelite/client/plugins/bank/ContainerCalculation.java
```diff
@@ -0,0 +1,121 @@
...
+	{
+		this.itemManager = itemManager;
+	}
+
+	@Nullable
+	ContainerPrices calculate(Supplier<Item[]> itemProvider)
```
deathbeam: What is the reason for having supplier here, when you are getting value from it right at start of the method? just pass items directly

Issue #8342, opened 2019-03-29
unable to launch please check log


oschaotic:
unable to launch please check log

i open runelite and it says unable to connect please check logs 
(note i was running osrs on runelite 2 hours prior and have been for 9 months)
this is the log  
[client.log](https://github.com/runelite/runelite/files/3022202/client.log)

Raelenna:
same problem

[client.log](https://github.com/runelite/runelite/files/3022430/client.log)

TheStonedTurtle:
```
[RuneLite] ERROR net.runelite.client.rs.ClientLoader - Error loading RS!
java.net.SocketTimeoutException: timeout
```

You both are having the same connection timed out issue which means you failed to connect to the Old School RuneScape Servers. Since these are hosted by jagex this isn't something we control/fix.

#8035 adds a more descriptive error pop-up, which may be the "fix" for this issue?

oschaotic:
half my guild is having the same issue also :(

oschaotic:
i have never experienced this before im so confused XD

TheStonedTurtle:
The problem is either http://oldschool.runescape.com/jav_config.ws doesn't load for you or the gamepack jar isn't. The gamepack jar url is based off the above config url's `codebase` and `initial_jar` values. 

I.E. http://oldschool55.runescape.com/gamepack_9623388.jar

oschaotic:
im on osb fine :( just runelite not co operating

AZ887:
check your firewall and add client into exception, this solved my problem and i googled mostly is network related issue

Raelenna:
Disabling Bitdefender's (my antivirus) firewall and using only windows firewall fixed it for me.

oschaotic:
turned off firewall completely isssues continuing but can log in but sometimes it will continue accour 4/10 times -no anti virus

PR #8343, opened 2019-03-29
SocketTimeoutException on read

XeonPowder:
#8342


XeonPowder:
Based on the client logs of linked issue, timeout exception occurs on read call of OkHTTP

XeonPowder:
On the issue linked:

Allowing runelite through firewall fixed their issue: #8342 (kinda)... As it was reported that there still is a ~50% chance that it does not work.

Abextm:
If you have >10 seconds of latency to the servers you will to be able to play they game. Regardless I highly doubt their issue is one of latency, but instead their network/isp not correctly routing packets and/or resolving dns

XeonPowder:
Would https://airbrake.io/languages/java-error-monitoring help?

Abextm:
Not in this case. because the issue is not something we can fix

PR #8357, opened 2019-03-30
Add Toggleable Wiki Orb

critzd:
There is now an actual config that can be changed by the user to enable/disable the wiki minimap orb. This does not affect the menu options for Quest Guides/Diary Wiki links. 

**Demo**

Orb is removed then added (not hidden!) based on config.

Changing of the config does not affect any menu items and will still go to wiki, and also will persist even on refresh of the tab.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
@@ -1,0 +1,20 @@
+package net.runelite.client.plugins.wiki;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("wiki")
+public interface WikiConfig extends Config
+{
+	@ConfigItem(
+		position = 1,
+		keyName = "orbEnable",
+		name = "Enable Wiki Minimap Orb",
+		description = "Enable the Wiki Minimap Orb to appear"
+	)
+	default boolean orbEnable()
+	{
+		return true;
+	}
+}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
@@ -30,6 +30,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Provider;
+import com.google.inject.Provides;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.MenuAction;
@@ -38,6 +39,7 @@ import net.runelite.api.NPC;
 import net.runelite.api.NPCComposition;
 import net.runelite.api.ObjectComposition;
 import net.runelite.api.coords.WorldPoint;
+import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.WidgetLoaded;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,44 @@
...
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("wiki")
+public interface WikiConfig extends Config
+{
+	@ConfigItem(
+		position = 1,
```
Abextm: You don't need this

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -366,4 +378,38 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			{
+				createOrb();
+			}
+		}
+	}
+
+	/**
+	 * This will create the minimap orb for the wiki plugin. This does not affect quest guide menu options.
+	 */
+	private void createOrb()
```
Abextm: these two methods should just be part of `add/removeWidgets`

critzd:
I'll be resolving these in a few days, currently out of state.

critzd:
Should be good now

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -148,8 +142,9 @@ private void onWidgetLoaded(WidgetLoaded l)
 
 	private void addWidgets()
 	{
+		spriteManager.addSpriteOverrides(WikiSprite.values());
```
Abextm: This should only be added when the widget is added

critzd: Went ahead and changed it to be only on startup and shutdown. (Add/remove respectively)

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -122,12 +122,14 @@ WikiConfig getConfig(ConfigManager configManager)
 	@Override
 	public void startUp()
 	{
+		spriteManager.addSpriteOverrides(WikiSprite.values());
```
Abextm: This should only be added when the widget is added

critzd: I'm sorry, I guess I don't quite understand. This is the same as the original code, and is only done on startup so I thought it's happening exclusively when it's added?

Abextm: this change is entirely about not adding the widget sometimes, so the sprites should only be added under the same condition, as such the sprites should be removed when the widget is removed

Kamielvf: You added a toggle to not show the widget, and in that case you would not need the sprite overrides. What it should do is add the overrides whenever the widget is toggled to show, and remove whenever it gets hidden.

Kamielvf: Wait you're commenting on an outdated change

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -366,4 +380,20 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+	{
+		if (event.getGroup().equals("wiki"))
+		{
+			if (!config.orbEnable())
+			{
+				clientThread.invokeLater(this::removeWidgets);
+			}
+			else
+			{
+				clientThread.invokeLater(this::addWidgets);
```
abextm: this will duplicate the widget if there are ever any other wiki config options

PR #8372, opened 2019-04-01
Added ability to change which coin value GroundItems color highlighting uses

### LINKED ISSUE #7737: Ground item plugin not respecting thresholds
Having an item on the ground be highlighted when the GE value is 50 and my highlight threshold is set to 250 is pretty annoying. It turns out that's because the high alch price, which I have hidden, exceeds the 250 threshold. I don't really care about high alch prices (that's why I have "Price Display Mode" set to GE only) so it just feels like the color indicators are pretty random, despite the exact values that I've set.

I'm proposing that a brand new setting needs to be added, "Value Calculation":
- Grand Exchange: Use GE price for color determination.
- High Alch: Use HE price for color determination.
- Both: Use whichever is higher.

While thresholds could be determined by the "Price Display Mode" setting, this would mean a default option would have to be chosen for the "Off" mode, since users probably still want highlighting even if they don't show the value.

abelbriggs1:
Fixes #7737.

Right now the color highlighting(low/medium/high/insane) uses whatever's highest of Grand Exchange and High Alchemy to determine an item's color. Not all players care about both, however.

This PR lets users select a specific mode for the color highlighting. Default is BOTH, which is the current functionality.

Example of usage: https://imgur.com/a/PRISDUB

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -179,9 +179,9 @@ public interface GroundItemsConfig extends Config
 		position = 11
 	)
 	default ValueCalculationMode valueCalculationMode()
-    {
-        return ValueCalculationMode.HIGHEST;
-    }
+	{
+		return ValueCalculationMode.HIGHEST;
+	}
 
 	@ConfigItem(
 		keyName = "highlightOverValue2",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -32,6 +32,7 @@ import net.runelite.client.config.ConfigItem;
 import net.runelite.client.plugins.grounditems.config.ItemHighlightMode;
 import net.runelite.client.plugins.grounditems.config.MenuHighlightMode;
 import net.runelite.client.plugins.grounditems.config.PriceDisplayMode;
+import net.runelite.client.plugins.grounditems.config.ValueCalculationMode;
 
 @ConfigGroup("grounditems")
 public interface GroundItemsConfig extends Config
@@ -172,10 +173,18 @@ public interface GroundItemsConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "highlightValueCalculation",
+		name = "Highlight Value Calculation",
+		description = "Configures which coin value is used to determine highlight color",
+		position = 11
+	)
+	default ValueCalculationMode valueCalculationMode() { return ValueCalculationMode.BOTH; }
+
+	@ConfigItem(
 		keyName = "highlightOverValue2",
 		name = "Highlight > Value",
 		description = "Configures highlighted ground items over either GE or HA value",
-		position = 11
+		position = 12
 	)
 	default int getHighlightOverValue()
 	{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -550,11 +552,23 @@ Color getHighlighted(String item, int gePrice, int haPrice)
...
 
+		ValueCalculationMode mode = config.valueCalculationMode();
 		for (Map.Entry<Integer, Color> entry : priceChecks.entrySet())
 		{
-			if (gePrice > entry.getKey() || haPrice > entry.getKey())
+			switch(mode)
 			{
-				return entry.getValue();
+				case GE:
+					if(gePrice > entry.getKey())
```
raiyni: https://github.com/runelite/runelite/wiki/Code-Conventions

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -171,11 +172,19 @@ default MenuHighlightMode menuHighlightMode()
 		return MenuHighlightMode.NAME;
 	}
 
+	@ConfigItem(
+		keyName = "highlightValueCalculation",
+		name = "Highlight Value Calculation",
+		description = "Configures which coin value is used to determine highlight color",
+		position = 11
+	)
+	default ValueCalculationMode valueCalculationMode() { return ValueCalculationMode.BOTH; }
```
Abextm: Add newlines

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -82,6 +82,8 @@
 import static net.runelite.client.plugins.grounditems.config.MenuHighlightMode.BOTH;
 import static net.runelite.client.plugins.grounditems.config.MenuHighlightMode.NAME;
 import static net.runelite.client.plugins.grounditems.config.MenuHighlightMode.OPTION;
+
```
Abextm: remove this newline

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/config/ValueCalculationMode.java
```diff
@@ -0,0 +1,21 @@
+package net.runelite.client.plugins.grounditems.config;
```
Abextm: Add a copyright header

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/config/ValueCalculationMode.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum ValueCalculationMode
+{
+    HA("High Alchemy"), // calc highlight by HA value
+    GE("Grand Exchange"), // calc by GE
+    BOTH("Both"); // use whatever is highest.
```
Abextm: maybe this should be `HIGHEST`?

abelbriggs1: I wasn't sure whether it should be called 'HIGHEST' or 'BOTH', but now that you mention it, I think it will be a bit clearer to the user if it's named 'HIGHEST'.

PR #6967, opened 2018-12-15
Add customizable timestamps

Trevor159:
I wanted a config option for seconds, I thought this was better. Added year, month, day simply because I was already making it customize-able not because I thought they were useful. Someone might want that who knows.

EDIT: replaces #6955

PR #8380, opened 2019-04-02
Add option for 12 hour clock format to chat timestamps

Lucwousin:
Add config for using am/pm in chat timestamps rather than a 24 hour clock.


Trevor159:
#6967 already covers this and more, I guess I need to rebase that though

Lucwousin:
My branch name was better though

PR #8384, opened 2019-04-02
prayer: fix dose indicator not accounting for super restores or sanfews

Hydrox6:
Currently, the Prayer Dose indicator will show when you can drink a prayer potion, even if you're using Super Restores or Sanfews which both restore more prayer. This leads to lost prayer points.


Lucwousin:
This also applies to sanfew serums, wasting 2.

Hydrox6:
Updated to include Sanfews, also now it uses the dominant potion type (as in the updated OP). ty @Lucwousin

PR #8409, opened 2019-04-04
Menu entry swap for buckets

M0dii:
Menu entry swap for buckets addition.


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -513,6 +513,10 @@ else if (config.swapChase() && option.equals("pick-up"))
 		{
 			swap("chase", option, target, true);
 		}
+		else if (config.swapBuckets() && option.equals("use"))
```
Twiglet1022: This is wrong because it will change the left-click to drop for any item with use as the first option, not just buckets. You'd also need to put this check after the custom shift click code or else custom shift click will stop working.

By the way there is already shift click drop for exactly this sort of thing anyway.

M0dii: Oh, how come I missed the whole point of entry swapper. 

I'm currently learning coding, how to use IDE's, Git, etc., sorry for my mistake.
Just wanted to contribute somehow, haha, will try harder next time:P

PR #8420, opened 2019-04-05
adding combat level notifiers to idlenotifier plugin

BrianJMRogers:
I found myself forgetting to pot up while in the nightmare zone or while doing slayer tasks. This addition to the idle notifier lets you receive notifications when you should pot up again


deathbeam:
dont we already have this in boost indicators

BrianJMRogers:
@deathbeam yes but that is pretty inflexible for when you have multiple boosts

Issue #8285, opened 2019-03-22
object markers render outside of draw distance


daviscook477:
reproduction steps:
1. use shift to mark an object so it is highlighted
2. get it outside of the draw distance
3. the highlight will still be drawn even though the object itself won't

example with table in edgeville bank:
![example_outside_highlights](https://user-images.githubusercontent.com/10566967/54799795-0907e200-4c35-11e9-8560-492f4233f7bb.png)

deathbeam:
I dont see a reason why this should be implemented, and also why is it marked as enhancement

genetic-soybean:
![Capture](https://user-images.githubusercontent.com/49249809/55657487-a4cf4b80-57fa-11e9-9179-6d669582a770.jpg)
Looks weird depending on gpu draw distance:
top left - 20 draw distance
bottom left - 70 draw distance
also with gpu plugin disabled you'll get bottom left

daviscook477:
Yeah when I reported it I thought this was a bug given that it didn't behave the same way as the other markers that wouldn't render outside of draw distance - not an enhancement. It really isn't that important but I thought it should at least be documented?

Castro3halo:
Can this be reviewed since Draw distances have been approved and improved over the years, Now we can click on objects that are out of the Object marker highlight distance. Clicking far off objects (that are clickable already) would be easier with this change

PR #8423, opened 2019-04-05
sets max distance for object markers

genetic-soybean:
closes: #8285


genetic-soybean:
alright, made the max draw distance configurable, this should better. Default being the same as the ground marker uses

PR #8021, opened 2019-02-27
Adding filtering to the World Hopper Plugin

RoyceMathews:
Fixes #4970 - Adding filtering to the World Hopper Plugin

* Adds a new button above the header used to sort the worlds. The button expands a panel with checkboxes for selecting filters.
* Checks WorldTypes to determine which worlds to display.
* `WorldHopperPlugin#updateList` is used since I wanted to update the list of worlds without triggering a world refresh that may be throttled.

PR #8426, opened 2019-04-06
add free/member filter to worldhopper plugin

DrRobosnarfen:
SubFilterDemo

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -200,6 +200,7 @@ public class WorldHopperPlugin extends Plugin
 			clientToolbar.addNavigation(navButton);
 		}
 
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -149,6 +149,11 @@
 	private Instant lastFetch;
 	private boolean firstRun;
 
+	public static final int BOTH = 2;
+	public static final int MEMBERS = 1;
+	public static final int FREE = 0;
+	protected int filterMode;
```
Adam-: This should probably just be a `SubscriptionFilterMode` in the panel itself.

Adam-:
There is also #8021 which is relevant to you maybe - but it is highly over-engineered imo.

DrRobosnarfen:
Moved the filtermode check to the panel. And yeah, I want to keep it simple.

Adam-:
Added in 1a73893065ea7eae835c4dfb58275812f9a5258e

PR #8432, opened 2019-04-06
Add theatre of blood npc exp modifiers to special counter plugin

HSJ-OSRS:
The exp modifiers were taken from Bitterkoekje's dps calc and should be accurate


HSJ-OSRS:
these are only for 5 man scaled the xp modifiers change in lower team sizes due to the npc's health getting lowered so an implementation that calculates xp modifier dynamically is needed but would also open up the door to raids 1 npcs also

PR #8434, opened 2019-04-06
Menu Entry swap for Empty Buckets

M0dii:
As title says - Menu Entry swap for Empty Buckets to drop them with single left click.


Dovry:
https://secure.runescape.com/m=news/shift-click-drop--qol?oldschool=1

PR #8439, opened 2019-04-07
idle notifier: allow animations fallback for interacting checks

### LINKED ISSUE #8429: Idle notifier doesn't work with cat
**Describe the bug**
With idle notifier, I don't get notifications when stopped fishing when I'm interrupted by my cat(being needy.) I get notifications normally though

**To Reproduce**
No clue, maybe just try fishing with a cat?

**Expected behavior**

 - OS:Windows
 - RuneLite version: latest

### LINKED ISSUE #9714: Idle Notification not sent when interrupted by interface popup
**Describe the bug**
A system notification is not sent when a character goes idle due to gaining a level.

**_Example:_**
If a fishing spot depletes/inventory fills up - a system notification is received when the character is idle.
However, if a Fishing level is gained - causing the character to idle - a system notification is not sent.

NOTE: I have already enabled _'Send notifications when focused'_ in the _RuneLite_ settings - it is not that.

**To Reproduce**
1. Start performing a non-combat skill of which you are about to obtain a level.
2. Gain the level, and wait as no notifications are sent even though the character ends up idle.

**Expected behavior**
A system notification should be sent on either a level granted, or properly detect idle status.
_--> It would be really cool to be able to toggle a notification upon a level being achieved!_

**Screenshots**
N/A

 - OS: Windows 10, Version 1809 _(OS Build 17763.678)_
 - RuneLite version: 1.5.31.1
 - Launcher version: 1.6.0

**Additional context**
N/A

Alexsuperfly:
adds back the fishing animations to check for animation idle
checks to see if the character has been interacting notified when attempting to send animation notification for the same activity, if so it skips the animation notification.

fixes #8429 
fixes #9714

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
@@ -48,8 +48,10 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import org.mockito.Mock;
+import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -84,6 +86,9 @@ public class IdleNotifierPluginTest
 	private NPC randomEvent;
 
 	@Mock
+	private NPC fishingSpot;
+
+	@Mock
 	private Player player;
 
 	@Before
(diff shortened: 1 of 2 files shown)


Alexsuperfly:
I spoke to the author of #8429 and sent him a compiled version with this in it, he says that he is receiving the notification when previously he would not have.

Twiglet1022:
I tried this out at sacred eels and barehanded sharks and it doesn’t give double notifications. :)

I think if the cat stopped you and later on after you stopped animating the fishing spot moved (so you stopped interacting) it might notify you again but that’s not necessarily a bad thing.

I don’t think there’s any fishing animations that have lots of small breaks in them (like the MLM mining notification) so this won’t result in spamming the user with idle notifications if using a zero ms idle notification delay.

runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
```diff
@@ -136,6 +147,36 @@ public void checkAnimationIdle()
...
+		animationChanged.setActor(player);
+		plugin.onInteractingChanged(new InteractingChanged(player, fishingSpot));
+		plugin.onAnimationChanged(animationChanged);
+		plugin.onGameTick(new GameTick());
+		when(player.getAnimation()).thenReturn(AnimationID.IDLE);
+		when(player.getInteracting()).thenReturn(null);
+		plugin.onAnimationChanged(animationChanged);
+		plugin.onInteractingChanged(new InteractingChanged(player, null));
+		plugin.onGameTick(new GameTick());
+		verify(notifier, atMost(1)).notify("[" + PLAYER_NAME + "] is now idle!");
```
Adam-: Why wouldn't this always notify?

PR #8444, opened 2019-04-08
Options to remove additional bank options

Jbleezy:
Adds the option to remove repeated bank options when the default quantity is changed, and also adds the option to remove Withdraw All but 1.

Remove Additional Bank Options

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -32,7 +32,7 @@ import net.runelite.client.config.ConfigItem;
 public interface MenuEntrySwapperConfig extends Config
 {
 	@ConfigItem(
-		position = -2,
+		position = -4,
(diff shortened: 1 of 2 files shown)


Trevor159:
To my knowledge we currently don't want to edit bank options at all.

Jbleezy:
These options just bring it back to how it was on release

Nightfirecat:
In addition to not wanting to touch bank options, we also never remove menu options and want to keep it that way.

Adam-:
I don't see anything wrong with removing withdraw-but-one or the custom withdraw-x amount. The way this approaches this I think can be simplified though by just checking the menu params. This is what it always used to be & wouldn't be advantageous.

Jbleezy:
Yes these changes just bring it back to how it used to be.

@Adam- It does remove your custom withdraw-x/deposit-x option, instead when the Quantity option on the bank options is changed to 5, 10, X, or All, it removes the extra option that is added. For example, normally with Quantity set to All, your options would become: All, 1, 5, 10, (custom x), X, All. My option changes it to: All, 1, 5, 10, (custom x), X. It removes the additional All menu entry since it is the new 1st option.

PR #8453, opened 2019-04-09
Fix item mapping of Saradomin's Blessed Sword and Saradomin's Tear

Twiglet1022:
When the Saradomin's Blessed Sword is used up or dismantled it returns the Saradomin's Tear, so these two items should both map to Saradomin's Tear. https://oldschool.runescape.wiki/w/Saradomin%27s_blessed_sword

Currently the tear is using the sara sword's GE value and sara's blessed sword has no GE value because of the wrong mapping.


runelite-client/src/main/java/net/runelite/client/game/ItemMapping.java
```diff
@@ -142,7 +142,7 @@
 	// Bounty hunter
 	ITEM_GRANITE_MAUL(GRANITE_MAUL, GRANITE_MAUL_12848),
 	ITEM_MAGIC_SHORTBOW(MAGIC_SHORTBOW, MAGIC_SHORTBOW_I),
-	ITEM_SARADOMINS_SWORD(SARADOMIN_SWORD, SARADOMINS_TEAR),
+	ITEM_SARADOMINS_TEAR(SARADOMINS_TEAR, SARADOMINS_BLESSED_SWORD),
```
deathbeam: This should be still named ITEM_SARADOMINS_BLESSED_SWORD

Twiglet1022: The actual tradeable item is the saradomin's tear, that's why it is like this. This will allow the blessed sword to sue the tear's GE value in the price lookup since it reverts into the tear.

deathbeam: That is not the point of the enum naming. E.g multiple items can map to same base item, so the name of the enum is based on mapped item and not result item.

deathbeam: E.g see <https://github.com/runelite/runelite/pull/8453/files#diff-401c37a2f94c7613ae99a619672a5e76R160>

Twiglet1022: Done

Issue #8431, opened 2019-04-06
Wiki Lookup


M0dii:
Lookup Wiki option (imo) should be implemented into bank and equipment menu.

![image](https://user-images.githubusercontent.com/11699611/55668732-72fbca80-5876-11e9-85fc-08d7670a2299.png)

![image](https://user-images.githubusercontent.com/11699611/55668731-7000da00-5876-11e9-9162-0d9fab74274e.png)

**Edit**
Also, would be great for Quests.

raiyni:
duplicate #7854

PR #8459, opened 2019-04-10
Wiki lookup plugin additions and improvements

syngfaa:
**As requested in issue #8431** 
The request is to make it possible to use the wiki lookup button to cast on items, I am not sure how that would be possible in some interfaces or tabs. I have instead been working on making it possible with adding an option while right clicking on items in some interfaces or tabs. Also, Wiki lookup on quests seems to be already working this way.

**What i did:**
- Change right click options from "Guide, quick guide, and wiki" to "Lookup Guide, Lookup Quick Guide and Lookup Guide to be consistent with casting the wiki lookup text"
- Add wiki lookup functionality to the following tabs and interfaces:
- Items in equipment tab
- Items in bank
- Items in inventory while bank is open
- Items in smithing interface
- Items in ge and search
- Items in ge inventory
- Items in crafting interface
- Items in music tab
- Items in prayer tab
- Items in spellbook tab
- Items in shop
- Items in shop inv

entory
Some example screenshots:
wiki lookup 1
Wiki lookup 4
Wiki lookup 9 ge search
Wiki lookup 7

**Suggestions and help would be appriciated, rather new on the runelite and open source development alltogether.**


syngfaa:
I don't think this is the best solution but because this is the way it already was implemented in the quest, achievement and skilltab i figured why not implement it also in other tabs and interfaces.

The reason i dont think this is the best solution because it is not consistent the way it works to lookup wiki on items/npcs/objects. this way some wiki lookup functionality is partly hidden (I personally did not know there was a right click lookup wiki guide functionality on the quest tab untill i read about it.

Any suggestions/ideas?

PR #8462, opened 2019-04-10
Add WidgetItemOverlay

Adam-:


runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
```diff
@@ -0,0 +1,56 @@
...
+
+public abstract class WidgetItemOverlay extends Overlay
+{
+	private final OverlayRenderer overlayRenderer;
+
+	public WidgetItemOverlay(OverlayRenderer overlayRenderer)
+	{
+		setPosition(OverlayPosition.DYNAMIC);
+		setPriority(OverlayPriority.LOW);
+		setLayer(OverlayLayer.ABOVE_WIDGETS);
```
deathbeam: You should probabky override getLayer instead of doing this so it cannot be changed, as the only point of this is to draw above widgets

runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
```diff
@@ -0,0 +1,56 @@
...
+		this.overlayRenderer = overlayRenderer;
+	}
+
+	public abstract void renderItemOverlay(Graphics2D graphics, int itemId, WidgetItem itemWidget);
+
+	@Override
+	public Dimension render(Graphics2D graphics)
```
deathbeam: Why not just do instanceof check in OverlayRenderer? Like the whole point of that class is to manage and draw overlays, and it would make way more sense because then this class would not need to receive OverlayRenderer in constructor and OverlayRenderer will not need to expose itemWidgets

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -578,4 +584,14 @@ private OverlayBounds buildSnapCorners()
 
 		return entries;
 	}
+
+	public void addWidgetItem(WidgetItem widget)
```
deathbeam: You have the entire list exposed and yet you also have this method

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -578,4 +584,14 @@ private OverlayBounds buildSnapCorners()
...
+	{
+		itemWidgets.add(widget);
+	}
+
+	public void resetWidgetItems()
```
deathbeam: see my previous comment

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -86,6 +89,9 @@
 	private boolean isResizeable;
 	private OverlayBounds snapCorners;
 
+	@Getter
+	private final List<WidgetItem> itemWidgets = new ArrayList<>();
```
deathbeam: also why is this in OverlayRenderer and not manager

runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -1338,10 +1344,30 @@ public static void renderWidgetLayer(Widget[] widgets, int parentId, int var2, i
...
-			widget.setRenderY(y + widget.getRelativeY());
 
-			HashTable<WidgetNode> componentTable = client.getComponentTable();
+			final int renderX = x + widget.getRelativeX();
+			final int renderY = y + widget.getRelativeY();
+			widget.setRenderX(renderX);
+			widget.setRenderY(renderY);
+
+			final int widgetType = widget.getType();
+			if (widgetType == WidgetType.GRAPHIC && widget.getItemId() != -1)
```
Abextm: You probably want to also add the check for if this is the dragged widget `parentID == -1412584499`

PR #8463, opened 2019-04-11
Add npc stats to NPCManager

TheStonedTurtle:
### Depends on runelite/static.runelite.net#92

Updates the NPCManager to use the npc stats scrapped from the osrs wiki. I've also removed the `npc_health.json` file as this data is now redundant

<details>
<summary>Included stats</summary>
<p>

```
hitpoints
combat level
slayer level
attack level
strength level
defence level
range level
magic level
stab bonus
slash bonus
crush bonus
range bonus
magic bonus
stab defence bonus
slash defence bonus
crush defence bonus
range defence bonus
magic defence bonus
bonus attack
bonus strength
bonus range strength
bonus magic damage
poison immune
venom immune
is dragon
is undead
is demon
```

</p>
</details>


deathbeam:
This json file should be hosted on s.r.n

TheStonedTurtle:
Alright i wasn't sure what was preferred since there were json files already in the client

deathbeam:
We put the giant ones there

TheStonedTurtle:
The PR has been updated to pull the stats from s.r.n. I'm assuming this code works since i can't really test it until the s.r.n pr is merged.

Twiglet1022:
I think it would be nice to also have an `isUndead` and `isDragon` based on the weaknesses containing "Salve amulet" and/or "Dragonbane weapons" too.

ShaunDreclin:
If those get added, `isDemon` would also be useful for mobs weak to Silverlight/Darklight/Arclight

Trevor159:
The best way to do that is probably checking if `Demonbane weapons`, `dragonbane weapons`, or `Salve amulet` are in the weaknesses section of the monster.

TheStonedTurtle:
This and the s.r.n pr have been updated to include a dragon, demon, and undead flag.

HSJ-OSRS:
accounting for scaling in raids 1 and 2 and adjusting the stats and xp modifiers accordingly would be nice idk how you would do it tho

TheStonedTurtle:
@HSJ-OSRS Since all this data is being scraped via the wiki and the wiki doesn't handle scaling npcs very well its sorta outside the scope of this PR. There has to be some algorithm the game uses to scale the npcs inside raids and i'd assume implementing the algorithm would be a better solution thanhard coding all possible scaled variations.

The wiki data for these npcs are also formatted weirdly. I.E. the Muttadile has 3 npcs ids, `|id = 7561,7562,7563` but the wiki lists stats for 4 variations. The way the wiki lists these variations is also different than the way it does it for all other npcs, where they put the ids for each variation in a different value. Normally all ids for that variation are stored in the `id#` value where `#` matches the stats for that variation, I.E. `combat1` is the combat level for all `id1` npc ids

PR #8475, opened 2019-04-11
Fix hd healthbars looking fully depleted on low health

Kamielvf:
This fixes an issue with hd healhtbars of the interface styles plugin looking like its fully depleted even though the actor has a slight amount of health left.

Before:
before

After:
after


runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -1524,4 +1525,37 @@ public EnumComposition getEnum(int id)
 		enumCache.put(id, rsEnum);
 		return rsEnum;
 	}
+
+	@Inject
+	private static boolean insideDraw2dExtras;
```
deathbeam: put this above methods

Adam-:
Fixed in #8592

PR #8477, opened 2019-04-11
Switch out addy and rune bars on slayer task overlay for mask variant (like other metal dragons)

nateisonfire:
Not to be merged until cache is updated.

Edit: Cache has been updated (thanks @Abextm )


makingcomment:
What about red and steel dragons?

nateisonfire:
> What about red and steel dragons?

Red dragon's use the mini version of their render (i think the sprite is from construction) and steel dragons have their sprite from construction as well.

makingcomment:
> > What about red and steel dragons?
> 
> Red dragon's use the mini version of their render (i think the sprite is from construction) and steel dragons have their sprite from construction as well.

That’s true. But what I was getting at was: If you’re changing the addy and rune dragon images to the masks, with the goal making things more consistent, shouldn’t steel and red dragons have their images changed as well?

nateisonfire:
@makingcomment Yeah, but I think the consistency we are trying to get at here is, "the overlay should like like the monster you are killing".

What looks closer to the image of a red dragon? a mini red dragon. (not a mask, an actual monster!)
What looks closer to the image of a steel dragon? A steel dragon. (not a mask, the actual monster!)
What looks closer to the image of a rune dragon? A rune dragon mask (not a runite bar)

PR #8503, opened 2019-04-13
Fix chat history wrapping

### LINKED ISSUE #1162: Chat history plugin: Line breaks in private messages aren't persisted
I don't condone the advertising in this screenshot :) but it's what alerted me to this bug.

The first four messages each spilled over to a second line before I hopped worlds. After the plugin restores them, they lose their line breaks and become truncated by the chatbox width.

image

Twiglet1022:
Fixes #1162

I had to edit mixins so I can't actually test one half of this change, but the reason for chat history not wrapping was because of the `nbsp()` method in the chat history plugin. It was a small hack to prevent chat history messages from triggering various plugins that detect chat messages. It did this by replacing spaces with non-breaking-spaces and that's what broke the wrapping.

However I figured that a better way of doing this now is to not send ChatMessage events for any chat message with a timestamp from before you logged in.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Constants.java
+++ b/runelite-api/src/main/java/net/runelite/api/Constants.java
@@ -89,4 +89,9 @@ public class Constants
 	 * the maximum framerate of 50 fps.
 	 */
 	public static final int CLIENT_TICK_LENGTH = 20;
+
+	/**
+	 * The text of the welcome message upon logging in.
+	 */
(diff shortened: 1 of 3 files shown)


Adam-:
I don't think this will work because the timestamps are applied after the event fires.

Twiglet1022:
That would be a problem. Is it possible to change that?

Twiglet1022:
Closing seeing as the idea behind this won't work.

PR #8517, opened 2019-04-14
Group simultaneous boost notifications into one to reduce notification spam

Twiglet1022:
Besides the chatbox spam the multiple notifications are particularly annoying with the way Windows queues up multiple notifications and sends them to you spaced out over the next half a minute or so.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
@@ -103,4 +103,15 @@ public interface BoostsConfig extends Config
 	{
 		return 0;
 	}
+
+	@ConfigItem(
+		keyName = "groupNotifications",
+		name = "Group Notifications",
+		description = "Configures whether or not to group notifications for multiple skills into a single notification",
+		position = 7
+	)
+	default boolean groupNotifications()
+	{
+		return false;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
@@ -100,6 +100,8 @@ public class BoostsPlugin extends Plugin
 	private int lastChangeUp = -1;
 	private boolean preserveBeenActive = false;
 	private long lastTickMillis;
+	private int boostedSkillsChanged = 0;
+	private String lastBoostedSkillChanged;
 
 	@Provides
 	BoostsConfig provideConfig(ConfigManager configManager)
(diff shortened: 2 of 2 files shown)


Nightfirecat:
I like this idea, but I'd prefer it list each skill that is getting low all in one notification. That is, instead of `Multiple skills are getting low!`, it should say `Attack, Defence, and Strength are getting low!`

Twiglet1022:
I did consider that and it would be easy enough to put in, but I'm not sure there's enough benefit considering the overlay already tells you which skills need boosting.

And there will be someone who somehow manages to generate a notification for all 23 skills at once. Of course I could cap it.

Nightfirecat:
It's helpful in that notifications can generate native OS toasts/notifications, and including the skill name(s) there would be useful. And I think the case for someone generating a notification with more than a few skills at once is an edge case not worth considering specially.

Twiglet1022:
Updated it so that it now looks like this:
image

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -233,18 +233,36 @@ public void onGameTick(GameTick event)
...
+				notifier.notify(boostedSkillsChanged.get(0) + " level is getting low!");
 			}
-			else if (boostedSkillsChanged > 1)
+			else
 			{
-				notifier.notify("Multiple skills are getting low!");
-				boostedSkillsChanged = 0;
+				String notification = "";
+				int i = 1;
+				for (final String skill : boostedSkillsChanged)
```
Nightfirecat: This should just be a `for (int i = 0; i < boostedSkillsChanged.size(); i++)` style loop.

Twiglet1022: Updated and retested with the change

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
```diff
@@ -122,4 +122,15 @@ default boolean notifyOnBoost()
...
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "groupNotifications",
+		name = "Group Notifications",
+		description = "Configures whether or not to group notifications for multiple skills into a single notification",
+		position = 8
+	)
+	default boolean groupNotifications()
```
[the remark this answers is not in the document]
Twiglet1022: Fair point! Thanks for the re-review. I've removed the config now and that's simplified the code. Retested with 3 boosts then 1 boost then 4 boosts just to make sure it all still works. Should be good now.

PR #8521, opened 2019-04-14
Swap Eat and Guzzle

MikeZLin:
Swaps the menu entry for eat and guzzle for dwarven rock cakes


deathbeam:
We already have shift-click configuration for this: https://puu.sh/zVEcD/aff42db5c7.gif

PR #8544, opened 2019-04-16
Separated Twitch chat from clan chat 

### LINKED ISSUE #7893: Separate Twitch chat colour from clan chat
Twitch chat, by default, currently uses the same colour as clan chat on fixed or a difficult to read red on transparent. When you change the colour for transparent clan chat, both change.

I'd like it if Twitch chat had it's own colour options to keep it separate and distinguishable from clan chat, while also allowing us to make it readable on transparent chat.

Fixed chat
Transparent chat

smurfyswag:
close #7893


runelite-client/src/main/java/net/runelite/client/plugins/twitch/TwitchPlugin.java
```diff
@@ -26,14 +26,18 @@
 
 import com.google.common.base.Strings;
 import com.google.inject.Provides;
+
+import java.awt.*;
```
xDemoN: oh no

Reasel: There is a way to set it so that you need something like 99 imports before inteliJ will shorten it to wildcards. I highly recommend it.

smurfyswag: Thank you I actually did not know about that, but I will definitely fix it.

Adam-:
So - where does this go if not to the clanchat? Just into `all`?

PR #8558, opened 2019-04-16
UI Customization

bigeggs666:
I was wondering if you guys could add some more UI customization to RuneLite in the form of like different fonts for the side bar plugins (personally always hated the fact I could never make the RS font bold in the UI) and different colored skins for the entire client etc.


Nightfirecat:
This should be opened as an issue, not as a pull request.

bigeggs666:
Sorry, first time using github.

PR #8561, opened 2019-04-17
Remove duplicate coordinate clue

bjornenalfa:
The duplicate entry causes the client to crash when trying to build the ImmutableMap.


Nightfirecat:
My bad on introducing this one. I'll look at writing a test for this since we really ought to be able to catch this at compile time.

PR #8562, opened 2019-04-17
reportbutton: Add Game Ticks option

mlvzk:
Counts game ticks since login and displays them on the Report button

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/ReportButtonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/ReportButtonPlugin.java
@@ -39,6 +39,7 @@ import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetInfo;
 import net.runelite.client.callback.ClientThread;
@@ -60,6 +61,8 @@ public class ReportButtonPlugin extends Plugin
 
 	private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter.ofLocalizedTime(FormatStyle.MEDIUM);
 	private static final DateFormat DATE_FORMAT = new SimpleDateFormat("MMM. dd, yyyy");
+
+	private int tickCounter;
 
 	private Instant loginTime;
 	private boolean ready;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/ReportButtonPlugin.java
```diff
@@ -61,6 +62,8 @@
 	private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter.ofLocalizedTime(FormatStyle.MEDIUM);
 	private static final DateFormat DATE_FORMAT = new SimpleDateFormat("MMM. dd, yyyy");
 
+	private int tickCounter;
```
devLotto: Please put this inbetween ``loginTime`` and ``ready`` and without newlines like so:
```java
private Instant loginTime;
private int tickCounter;
private boolean ready;
```

runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/ReportButtonPlugin.java
```diff
@@ -113,13 +116,21 @@ public void onGameStateChanged(GameStateChanged event)
 			case LOGGED_IN:
 				if (ready)
 				{
+					tickCounter = 0;
```
devLotto: Place this after ``loginTime = Instant.now();``

runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/ReportButtonPlugin.java
devLotto: Maybe it should also be renamed to something along the lines of ``ticksSinceLogin`` to be more clear about what it does.

runelite-client/src/main/java/net/runelite/client/plugins/reportbutton/TimeStyle.java
```diff
@@ -31,7 +31,8 @@
 	LOGIN_TIME("Login Timer"),
 	UTC("UTC Time"),
 	JAGEX("Jagex HQ Time"),
-	LOCAL_TIME("Local Time");
+	LOCAL_TIME("Local Time"),
+	TICK("Game Ticks");
```
devLotto: Change ``TICK`` to ``GAME_TICKS``

mlvzk:
Thank you, I commited the changes. I'm not sure if `updateReportButtonTime()` should be in `onGameTick`, should I at least add an `if` clause that runs it only if `config.time()` is `GAME_TICKS`?

devLotto:
Yeah, ``updateReportButtonTime`` should probably be in onGameTick like you currently have it, so that the tick count is updated when the tick changes. It won't have the same updating behaviour as the other time styles, but those don't require the same accuracy either.

You should probably add the ``config.time() == GAME_TICKS`` check to onGameTick.

mlvzk:
Changed, it's wrapped with `if` now

PR #8563, opened 2019-04-17
Add coordinate clue initialization test

Nightfirecat:
This will catch issues like the one fixed in d11db8e at compile time
rather than at runtime.


Nightfirecat:
Added in 8b5ff851239e793288a379fb037ea540c1a6d5f8 I guess

PR #8569, opened 2019-04-18
Add text labels to ground markers

Reasel:
This will add text labels to ground markers. They use the same color as the tile markers.

demo

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/ColorTileMarker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/ColorTileMarker.java
@@ -37,4 +37,5 @@ class ColorTileMarker
 {
 	private WorldPoint worldPoint;
 	private Color color;
+	private String label;
 }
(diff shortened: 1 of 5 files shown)


Abextm:
What exactly is the use case for this?

ShaunDreclin:
I was actually thinking about this the other day, I wanted to mark which minecart in Keldagrim goes to which destination but a coloured tile would be just as hard to remember as track 1 2 or 3 haha

Reasel:
I was thinking something along the lines of what ShaunDreclin said above. I personally mark things a lot, and although having different colors can be nice to differentiate I would love to be able to mark them.

Things like:

- Favorite safe spots.
- Item respawns
- NPC respawns
- Sometimes marking multi lines so I don't cross them.
- Lately I have used markers for zulrah and being able to label them by number instead of having to remember or color code them...


Its obviously something not everyone will use or want, as is apparent from the discussion on the discord, but I wanted it and made it in the case others agree.

deathbeam:
I believe I already closed issue that was asking for this because just coloring is enough, this is mostly just mess, so I dont see a point in implementing it (it just adds more clutter and its hardly readable)

Reasel:
Fair enough I guess.

PR #8580, opened 2019-04-19
stop overlay methods from triggering off of hidden items

Hydrox6:
Before:
before
After:
after


runelite-mixins/src/main/java/net/runelite/mixins/RSClientMixin.java
```diff
@@ -1353,7 +1353,7 @@ public static void renderWidgetLayer(Widget[] widgets, int parentId, int minX, i
 			final int widgetType = widget.getType();
 			if (widgetType == WidgetType.GRAPHIC && widget.getItemId() != -1)
 			{
-				if (renderX >= minX && renderX <= maxX && renderY >= minY && renderY <= maxY)
+				if (!widget.isSelfHidden() && renderX >= minX && renderX <= maxX && renderY >= minY && renderY <= maxY)
```
Adam-: This should be up on 1338?

PR #8583, opened 2019-04-19
Add daily notification for collection of ogre arrows from Rantz.

### LINKED ISSUE #8581: Daily Task Indicator - Ogre Arrows
Another useful feature to add to the daily task indicator options is picking up Ogre Arrows from Rantz if one has the Western Provinces diaries completed, since this is worth as much if not more than the other daily item pickups awarded by diaries!

Reasel:
Adds a notification for ogre arrow collection from Rantz with completion of at least easy western provinces diary.

Closes #8581

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksConfig.java
@@ -120,4 +120,15 @@ public interface DailyTasksConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		position = 9,
+		keyName = "showArrows",
+		name = "Show Claimable Ogre Arrows",
+		description = "Show a message when you can collect ogre arrows from Rantz."
+	)
+	default boolean showArrows()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksPlugin.java
```diff
@@ -247,6 +253,21 @@ else if (dailyReset)
 		}
 	}
 
+	private void checkArrows(boolean dailyReset)
+	{
+		if (client.getVar(Varbits.DIARY_WESTERN_EASY) == 1)
+		{
+			if (client.getVar(Varbits.DAILY_ARROWS_STATE) == 0)
```
Nightfirecat: `if (dailyReset || client.getVar(Varbits.DAILY_ARROWS_STATE) == 0)`

Both checks have the same body, so there's no reason to have them in separate if branches.

Reasel: I simply matched the file format that was already there. Should I fix the other functions as well? Only in places where this is also true?

Reasel:
Need to wait to check if the arrows can be claimed without a re-log.

runelite-client/src/main/java/net/runelite/client/plugins/dailytaskindicators/DailyTasksConfig.java
```diff
@@ -109,4 +109,15 @@ default boolean showBonemeal()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 8,
```
deathbeam: Fix your indetnation

Reasel:
The collection is similar to the battle staff collection in that it requires a relog in order to claim. I have added the necessary logic to show the play this. I think this is good to go.

xDemoN:
@Reasel Needs rebased

deathbeam:
needs rebase

deathbeam:
Also, @Adam- said he wants this off by default.

We already have a lot of daily notifs on by default, and this is not useful, so default off makes sense.

PR #8595, opened 2019-04-20
Bankvalue: calculate master scroll book value

### LINKED ISSUE #8590: master scroll book scroll value
I have to remove each scroll from the Master scroll book to check the true value of the book.

On examine, show total value of the Master scroll book including the book itself. Opening the scroll book once would be required to get accurate values and then kept in memory and updated when scrolls are added/lost from the book itself.

This data could also be added to the bank value itself.

Show the value when the book is opened inside the book itself similar to how the bank tab system works.

**Additional context**
-

SebastiaanVanspauwen:
Calculates the value of all the teleport scrolls in a master scroll book (if the item is in the bank) and adds them to bank value.

closes #8590


runelite-client/src/main/java/net/runelite/client/plugins/bank/BankCalculation.java
```diff
@@ -59,7 +60,26 @@
 		Varbits.BANK_TAB_EIGHT_COUNT,
 		Varbits.BANK_TAB_NINE_COUNT
 	);
-
+	private static final Map<Varbits, Integer> VARBIT_MASTERBOOK_ITEMIDMAP = new HashMap<Varbits, Integer>()
```
deathbeam: ImmutableMap.builder

Nightfirecat:
Are these varbits set on login? If not, we likely can't rely on their values.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankCalculation.java
```diff
@@ -177,6 +177,12 @@ void calculate()
 					continue;
 				}
 
+				//Limit quantity to 1 because master scroll books share teleport scrolls
+				if (itemId == ItemID.MASTER_SCROLL_BOOK)
```
deathbeam: This is still incorrect because it needs to count master scroll book x times, but teleports inside only once

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -281,6 +299,18 @@ public int getItemPrice(int itemID)
...
+		{
+			int totalPrice = 0;
+
+			for (Map.Entry<Varbits, Integer> entry : VARBIT_MASTERBOOK_ITEMIDMAP.entrySet())
+			{
+				final int amount = client.getVar(entry.getKey());
+				totalPrice += getItemPrice(entry.getValue()) * amount;
+			}
+
+			return totalPrice;
```
TheStonedTurtle: Shouldn't you also add the price of the scroll book here?

deathbeam: Yes, but not in this method as that would make getting correct price for multiple scroll books impossible

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -331,6 +319,25 @@ public int getItemPrice(int itemID)
...
+			price = totalEmptyBookPrice;
+
+			for (Map.Entry<Varbits, Integer> entry : VARBIT_MASTERBOOK_ITEMIDMAP.entrySet())
+			{
+				final int amount = client.getVar(entry.getKey());
+				price += getItemPrice(entry.getValue()) * amount;
+			}
+		}
+
+		return price;
```
deathbeam: so this works only for master scroll book? you should also make it work for normal items.

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -331,6 +319,25 @@ public int getItemPrice(int itemID)
 		return price;
 	}
 
+	public int getItemPrice(int itemID, int quantity)
+	{
+		int price = 0;
+
+		if (itemID == ItemID.MASTER_SCROLL_BOOK)
+		{
+			final int totalEmptyBookPrice = getItemPrice(MASTER_SCROLL_BOOK_EMPTY) * quantity;
```
deathbeam: just use itemId and not master_scrollbook_empty

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankCalculation.java
```diff
@@ -177,7 +177,15 @@ void calculate()
 					continue;
 				}
 
-				gePrice += (long) itemManager.getItemPrice(itemId) * quantity;
+				//If item is MASTER_SCROLL_BOOK get price for one filled book and the empty books
+				if (itemId == ItemID.MASTER_SCROLL_BOOK)
```
deathbeam: just use getItemPrice(item, quantity) ONLY

PR #8600, opened 2019-04-21
Add plugin displaying smelting info bar.

### LINKED ISSUE #8418: Smelting Infobox
I have been cooking and fishing and I noticed that there is an infobox like this:
https://runelite.net/img/features/fishing.png

It displays the Action Status (Fishing (Green) / NOT FISHING (Red); Cooking / NOT COOKING), and the count per hour.

I am suggesting the same InfoBox model for Smelting.
I noticed this infobox is absent (which might be a bug or simply wasn't developed) while smelting cannonballs out of steel bars, for example. 
For me it's particularly useful to see the bright red InfoBox saying "Not Fishing"  and the item unit count/unit count per hour so I can do my own math.

Sample Code from other plugins that may be useful for this:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java

Line 67

stepzhu:
Closes #8418 

This new plugin keeps track of the amount of bars/cannonballs smelted for the current session. 

With the plugin, it will look something like this:
Screen Shot 2019-04-20 at 9 05 07 PM

If no cannon balls were smelted during the session, it will only display bars smelted and vice versa. Likewise, if the player smelted both cannon balls and bars, then it will display both.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingOverlay.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2019, Stephen <stepzhu@umich.edu>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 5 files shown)


Adam-:
Aside from some minor codestyle violations this seems fine to me.

runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingOverlay.java
```diff
@@ -0,0 +1,136 @@
...
+		this.config = config;
+		this.xpTrackerService = xpTrackerService;
+		getMenuEntries().add(new OverlayMenuEntry(RUNELITE_OVERLAY_CONFIG, OPTION_CONFIGURE, "Smelting overlay"));
+	}
+
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		SmeltingSession session = plugin.getSession();
+		if (session.getLastItemSmelted() == null)
```
Adam-: This will NPE while the plugin is running

PR #8611, opened 2019-04-22
Fix chat history wrapping

### LINKED ISSUE #1162: Chat history plugin: Line breaks in private messages aren't persisted

Twiglet1022:
Fixes #1162

The reason that chat history doesn't wrap is because spaces are replaced by non-breaking spaces in order to prevent various plugins that detect chat messages from triggering incorrectly when chat history is replayed.

I've found a way to do this while still preserving wrapping by utilising the fact that "—" does not show up in-game.

<details><summary>Expand for before and after GIFs</summary>

Before
Before chat history wrapping fix

After
After chat history wrapping fix

</details>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chathistory/ChatHistoryPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chathistory/ChatHistoryPlugin.java
@@ -145,8 +145,8 @@ public class ChatHistoryPlugin extends Plugin implements KeyListener
 					.type(chatMessage.getType())
 					.name(chatMessage.getName())
 					.sender(chatMessage.getSender())
-					.value(nbsp(chatMessage.getMessage()))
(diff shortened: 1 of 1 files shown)


Adam-:
I would rather see this fixed more correctly than this, but I'm not sure how.

Twiglet1022:
I agree, though given that what it’s doing now is already a cheap trick it’s not really any worse is it?

Is there no hope for my previous idea of not firing the ChatMessage event for chat history messages? If such a thing were possible it would even solve the problem of chat history triggering on single words like !total and custom chat notifications.

Adam-:
That could be possible yes. I'd have to think about whether or not that fixes the problem.

Adam-:
fixed in e71b18b02b124bff69cc347238551d2e0d3bf29b

PR #8625, opened 2019-04-23
Fix high risk world type name, remove from PVP world types

### LINKED ISSUE #8605: Incorrect use of world masks
High risk standard world (W365) is incorrectly being defined as a PVP world.
The current masks for this world are: [MEMBERS, PVP_HIGH_RISK]
Whereas the masks for any PVP high risk world are:[MEMBERS, PVP, PVP_HIGH_RISK]

As current PVP high risk worlds are being defined as PVP worlds twice. 
would be ideal to change the PVP_HIGH_RISK mask to HIGH_RISK and remove the HIGH_RISK mask from PVP_WORLD_TYPES enum.
The world snapshot needs to be updated for this to work.

**Expected behavior**
High risk standard world (W365) masks should be [MEMBERS, HIGH_RISK]
and PVP high risk world mask should be [MEMBERS, PVP, HIGH_RISK]

### LINKED ISSUE #8624: Wilderness level attack range does not work on high-risk worlds
1. Log in to world 365 (the only non-pvp high-risk world)
2. Enter the wilderness
3. See error

Nightfirecat:
This world type is not specifically a PVP world type; at the time of
this commit, world 365 is a non-PVP high-risk world, where the world is
the same as a regular members' world, except that the Protect Item
prayer is disabled.

Note: this will not affect the world hopper plugin, as it both
explicitly references the high risk world type, and references the
http-api WorldType implementation, not the runelite-api WorldType which
is being modified in this commit.

Fixes #8605
Fixes #8624


Adam-:
I think this breaks the "next" hop keyboard shortcut when hopping in the wilderness.

xdesr:
https://github.com/xdesr/runelite/commit/aa736cc8998da671a498d7ad3cc08ac6ef490ca0
I imagine that would not break the next shortcut but the js at: https://api.runelite.net/runelite-1.5.21-SNAPSHOT/worlds.js needs to be updated. i'll read my own json file and do some testing.

Adam-:
the http-service worlds endpoint is basically mirroring the Jagex one. In it, high risk worlds are a thing so it shouldn't be removed from ours. It is not relevant to whether or not the world hopper respects it any.

Nightfirecat:
I'm thoroughly confused about what you mean by removing it from the http api. It's not that it needs to be removed, it's that it is poorly named and should be renamed.

Separately, it should be removed from `api.WorldType#isPvpWorld` (which is not used by the world hopper plugin), because high risk worlds are not PVP worlds.

Nightfirecat:
I can confirm by testing that the world hopper plugin on 98984c9 behaves the same as release.

Adam-:
Often pkers will use the quick hop up/down feature of the world hopper to cycle through worlds, we probably should not let them hop into high risk worlds like that.

PR #8641, opened 2019-04-25
Fix Mutated Zygomite fungicide weaknes threshold

deathbeam:
Based on Wiki its 0-7


deathbeam:
https://oldschool.runescape.wiki/w/Fungicide_spray

PR #8645, opened 2019-04-25
Bank: add runepouch contents to bank value

xdesr:
Runepouch contents added to bank calculations for GE/HA


deathbeam:
https://github.com/runelite/runelite/pull/8595 basically exactly same issue here

Adam-:
I don't really like cross-plugin dependencies like this. What is the issue you are referring to?

deathbeam:
The issue is that pricing for this should be implemented in ItemManager, so it will work also with e.g examine plugin etc

xdesr:
I have moved the pricing into ItemManager although the class depends on an enum from the runepouch plugin.
I assume that you are referring to ItemManager.getItemPrice() to add a quantity parameter and changing the return method to long.
I would like to additionally add an alchprice boolean parameter as currently the alchprice calculation is done inside the plugins.
An issue i have found with this is that a lot of items that can't be alched in game such as crystal shield, arclight etc... are currently displaying an alch price.
The wiki has a list of ~700 items but i'm not 100% sure how accurate this is.
Ideally a full list of all items non-alchable would fix this although i'm not sure the best method of going about this.

deathbeam:
Yeah, new method that accepts quantiy. But, I would start with implementing just that and adding runepouch price to it. Also, if it depends on something from RunePouch plugin then that thing should be moved outside of it. Alch price is different issue.

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -269,10 +280,12 @@ public void invalidateItemCompositionCache()
 	 * Look up an item's price
 	 *
 	 * @param itemID item id
+	 * @param quantity quantity
 	 * @return item price
 	 */
-	public int getItemPrice(int itemID)
+	public long getItemPrice(int itemID, int quantity)
```
Nightfirecat: Add an overload for this which calls this signature with a quantity of 1.

PR #8659, opened 2019-04-26
Added agility level to agility shortcuts

PeterMcteague:
With this commit, the agility plugin now shows the agility level required to pass shortcuts on hover/right-click/tooltip, with red/green coloring dependent on the players agility level.

This is useful because it means players do not have to refer to the map, skill guide or the wiki to see this information.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
@@ -245,4 +245,9 @@ public enum AgilityShortcut
 	{
 		return description + " - Level " + level;
 	}
+
+	public int getLevel()
+	{
+		return level;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -59,6 +59,7 @@ import net.runelite.api.events.GroundObjectDespawned;
 import net.runelite.api.events.GroundObjectSpawned;
 import net.runelite.api.events.ItemDespawned;
 import net.runelite.api.events.ItemSpawned;
+import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.WallObjectChanged;
 import net.runelite.api.events.WallObjectDespawned;
 import net.runelite.api.events.WallObjectSpawned;
@@ -71,6 +72,7 @@ import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.OverlayManager;
 import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
+import net.runelite.api.MenuEntry;
 
 @PluginDescriptor(
 	name = "Agility",
@@ -421,4 +423,39 @@ public class AgilityPlugin extends Plugin
 			}
 		}
 	}
+
+	@Subscribe
+	public void onMenuEntryAdded(MenuEntryAdded event)
+	{
+		if (config.showShortcutLevel())
+		{
+			MenuEntry[] entries = client.getMenuEntries();
+
+			//Skip anything where there's 0 chance it's a agi shortcut.
+			if (event.getTarget() == null || event.getTarget() == ""){
+				return;
+			}
+
+			Map<TileObject,Obstacle> obstacles = getObstacles();
+			for (Obstacle o : obstacles.values())
+			{
+				//Compare the event ID to nearby obstacle ID's
+				for (int obId: o.getShortcut().getObstacleIds()){
+					if (obId == event.getIdentifier()){
+						AgilityShortcut agilityShortcut = o.getShortcut();
+						for (MenuEntry entry : entries) {
+							//Check it's the non-examine
+							if (entry.getTarget().equalsIgnoreCase(event.getTarget()) && !entry.getOption().equalsIgnoreCase("examine") && !entry.getTarget().toLowerCase().contains("(level-")) {
+								//Append level with a color to the added entry and update it!
+								if (agilityShortcut.getLevel() <= getAgilityLevel()) entry.setTarget(entry.getTarget() + " <col=00ff00>(level-" + agilityShortcut.getLevel() + ")");
+								else entry.setTarget(entry.getTarget() + " <col=ff0000>(level-" + agilityShortcut.getLevel() + ")");
+								client.setMenuEntries(entries);
+								return;
+							}
+						}
+					}
+				}
+			}
+		}
+    }
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
@@ -99,6 +99,17 @@ public interface AgilityConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "addLevelsToShortcutOptions",
+		name = "Show Shortcut Requirement",
+		description = "Enable/disable showing agility requirements in shortcut right-click options",
+		position = 6
+	)
+	default boolean showShortcutLevel()
+	{
+		return true;
+	}
+
+	@ConfigItem(
 		keyName = "trapOverlay",
 		name = "Show Trap Overlay",
 		description = "Enable/disable the highlighting of traps on Agility courses",


runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
```diff
@@ -245,4 +245,9 @@ public String getTooltip()
 	{
 		return description + " - Level " + level;
 	}
+
+	public int getLevel()
```
Nightfirecat: Why was this added? This is already created via Lombok.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
 			}
 		}
 	}
+
+	@Subscribe
+	public void onMenuEntryAdded(MenuEntryAdded event)
+	{
+		if (config.showShortcutLevel())
```
Nightfirecat: This should be a guard clause.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+			//Skip anything where there's 0 chance it's a agi shortcut.
+			if (event.getTarget() == null || event.getTarget() == ""){
+				return;
+			}
+
+			Map<TileObject,Obstacle> obstacles = getObstacles();
+			for (Obstacle o : obstacles.values())
+			{
+				//Compare the event ID to nearby obstacle ID's
+				for (int obId: o.getShortcut().getObstacleIds()){
```
Nightfirecat: It could be worth refactoring AgilityShortcut to store `obstacleIds` as a `List` instead of an array so that this can be changed from `for (id : getObstacleIds()) { if (id == e.getIdentifier()) { /* ... */ } }` to become `if (o.getShortcut().getObstacleIds().contains(event.getIdentifier())`. This would have the added benefit of allowing this to become another guard clause instead of another giant nested block.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+					if (obId == event.getIdentifier()){
+						AgilityShortcut agilityShortcut = o.getShortcut();
+						for (MenuEntry entry : entries) {
+							//Check it's the non-examine
+							if (entry.getTarget().equalsIgnoreCase(event.getTarget()) && !entry.getOption().equalsIgnoreCase("examine") && !entry.getTarget().toLowerCase().contains("(level-")) {
+								//Append level with a color to the added entry and update it!
+								if (agilityShortcut.getLevel() <= getAgilityLevel()) entry.setTarget(entry.getTarget() + " <col=00ff00>(level-" + agilityShortcut.getLevel() + ")");
```
Nightfirecat: Don't use oneline `if`s.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+			for (Obstacle o : obstacles.values())
+			{
+				//Compare the event ID to nearby obstacle ID's
+				for (int obId: o.getShortcut().getObstacleIds()){
+					if (obId == event.getIdentifier()){
+						AgilityShortcut agilityShortcut = o.getShortcut();
+						for (MenuEntry entry : entries) {
+							//Check it's the non-examine
+							if (entry.getTarget().equalsIgnoreCase(event.getTarget()) && !entry.getOption().equalsIgnoreCase("examine") && !entry.getTarget().toLowerCase().contains("(level-")) {
+								//Append level with a color to the added entry and update it!
```
Nightfirecat: This isn't a useful comment.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+
+			Map<TileObject,Obstacle> obstacles = getObstacles();
+			for (Obstacle o : obstacles.values())
+			{
+				//Compare the event ID to nearby obstacle ID's
+				for (int obId: o.getShortcut().getObstacleIds()){
+					if (obId == event.getIdentifier()){
+						AgilityShortcut agilityShortcut = o.getShortcut();
+						for (MenuEntry entry : entries) {
+							//Check it's the non-examine
```
Nightfirecat: I believe you can use `MenuAction.EXAMINE` to compare against. Also, this comment isn't correct since it doesn't explain the comparison of the entry target to the event target nor the check for level text. Generally, comments which are explaining what the code is doing as opposed to why it's being done should be removed. Much of this code can be made more clear by being broken into smaller statements, using better variable naming, etc.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+
+			//Skip anything where there's 0 chance it's a agi shortcut.
+			if (event.getTarget() == null || event.getTarget() == ""){
+				return;
+			}
+
+			Map<TileObject,Obstacle> obstacles = getObstacles();
+			for (Obstacle o : obstacles.values())
+			{
+				//Compare the event ID to nearby obstacle ID's
```
Nightfirecat: This comment is also not helpful.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+		if (config.showShortcutLevel())
+		{
+			MenuEntry[] entries = client.getMenuEntries();
+
+			//Skip anything where there's 0 chance it's a agi shortcut.
+			if (event.getTarget() == null || event.getTarget() == ""){
+				return;
+			}
+
+			Map<TileObject,Obstacle> obstacles = getObstacles();
```
Nightfirecat: This is only used once, so does not need to be stored in a variable.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +423,39 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+				//Compare the event ID to nearby obstacle ID's
+				for (int obId: o.getShortcut().getObstacleIds()){
+					if (obId == event.getIdentifier()){
+						AgilityShortcut agilityShortcut = o.getShortcut();
+						for (MenuEntry entry : entries) {
+							//Check it's the non-examine
+							if (entry.getTarget().equalsIgnoreCase(event.getTarget()) && !entry.getOption().equalsIgnoreCase("examine") && !entry.getTarget().toLowerCase().contains("(level-")) {
+								//Append level with a color to the added entry and update it!
+								if (agilityShortcut.getLevel() <= getAgilityLevel()) entry.setTarget(entry.getTarget() + " <col=00ff00>(level-" + agilityShortcut.getLevel() + ")");
+								else entry.setTarget(entry.getTarget() + " <col=ff0000>(level-" + agilityShortcut.getLevel() + ")");
```
Nightfirecat: Don't use oneline `else`s.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
Nightfirecat: Use `ColorUtil` for the color string building. Also, append the `"(level-" + agilityShortcut.getLevel() + ')'` text outside of this if-else block; it is the same regardless of whether the condition is true.

runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
PeterMcteague: Whoops, I must have missed that.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
PeterMcteague: I've removed it.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
PeterMcteague: Ah I hadn't noticed those. I've added a guard case for some which I imagine are more common.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -25,6 +25,8 @@
 package net.runelite.client.plugins.agility;
 
 import com.google.inject.Provides;
+
```
Nightfirecat: Remove this newline
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -25,6 +25,8 @@
...
+import java.awt.*;
```
Nightfirecat: Don't use star imports: https://github.com/runelite/runelite/wiki/Code-Conventions

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +427,56 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+		{
+			return;
+		}
+
+		if (event.getTarget() == null || event.getTarget() == "")
+		{
+			return;
+		}
+
+		if (event.getType() == MenuAction.EXAMINE_OBJECT.getId()
```
Nightfirecat: Is there a specific menu action type that's used for agility obstacles? `GAME_OBJECT_FIRST_OPTION` probably? It might be better to guard against that than to skip these ones specifically.

eg.

```java
if (event.getType() != MenuAction.GAME_OBJECT_FIRST_OPTION.getId()
    && event.getType() != MenuAction.GAME_OBJECT_SECOND_OPTION.getId())
    // continue game object menu action checks here as needed
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,54 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+		for (Obstacle nearbyObstacle : getObstacles().values())
+		{
+			for (int obId: nearbyObstacle.getShortcut().getObstacleIds())
+			{
+				if (obId == event.getIdentifier())
+				{
+					AgilityShortcut agilityShortcut = nearbyObstacle.getShortcut();
+					for (MenuEntry entry : menuEntries)
+					{
+						if (entry.getTarget().equalsIgnoreCase(event.getTarget()) && !entry.getTarget().toLowerCase().contains("(level-"))
```
beaumitch: You can use the fact that "event" will always refer to the last menu option as the menu is built to avoid this for loop & unreliable string comparison.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,54 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+
+		//Guarding against non-first option because agility shortcuts are always that type of event.
+		if (event.getType() != MenuAction.GAME_OBJECT_FIRST_OPTION.getId())
+		{
+			return;
+		}
+
+		MenuEntry[] menuEntries = client.getMenuEntries();
+
+		for (Obstacle nearbyObstacle : getObstacles().values())
```
beaumitch: We can greatly reduce nesting & complexity here with streams and functional programming.

Untested but this logic could look something like this:
```
final int entryId = event.getIdentifier();
MenuEntry[] menuEntries = client.getMenuEntries();
for (Obstacle obstacle : getObstacles().values())
{
	AgilityShortcut shortcut = obstacle.getShortcut();
	if (shortcut.getObstaclesIds().stream().anyMatch(obstacleId -> entryId == obstacleId))
	{
		int level = shortcut.getLevel();
		String color = level <= getAgilityLevel() ? Color.GREEN : Color.RED;
		String requirementText = " (level-" + level + ")";
		MenuEntry entry = menuEntries[menuEntries.length - 1];
		entry.setTarget(event.getTarget() + ColorUtil.prependColorTag(requirementText, color));
	}
}
client.setMenuEntries(menuEntries);
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,54 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+
+	@Subscribe
+	public void onMenuEntryAdded(MenuEntryAdded event)
+	{
+		if (!config.showShortcutLevel())
+		{
+			return;
+		}
+
+		if (event.getTarget() == null || event.getTarget() == "")
```
beaumitch: This isn't necessary.

PeterMcteague: Great suggestions, thanks for your input

raiyni: why did you mark this resolved

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,41 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+		final int entryId = event.getIdentifier();
+		MenuEntry[] menuEntries = client.getMenuEntries();
+
+		for (Obstacle nearbyObstacle : getObstacles().values())
+		{
+			AgilityShortcut shortcut = nearbyObstacle.getShortcut();
+			if (Arrays.stream(shortcut.getObstacleIds()).anyMatch(i -> i == entryId))
+			{
+				MenuEntry entry = menuEntries[menuEntries.length - 1];
+				if (!entry.getTarget().toLowerCase().contains("(level-"))
```
raiyni: what is this check for

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,41 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+			return;
+		}
+
+		final int entryId = event.getIdentifier();
+		MenuEntry[] menuEntries = client.getMenuEntries();
+
+		for (Obstacle nearbyObstacle : getObstacles().values())
+		{
+			AgilityShortcut shortcut = nearbyObstacle.getShortcut();
+			if (Arrays.stream(shortcut.getObstacleIds()).anyMatch(i -> i == entryId))
```
beaumitch: Apparently getObstacleIds can return null resulting in an NPE here, it's definitely worth a little refactor of AgilityShortcut to use ImmutableList or ImmutableSet instead of an array to take care of multiple issues.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
beaumitch: ^ agree doesn't seem to be necessary unless there's an edge case we aren't considering

PeterMcteague: True, the "level-" check should never actually crop up. I think that was a check left over from when I was doing things in a much different way that I neglected to remove.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -421,4 +426,38 @@ private void onTileObject(Tile tile, TileObject oldObject, TileObject newObject)
...
+
+	@Subscribe
+	public void onMenuEntryAdded(MenuEntryAdded event)
+	{
+		if (!config.showShortcutLevel())
+		{
+			return;
+		}
+
+		//Guarding against non-first option because agility shortcuts are always that type of event.
```
Nightfirecat: This is mostly self-evident, so I don't think this comment is needed.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
PeterMcteague: Currently, I believe it to effectively be like this except shortcut is stored in a variable so I can get it's level requirement.

```
for (Obstacle nearbyObstacle : getObstacles().values())
		{
			AgilityShortcut shortcut = nearbyObstacle.getShortcut();
			if (Arrays.stream(shortcut.getObstacleIds()).anyMatch(i -> i == entryId))
```

Nightfirecat: Streams are pretty inefficient to initialize, but since this is only run once per menu opening at most, I think that'll be fine.

Adam-:
Is this useful? We already have a thing to change the overlay color based on whether or not you can use the shortcut.. Maybe we could make it show you when you examine it.

PeterMcteague:
> Is this useful? We already have a thing to change the overlay color based on whether or not you can use the shortcut.. Maybe we could make it show you when you examine it.

The overlay gives you an idea of whether you meet the requirement but the idea of this is to tell the player the exact level requirement if they're interested without having to open the map to check. Examine text could be a better place for it but is slightly more hidden away.

PR #8660, opened 2019-04-27
opponentinfo: Option to show both HP value and percentage

seandewar:
A simple and maybe niche feature -- but I've sometimes found it useful to see both the opponent's HP and HP percentage (as a number):

Healthbar

Currently, you can only see either the approx. HP value (if max HP is known) **_or_** the raw percentage. 

With this proposed _Both_ option, the percentage from the approx. health over max health ratio is shown too.

Config

This PR slightly changes the `ProgressBarComponent` to allow both its current value and percentage to be shown if wanted.

There should be enough room to display up to 9999 HP with the _Both_ setting without the label escaping the bounds of the panel by default, I think.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
@@ -164,7 +164,10 @@ class OpponentInfoOverlay extends Overlay
 			progressBarComponent.setBackgroundColor(HP_RED);
 			progressBarComponent.setForegroundColor(HP_GREEN);
 
-			if (lastMaxHealth != null && !opponentInfoConfig.showPercent())
+			final HitpointsDisplayStyle displayStyle = opponentInfoConfig.hitpointsDisplayStyle();
+
+			if ((displayStyle == HitpointsDisplayStyle.HITPOINTS || displayStyle == HitpointsDisplayStyle.BOTH)
+				&& lastMaxHealth != null)
 			{
 				// This is the reverse of the calculation of healthRatio done by the server
 				// which is: healthRatio = 1 + (healthScale - 1) * health / maxHealth (if health > 0, 0 otherwise)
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/components/ProgressBarComponent.java
```diff
@@ -77,6 +78,10 @@ public Dimension render(Graphics2D graphics)
 			case PERCENTAGE:
 				textToWrite = DECIMAL_FORMAT.format(pc * 100d) + "%";
 				break;
+			case BOTH:
+				textToWrite = DECIMAL_FORMAT_ABS.format(Math.floor(currentValue)) + "/" + maximum
```
deathbeam: reuse the code for percentages above, and code  for full format below.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -194,11 +197,15 @@ else if (opponent instanceof Player)
...
+					// Take the average of min and max possible healths
 					health = (minHealth + maxHealth + 1) / 2;
 				}
 
-				progressBarComponent.setLabelDisplayMode(ProgressBarComponent.LabelDisplayMode.FULL);
+				// Show both the hitpoint and percentage values if enabled in the config
+				final ProgressBarComponent.LabelDisplayMode progressBarDisplayMode = displayStyle == HitpointsDisplayStyle.BOTH ?
+					ProgressBarComponent.LabelDisplayMode.BOTH : ProgressBarComponent.LabelDisplayMode.FULL;
+
+				progressBarComponent.setLabelDisplayMode(progressBarDisplayMode);
```
Adam-: I think `setLabelDisplayMode` is missing on the other side of the branch

seandewar: Do you want an explicit call to `setLabelDisplayMode(LabelDisplayMode.PERCENT)` for when the maxHP isn't known/percentage requested in the config?

It's currently the default value for a newly constructed progress bar, which happens on line 163.

Adam-: Oh, I had misread that. This seems fine.

PR #8664, opened 2019-04-27
chat message manager: don't send messages if player is on tutorial island

Trevor159:
If we send messages on tutorial island it looks like this and can get in the way of the tutorial.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
@@ -27,12 +27,14 @@ package net.runelite.client.chat;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import java.awt.Color;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Objects;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


Trevor159:
updated to use region ids

runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
```diff
@@ -56,6 +58,8 @@
 @Singleton
 public class ChatMessageManager
 {
+	private static final Set<Integer> tutIslandRegions = ImmutableSet.of(12336, 12335, 12592, 12080, 12079, 12436);
```
deathbeam: UPPER_CASE

runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
```diff
@@ -550,6 +554,12 @@ public void process()
 
 	private void add(QueuedMessage message)
 	{
+		// Do not send message if the player is on tutorial island
+		if (TUTORIAL_ISLAND_REGIONS.contains(client.getLocalPlayer().getWorldLocation().getRegionID()))
```
deathbeam: You should also null-check player, probably. As add can be called from login screen, I think, even though it does nothing.

runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
```diff
@@ -550,6 +555,13 @@ public void process()
 
 	private void add(QueuedMessage message)
 	{
+		// Do not send message if the player is on tutorial island
+		final Player player = client.getLocalPlayer();
+		if (player == null || TUTORIAL_ISLAND_REGIONS.contains(player.getWorldLocation().getRegionID()))
```
deathbeam: This should be changed to player != null && to not change existing behaviour

PR #8669, opened 2019-04-27
Use xenial in Travis

deathbeam:
trusty is going EOL at end of this month and this also fixes openjdk11
builds.

Ref: https://travis-ci.community/t/oracle-jdk11-build-error/1865/6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 sudo: false
-dist: trusty
+dist: xenial
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


deathbeam:
I was not able to figure out anything better than to also change oraclejdk8 to openjdk8 because for some reason oraclejdk8 is simply not availbe on xenial :man_shrugging: 

Now both builds should be working. It's kind of stupid that something like this suddenly breaks and there is no announcement on Travis anywhere or anything.

PR #8678, opened 2019-04-28
Pin Runelite plugin by default

### LINKED ISSUE #6256: Pin RuneLite "plugin" by default

beaumitch:
Resolves #6256

noConfig

This only effects new users without a config CSV.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
@@ -650,7 +650,7 @@ public class ConfigPanel extends PluginPanel
 
 		if (config == null)
 		{
-			return Collections.emptyList();
+			return Collections.singletonList(RUNELITE_PLUGIN);
 		}
 
 		return Text.fromCSV(config);


raiyni:
this does nothing for most users though

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -650,7 +650,7 @@ void stopPlugin(Plugin plugin, PluginListItem listItem)
 
 		if (config == null)
```
Adam-: What happens when all plugins are unpinned? I think this becomes an empty string.

beaumitch:
> 
> this does nothing for most users though

To effect existing users I think we'd have to add an ugly key to the config specifically for overriding their current setting to avoid overwriting the setting on each config reload. This at least effects new users - which are the most likely to need this feature.

Adam-:
I wouldn't mind seeing the RL config just permanently pinned at the top. It's always hard for users to find.

beaumitch:
> 
> 
> I wouldn't mind seeing the RL config just permanently pinned at the top. It's always hard for users to find.

Since space isn't really an issue in that panel, I'm leaning in this direction, too.

deathbeam:
I said it multiple times before, but that is not because its not pinned at top, but because it is named "RuneLite". So simple solution is to rename it to something that makes sense when someone asks.

deathbeam:
> > I wouldn't mind seeing the RL config just permanently pinned at the top. It's always hard for users to find.
> 
> Since space isn't really an issue in that panel, I'm leaning in this direction, too.

I disagree, if we want to "pin" it at top, it should be possible to unpin it. "RuneLite" do not contains any settings you would want to change often, and the pin feature was made for that wasnt it, so it will be taking up useless space at top.

beaumitch:
> 
> 
> > > I wouldn't mind seeing the RL config just permanently pinned at the top. It's always hard for users to find.
> > 
> > 
> > Since space isn't really an issue in that panel, I'm leaning in this direction, too.
> 
> I disagree, if we want to "pin" it at top, it should be possible to unpin it. "RuneLite" do not contains any settings you would want to change often, and the pin feature was made for that wasnt it, so it will be taking up useless space at top.

This is a good point, potentially "Client" or something like that. No longer interested in seeing this merged, so I'll close for now. Tagging #6256 to preserve conversation.

PR #8696, opened 2019-04-30
daily tasks plugin: tidy up if statements and make less config options on by default

Trevor159:


Reasel:
Same changes are already done in #8583

deathbeam:
I like it being in separate commits more @Reasel . Also it would make reviewing #8583 easier

ShaunDreclin:
I'm pretty sure staves require a relog, this PR removed that message from the notification

PR #8708, opened 2019-05-01
Aggro area: only add a timer if the time left isn't negative

Lucwousin:
Fixes `java.lang.IllegalArgumentException: negative period!` when logging in if the monsters in the area you are aren't aggro'd to you anymore.


deathbeam:
Your commit author is different account than your github account, is that intentional?

Lucwousin:
Honestly I have no idea why it's like that. I could probably figure out how to change it if needed

deathbeam:
If you have no idea then yeah, you probably should. git commit --amend --author="" with your current account email should do it

Lucwousin:
Changed!

Issue #8709, opened 2019-05-01
Maven variable consistency


WilliamCollishaw:
The maven plugins and dependency version variables are inconsistent. They differ between hard coded and variable.
This issue will focus on converting all versions to variables and if the package is used across multiple modules, the version will be defined in the parent pom.
 EDIT: Prevent upgrading any packages. Focus on the variable consistency.

abextm:
The way it is now (and the way it should stay) is that anything where there are >1 module depending on a single dependency should be in `dependencyManagement` in the root pom. The properties are for when there are multiple dependencies that should be versioned together, (jogl and its natives for example).

Additionally we can't easily accept PRs relating to the root pom as we have to rebase them onto our internal branch, which for things like this means redoing them from scratch. Best case scenario is that is our internal branch has an inconsistent style, worst (and imo most probable) case it breaks in some non-obvious way.

PR #8710, opened 2019-05-01
Maven version variable consistency and dependency management

WilliamCollishaw:
#8709
Set maven package versions to variables and using dependency management in the parent pom when the package is used in multiple modules

~~Upgrade maven packages~~
~~Prevent trimming stack trace to maven surefire which caused newer versions to not display stacktrace in Netbeans.~~

~~Upgrade Maven jar plugin from 3.0.2 to 3.1.1~~
~~Upgrade Maven compiler plugin version from 3.6.1 to 3.8.0~~
~~Upgrade Maven checkstyle plugin version from 2.17 to 3.0.0~~
~~Upgrade Maven javadoc plugin from 3.0.0-M1 to 3.1.0~~
~~Upgrade Maven surefire plugin from 2.18 to 2.22~~
~~Upgrade Maven plugin api from 3.0.5 to 3.6.1~~
~~Upgrade Maven plugin annotations from 3.4 to 3.6.0~~
~~Upgrade Maven plugin plugin from 3.4 to 3.6.0~~
~~Upgrade Maven git commit id plugin from 2.2.5 to 2.2.6~~
~~Upgrade Maven jarsigner plugin from 1.4 to 3.0.0~~
~~Upgrade Maven resource plugin from 3.0.2 to 3.1.0~~


deathbeam:
You should split version normalization and version upgrades to separate commits

WilliamCollishaw:
I was somewhat conflicted between defining package versions in the pom of the module the package it is used when only used in one module, or defining all version variables in the parent pom. Any opinions?

deathbeam:
You usually do dependencyManagment in cases like this and have shared parent that includes that, and then you just do not specify versions in submodules

WilliamCollishaw:
> You should split version normalization and version upgrades to separate commits

Yeah, I can do this. Please see my previous comment regarding version normalization

Adam-:
I probably don't want to upgrade anything without having a reason to (security, stability, bug fixes, new features?). I also certainly wouldn't want to upgrade everything at once because when it inevitable breaks its harder to pinpoint. What we have now works without any issuses that I know of though.

WilliamCollishaw:
> I probably don't want to upgrade anything without having a reason to (security, stability, bug fixes, new features?). I also certainly wouldn't want to upgrade everything at once because when it inevitably breaks its harder to pinpoint. What we have now works without any issuses that I know of though.

Sure, I can reset the upgrades and focus on the dependency management consistency.

WilliamCollishaw:
Reset plugin upgrades, and set versions in parent pom dependency management for packages used over multiple modules.

WilliamCollishaw:
Merged master and resolved conflict from #8732

PR #8728, opened 2019-05-02
chatnotifications: add incoming private message notification

### LINKED ISSUE #7687: Add a PM received notification to the Chat Notifications plugin
It often happens that I am AFK. Friends do not always know this and talk to me. They get the feeling that I ignore them, while that is not the case. 

An extra toggle in the Chat Notifications plugin to receive a notification, containing the name and message of the friend, would be really nice.

odxs:
Resolves #7687

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
@@ -27,7 +27,10 @@ package net.runelite.client.plugins.chatnotifications;
 
 import com.google.common.base.Strings;
 import com.google.inject.Provides;
+
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
```diff
@@ -76,6 +79,9 @@
 	private String usernameReplacer = "";
 	private Pattern highlightMatcher = null;
 
+	// Private message cache used to avoid duplicate notifications from ChatHistory.
```
Nightfirecat: I don't understand the purpose of this. What problem is this solving?

odxs: Whenever you world hop and have ChatHistory enabled, you will be notified of messages from the past that you may have already received notifications about.

Adam-:
I would like to see the chathistory-resending-notifications problem fixed correctly first instead of hacking more things around it. We already have a couple of places that try to work around it.

PR #8729, opened 2019-05-02
Recolor also tag option in NPC indicators

deathbeam:
Without this the tag option will not be recolored with highlight menu
names enabled.


Adam-:
I think people would use the color to quickly identify npcs in piles, and noone is trying to quickly tag npcs. Also it's not really a *real* option anyway.

PR #8730, opened 2019-05-02
AgilityPlugin: Add "Laps to Goal" and toggles for overlay

### LINKED ISSUE #8677: Allow Agility plugin to count laps to goal, rather than laps to level.
https://i.imgur.com/fnID9N7.png <- Possible solution

Not game-changing, just something nice that could comfort people who are grinding out agility.

Thanks :)

beaumitch:
Resolves #8677 

agility_toggle_on
agility_toggle_off

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
@@ -30,6 +30,7 @@ import lombok.Setter;
 import net.runelite.api.Client;
 import net.runelite.api.Experience;
 import net.runelite.api.Skill;
+import net.runelite.api.VarPlayer;
 
 @Getter
 @Setter
@@ -39,6 +40,7 @@ class AgilitySession
 	private Instant lastLapCompleted;
 	private int totalLaps;
 	private int lapsTillLevel;
+	private int lapsTillGoal;
 
 	AgilitySession(Courses course)
 	{
@@ -61,11 +63,14 @@ class AgilitySession
 		} while (remainingXp < 0);
 
 		lapsTillLevel = remainingXp > 0 ? (int) Math.ceil(remainingXp / course.getTotalXp()) : 0;
+		int goalRemainingXp = client.getVar(VarPlayer.AGILITY_GOAL_END) - currentExp;
+		lapsTillGoal = goalRemainingXp > 0 ? (int) Math.ceil(goalRemainingXp / course.getTotalXp()) : 0;
 	}
 
 	void resetLapCount()
 	{
 		totalLaps = 0;
 		lapsTillLevel = 0;
+		lapsTillGoal = 0;
 	}
 }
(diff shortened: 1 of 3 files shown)


Adam-:
The commit message for this doesn't match what it is actually doing

deathbeam:
The toggles are very unnecessary. Also, imo, this should be integrated to XP tracker, because it already supports XP goals, and everything else uses it already.

Adam-:
If it isn't too much hackery putting it into the xptracker seems like a good idea.

deathbeam:
It isnt very trivial to move it. But it's doable (I had older PR for that, but I didnt finished it)

LukePeltier:
is there something preventing this change from being merged? Would love to see this feature

Adam-:
We were discussing putting it into the xp tracker and then referencing that (see the previous 2 comments) - but if noone wants to do that I guess we can just take this.

LukePeltier:
Yea. if it matters at all, i think the agility section would be a more discoverable place for this setting anyway

LukePeltier:
I just understood what was meant by moving it to the xp tracker... 😅

deathbeam:
This has nothing to do with agility section being "more discoverable place for this". Its just so it works also in XP tracker instead of being implemented just here, and then we could simply just take this information from there and display it also as overlay.

Even though now with on canvas xp trackers, the agility overlay would not even be necessary at all.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -61,11 +63,14 @@ void incrementLapCount(Client client)
 		} while (remainingXp < 0);
 
 		lapsTillLevel = remainingXp > 0 ? (int) Math.ceil(remainingXp / course.getTotalXp()) : 0;
+		int goalRemainingXp = client.getVar(VarPlayer.AGILITY_GOAL_END) - currentExp;
```
deathbeam: this should use XpTrackerService if this is going to be implemented like this. There is variable for endGoalXp (or something like that). Remove this and use that.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -61,11 +63,14 @@ void incrementLapCount(Client client)
 		} while (remainingXp < 0);
 
 		lapsTillLevel = remainingXp > 0 ? (int) Math.ceil(remainingXp / course.getTotalXp()) : 0;
```
deathbeam: This should just be merged together with goal XP and show next goal based on player's goal settings instead of displaying both

Adam-:
This seems like such a small thing that could move to the xptracker later anyway. I don't think anyone cares enough to move it there atm.

deathbeam:
This PR had unresolved comments (unrelated to the migration to XP tracker), why was it merged?

Adam-:
See https://github.com/runelite/runelite/pull/8730#issuecomment-531823116

deathbeam:
Again, my review comments were unrelated to merge with XP tracker.

Adam-:
I don't really mind it showing both. If we refactor it anyway then it would take the behavior of the tracker at least which is what you describe there.

PR #8738, opened 2019-05-03
Show loss rate of ammo

benjefferies:
Whilst training it's useful to know how much ammo per time unit you're
losing. This can help predict how much ammo needs to be bought for
training.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
@@ -25,22 +25,35 @@
 package net.runelite.client.plugins.ammo;
 
 import java.awt.image.BufferedImage;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.Temporal;
+
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.ui.overlay.infobox.Counter;
 import net.runelite.client.util.StackFormatter;
 
+@Slf4j
 class AmmoCounter extends Counter
 {
 	@Getter
-	private final int itemID;
-	private final String name;
+	private int itemID;
+	private String name;
+	private int total;
+	private Instant time;
 
 	AmmoCounter(Plugin plugin, int itemID, int count, String name, BufferedImage image)
 	{
 		super(image, plugin, count);
+		this.total = count;
 		this.itemID = itemID;
 		this.name = name;
+		this.time = Instant.now();
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


benjefferies:
@deathbeam this is ready to go and I've tested 👍

Adam-:
How does this work? Why wouldn't ava's break it?

benjefferies:
It works out how much ammo you have to start with and then based on how much ammo is being dropped it shows the rate of ammo loss. Ava's still has a drop rate

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
```diff
@@ -25,6 +25,11 @@
 package net.runelite.client.plugins.ammo;
 
 import java.awt.image.BufferedImage;
+import java.math.BigDecimal;
```
deathbeam: Fix your imports: https://github.com/runelite/runelite/wiki/Code-Conventions

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
```diff
@@ -33,14 +38,18 @@
 class AmmoCounter extends Counter
 {
 	@Getter
-	private final int itemID;
-	private final String name;
+	private int itemID;
```
deathbeam: this should still be final

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
```diff
@@ -33,14 +38,18 @@
 class AmmoCounter extends Counter
 {
 	@Getter
-	private final int itemID;
-	private final String name;
+	private int itemID;
+	private String name;
```
deathbeam: this too

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
```diff
@@ -33,14 +38,18 @@
 class AmmoCounter extends Counter
 {
 	@Getter
-	private final int itemID;
-	private final String name;
+	private int itemID;
+	private String name;
+	private int total;
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
```diff
@@ -33,14 +38,18 @@
 class AmmoCounter extends Counter
 {
 	@Getter
-	private final int itemID;
-	private final String name;
+	private int itemID;
+	private String name;
+	private int total;
+	private Instant time;
```
deathbeam: final

winterdaze-rs:
As far as I can tell this plugin takes note of the starting amount of ammo and time and then updates/calculates whenever the equipped ammo stack or weapon count changes, so it doesn't depend on the ava's type equipped (it's not an estimate based on % loss). However, I don't think this works with the blowpipe. Also, what happens when you add value to the stack, like if you buy ammo from the GE then equip it, it changes the stack so you'd get a "gain" rate? And if you pick up dropped ammo on the ground, that would affect the hourly loss rate too (is this the intended behaviour?)

benjefferies:
Not intended behaviour @winterdaze-rs. Maybe a simple solution would be to reset the counter if more than X added to the stack

winterdaze-rs:
Maybe just don't count any positive changes to the stack value then? If they're buying stuff from the GE or banking to re-supply they're not in combat anyways so the counter should be reset to be accurate, right? Or you could maybe re-work it to not depend on initial starting amounts and simply keep track of the differences only. Something like this, where you only use the sum of lost ammo? 

Previous Darts, Current Darts, Sum of Lost Darts
100, 100, 0 // Initially prev and current the same
100, 99, 1 // On negative change, update prev/current/sum
99, 98, 2 
98, 198, 2 // Player goes to resupply, don't update sum but update prev/current
198, 197, 3 // On negative change, update prev/current/sum

PR #8742, opened 2019-05-03
Added range to NPC Indicator

StaleDeveloper:
Given user option to edit the range the monsters will be highlighted to instead of the constant 15.

Works on both minimap and screen rendering.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
@@ -107,4 +107,14 @@ public interface NpcIndicatorsConfig extends Config
 	{
 		return false;
 	}
+	
+	@ConfigItem(
+		position = 7,
+		keyName = "drawDistance",
+		name = "Monster Render Distance",
+		description = "The distance that monsters will be highlighted up to.")
+	default int drawDistance()
+	{
+		return 3;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcMinimapOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcMinimapOverlay.java
@@ -56,9 +56,12 @@ public class NpcMinimapOverlay extends Overlay
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
+		
 		for (NPC npc : plugin.getHighlightedNpcs())
 		{
-			renderNpcOverlay(graphics, npc, npc.getName(), config.getHighlightColor());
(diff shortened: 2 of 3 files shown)


ShaunDreclin:
I wonder if this should be moved to a setting in `RuneLite`? Rather than individual overlay plugins having individual range options

StaleDeveloper:
> I wonder if this should be moved to a setting in `RuneLite`? Rather than individual overlay plugins having individual range options

Sounds plausible. Would make it easier for those who struggle power wise.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
```diff
@@ -107,4 +107,14 @@ default boolean showRespawnTimer()
...
 	}
+	
+	@ConfigItem(
+		position = 7,
+		keyName = "drawDistance",
+		name = "Monster Render Distance",
+		description = "The distance that monsters will be highlighted up to.")
+	default int drawDistance()
+	{
+		return 3;
```
Adam-: This should not default to 3. If merged, other people have to use your code too, and most people would consider this to be broken.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
```diff
@@ -107,4 +107,14 @@ default boolean showRespawnTimer()
 	{
 		return false;
 	}
+	
+	@ConfigItem(
+		position = 7,
+		keyName = "drawDistance",
+		name = "Monster Render Distance",
+		description = "The distance that monsters will be highlighted up to.")
```
Adam-:
```suggestion
		description = "The distance that monsters will be highlighted up to."
	)
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcMinimapOverlay.java
```diff
@@ -56,9 +56,12 @@
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
+		
 		for (NPC npc : plugin.getHighlightedNpcs())
 		{
-			renderNpcOverlay(graphics, npc, npc.getName(), config.getHighlightColor());
+			if ((int) Math.hypot(client.getLocalPlayer().getWorldLocation().getX() - npc.getWorldLocation().getX(), client.getLocalPlayer().getWorldLocation().getY() - npc.getWorldLocation().getY()) < config.drawDistance()){
```
Adam-: There is `WorldPoint.distanceTo` and also `LocalPoint.distanceTo`. One of them probably does what you want.

Issue #2134, opened 2018-04-29
Show countdown timer to prayer drain in Barrows


rherms:
I am doing a big Barrows grind right now and it would be useful for potting before fighting Dharok when he is in the tunnels and for knowing if I need to not be so afk and get out of the crypt quickly when fighting the other brothers. Anyone else interested in this feature? If so, I might take a stab at coding it!

Reasel:
Well does the prayer drain at a set rate each time as in 10 each time it procs or is it scaling? If the timer is based on entering then you should be able to time from each tick if you are in the crypts as well as underground. You would need to verify both of those in game yourself most likely and then add it to the barrows plugin.

Reasel:
Yeah it should be a simple thing and would be super useful.

FatalWaffles:
It should drain every 18 seconds, at which rate I'm not sure though.

PR #8746, opened 2019-05-04
Add barrows prayer drain timer

rherms:
Fixes https://github.com/runelite/runelite/issues/2134

This adds a timer that has a duration of 18.2 seconds and only appears inside Barrows crypts or tunnels. I've tested all situations of being in a crypt and then going into a tunnel, being in a crypt/tunnel and logging out, and dying in Barrows. It's my first PR so please point out any mistakes I made! Really hoping this can be in the next release. I've found it quite useful :)

238f546d2eb62ddd370a2ffca5aaf0a1

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
@@ -42,6 +42,7 @@ import net.runelite.api.ItemContainer;
 import net.runelite.api.NullObjectID;
 import net.runelite.api.ObjectID;
 import net.runelite.api.WallObject;
+import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.GameObjectChanged;
 import net.runelite.api.events.GameObjectDespawned;
 import net.runelite.api.events.GameObjectSpawned;
@@ -60,9 +61,11 @@ import net.runelite.client.chat.QueuedMessage;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.game.ItemManager;
+import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.OverlayManager;
+import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
 import net.runelite.client.util.StackFormatter;
 
 @PluginDescriptor(
@@ -89,11 +92,15 @@ public class BarrowsPlugin extends Plugin
 		WidgetInfo.BARROWS_PUZZLE_ANSWER3
 	);
 
+	private static final int CRYPT_REGION_ID = 14231;
+
 	@Getter(AccessLevel.PACKAGE)
 	private final Set<WallObject> walls = new HashSet<>();
 
 	@Getter(AccessLevel.PACKAGE)
 	private final Set<GameObject> ladders = new HashSet<>();
+
+	private boolean wasInCrypt = false;
 
 	@Getter
 	private Widget puzzleAnswer;
@@ -112,6 +119,12 @@ public class BarrowsPlugin extends Plugin
 
 	@Inject
 	private ItemManager itemManager;
+
+	@Inject
+	private SpriteManager spriteManager;
+
+	@Inject
+	private InfoBoxManager infoBoxManager;
 
 	@Inject
 	private ChatMessageManager chatMessageManager;
@@ -140,6 +153,8 @@ public class BarrowsPlugin extends Plugin
 		walls.clear();
 		ladders.clear();
 		puzzleAnswer = null;
+		wasInCrypt = false;
+		stopPrayerDrainTimer();
 
 		// Restore widgets
 		final Widget potential = client.getWidget(WidgetInfo.BARROWS_POTENTIAL);
@@ -152,6 +167,15 @@ public class BarrowsPlugin extends Plugin
 		if (barrowsBrothers != null)
 		{
 			barrowsBrothers.setHidden(false);
+		}
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals("barrows") && !config.showPrayerDrainTimer())
+		{
+			stopPrayerDrainTimer();
 		}
 	}
 
@@ -220,10 +244,23 @@ public class BarrowsPlugin extends Plugin
 	{
 		if (event.getGameState() == GameState.LOADING)
 		{
+			wasInCrypt = isInCrypt();
 			// on region changes the tiles get set to null
 			walls.clear();
 			ladders.clear();
 			puzzleAnswer = null;
+		}
+		else if (event.getGameState() == GameState.LOGGED_IN)
+		{
+			boolean isInCrypt = isInCrypt();
+			if (wasInCrypt && !isInCrypt)
+			{
+				stopPrayerDrainTimer();
+			}
+			else if (!wasInCrypt && isInCrypt)
+			{
+				startPrayerDrainTimer();
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPrayerDrainTimer.java
```diff
@@ -0,0 +1,56 @@
...
+import java.time.temporal.ChronoUnit;
+
+import net.runelite.api.SpriteID;
+import net.runelite.client.game.SpriteManager;
+import net.runelite.client.ui.overlay.infobox.InfoBoxPriority;
+import net.runelite.client.ui.overlay.infobox.Timer;
+
+class BarrowsPrayerDrainTimer extends Timer
+{
+    private static final Duration PRAYER_DRAIN_INTERVAL = Duration.of(18200, ChronoUnit.MILLIS);
```
raiyni: This can be a long

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPrayerDrainTimer.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2017, Adam <Adam@sigterm.info>
```
raiyni: Update copyright

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
```diff
@@ -53,6 +53,12 @@ public Timer(long period, ChronoUnit unit, BufferedImage image, Plugin plugin)
 		endTime = startTime.plus(duration);
 	}
 
+	public void reset()
```
Nightfirecat: If you change the fields to `protected` access level here, you can move this method into BarrowsPrayerDrainTimer as a private instance method instead of adding it here.

raiyni: I'll leave my comment from discord here. I'm not sure time should be mutable like this but I don't have a strong opinion.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -193,6 +212,10 @@ public void onGameObjectSpawned(GameObjectSpawned event)
 		{
 			ladders.add(gameObject);
 		}
+		if (BARROWS_SARCOPHAGI.contains(gameObject.getId()))
+		{
+			isSarcophagusPresent = true;
```
Adam-: This seems needlessly complex, since all of the sarcophagi are in one region, you can just check the current region

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
Adam-: I would much prefer just recreating a new timer.

deathbeam: Or just dont use timer at all, because it do not fits this scenario

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
rherms: What do you mean by region? Like checking the plane of a player?

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/Timer.java
rherms: How would you avoid a concurrent modification exception? How could you hook into the timer expiration to create a new one?

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
Adam-: If you get the worldloc of the player and getRegionID() there is one specific to barrows. It is a 64x64 map square.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -220,11 +235,35 @@ public void onGameStateChanged(GameStateChanged event)
...
+		{
+			if (wasInCrypt && !isInCrypt())
+			{
+				prayerDrainTimer.stop();
+			}
+			else if (!wasInCrypt && isInCrypt())
+			{
+				if (prayerDrainTimer == null)
+				{
+					prayerDrainTimer = new BarrowsPrayerDrainTimer(this, spriteManager);
```
rherms: Is there another place I can initialize this so I don't have to mark it as Nullable and lazily initialize it? I tried to do it in startUp, but that resulted in the spriteManager being unable to get the image.

rherms:
Implemented feedback from talking to deathbeam on Discord. This is looking a lot better now!

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -140,6 +154,7 @@ protected void shutDown()
 		walls.clear();
 		ladders.clear();
 		puzzleAnswer = null;
+		stopPrayerDrainTimer();
```
deathbeam: you should also reset wasInCrypt to false on plugin shutdown

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -271,4 +308,28 @@ else if (event.getGroupId() == WidgetID.BARROWS_PUZZLE_GROUP_ID)
 			}
 		}
 	}
+
+	@Subscribe
+	public void onLocalPlayerDeath(LocalPlayerDeath event)
+	{
+		stopPrayerDrainTimer();
```
Adam-: There are more ways to leave than just dieing, and i think this would trigger the gamestate change from above, so why is this here?

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -155,6 +171,15 @@ protected void shutDown()
 		}
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (!config.showPrayerDrainTimer())
```
Adam-: This should be checking the config group at least which changed.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -271,4 +308,28 @@ else if (event.getGroupId() == WidgetID.BARROWS_PUZZLE_GROUP_ID)
...
+	{
+		if (config.showPrayerDrainTimer())
+		{
+			infoBoxManager.addInfoBox(new BarrowsPrayerDrainTimer(this, spriteManager));
+		}
+	}
+
+	private void stopPrayerDrainTimer()
+	{
+		infoBoxManager.removeIf(t -> t instanceof BarrowsPrayerDrainTimer);
```
Adam-:
```suggestion
		infoBoxManager.removeIf(BarrowsPrayerDrainTimer.class::isInstance);
```

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
rherms: Oh, if a a gamestate change happens on death, then I probably don't need this.

Adam-: I think it does. Why did you add it? You should be testing whether or not the plugin behaves correctly when leaving or teleporting from the crypt.

rherms: I think I just added it because I saw it in another plugin. But you're right, the existing game state change logic catches all those cases.

Adam-:
@deathbeam did you want to wait on #8800 for this?

deathbeam:
No need to wait for it imo, I can always just update my PR after this will be merged.

PR #8748, opened 2019-05-04
IconTextField: make mouse listener affect the whole area

Trevor159:
I was looking into adding doubling clicking the search bar in the hiscore panel to lookup the local player like OSB used to have. I never realized we already had this because right now it only works if you double click the picture section of the search bar which is a low percent of the total area. This lets you also double click the textbox part of the search bar.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/components/IconTextField.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/components/IconTextField.java
@@ -169,6 +169,20 @@ public class IconTextField extends JPanel
 		add(clearButton, BorderLayout.EAST);
 	}
 
+	@Override
+	public void addMouseListener(MouseListener mouseListener)
+	{
+		super.addMouseListener(mouseListener);
+		textField.addMouseListener(mouseListener);
+	}
+
+	@Override
+	public void removeMouseListener(MouseListener mouseListener)
+	{
+		super.removeMouseListener(mouseListener);
+		textField.removeMouseListener(mouseListener);
+	}
+
 	public void addActionListener(ActionListener actionListener)
 	{
 		textField.addActionListener(actionListener);
--- a/runelite-client/src/main/java/net/runelite/client/ui/components/FlatTextField.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/components/FlatTextField.java
@@ -31,6 +31,7 @@ import java.awt.event.ActionListener;
 import java.awt.event.KeyListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.awt.event.MouseListener;
 import javax.swing.JPanel;
 import javax.swing.JTextField;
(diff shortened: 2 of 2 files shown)


Trevor159:
In case it wasn't obvious here is a picture



The left of the red line is what you can double click currently, with this PR you can click the whole box.

Adam-:
Shouldn't there be a removeMouseListener too?

Trevor159:
> Shouldn't there be a removeMouseListener too?

I can make one if wanted. It was not used at all in the client currently for these classes so I didn't.

deathbeam:
I agree, you should add also remove part, even if it is unused

Adam-:
Why is the icon text field mouse listener added to the panel and the flat text field not?

Trevor159:
> Why is the icon text field mouse listener added to the panel and the flat text field not?

I am not a swing expert so this could be wrong but the reason I did it is because in FlatTextField the JTextField covers the whole panel so adding the mouse listener to the panel would be redundant.

Adam-:
Why is it adding the listener to the jtextfield in flattextfield then if the panel already covers it?

deathbeam:
So, as far as I remember:
1. This functionality was originally intentional
2. Panel is not covering anything, you *need* to add listeners to child components in Swing if you want some hover effect or w/e, that is just how it works (unless it's just label, that one is pass-through I think)

runelite-client/src/main/java/net/runelite/client/ui/components/IconTextField.java
```diff
@@ -169,6 +169,20 @@ public void changedUpdate(DocumentEvent e)
 		add(clearButton, BorderLayout.EAST);
 	}
 
+	@Override
+	public void addMouseListener(MouseListener mouseListener)
+	{
+		super.addMouseListener(mouseListener);
```
deathbeam: should this really apply to whole icon text field? and not just text field and icon?

runelite-client/src/main/java/net/runelite/client/ui/components/IconTextField.java
```diff
@@ -169,6 +169,20 @@ public void changedUpdate(DocumentEvent e)
...
+		textField.addMouseListener(mouseListener);
+	}
+
+	@Override
+	public void removeMouseListener(MouseListener mouseListener)
+	{
+		super.removeMouseListener(mouseListener);
```
deathbeam: see my previous comment

raiyni:
Wouldn't this be fairly confusing behavior? Go in expecting to double click to highlight whole like (like every other text field that exists) and instead it does a lookup on your rsn.

deathbeam:
Maybe this could check if the text field is empty then?

Nightfirecat:
A suggestion from #8968 which I thought might help this discussion--maybe a right-click menu option would work here?

Trevor159:
I don't know why I ever cared enough to make this PR.

PR #8756, opened 2019-05-05
Menu Entry Swapper: Add 'Search' swap

ShaunDreclin:
To fix the regular annoyance of opening something, closing it by accident, opening it again, then having to right click and select search


Nightfirecat:
This is a niche feature that is best suited for the plugin hub, and is currently covered by the Menu swapper extented plugin there. See runelite/plugin-hub#42

ShaunDreclin:
I wonder if some things should be moved out of the main plugin, then? Seems like there are a few niche features in it

Trevor159:
while you are right, we do not like removing features from the main client.

PR #8761, opened 2019-05-05
Fix bug where !lvl command displays -1 for rank

JacobThompson:
The chat command !lvl displays -1 for the rank if the player has no experience. This update fixes that and display "unranked" instead of -1.


Adam-:
Please update your code style to match the project guidelines https://github.com/runelite/runelite/wiki/Checkstyle-with-IntelliJ-IDEA

PR #8763, opened 2019-05-05
Add plugins used in more than 1 location into plugin management.

WilliamCollishaw:
Remove second reference to maven-compiler-plugin version, as it is defined in the pluginManagement
Added maven-surefire-plugin version to pluginManagement as it is used in the parent and cache module
Add reproducible-build-maven-plugin version to pluginManagement as it is used in the parent and runelite-client module
Add spring-boot-maven-plugin version to pluginManagement as it is used in the http-service and cache-updater module


pom.xml
```diff
@@ -39,6 +39,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>1.8</java.version>
 		<lombok.version>1.18.4</lombok.version>
+		<spring.boot.version>1.5.6.RELEASE</spring.boot.version>
```
deathbeam: Any reason to have this as variable? It is used only once right

WilliamCollishaw: Right now the same version is used across the parent pom, http-service pom and cache adapter pom, maybe it should just be set once here in the parent and removed in the other two poms?

deathbeam: yeah

WilliamCollishaw: After rechecking my pull request, that is actually what I had already done, the variable is just defined once in the parent and used throughout the project 
Screen Shot 2019-05-12 at 11 44 58 AM

pom.xml
```diff
@@ -253,6 +250,22 @@
...
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-surefire-plugin</artifactId>
+					<!-- newer versions seem to not show stacktraces in Netbeans -->
+					<version>2.18</version>
+				</plugin>
+				<plugin>
+					<groupId>io.github.zlika</groupId>
+					<artifactId>reproducible-build-maven-plugin</artifactId>
+					<version>0.7</version>
```
deathbeam: Is this used at more than 1 place?

WilliamCollishaw: @deathbeam yes it is used in the runelite-client module. However a version was never set in that module previously.
Screen Shot 2019-06-01 at 8 23 54 AM

deathbeam: Ah alright

WilliamCollishaw:
@deathbeam if a rebase to the internal root pom is being considered I would suggest checking out #8753 as well

PR #8770, opened 2019-05-06
Adds foreman fee counter for Blast Furnace under level 60 smithing

### LINKED ISSUE #8769: Add blast furnace foreman fee counter
It's always frustrating to have a full inventory of coal/ore and you get the unfortunate message of "You must ask the foreman for permission to use the Blast Furnace ..." when under level 60 smithing. 

An overlay with a countdown timer should be sufficient to avoid this issue. Players can plan accordingly to pay the foreman at next bank chest visit.

Other than setting a timer on another device, etc this seems to be the best integrated solution.

**Additional context**
java_2019-05-06_01-17-35

bwhite94:
For players that are utilizing blast furnace and require to pay the
foreman fee every 10 minutes, this will assist in tracking the time
remaining before the next fee payment is required. This defaults to
disabled, but is bundled with the Blast Furnace plugin config.

Resolves #8769

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, Seth <Sethtroll3@gmail.com>
+ * Copyright (c) 2019, Brandon White <bmwqg@live.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -25,21 +26,31 @@
 package net.runelite.client.plugins.blastfurnace;
 
 import com.google.inject.Provides;
+import java.time.Duration;
+import java.time.Instant;
 import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
+import net.runelite.api.Client;
 import net.runelite.api.GameObject;
 import net.runelite.api.GameState;
 import static net.runelite.api.NullObjectID.NULL_9092;
 import static net.runelite.api.ObjectID.CONVEYOR_BELT;
+import net.runelite.api.Skill;
 import net.runelite.api.events.GameObjectDespawned;
 import net.runelite.api.events.GameObjectSpawned;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
+import net.runelite.api.widgets.Widget;
+import net.runelite.api.widgets.WidgetInfo;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.game.ItemManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.OverlayManager;
+import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
+import net.runelite.client.util.Text;
 
 @PluginDescriptor(
 	name = "Blast Furnace",
@@ -49,12 +60,15 @@ import net.runelite.client.ui.overlay.OverlayManager;
 public class BlastFurnacePlugin extends Plugin
 {
 	private static final int BAR_DISPENSER = NULL_9092;
+	private static final String FOREMAN_PERMISSION_TEXT = "Okay, you can use the furnace for ten minutes. Remember, you only need half as much coal as with a regular furnace.";
 
 	@Getter(AccessLevel.PACKAGE)
 	private GameObject conveyorBelt;
 
 	@Getter(AccessLevel.PACKAGE)
 	private GameObject barDispenser;
+
+	private ForemanTimer foremanTimer;
 
 	@Inject
 	private OverlayManager overlayManager;
@@ -68,6 +82,15 @@ public class BlastFurnacePlugin extends Plugin
 	@Inject
 	private BlastFurnaceClickBoxOverlay clickBoxOverlay;
 
+	@Inject
+	private Client client;
+
+	@Inject
+	private ItemManager itemManager;
+
+	@Inject
+	private InfoBoxManager infoBoxManager;
+
 	@Override
 	protected void startUp() throws Exception
 	{
@@ -79,11 +102,13 @@ public class BlastFurnacePlugin extends Plugin
 	@Override
 	protected void shutDown()
 	{
+		infoBoxManager.removeIf(ForemanTimer.class::isInstance);
 		overlayManager.remove(overlay);
 		overlayManager.remove(cofferOverlay);
 		overlayManager.remove(clickBoxOverlay);
 		conveyorBelt = null;
 		barDispenser = null;
+		foremanTimer = null;
 	}
 
 	@Provides
@@ -133,6 +158,33 @@ public class BlastFurnacePlugin extends Plugin
 		{
 			conveyorBelt = null;
 			barDispenser = null;
+		}
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		Widget npcDialog = client.getWidget(WidgetInfo.DIALOG_NPC_TEXT);
+		if (npcDialog == null)
+		{
+			return;
+		}
+
+		// blocking dialog check until 5 minutes needed to avoid re-adding while dialog message still displayed
+		boolean shouldCheckForemanFee = client.getRealSkillLevel(Skill.SMITHING) < 60
+			&& (foremanTimer == null || Duration.between(Instant.now(), foremanTimer.getEndTime()).toMinutes() <= 5);
+
+		if (shouldCheckForemanFee)
+		{
+			String npcText = Text.sanitizeMultilineText(npcDialog.getText());
+
+			if (npcText.equals(FOREMAN_PERMISSION_TEXT))
+			{
+				infoBoxManager.removeIf(ForemanTimer.class::isInstance);
+
+				foremanTimer = new ForemanTimer(this, itemManager);
+				infoBoxManager.addInfoBox(foremanTimer);
+			}
 		}
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/ForemanTimer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/ForemanTimer.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019, Brandon White <bmwqg@live.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -50,12 +59,17 @@
 {
 	private static final int BAR_DISPENSER = NULL_9092;
 
+	private static final Pattern FOREMAN_PERMISSION_PATTERN = Pattern.compile("Okay, you can use the furnace for ten minutes. Remember, you only need half as much coal as with a regular furnace.");
```
Nightfirecat: Why are you using a `Pattern` for a static string?

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -25,21 +25,30 @@
...
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.OverlayManager;
+import net.runelite.client.util.Text;
+
+import javax.inject.Inject;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static net.runelite.api.NullObjectID.NULL_9092;
```
Nightfirecat: Why are these object IDs imported?

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -135,4 +158,27 @@ public void onGameStateChanged(GameStateChanged event)
...
+			foremanCounter--;
+		}
+		else if (npcDialog != null)
+		{
+			String npcText = Text.sanitizeMultilineText(npcDialog.getText());
+			final Matcher mPermission = FOREMAN_PERMISSION_PATTERN.matcher(npcText);
+
+			if (mPermission.find())
+			{
+				foremanCounter = 100 * 10;
```
Nightfirecat: Define the duration as a final static variable, and use a `Duration` instead of measuring in ticks.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -135,4 +158,27 @@ public void onGameStateChanged(GameStateChanged event)
...
 	}
+
+	@Subscribe
+	public void onGameTick(GameTick tick)
+	{
+		Widget npcDialog = client.getWidget(WidgetInfo.DIALOG_NPC_TEXT);
+
+		if (foremanCounter != null && foremanCounter > 0)
+		{
+			foremanCounter--;
```
Nightfirecat: Rather than measuring in ticks, save the time at which the foreman permission text is last found, then compute the time remaining in the overlay's render function. The time left to display should be the static duration (10 minutes) minus the difference between that saved timestamp and the current time.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
```diff
@@ -0,0 +1,73 @@
...
+		this.client = client;
+		this.plugin = plugin;
+		this.config = config;
+		getMenuEntries().add(new OverlayMenuEntry(RUNELITE_OVERLAY_CONFIG, OPTION_CONFIGURE, "Foreman overlay"));
+	}
+
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		if (config.showForemanCounter())
```
Nightfirecat: Use a guard clause instead:

```java
if (!config.showForemanCounter())
{
	return null;
}

// ...rest of render code goes here
```

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
```diff
@@ -0,0 +1,73 @@
...
+import java.awt.Graphics2D;
+import java.awt.Dimension;
+
+import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
+import static net.runelite.client.ui.overlay.OverlayManager.OPTION_CONFIGURE;
+
+public class BlastFurnaceForemanOverlay extends Overlay
+{
+
+	private final Client client;
```
Nightfirecat: Why are you injecting a `Client` here? It doesn't seem to be used.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
```diff
@@ -0,0 +1,73 @@
+package net.runelite.client.plugins.blastfurnace;
```
Nightfirecat: Add license header.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceConfig.java
```diff
@@ -52,4 +52,15 @@ default boolean showBarDispenser()
...
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "showForemanCounter",
+			name = "Show foreman fee counter",
+			description = "If under level 60 smithing, displays time left before requiring payment",
+			position = 3
+	)
+	default boolean showForemanCounter()
```
Nightfirecat: Instead of having a config for this, display the overlay conditionally if the player is under 60 smithing.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Ah, it looks.like this was just shifted by IntelliJ when importing other classes. This is used in other portions of the plug in that are pre-existing.

Nightfirecat: Well, keep in mind that imports should all be in one sorted block without extra newlines separating them.

bwhite94: Resolved in d642d6b

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Resolved in d642d6b

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
```diff
@@ -0,0 +1,73 @@
...
+import net.runelite.client.ui.overlay.OverlayMenuEntry;
+import net.runelite.client.ui.overlay.OverlayPosition;
+import net.runelite.client.ui.overlay.components.LineComponent;
+import net.runelite.client.ui.overlay.components.PanelComponent;
+
+import javax.inject.Inject;
+import java.awt.Graphics2D;
+import java.awt.Dimension;
+
+import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
```
Nightfirecat: Same imports issue in this file

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -50,12 +55,17 @@
 {
 	private static final int BAR_DISPENSER = NULL_9092;
 
+	private static final String FOREMAN_PERMISSION_PATTERN = "Okay, you can use the furnace for ten minutes. Remember, you only need half as much coal as with a regular furnace.";
```
Nightfirecat: Since it's no longer a pattern, use a better var name.

```suggestion
	private static final String FOREMAN_PERMISSION_TEXT = "Okay, you can use the furnace for ten minutes. Remember, you only need half as much coal as with a regular furnace.";
```

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
```diff
@@ -135,4 +154,26 @@ public void onGameStateChanged(GameStateChanged event)
...
+
+		if (foremanCounter != null && foremanCounter > 0)
+		{
+			foremanCounter--;
+		}
+		else if (npcDialog != null)
+		{
+			String npcText = Text.sanitizeMultilineText(npcDialog.getText());
+
+			if (npcText.contains(FOREMAN_PERMISSION_PATTERN))
```
Nightfirecat: Use `.equals()` instead?

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
bwhite94: Now utilized for smithing skill lookup. Left in-tact.

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceConfig.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnacePlugin.java
bwhite94: Resolved in 0fadcde

runelite-client/src/main/java/net/runelite/client/plugins/blastfurnace/BlastFurnaceForemanOverlay.java
```diff
@@ -0,0 +1,97 @@
...
+	{
+		Instant initialForemanFeeInstant = plugin.getInitialForemanFeeInstant();
+		Duration foremanFeeDuration = plugin.getForemanFeeDuration();
+
+		Duration remainingDuration = foremanFeeDuration.minus(Duration.between(initialForemanFeeInstant, Instant.now()));
+
+		long minutes = remainingDuration.toMinutes();
+		long seconds = remainingDuration.getSeconds() % 60;
+
+		return String.format("%01d", minutes) + ":" + String.format("%02d", seconds);
```
Nightfirecat: I believe this can be `String.format("%01d:%02d", minutes, seconds);`

bwhite94:
Should be in pretty good shape now, let me know if you see anything else that could use some tweaking.

Adam-:
I think this would be better as a Timer than an overlay

bwhite94:
> I think this would be better as a Timer than an overlay

Oh wow. I didn't even realize that class existed. Assuming I would add an entry in `GameTimer.java` and then drop the config in `TimersPlugin.java`? Additionally, do you think the Coal bag would be a good image for this? Hover message would be something like "Foreman Fee" or similar for more clarity.

bwhite94:
Refactored to utilize timers per feedback. Removed commits from prior approach to clean up the change history. Much cleaner! Thanks, @Adam- 

The only thing I wasn't sure about a good approach for is `Duration.between(Instant.now(), foremanTimer.getEndTime()).toMinutes() <= 5` which basically just stops checking for the NPC dialog to avoid removing and re-adding the timer until the user closes the dialogue. Something like this was required, to avoid the timer not restarting if paid before completely elapsing.

bwhite94:
Reverted timer plugin changes and extended the timer within the blast furnace plugin per Adam's recommendation in Discord.

PR #8772, opened 2019-05-06
ba: use equipment ItemContainer instead of PlayerComposition

devLotto:
Makes equipment querying consistent across the client by changing the Barbarian Assault plugin to use the equipment ItemContainer instead of the local player's PlayerComposition.


Kamielvf:
This should probably use the item container change event now then?

devLotto:
@Kamielvf 

The IN_GAME_BA var isn't set when the ItemContainerChanged event is received, so the code as it is wouldn't work in an event callback. I could remove that ``client.getVar(Varbits.IN_GAME_BA) == 0`` check though. Maybe that is a decent solution, so we get the event in there atleast instead of polling the container.

devLotto:
Updated to use the item container change event. Tested and works well. 😃

Issue #8776, opened 2019-05-07
AFK notification on Plank Make


CoiNz1337:
Lunar plank make is now AFKable and decent magic xp, adding this to already made plugin "Idle Notifer would be really nice.

PR #8778, opened 2019-05-07
Add idle notifier to plank make

Macweese:


Hydrox6:
Get rid of the first 3 commits

Also, put `Closes #8776` in your first message in this thread

Macweese:
Was not meant

PR #8785, opened 2019-05-07
poison: don't display negative times in tooltip

seandewar:
Some actions can defer the next poison damage tick until after they finish.
When this happens, the poison plugin's tooltip can display a negative time:

This fixes that.


raiyni:
that only puts a bandaid on it

seandewar:
@raiyni Is there anything better?

seandewar:
Closing this -- #7967 already addresses this better.

PR #8786, opened 2019-05-07
xp tracker: display >1m exp with higher precision

devLotto:
Some users believe that the XP Tracker doesn't display experience values with enough precision when above 1m. For example, if you do skilling methods that give you between 1 and 1.1m xp/hr, it will just say 1m/hr when they want it to say e.g. "1.035m/hr".

This pull request changes the XP Tracker to display values over 1 million with higher precision.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
@@ -62,6 +62,14 @@ public class StackFormatter
 	);
 
 	/**
+	 * A more precise decimal number formatter, outputting thousandths
+	 */
+	private static final NumberFormat PRECISE_DECIMAL_FORMATTER = new DecimalFormat(
+		"#,###.###",
+		DecimalFormatSymbols.getInstance(Locale.ENGLISH)
+	);
+
+	/**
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
```diff
@@ -152,6 +160,24 @@ else if (quantity < 10_000_000)
 	 * with K after 100,000 and M after 10,000,000
 	 */
 	public static String quantityToRSDecimalStack(int quantity)
+	{
+		return quantityToRSDecimalStack(quantity, false);
```
Adam-: Is there anything calling this that doesn't want the more precise values? Just looks like the ge offer panel and the ammo counter.

Nightfirecat:
Would it be worth using an integer parameter for number of decimal digits rather than a boolean? I can envision sometime down the line some plugin or feature might want a different number of digits.

runelite-client/src/main/java/net/runelite/client/util/StackFormatter.java
deathbeam: Well, the point of the RS decimal stack method is to format numbers exactly like OSRS formats amounts, so I think having separate method for this is good.

PR #8774, opened 2019-05-07
Add user-defined cursor images

rherms:
Fixes https://github.com/runelite/runelite/issues/8764

To do this, I also added a generic File input and type for configs, which will hopefully be useful in the future.

bb1291a20a83f1ad31f4f11c45624b57

PR #8788, opened 2019-05-08
Allow a custom image cursor

### LINKED ISSUE #8764: Add a user-defined cursor option
I've had some people request the ability to use a cursor besides the ones we provide with the Custom Cursor plugin with RuneLite. Since we're unlikely to expand the cursors we currently provide to images not relating to RuneScape/OSRS, it might be useful to add a file select dialog and a "Custom" drop-down option to allow users to specify a cursor file from their own computers to use with the plugin, since the alternative is making them build and maintain a custom build of their own just a simple change like this.

rherms:
Fixes #8764 

Looks like we have no choice put to force users to put their image file in the runelite directory to avoid breaks between machines. Alternative implementation to #8774 and simpler because it just adds another dropdown choice instead of a file chooser. Result is the same though in terms of having my own crab cursor. 🦀

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursor.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursor.java
@@ -25,9 +25,11 @@
 package net.runelite.client.plugins.customcursor;
 
 import java.awt.image.BufferedImage;
+import javax.annotation.Nullable;
 import lombok.Getter;
 import net.runelite.client.util.ImageUtil;
 
+@Getter
 public enum CustomCursor
 {
 	RS3_GOLD("RS3 Gold", "cursor-rs3-gold.png"),
@@ -35,21 +37,22 @@ public enum CustomCursor
 	DRAGON_DAGGER("Dragon Dagger", "cursor-dragon-dagger.png"),
 	DRAGON_DAGGER_POISON("Dragon Dagger (p)", "cursor-dragon-dagger-p.png"),
 	TROUT("Trout", "cursor-trout.png"),
-	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png");
+	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png"),
+	CUSTOM_IMAGE("Custom Image");
 
 	private final String name;
-	@Getter
+	@Nullable
 	private final BufferedImage cursorImage;
+
+	CustomCursor(String name)
+	{
+		this.name = name;
+		this.cursorImage = null;
+	}
 
 	CustomCursor(String name, String icon)
 	{
 		this.name = name;
 		this.cursorImage = ImageUtil.getResourceStreamFromClass(CustomCursorPlugin.class, icon);
-	}
-
-	@Override
-	public String toString()
-	{
-		return name;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
@@ -25,10 +25,15 @@
 package net.runelite.client.plugins.customcursor;
 
 import com.google.inject.Provides;
+import java.awt.image.BufferedImage;
+import java.io.File;
+import javax.imageio.ImageIO;
 import javax.inject.Inject;
-import net.runelite.client.events.ConfigChanged;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.client.RuneLite;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.events.ConfigChanged;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.ClientUI;
@@ -38,8 +43,11 @@ import net.runelite.client.ui.ClientUI;
 	description = "Replaces your mouse cursor image",
 	enabledByDefault = false
 )
+@Slf4j
 public class CustomCursorPlugin extends Plugin
 {
+	private static final File CUSTOM_IMAGE_FILE = new File(RuneLite.RUNELITE_DIR, "cursor.png");
+
 	@Inject
 	private ClientUI clientUI;
 
@@ -76,6 +84,34 @@ public class CustomCursorPlugin extends Plugin
 	private void updateCursor()
 	{
 		CustomCursor selectedCursor = config.selectedCursor();
-		clientUI.setCursor(selectedCursor.getCursorImage(), selectedCursor.toString());
+
+		if (selectedCursor == CustomCursor.CUSTOM_IMAGE)
+		{
+			if (CUSTOM_IMAGE_FILE.exists())
+			{
+				try
+				{
+					BufferedImage image;
+					synchronized (ImageIO.class)
+					{
+						image = ImageIO.read(CUSTOM_IMAGE_FILE);
+					}
+					clientUI.setCursor(image, selectedCursor.getName());
+				}
+				catch (Exception e)
+				{
+					log.error("error setting custom cursor", e);
+					clientUI.resetCursor();
+				}
+			}
+			else
+			{
+				clientUI.resetCursor();
+			}
+			return;
+		}
+
+		assert selectedCursor.getCursorImage() != null;
+		clientUI.setCursor(selectedCursor.getCursorImage(), selectedCursor.getName());
 	}
 }


runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,24 @@ public void onConfigChanged(ConfigChanged event)
 	private void updateCursor()
 	{
 		CustomCursor selectedCursor = config.selectedCursor();
+		if (selectedCursor.equals(CustomCursor.CUSTOM_IMAGE))
```
raiyni: use `==`

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -25,14 +25,20 @@
 package net.runelite.client.plugins.customcursor;
 
 import com.google.inject.Provides;
+
+import javax.imageio.ImageIO;
```
deathbeam: Fix your imports

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
deathbeam: add empty line above

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,24 @@ public void onConfigChanged(ConfigChanged event)
...
+					BufferedImage image = ImageIO.read(file);
+					clientUI.setCursor(image, "Custom");
+					return;
+				}
+				catch (Exception e)
+				{
+					// fall through to setting the default cursor
+				}
+			}
+		}
```
deathbeam: add empty line below

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,24 @@ public void onConfigChanged(ConfigChanged event)
 	private void updateCursor()
 	{
 		CustomCursor selectedCursor = config.selectedCursor();
+		if (selectedCursor.equals(CustomCursor.CUSTOM_IMAGE))
+		{
+			// Try to load the user cursor from ~/.runelite/cursor.png
+			File file = new File(RuneLite.RUNELITE_DIR, "cursor.png");
```
deathbeam: final. this can be also static final constant at top

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,24 @@ public void onConfigChanged(ConfigChanged event)
...
+			if (file.exists())
```
deathbeam: empty line above

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -25,14 +25,20 @@
...
 import javax.inject.Inject;
 import net.runelite.api.events.ConfigChanged;
+import net.runelite.client.RuneLite;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.ClientUI;
 
+import java.awt.image.BufferedImage;
```
deathbeam: fix imports

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursor.java
```diff
@@ -34,7 +34,9 @@
 	RS3_SILVER("RS3 Silver", "cursor-rs3-silver.png"),
 	DRAGON_DAGGER("Dragon Dagger", "cursor-dragon-dagger.png"),
 	TROUT("Trout", "cursor-trout.png"),
-	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png");
+	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png"),
+	// will fall back to RS3 Gold if anything goes wrong
+	CUSTOM_IMAGE("Custom Image", "cursor-rs3-gold.png");
```
deathbeam: this should just have empty string

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursor.java
```diff
@@ -36,7 +36,7 @@
 	TROUT("Trout", "cursor-trout.png"),
 	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png"),
 	// will fall back to RS3 Gold if anything goes wrong
-	CUSTOM_IMAGE("Custom Image", "cursor-rs3-gold.png");
+	CUSTOM_IMAGE("", "cursor-rs3-gold.png");
```
Nightfirecat: The name should still be "Custom Image". It's the filename which should be an empty string.

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursor.java
```diff
@@ -36,7 +36,7 @@
 	TROUT("Trout", "cursor-trout.png"),
 	DRAGON_SCIMITAR("Dragon Scimitar", "cursor-dragon-scimitar.png"),
 	// will fall back to RS3 Gold if anything goes wrong
```
Nightfirecat: Accordingly, remove this comment.

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -46,6 +50,8 @@
 	@Inject
 	private CustomCursorConfig config;
 
+	private static final File customImageFile = new File(RuneLite.RUNELITE_DIR, "cursor.png");
```
deathbeam: static fields belong at top of the class, above instance fields

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,27 @@ public void onConfigChanged(ConfigChanged event)
...
+			{
+				try
+				{
+					BufferedImage image = ImageIO.read(file);
+					clientUI.setCursor(image, "Custom");
+					return;
+				}
+				catch (Exception e)
+				{
+					// fall through to setting the default cursor
```
Nightfirecat: `log.error()` the exception.

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -76,6 +82,27 @@ public void onConfigChanged(ConfigChanged event)
 	private void updateCursor()
 	{
 		CustomCursor selectedCursor = config.selectedCursor();
+
+		if (selectedCursor == CustomCursor.CUSTOM_IMAGE)
+		{
+			// Try to load the user cursor from ~/.runelite/cursor.png
+			File file = new File(RuneLite.RUNELITE_DIR, "cursor.png");
+
+			if (file.exists())
```
Nightfirecat: If the file doesn't exist, this should log that the file could not be found.

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -44,7 +46,9 @@
 )
 public class CustomCursorPlugin extends Plugin
 {
-	private static final File customImageFile = new File(RuneLite.RUNELITE_DIR, "cursor.png");
+	private static final Logger logger = LoggerFactory.getLogger(CustomCursorPlugin.class);
```
Nightfirecat: Annotate the class with `@Slf4j` and use the provided `log` object instead of this.

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
```diff
@@ -85,22 +89,24 @@ private void updateCursor()
 
 		if (selectedCursor == CustomCursor.CUSTOM_IMAGE)
 		{
-			// Try to load the user cursor from ~/.runelite/cursor.png
-			File file = new File(RuneLite.RUNELITE_DIR, "cursor.png");
```
Nightfirecat: I don't have strong feelings either way, but it seems like trying to load the custom cursor each time the cursor is updated is a good approach, rather than locking it in upon plugin initialization.

Adam-:
I don't think the average user would be capable of putting a cursor image into `~/.runelite`. What about being up a file selection dialog?

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorPlugin.java
deathbeam: This is not loading cursor, just creating File instance that is just reference to path

deathbeam:
That is not portable

rherms:
> I don't think the average user would be capable of putting a cursor image into `~/.runelite`. What about being up a file selection dialog?

That is what I first did at https://github.com/runelite/runelite/pull/8774

Adam-:
How hard would it be to store the cursor image in the config?

rherms:
Do we do this anywhere else currently? I imagine base64 encoding the image probably won't be too bad, but I am worried about what would happen if the user uploads some bogus or really large file.

Nightfirecat:
Given that #10237 has users place `notification.wav` into ~/.runelite, I feel we can have this plugin ask for the same.

Adam-:
I suppose so.... if you want to pick this up feel free

rherms:
In that case, I'll probably just reopen https://github.com/runelite/runelite/pull/8774 and touch it up.

rherms:
This is rebased and ready for another look.

PR #8789, opened 2019-05-08
regenmeter: notif before next hp regen & refactor existing code to use Constants.GAME_TICK_LENGTH

### LINKED ISSUE #2704: Add Health Regeneration Notification
NMZ suggestion: Making a notification appear whenever your health is about regenerate at a certain point such as at this point in the regeneration process. 
hp sugg

seandewar:
Adds a configuration option for notifying the player before their next HP regen occurs, and refactors existing code to utilise `Constants.GAME_TICK_LENGTH` where approperiate.

The following notification is sent:

Should be especially useful for NMZ. Its accuracy is limited to that of the Regeneration Meter plugin itself.
Closes #2704.

**Note:** the config accepts the amount of milliseconds before the regen to send the notification, but this is instead converted to the earliest corresponding 0.6s tick interval in the code. Maybe it would be a better idea to accept something like seconds instead?

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
@@ -57,4 +57,14 @@ public interface RegenMeterConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "notifyBeforeHpRegenDuration",
+		name = "Hitpoint Regen Notification (seconds)",
+		description = "Notify approximately when your next hitpoint is about to regen. A value of 0 will disable notification."
+	)
+	default int getNotifyBeforeHpRegenSeconds()
+	{
+		return 0;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2018 Abex
+ * Copyright (c) 2019, Sean Dewar <https://github.com/seandewar>
+ * Copyright (c) 2018, Abex
  * Copyright (c) 2018, Zimaya <https://github.com/Zimaya>
  * Copyright (c) 2017, Adam <Adam@sigterm.info>
  * All rights reserved.
@@ -30,6 +31,7 @@ import com.google.inject.Provides;
 import javax.inject.Inject;
 import lombok.Getter;
 import net.runelite.api.Client;
+import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.api.Prayer;
 import net.runelite.api.Skill;
@@ -37,6 +39,7 @@ import net.runelite.api.VarPlayer;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.VarbitChanged;
+import net.runelite.client.Notifier;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.plugins.Plugin;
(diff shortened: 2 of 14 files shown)


seandewar:
After a small discussion on the discord, I think it's better for the config option to accept seconds instead of milliseconds. I've modified the PR to reflect that change.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterConfig.java
```diff
@@ -57,4 +57,13 @@ default boolean showWhenNoChange()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "notifyBeforeHpRegenDuration",
+		name = "Hitpoint Regen Notification (seconds)",
+		description = "Notify approximately when your next hitpoint is about to regen. A value of 0 will disable notification.")
```
Adam-:
```suggestion
		description = "Notify approximately when your next hitpoint is about to regen. A value of 0 will disable notification."
	)
```

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -149,5 +155,18 @@ else if (currentHP > maxHP)
...
+		{
+			notifier.notify("[" + client.getLocalPlayer().getName() + "] regenerates their next hitpoint soon!");
+		}
+	}
+
+	private boolean shouldNotifyHpRegenThisTick(int ticksPerHPRegen)
+	{
+		// if the configured duration lies between two ticks, choose the earlier tick
+		final int ticksBeforeHPRegen = ticksPerHPRegen - ticksSinceHPRegen;
+		final int notifyTick = (int) Math.ceil(config.getNotifyBeforeHpRegenSeconds() / HP_REGEN_TICK_LENGTH);
```
deathbeam: This is basically game tick, don't we have constant for this somewhere? If no, then this is good opportunity to add it to Constants file

seandewar: Yeah, I noticed this. I wasn't sure whether to add it seeing that a few plugins already define their own version of the 0.6s tick constant in a few places.

If I do this, should I also update the other plugins to use the const from Constants within this PR? I could add that change as a seperate commit to this one and edit the name of the PR to reflect this.

Nightfirecat: A separate commit would be good, but that sounds like a good idea.

seandewar:
Latest revision adds `GAME_TICK_LENGTH` to Constants.java and refactors existing code to refer to it instead of redefining the same value in multiple places.

I've also used `GAME_TICK_LENGTH` in places that I think seem to somewhat loosely refer to the 600ms game tick interval (including Hooks.java, whose main loop does some work every 600 ms -- I could use some feedback on whether that change is approperiate).

Other places, such as BlastMineRock.java's `FUSE_TIME` constant (=4.2s; exactly 7 game ticks), I've decided to leave alone, as these seem fine enough.

seandewar:
@Adam- You want me to _flip_ the commits around?

Adam-:
Yes

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -30,6 +30,7 @@
 import javax.inject.Inject;
 import lombok.Getter;
 import net.runelite.api.Client;
+import net.runelite.api.Constants;
```
Adam-: ?

seandewar: 🤦 whoops

Adam-:
Better commit messages might be:

```
constants: add GAME_TICK_LENGTH and use where needed

Also update many usages of 20ms to CLIENT_TICK_LENGTH 
```

because 1) it is implied the commit modifies and codebase and 2) GAME_TICK_LENGTH is actually being added by the commit, not simply being used

and

`regenmeter: add option to notify before next hp regen`

PR #4044, opened 2018-06-27
Simple BEDMAS Calculator in Chat (Re-Opened)

IronApeiron:
**Note: Opened a new PR due to rebasing issues on my end. Had to re-fork to fix. Old PR here: https://github.com/runelite/runelite/pull/4016**

This is an attempt to close #3967. I implemented a simple BEDMAS calculator used via a chat command.  I feel like it's a "nice to have" feature. It is used in chat like the following:

`!calc <input expression with other sub-expressions involving math operations>`

As per feedback, this now uses exp4j's implementation. From https://www.objecthunter.net/exp4j/apidocs/index.html: 

Supported Operations:
> Addition: '2 + 2'
> Subtraction: '2 - 2'
> Multiplication: '2 * 2'
> Division: '2 / 2'
> Exponential: '2 ^ 2'
> Unary Minus,Plus (Sign Operators): '+2 - (-2)'
> Modulo: '2 % 2'
> the following functions are supported:
> 
> abs: absolute value
> acos: arc cosine
> asin: arc sine
> atan: arc tangent
> cbrt: cubic root
> ceil: nearest upper integer
> cos: cosine
> cosh: hyperbolic cosine
> exp: euler's number raised to the power (e^x)
> floor: nearest lower integer
> log: logarithmus naturalis (base e)
> log2: logarithm to base 2
> log10: logarithm to base 10
> sin: sine
> sinh: hyperbolic sine
> sqrt: square root
> tan: tangent
> tanh: hyperbolic tangent
> signum: signum of a value

The result is `Calc: <input expression> = <input expression evaluation>` if successful or `Calc: Unable to calculate expression: <input expression>` which may happen if the input expressions have unsupported symbols.

For example, `!calc (1+(2^2)/3.5)` yields `Calc: (1+(2^2)/3.5) = 2.142...` (all digits will be printed out in double precision).

Examples:
capture

Please note I'm new to the API so I may have missed something, or I might have made some logical errors. Perhaps there are better ways of optimizing my code. I'm open to suggestions!

PR #8792, opened 2019-05-09
!math added

JZomerlei:
Using !math equationAsString converts to a double input.

Displaying !math (4-2)^2 as PlayerName: (4-2)^2=4.0


JZomerlei:
Current issue is ((5)4) works but (4(5)) does not even though they both equal 20. Maybe some one can help me figure out a smart way of dealing with that issue, other than that it works quite well.

xdesr:
duplicate #4044

Nightfirecat:
Issues like the one you mentioned are why I feel rolling our own math library (because that's basically what this is) would be a bad idea. #4044 pulls in a trusted outside library to do it, which would be much preferred.

Issue #8618, opened 2019-04-22
Item charge overlay number is at wrong offset when in bank


Nightfirecat:
Comparing these two images:
![item-charge-wrong-offset-in-bank](https://user-images.githubusercontent.com/2199511/56539077-6359d280-6554-11e9-9003-dcf0cd4bd69f.png) ![item-charge-wrong-offset-outside-bank](https://user-images.githubusercontent.com/2199511/56539078-6359d280-6554-11e9-8900-7a6d3fce192c.png)

And this image, which shows only the games necklace in these two images overlaid:
![item-charge-wrong-offset-overlaid](https://user-images.githubusercontent.com/2199511/56539079-6359d280-6554-11e9-94e2-69483fe06a60.png)

The item charges number displayed in the inventory while the bank is open (in fixed viewport, not sure about others) is one pixel too far down and right than it is meant to be. This is reproduceable on master since the ItemWidget overlay change.

Nightfirecat:
And an animated gif showing the issue:
![item-charges-position-issue-animated](https://user-images.githubusercontent.com/2199511/56540312-e03a7b80-6557-11e9-96c3-99d888f25de4.gif)

xdesr:
This issue occurs with all recently updated Widget item overlays : rune pouch, slayer helm , item charges and inventory tags, when in the bank interface the itemWidgetCanvasBounds x,y  increase by 1 pixel

PR #8799, opened 2019-05-10
widget item overlay: Fix bank inventory widget bound position

Nightfirecat:
The bounds for item widgets within the bank inventory group were all one
pixel too far down and to the right--that is, having one too high x and
y values in their bounds. This patch fixes the bounds for such item
widgets.

Fixes runelite/runelite#8618


deathbeam:
Shouldn't this be fixed when the bounds are calculated? Also, why is this happening

xdesr:
This only fixes it for the bank inventory, the other interfaces (shop inventory, G.E. inventory etc..)  still have the issue

Nightfirecat:
Oh, wasn't aware this was an issue for other inventories too. I'll close this then, since there's likely a better fix for this problem.

PR #8800, opened 2019-05-10
Add loop timer infobox

deathbeam:
Add loop timer infobox that will keep looping through specified
interval.


deathbeam:
Updated:

Replaced BarrowsPrayerDrainTimer with LoopTimer

Issue #3894, opened 2018-06-19
DPS Meter (Similar to WoW's Recount)


shermaza:
**Describe the solution you'd like**
Add a plugin panel similar to WoW's _recount_ to track damage done per player in your clan chat. This can be reset via a plugin-panel.

**Describe alternatives you've considered**
Instead of clan chat, it could be by raid group as that's most likely where it would be used. However, with the clan chat, other group PvM like GWD or corp could be included.

**Additional context**
 Image examples from Recount's web page https://www.wowace.com/projects/recount/images - the latter screens are WoW specific, but the first one shows each player where their "bar" is a % full based on DPS, and their info is displayed as "username totalDamage (damagePerSecond %percentDamageDone)"

Jalopyy:
Only if we can add gearscore too

kulershk:
I dont think so it's possible, do jagex say who did how much damage? I think we only have hitsplat event who took damage not who dealt.

DeadeyeOo:
I start this already and used the constitution EXP drop to determine the damage of your next attack. Then calculate DPS based on how long your have been in combat. If you want I could clean it up and submit for review. 
![untitled](https://user-images.githubusercontent.com/37055804/41640262-f753410a-7415-11e8-9228-932f5b4e35ae.png)

shermaza:
I was watching @WooxSolo's stream today, it seems like he has a prototype for this on his client. He may have more input about how feasible it is for a full plugin.

CustardGod:
I believe Woox's version is a total damage plugin, not DPS. Granted I doubt it would be hard to change it to a DPS plugin

DeadeyeOo:
Oh yea I just checked it out in his stream, I do not know how he is calculating dmg for nearby players. Pretty awesome actually.
![capture](https://user-images.githubusercontent.com/37055804/41685161-77270f82-7494-11e8-9644-9b6a26a7d821.PNG)

CustardGod:
I believe he's re-written the animation api to allow him to find who does what damaged based on when their animation goes off and when the hitsplat shows up. Decimals because multiple hitsplats can show at the same time so the damage is averaged out between the players who did the damage

WooxSolo:
I do have a prototype for this yes. It works on all players nearby by waiting for the attack animation and then calculating the amount of ticks until the damage should appear and then counting the damage from the hitsplat. It is only an estimator for now though, since there is a problem with this approach.

[...]

PR #8803, opened 2019-05-11
ToB Damage Counter

akarhi:
The plugin counts the damage you have done and will add in the extra healing the boss have done to itself. After the fight or after your death it will print out a message the amount of damage you have done in % form. Along with that it has an option to change the raid party setup to 3 to 5 man and few chat message to appear or not. The default value is 4 (group members in total) since its the most common group to do

Might include the total amount of damage you have done to the boss but will be set as "false" as a default value

Healing; Maiden is the only boss that is true healing, does not go off percent base. The other two bosses does go off percentage healing.

Output:
image

Options:
image


Adam-:
I like the idea of a damage counter (this is basically #3894). However I don't like this approach. Specifically:

1) this should use the party system to share damages between clients
2) this should be able to easily support other things in the future (probably just cox?)
3) this should support any size tob party, not just 5 man

akarhi:
1. Which party system the runelite function one or the ingame one?
2. Should be supportable between raids
3. It should be supportable by any size, currently figuring out to grab the party system and making it automatic

pvmgod123:
Can anyone use this?

Adam-:
The RuneLite one. I made a start at https://github.com/Adam-/runelite/tree/dps which works for simple bosses like corp.

TheMursk:
I tested this and it's pretty neat. One thing you haven't taken into account is that Verzik is also healing when spawning crabs and during blood phaze. Maiden aswell ofc.

akarhi:
I have accounted the healing for all the bosses. Maiden is the actually amount, nylo and verzik is percentage base

PR #8805, opened 2019-05-11
log notifications

VincentJ711:
This adds a Runelite config option for logging notifications. My use case for this is handling the notifications outside of Runelite. I want to be able to run a NodeJS script that watches for certain notifications and performs actions based on them. Ie on log file change, if the notification i was looking for was added, perform XYZ.  Currently, at least on OSX, its very tricky finding where notifications are stored in the filesystem. This solves that problem.


deathbeam:
Remove the configuration option, and just use debug logging

deathbeam:
https://github.com/runelite/runelite/wiki/Code-Conventions also follow this

Issue #1464, opened 2018-04-12
Emojis


sadegr:
![konduit emojis](https://user-images.githubusercontent.com/20391979/38709437-06a1d710-3e89-11e8-9fbe-aa40fd174803.png)
This request is for an optional (togglable) plugin that allows players to see emojis sent by other players in OSRS chat. Konduit (another 3rd party OSRS client) already provides this feature (the screenshot shown above is from Konduit).

Players would be able to type Emoji shortcodes (as used on GitHub, Discord, etc. - see [here](https://gist.github.com/oliveratgithub/0bf11a9aff0d6da7b46f1490f86a71eb)) in public, private, and clan chat to cause the corresponding emoji to show up in the in-game chatbox for players with the emoji plugin enabled. The normal text would show up unmodified to players with the emoji plugin disabled. Emoji shortcodes could be typed by themselves or as part of a larger text string to cause emojis to show up, and multiple emojis could be used per line of text.

Example: typing `:ok_hand:` in public chat would cause the 👌 emoji to show instead of the `:ok_hand:` text to players with the emoji plugin enabled. To players with the emoji plugin disabled, the text itself (i.e. `:ok_hand:`) would show up as originally typed.

sethtroll:
If someone wanted to make the plugin and it was toggled off by default I would be okay with that. OSBuddy and Konduit both have this (I think) and we've always tried to keep the competition even in terms of plugins.

sanderdegroot:
I've attached a screenshot of a few Konduit emojis (it wouldn't have to be the same set of icons) along with the text that was typed underneath for reference.

GETrackerDan:
I've given it a thumbs up although I'm not a big fan of this in game and it just reminds me of private servers and a bit trashy if I'm honest. If somebody who has it toggled off doesn't see the emoji or :ok_hand: code/text and as @sethtroll has said toggled off as default I don't see an issue with it being built and added into the client. I'm sure there will be many people who would like this.

MESLewis:
How would the images be stored in the client? ATM we use mostly cache images, I am not sure if there is precedent set for where other images go.

sanderdegroot:
@MESLewis I would imagine a new folder in runelite-client/src/main/resources/ would be a pretty good place for them.

deathbeam:
[...]

PR #8811, opened 2019-05-12
Emojis

devLotto:
Disabled by default, this plugin replaces emoticons with their corresponding emoji image in the chat, in line with what has been discussed in #1464. The triggers must be separated from other words in a chat line for them to be replaced, and the triggers are case sensitive so people with the plugin off see emoticons with the correct capitalization.

| Emoji | Trigger | Emoji | Trigger | Emoji | Trigger |
| ------------- | ------------- |     :---:      |     :---:      |          ---: | ------------- |
| 🙂  | :)  | 😂 | =') | 🤠 | 3:) |
| 😊 | ^\_^ | 😀 | :D | 😄 | =D |
| 😉 | ;) | 😝 | X-P | 😛 | :P |
| 😋 | =P~ | 🤗 | >:D< | 😤 | :> |
| 🤔 | :‑? | 😕 | :/ | 😐 | =\| |
| 😑 | :\| | 😒 | :-\| | 🙁 | :( |
| ☹️ | =( | 😢 | :'( | 😭 | :\_(  |
| 😳 | :$ | 🤐 | :-# | 😣 | >_< |
| 😎 | 8-) | 😇 | O:) | 😈 | >:) |
| 😡 | >:( | 😯 | :‑O | 😮 | :O |
| 😱 | :-@ | 🙈 | X_X | 💃 | \\:D/ |
| 👌 | (Ok) | 👍 | (Y) | 👎 | (N) |
| ❤ | <3 | 💔 | </3 | 💤 | Zzz |
| 🐟 | <>< | 😸 | :3 | 🐶 | =3 |
| 🦀 | V(;,;)V | 🍴 | --E | 🍳 | --(o) |
| 🎉 | @@@ | 💦 | ;; | 👀 | O.O |
| 💩 | \~@\~ |

image


deathbeam:
What about adding `o.o` or `O.O` for :eyes:

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -0,0 +1,192 @@
...
+import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.OverheadTextChanged;
+import net.runelite.client.chat.ChatMessageManager;
+import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.util.ImageUtil;
+
+@PluginDescriptor(
+	name = "Emojis",
```
deathbeam: I would rename this to Emoticons instead of Emojis, as apparently Emojis are that `:things_like_this:` or something, idk

ShaunDreclin:
`;;` for 💦 would work too, it's usually paired with emotes like `^_^;;`

Just need something for 💩 haha

Being disabled by default seems a bit iffy though, a plugin like this isn't very useful if nobody else has it turned on. I think more people would want it enabled than not want it (not counting people who don't care either way)

deathbeam:
`:swing:` fit's the :hankey: emoji perfectly, imo

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
Adam-: Emojis are the images, Emoticons are text like `:-)`. This seems fine as is?

ShaunDreclin: Emoticons also cover the images, that's what MSN Messenger called them. This isn't really "emoji" support since it's a limited selection of them and they're just PNGs instead of being handled like Unicode emojis are

On the other hand it *is* using images of emoji rather than making our own emoticons, so it could go either way. 🤷‍♂️ (Would we want to make our own? Might be neat)

deathbeam: I don't really care, but, on the issue I was calling them emojis and someone told me they are not emojis :man_shrugging:

devLotto: Emoticons = character representation
Emoji = image representation

The plugin turns emoticons into emoji, which is why it should be called Emojis.

devLotto:
Added 💦, 👀 and 💩

devLotto:
@ShaunDreclin Having the chat be replaced with emojis might be a bit overwhelming for a new user of the client if they don't like them. I think having the plugin disabled by default is fine, because people who want it will likely know about its existence and enable it themselves.

PR #8813, opened 2019-05-12
add subdirectories to screenshots

### LINKED ISSUE #8711: Screenshot Subfolders
After playing for just over a month, I have over 2,300 screenshots in my screenshot folder. Windows is beginning to struggle to load the folder, and actually finding a given screenshot is an absolute nightmare.

The screenshots should be organised into subfolders. 
The actual subfolders themselves are fairly easy to decide, simply separate each of the selectable categories in the plugin into their own respective folders.

- Manual Screenshots
- Rewards
- Levels
- Kingdom Reward
- Pet
- PvP Kills
- Boss Kills
- Deaths
- Duels
- Valuable Drops
- Untradable Drops

Levels could also potentially be divided into further subfolders for each skill, since theoretically a maxed character could have 2245 screenshots in their levels folder. 

This would not only make searching for a specific screenshot much easier, it would massively reduce the loading time of the folder and the indexing when viewing it on windows.

You could do this manually, but that's inconvenient

gregg1494:
Closes #8711
Created directories for:
Manual Screenshots
Clue Scroll Rewards
Levels
Kingdom Rewards
Pets
PvP Kills
Boss Kills
Deaths
Duels
Valuable Drops
Untradable Drops
Quests

I was able to take actual screenshots for some of the basic ones like Levels, Deaths, Duels, and Manual Screenshots as those were easiest to do in game, and they worked correctly. I used ScreenshotPluginTest and check the return values for my screenShotSubDir variable and they are correct. I was not able to have the test take actual screen shots though.  

If someone knows how to actually take a screen shot when using the test calls let me know.


runelite-client/src/test/java/net/runelite/client/plugins/screenshot/ScreenshotPluginTest.java
```diff
@@ -113,8 +111,14 @@ public void testClueScroll()
...
 		screenshotPlugin.onChatMessage(chatMessageEvent);
 
-		assertEquals("medium", screenshotPlugin.getClueType());
-		assertEquals(28, screenshotPlugin.getClueNumber());
+		WidgetLoaded event = new WidgetLoaded();
+		event.setGroupId(CLUE_SCROLL_REWARD_GROUP_ID);
+		screenshotPlugin.onWidgetLoaded(event);
+
+
+		//assertEquals("medium", screenshotPlugin.getClueType());
```
Nightfirecat: Why are you commenting out old tests? This is not the right way to change the test suite for this plugin.

gregg1494: I commented it out to test if my implementation was correct. Didn't see the need to create a more tests when I could just check my value from there. I initially didn't comment it out but didn't know if having two assertEquals would mess up the check.

Nightfirecat: You can have any number of JUnit assertions in a single test case. The only important thing is that none of them fail.

gregg1494: Ok. Thanks for clarifying.

runelite-client/src/test/java/net/runelite/client/plugins/screenshot/ScreenshotPluginTest.java
```diff
@@ -36,8 +36,8 @@
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
-import static net.runelite.api.widgets.WidgetID.DIALOG_SPRITE_GROUP_ID;
-import static net.runelite.api.widgets.WidgetID.LEVEL_UP_GROUP_ID;
+
```
Nightfirecat: Revert the changes to this file.

runelite-client/src/test/java/net/runelite/client/plugins/screenshot/ScreenshotPluginTest.java
```diff
@@ -36,8 +36,7 @@
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
-import static net.runelite.api.widgets.WidgetID.DIALOG_SPRITE_GROUP_ID;
-import static net.runelite.api.widgets.WidgetID.LEVEL_UP_GROUP_ID;
+import static net.runelite.api.widgets.WidgetID.*;
```
Nightfirecat: These changes still aren't reverted. Don't use star imports here. The full diff should show no differences in this test file.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -751,4 +776,10 @@ int gettheatreOfBloodNumber()
 	{
 		return theatreOfBloodNumber;
 	}
+
+	@VisibleForTesting
+	String getScreenShotSubDir()
```
Nightfirecat: I guess this can be removed since it's no longer tested?

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -130,6 +130,8 @@
 		"You feel something weird sneaking into your backpack",
 		"You have a funny feeling like you would have been followed");
 
+	private String screenShotSubDir = "Manual Screenshots";
```
Nightfirecat: Instead of maintaining a field for the screenshot subdirectory, add a new signature to `takeScreenshot` which accepts a filename, subdirectory, and image. That way, these extra `screenShotSubDir = "Manual Screenshots";` resets can be removed and screenshot function calls can be more declarative.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
gregg1494: Ya that is no longer need. I will remove it.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
gregg1494: I still need a variable to track the the sub directory, are you suggesting just make it local to the function in which the subdirectory is created? I couldn't just pass the subdirectory name (for reasons like in the code below) where the takeScreenshot is called after a switch statement so I would need store the subdirectory based on the switch case.

https://github.com/runelite/runelite/blob/150d8722a889df943a5e3f7bc9811e863dc60d87/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java#L517-L555

Nightfirecat: I agree, a variable will be needed for the `GameTick` and `WidgetLoaded` subscribers, however this doesn't need to be a class-wide field.

gregg1494: Ok I will see what I can do.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -263,27 +263,31 @@ public void onGameTick(GameTick event)
 			return;
 		}
 
+		String screenshotSubDir = "";
```
Nightfirecat: Set this to `null`, and null-check it along with fileName before calling `takeScreenshot()`.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -563,8 +573,9 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
 	 * and optionally uploads it to an image-hosting service.
 	 *
 	 * @param fileName    Filename to use, without file extension.
+     * @param subDir    Sub directory where the file will be put.
```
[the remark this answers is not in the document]
gregg1494: I think I fixed it. I cant tell my check style is saying its good.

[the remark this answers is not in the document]
Nightfirecat: It's not fixed, it's still 5 spaces instead of `\t `.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -563,8 +573,9 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
 	 * and optionally uploads it to an image-hosting service.
 	 *
 	 * @param fileName    Filename to use, without file extension.
+	 * @param subDir    Sub directory where the file will be put.
```
Nightfirecat: To line up with the javadoc above it.
```suggestion
	 * @param subDir      Sub directory where the file will be put.
```

renfc01:
Hi, is this still being worked on?

PR #8815, opened 2019-05-12
KitType addendum

ollie-d:
Previous KitType.java started with CAPE(1) when there is indeed a neglected HELM(0). Furthermore, HEAD and JAW seem to always return -1, I'm not sure what their functionality is.


Kyleeld:
You’re also missing 
RING(12),
AMMUNITION(13);

I would also suggest rebasing and resubmitting a new Pr

ollie-d:
> 
> 
> You’re also missing
> RING(12),
> AMMUNITION(13);

KitType are intended for use on other players, which you can't grab RING and AMMO from. That's handled: https://static.runelite.net/api/runelite-api/net/runelite/api/EquipmentInventorySlot.html

I'm still not sure what the purpose of JAW and HEAD are, or if my naming index (0) HELM is appropriate. I haven't contributed to any public project before, sorry if I'm going about this incorrectly.

PR #8816, opened 2019-05-12
Format config enums as "Config Option" by default

devLotto:
Makes it so enum options that don't override toString already don't shout at you anymore.

For example, CONFIG becomes Config and CONFIG_OPTION becomes Config Option.

You can still override toString on the enums if you want the config option to display apostrophes for example.

**Before:**

image

**After:**

image


deathbeam:
Can you update this also for latest XP tracker config addition?

devLotto:
Updated

Adam-:
I really don't think this is solving an actual problem, and is introducing additional complexity for no reason.

deathbeam:
I dont think there was any problem in first place, this is basically just cleanup

PR #8822, opened 2019-05-12
Mining plugin

Barragek0:
Shows progress indicators for respawn times on all rocks that can be mined (including motherlode & amethyst)
Wq4DFma

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
@@ -1,0 +1,501 @@
+/*
+ * Copyright (c) 2017, Robin Weymans <Robin.weymans@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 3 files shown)


Alexgopen:
I see that you are hardcoding the respawn time for each ore, but isn't the respawn time variable upon server population? Or was that changed at some point?

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -0,0 +1,499 @@
...
+		}
+	}
+
+	@Getter(AccessLevel.PACKAGE)
+	private ArrayList<Obj> objects = new ArrayList<Obj>();
+	@Getter(AccessLevel.PACKAGE)
+	private ArrayList<WallObj> wallObjects = new ArrayList<WallObj>();
+
+	@Getter(AccessLevel.PACKAGE)
+	private String currentNode = "";
```
Adam-: this would be better served with an enum

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -0,0 +1,499 @@
...
+	public void onGameObjectDespawned(GameObjectDespawned event)
+	{
+		final GameObject object = event.getGameObject();
+		final Player player = client.getLocalPlayer();
+
+		if (object.getWorldLocation().distanceTo(player.getWorldLocation()) > 20)
+		{
+			return;
+		}
+		objects = removeDuplicates(objects);
```
Adam-: What is this for? It doesn't look like it can happen when each time you add a new Obj instance.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -0,0 +1,499 @@
...
+		if (isDepleted(object))
+		{
+			Iterator<Obj> iterator = objects.stream().distinct().iterator();
+			while (iterator.hasNext())
+			{
+				Obj obj = iterator.next();
+				GameObject o = obj.getGameObject();
+				if (object.getX() == o.getX() && object.getY() == o.getY() && object.getPlane() == o.getPlane())
+				{
+					objects.set(objects.indexOf(obj), new Obj(object, System.currentTimeMillis(), 0));
```
Adam-: Maybe you really want a `Map`?

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -0,0 +1,499 @@
...
+		else if (isRock(object))
+		{
+			Iterator<WallObj> iterator = wallObjects.stream().distinct().iterator();
+			while (iterator.hasNext())
+			{
+				WallObj obj = iterator.next();
+				WallObject o = obj.getWallObject();
+				if (object.getX() == o.getX() && object.getY() == o.getY() && object.getPlane() == o.getPlane())
+				{
+					wallObjects.set(wallObjects.indexOf(obj), new WallObj(object, System.currentTimeMillis(), 0));
```
Adam-: Why does this timer start both when the wall spawns and despawns?

Barragek0: This sets the timer to 0, so it no longer appears on the overlay, as when calling remove(wallObjects.indexOf(obj)) instead, it was causing a lot of issues.

The one below that (else if) is to add the respawn timer to the list.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
Barragek0: Was to cleanup the list for debug reasons, forgot to remove before committing.

Adam-:
This is covered by #8973

PR #8827, opened 2019-05-13
correct misunderstandings related to wall orientation

### LINKED ISSUE #8825: Document `WallObject`'s Orientation Parameters
It is very hard to interpret the `int` result from `GetOrientationA` and `GetOrientationB` does not seem to return anything other than `0`.

I would like some documentation on these functions to describe the meaning of their return values. If the output can be described by an enum I would like that as well.

Thank you!

abextm:
Closes #8825

Orientation for WallObject and DecorativeObject is not used for rendering, as their rotation is baked into their model. As such it not in the normal jagex angle format, and should not be used for projection.

### DIFF
--- a/runelite-mixins/src/main/java/net/runelite/mixins/RSDecorativeObjectMixin.java
+++ b/runelite-mixins/src/main/java/net/runelite/mixins/RSDecorativeObjectMixin.java
@@ -110,7 +110,29 @@ public abstract class RSDecorativeObjectMixin implements RSDecorativeObject
 	@Override
 	public Area getClickbox()
 	{
-		return Perspective.getClickbox(client, getModel(), getOrientation(), getLocalLocation());
+		Area clickbox = new Area();
+
+		LocalPoint lp = getLocalLocation();
+		Area clickboxA = Perspective.getClickbox(client, getModel(), 0,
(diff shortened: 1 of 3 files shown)


runelite-mixins/src/main/java/net/runelite/mixins/RSDecorativeObjectMixin.java
```diff
@@ -110,7 +110,29 @@ private RSModel getModel2()
 	@Override
 	public Area getClickbox()
 	{
-		return Perspective.getClickbox(client, getModel(), getOrientation(), getLocalLocation());
+		Area clickbox = new Area();
+
+		LocalPoint lp = getLocalLocation();
+		Area clickboxA = Perspective.getClickbox(client, getModel(), 0,
+			new LocalPoint(lp.getX() + getXOffset(), lp.getY() + getYOffset()));
```
Adam-: This offset is actually only sometimes applied. I don't quite understand it but it has something to do with the camera angle. I think it is so if you are facing the wall behind it it won't draw through the wall.

abextm: I don't see where the offset is conditionally applied. I only see where it conditionally renders the primary or secondary renderable.

Adam-:
You're right, I'm reading this wrong.

PR #8841, opened 2019-05-16
Some Ground Items QOL updates

Thource:
Added sorting on the item overlay, items now appear in the order that they'll be picked up

Changed item menu sorting:
  Hidden items now appear at the bottom of the menu, so that hidden items don't get picked up over lower value non-hidden items

  Added a config option to put hidden items below "Walk here"
    this option makes it so that you can't accidentally pick up hidden items

  Added a config option to sort by GE value
    this option means that you'll always pick up highest GE value items first with left click
    this option also sorts the overlay by GE value

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -36,6 +36,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.List;
+import java.util.Comparator;
 import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.Perspective;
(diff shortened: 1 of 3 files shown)


deathbeam:
You should not need to sort by GE value because jagex already puts most valuable stuff as left click option

Thource:
> 
> 
> You should not need to sort by GE value because jagex already puts most valuable stuff as left click option

I thought this too, but jagex actually sort by item value, if you drop a snapdragon seed and then a steel platebody, you'll see what I mean

deathbeam:
I think they put seeds/herbs under other drops? Not sure

Thource:
> 
> 
> I think they put seeds/herbs under other drops? Not sure

It's the same for other things too, leather and dwellberries example, same order as before: https://imgur.com/a/PovezvY

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -31,11 +31,8 @@
 import java.awt.Graphics2D;
 import java.awt.Polygon;
 import java.awt.Rectangle;
+import java.util.*;
```
WilliamCollishaw: Per [Runelite code conventions ](https://github.com/runelite/runelite/wiki/Code-Conventions), wild card imports should not be used

Thource: Sorry, Idea did this automatically, fixed now

Thource:
Just realised that this is actually on the too OP list here: https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features

Feel free to decline this

Adam-:
FWIW the sort order of item stacks is most valuable item (in item price, not ge price) at the top, the rest is ordered by when it was dropped. When an item is picked up, the client finds the next most expensive item and moves it to the top. I would rather not adjust the item pick up order though.

Issue #1422, opened 2018-04-12
Cluescroll (medium) - no answers needed.


mtarik550:
The cluescroll states that 'Luthus' will give you a question scroll in which you need to either answer with 0 or 33 however when i done the clue he did not give me any question clues instead he gave me my next cluescroll.

PR #8843, opened 2019-05-16
New Plugin: Walk Here

Thource:
Walk Here is a plugin that moves the Walk action to the top of the menu when holding control, it's useful for things like easily walking under monsters


Hydrox6:
As mentioned in #1422, this is not something we'd want. Sorry

PR #8844, opened 2019-05-16
emoji plugin: Eggplant emoji added.

aleios:
I have added the eggplant emoji to the client. I used the trigger "8=D". Feel free to suggest a better ascii representation. I simply went with the one it will likely be used for despite its lack of class.

The emoji was lifted from the same source twemoji where it has been scaled to the 13x13 size, filtered, and then horizontally mirrored. These changes appear inline with the license. I built the client and tested it. Please let me know of any quality issues.


devLotto:
Why is it horizontally mirrored? Anyways this is the emoji resized with the same method as the others: 

eggplant

kaicianflone:
@devLotto how would I go about resizing emoji icons from twemoji if I wanted to add more emojis? Does @aleios emoji not work if it wasn't horizontally mirrored?

Issue #8381, opened 2019-04-02
Farming timer optimization


ramononis:
**Problem to be solved:**
Currently, farming timers are very inaccurate due to RuneLite not knowning when growth ticks will happen. Planting right after an actual growth tick, but before a 'RuneLite' growth tick, will leave the timer off by a whole stage (10+ hours for hardwood trees!).  
My guess is that, as part of the rewrite of the farming skill last year, each player has a constant offset w.r.t. to [the standard schedule](https://oldschool.runescape.wiki/w/Farming#Growth_cycles).  
For example, my growth ticks happen exactly 28 minutes and 55 seconds before they should happen according to the wiki: for me, a fruit tree tick actually happens at 10:11:05AM GMT while it should happen at 10:40:00AM GMT. This offset appears to be universal to all plant types for me: herbs grow at XX:11, XX:31 and XX:51, etc.

**Proposed solutions**
This problem is solved by knowing what the offset is w.r.t. to the standard schedule.  
The easiest solution is to enable the user to manually input this offset.  
The hard solution is to let RuneLite measure this:  
When RuneLite witnesses a growth tick, it will save the 'offset' w.r.t. to the standard schedule and uses this to give exact predictions. However, witnessing a fast growing plant (i.e. short growth cycle) provides less _information_ as witnessing a slow growing plant: witnessing a herb growing 5 minutes late could mean a offset of 5, 25, 45... minutes, witnessing a magic tree growing 5 minutes late could mean a offset of 5, 45, 85... minutes. On the other hand, witnessing a hardwood tree growing enables us to predict all other plant types as hardwood cycles can be evenly divided by any other cycle.  
A rule we can use is: when a growth tick has been witnessed on a plant, for all plant types with smaller or equal growth cycles exact predictions can be made: witnessing magic trees growing also allows us to predict when herbs grow.  
We can also refine our prediction by NOT witnessing a growth tick: say we witnessed a herb growing at an offset of +5 minutes. This means that magic trees can grow at an offset of +5 or +25 minutes (+45 is the same as +5 as the growth cycle is 40 minutes). If we don't see a magic tree (or any other plant with a 40 minute growth cycle) growing at +5 offset, we know that trees must grow at an offset of +25. This can continue upward to plants with larger growth cycles by canceling possible offsets out. In the end, when we can predict hardwood trees (which have the largest offset of 640 minutes), we can predict every plant on the exact second.  
  
[...]

PR #8851, opened 2019-05-16
WIP: Add farming notifications: trees, fruit trees, herbs

hjdarnel:
I tried to model this after how bird houses are tracked. Instead of being updated from IN_PROGRESS to COMPLETE via `checkCompletion()` on username change and region change, we can just use the existing summaries and completion time used for the summary tab. Then I added some configs to determine whether or not to notify the user.

Looking for feedback 😃


abextm:
I don't want notifications for this until #8381is closed and proven working

hjdarnel:
Not sure how dated this is, but it could probably be updated once the farming code is optimized #8381

PR #8853, opened 2019-05-16
Display current world ping

gregg1494:
Capture

Plugin to display ping in the corner of screen under fps. Currently it pings the players current world once every minute.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/ping/PingOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/ping/PingOverlay.java
@@ -1,0 +1,58 @@
+package net.runelite.client.plugins.ping;
+
+import java.awt.Color;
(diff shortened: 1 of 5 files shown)


SRLJustin:
I feel once every minute is too long for a refresh, would be nice to have an option in the plugin to change the ping timer with a minimum value enforced

gregg1494:
Ya I have to talk to some others about the frequency in which the ping should happen.

deathbeam:
You should move the ping code from world plugin to some common place instead of copying it

gregg1494:
Ya I wanted to but didn't know where it would go. If someone knows where it should go I'll be happy to move it.

deathbeam:
i would just put it to `net.runelite.client.ping`, e.g new package

gregg1494:
Implemented into world hopper. #8871

PR #8856, opened 2019-05-17
Explorer's ring overlay

aleios:
Adds an overlay to the explorer's ring showing the remaining charges on High/Low-Alchemy, Run energy replenishment, and Teleports(If level-2 ring). 

ering-3

ering-2

Plugin has same options that rune pouch plugin provides including text color and switching the display mode to allow showing in inventory, tooltip or both.

ering-4

I think maybe the text needs to be changed. I am open to suggestion.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -487,7 +487,15 @@ public enum Varbits
 	/**
 	 * The active tab within the quest interface
 	 */
-	QUEST_TAB(8168);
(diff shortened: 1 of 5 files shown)


Nightfirecat:
This seems like it would be a much better fit to be added to the item charges plugin instead of being a standalone plugin.

aleios:
Yeah you're right. I don't know why I didn't add it there. I think I'll transplant the code here to it.

aleios:
Self closing this PR. Integrated into itemcharges plugin. Will add new PR when testing concluded.

PR #3719, opened 2018-06-09
Skill Calculator Expansion

TheStonedTurtle:
Thanks to @psikoi for helping to design the UI

## Skill Calculator Changes:
* Skill Planner Functionality added
* Banked Experience Functionality added
* Skill Calculator UI was updated to accommodate new functionality
* Added Config options to toggle new Functionality

## Planner Functionality:
* Right-Click an item to specify the amount you'd like to plan for
* Shows ending Level and Experience inside the **Target** box based on all planned actions.

## Banked Experience Functionality:
* Banked Experience shows exp from all items, regardless of required level
* Pulls inventory from entire bank regardless of visible tab
* Categorized items with toggle ability inside panel
* Shows ending Level and Experience inside the **Target** box
* Currently only supports single item training methods and fixed xp rates per raw resource
* Banked Experience currently works for:
	* Construction
	* Cooking
	* Crafting
	* Prayer
	* Smithing

## API Updates:
* Updated the **Skill** api to have a **getByName** method which returns the Skill if the passed string matches the name. All strings are upper-cased for compatibility.

## UI Component Updates:
* Added components to assist in creating **JComboBox** options with an icon and text.
* Updated **MaterialTab** to have RuneScape font as the default and to allow setting of Font, Selected Border, and Unselected Border.

## Summary of new UI Components
<details>
<p>

### ComboBoxIconEntry
* Requires an **Icon** and **String**.
* Should be added to your **JComboBox**
### ComboBoxIconRender
* Used for rendering **ComboBoxIconEntry** elements on a **JComboBox**
* Should be passed to **JComboBox.setRenderer()** 
* Supports no default selection (Must select the -1 index or null item, defaults to selecting first option)
* Settable default text for no default selection. (Default Value: "Select an option...")

</p>
</details>

## Media
<details>
<p>

#### Banked Experience

</p>
</details>

PR #8861, opened 2019-05-17
Banked prayer experience

SverreHassel:
Calculates the amount of prayer experience stored in the players bank. Initially started as a copy of the existing bank-plugin with modifications, but decided it was better to just move it into the "original". Commits #fb091b6 and #5060623 are the ones that modify existing code.


Adam-:
This seems like something which would be better suited to be in the skill calculator? Also it can apply to a lot more skills than prayer. This is basically what #3719 tried. Eventually I think we decided to leave it up to the various skill calculators on fansites since for most skills it is not so straight forward to calculate.

PR #8867, opened 2019-05-18
itemcharges: Add explorer's ring alchemy charge overlay.

aleios:
Adds a counter and info box from the item charges plugin to Explorer's Ring 1-4.

eringr-1

eringr-3

Also added the relevant varbits.


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -244,11 +245,51 @@ default boolean bindingNotification()
...
+	)
+	void explorerRing(int explorerRing);
+
+	@ConfigItem(
+		keyName = "explorerRingMaxCharges",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default int explorerRingMaxCharges()
```
deathbeam: This should not be config item

PR #8874, opened 2019-05-20
Refactor combat level calculations to use closed-form formulas.

### LINKED ISSUE #7411: Levels to next combat are not correct for most skills

emerald000:
Also move most calculations to the Experience utility class.
Fixes #7411.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Experience.java
+++ b/runelite-api/src/main/java/net/runelite/api/Experience.java
@@ -24,9 +24,6 @@
  */
 package net.runelite.api;
 
-import static java.lang.Math.floor;
-import static java.lang.Math.max;
-
 /**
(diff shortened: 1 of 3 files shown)


Adam-:
This needs tests added for the cases the existing code fails for.

abextm:
I much prefer #7528's implementation of this because its much simpler, and trying to be smart has failed too many times

Adam-:
The tests in this look correct and I like this more since its changing less of the plugin around. We can revisit #7528 if this turns out to be wrong too.

PR #8894, opened 2019-05-23
Fix wine fermentation timer to begin at appropriate time

abryant2113:
Previously, the wine fermentation timer would enter into a negative range if the user cooked an inventory full of wine and then tried to cook another inventory.

This change resets the fermentation timer each time the player is performing the wine cooking animation, giving a more accurate measure of the remaining time to fermentation.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
@@ -31,7 +31,11 @@ import java.time.Instant;
 import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
+import static net.runelite.api.AnimationID.COOKING_WINE;
 import net.runelite.api.ChatMessageType;
+import net.runelite.api.Client;
+import net.runelite.api.Player;
+import net.runelite.api.events.AnimationChanged;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameTick;
 import net.runelite.client.config.ConfigManager;
@@ -50,7 +54,8 @@ import net.runelite.client.ui.overlay.OverlayManager;
 @PluginDependency(XpTrackerPlugin.class)
 public class CookingPlugin extends Plugin
 {
-	private static final String WINE_MESSAGE = "You squeeze the grapes into the jug";
+	@Inject
+	private Client client;
 
 	@Inject
 	private CookingConfig config;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cooking/FermentTimerOverlay.java
```diff
@@ -72,7 +72,12 @@ public Dimension render(Graphics2D graphics)
 
 		panelComponent.getChildren().clear();
 
-		if (isMakingWine() || Duration.between(session.getLastWineMakingAction(), Instant.now()).getSeconds() < INITIAL_TIME)
+		if (isMakingWine())
+		{
+			session.updateLastWineMakingAction();
```
Nightfirecat: This should be updated in the plugin so it is not constantly updated once per render frame.

abryant2113: Agreed. Just updated this in the latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
```diff
@@ -121,6 +126,11 @@ public void onGameTick(GameTick gameTick)
 			{
 				fermentTimerSession = null;
 			}
+
+			if (client.getLocalPlayer().getAnimation() == COOKING_WINE)
```
Nightfirecat: You should never need to check for this _and_ check the difference in time since the last wine making action. Place this on L125 and change the previous `if` to `else if`.

abryant2113: Fixed.

Adam-:
Do the animation and chat message not happen at the same time? Why does this fix the problem? Is the chat message check still required?

raiyni:
Animation happens before chat message

abryant2113:
Yeah, the animation occurs as soon as the player mixes the grapes with the jugs of water. The chat message won't appear until the player can't make any more jugs of wine until the action is interrupted or the player runs out of supplies in their inventory.

Adam-:
I am referring to the "You squeeze the grapes into the jug"... message, which I think happens for each wine made?

abryant2113:
Nope. That message doesn't appear until player runs out of supplies. I actually just tested it by interrupting the wine-making action and that doesn't produce the message, nor does it happen after each jug of wine is made.

That message would appear for every jug of wine made if you were only making one jug of wine per inventory, but I doubt anyone is making one jug of wine, banking, and then repeating the process.

Adam-:
Is the code checking for the message and updating the last time even necessary then?

abryant2113:
Nope! Just removed the code involving the message checking and tested it out. Everything seems to work as intended. The newest commit should reflect these changes.

runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
```diff
@@ -117,11 +120,20 @@ public void onGameTick(GameTick gameTick)
...
+				fermentTimerSession.updateLastWineMakingAction();
+			}
+			else if (sinceCut.compareTo(statTimeout) >= 0)
 			{
 				fermentTimerSession = null;
 			}
 		}
+		else if (config.fermentTimer() && client.getLocalPlayer().getAnimation() == COOKING_WINE)
+		{
+			fermentTimerSession = new FermentTimerSession();
```
Adam-: This should just use the animation change event probably.

abryant2113: You mean remove the config.fermentTimer() condition, correct? Or is there a specific animation event that I'm unaware of?

abryant2113: Just found the class, nvm.

abryant2113: Fixed.

PR #8895, opened 2019-05-23
PluginManagerTest: Ensure config keyNames are not duplicated

Nightfirecat:
This prevents runtime errors if an erroneously duplicated keyName is
used in a plugin's configuration.

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/PluginManagerTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/PluginManagerTest.java
@@ -38,6 +38,7 @@ import java.applet.Applet;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/PluginManagerTest.java
```diff
@@ -95,6 +98,37 @@ public void before() throws IOException
 			if (pluginDescriptor != null)
 			{
 				pluginClasses.add(clazz);
+				continue;
+			}
+
+			// Ensure plugin configurations do not repeat keyNames
+			if (Config.class.isAssignableFrom(clazz))
```
deathbeam: You should just build list of config classes here and then make separate test, `@Before` is for preparing environment for tests, not for testing stuff

Nightfirecat: I've done just that. :+1:

ThatGamerBlue:
what about setting config values from inside the plugin, currently this requires a duplicate key name with a different method signature, from what i can see this test will throw an exception if that case exists

deathbeam:
@ThatGamerBlue nope, because this test has check for method being default

PR #8896, opened 2019-05-23
Added config for showing stat overlay for worn equipment

CubeLuke:
Added a new config for the Item Stat plugin that will allow for stats to be displayed via the tooltip within the worn equipment tab of the interface. The default setting for this will be false, just since the majority of players likely won't enjoy having the tooltips showing up all of a sudden.

image

Similar to some of the other configs for this plugin, if "Enable equipment stats" is false, then the item stat tooltip will not show up within the worn equipment tab.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
@@ -53,6 +53,16 @@ public interface ItemStatConfig extends Config
 	}
 
 	@ConfigItem(
+			keyName = "showInEquipmentGroup",
+			name = "Show in worn equipment",
+			description = "Will show equipment stat tooltips within worn equipment"
+	)
+	default boolean showInEquipmentGroup()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "geStats",
 		name = "Enable GE item information",
 		description = "Shows an item information panel when buying items in the GE"
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
@@ -143,7 +143,7 @@ public class ItemStatOverlay extends Overlay
 
 			if (stats != null)
 			{
-				final String tooltip = buildStatBonusString(stats);
+				final String tooltip = buildStatBonusString(stats, group);
 
 				if (!tooltip.isEmpty())
 				{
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -186,33 +186,37 @@ private String getChangeString(
...
-		if (s.isEquipable() && currentEquipment != null && c != null)
+		if (!(config.showInEquipmentGroup() && groupId == WidgetInfo.EQUIPMENT.getGroupId()))
 		{
-			final Item[] items = c.getItems();
-			final int slot = currentEquipment.getSlot();
+			final ItemEquipmentStats currentEquipment = s.getEquipment();
+			ItemContainer c = client.getItemContainer(InventoryID.EQUIPMENT);
 
-			if (slot != -1 && slot < items.length)
+			if (s.isEquipable() && currentEquipment != null && c != null)
```
Adam-: If I'm reading this correctly, all you need is to make `int groupId` to -> `boolean isEquipment` and then add `&& isEquipment` to here.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
```diff
@@ -52,6 +52,16 @@ default boolean equipmentStats()
 		return true;
 	}
 
+	@ConfigItem(
+			keyName = "showInEquipmentGroup",
```
Adam-:
```suggestion
		keyName = "showInEquipmentGroup",
```

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
```diff
@@ -52,6 +52,16 @@ default boolean equipmentStats()
...
+			name = "Show in worn equipment",
```
Adam-:
```suggestion
		name = "Show in worn equipment",
```

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
```diff
@@ -52,6 +52,16 @@ default boolean equipmentStats()
...
+			description = "Will show equipment stat tooltips within worn equipment"
```
Adam-:
```suggestion
		description = "Will show equipment stat tooltips within worn equipment"
```

Adam-:
This almost looks like a bug to me since it is cancelling out the item stats with what you have worn.

CubeLuke:
@Adam- Not sure I'm 100% following with what you meant by saying it seems like a bug? 

The overlay is grabbing the stats of both the currently equipped item, and the one you are hovering over, in order to do the stat comparison. If we keep `ItemStats` set to null, then we will always be showing the stats of the item we are currently hovering over. Since we are hovering over a currently equipped item, we need to skip the logic that is pulling stats from that currently equipped item, so that we can properly display the positive/negative stats from the hovered item (which happens to be the currently equipped). 

I might 100% be missing something here though since I'm new to the codebase, so lemme know!

Adam-:
I think it is a bug because it is supposed to show you the difference in what you are hovering vs what you are wearing, but if you hover over what you are wearing it is always 0 difference.

PR #8904, opened 2019-05-23
Added function key remapping

RALEx147:
This maps the 1~10 keys to their respective function keys for keyboard that don't support fn keys, or to simplify the process of using function keys. This is a separate plugin from the wasdcamera but has the exact same structure.


Alexsuperfly:
i would rather see this merged into the wasd camera and rename that plugin itself, seems pretty pointless to have two plugins essentially performing the same thing when one works fine

RALEx147:
Yeah I have that in a different version, but didn't know which was better

Adam-:
I think that this is better suited for AHK?

RALEx147:
The main reason for this plugin is to help mac users. I dont think ahk is as easy to setup.

PR #8915, opened 2019-05-24
Rename WASDCamera to add Fn key remapped functionality

RALEx147:
This is a revised edition of the keyboard remapping that adds functionality for the use of Fn keys through their numerical counterparts (1~10). This consolidates WASD and the FN keys to one plugin with separated boolean configs, but keeping the wasd customization.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -34,11 +34,26 @@ import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.VarClientStr;
 import net.runelite.client.callback.ClientThread;
+import net.runelite.client.config.Keybind;
+import net.runelite.client.config.ModifierlessKeybind;
 import net.runelite.client.input.KeyListener;
 import net.runelite.client.input.MouseAdapter;
 
 class KeyRemappingListener extends MouseAdapter implements KeyListener
 {
+	private static final Keybind ONE = new ModifierlessKeybind(KeyEvent.VK_1, 0);
+	private static final Keybind TWO = new ModifierlessKeybind(KeyEvent.VK_2, 0);
+	private static final Keybind THREE = new ModifierlessKeybind(KeyEvent.VK_3, 0);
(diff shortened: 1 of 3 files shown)


Adam-:
Can you rebase this and split the commit that renames the plugin from the one adding the new feature? It is much easier to review then.

RALEx147:
Any clue why github doesn't recognize WASDCameraListener changes to KeyRemappingListener in files changed, but the other two do.

Nightfirecat:
That's just a function of git's rename detection. Don't worry too much about that. So long as you have separate commits for the separate changes like Adam requested, it's fine. (Looks like you still need to rebase and squash into those commits though)

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -0,0 +1,350 @@
...
+			{
+				modified.remove(e.getKeyCode());
+				e.setKeyCode(m);
+			}
+		}
+	}
+	
+	private ModifierlessKeybind one()
+	{
+		return new ModifierlessKeybind(KeyEvent.VK_1, 0);
```
Adam-: Is there a reason you are making these each keystroke and not just keep keeping one of each? I think they are immutable.

RALEx147: I modeled it after the wasdcamera config

Adam-: Wasd config is like that because it is config. This isn't config.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -0,0 +1,314 @@
...
+				{
+					e.setKeyCode(KeyEvent.VK_LEFT);
+				}
+				else if (config.right().matches(e))
+				{
+					e.setKeyCode(KeyEvent.VK_RIGHT);
+				}
+				
+				
+				if (config.functionKeyRemap())
```
Adam-: why is this inside of the `if (config.wasdCameraRemap())`?

RALEx147: fucked it up after rebasing

PR #8916, opened 2019-05-24
duel arena chat command

### LINKED ISSUE #8909: Chat command for duel win/loss
!duels command for linking win/loss ratio

RALEx147:
Fixes #8909
<https://i.imgur.com/ZHfe2oh.png>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -65,6 +65,7 @@ import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.util.StackFormatter;
 import static net.runelite.client.util.Text.sanitize;
 import net.runelite.http.api.chat.ChatClient;
+import net.runelite.http.api.chat.Duels;
 import net.runelite.http.api.hiscore.HiscoreClient;
 import net.runelite.http.api.hiscore.HiscoreEndpoint;
 import net.runelite.http.api.hiscore.HiscoreResult;
@@ -88,6 +89,9 @@ public class ChatCommandsPlugin extends Plugin
 	private static final Pattern BARROWS_PATTERN = Pattern.compile("Your Barrows chest count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern KILL_DURATION_PATTERN = Pattern.compile("Fight duration: <col=ff0000>[0-9:]+</col>. Personal best: ([0-9:]+)");
 	private static final Pattern NEW_PB_PATTERN = Pattern.compile("Fight duration: <col=ff0000>([0-9:]+)</col> \\(new personal best\\)");
+	private static final Pattern DUEL_ARENA_WINS_PATTERN = Pattern.compile("You (were defeated|won)! You have(?: now)? won (\\d+) duels?");
+	private static final Pattern DUEL_ARENA_LOSSES_PATTERN = Pattern.compile("You have(?: now)? lost (\\d+) duels?");
+
 	private static final String TOTAL_LEVEL_COMMAND_STRING = "!total";
 	private static final String PRICE_COMMAND_STRING = "!price";
 	private static final String LEVEL_COMMAND_STRING = "!lvl";
@@ -97,6 +101,7 @@ public class ChatCommandsPlugin extends Plugin
 	private static final String QP_COMMAND_STRING = "!qp";
 	private static final String PB_COMMAND = "!pb";
 	private static final String GC_COMMAND_STRING = "!gc";
+	private static final String DUEL_ARENA_COMMAND = "!duels";
 
 	private final HiscoreClient hiscoreClient = new HiscoreClient();
 	private final ChatClient chatClient = new ChatClient();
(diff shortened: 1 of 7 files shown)


Adam-:
1) add a test case to `ChatCommandsPluginTest` wiith the duel messages
2) remove the mode modifications to the keyboard listener and skill abbreviations
3) maybe pack both wins and losses into one int so avoid having to do 2x web requests?

RALEx147:
The issue with part 3 is that duel area messages are in two lines, not sure if it is possible for onChatMessage to see previous chat messages.

RALEx147:
Another advanced feature this could have is to keep track of streaks, like how many you win in a row. But this may also increase web requests.

RALEx147:
I just implemented the win/losing streak functionality, but it calls a total of 3 web request. [Here are images of how they look](https://imgur.com/a/Noe7UVZ)

I haven't pushed it yet though

Nightfirecat:
Like Adam said, the wins and losses can be bitpacked (eg. `wins << 16 + losses`) to avoid an extra call. I don't know what others' feelings are on whether we should include the streak as well.

RALEx147:
Yeah I knew how to do it but didnt know I only had to do it for the chatClient versions of submitkc

RALEx147:
I also should refactor the bitpacking into two separate functions

Adam-:
I certainly wouldn't bitpack the wins/losses in the config, that was just a suggestion for abusing the kc service. Maybe it would be worth adding a separate endpoint for duels.

RALEx147:
Yeah this is a working version, but it could definitely can be improved

RALEx147:
I added support for a separate end point for the duels command, the plugin no longer works but if the web api is updated then all should be good.

Adam-:
Okay. It's not too hard to setup the api in intellij to run if you want to do that to test it. I will try to look at this tomorrow.

Adam-:
This is missing all of the changes to `http-service` required for the api to work.

Adam-:
Did you run http-service to test this?

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -87,6 +87,9 @@
 	private static final Pattern BARROWS_PATTERN = Pattern.compile("Your Barrows chest count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern KILL_DURATION_PATTERN = Pattern.compile("Fight duration: <col=ff0000>[0-9:]+</col>. Personal best: ([0-9:]+)");
 	private static final Pattern NEW_PB_PATTERN = Pattern.compile("Fight duration: <col=ff0000>([0-9:]+)</col> \\(new personal best\\)");
+	private static final Pattern DUEL_ARENA_WINS_PATTERN = Pattern.compile("You \\w+( \\w+)?! You have (now )?won (\\d+) duels");
```
Adam-: This seems like a fairly wide regex considering both test strings start with `You won!` - what other combinations are there?

RALEx147: `You were defeated! You have won x duels.` and `You won! You have now won x duels.`

It matches wins and losses so even if you lose it track know how many you have won, since it is separated in two messages. Also the first duel needs to parse both wins and losses.

RALEx147: so like the message is different for wins/losses

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -87,6 +88,9 @@
 	private static final Pattern BARROWS_PATTERN = Pattern.compile("Your Barrows chest count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern KILL_DURATION_PATTERN = Pattern.compile("Fight duration: <col=ff0000>[0-9:]+</col>. Personal best: ([0-9:]+)");
 	private static final Pattern NEW_PB_PATTERN = Pattern.compile("Fight duration: <col=ff0000>([0-9:]+)</col> \\(new personal best\\)");
+	private static final Pattern DUEL_ARENA_WINS_PATTERN = Pattern.compile("You (were defeated|won)! You have( now)? won (\\d+) duels");
```
Nightfirecat: Don't capture text you won't use when matching.
```suggestion
	private static final Pattern DUEL_ARENA_WINS_PATTERN = Pattern.compile("You (were defeated|won)! You have(?: now)? won (\\d+) duels");
```

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -87,6 +88,9 @@
...
+	private static final Pattern DUEL_ARENA_LOSSES_PATTERN = Pattern.compile("You have( now)?lost (\\d+) duels");
```
Nightfirecat:
```suggestion
	private static final Pattern DUEL_ARENA_LOSSES_PATTERN = Pattern.compile("You have(?: now)?lost (\\d+) duels");
```

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -233,6 +242,43 @@ public void onChatMessage(ChatMessage chatMessage)
...
+				winningStreak += 1;
+			}
+			else if (result.equals("were defeated"))
+			{
+				losingStreak += 1;
+				winningStreak = 0;
+			}
+			else
+			{
+				log.debug("Parse error");
```
Nightfirecat: This shouldn't be a debug log, this should probably be `log.warn`. Also, it should print some information about the text which caused said parsing error, otherwise this doesn't really help anyone figure out how to fix it.

Adam-:
@RALEx147 have you confirmed the varbit value is correct when outside of ba?

PR #8933, opened 2019-05-27
Refactor switch fallthrough and abstractions

AdamKBeck:
Moved the nominal part of the switch statement fallthrough to use a set, hopefully providing faster runtime. Moved repeated notification messaging to a helper function to reduce repeated code.


Nightfirecat:
Given that this PR has not been updated or maintained since it was opened some 9 months ago, and offers no justification or reasoning behind set lookup being faster than switch case testing (not considering the likely negligible performance impact of this change), I am closing this PR. If you are interested in revisiting this, I strongly recommend having more sound reasoning than "hopefully providing faster runtime"&mdash;bytecode differences, external references, and/or practical runtime comparisons would be ideal.

Issue #8946, opened 2019-05-28
Special attack bar can become hidden until relog


Nightfirecat:
After #8925 has been merged, the special attack bar can now be hidden (until relog) in this particular case:

1. Have no weapon equipped, or a weapon without any special attacks
2. Open the attack styles tab
3. Equip any weapon with a special attack
4. The special attack bar will not appear. This persists until logout.

This happens regardless of whether the "Hide auto retaliate" option is enabled, and even with the Attack Styles plugin disabled.

![GIF](https://user-images.githubusercontent.com/2199511/58519057-7b4ef280-81a1-11e9-840c-786e9b92bd58.gif)

PR #8947, opened 2019-05-28
attack styles: Attach spec bar redraw listener to parent

Nightfirecat:
This fixes a bug where the special attack bar would not be redrawn if
the combat tab was opened with any weapon equipped which did not have a
special attack bar.

Fixes runelite/runelite#8946


Nightfirecat:
Agreed, 593.1 seems to fully work without causing other problems. Updated the PR to use that.

PR #8949, opened 2019-05-29
menu entry swapper option

lxlxxlxl99:
menu entry swapper option: can we had a teleport item option only for quest cape ? please ?


deathbeam:
You can use shift-click configuration. also you opened pull request and not issue

PR #8953, opened 2019-05-30
world hop messages in game chat

VincentJ711:
gives the player the information needed to not hop into a world they've already logged into. my friend found the current "quick-hopping ..." message to be unreliable. he said it only worked "some" of the time. he asked if a list of the worlds he'd visited could be printed on login.


raiyni:
how is it unreliable? it prints the world you hop to

VincentJ711:
> how is it unreliable? it prints the world you hop to

let me rephrase. sometimes when he hopped, the message wouldn't print.

VincentJ711:
with my testing, i found the "quick-hopping ..." message was only printed when i hopped worlds using the sidebar hopper or the keyboard shortcut defined in the World Hopper config. Of course this works, but if you hop via the in game world hopper, the message is not printed. this commit makes it so the message is always printed regardless of where you hop from.

PR #8622, opened 2019-04-23
Prepare Barbarian Assault plugin for changes passed in the recent poll

Jacoblairm:
Change the role identifier from cape slot "Icons" to Role Horns in the inventory. In preparation for the recently passed poll question - allowing capes in Barbarian Assault. https://i.imgur.com/45vJHNe.png

Also null the gameTime after a game has complete to fix a minor bug causing the game duration announcement after completing the tutorial.

PR #8955, opened 2019-05-30
Fix BA plugin after latest game update.

Jacoblairm:
No longer rely on the previously used "icons" to determine the role, instead use the role widget currently loaded. Tested and working https://i.imgur.com/x0JnDsD.png

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
@@ -31,13 +31,8 @@ import java.awt.Image;
 import javax.inject.Inject;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
-import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemID;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.ChatMessage;
-import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -115,13 +110,38 @@ protected void shutDown() throws Exception
 	@Subscribe
 	public void onWidgetLoaded(WidgetLoaded event)
 	{
-		if (event.getGroupId() == WidgetID.BA_REWARD_GROUP_ID)
```
Hydrox6: Why have you removed the stuff that handles the Wave Times option?

Jacoblairm: I haven't, I just moved it into the switch statement.

Hydrox6: Ah, that's a confusing diff. Nvm then

Adam-:
Is the `gameTime = null` fix from #8622 important? should it be here?

Jacoblairm:
> Is the `gameTime = null` fix from #8622 important? should it be here?

Not too important but I just committed the fix.

Twiglet1022:
I've just tried it this out at some BA and it works great except for one issue which is that the timer resets to 30 secs when you close the horn of glory interface. It always disappeared when the horn of glory was up, but it would remember the time for when you closed it again.

It's pretty minor tbh because I doubt many people use the horn, but I find it useful sometimes.

I think it might be because you're now relying on the widget, so the way #8622 does it might be better because you still have the horn in your inventory when you open that interface.

Adam-:
@Jacoblairm what does it fix?

Jacoblairm:
> I've just tried it this out at some BA and it works great except for one issue which is that the timer resets to 30 secs when you close the horn of glory interface. It always disappeared when the horn of glory was up, but it would remember the time for when you closed it again.
> 
> It's pretty minor tbh because I doubt many people use the horn, but I find it useful sometimes.
> 
> I think it might be because you're now relying on the widget, so the way #8622 does it might be better because you still have the horn in your inventory when you open that interface.

Fixed this issue by checking if the currentRound has already been set during the round.

Adam-:
I meant, what is `gameTime = null` fixing? What happens if you don't add it?

Jacoblairm:
@Adam- https://i.imgur.com/fNbitFp.png occurs when you complete a tutorial and waveTime is not null (ie - after completing at least one wave). I just realized `gameTime = null` doesn't properly fix it; I have remediated the bug in the latest commit.

Adam-:
Is there a reason why this is waiting until you enter the following wave to tell you your previous waves time?

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -195,7 +177,7 @@ public void onVarbitChanged(VarbitChanged event)
 			{
 				overlay.setCurrentRound(null);
 
-				if (config.waveTimes() && gameTime != null)
+				if (config.waveTimes() && gameTime != null && client.isInInstancedRegion())
```
Adam-: Is this to try and avoid messages from the tutorials? How does that work as the tutorials are instanced?

Jacoblairm: Yes, it is. The tutorial uses the IN_GAME_BA varbit also. 
After exiting a tutorial there is a 2-3 tick delay for the IN_GAME_BA varbit to change from 1 -> 0 (2-3 ticks of being on the mainland). 
When finishing a real wave the IN_GAME_BA varbit change occurs when it is still in the BA instance. Checking if your in an instance still stops the tutorial from announcing wave duration.

Adam-: This is the type of things comments were invented for.

Adam-:
@Jacoblairm can you verify this is correct now?

Jacoblairm:
Yes that is correct.

PR #8978, opened 2019-06-01
Add support for pasting authenticator code on login screen

deathbeam:
- Add new LOGIN_SCREEN_AUTHENTICATOR game state
- Export otp and loginIndex from rs api
- Add support for pasting to authenticator pin (otp) field

Depends on !67

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
@@ -54,6 +54,7 @@ public class LoginScreenPlugin extends Plugin implements KeyListener
 {
 	private static final int MAX_USERNAME_LENGTH = 254;
 	private static final int MAX_PASSWORD_LENGTH = 20;
+	private static final int MAX_PIN_LENGTH = 6;
 
 	@Inject
 	private Client client;
@@ -163,7 +164,9 @@ public class LoginScreenPlugin extends Plugin implements KeyListener
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (!config.pasteEnabled() || client.getGameState() != GameState.LOGIN_SCREEN)
+		if (!config.pasteEnabled() || (
+			client.getGameState() != GameState.LOGIN_SCREEN &&
+			client.getGameState() != GameState.LOGIN_SCREEN_AUTHENTICATOR))
 		{
 			return;
 		}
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -182,16 +185,26 @@ public void keyPressed(KeyEvent e)
 					.toString()
 					.trim();
 
-				// 0 is username, 1 is password
-				if (client.getCurrentLoginField() == 0)
+				switch (client.getLoginIndex())
```
Nightfirecat: Maybe the values for this field should be documented in `Client`?

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -182,16 +185,26 @@ public void keyPressed(KeyEvent e)
...
-					// Truncate data to maximum username length if necessary
-					client.setUsername(data.substring(0, Math.min(data.length(), MAX_USERNAME_LENGTH)));
-				}
-				else
-				{
-					// Truncate data to maximum password length if necessary
-					client.setPassword(data.substring(0, Math.min(data.length(), MAX_PASSWORD_LENGTH)));
+					case 2:
+						// 0 is username, 1 is password
+						if (client.getCurrentLoginField() == 0)
```
Nightfirecat: I know this was added before, but same here. Documenting these values would be nice.

deathbeam: I think at that point it should be just changed to enum.I just didnt bothered with it originally because the PR was stuck for so long

deathbeam: Actually, its just 2 values, and its already documented. But I will update loginIndex

deathbeam:
Updated

Nightfirecat:
This will likely break some number of plugins which rely on determining when a player logs in by storing the previous game state and checking whether it is `LOGIN_SCREEN` (ex. ), or which perform specific behavior on the login screen. (ex. screenshot plugin not capturing screenshots on login screen) Also, I'd expect some plugins with `LOGIN_SCREEN` guard clauses to exhibit weird behavior because of the gamestate change event being fired.

abextm:
The same amount of events will file, except that it can be `LOGIN_SCREEN_AUTHENTICATOR` instead of `UNKNOWN`

Issue #8971, opened 2019-05-31
Music List Plugin - Filter/search context menu constantly expands


asherbasher:
**Describe the bug**
The right-click context menu of both the filter and search buttons on the music list will continue to add more and more of the same option. I believe it's each time a different track is automatically played each session. Please see my screenshot as it's much easier to explain in image format.

**To Reproduce**
1. Go to the music tab after having your music on auto and having been in a few different areas which would've played different songs
2. Right-click either the filter or search buttons

**Expected behavior**
The menu should not be adding duplicates of the "Toggle" or "Open Search" context entries.

**Screenshots**
![Context menu bug](https://user-images.githubusercontent.com/10160470/58712545-ad866d00-83b8-11e9-821b-75deb0fff857.png)

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.5.25
 - Launcher version: 1.6.0

andmcadams:
It looks like this happens when teleporting places. When I tried walking around to areas with different music, no new menu items were added.

**Environment (please complete the following information):**

- OS: Linux Mint 19 (Tara)
- RuneLite version: 1.5.25
- Launcher version: 1.6.2

Adam-:
Looks like this does only happen when teleporting places. Also I got it to happen to the quest list search thing too but it doesn't happen as reliably as it does with the music one for some reason.

andmcadams:
Yeah, it looks like there are assumptions that these buttons will only be created once. I've been looking at the quest one for a while now. On startup, it actually generates two of each button, but since they overlay each other, you can't tell the second button is underneath unless you right click.

Adam-:
fixed in c020717346f32ffa1695ad7aa412f32b6b01606f

PR #8982, opened 2019-06-01
Fix music list plugin growing search/filter menu bug

andmcadams:
This fix is in response to issue #8971. A simple name check is included to see if either of the search or filter buttons have been previously added, and doesn't add new buttons if this is the case. This seems to have been the cause of the bug initially, as two new buttons (and their options) would be added every time the onWidgetLoaded method was called with the correct id.

Assumes that the two buttons will either both be there or neither be there. Could always add more checks to handle cases of one but not the other being there if desired.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/musiclist/MusicListPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/musiclist/MusicListPlugin.java
@@ -123,6 +123,14 @@ public class MusicListPlugin extends Plugin
 			return;
 		}
 
+		//See if the search and filter buttons have already been added to avoid re-adding
+		if (header.getChildren() != null)
+		{
+			for (Widget w : header.getChildren())
+				if (w.getName().equals("Search") || w.getName().equals("All"))
+					return;
+		}
+
 		//Creation of the search and toggle status buttons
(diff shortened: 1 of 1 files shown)


Hydrox6:
Wouldn't it be a better idea to do a null check on `musicSearchButton` and `musicFilterButton`, and null them on Shutdown?

```java
private void addMusicButtons()
{
	Widget header = client.getWidget(WidgetInfo.MUSIC_WINDOW);

	if (header == null)
	{
		return;
	}

	if (musicSearchButton == null || musicFilterButton == null)
	{
		return;
	}

	...
```

Adam-:
I am just going to `deleteAllChildren`()` before adding them probably

andmcadams:
@Hydrox6  That does sound like a better idea. I think you would have to check for != null though. Otherwise the buttons will never be created.

@Adam-  That was my original idea, but that only works until you add some other widget in. If there's ever a separate plugin added that adds widgets, deleteAllChildren() will destroy those as well. Might not be an issue depending on how likely that is.

Adam-:
The plugin already makes this assumption as that is how `shutDown` works

andmcadams:
That's a good point. `shutDown` could always be changed too if the assumption should be avoided in general. Either way it isn't much work.

Adam-:
fixed in c020717346f32ffa1695ad7aa412f32b6b01606f

PR #8985, opened 2019-06-01
Fishing Plugin: fix visual bug when multiple of the same kind of fish are stacked

### LINKED ISSUE #1392: Duplicate fishing sprites

JordanWWood:
This fix resolves #1392

When rendering the fishing spots this change double checks if another spot of the same type has already rendered there. The way I stopped spots from stacking also seems to have fixed the floating icons that would previously occur.
Numbers have also been added to point out how many spots are stacked together. I used a small bit of code from PR #7897 to achieve this.
Image


Nightfirecat:
How does this behave when spots of different kinds are on the same tile?

JordanWWood:
I don't have access to the fishing guild and I spent a while world hopping in some areas where a fair few spots spawn and I couldn't find a single occurrence of it happening. I'm wondering if something changed at some point.

However, in the case two spots of differing types spawn on top of each other it should act as before since the main part of the visual issue was two icons of the same kind of fish appearing on top of each other. It may be interesting to have the icons appear to a more appealing way though.

JordanWWood:
Just spent a while looking into it just to be sure and it does as I said.
Multiple Spots
Another Angle

Adam-:
Let's try to avoid n^2 algorithms in the overlays. I think that you can do this by changing `fishingSpots` to a `LinkedHashMultimap` (assuming sorting its `values()` works as I think it does?). Also don't show a `(0)` if there are no duplicates.

JordanWWood:
I moved over to building the count before we enter the rendering loop. 
I also changed how it keeps track of what has already been rendered to make use of a multimap as you suggested so it can do a hashed lookup instead of looping.

The new count changes the number of iterations to n*2 instead of n^2

raiyni:
Why not do the counting during the spawn events in the plugin?

JordanWWood:
I was looking into doing it that way but I was running into issues with it. I decided to compromise and come back to it in the morning.
I saw your post on the Discord [Here](https://github.com/runelite/runelite/compare/master...raiyni:fishing-poc?expand=1). So I'll look into doing something similar later today

Adam-:
Fixed in 9fc04bf7e2469b4f77445c62464dedc9d117092b

PR #8990, opened 2019-06-02
falconry plugin

stealthAngel:
https://www.youtube.com/watch?v=QiJ06C2k19Y&feature=youtu.be


Nightfirecat:
The linked video is unavailable.

Also, what functionality does this plugin provide which isn't already available via the NPC Indicators plugin?

stealthAngel:
video link should be fixed now

stealthAngel:
you're right.... there is an npc indicator l0l

psikoi:
> you're right.... there is an npc indicator l0l

So I guess this can be closed?

Issue #8733, opened 2019-05-03
Quest List: Add menu options for all available filters


ShaunDreclin:
Having to click through them to get to the one you want isn't great, I'd like to be able to directly select one. (Especially if a fifth filter is added to resolve #8689)

Nightfirecat:
Would a right-click menu solve this issue? Sounds like that would be reasonable UI and UX.

ShaunDreclin:
Yee thats what I meant by menu options, should have been clearer there

Issue #8689, opened 2019-04-30
Hide completed quest functionality removed


Alexgopen:
**Describe the bug**
The functionality to hide completed quests was removed in this commit:
https://github.com/runelite/runelite/pull/8354

This seems like an oversight, and now your only options are displaying All Quests, ONLY Completed, ONLY In Progress, or ONLY Not Started.

Previously you could toggle to hide the completed quests so that you could work down your quest list, and this feature was eventually added to RS2/3 at one point.  

**To Reproduce**
1. Go to the quest tab
2. Click on the button that replaced the old hide completed button
3. You no longer have the ability to hide completed quests

**Expected behavior**
The previously existing hide completed quests feature should still be available.  It could be re-implemented as an additional state that you can toggle to with the new button.

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.5.21
 - Launcher version: 1.6.0

ShaunDreclin:
On the UI side, the option could be called "Show Not Completed" and have the icon be a half red and half yellow circle

PR #8992, opened 2019-06-02
Questlist Fixes

andmcadams:
Adds a "Not Completed" filter (#8689).
Creates a right click menu to select the filter (#8733).

Marking as WIP because:

Temporary `MINIMAP_ORB_HITPOINTS_VENOM` as the icon for the "Not Completed" filter. I'm not sure how to approach making a half red, half yellow icon (as suggested in #8689) since it looks like the whole icon has to be used.
Temporary "Show Next" action. It seems like it would be more helpful to give the name of the next filter that will be selected if you click, but if the colors of the icons are not explanatory for the filter then it doesn't seem clear which filter you are currently on. Mostly a problem for the "All" and "Not Completed Filters".
I left the "Toggle" verb in the menu as it was there previously. It doesn't make as much sense now with the menu in its current usage, but I'm not sure whether there's a preference to how it's changed.

EDIT: Added images below
Screen Shot 2019-06-02 at 8 11 00 AM The temporary "Not Completed" icon
Screen Shot 2019-06-02 at 8 10 19 AM Right click while on "All" filter
Screen Shot 2019-06-02 at 8 10 41 AM Right click while on "In Progress" filter.


andmcadams:
The "Toggle" verb was removed so now each menu option says "Show X".

andmcadams:
Out of date due to quest list changes (and probably many other updates)

PR #8996, opened 2019-06-02
Make skill calculator fields update in real time

### LINKED ISSUE #6667: When filter applied to stats calculator and skill button is clicked, keep filter applied

rbbi:
Uses the EventBus to listen for ExperienceChanged events and updates the "Current XP" and "Current Level" field values. I have added a configuration option "Real time updates" to toggle this behaviour.

Rewritten version of PR 7065

Fixes #6667

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018, Kruithne <kruithne@gmail.com>
  * Copyright (c) 2018, Psikoi <https://github.com/psikoi>
+ * Copyright (c) 2019, rbbi <31489752+rbbi@users.noreply.github.com>
  * All rights reserved.
  *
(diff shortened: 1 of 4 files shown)


deathbeam:
I don't think this really needs to be configurable

rbbi:
> I don't think this really needs to be configurable

Removed

PR #9010, opened 2019-06-03
mining plugin: draw overlay at correct height for mlm and amethyst

Twiglet1022:
It looked weird having the timers appear at floor level while the mining icons in MLM appear at height offset of 150. So I added support for an offset and used it to shift the overlay for veins in MLM to match the MLM plugin.

Did amethyst too because it looks nicer this way IMO.

MLM before:

MLM after:

Amethyst after:

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
@@ -27,6 +27,8 @@ package net.runelite.client.plugins.mining;
 import com.google.common.collect.ImmutableMap;
 import java.time.Duration;
 import java.util.Map;
+import lombok.AccessLevel;
+import lombok.Getter;
 import static net.runelite.api.ObjectID.ROCKS_11161;
 import static net.runelite.api.ObjectID.ROCKS_11360;
 import static net.runelite.api.ObjectID.ROCKS_11361;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
```diff
@@ -46,52 +46,52 @@
 
 enum Rock
 {
-	TIN(Duration.ofMillis(2300), ROCKS_11360, ROCKS_11361),
-	COPPER(Duration.ofMillis(2200), ROCKS_11161),
-	IRON(Duration.ofMillis(5300), ROCKS_11364, ROCKS_11365)
+	TIN(Duration.ofMillis(2300), 0, ROCKS_11360, ROCKS_11361),
```
Nightfirecat: Instead of adding `0` for each rock which doesn't need any offset, create a ctor overload which calls the new ctor with `0` height.

Twiglet1022: Is this possible? Since the constructor takes `(Duration, int, int...)`, how would it know that the second parameter is missing? It would just treat the first `int` from what is meant to be the `int... ` part as the second parameter instead wouldn't it?

Nightfirecat: Instead of removing the existing ctor, add a second one. Then they should look like this:

```java
Rock(Duration respawnTime, int... ids)
{
	this(respawnTime, 0, ids);
}
Rock(Duration respawnTime, int zOffset, int... ids)
{
	this.respawnTime = respawnTime;
	this.zOffset = zOffset;
	this.ids = ids;
}
```

You can see this behavior of defining multiple overloaded constructors in other classes like [AgilityShortcut](https://github.com/runelite/runelite/blob/c1b0ca4ffdf95e6e6979cdf2058b4522530f742a/runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java#L230-L242)

Issue #8176, opened 2019-03-11
Bring unknown ping worlds to bottom of world list when sorting in ascending


AdamtheNoob:
Right now you need to scroll down and either click on the quickly rising world with lowest ping or wait for all the worlds to be pinged.

Seven-Ate:
I was thinking about this yesterday; I think a better solution is to postpone resorting of the list until after all worlds have updated or to not clear out a world's ping until the new ping is found. I'm not very comfortable in async programming and having trouble implementing either solution, however.

Shingyx:
I've done this as part of https://github.com/runelite/runelite/pull/9020.

abextm:
put `Closes #8176` in your pr's description, rather than linking to it here

PR #9020, opened 2019-06-04
Improve world hopper sorting

Shingyx:
Put worlds with unknown ping at the bottom of the list and fix secondary sorting when sorting by activity, e.g. when sorting by "Ping" and then by "Activity".

Here's a couple of images comparing the results of sorting by "Ping" and then by "Activity":

Before | After
:-------------------------:|:-------------------------:
bad | good

Closes https://github.com/runelite/runelite/issues/8176

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.worldhopper;
 
+import com.google.common.collect.Ordering;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
@@ -33,6 +34,7 @@ import java.awt.event.MouseEvent;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
 import lombok.AccessLevel;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldTableRow.java
```diff
@@ -232,6 +232,11 @@ int getPing()
 		return ping;
 	}
 
+	boolean worldHasActivity()
+	{
+		return !world.getActivity().equals("-");
```
Shingyx: Another option is to remove this function and have the following in the other file instead.

```java
String a1 = r1.getWorld().getActivity();
String a2 = r2.getWorld().getActivity();
return a1.equals("-") == a2.equals("-")
	? a1.compareTo(a2) * (ascendingOrder ? 1 : -1)
	: a1.equals("-") ? 1 : -1;
```

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
```diff
@@ -159,24 +159,24 @@ void updateList()
 			switch (orderIndex)
 			{
 				case PING:
-					return Integer.compare(r1.getPing(), r2.getPing()) * (ascendingOrder ? 1 : -1);
+					// Leave worlds with unknown ping at the bottom
+					return r1.getPing() > 0 == r2.getPing() > 0
```
raiyni: This should be abstracted out into logic if statements IMO instead of nested ternaries.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
```diff
@@ -159,24 +159,24 @@ void updateList()
...
+						? Integer.compare(r1.getPing(), r2.getPing()) * (ascendingOrder ? 1 : -1)
+						: r1.getPing() <= 0 ? 1 : -1;
 				case WORLD:
 					return Integer.compare(r1.getWorld().getId(), r2.getWorld().getId()) * (ascendingOrder ? 1 : -1);
 				case PLAYERS:
 					return Integer.compare(r1.getUpdatedPlayerCount(), r2.getUpdatedPlayerCount()) * (ascendingOrder ? 1 : -1);
 				case ACTIVITY:
-					return r1.getWorld().getActivity().compareTo(r2.getWorld().getActivity()) * -1 * (ascendingOrder ? 1 : -1);
+					// Leave empty activity worlds on the bottom of the list
+					return r1.worldHasActivity() == r2.worldHasActivity()
```
raiyni: This should be abstracted out into logical if statements IMO instead of nested ternaries.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
chivesrs: runelite-client can use Guava, so https://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/collect/ComparisonChain.html would make this super clear what's going on.

Adam-:
This looks like a nice cleanup. Can you split the behavior change (which I think is just the ping sorting?) into a separate commit from the order refactoring?

PR #9021, opened 2019-06-04
Inventory Grid Plugin: Add color configuration

### LINKED ISSUE #13705: Allow the player to change the colours used in the Inventory Grid plugin
I think the grid is too bright, it's distracting and can make it harder to see where I'm dragging things.

I would like to be able to change the hex colour of the grid (both the white one and the green one), much like you can with other plugins.

I don't think there are any alternatives.

jplsek:
A friend wanted to change the color to something else, so allow changing the color of the grid and highlight.

(I think the original PR was going to have an option to change the color, but changing the alpha didn't exist at the time, so it was not put in. And someone will probably make a PR like this eventually.)

Closes #13705

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
@@ -24,6 +24,8 @@
  */
 package net.runelite.client.plugins.inventorygrid;
 
+import java.awt.Color;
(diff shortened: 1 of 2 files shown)


Nightfirecat (commented):
Change LGTM, but I'd rather we didn't reorder the configs from their current order. Presently, they appear as:

1. Drag delay
2. Highlight background
3. Show grid
4. Show item

Ideally, the added colors should appear between "Show grid" and "Show item"

jplsek:
Updated!

PR #9023, opened 2019-06-04
mining plugin: remove progress pie from mlm veins that respawned early

Twiglet1022:
Veins in MLM are on a variable timer. So far the longest and shortest respawns that I've witnessed have been +/- 15 seconds from the 108 seconds set in the plugin. This PR addresses one half of that by removing the progress pie once the vein has respawned.

Example of what I'm talking about, which this PR fixes:

Vein respawns before pie disappears

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningOverlay.java
@@ -43,6 +43,12 @@ import net.runelite.client.ui.overlay.components.ProgressPieComponent;
 
 class MiningOverlay extends Overlay
 {
+	// Range of Motherlode vein respawn time - not 100% confirmed but based on observation
+	static final int ORE_VEIN_MAX_RESPAWN_TIME = 123;
+	private static final int ORE_VEIN_MIN_RESPAWN_TIME = 90;
+	private static final float ORE_VEIN_RANDOM_PERCENT_THRESHOLD = (float) ORE_VEIN_MIN_RESPAWN_TIME / ORE_VEIN_MAX_RESPAWN_TIME;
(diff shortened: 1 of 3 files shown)


Twiglet1022:
Added a second commit that recolours the progress pie in MLM during the random segment. I thought green would look nice to indicate that it's close to respawning. It looks like this:

image

So far the 108 +/- 15 seconds seems to be working well. I've seen plenty veins respawn at exactly the 108 + 15 second mark and none respawn after that, but just for fun I'm gonna tweak the plugin to log all the respawn times around me and see what I get.

Twiglet1022:
I wrote something crude (it breaks if I move) to log the respawn times at MLM and collected 217 data points over the span of 40 mins. I don't know what's going on with the outliers, but for the vast majority of veins the respawn times lie in the range of 90 seconds to 122 seconds.

The shortest time I recorded was 89991 ms and the longest 121980ms, barring the outliers.

image

So I've tweaked the range from 93-123 seconds to 90-122 seconds based on what I've seen here.

JacobThompson:
it probably depends on the type of ore mined. can we compare datasets between the types of ore mined with this dataset to produce better results? i’m guessing the outliers are for runite ore.

Adam-:
Instead of keeping a list of respawned veins to skip, what about just removing them from the respawn list

Twiglet1022:
That makes a lot more sense, can't believe I missed the obvious solution there lol. I've redone the first commit to use that approach.

Twiglet1022:
> it probably depends on the type of ore mined. can we compare datasets between the types of ore mined with this dataset to produce better results? i’m guessing the outliers are for runite ore.

I don't know what ore was mined. I was just tracking veins spawning around me to collect their respawn times.

I don't think the client even knows what the ore is until it's deposited, so even if it was related to the ore then that information wouldn't be of any use.

PR #9024, opened 2019-06-05
interface styles: add RuneScape cross sprites option

devLotto:
Adds an option to Interface Styles that replaces left-click cross sprites with the ones in RuneScape.

k6Vg6XB4v1

yNQOm4HcA9


runescape-api/src/main/java/net/runelite/rs/api/RSClient.java
```diff
@@ -984,4 +984,8 @@
 	@Import("healthbarCache")
 	@Override
 	RSNodeCache getHealthBarCache();
+
+	@Import("crossSprites")
+	@Override
+	SpritePixels[] getCrossSprites();
```
Adam-: This can be `RSSpritePixels[]`

PR #9027, opened 2019-06-05
Add additional info to waterbirth coordinate clue

### LINKED ISSUE #9001: Master clue step at Askeladden should tell you to bring a pet rock and a rune thrownaxe.
image
Image of the clue location and hint.

jkybtw:
Fixes #9001.


deathbeam:
Maybe adding item requirement support would make sense instead?

jkybtw:
> Maybe adding item requirement support would make sense instead?

Technically the items arent required as someone else could let you in. I see this as the same additional info just like how fairy ring codes are.

PR #9047, opened 2019-06-07
EntityHider: prevent random config updates

Hermetism:
Prevent entity hider from updating its config when updating a different config group


Adam-:
This change seems fine to me - if you wouldn't mind reopening?

PR #9061, opened 2019-06-08
Added priority '1' to TRAINING_FISHING DiscordGameEvent - solves issu…

vardy:
## Issue

Issue #3323

Discord rich presence displays 'Training: Strength' or sometimes 'Training: Agility' whilst player is barbarian fishing due to the lack of increased priority for the fishing skill in the game event configuration.

## Fix

- Increased priority of TRAINING_FISHING type to 1.

## Caveats

I can't think of any other time fishing exp is gained as a byproduct of another action that should instead be prioritized however I could always be wrong - in which case a more elegant solution may be required.


vardy:
@deathbeam maybe but that could create exceptions - like for slayer where you can get more strength exp in one hit - which doesn't seem better than the current system

PR #9070, opened 2019-06-09
Show HA profit while casting HA on items

jkybtw:
d27dRNGBf5

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
@@ -96,6 +96,8 @@ class ItemPricesOverlay extends Overlay
 		// Tooltip action type handling
 		switch (action)
 		{
+			case ITEM_USE_ON_WIDGET:
+				if (!menuEntry.getTarget().contains("High Level Alchemy") || !config.showAlchProfit()) return null;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -96,6 +96,8 @@ public Dimension render(Graphics2D graphics)
 		// Tooltip action type handling
 		switch (action)
 		{
+			case ITEM_USE_ON_WIDGET:
+				if (!menuEntry.getTarget().contains("High Level Alchemy") || !config.showAlchProfit()) return null;
```
deathbeam: do not use single line ifs. and also, use break; and not early return

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -96,6 +96,9 @@ public Dimension render(Graphics2D graphics)
 		// Tooltip action type handling
 		switch (action)
 		{
+			case ITEM_USE_ON_WIDGET:
+				if (!menuEntry.getTarget().contains("High Level Alchemy") ||
```
deathbeam: By not using single line ifs I did not meant splitting single line if to 2 lines, I meant use { } like we do everywhere else

PR #9074, opened 2019-06-09
Allow Config of the Drag Delay for the Inventory Grid plugin

MrMagaw:


runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
```diff
@@ -60,4 +60,14 @@ default boolean showHighlight()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "dragDelay",
+		name = "Drag Delay",
+		description = "Number of ticks to wait after item press before showing grid"
```
Nightfirecat: This would not make sense to players, as players will interpret "ticks" to mean server ticks, which would be 0.6 sec per tick. I'd much prefer to see this changed to be measured in ms and compare durations against that.

MrMagaw: I totally agree, I didn't actually realize it wasn't in server ticks myself.

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridPlugin.java
```diff
@@ -43,12 +48,19 @@
...
 	@Inject
 	private OverlayManager overlayManager;
 
+	@Getter(AccessLevel.PACKAGE)
+	private int dragDelay;
+
 	@Override
 	public void startUp()
 	{
+		onConfigChanged(null);
```
Nightfirecat: Don't call subscriber methods directly. Create some method to be called from here and from `onConfigChanged()` which contains the code you want to run instead.

Adam-:
I don't see why you can't just access the config from the overlay, and do the conversion to client ticks there.

MrMagaw:
@Adam- I chose to persist the calculated value rather than recalculate every tick where you're dragging to speed things up at the cost of an int in memory. I guess it's not really needed, so if you want me to change it, I can.

Issue #9042, opened 2019-06-06
Mining plugin missing sandstone & granite


AmazingJesus:
Mining plugin, add missing ores, people can comment missing?

I noticed sandstone and granite is missing.

Adam-:
related: it seems coal outside of the p2p area of the mining guild has wrong times, too.

PR #9076, opened 2019-06-09
mining plugin: correct coal timer, missing copper rock and align other timers to game ticks

Twiglet1022:
Fixes issue from this comment: https://github.com/runelite/runelite/issues/9042#issuecomment-499691513

Wiki says 30 seconds, but on testing it out personally I found the respawn time to be 29.4 seconds.

All the timers should be multiple of 0.6 seconds to match game ticks so I also corrected those and overlays now match up perfectly with the ores respawning.

Also came across this copper rock which wasn't yet in the plugin while testing the copper rock timer.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
@@ -29,6 +29,7 @@ import java.time.Duration;
 import java.util.Map;
 import lombok.AccessLevel;
 import lombok.Getter;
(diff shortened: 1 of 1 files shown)


Adam-:
Did you test this outside of the guild, in the f2p area, and in the p2p area? I think that the area the plugin considers for the faster respawn rate is not correct.

Twiglet1022:
Yes I did. The plugin considers region 12183 to be the p2p area of the mining guild, and based on the map that looks correct. Region 12183 starts 1 tile north of the northern-most rock in the p2p section of the mining guild.

image

PR #9078, opened 2019-06-10
Build RuneLite with Gradle

MarkKoester:
Using Gradle instead of Maven to build RuneLite reduces project build times and add more flexibilty for customizing the build process. It also offers a better user-experience than Maven since it has a more modern, discoverable CLI, and it ditches XML for its build files and uses a Groovy-based DSL instead.

A longer explanation of the benefits of using Gradle over Maven can be found in the [Gradle vs Maven comparison](https://gradle.org/maven-vs-gradle/) section of the Gradle docs.

## Results
Below is a comparison of build times using both Gradle and Maven. Results were obtained from a clean build and from an incremental build, where only one source file was changed.

Tests were skipped in all cases.

### Clean build
| Build system     | Time (minutes )|
|------------------|----------------|
| Maven            | 1:27           |
| Gradle           | 1:32           |

### Incremental build
| Build system     | Time (minutes )|
|------------------|----------------|
| Maven            | 1:03           |
| Gradle           | 0:18           |

Although build times are approximately the same for clean builds, incremental builds are significantly faster with Gradle. 🚀

## Change summary
- Remove `pom.xml` from all projects
- Add `build.gradle` to all projects
- Add package name to antlr grammer
- Remove cache test dependency on itself
- Remove `cache` project dependency from CacheClientTest.java
- Upgrade mapstruct to 1.3 (Needed to work with jdk8 and the current version is deprecated)
- Moved Checkstyle files to the `checkstyle` directory to work with gradle defaults
- Change path to suppressions.xml in `checkstyle.xml`
- Using palantir:git-version plugin to get git info (http-api)
- Changed some property names since Gradle interprets names separated by `.` to be accessing an object
- The `User-Agent` header string in the `http-api` library (`RuneLiteAPI.java`) was altered slightly since we can no longer use the Maven plugin for retrieving git information. It still includes all the same information as before (the commit hash and an indicator for whether the git tree is dirty), but it's formatted slightly differently. If the header needs to remain exactly the same it'll require a bit more effort, but if the header is just being logged for reproducing issues it should be fine.


abextm:
We cannot effectively take contributions for things like this because it requires a large change of our internal build, build plugins, release process, etc.

abextm:
I also don't see where the script assembler is called, and its still packaged as a maven plugin, so I doubt this builds the client completely.

MarkKoester:
You're right, I missed that part. I'm going take a look at it.

I understand that this is a large change; I'm hoping that this can at least be a starting point if this is ever something that you want to tackle in the future. I feel like the benefits are well worth it.

MarkKoester:
I was able to convert the script-assembler-plugin to a gradle plugin fairly easily. I'm going to try to push my changes tomorrow.

I also noticed that the gradle build isn't creating a fat jar for the client, so I'll add that as well.

Adam-:
The only real issue I have with Maven is the inability to lock dependencies. We have mostly mitigated this on the release by locking them in the launcher. Other than the incremental compilations (which, apparently Maven has support for but I'm unsure if it works) - what else does Gradle give us?

N.B. if you are just editing the client, Intellij and most IDEs can incrementally compile just that without having to rerun Maven, which is what I generally do. You only have to rerun Maven if your dependencies change.

MarkKoester:
That's a good point about compiling via IntelliJ, that seems pretty useful. Is it reasonable to expect everyone to work that way? I feel like command line users would benefit from Gradle's CLI.

Setting performance aside, I think there's a huge benefit in being able to replace the XML build files with Gradle's DSL. There's a lot less noise in the build files and (in my opinion) it makes it a lot easier to understand what's going on during the build. It should also make it a lot easier to change the build process if changes ever need to be made in the future.

I also think Gradle's build scans can be useful. Here's a [sample build scan](https://scans.gradle.com/s/g3b5vtcpzhegk) from my builds yesterday. It could potentially make it a lot easier to troubleshoot build issues and to profile the build to improve it further.

Gradles's wrapper scripts can be helpful as well for making sure that everyone builds the project with the same Gradle version.

PR #9083, opened 2019-06-10
boosts plugin: disable notification with threshold -1 instead of 0

### LINKED ISSUE #8961: Optional Boost Threshold being able set to 0 (checkbox)
The feature in Runelite allows you to be notified when boosted stats are down, however, when set to 0 the feature is disabled. Being able to be notified at 0 would be helpful for players with prayer who used "preserve" to know when to turn their prayer off.

Since many features in runelite are disabled by setting to  0, I was considering adding a checkbox underneath the threshold amount indicator to say "set threshold to 0" or something similar.

As of now, I  set my  threshold to 1 and get notified when my boost is 1 away from being 0. 

I think this could be a good update for people trying to reach maximum efficiency doing tasks such as woodcutting with a dragon axe spec (and preserve prayer).  Perhaps consider adding the ability to set a notification threshold to 0 for other stats such as Prayer (in the idle notification). Stats like special attack make sense to not be notified at 0, but others could be really helpful! Thanks for reading.

Twiglet1022:
Closes #8961

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.boosts;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
 
 @ConfigGroup("boosts")
 public interface BoostsConfig extends Config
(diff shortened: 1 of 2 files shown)


Unmoon:
Maybe this should support negative numbers too, for people who boost down their stats? And instead have a checkbox for the toggle.

Twiglet1022:
I can’t really think of a use for setting it to be negative. What do you mean by “people who boost down their stats”?

Unmoon:
Stats can be negative, and you can do that with for example brews, wine or beer. People who want to hit 0's will want to keep their attack/strength at 0.

Adam-:
This will send a whole bunch of notifications to a whole bunch of people who probably don't want them.

Twiglet1022:
Closing for now. I'll reopen if I ever get round to doing this in a better way.

PR #9092, opened 2019-06-12
emoji: respect chat colour tags

### LINKED ISSUE #8842: Emoji plugin disregards chat recolor
**Describe the bug**
When typing emojis in clan chat, the colour of the message gets reset, I've got my clan chat colour set to blue but it gets reset if someone uses an emoji in a message

**To Reproduce**
1. Enable emoji plugin
2. Edit chat color plugin and change clan chat colour
3. Type some words in clan chat
4. Type some words followed by an emoji in clan chat

**Expected behavior**
Both lines of text should be the colour you chose

**Screenshots**
https://i.imgur.com/KUOEVJX.png

Hydrox6:
Fixes #8842

Makes the plugin parse the message with its tags, and make the replacement work such that it doesn't overwrite any tags that may be there.

`removeTags` has been changed to preserve the `<gt>` and `<lt>` tags. This also has the side effect of making `Chat Notifications` able to highlight `<gt>` and `<lt>`

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/emojis/Emoji.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/emojis/Emoji.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.emojis;
 import com.google.common.collect.ImmutableMap;
 import java.awt.image.BufferedImage;
 import java.util.Map;
+import lombok.Getter;
 import net.runelite.client.util.ImageUtil;
 
 enum Emoji
@@ -83,6 +84,7 @@ enum Emoji
 
 	private static final Map<String, Emoji> emojiMap;
 
+	@Getter
 	private final String trigger;
 
 	static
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Definitely feels to me like we should consider just changing `Text::removeTags` given that `<gt>` and `<lt>` are simply representations of characters rather than actual tags.

raiyni:
I think `removeTags` should be indiscriminate. Otherwise it turns into a different function unless you pass `allowable tags` to it

Adam-:
I agree we shouldn't make removeTags only remove some tags. What is the cause of the problem you are trying to fix?

Hydrox6:
When stripping tags, it'll remove `<gt>` and `<lt>`, mutilating chat messages that contain them. This causes problems for the Emoji plugin, as a number of Emoji have `<gt>` and `<lt>` in them. It also affects Chat Notifications, as it is unable to highlight `<gt>` and `<lt>`

Adam-:
So - the reason the colors are not taking effect is the emoji plugin is stripping tags? I don't see it doing that anywhere now.

PR #9099, opened 2019-06-13
Swapping 'wear' and 'wield' with 'use' on all mta alchemy room items

robinwithes:
For those unfamiliar with the alchemy room in the mage training arena: You are supposed to cast high alchemy on a random item specific to the arena that switches every 30 seconds, 4 out of 5 of those items are equippable but you get a message saying you can't when you try it (This option cancels all actions until you click continue). This menu swapper entry addition makes it less annoying by swapping the equip option with the use option.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -28,7 +28,6 @@ package net.runelite.client.plugins.menuentryswapper;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
-import java.util.Arrays;
 import java.util.Set;
 import javax.inject.Inject;
 import lombok.Getter;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -25,8 +25,10 @@
  */
 package net.runelite.client.plugins.menuentryswapper;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
+import java.util.Arrays;
 import java.util.Set;
 import javax.inject.Inject;
 import lombok.Getter;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Wouldn't this change make it more annoying? I'd rather try to equip something and not be stuck with a "Use ITEM -> " menu option activated if I misclicked.

robinwithes:
> 
> 
> Wouldn't this change make it more annoying? I'd rather try to equip something and not be stuck with a "Use ITEM -> " menu option activated if I misclicked.

No because the only reason to be there is high alching and would you rather high alch something with a equip option or use option? This change makes it so you can simply spam click it without it opening the "Can't equip" menu and pausing all gameplay.

Nightfirecat:
Ah, it interrupts the alchemy, that's why. :+1:

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -74,6 +76,8 @@
 	private static final String CONFIG_GROUP = "shiftclick";
 	private static final String ITEM_KEY_PREFIX = "item_";
 
+	public static final ImmutableList<Integer> MTA_ITEMS = ImmutableList.of(6893, 6894, 6895, 6897); //item id's of mage training arena alchemy room items
```
Nightfirecat: `private`, and reference `ItemID` definitions instead of using hardcoded ints.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -569,6 +573,10 @@ else if (config.swapTeleportItem() && option.equals("wear"))
 			swap("rub", option, target, true);
 			swap("teleport", option, target, true);
 		}
+		else if (config.swapMtaItems() && MTA_ITEMS.contains(eventId) && Arrays.asList("wield", "wear").contains(option))
```
Nightfirecat: Don't construct an array each time the full conditional runs. Either define the array as a static variable, or use two separate comparisons. (ie. `("wield".equals(option) || "wear".equals(option))`)

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -569,6 +576,10 @@ else if (config.swapTeleportItem() && option.equals("wear"))
 			swap("rub", option, target, true);
 			swap("teleport", option, target, true);
 		}
+		else if (config.swapMtaItems() && MTA_ITEMS.contains(eventId) && ("wield".equals(option) || "wear".equals(option)))
```
Nightfirecat: I guess it should be `(option.equals("wield") || option.equals("wear"))` for consistency with the other conditional checks in this chain. My bad for recommending the other way around first.

PR #9110, opened 2019-06-15
Adjust alignment of icons in 2005 interface style and fix inventory icon

sadegr:
I noticed that the alignment of a few of the icons in the 2005 interface style were off by a few pixels (most notably the music icon being too far down and the log out icon being too far to the right). I've gone ahead and updated the alignments of all of them to match the following screenshot (supposedly taken in 2006, but the HUD did not change between 2005/2006 AFAIK):
TrYePoN 1
[Source](https://www.reddit.com/r/2007scape/comments/2me0bd/can_we_revisit_the_2006_client_for_the_2nd/)

Also, the inventory icon currently used in the 2005 interface is much smaller than it was originally in 2005, so I've updated the sprite. The new sprite was ripped from the above screenshot as well (both the screenshot and sprite are PNGs). I hope there aren't any copyright issues in doing so, but since the other 2005 interface icons were presumably also either ripped from old screenshots or obtained from some client cache, I figured this would be ok.

Before:
before

After:
after


Adam-:
I have noticed this too, particularly the inventory icon. I'm not sure where we got our existing icon but I think it was from the existing cache? Maybe someone can dump it from a 317 cache and we can compare what we have now with that. CC @Joshua-F

Joshua-F:
Here's all of the images directly exported from a (close to) 317 cache.

sprite_21_12
sprite_21_0
sprite_21_1
sprite_21_2
sprite_21_3
sprite_21_4
sprite_21_5
sprite_21_6
sprite_21_7
sprite_21_8
sprite_21_9
sprite_21_10
sprite_21_11

[sideicons.zip](https://github.com/runelite/runelite/files/3292769/sideicons.zip)

sadegr:
Thanks @Joshua-F . I've used the inventory icon from that zip file, although I had to make a few adjustments to it:
1. It had a much higher DPI than the other HUD icons in the RuneLite repo, so I've adjusted the DPI to match the others
2. It had a different resolution than the other icons in the RuneLite repo, causing the client to display it scaled up. I've changed the resolution to match the others

Screenshot:

Issue #9011, opened 2019-06-03
Item Prices plugin doesn't show High Alchemy profit unless High Alchemy values are enabled


fuzxi:
**Describe the bug**
Using the Item Prices plugin, the tooltip doesn't show High Alchemy profit unless High Alchemy value is also enabled.

**To Reproduce**
1. Enable Item Prices plugin.
2. Enable High Alchemy Values and High Alchemy Profit options.
3. Hover over any alchable item. Value and profit will be shown.
4. Disable High Alchemy Values option.
5. Hover over any alchable item. Neither value nor profit will be shown.

**Expected behavior**
The tooltip should show the High Alchemy profit. Showing both the value and the profit is often redundant; showing just the profit is preferable.

**Screenshots**
![2019-06-03_15-29-35](https://user-images.githubusercontent.com/33790641/58838884-da36c080-8614-11e9-8e1b-80a671e75dd2.png)
![2019-06-03_15-34-01](https://user-images.githubusercontent.com/33790641/58838998-26820080-8615-11e9-8e7e-2885cbd227b2.png)

**Environment (please complete the following information):**
 - OS: Void Linux 4.19.46_1
 - RuneLite version: 1.5.25
 - Launcher version: 1.6.2

Serpa88:
#9111

PR #9111, opened 2019-06-15
Fix high alch profit

Serpa88:


Adam-:
What is this fixing exactly?

Serpa88:
Fixes #9011

gregg1494:
#9014

Adam-:
Okay. I took #9014 instead.

PR #9112, opened 2019-06-15
clues: correct falo the bard typos

Twiglet1022:
The osrs wiki had the wrong text (hope -> hopes). Here is an image of the clue in game after correcting the text in the plugin:

image

Found another one, wiki text was missing a comma. Updated:

image


Twiglet1022:
Found another one so just added another commit

PR #6977, opened 2018-12-16
items stats: add option to show weight

Dava96:
items stats: add option to show weight
resolves #6939

PR #9114, opened 2019-06-15
Added ability to toggle on/off the item weight tooltip in the Item Stats Plugin

Adelaidian:
Added ability to toggle on/off the item weight tooltip in the Item Stats Plugin. Having the weight displayed for every item in the game can be annoying, so this modification to the Item Stats plugin will allow users to toggle the weight tooltip off.


Twiglet1022:
#6977 already does this

PR #9123, opened 2019-06-16
mining plugin: only show timers on same level in motherlode mine

### LINKED ISSUE #9120: Mining plugin: In MLM, only show respawn times for the ores on the floor that you're on
If I'm mining at the upper level, I don't care about respawn timers for the lower level. These respawn timers clog up the screen and should not be visible at all.

Twiglet1022:
Fixes #9120


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningOverlay.java
```diff
@@ -108,4 +118,16 @@ public Dimension render(Graphics2D graphics)
...
 	}
+
+	/**
+	 * Checks if the given point is "upstairs" in the mlm.
+	 * The upper floor is actually on z=0.
+	 *
+	 * @param localPoint
+	 * @return true if localPoint is at same height as mlm upper floor
+	 */
+	boolean isUpstairsMotherlode(LocalPoint localPoint)
```
Nightfirecat: We already have a similar function available in the MLM plugin. Seems like it might be worthwhile to move it to a common location and reference it from both plugins?

Twiglet1022: Is this what you mean? https://github.com/runelite/runelite/compare/master...Twiglet1022:mlm-same-level-timers-2

PR #9137, opened 2019-06-18
Add bank overlay to show current most expensive items.

TalCohen:
Adds the option to show the most (currently 10, could be configurable) expensive items in your bank, by GE price.

Oftentimes I'm clearing out my bank and am wondering which items are contributing the most to my total. I made this recently and thought I'd share it.

It currently works per tab, or for the whole bank if "all items" tab is shown, as you'd expect.

Some screenshots:
### Random Tab
image

### Another Random Tab
image

### All Items Tab
image

### Removing an item
image

### Tab with less than 10 items with a GE price
image

### The New Bank Config (I don't mind changing the config entry name)
image

UI code is not my strong-suit, but it seems to work fine. Someone better at it can come along and redesign it, if they'd like.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
@@ -96,7 +96,8 @@ class BankItemsOverlay extends Overlay
 
 		panelComponent.getChildren().clear();
 
-		if (expensiveItems == null || bankCalculation.isBankDifferent(bankItems)) {
+		if (expensiveItems == null || bankCalculation.isBankDifferent(bankItems)) 
+		{
 			expensiveItems = getExpensiveItems(bankItems);
 			Arrays.sort(expensiveItems, Comparator.comparing(ItemPrice::getPrice).reversed());
 		}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2019, Tal Cohen <https://github.com/talcohen>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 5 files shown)


BrendanJones44:
Wow, this is such a cool idea! Thanks for putting this together for the community :D

raiyni:
What does this look like in fixed mode

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
```diff
@@ -0,0 +1,166 @@
...
+		return panelComponent.render(graphics);
+
+	}
+
+	private BufferedImage getImage(Item item)
+	{
+		return itemManager.getImage(item.getId(), item.getQuantity(), true);
+	}
+
+	private ItemPrice[] getExpensiveItems(Item[] items)
```
raiyni: This should be calculated in the plugin.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
```diff
@@ -0,0 +1,166 @@
...
+				if (itemPrice.getItem().getQuantity() > 0)
+				{
+					final BufferedImage image = getImage(itemPrice.getItem());
+
+					if (image != null)
+					{
+						ImageComponent imageComponent = new ImageComponent(image);
+						panelComponent.getChildren().add(imageComponent);
+
+						TitleComponent titleComponent = TitleComponent.builder()
```
raiyni: Why is this a title component

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
```diff
@@ -0,0 +1,166 @@
...
+		panelComponent.getChildren().clear();
+
+		if (expensiveItems == null || bankCalculation.isBankDifferent(bankItems)) 
+		{
+			expensiveItems = getExpensiveItems(bankItems);
+			Arrays.sort(expensiveItems, Comparator.comparing(ItemPrice::getPrice).reversed());
+		}
+
+
+		for (int i = 0; i < ITEM_SIZE; i++)
```
raiyni: Why does maintaining a 10 item size matter?

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
TalCohen: Why not?

As I mentioned above, UI isn't a strong-suit of mine. In fact, I have zero experience with Java AWT other than this change. With that said, I think it would be more helpful to explain why you think a `TitleComponent` isn't the correct solution here and which alternative you prefer. That would really help me out, along with any other readers with similarly little experience with the code base.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
TalCohen: It doesn't. It's just a random number I chose, and mentioned it can be configurable if we thought appropriate.

Some of this logic is from how I saw it was happening in `InventoryViewerOverlay`, so I carried it over. I decided to keep the idea of having a non-resizing panel, which is what this (and the placeholder additions below) tries to accomplish.

If you can think of a more intuitive design, I'd be more than happy to change it.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
TalCohen: Why is that? This is the only place it's used.

TalCohen:
> What does this look like in fixed mode

Ehh, not the best:

image

The panel is moveable, but not resizable right now. I'd love to learn a better way to present this information on screen, though.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankItemsOverlay.java
Nightfirecat: Because you're running this one per frame in the overlay, when it only needs to be run once each time the bank is updated.

TalCohen: FWIW, it *is* run only each time the bank is updated

Nightfirecat: You're right, though it is comparing the bank contents to the previously stored bank contents every frame the bank is open.
The point is, raiyni suggested this be tracked and calculated in the plugin because you then have access to `@Subscribe`r methods for event-based programming, rather than being fully declarative in this overlay. That would have the benefit of running this code (and the bank comparison) exactly one time upon changes instead of 50x/second.

raiyni:
Better served as a hub plugin and some existing plugins might already fill this feature.

PR #9140, opened 2019-06-18
Display item value heatmap in bank

### LINKED ISSUE #2062: [Request] List of most valuable items in bank
Similar to runeloader's feature

raiyni:
Closes #2062
Supersedes/Closes #9137
Supersedes/Closes #7129

d1ci3659II


TalCohen:
This is nice, I like it. A few questions:

* What do you envision the final product will look like? I’m assuming the colored circles are placeholders?
* What do the different colors mean and how would users know?

One of the biggest things I wanted was to be able to not only see the most expensive items, but also see their ordering and how much they were actually valued. I wonder if there’s any nice way we could display that information.

marcsasson:
Any progress on this?

runelite-client/src/main/java/net/runelite/client/plugins/bank/HeatmapItemOverlay.java
```diff
@@ -0,0 +1,105 @@
...
+import net.runelite.api.ItemComposition;
+import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.api.widgets.WidgetItem;
+import net.runelite.client.game.ItemManager;
+import net.runelite.client.ui.overlay.WidgetItemOverlay;
+import net.runelite.client.util.ImageUtil;
+
+public class HeatmapItemOverlay extends WidgetItemOverlay
+{
+	private static final int ALPHA = 204;
```
deathbeam: I would maybe reduce the alpha by a bit, at least from your GIF in main post it looks kind of bad

raiyni: Alpha lowered by 30% 
java_9obQZkVktp

raiyni:
This was added as an external plugin.

PR #9158, opened 2019-06-20
Add inventory setup plugin

dillydill123:
inventory_setups
config jpeg

This plugin is designed to help players remove screenshots of their different bossing/farming/skilling setups off their computers and let RuneLite keep track of them instead. I had this idea when I became frustrated trying to remember what my setup was for doing Barrows runs on my Ironman. I was surprised to see something like this didn't already exist. I asked around and there seemed to be a interest for it, so I created it. It's easy to add a new setup. All you do is gear up and press the + button and add a name! This uses the configuration manager to allow these setups to persist passed runelite instances, but do not persist further than that (e.g., not account based).

I have also added optional highlighting to inform users when their inventory is not matching a setup. As of now, the order of the items matters when highlighting, but in the future I may add the ability to disregard the order and only highlight if the item is not there. I don't think this is a major feature, and the plugin is useful in it's current state.

I understand that there is already the bank tags plugin to help remedy the problem I am trying to solve. The biggest difference that this plugin offers is the ability to keep track of things like food and potions and other supplies for an inventory setup. I also think that some will find a visual representation of the setup more appealing than just items in the bank. But personally in my opinion this plugin will work best in tandem with the bank tag plugin, as shown in the gif above.

Thank you for making RuneLite open source and allowing me to make this. I had a lot of fun creating it and showing it to friends. I hope it can be added in!


Adam-:
Historically we have just referenced people to use the bank tag plugin for this, since you can easily tag your current inventory inventory and equipment and then get a nice "tab" view of it. So I'm not sure this adds anything new or not. Maybe this is slightly easier?

As an aside, I think the current UI for this looks pretty clunky and would like to see something more polished if we decided we wanted to add it.

dillydill123:
Thanks for the reply. The only "new" thing it adds would be tracking quantities of supplies. But as you said it may be easier for others than the bank tags.

I'll work on the UI. My plan currently is to make it similar to the Screen Markers plugin. Each setup will have its own block that contains inventory setup specific settings. This would remove the need for a config. From there the user will be able to view a setup via button. I'll have to consider how to improve the panel that actually shows the inventory and equipment. If anyone has any tips I would appreciate them.

dillydill123:
Closing because it’s on the plugin hub now.

PR #9161, opened 2019-06-20
Items Kept - Fix edge case death prices and add test file

TheStonedTurtle:
Updated the death price to ignore the `UntradeableItemMapping` file.

This file is used for getting the liquidation value of items, such as Graceful pieces which can be sold for marks of grace which can be used to buy amylase crystals. This made the pieces value tied to the price of amylase crystals which is incorrect for this use case.

While checking these i noticed skilling outfits have a value offset so i added these as well.

This change does not effect the price calculation for the value widget.


Adam-:
I'd rather see an overload of `getItemPrice` probably.

Nightfirecat:
I've mentioned this in Discord, but I'll mention it here for visibility as well: particularly since we're starting to see more bug reports on specific cases which are incorrectly handled by this plugin, we should start building a test suite to ensure these cases are covered properly. This PR would be a reasonable place to start adding those IMO

TheStonedTurtle:
I added a bunch of tests, all of which are passing, but i had to move the kept/lost filtering logic to its own method to actually test that functionality.

I'll be working on fixing #9167 next, and add tests for those cases.

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -299,65 +326,128 @@ else if (keepCount > 0)
...
+						if (boxableItem.getId() == boxedId)
+						{
+							if (boxableItem.getQty() > 0)
+							{
+								boxableItem.setQty(boxableItem.getQty() - 1);
+								keptItems.add(new TempItem(boxedId, 1));
+							}
+							else
+							{
+								lostItems.remove(boxableItem);
```
Adam-: There is no test coverage of this? This will throw a concurrent modification exception

runelite-client/src/main/java/net/runelite/client/plugins/itemskeptondeath/ItemsKeptOnDeathPlugin.java
```diff
@@ -223,73 +232,92 @@ private void rebuildItemsKeptOnDeathInterface()
...
+	}
+
+	/**
+	 * Calculates which items will be kept/lost. first list is kept items, second is lost.
+	 * @param inv players inventory
+	 * @param equip players equipement
+	 * @return list of items kept followed by a list of items lost
+	 */
+	@VisibleForTesting
+	List<List<ItemStack>> calculateKeptLostItems(final Item[] inv, final Item[] equip)
```
abextm: I'd much rather this return a static inner class

Adam-: You could also then move back `hasAlwaysLost` then too probably

PR #9170, opened 2019-06-21
widget overlay: Don't draw empty wilderness K/D box

Nightfirecat:
The Wilderness and PVP kill-death information box is created while in
these areas regardless of whether the setting (configured via the notice
board at the Edgeville bank) is enabled to show the text, meaning the
widget contains only empty text widgets when the setting is disabled
rather than being null, causing a bounding box to still be drawn and
affecting other snapped widget layout. This commit adds a child class of
WidgetOverlay specific to this widget and prevents it from being drawn
when the setting to show this information is disabled.

Fixes runelite/runelite#6089

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -631,7 +631,16 @@ public enum Varbits
 	 * 0 = on
 	 * 1 = off
 	 */
-	BOSS_HEALTH_OVERLAY(12389);
+	BOSS_HEALTH_OVERLAY(12389),
+
+	/**
+	 * Whether the PVP kill-death stats widget should be drawn while in the wilderness or in PVP worlds.
+	 *
+	 * 0 = Disabled
+	 * 1 = Enabled
+	 */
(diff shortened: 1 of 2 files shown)


deathbeam:
So this does not solve the XP tracker issue right?

Nightfirecat:
As far as I can tell, the XP tracker issue isn't actually an issue anymore. If I set my XP drops duration to 1 second and let it disappear, its widget is fully removed from the widget tree and other overlays in the same snap corner move closer to the corner.

If there's some other issue I'm not thinking of, then probably not. This PR is pretty specifically geared to fix the pvp K/D widget issue.

Adam-:
Is there a reason we cannot just change the widget we are moving to the parent of whatever it is set to now instead?

Nightfirecat:
90.14 is a viewport-sized container, so I don't think we want to make that draggable. That also wouldn't solve the problem since that is also not hidden when the K/D text is hidden.

GIF

Nightfirecat:
Confirming here since I thought I had run into this not working since last commenting here, this PR is still working.

Nightfirecat:
I've updated this PR to be more specific in its approach to fix this issue.

PR #9182, opened 2019-06-22
EmoteClue: Add null checks for stashUnit

Nightfirecat:
Some clues may have item requirements, but no STASH unit, which would
cause NPEs during rendering. This commit also marks stashUnit as
nullable so this issue can hopefully be avoided in the future.


Nightfirecat:
For reference, this can be tested with any of the following clue texts:

- "Cheer at the games room. Have nothing equipped at all when you do."
- "Panic on the pier where you catch the Fishing trawler. Have nothing equipped at all when you do."
- "Panic in the heart of the Haunted Woods. Beware of double agents! Have no items equipped when you do."
- "Bow to Brugsen Bursen at the Grand Exchange."

PR #9186, opened 2019-06-22
tile indicators: fix phantom clicks

### LINKED ISSUE #9179: Tile Indicators Highlight hovered tile phantom clicks

**Describe the bug**
Clicking on a wall with void behind or clicking out of range doesn't produce any input until the cursor hovers a tile in range.

**To Reproduce**
1. Enable **Tile indicators** and **Highlight hovered tile**
2. Go to POH and walk all the way to the border to see the void
3. Click the void
4. Move the cursor onto a tile in range
5. Now the click in the void registers, producing unexpected behavior.

**Expected behavior**
When clicking the void, nothing at all should happen, resulting in the same behavior as not having the Highlight hovered tile enabled.

**Screenshots**
RuneLite_012L3rnVZS

 - OS:  Windows 10
 - RuneLite version: 1.5.27
 - Launcher version: 1.6.0

Hydrox6:
Closes #9179

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Client.java
+++ b/runelite-api/src/main/java/net/runelite/api/Client.java
@@ -1632,4 +1632,6 @@ public interface Client extends GameEngine
 	void draw2010Menu();
 
 	void resetHealthBarCaches();
+
+	void setViewportWalking(boolean walking);
 }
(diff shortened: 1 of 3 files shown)


deathbeam:
This is also used in PartPingOverlay

Hydrox6:
Pinging doesn't cause any phantom clicks, so adding it there would be unneeded

deathbeam:
Did you actually tested that? Because it's using literally the same code

Hydrox6:
I went and tested it, and it doesn't cause Phantom clicks

deathbeam:
Well, it should. You need to actually be in party, or remove the checks around, and also receive pings, or remove that check as well, to verify it.

Adam-:
fixed in a3e9b790dd92d8fe7eb12d35de6db250f1f78328

PR #9190, opened 2019-06-23
Add Tzhaar Timers Timer Plugin

### LINKED ISSUE #1347: Ingame timer for fight caves
A timer that shows how long current fight caves is going on for. Similiar to Raids

Could also add it for inferno...cant think of anything else to add it to

winterdaze-rs:
Closes #1347 

Adds a timer within the Fight Caves and Inferno. Is slightly less accurate when hopping worlds/pausing the caves a lot, but that is to be expected.

Some testing has been done by @Kamielvf [[image]](https://cdn.discordapp.com/attachments/436089071144534016/572620695335337984/unknown.png)


runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersConfig.java
```diff
@@ -0,0 +1,48 @@
...
+	String CONFIG_TIME = "time";
+	String CONFIG_STARTED = "started";
+	String CONFIG_LASTTIME = "lasttime";
+	@ConfigItem(
+			keyName = "tzhaarTimers",
+			name = "Show timer",
+			description = "Display elapsed time in the Fight Caves and Inferno",
+			hidden = true
+	)
+	default boolean tzhaarTimers()
```
Kamielvf: If the plugin only has an on/off state possible, you do not need a config. Anything done on config change can be done on plugin startUp/shutDown.

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,297 @@
...
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_TIME);
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_STARTED);
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_LASTTIME);
+	}
+
+	private void saveConfig()
+	{
+		if (startTime != null)
+		{
+			resetConfig();
```
Kamielvf: I don't think this is needed here because you are overwriting the config already anyway

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,297 @@
...
+@Slf4j
+public class TzhaarTimersPlugin extends Plugin
+{
+	private static final String START_MESSAGE = "Wave: 1";
+	private static final String WAVE_MESSAGE = "Wave:";
+	private static final String DEFEATED_MESSAGE = "You have been defeated!";
+	private static final String INFERNO_COMPLETE_MESSAGE = "Your TzKal-Zuk kill count is:";
+	private static final String FIGHT_CAVES_COMPLETE_MESSAGE = "Your TzTok-Jad kill count is:";
+	private static final String INFERNO_PAUSED_MESSAGE = "The Inferno has been paused. You may now log out.";
+	private static final String FIGHT_CAVE_PAUSED_MESSAGE = "The Fight Cave has been paused. You may now log out.";
```
Kamielvf: You could precompile a regex for these since they are so similar.
```
Your (TzTok-Jad|TzKal-Zuk) kill count is:
The (Inferno|Fight Cave) has been paused. You may now log out.
```

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,297 @@
...
+	{
+		return configManager.getConfig(TzhaarTimersConfig.class);
+	}
+
+	@Override
+	protected void shutDown() throws Exception
+	{
+		removeTimer();
+		resetConfig();
+		startTime = null;
```
Kamielvf: There's a few places these 3 are all nulled, maybe put them in a method and just call that

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,297 @@
...
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_STARTED);
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_LASTTIME);
+	}
+
+	private void saveConfig()
+	{
+		if (startTime != null)
+		{
+			resetConfig();
+			if (lastTime == null)
```
Kamielvf: Is this ever null at all here?

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,297 @@
...
+	private void resetConfig()
+	{
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_TIME);
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_STARTED);
+		configManager.unsetConfiguration(TzhaarTimersConfig.CONFIG_GROUP, TzhaarTimersConfig.CONFIG_LASTTIME);
+	}
+
+	private void saveConfig()
+	{
+		if (startTime != null)
```
Kamielvf: Use an early return instead, so 
```
if (startTime == null)
{
	return;
}

//rest of method
```

Kamielvf:
Side note; maybe this should be in the timers plugin instead of a separate one, but I don't feel too strongly about that.

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
winterdaze-rs: It is null if you haven't logged out/hopped or paused yet I believe, or if you force a log and don't complete the wave

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,279 @@
...
+
+	private Instant startTime;
+
+	private Instant lastTime;
+
+	private Boolean started = false;
+
+	private boolean loggingIn;
+
+	private static final String CONFIG_GROUP = "tzhaartimers";
```
Kamielvf: Any constant should be put at the top with the others

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimers.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.client.ui.overlay.infobox.InfoBox;
+import java.awt.image.BufferedImage;
+import java.awt.Color;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+@Slf4j
```
Nightfirecat: This class does not use any `log` statements, so remove this.

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.regex.Matcher;
+import static net.runelite.api.ItemID.FIRE_CAPE;
+import static net.runelite.api.ItemID.INFERNAL_CAPE;
+
+@PluginDescriptor(
+	name = "Tzhaar Timers",
+	description = "Display elapsed time in the Fight Caves and Inferno",
+	tags = {"inferno", "fight", "caves", "cape", "timer", "tzhaar"}
+)
+@Slf4j
```
Nightfirecat: This class does not use any `log` statements, so remove this.

runelite-client/src/main/java/net/runelite/client/plugins/tzhaartimers/TzhaarTimersPlugin.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.regex.Matcher;
+import static net.runelite.api.ItemID.FIRE_CAPE;
+import static net.runelite.api.ItemID.INFERNAL_CAPE;
+
+@PluginDescriptor(
+	name = "Tzhaar Timers",
+	description = "Display elapsed time in the Fight Caves and Inferno",
+	tags = {"inferno", "fight", "caves", "cape", "timer", "tzhaar"}
+)
+
```
Nightfirecat:
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -128,6 +140,11 @@
 	private int lastAnimation;
 	private boolean loggedInRace;
 	private boolean widgetHiddenChangedOnPvpWorld;
+	private ElapsedTimer tzhaarTimer;
+	private Instant tzhaarStartTime;
+	private Instant tzhaarLastTime;
+	private Boolean tzhaarStarted = false;
```
Nightfirecat: As was discussed on Discord, why is this needed in addition to `tzhaarStartTime`? Is this always used as a `tzhaarStartTime != null` check?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +638,113 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+		{
+			removeTzhaarTimer();
+			resetTzhaarConfig();
+			resetTzhaarVars();
+		}
+	}
+
+	private boolean checkInFightCaves()
+	{
+		return client.getMapRegions() != null && ArrayUtils.contains(client.getMapRegions(), 9551);
```
Nightfirecat: These region IDs should be defined as constants.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +638,113 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+				tzhaarStartTime = tzhaarStartTime.plus(Duration.between(tzhaarStartTime, now)).minus(Duration.between(tzhaarStartTime, tzhaarLastTime));
+				tzhaarLastTime = null;
+			}
+
+			createTzhaarTimer(tzhaarStartTime, tzhaarLastTime);
+		}
+
+	}
+
+	private void updateInfoBoxState()
```
Nightfirecat: This isn't really what this method is doing. It's possibly removing the timer and resetting config/vars. It should be named appropriately.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +638,113 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+	private void createTzhaarTimer(Instant startTime, Instant lastTime)
+	{
+		if (tzhaarTimer != null)
+		{
+			infoBoxManager.removeInfoBox(tzhaarTimer);
+		}
+
+		if (checkInFightCaves())
+		{
+			tzhaarTimer = new ElapsedTimer(itemManager.getImage(FIRE_CAPE), this, startTime, lastTime);
```
Nightfirecat: Would the Jad and Zuk pets be better images for these timers?

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,85 @@
...
+	{
+		if (startTime == null)
+		{
+			return "";
+		}
+
+		if (lastTime == null)
+		{
+			Duration elapsed = Duration.between(startTime, Instant.now());
+			time = LocalTime.ofSecondOfDay(elapsed.getSeconds());
```
Nightfirecat: This is identical in each branch, move it after this check.

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,85 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+public class ElapsedTimer extends InfoBox
+{
+	private final Instant startTime;
+	private LocalTime time;
+	private Instant lastTime;
```
Nightfirecat: What is this for? Setting a paused duration timer? Some doc to explain that would be nice

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
winterdaze: I don't think so, because slayer task uses the bosses as icons already, so if you were on task you'd have 2x jad or 2x zuk infoboxes

winterdaze:
added a test case, first time working with mockito so hopefully it's okay. not sure if using Thread.sleep is permitted but I needed a way to add time to the timer, alternatives welcome if it's not.

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public void updateTimer()
+	{
+		time = Duration.between(startTime, lastTime == null ? Instant.now() : lastTime);
+	}
+
+	private String formatTime(Duration time)
+	{
+		if (time.toHours() > 0) return DurationFormatUtils.formatDuration(time.toMillis(), "HH:mm", true);
+		return DurationFormatUtils.formatDuration(time.toMillis(), "mm:ss", true);
```
Nightfirecat: Don't use one-line `if`s. Given the only difference is the format string, this would be a simpler way to do this:


```suggestion
		final String formatString = time.toHours() > 0 ? "HH:mm" : "mm:ss";
		return DurationFormatUtils.formatDuration(time.toMillis(), formatString, true);
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +640,112 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+			{
+				int wave = Integer.parseInt(matcher.group(1));
+				if (wave != 1)
+				{
+					return;
+				}
+
+				tzhaarStartTime = now;
+			}
+
```
Nightfirecat:
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +640,112 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+		if (checkInFightCaves())
+		{
+			tzhaarTimer = new ElapsedTimer(itemManager.getImage(FIRE_CAPE), this, startTime, lastTime);
+			infoBoxManager.addInfoBox(tzhaarTimer);
+		}
+		else if (checkInInferno())
+		{
+			tzhaarTimer = new ElapsedTimer(itemManager.getImage(INFERNAL_CAPE), this, startTime, lastTime);
+			infoBoxManager.addInfoBox(tzhaarTimer);
+		}
```
Nightfirecat: These blocks are almost identical and can be replaced with the following:

```java
int imageItem = -1;

if (checkInFightCaves())
{
	imageItem = FIRE_CAPE;
else if (checkInInferno())
{
	imageItem = INFERNAL_CAPE;
}

if (imageItem == -1)
{
	return;
}

tzhaarTimer = new ElapsedTimer(itemManager.getImage(imageItem), this, startTime, lastTime);
infoBoxManager.addInfoBox(tzhaarTimer);
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +640,112 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+				tzhaarStartTime = now;
+			}
+
+			else if (tzhaarLastTime != null)
+			{
+				tzhaarStartTime = tzhaarStartTime.plus(Duration.between(tzhaarStartTime, now)).minus(Duration.between(tzhaarStartTime, tzhaarLastTime));
+				tzhaarLastTime = null;
+			}
+
+			createTzhaarTimer(tzhaarStartTime, tzhaarLastTime);
```
Nightfirecat: Why is this called each time a new wave is started? This should only happen when tzhaarLastTime is different than the value in the active timer right?

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,83 @@
...
+package net.runelite.client.plugins.timers;
+
+import net.runelite.client.ui.overlay.infobox.InfoBox;
+import java.awt.image.BufferedImage;
+import java.awt.Color;
+import java.time.Duration;
+import java.time.Instant;
+import org.apache.commons.lang3.time.DurationFormatUtils;
+
+public class ElapsedTimer extends InfoBox
```
Adam-, line 34:
```suggestion
class ElapsedTimer extends InfoBox
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,83 @@
...
+{
+	private final Instant startTime;
+	private Duration time;
+	private final Instant lastTime;
+
+	// Creates a timer that counts up if lastTime is null, or a paused timer if lastTime is defined
+	public ElapsedTimer(BufferedImage image, TimersPlugin plugin, Instant startTime, Instant lastTime)
```
Adam-, line 41:
```suggestion
	ElapsedTimer(BufferedImage image, TimersPlugin plugin, Instant startTime, Instant lastTime)
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.awt.image.BufferedImage;
+import java.awt.Color;
+import java.time.Duration;
+import java.time.Instant;
+import org.apache.commons.lang3.time.DurationFormatUtils;
+
+public class ElapsedTimer extends InfoBox
+{
+	private final Instant startTime;
+	private Duration time;
```
Adam-, line 37: There's no reason for this to be stored, it should just be computed as necessary

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -618,6 +640,112 @@ else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+
+	private void removeTzhaarTimer()
+	{
+		infoBoxManager.removeInfoBox(tzhaarTimer);
+		tzhaarTimer = null;
+	}
+
+	private void createTzhaarTimer(Instant startTime, Instant lastTime)
+	{
+		if (tzhaarTimer != null)
```
Adam-, line 734: removeTzhaarTimer()

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Adam-: I think this should only be called if one of the above two ifs are true

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -662,12 +789,30 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
+				removeTzhaarTimer();
+				saveTzhaarConfig();
 				removeTbTimers();
 				break;
 			case LOGGED_IN:
 				loggedInRace = true;
+				if (loggingIn)
+				{
+					loggingIn = false;
+					loadTzhaarConfig();
```
Adam-, line 813: What's the purpose of this load-into-memory-then-wipe-from-config pattern you have going here? I think you can just use the values from config and simplify most of this?

winterdaze:
updated with the feedback from last review

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -639,6 +660,118 @@ else if (event.getMessage().contains(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+			if (message.contains(TZHAAR_DEFEATED_MESSAGE) || matcher.matches())
+			{
+				removeTzhaarTimer();
+				resetTzhaarConfig();
+				resetTzhaarVars();
+				return;
+			}
+
+			Instant now = Instant.now();
+			matcher = TZHAAR_PAUSED_MESSAGE.matcher(message);
```
Nightfirecat, line 678: Rather than reassigning `matcher` multiple times, using distinct names for each matcher would be much clearer when reading this branch.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -639,6 +660,118 @@ else if (event.getMessage().contains(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+		tzhaarLastTime = null;
+	}
+
+	private void removeTzhaarTimer()
+	{
+		infoBoxManager.removeInfoBox(tzhaarTimer);
+		tzhaarTimer = null;
+	}
+
+	private void createTzhaarTimer(Instant startTime, Instant lastTime)
```
Nightfirecat, line 750: This is always called with `tzhaarStartTime` and `tzhaarLastTime` as its arguments. Hence, these can be referenced directly instead of being passed as arguments.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -639,6 +660,118 @@ else if (event.getMessage().contains(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+				{
+					return;
+				}
+
+				tzhaarStartTime = now;
+				createTzhaarTimer(tzhaarStartTime, tzhaarLastTime);
+			}
+			else if (tzhaarLastTime != null)
+			{
+				tzhaarStartTime = tzhaarStartTime.plus(Duration.between(tzhaarStartTime, now)).minus(Duration.between(tzhaarStartTime, tzhaarLastTime));
```
Nightfirecat, line 705: Isn't this just `startTime.plus(between(lastTime, now))`?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -639,6 +660,118 @@ else if (event.getMessage().contains(KILLED_TELEBLOCK_OPPONENT_TEXT))
...
+		infoBoxManager.removeInfoBox(tzhaarTimer);
+		tzhaarTimer = null;
+	}
+
+	private void createTzhaarTimer(Instant startTime, Instant lastTime)
+	{
+		if (tzhaarTimer != null)
+		{
+			removeTzhaarTimer();
+		}
```
Nightfirecat, line 755: The timer doesn't need to be nonnull to attempt to remove it from infoboxManager. Hence, this reads more neatly as just `removeTzhaarTimer();` without the null check.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -683,12 +815,28 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
+				loggingIn = true;
+				break;
+			case LOADING:
+				if (!loggingIn)
+				{
+					updateTzhaarTimerOnLoading();
+				}
+				break;
 			case HOPPING:
+				loggingIn = true;
```
Nightfirecat, line 828: Add an explicit comment indicating the fallthrough that happens here
```suggestion
				loggingIn = true;
				// fall through to LOGIN_SCREEN case
```

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -233,4 +242,65 @@ public void testEndurance()
...
+		}
+
+		// test timer pause
+		chatMessage = new ChatMessage(null, ChatMessageType.GAMEMESSAGE, "", TZHAAR_PAUSED_MESSAGE, "", 0);
+		timersPlugin.onChatMessage(chatMessage);
+		verify(infoBoxManager).removeInfoBox(captor.capture());
+		captor = ArgumentCaptor.forClass(InfoBox.class);
+		verify(infoBoxManager, times(2)).addInfoBox(captor.capture());
+		ElapsedTimer timer = (ElapsedTimer) captor.getValue();
+		assertEquals("00:01", timer.getText());
```
Nightfirecat, line 273: The elapsed time testing should be part of a dedicated ElapsedTimerTest suite. I'm not convinced it's a good idea to test this anyway since this explicitly requires 3 seconds of CPU time to wait on sleeping threads. In any event this test case should just be ensuring appropriate responses to expected chat messages.

winterdaze: How does one "ensure appropriate responses" if the test doesn't let the timer run? Like, if you don't create an actual timer how do you test to see if it pauses correctly or not? I made the test per adam's recommendations:
> I would test 1) create on first wave 2) pause it and see if it pauses 3) unpause it and see if it unpauses 4) die and see if it clears at the very least
> eg. send it a start wave message, capture the value passed to addInfoBox() and 1) check it is a elapsed timer 2) check the startTime/lastTime etc of that timer is something you expect

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
winterdaze: I believe the original version of my code used distinct names and Kamiel recommended using just one matcher variable in early review of the plugin when cleaning up the code, so I'd like to confirm this is actually what is wanted before i change it, again.

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
Nightfirecat: If anything, the time comparisons should be in an ElapsedTimerTest suite. The plugin should not test that the timer works--only that its state is as expected at various times. So for the examples given:

1) Create on first wave: verify the infobox was added and that it is an ElapsedTimer
2) Pause it and see if it pauses: verify the added ElapsedTimer has a non-null lastTime
3) Unpause it and see if it unpauses: save the result from the paused timer, verify that the start time is not equal after being unpaused
4) Die and see if it clears: verify that the infobox was removed (and no more were added)

winterdaze:
updated from nightfirecat's review

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
```diff
@@ -0,0 +1,80 @@
...
+import net.runelite.client.ui.overlay.infobox.InfoBox;
+import java.awt.image.BufferedImage;
+import java.awt.Color;
+import java.time.Duration;
+import java.time.Instant;
+import org.apache.commons.lang3.time.DurationFormatUtils;
+
+class ElapsedTimer extends InfoBox
+{
+	@Getter
```
Nightfirecat, line 37: Annotate the class with `@Getter` instead of the fields since they all have generated getters.

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -55,6 +56,7 @@
 import static org.mockito.Mockito.when;
 import org.mockito.junit.MockitoJUnitRunner;
 
+@Slf4j
```
Nightfirecat, line 59: This annotation is unused

runelite-client/src/main/java/net/runelite/client/plugins/timers/ElapsedTimer.java
winterdaze: fixed

Nightfirecat:
I've pushed an update to the added tests which is more in line with what I was imagining. I'll give the rest of the PR another review shortly, but I think this is likely fine now.

PR #8974, opened 2019-06-01
Implement beginner hot-cold clue solutions 

Nightfirecat:
This PR adds a generic hot-cold solver class, as well as abstracts the concepts of temperature and change-of-temperature into interfaces for general use. It uses this class to implement master-level hot-cold clues, and to add an implementation for beginner-level hot-cold clues.

This addresses the hot-cold portion of #8504

PR #9198, opened 2019-06-24
Add support for beginner Hot Cold clues.

emerald000:


Adam-:
According to nightfirecat, this doesn't account for the dig boxes of beginner clues being smaller than masters. (they are 3x3). See also #8974

emerald000:
It does work correctly. The four highlighted tiles mark the corners of the dig zone. Standing one tile outside shows the correct location.

Issue #9097, opened 2019-06-13
Allow arbitrary key remaps


AildirMike:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always
frustrated when [...]
- The key Remapping feature remaps the F keys to 1, 2, 3 and so on.. However on my Azerty Keyboard their default is &, é, ", ', and so on.. Making the feature unusable for me.

**Describe the solution you'd like**
- Is there a possibility for other keyboard to be supported by runelite?

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've
considered.
- X

**Additional context**
- X

Adam-:
fixed in 3125b74f4fb4cbfb05e90d78a61c9ad216b1b08e

PR #9200, opened 2019-06-25
Add custom mapping for F keys, Add toggle to consume Esc presses in chat

birdsoup:
Addresses #9097

* Add custom mapping for Fn keys and the Esc key. 
* Consume Esc presses when the chat is in focus

image

This is mostly the same as #7623, with the following differences:
* Is based on the most recent version of the Key Remapper plugin
* Supports re-binding F11, F12, and Esc
* Consumes Esc presses when chatting (prevents unintentionally switching tabs when canceling the chat).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -92,10 +92,153 @@ public interface KeyRemappingConfig extends Config
 		position = 6,
 		keyName = "fkeyRemap",
 		name = "Remap F Keys",
-		description = "Configures whether F-Keys are Remapped to 1 (F1) through 0 (F10), '-' (F11), and '=' (F12)"
+		description = "Configures whether F-Keys use remapped keys"
 	)
 	default boolean fkeyRemap()
 	{
 		return false;
+	}
+
+	@ConfigItem(
+		position = 7,
+		keyName = "f1",
+		name = "F1",
+		description = "The key which will replace {F1}."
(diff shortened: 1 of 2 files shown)


TzZek:
Yes pls

Adam-:
I don't understand this esc thing. When typing if escaping changes tabs then we should fix just that, not add a config option to fix that.

birdsoup:
@Adam- Fair point. I was originally just trying to not change the existing default functionality, but it does make more sense to leave the configuration part out. I've updated the PR to remove the configuration.

Adam-:
This escape thing looks like a bug to me since it happens on the release regardless of remap camera or fkeys.

birdsoup:
> This escape thing looks like a bug to me since it happens on the release regardless of remap camera or fkeys.

I don't think that the added functionality is a bug. Currently the plugin requires pressing enter to chat (and escape to cancel chatting) when enabled, regardless of whether or not the camera or the f-keys are remapped. This addition only suppresses esc presses when the user is chatting.

Adam-:
Currently on the release if you enable the plugin, change tabs, hit enter, type something and then decide to hit esc, it will both take you out of typing mode and change your tab to your esc keybind, regardless of the wasd or fkey settings. This is a bug.

birdsoup:
Oh, I misread your comment. My bad

Adam-:
I've fixed this bug in 8a4d6172b2e15b78b9321d79bc691577f97fa087

Adam-:
added in 3125b74f4fb4cbfb05e90d78a61c9ad216b1b08e and 39e66af62346949dc77e339ebfb34014b0a3b220

PR #9201, opened 2019-06-25
Wintertodt plugin: Round start notification

kushagharahi:
image
image


runelite-api/src/main/java/net/runelite/api/VarPlayer.java
```diff
@@ -162,7 +162,9 @@
 	MUSIC_TRACKS_UNLOCKED_16(1009),
 	MUSIC_TRACKS_UNLOCKED_17(1338),
 	MUSIC_TRACKS_UNLOCKED_18(1681),
-	MUSIC_TRACKS_UNLOCKED_19(2065);
+	MUSIC_TRACKS_UNLOCKED_19(2065),
+
+	WINTERTODT_TIMER(1142);
```
Adam-: What is this value?

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -182,6 +184,24 @@ public void onGameTick(GameTick gameTick)
...
+		if (config.notifyWhenRoundStarting())
+		{
+			int currentTimerValue = client.getVar(VarPlayer.WINTERTODT_TIMER);
+			if (currentTimerValue > 0)
+			{
+				log.debug("Wintertodt time left until round start " + currentTimerValue);
+
+				if (currentTimerValue < 50000 && currentTimerValue != 0)
+				{
+					notifier.notify("Round is about to start");
```
Adam-: This seems like it would cause notifications to be spammed whenever the time was low

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -182,6 +184,24 @@ public void onGameTick(GameTick gameTick)
...
 
+	@Subscribe
+	public void onVarbitChanged(VarbitChanged varbitChanged)
+	{
+		if (config.notifyWhenRoundStarting())
+		{
+			int currentTimerValue = client.getVar(VarPlayer.WINTERTODT_TIMER);
+			if (currentTimerValue > 0)
+			{
+				log.debug("Wintertodt time left until round start " + currentTimerValue);
```
Adam-:
```suggestion
				log.debug("Wintertodt time left until round start: {}," currentTimerValue);
```

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
kushagharahi: I've used this notification from 70-97 now and it notifies at max two times

Adam-: That may just be your system notifications being smart. You can run the client in debug and see notifier logs.

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
kushagharahi: This is the varplayer value used for the countdown timer to the next Wintertodt match

kushagharahi: Went ahead and added a comment above the enum to better clarify

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
kushagharahi: That makes sense, I will check this tonight and add a debounce if needed

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
Adam-: What I meant was - what is it? Seconds? MS? game ticks? A combination of both?

kushagharahi: It ends up being minutes/seconds on the client interface but I couldn't quite figure out how to translate the value into seconds

I have a sample of what the values look like when sent to the client in the description of the PR

Adam-:
Judging by the scripts (2755, 2757) it looks like `%varbit7980 * 30` is the time in seconds

Joshua-F:
Was told to put this here, varp 1142 is what varbit 7980 uses which is why the script registers a hook for `var1142`.

```ini
[7980]
basevar = 1142
startbit = 14
endbit = 30
```

kushagharahi:
@Adam- @Joshua-F Awesome I'll make those changes tonight! Out of curiosity how did y'all figure that out?

Joshua-F:
You can find the usages of that varbit in the scripts Adam said.

https://github.com/RuneStar/cs2-scripts/blob/a9b97e1d8b61f66fe892cd84b2dd869142d30c00/scripts/%5Bproc%2Cwint_timer_resynch%5D.cs2
https://github.com/RuneStar/cs2-scripts/blob/a9b97e1d8b61f66fe892cd84b2dd869142d30c00/scripts/%5Bproc%2Cwint_timer_countdown%5D.cs2

Adam-:
Since the timer only appears to have 30s resolution we will have to wait for it to hit ~30s - then start a countdown maybe ourself.

kushagharahi:
@Adam- @Joshua-F 
Based on your findings I was able to make it configurable by time, so you can now select 15, 10, 5 seconds remaining or none. Check it out and let me know what you think

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -182,6 +186,29 @@ public void onGameTick(GameTick gameTick)
 		checkActionTimeout();
 	}
 
+	@Subscribe
+	public void onVarbitChanged(VarbitChanged varbitChanged)
+	{
+		if (config.notifyWhenRoundStarting() != TimerAlertOptions.NONE)
```
Adam-: Generally the way we do varbit changes is to store the old value and then detect a change, and then check things like config etc, since it is faster. You can also use it to detect the duplicate alerts.

Adam-:
I don't see anything wrong with just a standard int input for the config. Set it to like 5 as default.

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
kushagharahi: Done

kushagharahi:
@Adam- There are no validations for an int input config.
Also `varbit 7980` doesn't always start at precisely 80s, so if a user sets the alert from 80s to ~75s it might not alert due to this line `if (previousSecondsToNextRound > timeToNotify && secondsToNextRound <= timeToNotify)` since the previousSecondsToNextRound have been set to a previous varbit value.

Hope that makes sense, I know it was a bit confusing. If we are ok with those tradeoffs I will happily make that config change.

Adam-:
1) use `@Range`
2) Using 0 to disable things is a fairly common thing we do

Adam-:
@kushagharahi can you try this?

kushagharahi:
@Adam- Yep works great

runelite-client/src/test/java/net/runelite/client/plugins/wintertodt/WintertodtPluginTest.java
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright (c) 2019, Adam <Adam@sigterm.info>
```
kushagharahi:
```suggestion
 * Copyright (c) 2019, Kusha Gharahi<kusha.me>
```

@Adam- I would like to claim the copyright here, if possible please.

Adam-: So add yourself to it.

PR #9086, opened 2019-06-11
WorldMapPlugin: Show quest state

connorjclark:
Found it simpler to restart than rebase #8199.

I'd like to follow up this PR with adding the locations to the Quest enum, and removing the QuestLocation enum.

I think the icon for "In Progress" quests should be yellow, to match the color in the quest log. I just copied the images from the original PR, but I could remake the icon as yellow tomorrow.

There are some quests that are started in the same location. The logic will only show the latest non-completed quest, else it will show the last quest.

image

image

image

Also, I figured rolling this into the existing quest start config option made sense, instead of making a new config option.

PR #8442, opened 2019-04-08
World Map: Identify Both Shield of Arrav Quest Start Points

xDemoN:
Closes #8437 

Let me know if you'd like the syntax to be different.

image

image

PR #9233, opened 2019-06-28
Shield of Arrav Quest Icon Fix - #9215

### LINKED ISSUE #9215: Quest icon not ticked when quest completed
Screenshot 2019-06-26 at 20 03 28

Yea I have a quest cape and for some reason there isn't a tick next to this quest icon when all the other ones have a tick next to it?

basicDamen:
adds second quest starting point for shield of arrav (using Charlie the Tramp to start the quest instead of Reldo)

[ closes #9215 ]


xDemoN:
This is a regression because I fixed it in #8442 and for some reason it was removed in #9086 

You should update the ENUM to use the same naming convention as #8442

Adam-:
Additionally the coordinate from #8442 is slightly different so I would just use that too.

deathbeam:
Updated

PR #9236, opened 2019-06-28
npc indicators: Fix long respawn timer bug

Nightfirecat:
By killing a monster which has a known respawn timer and leaving its
respawn area (by walking away or teleporting) and returning to it some
time after it has respawned, the plugin can erroneously overwrite its
respawn time with a much longer one. This commit fixes that issue by
assuming that the lower of the observed respawn time and the previously
recorded respawn time is the correct one.


Adam-:
This is just:
```
int respawnTime = client.getTickCount() + 1 - mn.getDiedOnTick();
if (mn.getRespawnTime() == -1 || respawnTime < mn.getRespawnTime()) {
  mn.setRespawnTime(respawnTime);
}

PR #9255, opened 2019-07-01
fix duplicate tob and cox loot tracker

### LINKED ISSUE #4754: Loot tracker adds multiple entries if you open ToB reward chest more than once
Steps to reproduce: finish a raid and open the chest, close the interface and open again. Will result in multiple loot entries.

winterdaze-rs:
closes #4754 for both ToB and CoX

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -62,6 +62,7 @@ import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.WidgetLoaded;
+import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.widgets.WidgetID;
 import net.runelite.client.account.AccountSession;
 import net.runelite.client.account.SessionManager;
@@ -113,11 +114,17 @@ public class LootTrackerPlugin extends Plugin
 		13113, "Larran's small chest"
 	);
 
+	private static final String CONFIG_LOOTED = "looted";
+	private static final String CONFIG_GROUP = "loottracker";
+
 	@Inject
 	private ClientToolbar clientToolbar;
 
 	@Inject
 	private ItemManager itemManager;
+
+	@Inject
+	private ConfigManager configManager;
 
 	@Inject
 	private SpriteManager spriteManager;
@@ -147,6 +154,9 @@ public class LootTrackerPlugin extends Plugin
 
 	@Getter(AccessLevel.PACKAGE)
 	private LootTrackerClient lootTrackerClient;
+
+	private Boolean looted = false;
+	private boolean loggingIn;
 
 	private static Collection<ItemStack> stack(Collection<ItemStack> items)
 	{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -318,6 +328,42 @@ public void onPlayerLootReceived(final PlayerLootReceived playerLootReceived)
...
+					resetConfig();
+				}
+				break;
+			case LOGGING_IN:
+				loggingIn = true;
+				break;
+			case LOADING:
+				if (!loggingIn)
+				{
+					if (!client.isInInstancedRegion())
```
Nightfirecat: Collapse this into one `if` statement.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -318,6 +328,42 @@ public void onPlayerLootReceived(final PlayerLootReceived playerLootReceived)
...
+						looted = false;
+					}
+				}
+				break;
+			case HOPPING:
+				loggingIn = true;
+			case LOGIN_SCREEN:
+				saveConfig();
+				break;
+			default:
```
Nightfirecat: If a default case is empty, it can be removed without consequence.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -330,7 +376,15 @@ public void onWidgetLoaded(WidgetLoaded event)
 				break;
 			case (WidgetID.CHAMBERS_OF_XERIC_REWARD_GROUP_ID):
 				eventType = "Chambers of Xeric";
-				container = client.getItemContainer(InventoryID.CHAMBERS_OF_XERIC_CHEST);
+				if (!looted)
```
Nightfirecat: Make this a guard clause instead.

```java
if (looted)
{
	return;
}
looted = true;
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -339,7 +393,15 @@ public void onWidgetLoaded(WidgetLoaded event)
 					return;
 				}
 				eventType = "Theatre of Blood";
-				container = client.getItemContainer(InventoryID.THEATRE_OF_BLOOD_CHEST);
+				if (!looted)
```
Nightfirecat: Guard clause, like above.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -148,6 +155,9 @@
 	@Getter(AccessLevel.PACKAGE)
 	private LootTrackerClient lootTrackerClient;
 
+	private Boolean looted = false;
```
Nightfirecat: Why is this a boxed type? Also, `boolean` fields initialize to `false`, so it does not need to be declared.

winterdaze-rs: it is Boolean so it can be saved/loaded from the config

Nightfirecat: Mentioned in Discord, but this becomes boxed when being coerced to `Object` by `setConfiguration()`. You can change this to `boolean` and test locally to see for yourself that this works fine that way.

Adam-:
1) why is this storing looted status in the *config*?
2) why are you commiting client jars to the repo?

PR #9256, opened 2019-07-01
Hot cold fixes

Nightfirecat:
Bugfixes/improvements for the hot-cold port.

1. Hot-cold temperature gets the wrong temperature of the string "The device is warm, but colder than last time."
2. When a locator orb is activated at a location where it is visibly shaking, it resets the possible locations list instead of displaying the location.
3. When reading a new hot-cold clue, if a location had been marked from a prior hot-cold clue in the session, the final location is not cleared.

cc: @Hydrox6, as he helped point out some of these bugs. This branch should address the issues you found

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/hotcold/HotColdTemperature.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/hotcold/HotColdTemperature.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import javax.annotation.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import net.runelite.client.util.Text;
 
 @AllArgsConstructor
 @Getter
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/hotcold/HotColdTemperature.java
```diff
@@ -74,6 +75,7 @@
 	private final int maxDistance;
 
 	private static final String DEVICE_USED_START_TEXT = "The device is ";
+	private static final Pattern COMMA_PATTERN = Pattern.compile(",");
```
deathbeam: We have Text.fromCSV

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/hotcold/HotColdTemperature.java
```diff
@@ -93,11 +94,12 @@ public static HotColdTemperature getFromTemperatureSet(final Set<HotColdTemperat
 			return null;
 		}
 
+		final String messageStart = Text.fromCSV(message).get(0);
```
deathbeam: Shouldnt this also check if the split is not empty? just in case

Nightfirecat: This method already does an early return if the message doesn't start with `DEVICE_USED_START_TEXT` ("The device is "), so there will always be something in the first split.

deathbeam: Alright

PR #9258, opened 2019-07-01
Teleblock timer not showing fix

### LINKED ISSUE #9257: Teleblock message now names the player who cast the spell, does not trigger timer
The new message is of the format: "A Tele Block spell has been cast on you by **Zezima**. It will expire in 5 minutes, 0 seconds."

An example: https://clips.twitch.tv/ShortAmericanLlamaHeyGuys

jordibenck:
Teleblock timer is not showing after message because of the new message format.
Closes #9257


Adam-:
It is probably worth at least precompiling these patterns. At this point maybe we should just make one pattern which parses out the time and uses it to create the timer? + Add some tests.

Adam-:
I didn't realize that there wasn't a way to make dynamic len timers. Go back to what you had before but with the test and with pattern precompiling, and that should be good.

Issue #2561, opened 2018-05-10
Client moves down when opening and closing the side bar


imapersonnow:
Since 10th May update when I open and close the side bar, the client moves down a few pixels each time. See the video below.

[https://streamable.com/vvysm](url)

I'm running on Ubuntu 16.04 with disabled hardware acceleration.

RomeuG:
Same. I am on ArchLinux btw :^)

imapersonnow:
I should add that I upgraded from Ubuntu 16.04 running Unity to Ubuntu 18.04 running Gnome and this issue is gone.

abextm:
This seems to stem from AWT adjusting its position depending on what it thinks the size of the window's decorations are (the window's Insets). Under X11 it does this by emitting a `_NET_REQUEST_FRAME_EXTENTS` and waiting for `_NET_FRAME_EXTENTS` back. If your wm does not report this correctly, it will make the window move every time we call `Frame::setBounds`. In the case of `i3` and `i3-gaps` it [will always report the font height + 4](https://github.com/i3/i3/blob/master/src/render.c#L25), regardless of if you have decorations enabled or not.

So there are a number of places this behavior can be avoided:
1) Your WM can provide accurate info, knowing that the JRE needs it to operate correctly, despite X11's spec
2) The JRE can try to handle this case better, though I don't know if this is possible
3) We can be careful to not emit needless `setBounds` calls, but that requires a rewrite of `ContainableFrame` because the way it is now is unsafe to use calls that are not `setBounds` due to the private field `Component::boundsOp` becoming out of sync with the data passed to `setBounds`

Septem151:
> This seems to stem from AWT adjusting its position depending on what it thinks the size of the window's decorations are (the window's Insets). Under X11 it does this by emitting a `_NET_REQUEST_FRAME_EXTENTS` and waiting for `_NET_FRAME_EXTENTS` back. If your wm does not report this correctly, it will make the window move every time we call `Frame::setBounds`.

I'm trying to find where this call is made in awesome-wm, however I'm [not able to find it](https://github.com/awesomeWM/awesome/blob/v4.3/objects/client.c) since I'm not all too familiar with C. However, I did find multiple mentions of the `client:geometry()` call [excluding border from X and Y](https://github.com/awesomeWM/awesome/blob/v4.3/lib/awful/client.lua#L842) which could be why the client window shifts.

This does seem to be the case, as when border width is set to 0, the RuneLite window remains in the same position.

[...]

Issue #2521, opened 2018-05-09
Looking up username with only numbers


Timrwwatson:
I tried to lookup the username '6500' today at zeah rc and this didnt return anything in the highscores tab.

On the osrs HiScores webpage this player has 2000+ total level, I think the lookup doesn't handle an all integer name correctly.

I tried typing the name in as well as right clicking on the player and selecting the op Lookup.

Timrwwatson:
This was caused by earlier in the play session i clicked the UIM button, and I didn't realise this stayed selected, if this was more clear or became de-selected by default etc

PR #9270, opened 2019-07-03
runelite-client: Don't try to contain to screen when contain is off

abextm:
Java's GraphicsConfiguration::getBounds() is completely nonfunctional for me, returning nearly random values, which causes the our window to move itself to random places, often outside of my displays' bounds

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ContainableFrame.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ContainableFrame.java
@@ -113,23 +113,27 @@ public class ContainableFrame extends JFrame
 		if (forcedWidthIncrease || expandResizeType == ExpandResizeType.KEEP_GAME_SIZE)
 		{
 			final int newWindowWidth = getWidth() + increment;
-			final Rectangle screenBounds = getGraphicsConfiguration().getBounds();
-			final boolean wouldExpandThroughEdge = getX() + newWindowWidth > screenBounds.getX() + screenBounds.getWidth();
 			int newWindowX = getX();
 
-			if (wouldExpandThroughEdge)
+			if (containedInScreen)
 			{
-				if (!isFrameCloseToRightEdge() || isFrameCloseToLeftEdge())
+				final Rectangle screenBounds = getGraphicsConfiguration().getBounds();
+				final boolean wouldExpandThroughEdge = getX() + newWindowWidth > screenBounds.getX() + screenBounds.getWidth();
+
+				if (wouldExpandThroughEdge)
 				{
-					// Move the window to the edge
-					newWindowX = (int)(screenBounds.getX() + screenBounds.getWidth()) - getWidth();
+					if (!isFrameCloseToRightEdge() || isFrameCloseToLeftEdge())
+					{
+						// Move the window to the edge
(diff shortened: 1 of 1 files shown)


deathbeam:
Im pretty sure this issue is global to most unix window managers. I think this issue is related: https://github.com/runelite/runelite/issues/2561

abextm:
#2521 is because it swing thinks the window has a ~20px titlebar always on x11, and tries to move the content pane to be that far down from the window's bounds. This is some other (probably unrelated) x11 java bug.

themrrobert:
I do not like the changes to ContainableFrame.java in d72a1b9456ac2757c7012a6f2b35c679e3f490bb 

When my window is on the right of the screen, which it always is, the frame now pops up off screen to the right. The old behavior is restored when enabling Contain to screen, but this is undesirable as I frequently move the client off-screen in order to work. 

If this is a workaround to an x11 bug, then an additional option should be provided for that, instead of breaking existing functionality to fix a bug.

Thanks :)

Adam-:
For me, this causes the client to expand out of bounds. How is this helping you if it is still using `getGraphicsConfiguration().getBounds()`?

abextm:
`getGraphicsConfiguration().getBounds()` is locked behind a config, which I have disabled, so it's values don't get used.

Adam-:
The config option is meant to prevent dragging the client off screen. Using it to determine if the client expanding should be allowed to expand off screen seems unrelated. Generally I think when someone expands the client without the client locked to the screen the expected behavior would not be the client expanding off screen.

To me this seems like a major UX regression. If your problem is truly specific to *nix WMs I would rather see a different approach taken since so few users use nix.

Have you tried either trying to sanity check the values to determine if they are bogus, or at the very least just disable the window expand logic on non-Windows/osx?

abextm:
The config option is described as `Makes the client stay contained in the screen when attempted to move out of it.`, which applies to the client resizing or being moved by the user. If you want we can reword the config option to what you think it should be, and replace this check with an OS == Linux check instead, because as far as I can tell most of openjdk's X11 code is mostly unmaintained and not supported.
Additionally the config description contains `Note: Only works if custom chrome is enabled.` which seems to be wrong because this does atleast attempt to work without custom chrome on.

Adam-:
It looks like `setContainedInScreen` checks the titlebar, so it doesn't run. Not sure why.

The behavior of the release is that it will expand and keep itself in the bounds of the current screen, even when you have multiple screens, with the config option off (which is default)

deathbeam:
`Generally I think when someone expands the client without the client locked to the screen the expected behavior would not be the client expanding off screen.`

I think its exactly opposite. This behaviour should be at least configurable, and now it is. If people want to split it, they should open new issue for that. NFC why is this being discussed on PR. 

`
It looks like the call to setContainedInScreen checks the titlebar, so it doesn't run. Not sure why.
`

The setContainedinScreen checks the titlebar, because without custom window chrome (e.g no "title bar"), OS should handle *everything* window related, and not this dumb stuff we are doing with Swing.

Adam-:
The previous behavior of:

1) not locking moving to any one screen
2) moving the client to not expand off of the current screen

is no longer possible now. I think this is a good default behavior of the client and should be kept as the default behavior.

PR #9282, opened 2019-07-04
Added untradeable bond item price

RRickkert:
This pull request is to add the item price to untradeable bonds, allowing the bank value to also count bonds in it's total bank value. They're worth exactly 0.9x the value of a tradeable bond.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
@@ -282,6 +282,11 @@ public class ItemManager
 		{
 			return 1000;
 		}
+		if (itemID == OLD_SCHOOL_BOND_UNTRADEABLE)
+		{
+			// 1 untradeable bond = 0.9 * tradable bond = tradable bond - 1/10th of a tradable bond
+			return Math.round(getItemPrice(ItemID.OLD_SCHOOL_BOND) * 0.9f);
+		}
 
 		UntradeableItemMapping p = UntradeableItemMapping.map(ItemVariationMapping.map(itemID));
 		if (p != null)


RRickkert:
A cleaner way would've been to modify the `UntradeableItemMapping` to support float quantities, adding the untradeable bond as an entry.
The issue with that implementation is the usage of the `UntradeableItemMapping` in the `getItemPrice` method:
```Java
UntradeableItemMapping p = UntradeableItemMapping.map(ItemVariationMapping.map(itemID));
if (p != null)
{
	return getItemPrice(p.getPriceID()) * p.getQuantity();
}
```
Where `ItemVariationMapping.map(itemID)` would "map the untradeable bond to a tradeable bond" resulting in the `UntradeableItemMapping` looking for a "tradeable bond key" (which doesn't exist, since only an untradeable bond key exists). Creating a "tradeable bond entry" would result in an infinite loop:

`ItemVariationMapping` returns tradeable bond id, `UntradeableItemMapping` returns mapping to tradeable bond id, `getItemPrice on tradeable bond id` will call the same line of code again. `ItemVariationMapping` returns tradeable bond id again, `UntradeableItemMapping` ... etc.

RRickkert:
Modified implementation based on feedback from @TheStonedTurtle and @Hydrox6 , using the method described in the comment, but with a clean workaround for the infinite loop issue.

> I'd just send the actual item ID and the variation mapped ID to the UntadeableItemMapping.map and have that check both
> check UntradeableItemMapping with the unmapped itemID
as well as the mapped itemID

RRickkert:
Moved the code to support `UntradeableItemMapping`'s optional itemId in the `map` method to a new method in the `ItemManager` class instead, since it didn't belong there.

deathbeam:
Needs rebase

PR #9283, opened 2019-07-04
Show Remaining Levels and XP in skill calculator

### LINKED ISSUE #9260: Show Remaining XP in skill calculator
No

On the xp calculator, when u put two level in to see the xp it will give you both xp that each level has but it would subtract the xp to tell you how much xp you need till u would get to the level.
ex:
Mining -
Level: 2 (current level) = 83
Level: 5 (desired level) = 388
Remaining XP = 305

I usually have to open the calculator and type in the digits to subtract so i can find something like and xp per hour and calculator how long it would take to do something but having to type all the digits in, especially when it gets to the millions, is tedious and i feel like it can be easily resolved with a "remaining xp" variable somewhere. 

**Additional context**
I think this would be a small very useful function.

DSerp:
Closes #9260 

image

I am open to other implementations

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -339,6 +339,8 @@ class SkillCalculator extends JPanel
 		uiInput.setCurrentXPInput(currentXP);
 		uiInput.setTargetLevelInput(targetLevel);
 		uiInput.setTargetXPInput(targetXP);
+		uiInput.setRemainingLevelsInput(targetLevel - currentLevel);
+		uiInput.setRemainingXPInput(targetXP - currentXP);
 		calculate();
 	}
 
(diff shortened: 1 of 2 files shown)


Kagiri12:
@Serpa88 What about adding a button that sorts the list for least actions or most xp per item made or done?

xDemoN:
This also closes #4816 and supersedes #5789

psikoi:
Are those editable? If they're not, they're styled as if they are, that might be confusing.

Serpa88:
They are not editable. I thought the same but I also wanted them to look similar as well, so I am not sure about it.

psikoi:
Make them regular labels I guess

psikoi:
Also, you're not just going to leave "Remaining Experie..." right?

Serpa88:
Well I don't think regular labels is right either though

psikoi:
You could try adding a rectangular border to the label, this way, it would have the same shape as the text fields above, but not the same border or background, try it out

Serpa88:
> @Serpa88 What about adding a button that sorts the list for least actions or most xp from 1 item made or done?

That would have to be in another pull request

deathbeam:
Just set the fields to disabled and make them grey.

I also don't think this should be implemented, but its spamming Discord.

Issue #9301, opened 2019-07-07
Motherlode rock obstacles are not being marked after logging in or entering the area


Quasindro:
**Describe the bug**
The rock obstacles in MLM do not get marked when a player logs in or has just entered the area.

**To Reproduce**
1. Be in MLM
2. Stand near any rock obstacle
3. Log out
4. Log back in
5. The obstacle tiles are not marked in red, even though they should be

**Expected behavior**
The rock obstacles' tiles should be marked reliably when the config setting is on

**Screenshots**
After logging in:
![image](https://user-images.githubusercontent.com/10108473/60766475-82bfc080-a0aa-11e9-8a5a-b40c56060f70.png)

When the rocks respawn after being broken
![image](https://user-images.githubusercontent.com/10108473/60766481-b569b900-a0aa-11e9-80e1-3ddd68637a6e.png)

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.5.29-SNAPSHOT
 - Launcher version: Unknown

**Additional information**
I already have an idea as to what may be wrong in the plugin, so I will submit a small PR shortly.

Adam-:
fixed in #9302

PR #9302, opened 2019-07-07
motherlode: Perform MLM location check when the player is logging in or entering an area

Quasindro:


Quasindro:
fixes #9301

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -376,6 +376,8 @@ public void onGameStateChanged(GameStateChanged event)
 			// on region changes the tiles get set to null
 			veins.clear();
 			rocks.clear();
+
+			inMlm = checkInMlm();
```
Adam-: If this is being moved to here, you don't need it below in the `LOGGED_IN` case

Adam-:
Do you know why this fixes the issue? Is it that the spawn events happen during loading? I wonder how this ever worked..

Quasindro:
Yeah, the obstacles seem to have been an afterthought. When I checked all the *spawned events and compared them with the gamestates, it seems that the wallobjects spawn correctly after the GameState is set to LOGGED_IN and the checkInMlm() method has already ran. The gameobjects, however, are being loaded when the client is still loading.

Committed the change suggested in a review. You were right, removing that check did not change anything.

PR #9323, opened 2019-07-09
Add Hespori to loot tracker

### LINKED ISSUE #9249: Track Hespori loot in Loot Tracker
I know the Hespori is a farming patch and not technically and "actor" dropping items on the ground, but it would make sense for its drops to be tracked as loot.  It is a combat boss, after all.

Decorth:
The looted message "You have successfully cleared this patch for new crops." seems to be unique to the Hespori patch.

Closes #9249

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -105,6 +105,11 @@ public class LootTrackerPlugin extends Plugin
 	private static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
 	private static final String HERBIBOR_EVENT = "Herbiboar";
 
+	// Hespori loot handling
+	private static final String HESPORI_LOOTED_MESSAGE = "You have successfully cleared this patch for new crops.";
+	private static final String HESPORI_EVENT = "Hespori";
+	private static final int HESPORI_REGION = 5021;
+
 	// Chest loot handling
 	private static final String CHEST_LOOTED_MESSAGE = "You find some treasure in the chest!";
 	private static final Pattern LARRAN_LOOTED_PATTERN = Pattern.compile("You have opened Larran's (big|small) chest .*");
@@ -432,6 +437,12 @@ public class LootTrackerPlugin extends Plugin
 			return;
 		}
 
+		if (HESPORI_REGION == client.getLocalPlayer().getWorldLocation().getRegionID() && message.equals(HESPORI_LOOTED_MESSAGE))
+		{
+			eventType = HESPORI_EVENT;
+			takeInventorySnapshot();
+		}
+
 		// Check if message is for a clue scroll reward
 		final Matcher m = CLUE_SCROLL_PATTERN.matcher(Text.removeTags(message));
 		if (m.find())
@@ -464,7 +475,7 @@ public class LootTrackerPlugin extends Plugin
 	@Subscribe
 	public void onItemContainerChanged(ItemContainerChanged event)
 	{
-		if (eventType != null && (CHEST_EVENT_TYPES.containsValue(eventType) || HERBIBOR_EVENT.equals(eventType)))
+		if (eventType != null && (CHEST_EVENT_TYPES.containsValue(eventType) || HERBIBOR_EVENT.equals(eventType) || HESPORI_EVENT.equals(eventType)))
 		{
 			if (event.getItemContainer() != client.getItemContainer(InventoryID.INVENTORY))
 			{


Decorth:
Woops missed off a bracket

Decorth:
Closes #9249

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -408,6 +412,12 @@ public void onChatMessage(ChatMessage event)
 
 			return;
 		}
+		
+		if (message.equals(HESPORI_LOOTED_MESSAGE))
```
Nightfirecat: This message is not unique (it is used for trees, bushes, etc.) and it is possible to get loot from clearing tree patches in the form of tree roots. Before checking for message equality, check that the client's region is equal to the hespori region.

Decorth: Okay I was pretty confused because I tested the trees and bushes etc, and the message did not come up. However, you are right, it comes up when you clear a dead tree (can you get loot from dead stuff,? Just out of interest). I'll make the necessary changes and check the client region.

Nightfirecat: afaik only healthy normal trees (that is, not fruit or hardwood trees) give any loot when dug up.

Decorth:
The loot tracker will now check if the players current map region is the same as the Hespori map region and then will check if the string  "You have successfully cleared this patch for new crops." occurs in the chat.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -103,6 +103,11 @@
 	private static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
 	private static final String HERBIBOR_EVENT = "Herbiboar";
 
+	// Hespori loot handling
+	private static final String HESPORI_LOOTED_MESSAGE = "You have successfully cleared this patch for new crops.";
+	private static final String HESPORI_EVENT = "Hespori";
+	private static final int [] HESPORI_REGION = {4764, 4765, 4766, 5020, 5021, 5022, 5276, 5277, 5278};
```
raiyni: Hespori is 1 region

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -409,6 +415,12 @@ public void onChatMessage(ChatMessage event)
 			return;
 		}
 
+		if (Arrays.equals(HESPORI_REGION, client.getMapRegions()) && message.equals(HESPORI_LOOTED_MESSAGE))
```
Nightfirecat: Use `client.getLocalPlayer().getWorldLocation().getRegionID()` and compare against a single region instead of comparing against the full region array.

Decorth: Okay sure thanks

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -103,6 +103,11 @@
 	private static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
 	private static final String HERBIBOR_EVENT = "Herbiboar";
 
+	// Hespori loot handling
+	private static final String HESPORI_LOOTED_MESSAGE = "You have successfully cleared this patch for new crops.";
+	private static final String HESPORI_EVENT = "Hespori";
+	private static final int HESPORI_REGION = 4765;
```
Nightfirecat: That's not the region. Look at my screenshot again.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -304,6 +309,7 @@ public void onNpcLootReceived(final NpcLootReceived npcLootReceived)
 	@Subscribe
 	public void onPlayerLootReceived(final PlayerLootReceived playerLootReceived)
 	{
+
```
Nightfirecat: Remove this empty line
```suggestion
```

deathbeam:
Any update on this?

Nightfirecat:
I've squashed and fixed the last couple things I noticed. Tested ingame and found that it works fine--across a full farm run, only hespori was tracked as loot, which is corect.

PR #9328, opened 2019-07-10
mta: fix Telekinetic solver not moving

### LINKED ISSUE #3946: MTA plugin: Telekinetic solver fails to update

raiyni:
The bug happened when the Telekinetic projectile existed on the same game tick as the guardian moving.

Closes #3946


Adam-:
AFAICT this is due to a race with the projectiles being culled at the time the npc update moves the guardian. I can find no reason why this logic existed in the first place.

PR #9345, opened 2019-07-12
runelite-client: Allow partial screen containment again

abextm:
See #9270


runelite-client/src/main/java/net/runelite/client/ui/ContainableFrame.java
```diff
@@ -27,23 +27,40 @@
...
 public class ContainableFrame extends JFrame
 {
+	@RequiredArgsConstructor
+	public enum Mode
+	{
+		ALWAYS("Always"),
+		RESIZING("Resizing"),
+		NEVER("Never");
+
+		public final String name;
```
deathbeam: oh, actually. you do not need to do this anymore, config manager is handling simple naming like this now, so remove both name and toString

PR #9354, opened 2019-07-12
Fix MSAA white pixels

Toocanzs:
Fixes the white pixels at the edges of objects when using MSAA.

Before:
Before
After:
After


Adam-:
With smoothbanding on, what happens?

Toocanzs:
It looks normal. For some reason smooth banding doesn't need centroid

PR #9358, opened 2019-07-13
Add sprite override for button to switch between friends/ignore list in 2005 interface style

sadegr:
I could have sworn we used to already have this at one point. Maybe we didn't and I'm just losing my mind.

Either way, the sprite IDs for the buttons at the top of the friends/ignore list differ from the sprite IDs of the icons on the interface, so we need separate sprite overrides for the buttons. I took the sprites of the 2005 icons and edited them to make them look nice scaled down - if anyone wants to take a shot at making them look better, I'm all for it.

The 2010 friends/ignore list icons are a lot more detailed than the ones in the 2005 interface and I don't think they'd look good scaled down, so I haven't created sprite overrides for the buttons in the 2010 interface (for now).


Adam-:
Whats up with the file sizes on these images?

sadegr:
Oops, used the wrong compression settings on those PNGs. I've losslessly compressed them to the max now and removed any EXIF data.

I force pushed to my branch because one of my commits contained an email address I want to keep private; I rebased the branch to use my personal email in that commit instead.

Issue #1865, opened 2018-04-24
TzTok-Jad task does not register properly


rherms:
When you have a Jad task, the slayer counter goes to 0 once you kill any fight caves creature instead of after you fight Jad.

rherms:
🎉 nice!

PR #9360, opened 2019-07-13
slayerplugin: Correct Jad and Zuk task tracking

Nightfirecat:
This adds a field to Task which dictates the expected exp gain of the
task. When set to a non-zero value, it will prevent slayer exp gains
from registering as a kill unless they are of exactly that amount.

Closes runelite/runelite#1865

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -535,7 +535,15 @@ public class SlayerPlugin extends Plugin
 			return;
 		}
 
-		killedOne();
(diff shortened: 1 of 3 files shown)


Adam-:
Does the task need to be cleared when you die or leave?

Nightfirecat:
It should already be cleared for fight caves [via L481-485 of the plugin currently](https://github.com/runelite/runelite/blob/570de63d3d05b360955fcce33e2ddc2d69543147/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java#L481-L485), and I believe that message is the same for dying or voluntarily leaving the fight caves. As for the inferno, I'm not sure what the chat message for that event looks like, and can't find out without getting a zuk task. I'd be happy to add that in if someone knows what that looks like, but I don't have that myself.

PR #9361, opened 2019-07-13
slayerplugin: Fix first slayer kill not being registered

Nightfirecat:
Because a user can have 0 slayer experience on a new account, we cannot
rely on 0 as the pre-login value. Prior to this change, a player getting
their first slayer kill in the described scenario would cause it not to
be counted toward the tracked task.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -204,7 +204,7 @@ public class SlayerPlugin extends Plugin
 	private int points;
 
 	private TaskCounter counter;
-	private int cachedXp;
+	private int cachedXp = -1;
 	private Instant infoTimer;
 	private boolean loginFlag;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -217,15 +217,19 @@ protected void startUp() throws Exception
...
-			&& config.amount() != -1
-			&& !config.taskName().isEmpty())
+		if (client.getGameState() == GameState.LOGGED_IN)
 		{
-			points = config.points();
-			streak = config.streak();
-			setExpeditiousChargeCount(config.expeditious());
-			setSlaughterChargeCount(config.slaughter());
-			clientThread.invoke(() -> setTask(config.taskName(), config.amount(), config.initialAmount(), config.taskLocation()));
+			clientThread.invoke(() -> cachedXp = client.getSkillExperience(SLAYER));
```
Adam-: This is actually safe off of the client thread

PR #9370, opened 2019-07-14
Added Ring of Forging to Item Charges plugin.

### LINKED ISSUE #8865: Add "Ring of Forging" to the "Item Charges" plugin
Currently, Runelite's Item Charges plugin does not have any tracking features for the Ring of Forging. This item currently sees more use recently due to being introduced to Free-to-Play players.

The Item Charges plugin can be updated to add useful "charge count" features similar to that of the "Dodgy Necklace" tracked features.

A notification (and/or audible warning) could be added for when Runelite sees the message "Your Ring of Forging has melted." in the chatbox.

zachtyll:
Fixes #8865 Add "Ring of Forging" to the "Item Charges" plugin.


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -395,6 +453,10 @@ private void updateJewelleryInfobox(ItemWithSlot type, Item[] items, EquipmentIn
 
 		if (itemWithCharge == null)
 		{
+			if (id == ItemID.RING_OF_FORGING && type == ItemWithSlot.RING_OF_FORGING)
```
Adam-: Add this to the bottom with `else if`

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -182,6 +197,24 @@ public void onChatMessage(ChatMessage event)
 			{
 				notifier.notify("Your Ring of Recoil has shattered");
 			}
+			else if (ringOfForgingCheckMatcher.find())
+			{
+				updateRingOfForgingCharges(Integer.parseInt(ringOfForgingCheckMatcher.group(1)));
```
Adam-: this will error when the captured group is `one`. there should be a test for this too.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeOverlay.java
```diff
@@ -69,6 +69,15 @@ public void renderItemOverlay(Graphics2D graphics, int itemId, WidgetItem itemWi
...
+		if (itemId == ItemID.RING_OF_FORGING)
+		{
+			if (!config.showRingOfForgingCount())
+			{
+				return;
+			}
+
+			charges = config.ringOfForging();
+		}
 		if (itemId == ItemID.DODGY_NECKLACE)
```
Adam-: `else if`

Adam-:
Whats up with all of the basket/bellows changes to this?

zachtyll:
> Whats up with all of the basket/bellows changes to this?

When I renumbered the configs the basket and the bellows appeared right in the middle of them. So rather than go back and change all the changes that had just been done they were put at the end positions.

Adam-:
You should really not be resorting any of the config like this in this PR. What happened to that test you were writing? Ideally this should have tests for each matcher.

zachtyll:
The new configurations could be appended to the end of the configurations file and the old ones put back into the order they are in the master repo. The tests are not located within the Itemcharges plugin folder where the git instance was set up so they did not get added to the commit. This can be changed. Would it be a good idea to include tests for the binding necklace at the same time, as the syntax for the tests would be almost identical?

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -34,273 +34,313 @@
 public interface ItemChargeConfig extends Config
 {
 	@ConfigItem(
-		keyName = "veryLowWarningColor",
-		name = "Very Low Warning Color",
-		description = "Configure the color of the overlay when charges are very low",
-		position = 1
+			keyName = "veryLowWarningColor",
```
Nightfirecat: This PR is applying the wrong tab depth to every single `@ConfigItem` annotation in this file. Please fix the indentation. (only one level further than the `@ConfigItem`)

tabs

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -34,273 +34,313 @@
...
-		description = ""
+			keyName = "explorerRing",
+			name = "",
+			description = ""
 	)
 	void explorerRing(int explorerRing);
 
+	@ConfigItem(
+			keyName = "showRingOfForgingCount",
+			name = "Ring of Forging Count",
```
Nightfirecat:
```suggestion
			name = "Show Ring of Forging Charges",
```

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -34,273 +34,313 @@
...
+		return true;
+	}
+
+	@ConfigItem(
+			keyName = "RingOfForgingNotification",
+			name = "Ring of Forging Notification",
+			description = "Configures if the Ring of Forging breaking notification is shown",
+			position = 21
+	)
+	default boolean ringOfForgingNotification()
```
Nightfirecat: Move this notification default method below the `ringOfForging` hidden and setter methods.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -75,10 +75,17 @@
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
 		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
 	private static final String BINDING_BREAK_TEXT = "Your Binding necklace has disintegrated.";
+	private static final Pattern RING_OF_FORGING_CHECK_PATTERN = Pattern.compile(
+			"You can smelt ([0-9+]+|one) more pieces of iron ore before a ring melts\\.");
```
Nightfirecat: Same indent issue here (and on the other added static vars), only increase the indentation level by one.
```suggestion
		"You can smelt ([0-9+]+|one) more pieces of iron ore before a ring melts\\.");
```

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -225,6 +240,31 @@ else if (bindingNecklaceCheckMatcher.find())
 
 				updateBindingNecklaceCharges(charges);
 			}
+			else if (ringOfForgingCheckMatcher.find())
+			{
+
```
Nightfirecat: Remove this empty line.
```suggestion
```

runelite-client/src/test/java/net/runelite/client/plugins/itemcharges/ItemChargePluginTest.java
```diff
@@ -52,6 +52,16 @@
 	private static final String PROTECT_1 = "Your dodgy necklace protects you. <col=ff0000>It has 1 charge left.</col>";
 	private static final String BREAK = "Your dodgy necklace protects you. <col=ff0000>It then crumbles to dust.</col>";
 
+	private static final String CHECK_RING_OF_FORGING_FULL = "You can smelt 140 more pieces of iron ore before a ring melts.";
+	private static final String CHECK_RING_OF_FORGING_ONE = "You can smelt one more pieces of iron ore before a ring melts.";
+	private static final String USED_RING_OF_FORGING = "You retrieve a bar of iron.";
+	private static final String BREAK_RING_OF_FORGING = "Your Ring of Forging has melted.";
+
+	private static final String CHECK_BINDING_NECKLACE_FULL = "You have 16 charges left before your Binding necklace disintegrates.";
```
Nightfirecat: Binding necklace tests should be added in a separate commit, if not a fully separate PR.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -75,10 +75,17 @@
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
 		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
 	private static final String BINDING_BREAK_TEXT = "Your Binding necklace has disintegrated.";
+	private static final Pattern RING_OF_FORGING_CHECK_PATTERN = Pattern.compile(
+		"You can smelt ([0-9+]+|one) more pieces of iron ore before a ring melts\\.");
+	private static final Pattern RING_OF_FORGING_USED_PATTERN = Pattern.compile(
```
Nightfirecat: This is a static string with no variance, so use a `String` and use `String#contains()` to check for it within the chat message subscriber. (see `BINDING_BREAK_TEXT`'s usage for an example)

```java
private static final String RING_OF_FORGING_USED_TEXT = "You retrieve a bar of iron.";
```

Adam-: Assuming the test is correct, `equals` should be fine.

Nightfirecat: The same change can be made to `RING_OF_FORGING_USED_PATTERN`, it should be a String instead.

Adam-:
Your test has a check for: `"You can smelt one more pieces of iron ore before a ring melts."` - did you get this from the game? It is not grammatically correct

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -75,10 +75,15 @@
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
 		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
 	private static final String BINDING_BREAK_TEXT = "Your Binding necklace has disintegrated.";
+	private static final Pattern RING_OF_FORGING_CHECK_PATTERN = Pattern.compile(
+		"You can smelt ([0-9+]+|one) more pieces of iron ore before a ring melts\\.");
+	private static final String RING_OF_FORGING_USED_TEXT = "You retrieve a bar of iron.";
+	private static final String RING_OF_FORGING_BREAK_TEXT = "Your Ring of Forging has melted.";
```
Nightfirecat: This message has some color applied, so this is not the correct exact string.

Ring of forging melt text

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -75,10 +75,15 @@
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
 		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
 	private static final String BINDING_BREAK_TEXT = "Your Binding necklace has disintegrated.";
+	private static final Pattern RING_OF_FORGING_CHECK_PATTERN = Pattern.compile(
+		"You can smelt ([0-9+]+|one) more pieces of iron ore before a ring melts\\.");
```
Alexsuperfly: when only one charge remains pieces -> piece

zachtyll:
> Your test has a check for: "You can smelt one more pieces of iron ore before a ring melts." - did you get this from the game? It is not grammatically correct

I just re-checked and it says "You can smelt one more piece of iron ore Before a ring melts.". Therefore it won't work as intended. When I tested this in game it was however not noticeable since we never see the item charge at 0 as the ring melts away. I could change the pattern to decrease to 0 at the last charge.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
zachtyll: Oh alright! I wasn't aware that colour was coded into the string itself. That explains why the recoil text looks odd too. Do you recommend using the same syntax and if so, which value is the purple text?

Nightfirecat: I'd recommend using the same syntax as the recoil text. Sadly, I don't know the value (not my screenshot), but the color from the image is `rgb(127, 0, 127)`, which would be `"<col=7f007f>"`.

zachtyll: Ok thanks! I just checked in game and the next Ring of Forging has 140 charges so the pattern matcher must have caught the message. I will alter the code anyway for the sake of correctness.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
zachtyll: Thank you! I added the greedy quantifier '?' we see in the other pattern matchers when the syntax changes on the last charge.

Adam-: You should be adding test cases which contain the actual text the game uses. At least one of your current test cases is just made up.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -75,10 +75,15 @@
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
 		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
 	private static final String BINDING_BREAK_TEXT = "Your Binding necklace has disintegrated.";
+	private static final Pattern RING_OF_FORGING_CHECK_PATTERN = Pattern.compile(
+		"You can smelt ([0-9+]+|one) more pieces? of iron ore before a ring melts\\.");
+	private static final String RING_OF_FORGING_USED_TEXT = "You retrieve a bar of iron.";
```
Nightfirecat: Isn't this message the same for smelting a bar of iron whether or not you have a ring of forging equipped? As-is, it would change the charge count even if you have no ring on.

zachtyll: Yes you are right! Is there a way to detect the itemID for what is equipped in the ring slot?

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -32,14 +32,8 @@
...
-import net.runelite.api.ChatMessageType;
-import net.runelite.api.Client;
-import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemContainer;
-import net.runelite.api.ItemID;
-import net.runelite.api.Varbits;
+
+import net.runelite.api.*;
```
Nightfirecat: Don't use wildcard imports: https://github.com/runelite/runelite/wiki/Code-Conventions#imports

zachtyll:
The slayer plugin makes use of getters and setters for testing the different bracelets one would use when slaying. Is this a recommended way of implementing tests for things that need to be worn or is there a better way to do it?

PR #9378, opened 2019-07-15
XP tracker: Fix issue where progress bar shows 100.00% before level up

jdkleuver:
Saw this in discord:
2019-07-15-200239_679x101_scrot
Before:
2019-07-15-200045_227x63_scrot
After:
2019-07-15-200149_237x64_scrot

My first pull request so let me know if I did anything wrong!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
@@ -28,6 +28,7 @@ package net.runelite.client.plugins.xptracker;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
+import java.math.RoundingMode;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


deathbeam:
Should this be really solved like this? this looks like rounding issue, so maybe avoid rounding

abextm:
Probably could just configure the DecimalFormat to round down

jdkleuver:
I took your feedback into account, see new commit

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -226,6 +227,7 @@ private void rebuildAsync(boolean updated, boolean skillPaused, XpSnapshotSingle
 
 			// Update progress bar
 			progressBar.setValue((int) xpSnapshotSingle.getSkillProgressToGoal());
+			TWO_DECIMAL_FORMAT.setRoundingMode(RoundingMode.DOWN);
```
deathbeam: this should be in static { } constructor, e.g

```java
static
{
    			TWO_DECIMAL_FORMAT.setRoundingMode(RoundingMode.DOWN);
}
```

jdkleuver:
@deathbeam is that change ok? tbh I've never used a static block like that before so thanks for bringing it to my attention

deathbeam:
Yeah, looks fine

PR #9380, opened 2019-07-15
Fixes #9338. Adds Nets, Greegrees, Bones and Waterskins to Item Ident…

### LINKED ISSUE #9338: Item Identification Suggestion (Fishing Nets, Greegrees, Bones, etc.)
Not a problem.

Item Identification should also tag fishing nets, as they are hard to distinguish between small, big, drift net, etc. Additionally waterskins (0-4) all have the same model but no graphical way to distinguish what has what amount in the bank, but Item Charges does show this while in inventory. Bones as well are hard to distinguish between what types are held as many use the same graphic model. Same logic for monkey greegrees as they are tinted but hard to tell which is which although these are so rarely used I don't know if its worth adding the dev time for them.

Alternatives could be to just tint the items different colors with an overlay or have the item highlight tags show in the bank as well, but this seems like a best fit for the Item ID plugin.

**Additional context**
N/A

zachtyll:
…ification plugin.


runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
```diff
@@ -119,7 +119,65 @@
...
+	SHEEP_2(Type.BONE, "Sheep 2", "S2", ItemID.SHEEP_BONES_2),
+	SHEEP_3(Type.BONE, "Sheep 3", "S3", ItemID.SHEEP_BONES_3),
+	SHEEP_4(Type.BONE, "Sheep 4", "S4", ItemID.SHEEP_BONES_4),
+	SUPERIOR_DRAGON(Type.BONE, "Sup Drag", "SD", ItemID.SUPERIOR_DRAGON_BONES),
+	WOLF(Type.BONE, "Wolf", "WF", ItemID.WOLF_BONES),
+	WYRM(Type.BONE, "Wyrm", "WR", ItemID.WYRM_BONES),
+	WYVERN(Type.BONE, "Wyvern", "WV", ItemID.WYVERN_BONES),
+	ZOGRE(Type.BONE, "Zogre", "Z", ItemID.ZOGRE_BONES),
+
+	//Waterskins
```
xDemoN: I think waterskins would make more sense in the item charges plugin

zachtyll: Yes I thought so too but appearently the user wanted it for seeing which waterskins hold which charges in their bank.
If this is superflous it could be removed.

deathbeam:
Do we really need this? These items are pretty different from eachother

PR #9382, opened 2019-07-16
Add Splash screen

TheStonedTurtle:
Revised version of #3497. Related to runelite/launcher#64

D20UsA5R0B

There's also a commit that adds client-sided launcher version check. This prevents the client from loading if launcher is outdated.

<details>
<p>

splash-screen:

no-splash:

</p>
</details>

The Splash Screen code can be found here:
<https://github.com/TheStonedTurtle/splashscreen>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -34,6 +34,7 @@ import java.io.File;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
 import java.util.Locale;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.inject.Provider;
 import javax.inject.Singleton;
@@ -82,7 +83,9 @@ public class RuneLite
 	private static final File LOGS_DIR = new File(RUNELITE_DIR, "logs");
 	private static final File LOG_FILE = new File(LOGS_DIR, "client.log");
 	private static final RuneLiteProperties PROPERTIES = new RuneLiteProperties();
-	private static final String EXPECTED_LAUNCHER_VERSION = "Launcher 1.6.0";
+	// Launcher was updated to check it's own version starting at v2.
+	// If they are using a version below 2 (Launcher 1.X.X) the client needs to force them to dl the updated launcher
+	private static final Pattern OUTDATED_LAUNCHER_VERSION = Pattern.compile("Launcher 1\\..*");
 
 	@Getter
 	private static Injector injector;
@@ -219,7 +222,7 @@ public class RuneLite
 		setSplashMessage("Checking launcher version", 0);
 		final String launcherVersion = RuneLiteProperties.getLauncherVersion();
 		if (launcherVersion != null &&
-			!EXPECTED_LAUNCHER_VERSION.equalsIgnoreCase(launcherVersion))
+			OUTDATED_LAUNCHER_VERSION.matcher(launcherVersion).matches())
 		{
 			SwingUtilities.invokeAndWait(() ->
 			{
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -81,6 +81,7 @@ public class RuneLite
 	public static final File SCREENSHOT_DIR = new File(RUNELITE_DIR, "screenshots");
 	private static final File LOGS_DIR = new File(RUNELITE_DIR, "logs");
 	private static final RuneLiteProperties PROPERTIES = new RuneLiteProperties();
+	private static final String EXPECTED_LAUNCHER_VERSION = "Launcher 1.6.0";
 
 	@Getter
 	private static Injector injector;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -212,6 +220,32 @@ public ClientUpdateCheckMode convert(String v)
 			}
 		});
 
+		SwingUtilities.invokeAndWait(() ->
+		{
+			RuneLiteSplashScreen.setTheme();
+			if (options.has("no-splash"))
+			{
+				SPLASH_SCREEN = new RuneLiteSplashScreen(new File(LOGS_DIR, "client.log"));
```
deathbeam: this new File(LOGS_DIR, "client.log") should be just LOG_FILE constant at top

deathbeam: also, honestly, these constructors do not make sense.Why is file not at first place in both scenarios, and why no-splash still needs splash screen, and why adding string to splash screen constructor changes behaviour of it?

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -212,6 +220,32 @@ public ClientUpdateCheckMode convert(String v)
...
+				SPLASH_SCREEN = new RuneLiteSplashScreen("Runelite " + PROPERTIES.getVersion(), new File(LOGS_DIR, "client.log"));
+			}
+		});
+
+		SPLASH_SCREEN.setMessage("Checking launcher version", 0);
+		final String launcherVersion = RuneLiteProperties.getLauncherVersion();
+		if (launcherVersion != null &&
+			!EXPECTED_LAUNCHER_VERSION.equalsIgnoreCase(launcherVersion))
+		{
+			SwingUtilities.invokeAndWait(SPLASH_SCREEN::invalidVersion);
```
deathbeam: Why is invalidVersion part of SPLASH_SCREEN? will it be used also in launcher? if no, why is it there

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockManager.java
```diff
@@ -57,7 +57,12 @@
 	private final List<Stopwatch> stopwatches = new ArrayList<>();
 
 	@Getter
-	private ClockTabPanel clockTabPanel = new ClockTabPanel(this);
+	private ClockTabPanel clockTabPanel;
+
+	ClockManager()
+	{
+		SwingUtilities.invokeLater(() -> clockTabPanel = new ClockTabPanel(this));
```
deathbeam: This should be separate commit, and also explain why are you doing this

runelite-client/src/main/java/net/runelite/client/RuneLite.java
TheStonedTurtle: I create the splash screen even when no-splash is specified so I didn't have to null-check SPLASH_SCREEN at every `SPLASH_SCREEN.setMessage()`. It also lets me use the splash screens dialog popups which i think should be displayable regardless of if the splash screen is being displayed.

runelite-client/src/main/java/net/runelite/client/RuneLite.java
TheStonedTurtle: I created the version check for the launcher but in order to force people to download the updated launcher we needed a version check in the client (at least temporarily)

runelite-client/src/main/java/net/runelite/client/RuneLite.java
deathbeam: I would rather do proxy method and null checks. If you want just the dialog popups, separate them from splash screen

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockManager.java
TheStonedTurtle: The PR i linked had this change as it was causing a component creation off EDT issue for this file. It looks like this is no longer an issue though as I reverted this change and am not getting the error anymore. I don't understand why it's no longer an issue but I've reverted this change

TheStonedTurtle:
I've rebased the PR after #9428 was merged. I also updated the ClientUI error message referencing the log to use the splash screen error prompt.

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -77,6 +80,14 @@
 	public static final File RUNELITE_DIR = new File(System.getProperty("user.home"), ".runelite");
 	public static final File PROFILES_DIR = new File(RUNELITE_DIR, "profiles");
 	public static final File SCREENSHOT_DIR = new File(RUNELITE_DIR, "screenshots");
+	private static final File LOGS_DIR = new File(RUNELITE_DIR, "logs");
+	private static final File LOG_FILE = new File(LOGS_DIR, "client.log");
+	private static final RuneLiteProperties PROPERTIES = new RuneLiteProperties();
+	// Launcher was updated to check it's own version starting at v2.
+	// If they are using a version below 2 (Launcher 1.X.X) the client needs to force them to dl the updated launcher
+	private static final Pattern OUTDATED_LAUNCHER_VERSION = Pattern.compile("Launcher 1\\..*");
```
abextm: rather than having to do this with regex, which is difficult, it would probable be easier to just use strcmp with the last good version, and hope we don't need to release a 2 digit version number

PR #6701, opened 2018-11-26
Nmz points per hour and NMZ total points on overlay

Haashi:
Fixes : #6301 #6268 

image

image

This is my first contribution so hopefully this get some suggestions for improvement.

PR #9388, opened 2019-07-16
nmz: Add total and per hour points to NMZ overlay

### LINKED ISSUE #6268: Add Points Per Hour marker at NMZ
I'm trying to maximise my points per hour at NMZ and can figure out the amount of points I'm getting per hour by figuring out how many doses of overloads I've taken and multiplying it until it reaches the equivalent of 1 hour of doses of overloads

for example, 100k points in the first 5 minutes means 1.2m points per hour. 150k points in 10 minutes means 900k points per hour

not sure of the formula required but surely this wouldn't be super complex
pph

### LINKED ISSUE #7607: Show total Nightmare Zone points
When inside a Nightmare Zone dream (session) both the vanilla client Nightmare Zone interface and Runelite's Nightmare Zone plugin show the number of Nightmare Zone points earned within the dream you're currently in, but not the total points (i.e. points you had before you entered the dream + points earned during the current dream). This is inconvenient because it may take multiple dreams to get the number of points needed to imbue something, in which case you have to remember how many points you had before you entered the dream in order to know when you have earned enough points rather than being able to see it in the interface.

Would it be possible to show the total number of Nightmare Zone points in the Nightmare Zone plugin's overlay, in addition to the number of points earned just in the current dream already shown?

adwitkow:
closes #6268 and closes #7607 and closes #6701 

nmz-points

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZoneOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZoneOverlay.java
@@ -30,6 +30,7 @@ import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.ItemID;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
+import net.runelite.api.VarPlayer;
 import net.runelite.api.Varbits;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetInfo;
@@ -101,10 +102,21 @@ class NightmareZoneOverlay extends Overlay
 
 		renderAbsorptionCounter();
 
+		final int currentPoints = client.getVar(Varbits.NMZ_POINTS);
+		final int totalPoints = currentPoints + client.getVar(VarPlayer.NMZ_REWARD_POINTS);
+
 		panelComponent.getChildren().clear();
 		panelComponent.getChildren().add(LineComponent.builder()
 			.left("Points: ")
-			.right(StackFormatter.formatNumber(client.getVar(Varbits.NMZ_POINTS)))
+			.right(StackFormatter.formatNumber(currentPoints))
+			.build());
+		panelComponent.getChildren().add(LineComponent.builder()
+			.left("Points/Hour: ")
+			.right(StackFormatter.formatNumber(plugin.getPointsPerHour()))
+			.build());
+		panelComponent.getChildren().add(LineComponent.builder()
+			.left("Total: ")
+			.right(StackFormatter.formatNumber(totalPoints))
 			.build());
 
 		return panelComponent.render(graphics);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
@@ -25,8 +25,11 @@
 package net.runelite.client.plugins.nightmarezone;
 
 import com.google.inject.Provides;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZoneOverlay.java
```diff
@@ -148,6 +175,39 @@ public void removeAbsorptionCounter()
...
+		Instant now = Instant.now();
+
+		if (nmzSessionStartTime == null)
+		{
+			nmzSessionStartTime = now;
+		}
+
+		if (lastPoints == currentPoints)
+		{
+			// prevent Points/Hour from flickering aggresively
```
Nightfirecat: I'd kind of like to see this calculation happen in the plugin per game tick so that we can keep the logic consolidated there and avoid having this problem. That would also significantly simplify the overlay logic to simply building a panel with values fetched from the plugin.

Quasindro: Pushed the requested changes.

I think I screwed up by squashing the commits, didn't I? Apologies for that.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
```diff
@@ -193,6 +211,32 @@ private void checkAbsorption()
...
+
+		if (!timeSinceStart.isZero())
+		{
+			return (int) ((double) currentPoints * (double) HOUR.toMillis() / (double) timeSinceStart.toMillis());
+		}
+
+		return pointsPerHour;
+	}
+
+	private void resetPointsPerHour()
```
Nightfirecat: This should also be called on plugin shutdown.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
```diff
@@ -193,6 +213,32 @@ private void checkAbsorption()
...
+		}
+
+		Duration timeSinceStart = Duration.between(nmzSessionStartTime, now);
+
+		if (!timeSinceStart.isZero())
+		{
+			return (int) ((double) currentPoints * (double) HOUR.toMillis() / (double) timeSinceStart.toMillis());
+		}
+
+		return pointsPerHour;
```
Adam-: This should probably be 0? Or -1 and then don't show it in the overlay?

Adam-:
This seems fine. There is also #6701 which it closes too.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
adwitkow: Changed it to 0. I think that restricting the overlay based on -1 there would be unnecessary since it would be -1 only during the first tick of a NMZ session.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
```diff
@@ -67,6 +71,9 @@
 	@Inject
 	private NightmareZoneOverlay overlay;
 
+	@Getter
+	private int pointsPerHour;
+	private Instant nmzSessionStartTime;
```
deathbeam: Add empty line below (and above)

adwitkow: added

PR #9390, opened 2019-07-17
GE Plugin: Add offer data to status widget

### LINKED ISSUE #9377: Add OSbuddy price data on sell offers
Would it be possible to add the OSBuddy price data on your current sell offers after the offer has gone through, instead of it only being there when you are making the initial offer. Sometimes I put items in at an offer and I forget what the current market was at.

When posting the offer:
https://i.imgur.com/QBoZLyy.png

When you recheck the sell offer:
https://i.imgur.com/SZ6CQhM.png

Serpa88:
Closes #9377 

image


Nightfirecat:
I don't remember where this was discussed, but Adam thought this would likely be better handled via script callback from the script which generates this text, as we've had issues with text getting duplicated or becoming too long in the past because of code touching this widget in multiple places.

DSerp:
Brought up to date and cleaned up code a bit. This would also show GE limits since this still uses setExamineText and finds the itemId without the PlayerVar

PR #1286, opened 2018-04-05
Motherlode Mine - Sack overlay and region check

ypperlig:
The following changes have been done:

* Added red color to the "Pay-dirt in sack" overlay when the number of pay-dirt in sack exceeds or is equal to the max allowed.
* Added "Deposits left" to sack overlay. This shows the number of "inventories" left before the sack will be full. This is colored red when deposits left are 0.
* Removed "You are mining" / "You are not mining" as this could not be shown accurately. This showed "You are not mining" when animation stopped, although the character would continue mining straight after.
* Added isInMLM (region check) to most functions

Screenshots:
Overlay
OverlayRed

PR #9402, opened 2019-07-19
Block LMS player loots from loot tracker

psikoi:
Prevents any loot from player kills inside Last Man Standing from appearing in the loot tracker

Also fixed a typo in the herbiboar bit

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -105,7 +105,7 @@ public class LootTrackerPlugin extends Plugin
 
 	// Herbiboar loot handling
 	private static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
-	private static final String HERBIBOR_EVENT = "Herbiboar";
+	private static final String HERBIBOAR_EVENT = "Herbiboar";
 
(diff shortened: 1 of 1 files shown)


deathbeam:
It should be always saved, like hidden drops always are, and just be hidden from UI.

Also, instead of config option, imo simply skipping player drop events inside of LMS is better, because noone cares about that there (and that was also basically the request of the reporter)

psikoi:
@deathbeam updated, as requested

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -115,6 +116,9 @@
 		13113, "Larran's small chest"
 	);
 
+	// Last man standing map regions
+	private static final int[] LAST_MAN_STANDING_REGIONS = {13658, 13659, 13914, 13915, 13916};
```
ypperlig: Should this be an immutable set? Reference:
https://github.com/runelite/runelite/pull/1286#discussion_r179928302

psikoi: Good call, should be good now

deathbeam:
needs rebase

Adam-:
I don't see why the items need to be saved at all. They aren't even real.

deathbeam:
Yeah, for LMS specifically my previous comment about saving is not relevant. The original change was about hiding player loot (by not saving it).

PR #9407, opened 2019-07-19
Woodcutting: Add respawn timer

### LINKED ISSUE #9397: Tree respawn timer
I feel like we’re missing something, when mining rocks we see a respawn timer but when cutting trees we don’t, this is really the case with yew+ level trees

I would like to see the exact same solution added to this like the one added to ores, but maybe a slider to change the size of the timer circle or a number only view (time in text)

Dava96:
t8XEHRRil6

closes #9397

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
@@ -25,20 +25,52 @@
 package net.runelite.client.plugins.woodcutting;
 
 import com.google.common.collect.ImmutableMap;
+import java.time.Duration;
 import java.util.Map;
+import lombok.AccessLevel;
 import lombok.Getter;
+import static net.runelite.api.ObjectID.TREE;
+import static net.runelite.api.ObjectID.TREE_1277;
+import static net.runelite.api.ObjectID.TREE_1278;
+import static net.runelite.api.ObjectID.TREE_1279;
+import static net.runelite.api.ObjectID.TREE_1280;
+import static net.runelite.api.ObjectID.OAK_TREE;
+import static net.runelite.api.ObjectID.OAK_TREE_4540;
+import static net.runelite.api.ObjectID.OAK_10820;
+import static net.runelite.api.ObjectID.WILLOW;
+import static net.runelite.api.ObjectID.WILLOW_10833;
+import static net.runelite.api.ObjectID.WILLOW_10831;
+import static net.runelite.api.ObjectID.MAPLE_TREE;
+import static net.runelite.api.ObjectID.MAPLE_TREE_10832;
+import static net.runelite.api.ObjectID.TEAK;
+import static net.runelite.api.ObjectID.MAHOGANY;
+import static net.runelite.api.ObjectID.YEW;
+import static net.runelite.api.ObjectID.MAGIC_TREE_10834;
 import static net.runelite.api.ObjectID.REDWOOD;
+import static net.runelite.api.ObjectID.REDWOOD_29669;
+import static net.runelite.api.ObjectID.REDWOOD_29671;
 import static net.runelite.api.ObjectID.REDWOOD_29670;
 
 @Getter
 enum Tree
 {
-	REDWOOD_TREE_SPAWN(REDWOOD, REDWOOD_29670);
+	REGULAR_TREES(Duration.ofSeconds(WoodcuttingTimersOverlay.TREE_MAX_RESPAWN_TIME), TREE, TREE_1277, TREE_1278, TREE_1279, TREE_1280),
+	OAK_TREES(Duration.ofMillis(8500), OAK_TREE, OAK_TREE_4540, OAK_10820),
+	WILLOW_TREES(Duration.ofMillis(8500), WILLOW, WILLOW_10833, WILLOW_10831),
+	MAPLE_TREES(Duration.ofSeconds(35), MAPLE_TREE, MAPLE_TREE_10832),
+	TEAK_TREES(Duration.ofMillis(8500), TEAK),
+	MAHOGANY_TREES(Duration.ofMillis(8500), MAHOGANY),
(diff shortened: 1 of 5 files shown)


Dava96:
Are there any changes I need to make to this?

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
```diff
@@ -0,0 +1,135 @@
...
+			LocalPoint loc = LocalPoint.fromWorld(client, worldPoint);
+
+			// This is used to get the X and Y for centre location to work
+			LocalPoint tempLoc = LocalPoint.fromWorld(client, worldPoint.getX(), worldPoint.getY());
+			LocalPoint centreLoc = new LocalPoint(
+				tempLoc.getX() + Perspective.LOCAL_TILE_SIZE * (client.getPlane() - 1) / 2,
+				tempLoc.getY() + Perspective.LOCAL_TILE_SIZE * (client.getPlane() - 1) / 2);
+			if (loc == null || percent > 1.0f)
+			{
+				it.remove();
```
Nightfirecat: Why is the overlay responsible for removing trees which have respawned? That logic should be contained in the client, and the overlay should be responsible only for drawing the respawn timers.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
```diff
@@ -0,0 +1,135 @@
...
+				pieBorderColor = DARK_GREEN;
+			}
+
+			// The pie is already centred on these trees
+			if (tree == Tree.OAK_TREES || tree == Tree.YEW_TREES || tree == Tree.MAHOGANY_TREES || tree == Tree.TEAK_TREES || tree == Tree.REDWOODS)
+			{
+				ProgressPieComponent ppc = new ProgressPieComponent();
+				ppc.setBorderColor(pieBorderColor);
+				ppc.setFill(pieFillColor);
+				ppc.setPosition(point);
```
Nightfirecat: All of the code in this block is duplicated below, with the single difference being `setPosition(point)` versus `setPosition(centrePoint)`. Please de-duplicate this.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
```diff
@@ -0,0 +1,135 @@
...
+			Tree tree = treeRespawn.getTree();
+
+			// Recolour pie on regular trees within the range they spawn
+			if (tree == Tree.REGULAR_TREES && percent > TREE_RANDOM_PERCENT_THRESHOLD)
+			{
+				pieFillColor = Color.GREEN;
+				pieBorderColor = DARK_GREEN;
+			}
+
+			// The pie is already centred on these trees
```
Nightfirecat: Why is that the case? Some explanation is needed here.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
```diff
@@ -0,0 +1,135 @@
...
+import net.runelite.api.coords.LocalPoint;
+import net.runelite.api.coords.WorldPoint;
+import net.runelite.client.ui.overlay.Overlay;
+import net.runelite.client.ui.overlay.OverlayLayer;
+import net.runelite.client.ui.overlay.OverlayPosition;
+import net.runelite.client.ui.overlay.components.ProgressPieComponent;
+
+class WoodcuttingTimersOverlay extends Overlay
+{
+	// Range of Trees respawn time
```
Nightfirecat: This should be explicit in explaining that regular trees have a variable respawn time.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
Dava96: This is how it's done with the mining timers overlay, which is the reason I've done it here

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTimersOverlay.java
Dava96: It is like this because some trees are on a different offset than others which makes the timer to display incorrectly like the magic tree stump. Although there could be a better way to do it than what I've done.
@Nightfirecat 
centred vs uncentred

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/Tree.java
```diff
@@ -25,20 +25,52 @@
...
+	WILLOW_TREES(Duration.ofMillis(8500), WILLOW, WILLOW_10833, WILLOW_10831),
+	MAPLE_TREES(Duration.ofSeconds(35), MAPLE_TREE, MAPLE_TREE_10832),
+	TEAK_TREES(Duration.ofMillis(8500), TEAK),
+	MAHOGANY_TREES(Duration.ofMillis(8500), MAHOGANY),
+	YEW_TREES(Duration.ofMinutes(1), YEW),
+	MAGIC_TREES(Duration.ofMinutes(2), MAGIC_TREE_10834),
+	REDWOODS(Duration.ofMinutes(2), REDWOOD, REDWOOD_29669, REDWOOD_29670, REDWOOD_29671);
 
+	private final Duration respawnTime;
+	@Getter(AccessLevel.PACKAGE)
```
Nightfirecat: Why does this have a getter? Nothing uses it.

thelg59:
This would be cool to see, if it's still a possibility.

Adam-:
added in 3fd8fb79aa171c924af5f3cb577ec9dce404e4da

PR #9414, opened 2019-07-20
Timers Plugin: Remove Tb when TeleBlocking Opponent Dies

### LINKED ISSUE #9409: Teleblock timer should be removed when killing the player who cast it on you
Upon killing the player who cast the spell on you, your tele block is removed. This adds a chat message in the format: (color unknown, probably the same color as TB message)

```
<col=??????>Your Tele Block has been removed because you killed Zezima.
```

rlw0014:
Closes  #9409

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -104,6 +104,7 @@ public class TimersPlugin extends Plugin
 	private static final String SUPER_ANTIFIRE_DRINK_MESSAGE = "You drink some of your super antifire potion";
 	private static final String SUPER_ANTIFIRE_EXPIRED_MESSAGE = "<col=7f007f>Your super antifire potion has expired.</col>";
 	private static final String SUPER_ANTIVENOM_DRINK_MESSAGE = "You drink some of your super antivenom potion";
+	private static final String KILLED_TELEBLOCK_OPPONENT_TEXT = "<col=4f006f>Your Tele Block has been removed because you killed ";
 
 	private static final Pattern DEADMAN_HALF_TELEBLOCK_PATTERN = Pattern.compile("<col=4f006f>A Tele Block spell has been cast on you by (.+). It will expire in 1 minute, 15 seconds.</col>");
 	private static final Pattern FULL_TELEBLOCK_PATTERN = Pattern.compile("<col=4f006f>A Tele Block spell has been cast on you by (.+). It will expire in 5 minutes, 0 seconds.</col>");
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -108,6 +108,7 @@
 	private static final Pattern DEADMAN_HALF_TELEBLOCK_PATTERN = Pattern.compile("<col=4f006f>A Tele Block spell has been cast on you by (.+). It will expire in 1 minute, 15 seconds.</col>");
 	private static final Pattern FULL_TELEBLOCK_PATTERN = Pattern.compile("<col=4f006f>A Tele Block spell has been cast on you by (.+). It will expire in 5 minutes, 0 seconds.</col>");
 	private static final Pattern HALF_TELEBLOCK_PATTERN = Pattern.compile("<col=4f006f>A Tele Block spell has been cast on you by (.+). It will expire in 2 minutes, 30 seconds.</col>");
+	private static final Pattern KILLED_TELEBLOCK_OPPONENT = Pattern.compile("<col=4f006f>Your Tele Block has been removed because you killed (.+).</col>");
```
Nightfirecat: We don't actually end up using the matched group at all, so this would be much better off being a `String`, and matched using `event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT)`:

```suggestion
	private static final String KILLED_TELEBLOCK_OPPONENT_TEXT = "<col=4f006f>Your Tele Block has been removed because you killed ";
```

Alexsuperfly:
do you have to kill them, or can they be killed by another player

rlw0014:
> do you have to kill them, or can they be killed by another player

just tested this. you have to kill your opponent who teleblocked you for your teleblock status to be removed so the PR _should_ be correct now

Nightfirecat:
I believe this happens whenever the player who cast the tele block dies (regardless of who killed them) but that's not really too important for this code. The important thing is that we correctly remove the timer upon receiving that message. I'll test this ingame this weekend and let you know whether it works.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -542,6 +543,10 @@ else if (DEADMAN_HALF_TELEBLOCK_PATTERN.matcher(event.getMessage()).find())
 			{
 				createGameTimer(DMM_HALFTB);
 			}
+			else if (event.getMessage().startsWith(KILLED_TELEBLOCK_OPPONENT_TEXT))
```
Adam-: why `startsWith`? what is the full message?

Nightfirecat: The full message is `"<col=4f006f>Your Tele Block has been removed because you killed (.+).</col>"`, where `(.+)` is the name of the player who teleblocked you.

PR #9422, opened 2019-07-21
Precompile the split regex in RaidsPlugin

WilliamCollishaw:
Performance enhancement, prevents Pattern.compile(regex) being run more than necessary.


runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -80,7 +80,7 @@
 	private static final String RAID_COMPLETE_MESSAGE = "Congratulations - your raid is complete!";
 	private static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("###.##");
 	static final DecimalFormat POINTS_FORMAT = new DecimalFormat("#,###");
-	private static final String SPLIT_REGEX = "\\s*,\\s*";
+	private static final Pattern SPLIT_REGEX = Pattern.compile("\\s*,\\s*");
```
deathbeam: When changing this, I would just use Text.fromCSV instead

deathbeam:
Updated to use Text.fromCSV

WilliamCollishaw:
@deathbeam changes look good to me!

PR #9423, opened 2019-07-21
grounditems: add despawn timers

Adam-:
Co-authored-by: Matthew Smith <matthew.smith813@gmail.com>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
@@ -45,10 +45,8 @@ class GroundItem
 	private int gePrice;
 	private int offset;
 	private boolean tradeable;
-	/**
-	 * Is loot received from a drop
-	 */
-	private boolean isMine;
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
```diff
@@ -43,7 +45,14 @@
 	private int gePrice;
 	private int offset;
 	private boolean tradeable;
-	private boolean isMine;
+	private boolean isPvpLoot;
```
deathbeam: should this be just Enum or something? or boolean? Because it can be either pvp or pvm, not both.

Adam-: Could do an enum I guess but there are 3 separate things: pvm, pvp, neither.

deathbeam: Yeah, I realized after I commented. Then maybe 3 state enum (PvP, PvM, Unknown)?

PR #9435, opened 2019-07-23
Add option to highlight items already stacked in inventory

### LINKED ISSUE #9332: Automatically highlight stackable items which you already have a stack of
When doing slayer I want to highlight and pick up the stackable items but don't want to have to have a bunch of random items highlighted for when im not doing slayer, and don't want to have to re highlight and unhighlight a bunch of items for every single task. 

A solution for this would be if stackable items which you already have a stack of in your inventory could be automatically highlighted.

dekvall:
Adds an option to the Ground Items plugin to highlight stackable items which already exists in the players inventory. This is useful when training slayer.

Closes #9332

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -369,4 +369,15 @@ public interface GroundItemsConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "inventoryStacks",
+		name = "Inventory Stacks",
+		description = "Highlights items already stacked in inventory",
+		position = 29
+	)
+	default boolean inventoryStacks()
+	{
+		return false;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -44,6 +44,8 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
@@ -51,7 +53,9 @@ import lombok.Setter;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.GameState;
+import net.runelite.api.InventoryID;
 import net.runelite.api.ItemComposition;
+import net.runelite.api.ItemContainer;
 import net.runelite.api.ItemID;
 import net.runelite.api.ItemLayer;
 import net.runelite.api.MenuAction;
@@ -66,12 +70,14 @@ import net.runelite.api.events.ClientTick;
 import net.runelite.api.events.ConfigChanged;
 import net.runelite.api.events.FocusChanged;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.ItemDespawned;
 import net.runelite.api.events.ItemQuantityChanged;
 import net.runelite.api.events.ItemSpawned;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.client.Notifier;
+import net.runelite.client.callback.ClientThread;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.NpcLootReceived;
@@ -135,6 +141,7 @@ public class GroundItemsPlugin extends Plugin
 
 	private List<String> hiddenItemList = new CopyOnWriteArrayList<>();
 	private List<String> highlightedItemsList = new CopyOnWriteArrayList<>();
+	private final Set<String> inventoryStacksSet = new HashSet<>();
 
 	@Inject
 	private GroundItemInputListener inputListener;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -358,4 +358,26 @@ default boolean collapseEntries()
...
+
+	@ConfigItem(
+			keyName = "inventoryStacksColor",
+			name = "Stackable color.",
+			description = "Configures the highlight color for items already stacked in inventory",
+			position = 29
+	)
+	default Color inventoryStacksColor()
+	{
+		return Color.decode("#2AECFA");
```
Nightfirecat: This should just use the existing "highlight" color instead of having its own highlight color imo.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -667,4 +687,38 @@ private void notifyHighlightedItem(GroundItem item)
...
+		{
+			editInventoryStacks();
+		}
+	}
+
+	void editInventoryStacks()
+	{
+		final ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
+
+		if (itemContainer != null)
```
Nightfirecat: Use a guard clause instead. eg.

```java
if (itemContainer == null)
{
	return;
}

// code continues here
```

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -667,4 +687,38 @@ private void notifyHighlightedItem(GroundItem item)
...
+					.collect(Collectors.toSet());
+
+			if (!inventoryStacksSet.equals(stackedItemsInInventory))
+			{
+				inventoryStacksSet = stackedItemsInInventory;
+				inventoryStacks = CacheBuilder.newBuilder()
+						.maximumSize(32L)
+						.expireAfterAccess(10, TimeUnit.MINUTES)
+						.build(new WildcardMatchLoader(new ArrayList<>(inventoryStacksSet)));
+
```
Nightfirecat: Remove empty line

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -667,4 +687,38 @@ private void notifyHighlightedItem(GroundItem item)
...
+			}
+
```
Nightfirecat: Remove empty line

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -667,4 +687,38 @@ private void notifyHighlightedItem(GroundItem item)
...
+	}
+
+	void editInventoryStacks()
+	{
+		final ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
+
+		if (itemContainer != null)
+		{
+			final Set<String> stackedItemsInInventory = Stream.of(itemContainer.getItems())
+					.map(item -> itemManager.getItemComposition(item.getId()))
```
Nightfirecat: This should only indent one time further than the previous line.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
dekvall: My thinking was that the existing highlight color was "sacred" in a way and reserved for items that you "really would want to" pick up, while this option is for items that you "might as well" pick up. I felt that different player behavior should present in different colors. 

With that said, this is my own opinion and maybe I am misunderstanding how the highlight list is meant to be used.

Should these stacks work like the normal highlight list in every other way as well? Right now, if the item is in the hidden list, but also in the player's inventory, the item is not highlighted. This is to avoid cluttering the screen with less valuable stackable drops the player might have in its inventory, such as coins or runes. The player also don't get desktop notifications for items that are already stacked in its inventory, should they?

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -565,12 +580,17 @@ Color getHighlighted(String item, int gePrice, int haPrice)
...
 		}
 
-		// Explicit hide takes priority over implicit highlight
+		// Explicit hide takes priority over implicit highlight and inventory stacks
 		if (TRUE.equals(hiddenItems.getUnchecked(item)))
 		{
 			return null;
 		}
 
+		if (config.inventoryStacks() && inventoryStacks != null &&  TRUE.equals(inventoryStacks.getUnchecked(item)))
```
Nightfirecat: There's a double space in this `if` statement.

```suggestion
		if (config.inventoryStacks() && inventoryStacks != null && TRUE.equals(inventoryStacks.getUnchecked(item)))
```

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -164,6 +170,7 @@
 	private final Map<Integer, Color> priceChecks = new LinkedHashMap<>();
 	private LoadingCache<String, Boolean> highlightedItems;
 	private LoadingCache<String, Boolean> hiddenItems;
+	private LoadingCache<String, Boolean> inventoryStacks;
```
Nightfirecat: Caches are only used for hidden/highlighted items so that they support wildcard matching. Since stacked inventory items aren't specified via user input--and thus will not include wildcards--they won't need a cache. Use `inventoryStacksSet` instead.

dekvall:
Fixed review comments and adhered to the rules of `getItemComposition` by letting it run on `ClientThread`, I missed that assertion earlier.

dekvall:
Superseded by #13858

PR #9451, opened 2019-07-25
Use npcId to get health from NPCManager

### LINKED ISSUE #9222: Tutorial Island Giant Rat listed with wrong HP in combat overlay
**Describe the bug**
Tutorial island giant rat shows up with 5 hp in the combat overlay, it actually has 3.

**To Reproduce**
1. Go to tutorial island
2. Enter combat with a giant rat with the combat overlay enabled
3. HP is tracked starting from 5/5

**Expected behavior**
Max HP listed as 5, rat actually has 3.

 - OS: Debian
 - RuneLite version: 1.5.27
 - Launcher version: 1.6.2

dekvall:
Since the `ItemStats` plugin uses ItemIDs to lookup stats for items, we should use npcIDs when looking up npchealth. The json file was grabbed from [this pull request](https://github.com/runelite/static.runelite.net/blob/242a657f4f385d632afd44fe6f599a3d9a8ce39a/npc/stats.json) and after a while of testing it seems accurate up to the point of tutorial island rats. 

The rats on tutorial island have an npcID of `3313` and according to the original json should have a health of 5. I believe this is because there are no `Giant rat` with a health of 3 anywhere on the wiki and thus the scraper script figured that a health is appropriate as that is the health of all other level 3 Giant rats. I visited all places with level 3 Giant rats listed on the [Giant rat page](https://oldschool.runescape.wiki/w/Giant_rat) on the wiki and could not find any other rat with an id of `3313`. Thus, I deemed it safe to set the value for that entry to 3. 

Closes #9222


TheStonedTurtle:
How exactly is this different from #8463?

dekvall:
It's not, forgive me. For some reason i only saw the pr to static.runelite.
In essence I was just trying to fix #9222 and this seemed like the easiest change. 

I will close this now.

PR #9473, opened 2019-07-26
skill calc: SotE additions

Alexsuperfly:
I did a few laps with and without taking a portal and the avg came to 1220 so that's what I went with. If its believed it should always account for taking a portal I can change that.


Alexsuperfly:
Expanded the scope to also include Crystal Trees in farming

Issue #9459, opened 2019-07-26
Song of the Elves changes


Alexsuperfly:
### The location of some/all the elves have changed to Prifddinas in post quest state
- [x] Eluned - southwest Prifddinas by the dye trader (hard cipher clue) #10887
- [x] Oronwen still in Lletya (anagram clue)
- [x] Ysgawyn - dead (master anagram clue) - possibly replaced by "RUE GO" anagram
- [ ] Lord Iorwerth - SSW Prifddinas by the teleporter (hard cryptic clue) - apparently this is a different Lord Iorwerth, meaning the clue either hasnt changed or is removed from the possible steps pool since the real Lord Iorwerth perished in the quest
- [x] Bow near lord Iorwerth (master emote clue) changed to "Bow in the Iorwerth Camp." #9471

### New clue steps #9471
- [x] medium anagram "HICK JET" - Jethick in West Ardougne answer: 38 #9455 
- [x] elite cryptic "Has no one told you it is rude to ask a lady her age?" - Lady Tangwen Trahaearn SSE Prifddinas by the teleporter
- [x] master emote "Beckon by a collection of crystalline maple trees. Beware of double agents! Equip Bryophyta's staff and a nature tiara."
- [x] master cryptic "Elvish onions." dig onion patch east of Prifddinas allotment patches
- [x] master anagram "RUE GO" - Goreu in Lletya
- [x] hard cryptic "A general who sets a 'shining' example." Speak to General Hining found at Tyras Camp. #9613
	
### New Timers
- [x] Divine potions #9493
- [x] Zalcano boss #9469 
- [x] Farming patches #9465 
- [x] gauntlet pb tracking #9501
- [x] agility course pb and kc #9492 

### New World map icons #9472 
- [x] SotE quest start
- [x] Rare trees
- [x] Transportation
- [x] agility shortcuts
- [x] update elf camp scroll to Iorwerth camp scroll
- [x] farming patches
- [x] teleport crystal
- [x] minigame

### New overlays
- [x] Prifddinas agility course obstacles #9466 
- [x] Prifddinas mining rocks #9463
- [x] Crystal Implings #9478 

### New loot tracking
- [x] Gauntlet chest #9462 
- [x] Prifddinas crystal chest #9470

### Others
- [x] Prifddinas agility course skill calc addition #9473
- [x] crystal tree skill calc #9473
- [x] Gauntlet consumable information #9461 
- [x] Gauntlet kc tracking #9480 
[...]

PR #9478, opened 2019-07-27
Implings plugin: Add Crystal implings

Dava96:
Should be added to #9459


Adam-:
These look like they are a lower level than dragon imps

PR #9480, opened 2019-07-27
Gauntlet: Add !kc command

rbbi:
Support for regular & corrupted Gauntlet killcounts, as required by #9459

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -83,7 +83,7 @@ import org.apache.commons.text.WordUtils;
 @Slf4j
 public class ChatCommandsPlugin extends Plugin
 {
-	private static final Pattern KILLCOUNT_PATTERN = Pattern.compile("Your (.+) (?:kill|harvest|lap) count is: <col=ff0000>(\\d+)</col>");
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -1346,6 +1356,16 @@ private static String longBossName(String boss)
 			case "raids 2":
 				return "Theatre of Blood";
 
+			// The Gauntlet
+			case "gaunt":
+			case "gauntlet":
+				return "Gauntlet";
+
+			// Corrupted Gauntlet
+			case "cgaunt":
```
rbbi: If anyone has any better suggestions for these, please shout :+1:

mfboulos: I would suggest adding variants of "Crystalline Hunllef" since gauntlet kc and hunllef kc are synonymous. As for variations of Corrupted Gauntlet, I'd wait a bit to see what the community calls it. Some call it hard mode, challenge mode, something will come from that I guess.

Adam-:
I changed this to just add `completion` to the existing kc regex.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
Adam-: `hunllef` and `chunllef` maybe? I feel like Hunllef is harder to spell/remember than gauntlet.

PR #9482, opened 2019-07-27
Fix WorldMapOverlay sitting over clue helper

Poppro:
Before:
Screen Shot 2019-07-26 at 8 47 56 PM

After:
Screen Shot 2019-07-26 at 8 45 33 PM


deathbeam:
Im pretty sure this is not the way to go. I dont think in this case this "issue" can be fixed

Poppro:
What issues do you perceive being created by the proposed priority reordering?

Adam-:
Why is the clue scroll overlay rendering above the world map at all? Currently it should render UNDER_WIDGETS which should be under the map. I think this makes sense.

deathbeam:
Because it moves layers when not snapped like every other overlay @Adam- 

This "priority reordering" will make clue scroll overlay never render under widget, and also not snap with other overlays we have defined this way @Poppro

Adam-:
Maybe we could move the non-snapped widgets to below-map then (is that a thing?) instead of above map? Do we have any reason why we would ever want to draw anything over the map that isn't part of the map?

deathbeam:
Well, we dont have "below-map" layer atm, but I guess we can add it, somehow, somewhere? As long as it will still be above other interfaces

Adam-:
Isn't that what `ABOVE_WIDGETS` is?

deathbeam:
Oh, is that below map? then i guess that works

Poppro:
Looks like ABOVE_WIDGETS is rendering above the map in this case. BELOW_WIDGETS renders under the map but also renders under the clue scroll interface which could be perceived as undesirable

abextm:
We cannot make a create a `below the map but above the chat/minimap/inventory area` layer because the world map is below all of those in the TLI, unless we wanted to:
 - Start doing weird hacks with the TLI component trees, which sounds fragile
 - or draw the overlays to a bitmap, and conditionally blit that at separate times depending on the state and location of the world map
 - or draw the overlays at a different time depending on their location, which is not possible for non layouted overlays, and would be very odd when crossing the boundary between these two zones with the map open

abextm:
Another option would be to just put a clip mask on the `ABOVE_WIDGETS` layer that excludes the world map, should it be open. And a big warning in `ABOVE_WIDGETS`'s  javadoc about how its not above all widgets as it's name implies

deathbeam:
I think the simple solution to actual issue here is to simply lower world map overlay priority to lowest so the icons will be below other overlays that are moved to above_widgets layer

Poppro:
The problem with the above suggestion is that WorldMapOverlay is assigned to ABOVE_MAP layer, which takes precedence over all ABOVE_WIDGET overlays. It seems to have been assigned ABOVE_MAP because the same overlay is used for both the map widget and the full screen map. My latest pull request addresses the issue by dynamically changing the WorldMapOverlay’s assigned layer based on whether full screen map is enabled

Adam-:
Even if the icons are `lowest` it is still above the highest `ABOVE_WIDGETS`. What about just putting a flag onto `Overlay` for whether or not it should draw when the fullscreen map is open, and then get rid of the `ABOVE_MAP` layer and move everything to ABOVE_WIDGETS` and then assign them low priorities?

Poppro:
@Adam- I think your suggestion is reasonable. Seems like maybe less of a hack than dynamically changing layers during run time. I guess this pull request isn't really focused on the clue helper anymore but rather just drawing on the map in general.

Adam-:
There are a couple more overlays which are also under widgets + low:

corp damage
raids
npc aggro
nmz
agility lap counter
barrows
team capes

Do we care about these? I think they would exhibit the same issue of drawing over the map when they are not in a snap corner.

Adam-:
```
[2:05 PM] abex: Is there a reason you can't just render the above_map layer below the above_widgets one?
[2:07 PM] abex: And not have to refractor everything that uses it and introduce 2 more foot guns into Overlay
```

Abex just mentioned this on Discord which actually makes sense. That would be ~1 line. Is there a downside to that?

deathbeam:
Well only difference between above map and above widget layer is that one has fullscreen check and one doesnt, so not really, and I think it makes most sense as well

Poppro:
Hmm... yea that is a pretty simple & effective solution. The only overlay (that i can identify) that would need a layer refactor is DevToolsOverlay, which might be switchable to ALWAYS_ON_TOP. It's worth noting that the only real purpose for the ABOVE_MAP layer with this change would be to draw directly on the map. As long as that's understood by devs I don't see a problem with Abex's suggestion

Adam-:
Why does the DevToolsOverlay need to move? Because it doesn't draw over the map?

Poppro:
@Adam- Yea, DevToolsOverlay never draws over the map (at least that I can identify?). I put the overlay into ABOVE_WIDGETS with priority HIGHEST to preserve the original functionality of the plugin (was in ABOVE_MAP, rendering above everything in ABOVE_WIDGETS).

PR #9487, opened 2019-07-27
Fix examine coins extra line

### LINKED ISSUE #9483: Examine Coins Extra Line
**Describe the bug**
When examining coins, the amount is listed in the chat box. An additional line item is added containing the exact same information as well as redundant information (GE Price, HA value). When the item examined are coins, the message should not be populated.

**To Reproduce**
1. Have a stack of coins in your inventory with Price enabled in Chat Commands
2. Right click on coins and select examine
3. Witness result in chat window

**Expected behavior**
Ideally, only 1 message is generated and the GE Average and HA value fields are not populated

**Screenshots**
image

 - OS: Windows
 - RuneLite version: 1.5.29.1
 - Launcher version: 1.6.0

dekvall:
Removed the GE and HA price message when examining coins since it's redundant information.
coins_correct

Closes #9483

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
@@ -37,6 +37,7 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.ItemComposition;
+import net.runelite.api.ItemID;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.MenuOptionClicked;
@@ -201,6 +202,12 @@ public class ExaminePlugin extends Plugin
 		{
 			final int itemId = pendingExamine.getId();
 			final int itemQuantity = pendingExamine.getQuantity();
+
+			if (itemId == ItemID.COINS_995)
+			{
+				return;
+			}
+
 			itemComposition = itemManager.getItemComposition(itemId);
 
 			if (itemComposition != null)


runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -203,7 +204,7 @@ public void onChatMessage(ChatMessage event)
 			final int itemQuantity = pendingExamine.getQuantity();
 			itemComposition = itemManager.getItemComposition(itemId);
 
-			if (itemComposition != null)
+			if (itemComposition != null && itemId != ItemID.COINS_995)
```
deathbeam: This should be moved below `endingExamine pendingExamine = pending.pop();`, just check id and type there and do early return

dekvall:
The reason I didn't do an early return is because the only issue I wanted to solve was that the GE and HA prices were shown for coins on examine. There is some logic below this where coins are submitted to the `examineClient`in two cases:
 - If the amount of coins is `< 100k` the examine text shows `Lovely money!`
 - If the coins are examined from the bank interface commas are added to the amount. `6192812 x Coins` becomes `6,192,812 x Coins.`

Since I am not too familiar with the `examineClient` and the underlying API I resorted to just make sure that no GE and HA values were shown for that item as the ExamineAPI felt outside the scope of the issue. 

Is this an existing oversight and I don't have to worry about using an early return for coins?

runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -195,6 +196,11 @@ public void onChatMessage(ChatMessage event)
 
 		log.debug("Got examine for {} {}: {}", pendingExamine.getType(), pendingExamine.getId(), event.getMessage());
 
+		if ((type == ExamineType.ITEM || type == ExamineType.ITEM_BANK_EQ) && pendingExamine.getId() == ItemID.COINS_995)
```
Adam-: This should just be down around ~210

dekvall: Isn't that the opposite of what db said? But i think it fits better there as well :)

dekvall:
Moved early return to L208.

dekvall:
Squashed.

PR #9489, opened 2019-07-27
Make Gauntlet timer and new boss healthbars layoutable

Hydrox6:
Makes the Gauntlet timer and the new healthbars from SotE, Gauntlet, and Zalcano layoutable

Up for suggestions on the naming, as I wasn't too sure what to call the healthbar


Bram91:
Zalcano healthbar goes invisible if its moved below the height of the chatbox in resizable

image


might want to limit the movement like is done with in-game xp tracker

Hydrox6:
We don't actually limit where widgets can move, as they are limited by their parents. It seems the parent of this widget is vastly bigger than the viewport, and I don't think there's much that can be done about that without making this PR an overhaul of how the system works

PR #9490, opened 2019-07-27
Fixed null pointer exception in SlayerPlugin.java

AttackOfTheMoons:


AttackOfTheMoons:
> 
> 
> LGTM. `Task.getTask()` should be marked `@Nullable` so we don't run into this again.

added

PR #9491, opened 2019-07-27
Bank Plugin - Add option to enlarge bank pin numbers

TheStonedTurtle:
Title. Config default is off.

qHAKXmd0HU


deathbeam:
I don't think we should mess with bank pin interface

PR #9494, opened 2019-07-28
Swap pickpocket with talk-to on any NPC

Krikke93:
Due to the recent addition to the pickpocket table of Elves (crystal shards + eternal teleport crystal), I felt there might be a need for a left-click pickpocket option, much like H.A.M. members. 
Because of the similarity, I renamed "Pickpocket on H.A.M." to simply "Pickpocket" and modified the behavior to function on any NPC, rather than H.A.M. members alone.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -194,8 +194,8 @@ public interface MenuEntrySwapperConfig extends Config
 
 	@ConfigItem(
 		keyName = "swapPickpocket",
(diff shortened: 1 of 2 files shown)


Adam-:
We have pretty obviously done this intentionally as there is the explicit ham check in there -- it is to keep parity with OSB. I don't think we should be changing that at this time.

Krikke93:
> We have pretty obviously done this intentionally as there is the explicit ham check in there -- it is to keep parity with OSB. I don't think we should be changing that at this time.

I see, excuses for not knowing this was a thing. May I ask why keeping parity with OSB is necessary in this case? Or maybe you could link me to a post explaining?

Adam-:
I am not comfortable pushing the thieving swapping further than what has been already established.

deathbeam:
This was rejected in past, so I have to close this, sorry

Issue #5175, opened 2018-08-28
NMZ plugin addons


itsSimplex:
Wondering if there can be an addon that shows the time till the next powerup and the time till the powerup ends once activated

fairyfoam48:
Do you know how the power ups are programmed? They almost seem random with regards to when they appear. A timer feature that lets you know how long they take until they end would be helpful but also another timer that tells you how long until the power up disappears (if you haven't activated it) would be very useful.

PR #9499, opened 2019-07-28
nmz: Add powerup timers and add an option to show powerup objects on minimap

### LINKED ISSUE #2278: NMZ Plugin: show power up names on minimap and highlight their tiles
Show the spawned power up names on the minimap and mark their tile, making it easier for AFK people to find the location of the spawned power up.

adwitkow:
Closes #2278
Addresses #5175 (only half of the issue since the spawns seem to be 100% random, from timings it seems that it can take somewhere from 30 seconds to 10 minutes between one spawn and another)

The overlay logic is based on the existing notification config items as I felt that adding a separate config would be redundant. If someone does not care about a particular powerup (ultimate force I'm looking at you), they disable the notification and thus the powerup won't be shown in the overlay or on the minimap.

Red: Spawned powerup, hasn't been taken by player yet
Green: Active powerup

nmz-overlay
nmz-overlay-with-minimap

I hope I didn't mess up the commits, I had some issues with branching that I had to force-push out. In the screenshots you can see the point counters from my other PR, they should not have been there. My bad.


deathbeam:
Shouldn't the powerup timers be just infoboxes like every other timer?

adwitkow:
I was considering that, but I could not find any nice RS sprites for it. I am not a good artist, that is for sure.

adwitkow:
Pushed the changes
Initial comment updated

adwitkow:
Rebased to stay up-to-date, added different colors to minimap rectangles and updated the PR description. Also, I did a little housekeeping in the commits, hopefully it looks good.

PR #3116, opened 2018-05-22
Animate Config Panel Toggle Buttons

RubenG123:
Adds animations to the toggle switches for the plugins in the config panel.

This was achieved by removing the call to rebuild the panel whenever a toggle switch was clicked, which I didn't seem to find a reason for. Though correct me if I'm wrong! And feel free to provide any feedback or point out any improvements.

EDIT: credit to @psikoi for the animation frame drawings

PR #9500, opened 2019-07-28
[WIP] Animate the toggle switches for turning on/off plugins.

RubenG123:
Changes the toggle switches in the plugin list to animate to their new state when clicked. Also cleaned up the plugin list item code a bit as there were some redundant calls in a few places.

I removed the event handler that refreshes the plugin list to enable this to work. It doesn't really seem to be nesesarry to refresh the whole thing when one plugin gets changed. But let me know if you know otherwise!

Credit to @psikoi for the animation frame drawings which he gave me a very long time ago

-----------

Marked as WIP as I need to make a fairly large refactor with the implementation


Adam-:
see also #3116

RubenG123:
Force pushes were to fix formatting to pass CI

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPlugin.java
```diff
@@ -103,12 +102,6 @@ protected void shutDown() throws Exception
 		clientToolbar.removeNavigation(navButton);
 	}
 
-	@Subscribe
-	public void onPluginChanged(PluginChanged event)
```
deathbeam: Why was this removed?

RubenG123: I removed the event handler as it was calling Configpanel::refreshPluginList() whenever a single plugin was either enabled or disabled. (This is also the only time its called outside of the ConfigPanel ctor)

```java
void refreshPluginList()
{
	// update enabled / disabled status of all items
	pluginList.forEach(listItem ->
	{
		final Plugin plugin = listItem.getPlugin();
		if (plugin != null)
		{
			listItem.setPluginEnabled(pluginManager.isPluginEnabled(plugin));
		}
	});

	if (showingPluginList)
	{
		openConfigList();
	}
}
```

As you can see its just rebuilding the whole plugin list. Unless this is some hacky fix for an unknown issue, it seems redundant to me. And rebuilding the plugin list prevents me from performing my animations. 

I guess if you wanted to keep this, I could add a parameter to the event that lets the refreshPluginList() method know to fire an animation for the updated plugin?

Adam-:
@psikoi can you review this?

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPlugin.java
deathbeam: It is here so the panel will update when plugin is toggled outside of config panel (so, for example GPU plugin, or logging in/out of rl account)

RubenG123: @deathbeam Oh right, thanks! I will definitely need to change a lot of my impl then

RubenG123: @deathbeam I can't seem to find any code that toggles the state of plugins outside of actually clicking the button itself, can you point me in the right direction? I also think I may have found a bug with what you are describing

RubenG123:
On second thought, I don't think this is worth doing so I'll just close this for good. Some people may even prefer it to not even be animated.

Issue #8454, opened 2019-04-09
!pb doesn't pick up some boss timers


MagicfTail:
This is known to affect:

- [ ] Theatre of Blood completion time
- [x] Chambers of Xeric completion time
- [x] Fight Caves (TzTok-Jad)
- [x] Inferno (TzKal-Zuk)
- [x] Grotesque Guardians

Adam-:
This is not specific to tob, but probably to any activity that has the pb time in a separate message from the kc.

Alexsuperfly:
Inferno and Fight Caves also fall under this


![](https://cdn.discordapp.com/attachments/361679294679744524/566191511491051520/bb89ef0e99bc6e7d5e362c539b1b7fb4.png)

![](https://i.imgur.com/tHfpsZ9.png)

Alexsuperfly:
Grotesque Guardians seems to have the same problem but its even more annoying to account for since the duration is sent before the boss kill.

image provided by someone else in the discord
![](https://image.prntscr.com/image/HBr_6AeLTOexBb6jbZskiw.png)

Nightfirecat:
Renamed issue and updated original post to reflect that this isn't a ToB-specific issue.

Alexsuperfly:
if we're going to do ToB might as well put CoX in there 

![](https://cdn.discordapp.com/attachments/361679294679744524/566733979282309130/d165036a80ab6fb3a3b7f82668342af3.png)

Adam-:
Here is the other zuk message: https://cdn.discordapp.com/attachments/301497432909414422/585270372761337856/unknown.png

Alexsuperfly:
![image](https://user-images.githubusercontent.com/7191512/62170822-c77df680-b2fa-11e9-87bb-311cbc164567.png)

Adam-:
Zuk/Jad fix is in 1ce2e5aa930d122d1c543c497f9990b2cde8b235 and gg in 1b2219374ff5dbc5c3380ff596779a06a34e7040

PR #9501, opened 2019-07-29
Gauntlet !pb

mfboulos:
Addresses !pb for the gauntlet, as requested in #9459.

I modified existing PB and kill duration patterns, I don't know of any false positives that would incur so feel free to strike that one down if necessary.

I haven't actually tested this because the only account I have that's able to do so is a hardcore, and I don't want to risk the client crashing or anything like that while I'm in the gauntlet :)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -87,8 +87,9 @@ public class ChatCommandsPlugin extends Plugin
 	private static final Pattern RAIDS_PATTERN = Pattern.compile("Your completed (.+) count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern WINTERTODT_PATTERN = Pattern.compile("Your subdued Wintertodt count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern BARROWS_PATTERN = Pattern.compile("Your Barrows chest count is: <col=ff0000>(\\d+)</col>");
-	private static final Pattern KILL_DURATION_PATTERN = Pattern.compile("(?:Fight|Lap) duration: <col=ff0000>[0-9:]+</col>. Personal best: ([0-9:]+)");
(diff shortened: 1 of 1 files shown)


Bram91:
Just tested it in the regular gauntlet, working as expected for that. Wont be able to try corrupted since I still struggle with that myself.

Adam-:
This should probably include tests too. Does anyone have a screenshot of the completion messages?

SebastiaanVanspauwen:
Regular gauntlet:
image
Corrupted gauntlet:
image

Adam-:
I'm not sure if it is safe to assume eg `Challenge duration: 10:24. Personal best: 7:59` is always the Gauntlet like this is doing?

mfboulos:
The Gauntlet follows the same sort of patterns as the kill duration and PB patterns. I asked a user to help me with the corrupted gauntlet completion pattern after a completion and he sent me this:

image

Which is why I handled it by generalizing those patterns to better suit how they're used (if and only if `lastBossKill` is non-null) and simply using the variation in patterns between normal and corrupt gauntlets to populate `lastBossKill` with the respective string name.

Adam-:
What about https://github.com/runelite/runelite/compare/master...Adam-:gauntpb ? It is slightly less elegant but should be more robust.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -290,6 +291,13 @@ else if (result.equals("were defeated"))
 
 			setKc("Barrows Chests", kc);
 		}
+		
+		matcher = GAUNTLET_PATTERN.matcher(message);
+		// TODO: set lastBossKill in kc command instead
```
Adam-: what is this referring to?

mfboulos: I had a hunch that the messages were being sent as one, and that I could do that. Meant the kc message, but either way it's not relevant anymore. Must have left that in on accident, apologies.

Adam-:
The reason this doesn't work is the same issue as #8454 where it doesn't work with gg, since they also send an out of order pb message. So I've fixed that and then fixed this in e27459f1f0a8fd06ba685baea258bd210d5e21fa

PR #9503, opened 2019-07-29
customcursor: option for a larger version for resolutions above 1600x900.

AlexanderVansp:
Option to toggle a larger version of the custom cursors.
Doubles in size for resolutions larger than 1600x900.
(ignore color)

Before:
Untitled 2

After:
Untitled

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorConfig.java
@@ -32,6 +32,7 @@ import net.runelite.client.config.ConfigItem;
 public interface CustomCursorConfig extends Config
 {
 	@ConfigItem(
+		position = 0,
 		keyName = "cursorStyle",
 		name = "Cursor",
 		description = "Select which cursor you wish to use"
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorConfig.java
```diff
@@ -40,4 +41,15 @@ default CustomCursor selectedCursor()
 	{
 		return CustomCursor.RS3_GOLD;
 	}
+
+	@ConfigItem(
+		position = 1,
+		keyName = "bigScreenStyle",
+		name = "Big screen mode",
+		description = "Enlarges the cursor when having a screen resolution of 1920x1080 or above."
```
Darkkgreen: 1600x900, not 1080p

runelite-client/src/main/java/net/runelite/client/plugins/customcursor/CustomCursorConfig.java
```diff
@@ -40,4 +41,15 @@ default CustomCursor selectedCursor()
 	{
 		return CustomCursor.RS3_GOLD;
 	}
+
+	@ConfigItem(
+		position = 1,
+		keyName = "bigScreenStyle",
+		name = "Big screen mode",
```
Nightfirecat: I think this feature would be much better received if it were just a simple "Large cursor" toggle, which didn't depend on screen size. I can imagine some players wanting the cursor to be double-sized even at smaller resolutions as a matter of personal preference. In addition, I don't really see a benefit to restricting it to larger screens in this way.

Adam-:
This seems like the type of thing we could be able to detect automatically instead of requiring the user to specify they have a large resolution screen. Is that possible?

deathbeam:
My system cursor is properly scaled, but with custom cursors, it is suddenly tiny. As @Adam- said, this should be done automatically (or well, this, I would just prefer to scale cursors somehow based on resolution/DPI). Regardless, toggle for this sounds like lazy solution that works only for author of the issue (as obviously, people have different resolutions).

PR #9511, opened 2019-07-29
Sote new tool animations

Alexsuperfly:


deathbeam:
Your last commit has typo (crytal axe instead of crystal axe)

PR #9512, opened 2019-07-29
Add infernal cape and infernal max cape to master clue scroll step

### LINKED ISSUE #9510: Emote master clue item helper - tzhaar gem store
**Describe the bug**
Clue step: Cry in the Tzhaar gem store. Beware of the double agents! Equip a fire cape and a TokTz-Xil-Ul.

Regular infernal cape, fire max cape, and infernal max cape also count for this step, so the item checker should also show a checkmark if you have any of those in your inventory.

Darkkgreen:
Just added Infernal cape and Infernal max cape to the step item check.

Fixes #9510


deathbeam:
Oops, do not noticed it has incorrect wording for commit when merging. Welp, it's fine

PR #9528, opened 2019-08-01
bank tags: properly open saved tab

### LINKED ISSUE #9401: Bank tags sometimes does not remember last tag tab
**Describe the bug**
The bank interface is not remembering which tag tab I was last on. It just resets to the vanilla normal tab whenever I press esc or click the close button in the corner. Restarting the client seems to temporarily fix it, but it sporadically breaks again and multiple users are having this issue after the recent update.

**To Reproduce**
1. Open Bank
2. Select bank tag tab
3. Close Bank
4. Bank tag tab is not remembered

**Expected behavior**
Closing the bank interface while viewing a tag tab should display that tab when opening the interface again

**Screenshots**

 - OS: Windows 10
 - RuneLite version: 1.5.29
 - Launcher version: 1.6.2

**Additional context**
Logs didn't seem to have information related but I will include them anyways
[client.log](https://github.com/runelite/runelite/files/3409174/client.log)

trimbe:
Depends on !75

Jagex appears to be syncing the current tab varbit more aggressively now, so if the active tab was not the first tab before switching to a tag tab, bank layout would misbehave when the bank was reopened.

Closes #9401

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
@@ -327,7 +327,7 @@ public class TabInterface
 		switch (event.getOp())
 		{
 			case Tab.OPEN_TAG:
-				client.setVarbitValue(client.getVarps(), Varbits.CURRENT_BANK_TAB.getId(), 0);
+				client.setVarbit(Varbits.CURRENT_BANK_TAB, 0);
 				Widget clicked = event.getSource();
 
 				TagTab tab = tabManager.find(Text.removeTags(clicked.getName()));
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -215,6 +215,8 @@ public void init()
 
 		if (config.rememberTab() && !Strings.isNullOrEmpty(config.tab()))
 		{
+			// the server will resync the last opened vanilla tab when the bank is opened
+			client.setVarbitValue(client.getVarps(), Varbits.CURRENT_BANK_TAB.getId(), 0);
```
Adam-: this is probably just `setSetting`, though both are marked as devtools only...

raiyni:
This code doesn't fix the issue for me. It appears the bank closes and opens on the same tick. Which was observed through logging. So the isHidden check doesn't get called in the update method.

PR #9544, opened 2019-08-01
Update prif agility course

Alexsuperfly:
made the calc this time just always account for taking the portal, since thats what people will be wanting to do anyways especially since they are highlighted now


Adam-:
is 1337 *always* the xp you get if you take the portal?

Alexsuperfly:
split the agility changes into two commits per your request in discord

PR #9547, opened 2019-08-01
Entityhider: add hide random events

weaman10:
should have an option to hide random events with the option to show Genie?


Darkkgreen:
What is even happening here.

weaman10:
wow just noticed my bad... can i even remove it?

PR #9557, opened 2019-08-03
Snapshot all grahics2d properties in safeRender

deathbeam:
This prevents overlay graphics configurations leaking to other overlays.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -27,10 +27,14 @@ package net.runelite.client.ui.overlay;
 import com.google.common.base.MoreObjects;
 import com.google.common.primitives.Ints;
 import java.awt.Color;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This PR does seem to fix the world map tooltip bug I noticed earlier.

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -462,11 +466,15 @@ else if (position == OverlayPosition.TOOLTIP)
...
 		}
 
-		// Reset the default color
-		graphics.setColor(Color.WHITE);
+		// Save graphics2d properties so we can restore them later
+		final AffineTransform transform = graphics.getTransform();
+		final Stroke stroke = graphics.getStroke();
+		final Composite composite = graphics.getComposite();
+		final Paint paint = graphics.getPaint();
+		final Color color = graphics.getColor();
```
Adam-: This should be able to work with only `Paint` since it s superclass of color

Issue #8631, opened 2019-04-24
Loot Tracker hiding items over zealously 


MatzeBro:
When you Hide 'Dragon Bones' 'Babydragon bones' are also hidden.

psikoi:
/loottracker/LootTrackerPlugin.java
At line 576

Change .contains to .equals, for anyone looking for a quick first contribution :P

pattyjogal:
I don't know how you were getting that behavior, @MatzeBro 

The flow of items being checked for being ignored looks like this:

```java
// Ignored items array is defined
private List<String> ignoredItems = new ArrayList<>();
...
// Parses the item names from the CSV
ignoredItems = Text.fromCSV(config.getIgnoredItems());
...
// isIgnored method
boolean isIgnored(String name)
{
	return ignoredItems.contains(name);
}
```
There should be no substring-ing going on here; `isIgnored` merely checks if the input name is in the `ignoredItems` array.

Nightfirecat:
I can confirm that I did get a loot tracker entry containing `Big bones` when `Bones` was on my ignored items, so I'm pretty sure this isn't an actual bug.

Nightfirecat:
As no further information was provided and nobody else has been able to reproduce this, I am closing this issue. If you have some way to reproduce this issue, please comment and re-open.

PR #9571, opened 2019-08-04
Change the loottracker hide feature to be less overzealous

pattyjogal:
Per #8631, the Loot Tracker plugin would hide items based off of substrings. Fixed to do an exact comparison.


Krysaczek:
I don't see anything wrong with Ignored items function as it is. #8631 is wrong, the search is case sensitive and because of that the 'Dragon Bones' string doesn't hide anything. The correct string 'Dragon bones' hides the exact match only and 'Babydragon bones' are still visible.

pattyjogal:
My mistake, I took the issue feedback as is, and didn't investigate further.

pattyjogal:
Yea, it seems like that's right. I don't know what the issue was referring to then, but seems without further explanation, this should be closed.

PR #9577, opened 2019-08-05
WIP: Allow searching by value in banks

Bram91:
Added searching by value in banks.

- `<` smaller than
- `>` larger than
- ` -` intervals
ie:
`value:<5m` will show all itemstacks below 5 million GP
`value:100k-1b` will show all itemstacks starting at 100k up to 1billion GP


runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
```diff
@@ -54,7 +57,7 @@
 	private final ConfigManager configManager;
 	private final ItemManager itemManager;
 	private final ClueScrollService clueScrollService;
-
+	private Item[] bankItems;
```
raiyni: Tag manager shouldn't be aware of the itemcontainer items.

Bram91: Relocated the itemcontainer items array to its own class.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankItemCache.java
```diff
@@ -0,0 +1,23 @@
+package net.runelite.client.plugins.banktags;
+
+import net.runelite.api.Item;
+
+public class BankItemCache {
```
deathbeam: Remove this class, and just inject client and call client.getItemContainer (or w/e it is named atm) in tag manager

Bram91:
Removed all the logic from tagmanager and placed it in BankTagsPlugin also added `HA:` search

PR #9578, opened 2019-08-05
Fix hiding attack styles when switching weapons

### LINKED ISSUE #9564: Combat styles are not hidden when switching between bludgeon and bow
https://twitter.com/piksubtw/status/1157692363948601345

dekvall:
Hiding attack styles primarily relies on a subscription to onWidgetHiddenChanged but when swapping between bludgeon to bow this event fires before onVarbitChanged which means that when swapping quickly between the two the longrange option will not be hidden.

There is also a NPE fix.

Closes #9564

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -105,11 +105,7 @@ public class AttackStylesPlugin extends Plugin
 
 	private void start()
 	{
-		updateWarnedSkills(config.warnForAttack(), Skill.ATTACK);
-		updateWarnedSkills(config.warnForStrength(), Skill.STRENGTH);
-		updateWarnedSkills(config.warnForDefence(), Skill.DEFENCE);
-		updateWarnedSkills(config.warnForRanged(), Skill.RANGED);
-		updateWarnedSkills(config.warnForMagic(), Skill.MAGIC);
+		resetWarnings();
 		attackStyleVarbit = client.getVar(VarPlayer.ATTACK_STYLE);
 		equippedWeaponTypeVarbit = client.getVar(Varbits.EQUIPPED_WEAPON_TYPE);
 		castingModeVarbit = client.getVar(Varbits.DEFENSIVE_CASTING_MODE);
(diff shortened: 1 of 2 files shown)


Adam-:
Okay. Why not call processWidgets from onVarbitChanged then? This looks like it will update the widgets 5x on login.

dekvall:
You are right, right now process processWidgets are already called 30 times on login, after this change it will be called 39 times. Moving it to onVarBitChanged calls it 5k+ times on login, and just just moving it from onWidgetHiddenChanged doesn't work as the widgets that needs to be hidden need to be open (i think). I didn't think the increase in calls would be that big of a deal compared to what was already happening. But i can look into a more thorough rewrite if you want, it would be nice to avoid the existing amount of calls.

dekvall:
Lowered the amount of processWidgets calls from 39 to 1 on login.

Adam-:
Why is it called 30 times on login? Also moving it to on varbit change would not call it 5k+ times unless the variable changed 5k times so that makes no sense

dekvall:
I messed up the onVarbitChange test, i just counted the number of times the event fired on login, not how many times the specific varbits changed. My bad.

It was originally called 30 times on login because it was fired for every widget rendered in the COMBAT_GROUP as their hidden status was changed to visible on render.

Adam-:
I would refactor onVarbitChanged to have one branch for any of the 3 vars changing

dekvall:
Refactored.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -163,6 +163,7 @@ public void onWidgetLoaded(WidgetLoaded event)
 		}
 
 		loadedCombatStyleWidget = true;
+		processWidgets();
```
Adam-: so... is this necessary or not?

dekvall: I didn't think it was necessary up until now, but if you use your f-keys to navigate to the combat tab the moment you log in you will see the disabled options for a split second before the varbits are set.

Adam-: Is the widget hidden change event still necessary if we are now processing widgets when the var changes?

dekvall: Yes, without it the entire plugin breaks. The attack options seem to be living a life of their own.

Adam-: In what cases can the visibility of the combat widgets change when the vars do not?

dekvall: In my experience, it looks as if the attack style widgets reset their visibility whenever you go to the combat tab. So you have to set it to hidden before they are properly loaded.

Adam-: It looks like there is a hidden even fired when the tab is selected and when it is deselected. `loadedCombatStyleWidget` will get reset whenever you teleport so that will break the plugin

dekvall: I seem to have misinterpreted that event. Thanks, I think it's fixed now.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -63,7 +64,7 @@
 public class AttackStylesPlugin extends Plugin
 {
 	private int attackStyleVarbit = -1;
-	private int equippedWeaponTypeVarbit = -1;
+	private int equippedWeaponTypeVarbit = 0; //The player is briefly considered unarmed during login
```
Adam-: What is the purpose of this?

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -181,41 +186,46 @@ private void processWidgets()
...
-			updateWarnedSkills(config.warnForMagic(), Skill.MAGIC);
+			case LOGGED_IN:
+				updateWarnedSkills(config.warnForAttack(), Skill.ATTACK);
+				updateWarnedSkills(config.warnForStrength(), Skill.STRENGTH);
+				updateWarnedSkills(config.warnForDefence(), Skill.DEFENCE);
+				updateWarnedSkills(config.warnForRanged(), Skill.RANGED);
+				updateWarnedSkills(config.warnForMagic(), Skill.MAGIC);
+				break;
+			case LOGIN_SCREEN:
+			case LOGIN_SCREEN_AUTHENTICATOR:
```
Adam-: This is unnecessary

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
dekvall: The purpose was to avoid an unnecessary weapon switch and thereby a call to processWidgets on login. Should i revert it?

Adam-: This probably isn't always 0 on login? If you change accounts etc it would be whatever it was on the last account

dekvall: The varbit is always set to 0 on login, then updated if you have something equipped. But yea, since i never reset these values when the player logs out they will be kept at their current values and the call will processWidget call will happen twice anyway.

I guess i can reset it on login screen along with widgetHasBeenLoaded to mitigate this.

dekvall:
Squashed to two commits.

I decided to remove the prevention of extra calls to processWidgets on login as it's more trouble than it's worth. As you said; the overhead is pretty small and fixing it makes the code more complex than required.

Adam-:
Sounds good

dekvall:
updated commit message.

PR #9585, opened 2019-08-05
Fix loot tracker order on client reload

### LINKED ISSUE #9575: LootTracker: Loot Drops From Server are in Reverse Order
**Describe the bug**
When reloading the client the drops that are loaded from prev sessions are in reverse order

**To Reproduce**
1. switch loot tracker to individual kill
2. Kill a mob
3. kill another mob
4. restart runelite
5. open loot tracker and see drops have been reversed

**Expected behavior**
the loot tracker should always have the same order of drops

**Screenshots**
Kill Order: 
rabbit -> guard -> guard

Loot Tracker before reloading runelite:
image

Loot Tracker after reloading runelite:
image

Loot Tracker after killing imp after reload:
image

Expected:
(top -> bottom)
imp -> guard -> guard -> rabbit

 - OS: Windows 10
 - RuneLite version: 1.5.30.1
 - Launcher version: 1.6.0

dekvall:
The panel is currently rebuilt with the earliest loots at the top.
For consistency's sake it should be the other way around.

Closes #9575

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
@@ -378,7 +378,7 @@ class LootTrackerPanel extends PluginPanel
 	void add(final String eventName, final int actorLevel, LootTrackerItem[] items)
 	{
 		final String subTitle = actorLevel > -1 ? "(lvl-" + actorLevel + ")" : "";
-		final LootTrackerRecord record = new LootTrackerRecord(eventName, subTitle, items, System.currentTimeMillis());
+		final LootTrackerRecord record = new LootTrackerRecord(eventName, subTitle, items);
 		records.add(record);
 		LootTrackerBox box = buildBox(record);
 		if (box != null)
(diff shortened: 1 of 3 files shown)


Adam-:
`LootRecord` has a time, maybe it would be better to just do a sort by that?

dekvall:
It's sorted by time in the when selected from the database though, so the result would be the same.

Adam-:
Yeah but it would be less error prone to not rely on that. If we did, we could just reorder it in the query?

dekvall:
I guess I just don't see a case where the query for the last N monsters killed would not rely on time.

deathbeam:
I would just order this by time field on record like @Adam- said.

dekvall:
Alright, will do.

dekvall:
Updated to sort by timestamp instead of reversing

deathbeam:
I think we set the time to -1 intentionally before, but I dont remember why :/

Adam-:
It looks to me like `timestamp` isn't actually used? It was added in 4dd8322121e190cca16a22f0a865ecf721b36223 but I don't see it used there either..

Adam-:
Can we just sort the `LootRecord` instead and then remove this timestamp field?

dekvall:
Sorted the `LootRecord` entries and removed the timestamp field from `LootTrackerRecord`.

PR #9593, opened 2019-08-06
update Prifddinas agility course xp values to fix lap counter

4ilo:
Last game update changed the xp rewarded for the last obstacle.
Due to this change, the lap counter would not show on the Prifddinas agility course.

By changing the xp values this problem is solved.


xdesr:
duplicate #9544

Alexsuperfly:
I believe my pr is better suited because it accounts for taking a portal during the lap which grants more xp

dekvall:
If the xp difference is substantial maybe we should add both, like we have done for smithing for gold bars with/without goldsmith gauntlets.

PR #9599, opened 2019-08-06
prayer: display prayer time remaining in prayer orb

### LINKED ISSUE #9579: Prayer Timer
No. It's a matter of convenience.

For something that works like a timer for an antipoison. 
(icon/timer in a box located where other ones currently do if buffs exist, pic for example:)

Hover mouse over prayer orb next to minimap to see how much time I currently have left until prayer points reach 0 with whatever prayer is currently active.

**Additional context**
Only suggesting because I feel it'd be convenient to see a timer on the screen when a prayer activates, instead of hovering over the prayer orb next to the minimap. If you guys agree, cool; if not, oh well.

dekvall:
Adds the option to display the time for which you are able to keep your current prayers up instead of prayer points in the prayer orb. When the option is active the, prayer points are shown in the prayer tooltip instead of time.

prayer_plugin

Closes #9579


runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerPlugin.java
```diff
@@ -296,4 +308,50 @@ private void removeOverheadsIndicators()
...
+	private void setPrayerOrbText(String text)
+	{
+		Widget prayerOrbText = client.getWidget(WidgetInfo.MINIMAP_PRAYER_ORB_TEXT);
+		if (prayerOrbText != null)
+		{
+			prayerOrbText.setText(text);
+		}
+	}
+
+	private double getPrayerDrainRate(Client client)
```
Nightfirecat: This can be `static`
```suggestion
	private static double getPrayerDrainRate(Client client)
```

dekvall: Didn't think of that, thanks!

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerPlugin.java
```diff
@@ -304,4 +317,57 @@ private void removeOverheadsIndicators()
 		infoBoxManager.removeIf(entry -> entry instanceof PrayerCounter
 			&& ((PrayerCounter) entry).getPrayerType().isOverhead());
 	}
+
+	private void setPrayerOrbText(String text)
```
Nightfirecat: This should also be reset on plugin shutdown.

dekvall: I'm not sure i understand, the prayer orb text has to be changed every game tick to show time so when the plugin is turned off, it resumes to use prayer points. Setting it explicitly on shutdown feel unnecessary (?)

Nightfirecat: Does the client set that text each tick via some script? If so this is fine.

dekvall: Upon closer inspection it seems like script 82 sets it on prayer turn off or when you lose a prayer point, so I have used too draining prayers when testing.

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerPlugin.java
```diff
@@ -176,6 +184,11 @@ public void onGameTick(GameTick tick)
 			barOverlay.onTick();
 		}
 
+		if (config.replaceOrbText() && isAnyPrayerActive())
+		{
+			setPrayerOrbText(getEstimatedTimeRemaining(true));
+		}
```
Nightfirecat: Given your comment regarding the orb text updating after script 82 runs, maybe this should be moved to an `onPostScriptFired` subscriber?

dekvall: Yes that is probably a good idea

Nightfirecat: On second thought, this would cause the remaining time to not appear when toggling individual prayers until a point is drained, so that's not great. That requires this code being run each game tick, so this should be fine after all.

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
```diff
@@ -114,11 +112,19 @@ public Dimension render(Graphics2D graphics)
...
 		{
-			final String tooltip = "Time Remaining: " + getEstimatedTimeRemaining() +
-				"</br>" +
-				"Prayer Bonus: " + prayerBonus;
+			final StringBuilder sb = new StringBuilder();
 
-			tooltipManager.add(new Tooltip(tooltip));
+			if (config.replaceOrbText())
+			{
+				sb.append("Prayer points remaining: " + client.getBoostedSkillLevel(Skill.PRAYER));
```
Adam-:
```suggestion
				sb.append("Prayer points remaining: ".append(client.getBoostedSkillLevel(Skill.PRAYER));
```

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
```diff
@@ -114,11 +112,19 @@ public Dimension render(Graphics2D graphics)
...
+			}
+			else
+			{
+				sb.append("Time Remaining: " + plugin.getEstimatedTimeRemaining(false));
```
Adam-:
```suggestion
				sb.append("Time Remaining: ".append(plugin.getEstimatedTimeRemaining(false)));
```

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
```diff
@@ -114,11 +112,19 @@ public Dimension render(Graphics2D graphics)
...
+			}
+
+			sb.append("</br>" + "Prayer Bonus: " + plugin.getPrayerBonus());
```
Adam-: here too

runelite-client/src/main/java/net/runelite/client/plugins/prayer/PrayerDoseOverlay.java
Nightfirecat: Don't you mean `sb.append(str1).append(str2)`? Not `sb.append(str1.append(str2))`?

PR #9603, opened 2019-08-07
Default option for teleport crystal changed to prifddinas

Wazikamawata:
Thought it'd be nice to have a toggle for the teleport crystal now that prifddinas is accessible, and most people post-SOTE are probably only travelling to prif now


Alexsuperfly:
this is already possible with shift click imo, just adds more useless clutter to that plugin

deathbeam:
This can already be done with shift-click, so closing, sorry

PR #9604, opened 2019-08-07
Reset skill states when starting XpTracker

### LINKED ISSUE #9089: XP Tracker panel doesn't show overall XP if XP Tracker plugin is toggled off and on again
**Describe the bug**
If you toggle the XP Tracker plugin off and on again the overall XP counter will fail to show up and the error panel that reads "You have not gained experience yet." doesn't get removed.

It comes back if you either:
1. Return to the login screen, log back in and then gain some xp in a skill that is already in the XP Tracker panel.
2. Gain construction XP twice.

**To Reproduce**
1. Having enabled the XP Tracker plugin, gain some XP so that the skill shows up in the XP Tracker panel.
2. Toggle the XP Tracker plugin off and on again which will display an empty XP Tracker panel with just the error message "You have not gained experience yet."
3. Gain XP twice so that the XP Tracker panel starts showing a skill.
4. Observe no overall XP and the error message remains.

**Expected behavior**
The overall XP counter should show up and the error message be removed.

**Screenshots**
What the XP tracker should look like:
image

How it looks after it breaks:
image

 - OS: Windows 10
 - RuneLite version: 1.5.26
 - Launcher version: 1.6.0

**Additional context**
I don't really know the best way to fix this, but the issue seems to be with this section of code:

https://github.com/runelite/runelite/blob/ad16bf5b3f55223d775936ae9b20ffb421c1a4d1/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java#L333-L344

Line 338 is the only time overall XP is initialised, but this is based on the assumption that you're logging into the game, which apparently happens when construction XP is updated. To test this out I tried gaining some construction XP and it turns out that if you gain some construction XP twice then overall XP will get initialised properly again without needing to return to the login screen.

dekvall:
When resetting the XpState on startup I was unable to reproduce the issue and everything works as expected.

Closes #9089

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
@@ -142,6 +142,11 @@ public class XpTrackerPlugin extends Plugin
 	{
 		xpPanel = new XpPanel(this, xpTrackerConfig, client, skillIconManager);
 
+		for (Skill skill : Skill.values())
+		{
+			resetSkillState(skill);
+		}
+
 		final BufferedImage icon = ImageUtil.getResourceStreamFromClass(getClass(), "/skill_icons/overall.png");
 
 		navButton = NavigationButton.builder()


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -142,6 +142,11 @@ protected void startUp() throws Exception
 	{
 		xpPanel = new XpPanel(this, xpTrackerConfig, client, skillIconManager);
 
+		for (Skill skill : Skill.values())
```
Adam-: this should just be `xpState.reset()`

dekvall: `xpState.reset()` doesn't initialize the skills to the current xp and the issue persists.

Adam-: What happens if you toggle the plugin while logged out?

dekvall: Nothing special from what I can tell.

Adam-: I believe it doesn't zero the skill xps out if you logout. Is that bad?

dekvall: No, because you might want to continue your skilling session after a break.

dekvall:
Adam made a cleaner fix for this.

PR #9607, opened 2019-08-07
LootTracker: Add Kingdom of Miscellania

dabolink:
saw some inventory checks have region checks but not 100% sure as to why. I got the region id for managing miscellania (10044) and can add if needed.

image


runelite-api/src/main/java/net/runelite/api/InventoryID.java
```diff
@@ -53,6 +53,10 @@
 	 * Monkey madness puzzle box inventory.
 	 */
 	MONKEY_MADNESS_PUZZLE_BOX(221),
+	/**
+	 * Kingdom Of Miscellania reward inventory
```
daelynj: Let's add a period to stay consistent.

PR #9610, opened 2019-08-07
bank plugin: improve responsiveness of bank searches

### LINKED ISSUE #9569: Potentially faster search in bank by calling 'relayout' earlier
The bank tags open slightly faster than the search results for a global bank search. If it were possible to call the relayout function earlier, this might slightly increase the speed at which bank searches return results. It might also degrade performance, experimentation is therefore required. 

Quote from trimble in Discord: "iirc the bank does a layout every tick when a search is active. i think we could prevent that and just layout when the input string/itemcontainer changes and actually improve the performance?" 

Could be implemented together with or related to #3994

trimbe:
The client will normally layout the bank during a search only once every 40 client ticks, causing slow response times after the search input is updated. Instead, hook the search refresh script that is called every client tick, and if the bank hasn't been laid out on the current tick, and the search input has changed, lay it out early.

Closes #9569

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
@@ -45,11 +45,14 @@ import net.runelite.api.ItemComposition;
 import net.runelite.api.ItemContainer;
 import net.runelite.api.ItemID;
 import net.runelite.api.MenuEntry;
+import net.runelite.api.ScriptID;
+import net.runelite.api.VarClientStr;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.MenuShouldLeftClick;
 import net.runelite.api.events.ScriptCallbackEvent;
+import net.runelite.api.events.ScriptPostFired;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetID;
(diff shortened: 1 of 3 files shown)


dekvall:
This is one of those things that makes you wonder how you ever could go back.

notjuanortiz:
Bumping for a great PR.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -176,6 +184,35 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+			{
+				searchButtonBackground.setOnTimerListener((Object[]) null);
+				searchButtonBackground.setHasListener(false);
+			}
+
+			clientThread.invokeLater(() -> bankSearch.layoutBank());
+			searchString = searchVar;
+		}
+
+		if (client.getVar(VarClientInt.INPUT_TYPE) != InputType.SEARCH.getType() && Strings.isNullOrEmpty(client.getVar(VarClientStr.INPUT_TEXT)))
```
Nightfirecat: Why is a second call of `client.getVar(VarClientStr.INPUT_TEXT)` needed? `searchVar` should have that value already.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/BankSearch.java
```diff
@@ -77,11 +64,29 @@ public void search(InputType inputType, String search, boolean closeInput)
...
+
+		if (scriptArgs == null)
+		{
+			return;
+		}
+
+		client.runScript(scriptArgs);
+	}
+
+	public void reset(Boolean closeChat)
```
Nightfirecat, line 89: Why is this method signature changed to accept a boxed boolean?

trimbe: Oops, that slipped past in the rebase.

PR #9628, opened 2019-08-09
Med Clue: Light source item requirement

### LINKED ISSUE #9626: Change Medium Clue Guide
Medium emote clue step says lumbridge swamp and not under, doesn't tell people to bring a lightsource. (https://gyazo.com/6747a8533804a922c5d63a35830acb25)

Change the guide to "Below Lumbridge Swamp (Bring Light). The clue for baxtorian falls tells you to bring a rope so I see no reason as to why this couldn't say to bring a light.

Dava96:
closes #9626 

image

(I'm aware the picture isn't the correct clue but I was using it for testing)


Alexsuperfly:
theres a lot more sources of light than just that

Dava96:
> This both (a) doesn't clarify the location to be the Lumbridge swamp caves, and (b) doesn't really address the light source issue since it doesn't have an exhaustive listing of light sources. This would be a case where having a "Notes" section for emote clues would be useful. (I think there are other issues where such a section could have been helpful as well? We may want to use that as a more general solution to this issue)

If I was to add a note section what other example are there other than this one for it to be used? The double agent clues?

dekvall:
A notes section could include a shortened version of the field with the same name [here](https://oldschool.runescape.wiki/w/Treasure_Trails/Guide/Emote_clues). For example, you need a lockpick to do the emote in the magic axe hut or that you should turn Auto-retaliate off when doing the emote with Toktz-xil-ul to not throw it away etc.

Just go through the clues and add what seems useful :)

Adam-:
There are some other clues which require light sources. The first to come to mind is the ones where you have to go through the cave horrors.

Dava96:
I'll add it and I'll go through the list tomorrow, to see if there are any other clues that require a light source.

deathbeam:
I think just making like LightSourceRequirement helper class (i think we have similar one for spade) would be enough, and then people can use it in future

dekvall:
Or maybe even an ItemRequirement/ItemRecommendation helper for clues that requires other items not not explicitly required.

 - Sled for Trollweiss clue
 - Axe and machete for Kharazi jungle clue
 - Anti-dragon shild for KBD clue
 - Lockpick for the magix axe hut
 - Nosepeg for top of slayer tower clue
 - Ring of visibilty for shadow dungeon clue
 - Slashing weapon or knife for lava maze clue
 - Crossbow and grapple for observatory clue
 - Clue box for any clue in the wilderness

drahenshaw:
@Dava96, is this PR still being worked on? Do you need help in any way? I'm looking for a place to contribute. Thanks!

Issue #9549, opened 2019-08-02
[chatcommands] Add shorthands for Gauntlet and Corrupted Gauntlet kc


Jimbobiss:
Currently you have to write '!kc The Corrupted Gauntlet' to get a kc print out. This is rather heavy handed, so having some alternatives available would be useful such as '!kc tcg' or '!kc cgaunt'. You get the idea.

Would be great if possible, but if not I suppose no worries.

Adam-:
This is a bug. The row tracks it as `The Gauntlet` and we track it as `Gauntlet`

PR #9631, opened 2019-08-10
Fix Gauntlet and corrupted gauntlet abbreviations

Rdeneve:
Fixes Issue #9549


Adam-:
This doesn't fix the issue because the chatcommands plugin still adds the kcs without the leading `The`. We either need to change chatcommands to do this, or change the row scraper to remove it.

Adam-:
this was fixed in c7d89860ab34f2b1157543af12b25ad51239ac87

PR #9634, opened 2019-08-10
SkillCalc: Clear button does not reset the search bar

### LINKED ISSUE #9545: SkillCalc: Clear button does not reset the search results
clearbuttonfail

Bonsaigin:
Closes #9545 

search-bar-clear


Dava96:
You should probably add this to the Grand exchange search. Since they behave in the same way.

Bonsaigin:
> You should probably add this to the Grand exchange search. Since they behave in the same way.

I won't do that as part of this pull request.

The Grand Exchange plugin search is a little different, as the search isn't updated on each keystroke. 
 Maybe they shouldn't be different at all.  Feel free to open an issue for the GE plugin and the desired functionality can be discussed there.

raiyni:
It performs the same function.

PR #9641, opened 2019-08-11
notes: strip input of invisible control characters before saving

### LINKED ISSUE #9248: Symbols appearing in Notes plugin window in numerical values with commas
I thought this would've been bought up before but I wasn't able to find a bug report, open or closed on this. I could've missed it though, sorry for wasting your time if that's the case.

I get this bug a lot so thought I'd see about making a bug report.

Cheers guys. Yell at me if you need any more information, though I don't think there's much to tell, other than that I only play on one account.

dekvall:
When copying from windows calculator the characters U+202D (LEFT-TO-RIGHT OVERRIDE) and U+202C (POP DIRECTIONAL FORMATTING) are added to the beginning and the end of the desired content respectively. If they are copied into the notes plugin they will yield weird characters after a client reload when the notes are restored from a save.

This commit removes all invisible:
 - control characters 0x00-0x0F and 0x7F-0x9F
 - formatting indicators
 - code points reserved for private use
 - unassinged characters

Where number 2 in the list solves the case of windows calculator.
Invisible characters part of a surrogate pair in UTF-16 encoding are not removed due to keep emoji support for simple emojis in the plugin.

Closes #9248

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/notes/NotesPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/notes/NotesPanel.java
@@ -140,8 +140,9 @@ class NotesPanel extends PluginPanel
 				try
 				{
 					// get document text and save to config whenever editor is changed
-					String data = doc.getText(0, doc.getLength());
+					String data = doc.getText(0, doc.getLength()).replaceAll("[\\p{Cc}\\p{Cf}\\p{Co}\\p{Cn}]", "");
 					config.notesData(data);
+					setNotes(data);
 				}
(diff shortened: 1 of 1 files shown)


Adam-:
It would be better to use https://guava.dev/releases/19.0/api/docs/com/google/common/base/CharMatcher.html#INVISIBLE

dekvall:
`CharMatcher INVISIBLE` breaks surrogate pairs, hence it will break unicode smileys. I do not see any way to filter from the separate categories individually.

deathbeam:
I think removing unicode smileys is fine

PR #9647, opened 2019-08-12
Remove duplicates from tag icon search

### LINKED ISSUE #9644: Ensouled heads show up twice in the tag tabs icon search
image

dekvall:
Use `ItemVariationMapping` in the chatbox search to not show multiple versions of the same item.

Closes #9644

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java
@@ -44,6 +44,7 @@ import net.runelite.api.widgets.WidgetTextAlignment;
 import net.runelite.api.widgets.WidgetType;
 import net.runelite.client.callback.ClientThread;
 import net.runelite.client.game.ItemManager;
+import net.runelite.client.game.ItemVariationMapping;
 
 @Singleton
 public class ChatboxItemSearch extends ChatboxTextInput
@@ -294,8 +295,10 @@ public class ChatboxItemSearch extends ChatboxTextInput
 			// The client assigns "null" to item names of items it doesn't know about
 			if (!name.equals("null") && name.toLowerCase().contains(search))
 			{
-				// This may already be in the map due to canonicalize mapping the item to something we've already seen
-				results.putIfAbsent(itemComposition.getId(), itemComposition);
+				int itemId = ItemVariationMapping.map(itemComposition.getId());
+
+				// This may already be in the map due to canonicalize and itemvariation mapping
+				results.putIfAbsent(itemId, itemComposition);
 			}
 		}
 	}


raiyni:
This won't let you use explicit potion doses and such as icons.

deathbeam:
I think that is probably fine

raiyni:
I think that is bad ux, maybe make something that hides items with the same pixels.

dekvall:
I have done some experimenting in comparing icons present in the search results and to me it doesn't really seem viable for a couple of reasons. Hopefully someone else will have some better ideas.
 - Using `Object::hashCode`
    Unfortunately `BufferedImage` doesn't override this method `Object` and it's based on Object identity. 

 - Comparing the raw RGB values 
    I checked the search results for equality by just iterating over every pixel, this worked for some items, but not all. ex. super combat potion. 

 - Using `DataBuffer` to do compare the underlying data. 
    Same result as above, worked for most, but some items still had duplicates. 

The pixel and data matching also used a considerable amount of computing power when searching. I guess it could be simplified by comparing less pixels. But i feel that if we really want to remove duplicate icons, we should do it 100%. 

 - Improvements could include:
    - Setting arbritary bounds in pixel comparisons, we could try setting these bounds until every item is accounted for. But at that point we could rather make a json that stores all icon duplicates.
    - `pHash` - perceptual hashing, is something we could try but I don't really want to use a third party library. 
    - A multitude of other image processing methods, But, I really feel like that is overkill for the purpose of only matching search icons.

If someone figures out a smart (and preferably simple) way to do this efficiently. Please do tell so we can show all available icons. My current preferred potential solution is an `IconVariationMapping` implementation, with variations evaluated offline, though I don't really see any use cases for it other than this. 

Side note, I like the current `ItemVariationMapping` implementation and think it's preferred to showing duplicate items.

abextm:
we can just compare the rendering fields in the item composition

Darkkgreen:
I know this will probably sound pretty dumb but why shouldn't we define which items are variants and just exclude them using their IDs? I mean, I'm pretty sure the 'is this item a variant?' question will pop when programming other plugins and comparing images or pixels or whatever doesn't seem efficient to me

deathbeam:
We already have item variation mapping just for this.

Potion doses are still variants, but I guess @raiyni wants to preserve them for this specific scenario. For everything else, the variation mapping still applies.

raiyni:
Really anything that gets gobbled up by variations but still has a different image. Which may only be potions?

Darkkgreen:
> We already have item variation mapping just for this.
> 
> Potion doses are still variants, but I guess @raiyni wants to preserve them for this specific scenario. For everything else, the variation mapping still applies.

Oh, I see

Adam-:
I'm fine with this unless we think we need to extend it to all items with identical models?

dekvall:
Closing in favor of #9710

PR #9655, opened 2019-08-14
ItemsKeptOnDeath - Fix avernic defender price

TheStonedTurtle:
Adds a 4040k offset to the Avernic Defender to match the jagex death price. Image from discord.


Nightfirecat:
Can this add the skulled avernic+b ring(i) test case? (where the avernic is expected to be kept with protect item) The more the better.

TheStonedTurtle:
I've added the new test case and moved the existing test into the new function

PR #9667, opened 2019-08-15
Update NPCs in npc_health.json

xDemoN:
Supersedes #9311 

npc_health.json has not been updated in 6 months (and maybe not from the new wiki either). The osrs-data wiki dumper from deathbeam was used to pull all of this data on 8/15/19.

PR #9311, opened 2019-07-08
Update npc_health.json Forthos Dungeon

Decorth:

PR #9679, opened 2019-08-17
NPC Health: Update with 8/15 Wiki Scrape

xDemoN:
Supersedes #9311
Supersedes #9667 

For the time being, I'm sticking with just updating the current file as it's outdated and we now have the wiki scraper.

I'll open another PR later once I figure out doing it by NPC ID instead of name.


xDemoN:
Went through all of the differences one by one, now it's ready for review

dekvall:
We scrape health information off the wiki as of f6640844b5d58475a6bf07a8842c7b3522c807b7

PR #9690, opened 2019-08-18
clues: fixes sherlock clue step so that agility cape, agility cape(t) and max cape all count as part of the graceful set.

### LINKED ISSUE #9685: Agility cape does not count for graceful for sherlock master clue
when you get the rellekka rooftop sherlock master step and use agi cape untrimmed, it does not show as complete graceful outfit in the indicator (don't know about trimmed cape)

thomas-cedeno:
…and max cape all count as part of the graceful set. Fixes #9685


thomas-cedeno:
Fixes [#9685](https://github.com/runelite/runelite/issues/9685)

Alexsuperfly:
put that in the description and it will auto close the issue, otherwise it only links to it

Alexsuperfly:
it needs the fixes part too

Issue #9559, opened 2019-08-03
Add Menu Entry Swap for Teleport Crystal and Eternal Teleport Crystal


Jademalo:
**Describe the solution you'd like**
An extra option added to the Menu Entry Swapper plugin that allows the Left Click of the Teleport Crystal and Eternal Teleport Crystal to be swapped between Lletya and Prifddinas.

**Additional context**
This is a fairly simple request, since the crystal now has two options on it and I've been finding a lot more use for the Prif teleport.

OneProGoober:
Not saying it's a bad idea, but wouldn't shift chift click configurations do the trick?

Jademalo:
Shift click works certainly, but it's still an extra key to press instead of a pure one click. I'd rather have Prif be primary with a shift click Lletya personally.

Nightfirecat:
I don't think we're likely to prioritize one teleport location over another, given that we use the same behavior (that is, not adding configurations for them) for all of the various other teleport items which can teleport to multiple locations. (Ardy cloak, Morytania legs, Pharaoh's Sceptre, etc.)

Jademalo:
That's fairly reasonable, though to be fair, those three are wear on the left click by default rather than the teleport. I can't really think of another example of a pure teleport item with two different options.

Alexsuperfly:
it has two options, one on left click the other you can make shift left click, thats good enough, and all that will be done, this issue is invalid imo.

deathbeam:
As @Alexsuperfly said

PR #9691, opened 2019-08-18
Add Teleport Crystal to Menu Entry Swapper

Dennis-III:
Update the Menu Entry Swapper to include the Teleport Crystal; allowing Lletya to be swapped with Prifddinas


Alexsuperfly:
imo this is unnecessary, relevant discussion #9559

Dennis-III:
> imo this is unnecessary, relevant discussion #9559

I've seen the discussion and the OP makes good points. Swapping the teleport destination on the Teleport Crystal is a QOL improvement for many players that frequently visit Prifddinas, especially those that return to Zalcano after healing up at a POH.

deathbeam:
Just use shift-click. I dont see how are people failing to understand this. We are not going to add more menu entry swaps for items when you can simply just use shift-click. Also, the original issue was closed as wontfix, so closing this PR as well, sorry

PR #9692, opened 2019-08-18
worldmap: add on-click jumping for unambiguous single-target transports

adwitkow:
worldmap-jump

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlayMouseListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlayMouseListener.java
@@ -68,8 +68,12 @@ public class WorldMapOverlayMouseListener extends MouseAdapter
 				{
 					if (worldMapPoint.isJumpOnClick())
 					{
-						// jump map to position of point
-						WorldPoint target = worldMapPoint.getWorldPoint();
+						// jump map to target, or position of point
+						WorldPoint target = worldMapPoint.getTarget();
+						if (target == null)
+						{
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlayMouseListener.java
```diff
@@ -68,8 +69,18 @@ public MouseEvent mousePressed(MouseEvent e)
 				{
 					if (worldMapPoint.isJumpOnClick())
 					{
+						WorldPoint target;
+						if (worldMapPoint instanceof TransportationPoint)
```
deathbeam: This is wrong. You should not use class from plugin in core package. Just use getWorldPoint

adwitkow: getWorldPoint won't work I think:
image

I will push another solution later today.

deathbeam: The probably just extend the worldMapPoint api

adwitkow: Pushed the changes.

Now that I think about it, I should have probably just changed the jumpOnClick from boolean to WorldPoint, that way it could omit the target null-checking in the mouse listener. But I may be overthinking it, so I'm not changing anything until the review.

adwitkow:
pushed & squashed the changes

Issue #9689, opened 2019-08-18
Feral vampyres not counted towards vampire slayer task


Kamielvf:
The counter from the slayer plugin does not decrease as these Feral vampyres are slain, but they are a valid slayer target.

Alexsuperfly:
this bug was introduced in #9490 with the null check return before the killedOne() method in onExperienceChanged

Nightfirecat:
Hm, maybe we should change the fix from 9490 to allow null tasks to decrement the task counter by 1? Seems like an easier fix than fixing each case of an alternate monster being allowed as a slayer kill individually.

deathbeam:
I would rather just add the special cases

PR #9694, opened 2019-08-18
slayer: allow unknown tasks to still decrement counter

Alexsuperfly:
null tasks are technically valid, it only means they arent explicitly defined in the Task enum.  allow them through so that if there is a task capture failure the counter will still work.  
This functionality was especially useful when Konar was released as every one of her tasks were not captured properly and ended up being null, this allowed people to have counters that worked as they expected until the release that captured her tasks correctly.


Alexsuperfly:
this does fix the cause of the regression displayed in #9689 but the task itself should be corrected in the enum so it can become valid instead of silently fail

Kamielvf:
See https://github.com/runelite/runelite/issues/9689#issuecomment-522313147 also.

@deathbeam

Nightfirecat:
I suppose this should have a test added to cover this case.

PR #9703, opened 2019-08-20
Spelling/formatting and readability fixes.

CraigCraig:


runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -188,14 +188,14 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
 		// Return the text
 		if (inSeconds)
 		{
-			return Integer.toString((int) Math.floor(secondsLeft)) + "s";
+			return Math.floor(secondsLeft) + "s";
```
Adam-: This isn't the same thing. With secondsLeft = `5.0d` it would become `5.0s` instead of `5s` like it is now.

CraigCraig: Ah yes, I see. The cast to (int) fixes this, only meant to remove the unneeded cast to a string.

PR #9711, opened 2019-08-21
InfoBox separation support for stat boosts

jochim-vd:
Added a toggle to separate stat boosts infoboxes from boss timers, slayer, etc. infoboxes. This functionality could also be easily expanded upon by adding additional rules to the InfoBoxType enum.

One group:
Screenshot 2019-08-21 at 09 49 08

Separated:
Screenshot 2019-08-21 at 09 48 46

Menu option:
Screenshot 2019-08-21 at 11 21 39

addresses #1785


Adam-:
I don't like this because it makes the overlay renderer aware of the different infobox implementations. Also as soon as we do something like this we'll get 50 requests for things other than the combat boxes. If someone wants to make a way to dynamically split up the infoboxes that would be neat I think.

I've also wanted to try to make each of them into separate overlays for awhile so we could shift+right click configure them individually.

jochim-vd:
Alright I'll look into making them individually movable. Maybe a snapping functionality similar to the one we've got would be nice so we can still group infoboxes together?

raiyni:
See https://github.com/runelite/runelite/pull/5868 for a previous attempt. You need to prerender the separate layers off screen to prevent flicker. I also don't think using multimap filters is efficient.

jochim-vd:
Infoboxes now extend Overlay making them individually movable, InfoBoxOverlay and InfoBox were merged to achieve this. Currently wrapping and orientation functionality is broken because infoboxes aren't albe to be grouped together yet.

Is it worth it continuing in this direction or would another approach be more feasable?

Example:
Screenshot 2019-08-22 at 05 19 42

raiyni:
now they can't even be grouped

jochim-vd:
True, this is still far from complete. But now that the infoboxes are overlays the already implemented snapping feature could be extended to offer more flexibility. Perhaps extra (movable) zones could be added providing a ton of customizability.

Adam-:
Movable snap corners (and the ability to create new ones) could work, assuming they had logic to layout infobox overlays (which is mostly what is in InfoBoxOverlay now).

We could also try passing the right-click menu handling into the overlay somehow so that the infobox overlay can handle that itself. That would let us get per-infobox right click options at least, which is mostly what I wanted out of individual overlays anyway. Then all we would need is another InfoboxOverlay.

deathbeam:
I liked @raiyni's version way more than this. Infoboxes grouped per plugin is all that is needed. There is no scenario where it would need to be separated differently, neither there is a scenario where we would need right click menu for specific infobox, its anyway just per plugin, so each plugin infobox group could have plugin-specific right click menu as well.

PR #9712, opened 2019-08-21
slayer: correct Vampyre task spelling

### LINKED ISSUE #9689: Feral vampyres not counted towards vampire slayer task

Alexsuperfly:
fixes #9689 

Jagex uses British spelling. [Vampyre](https://oldschool.runescape.wiki/w/Vampyre)


deathbeam:
```
	public static final int FLYING_FEMALE_VAMPIRE = 3764;
	public static final int FLYING_FEMALE_VAMPIRE_3765 = 3765;
	public static final int FLYING_FEMALE_VAMPIRE_3766 = 3766;
	public static final int FLYING_FEMALE_VAMPIRE_3767 = 3767;
```

for some reason they also have this. Maybe plugin should count for both?

Alexsuperfly:
i have no idea where those would be but that does remind me that vyrewatch exist, and im pretty sure they count as vampyres

PR #9716, opened 2019-08-22
loot tracker: truncate name instead of kills

### LINKED ISSUE #5673: Loot Tracker: Truncate NPC name rather than kill amount

Hydrox6:
Closes #5673

short
long

Doesn't break with high killcount, even up to the limit of a `long` (though it's not really usable at that point anyway)
image


Adam-:
@psikoi can you review this?

PR #9720, opened 2019-08-22
chat: fix private chat username recolouring, improve responsiveness to config settings

### LINKED ISSUE #6351: Split Private Chat Recolour Not Fully Working
**Describe the bug**
The split private chat recolour does not work for the colon (':') nor the '_ has logged in/out' text. 
I have both transparent and normal recoloured, with both names and messages. 

**Screenshots**

screen shot 2018-11-05 at 4 29 51 pm

 - Mac OS
 - RuneLite version: 1.4.24
 - Launcher version: Unknown

adwitkow:
closes #6351

image

I was not able to test how MODPRIVATECHAT type works, so it may be caught in a crossfire unnecessarily.

We should maybe consider adding a setting to color From/To tags and the colon in a different color than the username itself after the config grouping PR is merged.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
@@ -119,12 +119,8 @@ public class ChatMessageManager
 
 		switch (chatMessageType)
 		{
-			case MODPRIVATECHAT:
-			case PRIVATECHAT:
-			case PRIVATECHATOUT:
-				usernameColor = isChatboxTransparent ? chatColorConfig.transparentPrivateUsernames() : chatColorConfig.opaquePrivateUsernames();
-				break;
-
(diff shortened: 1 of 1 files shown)


adwitkow:
So you would prefer basically wrapping the From/To, the username and the colon all separately?

Either way, coloring the username on chat message instead of on script callback still lets you do things like this when you change config after each message, which is inconsistent as From/To gets recolored correctly, while the username doesn't change at all until you relog:
image


Also, the login/logout messages don't have their own events, if I remember correctly they only appear in chat filter callbacks - and using them to recolor the messages feels hackish.

Apologies for a delayed response, my routine is a bit ruined this week due to a dead clutch in my car 😎

Adam-:
Hmm, then it does make sense to use the script callback. I would add a comment in `onChatMessage` explaining the pm username recolor is handled by the script then.

If you want to also get the log in/out messages, we would have to locate the scripts which add them and add in the eventing there, but we can do that separately probably.

adwitkow:
Changes pushed, initial message updated

PR #9723, opened 2019-08-23
Add toggle to not display combat boosts

chestnut1693:
Being only interested in the stat change timer and not the cluttering of the combat boosts, there is now a toggle to not display combat boosts.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
@@ -38,15 +38,23 @@ public interface BoostsConfig extends Config
 		NEVER
 	}
 
+	enum DisplayBoosts
+	{
+		NONE,
+		COMBAT,
+		NON_COMBAT,
+		BOTH
+	}
+
 	@ConfigItem(
-		keyName = "enableSkill",
-		name = "Enable Skill Boosts",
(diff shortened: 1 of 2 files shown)


Nightfirecat:
This would be better as a config dropdown than as a second checkbox at this point.

chestnut1693:
I've changed the checkboxes to a dropdown. Is it still possible to squash these commits?

Adam-:
Yes, you can squash them and force push.

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
```diff
@@ -38,15 +38,23 @@
...
+		BOTH
+	}
+
 	@ConfigItem(
-		keyName = "enableSkill",
-		name = "Enable Skill Boosts",
-		description = "Configures whether or not to display skill boost information",
+		keyName = "displayBoosts",
+		name = "Display Boosts",
+		description = "Configures whether or not to display boost information",
```
Adam-: This should probably be updated to reflect what the setting does. `Configures what skill boosts to display`?

chestnut1693: Should SKILLS from the enum then be changed to NON_COMBAT?

PR #9733, opened 2019-08-25
examine: fix pricecheck for items in seedbox

dekvall:
The menu option for Examine in the seedbox widget has a <col=9040> tag appended to it which explains the removal of tags. I also added the Seed Box to WidgetInfo.


dekvall:
Removed.

PR #9735, opened 2019-08-25
Add minnow to UntradeableItemMapping

dekvall:
One minnow is 1/40 of a raw shark and I added a field to the the UntradeableItemMapping to handle cases like this. The price check of minnows will not be 100% accurate at all times because of rounding errors but it will be a lot more accurate than it was before.


Unmoon:
Resolves #9802 

I'm thinking this should just use floats?

dekvall:
Updated to use floats.

PR #9738, opened 2019-08-25
fishing: Set status to NOT fishing when the flying fish is active

### LINKED ISSUE #8584: Minnow Fishing Plugin Adit
When fishing Minnows, and the 'shark' thing pops out of the water and eats your Minnows, the fishing overlay should indicate that you're not actually 'fishing' 

After recently transitioning from OSBuddy (OSBuddy have this feature, their plugin says 'Not ACTUALLY Fishing' when it happens. 

This is great for when I'm watching youtube or whatever and I can see the red text in the overlay in my peripheral vision.

dekvall:
Since you are not getting fish while the flying fish is active, you are not actually fishing any minnows.

Closes #8584

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
@@ -29,6 +29,7 @@ import java.awt.Dimension;
 import java.awt.Graphics2D;
 import javax.inject.Inject;
 import net.runelite.api.Client;
+import net.runelite.api.GraphicID;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -73,7 +74,7 @@ public Dimension render(Graphics2D graphics)
 
 		panelComponent.getChildren().clear();
 		if (client.getLocalPlayer().getInteracting() != null && client.getLocalPlayer().getInteracting().getName()
```
deathbeam: split each condition to separate line

PR #9741, opened 2019-08-26
add mouse listener to reset search results when clear button is pressed

### LINKED ISSUE #9545: SkillCalc: Clear button does not reset the search results

iamhating:
This pull request addresses the unwanted behavior detailed in this [issue](https://github.com/runelite/runelite/issues/9545).

Currently, when a user enters text into the search bar, then clicks the `x` button to clear the search text, the skill calculator does not reset the search result. The implementation in this pull request will reset the list of results to the default list, which is the list of all items associated with the skill calculator of the selected skill.

Closes #9545

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -89,6 +89,7 @@ class SkillCalculator extends JPanel
 		searchBar.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		searchBar.setHoverBackgroundColor(ColorScheme.DARK_GRAY_HOVER_COLOR);
 		searchBar.addKeyListener(e -> onSearch());
+		searchBar.addMouseListener(e -> onClear());
 
 		setLayout(new DynamicGridLayout(0, 1, 0, 5));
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -89,6 +89,7 @@
 		searchBar.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		searchBar.setHoverBackgroundColor(ColorScheme.DARK_GRAY_HOVER_COLOR);
 		searchBar.addKeyListener(e -> onSearch());
```
dekvall: Is there any problem with just replacing this line with:
```
searchBar.getDocument().addDocumentListener(new DocumentListener()
{
	@Override
	public void insertUpdate(DocumentEvent e)
	{
		onSearch();
	}

	@Override
	public void removeUpdate(DocumentEvent e)
	{
		onSearch();
	}

	@Override
	public void changedUpdate(DocumentEvent e)
	{
		onSearch();
	}
});
```
Or would that allow us to skip the addition of the mouse listeners?

iamhating: Didn't think of that. That works too since clearing the textfield would mean changing the document. I'll make those changes?

PR #9744, opened 2019-08-26
Update Magic Logbow Western Provinces Diary Step

dabolink:


dekvall:
Have not done the quest and this is the new text.

Issue #9748, opened 2019-08-27
'Ignore NPC' list for loot tracker


reowmeowmew:
I like to keep bosses/good drops in my loot tracker, but often have to reset a bunch of different things I kill throughout the day/week to make it so they're not there anymore taking up space.

I know you can now collapse each one to help a little bit, but I think it would be nice to have a right click 'do not track' to add certain monsters to that list, and a button to view your filtered list.

https://i.imgur.com/eYJ3K5A.png

Would be nice to completely get rid of all this ever even being tracked

Hydrox6:
Duplicate of #7482

PR #9752, opened 2019-08-27
[Loot Tracker] Add ability to hide specific NPCs/Events

dabolink:
Hide ignored npcs kills and value from overall calculator
Add right click hide NPC
Dim label of NPCs that are hidden when hideHiddenItems is toggled

show hidden events

hide hidden events

my gif generator still seems to be having issues. will update when I find a replacement

was on my todo list as well as saw #9748


Adam-:
this was added in #10699

PR #9756, opened 2019-08-28
WidgetInspector: Add widget picker

abextm:
Depends on !79

https://cdn.discordapp.com/attachments/419891709883973642/616076574734942208/2019-08-27_19-06-52.mp4


runelite-api/src/main/java/net/runelite/api/Client.java
```diff
@@ -1660,4 +1666,9 @@
 	 * Returns the max item index + 1 from cache
 	 */
 	int getItemCount();
+
+	/**
+	 * Makes all widgets behave as if they are {@link WidgetConfig#WIDGET_USE_TARGET}
+	 */
+	void setAllWidgetsAreOpTargetable(boolean value);
```
Adam-: Would it be better to just give it a mask to apply?

abextm: I don't think there are any other bits are useful to set globally

PR #9764, opened 2019-08-29
Inventory grid only shows once item is dragged

Jbleezy:
Inventory Grid Improvements

- Inventory Grid does not show up until item is dragged 5 pixels away, which is when the client recognizes the action as moving the item and not clicking it
- Can hold the item in place as long as you want, and the grid will not show up
- Grid shows up immediately after dragging the item
- Removed the option for drag delay as it is no longer necessary with these improvements

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
@@ -54,6 +54,10 @@ class InventoryGridOverlay extends Overlay
 	private final Client client;
 	private final ItemManager itemManager;
 
+	private static final int initialMousePointAreaRadius = 5;
+	private static Rectangle initialMousePointArea = null;
+	private static boolean passedInitialMousePointArea = false;
+
 	@Inject
 	private InventoryGridOverlay(InventoryGridConfig config, Client client, ItemManager itemManager)
 	{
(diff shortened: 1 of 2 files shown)


Adam-:
I can't tell if this is describing the current behavior or what you want it to do.

Jbleezy:
This is describing the behavior that it has been changed to. Normally the grid just shows up after x amount of milliseconds of holding an item. It doesn't take into account if you drag the item or not.

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
```diff
@@ -54,6 +53,10 @@
 	private final Client client;
 	private final ItemManager itemManager;
 
+	private static final int initialMousePointAreaRadius = 5;
```
deathbeam: UPPER_CASE if this is constant. Also move it at top

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
```diff
@@ -54,6 +53,10 @@
...
+	private static Rectangle initialMousePointArea = null;
+	private static boolean passedInitialMousePointArea = false;
```
deathbeam: Why are these static?

deathbeam:
Any update?

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
Jbleezy: They are static because they need to retain the value between function calls.

deathbeam: but render isnt static.

Hydrox6: It doesn't need to be `static` to retain value across function calls, it just needs to not be `final`.

PR #9771, opened 2019-08-29
runecraft: give minimap icons their own overlay

Hydrox6:
Moving the AbyssOverlay down to the correct layer stopped the runes from displaying on the minimap. Why an overlay was trying to do both Scene and Minimmap rendering I have no idea, but I've moved the minimap code to its own overlay. I've also moved the "what rifts to render" stuff to the plugin, since it's now common

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/AbyssMinimapOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runecraft/AbyssMinimapOverlay.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018, Seth <Sethtroll3@gmail.com>
+ * Copyright (c) 2019 Hydrox6 <ikada@protonmail.ch>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/runecraft/AbyssOverlay.java
```diff
@@ -127,7 +114,7 @@ private void highlightDarkMage(Graphics2D graphics)
 	private void renderRifts(Graphics2D graphics, DecorativeObject object)
 	{
 		AbyssRifts rift = AbyssRifts.getRift(object.getId());
-		if (rift == null || !rifts.contains(rift))
```
Adam-: Then this field needs to be removed

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/AbyssMinimapOverlay.java
```diff
@@ -0,0 +1,109 @@
...
+import net.runelite.client.ui.overlay.Overlay;
+import net.runelite.client.ui.overlay.OverlayLayer;
+import net.runelite.client.ui.overlay.OverlayPosition;
+import java.awt.Dimension;
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AbyssMinimapOverlay extends Overlay
```
Adam-: `class AbyssMinimapOverlay extends Overlay`

runelite-client/src/main/java/net/runelite/client/plugins/runecraft/AbyssMinimapOverlay.java
```diff
@@ -0,0 +1,109 @@
...
+			if (miniMapImage != null)
+			{
+				graphics.drawImage(image, miniMapImage.getX(), miniMapImage.getY(), null);
+			}
+		}
+
+		return null;
+	}
+
+	public BufferedImage getImage(AbyssRifts rift)
```
Adam-:
```suggestion
	private BufferedImage getImage(AbyssRifts rift)
```

Hydrox6: Turns out it was public in the original overlay, weirdly. Either way, fixed

PR #4012, opened 2018-06-25
slayer plugin: do not show task infobox on login

Jasper-ketelaar:
Fixes #3987.

This PR adds a boolean in setTask that specifies whether the method was called by performing an action (e.g checking your slayer gem) or if it was called by logging in/starting up. 

Previously if the user were to log in the Instant used to time out the infobox was set to be current. This caused unwanted behaviour where the infobox would appear for the first five minutes of being logged in eventhough the user had not performed any slayer related actions.

PR #9777, opened 2019-08-30
slayer: fix the plugin computing on free-to-play worlds

### LINKED ISSUE #3628: Slayer task showing up after losing members in a F2P world
Please check if your issue is not a duplicate by [searching existing issues](https://github.com/runelite/runelite/issues)

| Question    | Answer
| ------------| ---------------
| Java Version | x.y.z
| Operating System | e.g. Ubuntu/Windows/MacOS

<details>
 <summary>stack trace</summary>
 
 ```
  %paste your stack trace here%
 ```
</details>

adwitkow:
closes  #3628

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -52,6 +52,7 @@ import net.runelite.api.MessageNode;
 import net.runelite.api.NPC;
 import net.runelite.api.NPCComposition;
 import static net.runelite.api.Skill.SLAYER;
+import net.runelite.api.WorldType;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Technically, #4012 fixes this as well, and I rather prefer that solution personally.

adwitkow:
While #4012 does seem more elegant, while using it I can make the infobox appear on a free-to-play world by flipping the Task Infobox config item.

Adam-:
I much prefer #4012, though I feel like it will probably come back as a config option... maybe we can  just deal with that then, but it should really only have a world check ~1 place. Maybe two if we really care to block the config change thing but I don't think many people will hit that.

Adam-:
If you really want to block the npc highlights I suppose we can do that. The rest of this I think either can't happen (the chat messages) or shouldn't be blocked (!task)

Nightfirecat:
Most of this is covered now that #4012 has been merged. If blocking NPC highlights is wanted, this PR can be reopened with those specific changes, or it can be added in a separate PR.

PR #9778, opened 2019-08-30
Raids: recommended items display

Alexsuperfly:
Adds option to show common recommended items for raid rooms in the scouting overlay

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
@@ -28,6 +28,11 @@ import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.FontMetrics;
 import java.awt.Graphics2D;
+import java.awt.Point;
+import java.awt.Rectangle;
+import java.awt.image.BufferedImage;
+import java.util.HashSet;
+import java.util.Set;
 import javax.inject.Inject;
 import lombok.Getter;
 import lombok.Setter;
@@ -35,6 +40,7 @@ import net.runelite.api.ClanMemberManager;
 import net.runelite.api.Client;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
+import net.runelite.client.game.ItemManager;
 import net.runelite.client.plugins.raids.solver.Room;
 import net.runelite.client.ui.overlay.Overlay;
 import static net.runelite.client.ui.overlay.OverlayManager.OPTION_CONFIGURE;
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Can you provide a screenshot of what this looks like?

Alexsuperfly:
apologies, i thought i had one in there and never looked back to check

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
```diff
@@ -40,28 +47,38 @@
...
+import net.runelite.client.ui.overlay.components.ComponentOrientation;
+import net.runelite.client.ui.overlay.components.ComponentConstants;
+import net.runelite.client.util.ImageUtil;
 
 public class RaidsOverlay extends Overlay
 {
 	private static final int OLM_PLANE = 0;
+	private static final int SMALL_IMAGE_SIZE = 24;
+	private static final int REGULAR_IMAGE_SIZE = 32;
+	private static final int BOLD_IMAGE_SIZE = 27;
```
Nightfirecat: Is it intentional for the bold image size to be smaller than the regular image size?

Alexsuperfly: The text it sits next to would be so much larger with bold that I had to make the images smaller, and dynamically changing the panel size was an absolute bear that i stopped trying to fight after a few days and went with this.

Nightfirecat: Is the overlay smaller width with the small font? If not, then why is the size larger at regular size? (If it is, say so and dismiss this comment)

Alexsuperfly: The way the base panel is made, is it has a default minimum size which only the bold font really shows.  The small is set because the height of the panel is fully dynamic with the font and so if the images arnt scaled for that they fall off the bottom, and again changing that was difficult for it to do well so i just went with making them smaller for that case too.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
```diff
@@ -176,16 +193,65 @@ else if (plugin.getRoomBlacklist().contains(room.getName().toLowerCase()))
...
 						.right(name)
 						.rightColor(color)
 						.build());
 					break;
 			}
 		}
 
-		return panelComponent.render(graphics);
+		//if size is 1 the only thing in it is -1 which isnt an item
+		if (!showItems || recommendedItems.size() == 1)
```
Nightfirecat: Since `recommendedItems` is used beyond this point to loop over each item ID, it would be better to simply remove `-1` before doing this check. That way, it also does not need to be taken into account later on either.

```suggestion
// Remove -1 which is the item ID placeholder, not a real item
recommendedItems.remove(-1);

if (!showItems || recommendedItems.isEmpty())
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
```diff
@@ -176,16 +193,65 @@ else if (plugin.getRoomBlacklist().contains(room.getName().toLowerCase()))
...
+		itemImages.getChildren().clear();
+		itemImages.setBorder(new Rectangle(0, 0, 0, 0));
+		itemImages.setPreferredLocation(new Point(ComponentConstants.STANDARD_BORDER, verticalOffset));
+		itemImages.setBackgroundColor(null);
+		itemImages.setWrapping(2);
+		itemImages.setOrientation(ComponentOrientation.HORIZONTAL);
+
+		for (int item : recommendedItems)
+		{
+			final BufferedImage image = item != -1 ? getImage(item, imageSize) : null;
```
Nightfirecat: Following the comment above, this is simplified now that `-1` is not contained in the item set.
```suggestion
			final BufferedImage image = getImage(item, imageSize);
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsOverlay.java
```diff
@@ -176,16 +193,65 @@ else if (plugin.getRoomBlacklist().contains(room.getName().toLowerCase()))
...
+			if (image != null)
+			{
+				itemImages.getChildren().add(new ImageComponent(image));
+			}
+		}
+
+		//render the items over the main panel so they aren't occluded then return the dimensions of the main panel
```
Nightfirecat:
```suggestion
		// render the items over the main panel so they aren't occluded then return the dimensions of the main panel
```

Alexsuperfly:
this was added in runelite/plugin-hub#417

PR #9783, opened 2019-08-31
hiscore: use a builder to build the tooltips

### LINKED ISSUE #6368: More information on tooltips for hiscore entries
I, like a lot of newer players; are not 100% familiar with all of the entries on the hiscores, and some entries have no information as to what the rank is. (The bottom two ranks and the second to bottom right rank)

All other entries say the skill name of the entry, for consistancy and to save players the trouble of going to the hiscores to try and find out what those are, perhaps the tooltip could read: 
"Last man standing"   
"Rank: X"                   
etc
image

adwitkow:
- instead of using a raw html string to build the tooltips, we can use builder components and convert them on return
- hiding the elements based on availability is easier now (instead of showing three null values for a tooltip - `rank: unranked`, `xp: unranked` and `remaining xp: unranked` for some reason - we can just omit anything but `rank: unranked`)
- clue scrolls are still bad but they have a title now
- I am not sure how the `closingTags` ever worked
- also, closes #6368 because titles were added

I am still learning lombok so pls no bully

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
@@ -39,7 +39,9 @@ import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.EnumSet;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


raiyni:
What do we need this for to build swing tooltips?

adwitkow:
I dunno, feel free to actually take a look at the PR instead of posting unhelpful comments.

raiyni:
I did, that is why I am asking. It looks like it is specially wrote for the high scores panel.

Nightfirecat:
Titles were added in 2f85f574875f13ad60317ecfc2a4a79727ccef04. Given that and the inactivity here, I am closing this PR.

PR #9789, opened 2019-09-01
Added idle movement notifications to the Idle Notifier plugin

chowder:
Works similar to the Idle Animations component - sends a notification after `waitDuration` milliseconds if the player's character stops moving.

Some may find it useful for activities with long walking durations, like Runecrafting or Fishing.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -54,10 +54,21 @@ public interface IdleNotifierConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "movementidle",
+		name = "Idle Movement Notifications",
+		description = "Configures if idle movement notifications are enabled e.g. running, walking",
+		position = 3
+	)
+	default boolean movementIdle()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "logoutidle",
 		name = "Idle Logout Notifications",
 		description = "Configures if the idle logout notifications are enabled",
-		position = 3
+		position = 4
 	)
 	default boolean logoutIdle()
 	{
@@ -68,7 +79,7 @@ public interface IdleNotifierConfig extends Config
 		keyName = "timeout",
 		name = "Idle Notification Delay (ms)",
 		description = "The notification delay after the player is idle",
-		position = 4
+		position = 5
 	)
 	default int getIdleNotificationDelay()
 	{
@@ -79,7 +90,7 @@ public interface IdleNotifierConfig extends Config
 		keyName = "hitpoints",
 		name = "Hitpoints Notification Threshold",
 		description = "The amount of hitpoints to send a notification at. A value of 0 will disable notification.",
-		position = 5
+		position = 6
 	)
 	default int getHitpointsThreshold()
 	{
@@ -90,7 +101,7 @@ public interface IdleNotifierConfig extends Config
 		keyName = "prayer",
 		name = "Prayer Notification Threshold",
 		description = "The amount of prayer points to send a notification at. A value of 0 will disable notification.",
-		position = 6
+		position = 7
 	)
 	default int getPrayerThreshold()
 	{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -53,11 +53,22 @@ default boolean interactionIdle()
...
 
+	@ConfigItem(
+			keyName = "movementidle",
+			name = "Idle Movement Notifications",
+			description = "Configures if idle movement notifications are enabled e.g. running, walking",
+			position = 3
+	)
+	default boolean movementIdle()
+	{
+		return true;
```
Kamielvf: This should definitely not default to on

Adam-:
You can probably just use World location for this, since it only needs to be tested per game tick. Local points change per client tick.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
chowder: I agree, I've changed this now, thanks.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -638,6 +651,35 @@ private boolean checkAnimationIdle(Duration waitDuration, Player local)
...
+	{
+		if (lastPosition == null)
+		{
+			lastPosition = lastNotifiedPosition = local.getWorldLocation();
+			return false;
+		}
+
+		WorldPoint position = local.getWorldLocation();
+
+		if (lastPosition.distanceTo(position) == 0)
```
Adam-:
```suggestion
		if (lastPosition.equals(position))
```

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -638,6 +651,35 @@ private boolean checkAnimationIdle(Duration waitDuration, Player local)
...
+		{
+			if (local.getAnimation() == IDLE
+				&& lastNotifiedPosition.distanceTo(position) != 0
```
Adam-: Can't `lastNotifiedPosition` just be replaced with a `boolean` like the other notify methods?

chowder: It could, yeah; I've changed this now, thanks.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -638,6 +651,37 @@ private boolean checkAnimationIdle(Duration waitDuration, Player local)
...
+		WorldPoint position = local.getWorldLocation();
+
+		if (lastPosition.equals(position))
+		{
+			if (notifyPosition
+				&& local.getAnimation() == IDLE
+				&& Instant.now().compareTo(lastMoving.plus(waitDuration)) >= 0)
+			{
+				notifyPosition = false;
+				// Return true only if we weren't just breaking out of an animation
```
Adam-: You could maybe elaborate on this? I think this is so if there is another idle notification pending due to the animation it will instead defer to that?

Adam-:
This seems good now.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
chowder: Yes that's right. 

This is to cover the case where a player makes an interaction (from a distance) that also starts an animation. So if a player chops a tree, and the tree falls, the idle movement notification won't fire.

geheur:
@chowder do you remember why this is designed to wait until the player's animation is idle to fire off the idle movement notification? This seems unintuitive - when the player stops moving it should either fire a notification or not, not delay it until some animation finishes (which could be minutes later).

chowder:
Hey @geheur,

I don't think it *waits* until the player's animation is idle to fire the notification. Rather, it only fires the notification IFF the player's model animation at the time movement has stopped is IDLE. 

The rationale being that if I were to click on a tree from a distance to chop it, the notification shouldn't fire when the player reaches the tree to start chopping it - rather, the idle animation notification (if enabled) should be triggered once the tree has been chopped. 

Though I think I understand the behaviour you are describing as well. It was just in my experience doing it this way allowed me to leave both options toggled most of the time, instead of needing to disable the idle movement notifications for certain skilling activities like fishing, mining, and woodcutting.

PR #9793, opened 2019-09-02
ClientLoader: Cache the patched jar

abextm:
This works by checking the mtime in the first file in the vanilla jar against what was it was when the client was last patched. This adds a small amount of time to the slow (patching) path, but saves much more when we are able to hit the cache.

Depends on !80


abextm:
This functionality was added as part of a60d0f1a217c84fad98d61f270231103d4a62ff5

Issue #9488, opened 2019-07-27
Make "Guzzle" the left click option on rock cake


ChiSox:
**Is your feature request related to a problem? Please describe.**
Right clicking and selecting guzzle is tedius in NMZ. Shift click is an option, still takes two hands so is not as ideal.

**Describe the solution you'd like**
Add "Guzzle" as an option in the menu swapper plug-in.

**Describe alternatives you've considered**
You can add it to shift-click configuration, but this is a perfect fit for menu swapper. It's something done by many players and shift-click takes two hands instead of one.

**Additional context**
Guzzling rock cakes for NMZ is used very often by tons of players. It is the recommended method to train on the wiki. This would be a huge QoL for everyone. It has advantages over shift-click in that it takes one hand, and making it native in menu swapper would give it wider use by everyone. I wasn't familiar with shift-click until today myself.

Hydrox6:
Duplicate of #1860
TL;DR: use Shift Click configuration to make it work on Shift click
https://github.com/runelite/runelite/wiki/Menu-Entry-Swapper#shift-click-configuration

PR #9138, opened 2019-06-18
[Menu entry swapper] Switch eat with guzzle on dwarven rock cake

robinwithes:
Something to help out with dharokking in places like nightmarezone and giant mole.

Issue #7109, opened 2018-12-30
add dwarven rock cake to menu swapper


javedm1:
can we add the dwarven rock cake to the menu swapper? so guzzle can be made the left click/first option?

thanks for reading

Abextm:
Use shift click customization https://github.com/runelite/runelite/wiki/Menu-Entry-Swapper#shift-click-configuration

Hydrox6:
Duplicate of #1860

PR #9796, opened 2019-09-02
Swap eat for guzzle for the dwarven rock cake

EdwardSalter:
Guzzling the Dwarven rock cake provides damage of 10% vs 1hp when eating. This should make reaching 1 player hitpoint much less click intensive.


Hydrox6:
As has been mentioned in the 3 closed issues (#1860, #7109, #9488) and the 3 closed PRs (#2118, #8521, #9138), we'd rather Shift Click configuration is used for this use case

PR #9829, opened 2019-09-08
ChatCommandsPlugin: add !kc penance queen feature

knightjoe2222:
When the final reward widget appears after wave 10 in Barbarian Assault, increment and submit a user's penance queen kc. 

Usage: !kc [ba | pq | penance]

Basically a barbarian assault round counter, but since the penance queen is a boss I figured it needed a !kc command.

Since the game does not track # of completed ba games, all users start at 0. 

The code for identifying the final BA reward widget was based off of [BarbarianAssaultPlugin.java](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java#L119).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -49,6 +49,7 @@ import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.vars.AccountType;
 import net.runelite.api.widgets.Widget;
+import static net.runelite.api.widgets.WidgetID.BA_REWARD_GROUP_ID;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Is there no way to check how many gambles you've done in the past?

Adam-:
Is there really no way to get the queen kc at all? Counting wave 10 completions seems very error prone since it won't show the correct number, only what it has been able to count.

knightjoe2222:
@Adam- , there is no current way to get the queen kc in game. When you open the BA rewards shop, there is a “counter” for queen kills, but it decrements every time a reward is purchased that requires a queen kc. It also will cap when a user reaches the maximum points per role. it’s not an accurate measure. 

It seems like there are only two options if this addition is to be merged in, either start everyone at 0 or start from that reward shops number. I can understand if you don’t want a kc counter that doesn’t track the exact number of queen kills in a player’s history. As a frequent BA player, this is just something small that will benefit BA newbs going for torso as well as another metric for casual BA’ers to flex.

PR #9835, opened 2019-09-08
cluescrolls: Reset clue on new beginner/master step 

### LINKED ISSUE #9830: 2 Hot n  Colds in a row results in 0 possible locations
Doing beginners clues, I noticed this:

If you get 2 Hot/Cold clue steps in a row, for the second one you'll always have to drop the clue and pick it back up, as it shows 0 possible locations.

To reproduce, do a lot of beginner clues (??) until you get 2 Hot/Colds' in a row.

Nightfirecat:
When receiving a new beginner or master clue step (which can be detected
by checking the item ID shown in the chat dialog), the clue ID does not
change, because all beginner and master clues share a single ID. Hence,
we can reset the current clue when the "You've found a new clue!" dialog
appears to prevent stale clue information from persisting between steps.

Closes #9830


Nightfirecat:
I have realized that we can extend this behavior to master clues as well, so I've updated the contents of this PR further to handle that.

PR #9887, opened 2019-09-19
Bug/9875 Change nmz label Points/Hour

### LINKED ISSUE #9875: NMZ Points\Hour overlaps points when over 1,000,000
**Describe the bug**
NMZ Points\Hour overlaps points when over 1,000,000
https://i.imgur.com/R1Vsdk4.png

BramManuel:
Fixes #9875 

Chosen result based on that points is already defined above this label.("Points") P/h would be to cryptic and short to understand, so we chose for P/Hour.


makingcomment:
How about "Points/hr"?

Nightfirecat:
Alternately, if we cannot shorten the label, we can widen the panel. "Points/hr" seems like a reasonable abbreviation.

BramManuel:
I thought about widening the panel, but that looks ugly compared to the enemy health-bar. So i changed the label to "Points/hr: ".

dekvall:
Implemented in #10241

PR #9899, opened 2019-09-21
Add volume overrides

### LINKED ISSUE #9754: Volumes and Mute in SideBar
Think it's faster and easier if there was a option to add the volume control and mute button without touching the volumes in the side bar :D

Adam-:
Closes #9754

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicConfig.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2019, Adam <Adam@sigterm.info>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


EvanWalden:
I tried this and ran into an issue where the ingame setting ends up overriding the plugin setting after short delay.

To reproduce:

1. Set "Music Volume" to something low like 63
2. Set the in game volume to the middle dot.
3. Set the in game volume to the rightmost dot.

The volume will dip down to the plugin setting (63), then jump up to the ingame setting (255).

EvanWalden:
Yep, that fixed it.

EvanWalden:
While playing around with the setting I ran into another issue.

If I:

1. Set the plugin music value to 255.
2. Set the ingame music slider to 0. 

The music will stop, and can only be started again by setting the plugin music value to 0 and fiddling with the slider.

It doesn't seem to happen with the sound effects.

Adam-:
Thats pretty much intentional atm, its because when the slider hits 0 the game stops playing the music which I can't do much about. I think its actually fine. As long as the volume is above 0 it will use the plugin setting.

Seppobb:
@Adam- @EvanWalden  I belive this plugin/update on RL has been broken due to Jagex updating the settings menu. I tried searching for a solution to this but couldn't find one. Will this be adapted in the RL itself, or maybe the music plugin?

PR #9907, opened 2019-09-22
client: show names of members items in F2P

### LINKED ISSUE #5597: Display "Members objects" by their actual name
Is this feasible? If so, it would be a fantastic QoL improvement for ex-members.

### LINKED ISSUE #13282: Loot tracker displays items as Members objects
**Describe the bug**
Loot tracker displays all members items as members objects and doesn't stack all of them properly.

**To Reproduce**
1. Open runelite client
2. Log in an F2P account
3. Log out
4. Log in an P2P account
5. Now P2P account has loot tracker items displayed as members objects

**Expected behavior**
Runelite still recognizes members objects and displays them correctly both on F2P and P2P worlds. 
Or atleast after logging into P2P account, makes sure items have names instead of members objects.

**Screenshots**

 - OS: Windows 7
 - RuneLite version: 1.7.1
 - Launcher version: 1.5.2

Hydrox6:
Depends on !10
Closes #5597
Closes #13282

image

This PR now also updates sidebar-based plugins to always use the real item name even when on F2P

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/ItemComposition.java
+++ b/runelite-api/src/main/java/net/runelite/api/ItemComposition.java
@@ -32,11 +32,21 @@ import javax.annotation.Nullable;
 public interface ItemComposition extends ParamHolder
 {
 	/**
-	 * Gets the item's name.
+	 * Gets the item's name as it appears in game.
+	 * On a members server, this is always the item's actual name.
+	 * On a free server, this will be the actual name, with " (Members)" appended to it, e.g. Twisted bow (Members)
 	 *
-	 * @return the name of the item
+	 * @return the name of the item as it appears in game
 	 */
 	String getName();
+
+	/**
+	 * Gets the real item name, even if the player is on a F2P server.
+	 * Unlike {@link ItemComposition#getName()}, this will not have " (Members)" at the end on F2P servers.
+	 *
+	 * @return the real name of the item
+	 */
+	String getMembersName();
 
 	/**
 	 * Sets the item's name.
--- a/runelite-client/src/main/java/net/runelite/client/party/PartyService.java
+++ b/runelite-client/src/main/java/net/runelite/client/party/PartyService.java
@@ -102,7 +102,7 @@ public class PartyService
 			{
 				final int itemId = r.nextInt(client.getItemCount());
 				final ItemComposition def = client.getItemDefinition(itemId);
-				final String name = def.getName();
+				final String name = def.getMembersName();
 				if (name == null || name.isEmpty() || name.equalsIgnoreCase("null"))
 				{
 					continue;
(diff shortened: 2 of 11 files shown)


Crystalknoct:
Would this only show in f2p worlds or only if enabled? Judging by the code just by enabled.

Hydrox6:
It only affects F2P, since `Members object`s aren't a thing in P2P

dekvall (dismissed):

runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentificationPlugin.java
```diff
@@ -54,11 +63,22 @@ ItemIdentificationConfig getConfig(ConfigManager configManager)
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		client.setOverrideMembersItemNames(config.membersObject());
```
Adam-: this call is not threadsafe

Nightfirecat: There's still a call to this in `startUp`, which runs on EDT, not on client thread.

Hydrox6: Whoops, can't believe I forgot that. Fixed it.

Nightfirecat (approved):
LGTM pending internal PR

Hydrox6:
After a chat with abex, it was decided to expose the true name of the item to be usable by sidebar-based plugins. This also ends up fixing multiple other issues of `Members item` being shown when the client is loaded up on a F2P world.

runelite-api/src/main/java/net/runelite/api/ItemComposition.java
```diff
@@ -38,6 +38,13 @@
 	 */
 	String getName();
```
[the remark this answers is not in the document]
Hydrox6: Updated the javadoc. Still not entirely confidant with writing good docs but hopefully this is good enough.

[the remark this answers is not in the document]
Nightfirecat: I think the content is fine, but the formatting isn't right as-is. Since Javadoc is just HTML, this ends up yielding the following formatting when generated:

> #### getName()
> java.lang.String getName()
>
> Gets the items name as it appears in game. On a members server, this is always the item's actual name. On a free server, this depends on [Client.setOverrideMembersItemNames(boolean)](): * If it has not been called, or called with false, members items will have the name "Members object" * If it has been called with true, members items will have their actual name with " (members)" appended eg. Twisted bow (members)
>
> **Returns:**
> the name of the item as it appears in game

Primarily, this just needs the list to have list markup be HTML rather than markdown, eg.

```
[ ... ] this depends on {@link Client#setOverrideMembersItemNames}:
<ul>
  <li>If it has not been called, or called with false, members items will have the name "Members object"</li>
  <li>If it has been called with true, members items will have their actual name with " (members)" appended eg. "Twisted bow (members)"</li>
</ul>

[ ... ]
```

Nightfirecat (approved):
LGTM

abextm (approved):

Adam-:
added in https://github.com/runelite/runelite/commit/09be0b63ce222624594b14bbda9502ed660c1529 and https://github.com/runelite/runelite/commit/fd2c4a40af8102890c742dd014e03bac5e608e77

PR #9920, opened 2019-09-25
Notes: add font options

XrioBtw:
New options:
- Font
- Font style
- Font size
- Font colour

new_config_options

example_comic_sans_ms

example_consolas

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/notes/FontNotes.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/notes/FontNotes.java
@@ -1,0 +1,21 @@
+package net.runelite.client.plugins.notes;
+
+import java.awt.Font;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import net.runelite.client.ui.FontManager;
+
+@Getter
+@RequiredArgsConstructor
+public enum FontNotes
+{
+	RUNESCAPE_STANDARD(FontManager.getRunescapeFont()),
+	RUNESCAPE_BOLD(FontManager.getRunescapeBoldFont()),
+	RUNESCAPE_SMALL(FontManager.getRunescapeSmallFont()),
+	ARIAL(new Font("Arial", Font.PLAIN, 16)),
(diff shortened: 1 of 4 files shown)


jplsek:
Since the settings saved are meant to be cross platform, the font family chosen may not be available when syncing RuneLite on multiple operating systems. (That's also a reason why you can't configure the screenshot location yet, for example.) It may be best to remove the "Font" and "Font type" config, or change it such that the user can only change between the three RS fonts.

XrioBtw:
@jplsek I am fairly sure that a default fallback font will be selected if the current font is not available on the your operating system.
The most import part of this PR, along with font size and colour, is the ability to use a monospaced font. Neither of the three RS fonts are monospaced. I can add an unofficial monospaced runescape font to RuneLite resources if this is the preferable approach.

deathbeam:
I would rather just prefer switching to monospaced font for notes, I don't really see a point in having so extensive configuration for such simple plugin.

XrioBtw:
@deathbeam but what about the fancy _italic_, **bold** and colour possibilities!?

runelite-client/src/main/java/net/runelite/client/plugins/notes/FontNotes.java
```diff
@@ -0,0 +1,18 @@
+package net.runelite.client.plugins.notes;
```
Nightfirecat: This is missing a copyright header.

runelite-client/src/main/java/net/runelite/client/plugins/notes/NotesConfig.java
```diff
@@ -48,4 +57,52 @@ default String notesData()
...
+
+	@ConfigItem(
+		position = 3,
+		keyName = "fontColour",
+		name = "Font colour",
+		description = "The colour of the font"
+	)
+	default Color fontColour()
+	{
+		return new Color(198, 198, 198);
```
Nightfirecat: This color should be defined as a constant somewhere. It's currently only defined in RuneLite.colorschemes.

PR #9926, opened 2019-09-26
Menu Entry Swapper: Quick-leave instead of Leave tomb at Pyramid Plunder

### LINKED ISSUE #9912: Menu Entry Swapper: Quickleaving Pyramid Plunder
I would like for a menu entry swapper option on the tomb door in Pyramid Plunder so you can quick leave.

Emil-Hammer:
Adds a menu entry swapper option for Quick-Leaving instead of Leave tomb at Pyramid Plunder.

Closes #9912


Adam-:
This seems like a bad default, I've accidentally clicked the exits before thinking they were doors.

Emil-Hammer:
> This seems like a bad default, I've accidentally clicked the exits before thinking they were doors.

You're correct! I will change it so it's not on per default.

Adam-:
added in 5344cd6ddbd0e435bad7a088dbc266738319a5fc

PR #9939, opened 2019-09-30
remove demonic gorilla plugin

Trevor159:
While other things are up to interpretation jagex were very clear they do not want this.


chiragbharadwaj:
I still think this is up to interpretation. They only said that the client cannot tell you what to pray. What about a simpler plugin that at least tells you how many attacks have happened before the last prayer switch? This would at least ease you from counting attacks while still being compliant with the newspost.

I don't think we necessarily have to scrap the plugin altogether as a consequence.

Trevor159:
abex did it

PR #9940, opened 2019-09-30
agility: account for bonus exp from agility pyramid

### LINKED ISSUE #4213: Agility pyramid laps to level is inaccurate

Alexsuperfly:
fixes #4213
supersedes #5498


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -53,8 +53,15 @@ void incrementLapCount(Client client, XpTrackerService xpTrackerService)
 		final int currentExp = client.getSkillExperience(Skill.AGILITY);
 		final int goalXp = xpTrackerService.getEndGoalXp(Skill.AGILITY);
 		final int goalRemainingXp = goalXp - currentExp;
+		double courseTotalExp = course.getTotalXp();
+		if (course == Courses.PYRAMID)
+		{
+			// agility pyramid has a bonus exp drop on the last obstacle that scales with player level and caps at 1000
+			// the bonus is not already accounted for in the total exp number in the courses enum
```
Nightfirecat: This line is inaccurate. `totalXp` includes all experience gained during a lap of the course, including the final obstacle.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
```diff
@@ -53,8 +53,15 @@ void incrementLapCount(Client client, XpTrackerService xpTrackerService)
...
+			courseTotalExp += Math.min(course.getLastObstacleXp() + 8 * client.getRealSkillLevel(Skill.AGILITY), 1000);
```
Nightfirecat: Based on my comment above, this isn't correct. Additionally, this would not be accurate for any goal level which is more than 1 level away and is below the bonus xp cap. (related: #10076)

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
Alexsuperfly: that is what it should be, but for the agility pyramid it does not, and the reason the comment is there at all is to explain that and why the code under it is doing what its doing.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
Alexsuperfly: im not sure being level aware is something this can do, its miles more accurate that currently though.  If you have any suggestions on how to get it to be level aware i would welcome them.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
Nightfirecat: Given that the pyramid course already checks for a lap completion specifically by the ending location, I would rather see its `Course` definition changed so the ending obstacle xp can be fully replaced here instead.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
Nightfirecat: No suggestions as of now, but it would likely be good to add a `// TODO` comment or something of the sort.

PR #9950, opened 2019-10-01
runelite-client: Remove Menu Entry Swapper Plugin

graingert:
as per Jagex's wishes

> (we consider unacceptable) features which might modify how you select actions in the game or change how you click on menus, players, or NPCs, especially where the feature leads to you using fewer clicks than you would when using our client.


abextm:
We currently unsure if/how Jagex's statement affects the menu entry swapper, and we will not be removing it until we get further clarification from them directly.

graingert:
> We currently unsure if/how Jagex's statement affects the menu entry swapper, and we will not be removing it until we get further clarification from them directly.

I'm glad to hear. This is probably one of my most used QoL features

whartd:
> We currently unsure if/how Jagex's statement affects the menu entry swapper, and we will not be removing it until we get further clarification from them directly.

Can you explain what about their comment leaves ambiguity / lack of clarity? It seems very explicit to me.

Adam-:
If you take the guidance literally it would definitely be prohibited. However if you take it as *guidance* you can see that their intent is they are trying to prohibit features which unbalance the game (whether that be either unbalance skilling rates or high-level content which has high-level rewards). Most of our MES (maybe even all? except for the bones swap?) doesn't unbalance anything, and it is a far cry from all of the more explicit points they provided, so I think it is probably not what their intent was. So, as we already stated, we are trying to get further clarification on this first before we make any drastic moves.

whartd:
I guess I'm dragging into this a bias of having never thought the plugin was fair. The whole game is balanced around increased effort giving increased speed or efficiency. While some instances may be considered arbitrary, I see it as the role of polls alone to determine that. We have already had instances, such as gnome gliders and cata entrance, where jagex and the community decided it was right to change the left click option. As a result, I see this plugin as intentionally skipping/bypassing/ignoring the polling system.

I can draw my own line in the sand in the plugin. I e. Banking 1t faster with left click is, to me, overpowered considering hm activities it buffs. Enchanting left click w/ eluned seems fair to me since it's done infrequently and players always run up from a distance, making right clicking trivial in effort. But I don't think that's my job to decide. I think that responsibility belongs in the hands of the polling system, the precedent I see as having been set.

I'll be honest, I know it's one of the most popular plug-ins, but I hope jagex says to gut it.

PR #9955, opened 2019-10-01
WidgetItemOverlay: add clipping for bank

### LINKED ISSUE #10437: Runepouch overlay in bank
**Describe the bug**
When the runepouch is filled in the bank, and you scroll past it, the overlay gets over the 'frame' of the bank screen.

**To Reproduce**
1. put a filled runepouch in the bank relatively low on the first tab, so you have to scroll to it
2. scroll towards the runepouch location and back 
3. you will see that the overlay gets over the 'frame' 

**Expected behavior**
The overlay has to be between of the runepouch icon, and the frame of the bank screen

**Screenshots**
image

 - OS: Windows 10
 - RuneLite version: 1.5.42.3
 - Launcher version: 1.6.0

**Additional context**
It's not top priority, but in my opinion it is quite disturbing for some reason

raiyni:
Depends on !85
Closes #10437

Allows drawing partial overlays at the top of the bank.
* Live
72IwZCoGNx

* With this PR
java_8OKA4LxgbJ

Fixes the graphics clipping for the bottom of the bank, so items don't draw outside of the container bounds.
* Live
IM8mDYB5Tq

* With this PR
java_79bTFLnEaZ

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
@@ -26,6 +26,7 @@ package net.runelite.client.ui.overlay;
 
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.awt.Rectangle;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Adam-:
We've had performance issues in the past with `setClip` - see 3e35fa33fae0742fad3cb22a0cb26f5546c3c16b. Not sure how much that will affect this.

raiyni:
On my computer, there was no perceivable performance impact at max bank size. Regardless, I added checking for intersection of items so that only 16 (top and bottom row) clipping bounds are set at once instead of every item visible setting one.

dekvall:
This closes #10437

Adam-:
I've added most of this in 9a103f8ccd4928c6da8679ace7e02f1f9f874a63

raiyni:
Provided commit covers this public PR, the overlay won't clip at the top border of the parent still.

PR #9961, opened 2019-10-02
object markers: use id instead of name to identify objects

### LINKED ISSUE #9959: MLM Ladder object marker glitch
**Describe the bug**
A random invisible large object seems to be highlighted using the object marker plugin when both the upper and lower ladders are marked in MLM.

**To Reproduce**
1. Hold down shift and right click "mark object" on upper and lower ladder in MLM
2. Hop worlds or logout/in
3. You will see an extra invisible object marked that can't be removed
(Only way to get rid of it is to disable object marker plugin

**Expected behavior**
Only the actual lower and upper ladder used to highlight.

**Screenshots**
Screenshots 
image
image

 - OS: Windows 10
 - RuneLite version: Current as of 10/2
 - Launcher version: Current as of 10/2

dekvall:
With the addition of the ability to mark WallObjects some normal object markers created unremovable artifacts when hopping. This is basically because when getting the ObjectComposition of many objects they have the name "null" even some walls. Namely, when marking the top part of the ladder to upper level mlm, which has the name "null". When we then log in again the wall just below it is a part of the same tile with the same name "null" which adds it to the highlighted objects of the region. Since we can never select that wall again, we can never unmark it.

Closes #9959

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
@@ -61,8 +61,11 @@ class ObjectIndicatorsOverlay extends Overlay
 	{
 		for (TileObject object : plugin.getObjects())
 		{
-			if (object.getPlane() != client.getPlane())
+			if (object.getPlane() != client.getPlane() && !(object instanceof WallObject))
 			{
+				//If veins in mlm upper level are depleted when the player logs in, the plane is 1
+				//But if the player sees it get depleted it will be on plane 0
+				//I don't know if this extends to other WallObjects.
 				continue;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
```diff
@@ -61,8 +61,11 @@ public Dimension render(Graphics2D graphics)
 	{
 		for (TileObject object : plugin.getObjects())
 		{
-			if (object.getPlane() != client.getPlane())
+			if (object.getPlane() != client.getPlane() && !(object instanceof WallObject))
```
Nightfirecat: This should be in a separate commit, if you intended to include it in this PR

PR #9973, opened 2019-10-04
Add additional reset configuration for plugins

raiyni:
Some plugins create config keys that aren't reset by the config reset button (IE banktags).

This adds an overridable method to the Plugin class that that is called on reset button click.


deathbeam:
Hmm, also, what if plugin needs reset functionality but it does not have any configuration? It would be nice to also handle that scenario, but not sure if it's fitting in this PR.

PR #9976, opened 2019-10-04
inventory viewer: add the ability to minimize the overlay

### LINKED ISSUE #4148: Add option to minimize inventory viewer
Add the option to minimize the inventory overlay without having to turn off the plugin. Would simply be a minimize sprite.

raiyni:
Closes #4148

This adds an overlay menu option to inventory viewer `Toggle Inventory Viewer`. When clicked, it collapses the overlay down to 1 item image.

Collapsed
java_kT0raKQv7r


runelite-client/src/main/java/net/runelite/client/plugins/inventoryviewer/InventoryViewerPlugin.java
```diff
@@ -54,4 +60,14 @@ public void shutDown()
 	{
 		overlayManager.remove(overlay);
```
deathbeam: you should also reset the minimized state here.

Issue #6983, opened 2018-12-16
"Next + Restore" should not display when boosted skill is not displayed


Nightfirecat:
**Describe the bug**
When boosting a skill which is not displayed (eg. non-combat skill when "Enable Skill Boosts" is disabled), the next buff/debuff change text appears on its own. I feel this should not be shown unless the skill boost information is also shown.

![2018-12-16_11-50-29](https://user-images.githubusercontent.com/2199511/50058407-c0411880-016f-11e9-88a0-d3fd4fb3717a.png)

**To Reproduce**
1. Disable "Enable Skill Boosts", set "Display next buff change" to"BOOSTED"
2. Boost any non-combat skill and wait 60s for the first level decrement
3. See behavior described above

**Expected behavior**
If no boosted skills are displayed, no buff/debuff timers should be displayed

PR #9980, opened 2019-10-05
boosts: Hide restore timer when no boosts are visible

Nightfirecat:
Fixes runelite/runelite#6983

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
@@ -27,7 +27,7 @@ package net.runelite.client.plugins.boosts;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
 import java.util.Arrays;
-import java.util.HashSet;
+import java.util.EnumSet;
 import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Pushed some commits that (probably) fix that bug and the stream allocation each render. Haven't tested, but will test and squash when I have time tomorrow.

deathbeam:
Build is failing

Nightfirecat:
I've fixed the build failing and addressed the "ALWAYS" boost showing behavior.

runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
```diff
@@ -282,18 +284,20 @@ private void updateShownSkills()
...
 		// Reset is boosted
 		isChangedDown = false;
 		isChangedUp = false;
+		skillsToDisplay.clear();
 
 		// Check if we are still boosted
 		for (final Skill skill : Skill.values())
 		{
-			if (!BOOSTABLE_COMBAT_SKILLS.contains(skill) && !BOOSTABLE_NON_COMBAT_SKILLS.contains(skill))
+			if (!shownSkills.contains(skill))
```
Nightfirecat: Thinking about this more, it would be nice to have `isChangedUp` and `isChangedDown` set regardless of whether the given skill should be shown, and gate the +/- restore timer based on whether any skills will be shown (and the "ALWAYS" timer config). That way, if you changed your config at some point after the timer is known, you wouldn't have to wait another full cycle for another buff/debuff for the plugin to figure it out again. I'll look at this later today, I think it should be pretty straightforward to include that behavior.

Nightfirecat: Actually, that problem is a fair bit harder to solve, and is a use case which I don't think will be encountered often enough to be worth implementing.

PR #9982, opened 2019-10-05
Fix typo in barrows chest emote clue

eric-weaver:
Changes clue string text from `Do a jig at the barrow's chest` to `Do a jig at the barrows chest` to fix this master clue not showing up properly


dekvall:
Are you sure this is right? According to the [wiki](https://oldschool.runescape.wiki/w/Treasure_Trails/Guide/Emote_clues) and [this video](https://youtu.be/lefcHLz1_3g?t=1), although from 2016, the current clue text should be correct. Was this recently updated?

eric-weaver:
Yes, I had just gotten this clue and verified that this was broken before and fixed with this change. I meant to screenshot it but forgot

eric-weaver:
No screenshot, but here's the client warning log

> 2019-10-05 09:20:49 [Client] WARN  n.r.c.p.cluescrolls.ClueScrollPlugin - Encountered unhandled clue text: Do a jig at the Barrows chest. Beware of double agents! Equip any full barrows set.

PR #9983, opened 2019-10-05
xptracker: fix out-of-sync add to canvas menu

### LINKED ISSUE #9788: XP tracker sidebar does not update context menu after adding a tracker to canvas ingame
Adding a skill tracker trough the ingame stats panel will not change the context menu of said skill in the sidebar.
To remove the tracker you will first have to add it from the sidebar, and then the context menu will allow you to remove it.

Expected: The context menu in the sidebar will update as a tracker is added to canvas ingame

image

Hoffs:
Fixes #9788

Changes so that `XpInfoBox` uses same method of determining whether to show `Add to canvas` or `Remove from canvas` button by calling `hasOverlay`. Previously `XpInfoBox` would rely solely on its own state which meant that if canvas is added from anywhere else (e.g. skill context menu) this action is not known by the `XpInfoBox`.


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -139,18 +139,18 @@
 		popupMenu.add(resetOthers);
 		popupMenu.add(pauseSkill);
 		popupMenu.add(canvasItem);
+		popupMenu.addPopupMenuListener(new SimplePopupMenuListener(e ->
```
deathbeam: better (or well, simpler) would be to just use anonymous implementations here (we already do it below, no reason to overcomplicate things)

Hoffs: I'm not sure I follow, but PopupMenuListener interface is not functional so you can't do straight lambda like on L145. Anonymous implementation expands to
```java
		popupMenu.addPopupMenuListener(new PopupMenuListener()
		{
			@Override
			public void popupMenuWillBecomeVisible(PopupMenuEvent popupMenuEvent)
			{				
			}

			@Override
			public void popupMenuWillBecomeInvisible(PopupMenuEvent popupMenuEvent)
			{
			}

			@Override
			public void popupMenuCanceled(PopupMenuEvent popupMenuEvent)
			{
			}
		});
```
Hence why `PopupMenuOpenListener` is a functional interface for the required action and `SimplePopupMenuListener` is a simple wrapper around entire 3 method interface.

Hoffs: Is the implementation of all 3 methods inline better?

deathbeam: yeah, that is exactly what I meant

PR #9998, opened 2019-10-06
NPC unaggro timer: Added a notification feature to the Unaggro plugin

h0dgep0dge:
Not necessarily looking to have this merged as-is, feedback is welcome

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
@@ -52,6 +52,7 @@ import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.NpcSpawned;
 import net.runelite.api.geometry.Geometry;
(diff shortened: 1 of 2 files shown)


h0dgep0dge:
Thanks, I wasn't aware that there were additional steps to getting the code merged. I'd expected something along the lines of an interrogation and defense lol sorry if there's a prescribed procedure I've missed somewhere

Nightfirecat:
No need for that. Good code and a good use case is enough to get code merged most of the time.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -125,6 +129,9 @@
 	private WorldPoint lastPlayerLocation;
 	private WorldPoint previousUnknownCenter;
 	private boolean loggingIn;
+
```
Nightfirecat: Remove this extra newline.
```suggestion
```

Nightfirecat: Also, this should be reset along with the other plugin fields in `shutDown()`.

PR #9999, opened 2019-10-06
npchighlight: Apply correct highlights to changed NPCs

Nightfirecat:
This change ensures that tagged NPCs will be tracked throughout any
NpcChanged events, but highlighted NPCs will only display an overlay
when the current NPC name matches the config list.

Fixes runelite/runelite#5161
Closes runelite/runelite#6300


Nightfirecat:
I've finally gotten back around to this PR and rewritten it. The previous solution was pretty overengineered and probably isn't the right approach given the plugin's current state.

PR #10012, opened 2019-10-08
Skill calculator

dan-ashe:


Nightfirecat:
Can this be rebased and squashed so the individual features/functionality changes are in separate commits? As-is, it's not clear from looking at the whole of the changes exactly how much has been changed.

dan-ashe:
Sure, Ill be home 4pm UK time, I'll do it then

PR #10016, opened 2019-10-09
Music - Respect in-game volume mutes

TheStonedTurtle:
Adds a new config option to control whether the Volume overrides are always applied. This is meant to allow players to mute the game sounds without having to adjust the config settings.

Config defaults to true which does not change the current implementation.


TheStonedTurtle:
The PR has been updated to always respect the in-game volume mutes instead of basing it off a new config option.

PR #10021, opened 2019-10-10
Add teleport icon for Stony Basalt under 73 Agility

MESLewis:
Stony Basalt teleports you to a slightly different location with less than 73 agility.


Nightfirecat:
This is apparently gated by the agility level _and_ the Fremennik hard diary.

MESLewis:
You are correct, I'm going to look at adding it as a conditional map icon based on those requirements. I'll update the pr when I have more progress.

deathbeam:
There are multiple shortcuts we already have listed that also have diary requirements but we dont have support for that yet afaik, so I think this is fine?

Issue #10034, opened 2019-10-11
Add Player Kill value in chat


pbs1234:
Would like to request a simple feature that would enable you to see the GP value of the player kill as shown in lootracker. So defeated player xx value xx. 

This  way you do not have to have lootracker open all the time ( world switcher ) and see how much ur last kill is worth very quickly.

I'v seen this on some other clients its really nice feature.

Nightfirecat:
Is there a specific format that you've seen this message in? Ex. "You've killed *Zezima* for 73.5k loot."

pbs1234:
> Is there a specific format that you've seen this message in, or like to see it if it's added to RuneLite? Ex. "You've killed _Zezima_ for 73.5k loot."

Format is not too important, just regular message as it is just added loot value at the end is fine. So ur example would be completly fine.

PR #10035, opened 2019-10-11
loottracker: Show chat message after NPC and PVP kills

Nightfirecat:
Closes runelite/runelite#10034

Kill Noob 2019-10-10_19-31-39
Kill Zezima 2019-10-10_19-32-05

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
@@ -107,4 +107,24 @@ public interface LootTrackerConfig extends Config
 		description = ""
 	)
 	void setIgnoredEvents(String key);
+
+	@ConfigItem(
+		keyName = "npcKillChatMessage",
+		name = "Show chat message for NPC kills",
+		description = "Adds a chat message with monster name and kill value when receiving loot from an NPC kill."
+	)
+	default boolean npcKillChatMessage()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "pvpKillChatMessage",
+		name = "Show chat message for PVP kills",
+		description = "Adds a chat message with player name and kill value when receiving loot from a player kill."
+	)
+	default boolean pvpKillChatMessage()
+	{
+		return false;
+	}
 }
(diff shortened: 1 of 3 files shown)


deathbeam:
So maybe something like this for configuration:

```
enum LootMessage {
  PLAYER,
  NPC,
  BOTH,
  OFF
}
```

Nightfirecat:
I agree it wouldn't be hard, but it would be incredibly spammy a lot of the time.

deathbeam:
Can it? Because the low valuable drop is per item, and this would be per npc.

Nightfirecat:
I've added a second configuration flag to display a similar chat message for NPC kills.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -90,4 +90,24 @@ default boolean syncPanel()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "npcKillChatMessage",
+		name = "Show chat message for NPC kills",
+		description = "Adds a chat message with monster name and kill value when receiving loot from an NPC kill."
+	)
+	default boolean npcKillChatMessage()
```
deathbeam: I would make this config enum (OFF, NPC, PLAYER, BOTH)

Nightfirecat: Do we expect chat messages would be desired for other events besides PVP/NPC kills? If so, a dropdown would not scale for that.

deathbeam: Oh right, now we also support bunch of other stuff

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -394,6 +402,26 @@ public void onNpcLootReceived(final NpcLootReceived npcLootReceived)
 		final int combat = npc.getCombatLevel();
 
 		addLoot(name, combat, LootRecordType.NPC, items);
+
+		if (config.npcKillChatMessage())
```
deathbeam, line 406: This can be 1 method where you pass items and name

Adam-:
This does not work with noted items.

Nightfirecat:
#11529 has been opened to address that issue.

PR #10036, opened 2019-10-11
runelite-api: Add some more documentation

abextm:


abextm:
They are paired interface methods only afaik

runelite-api/src/main/java/net/runelite/api/IndexedSprite.java
```diff
@@ -25,119 +25,39 @@
...
-	 */
-	void setOriginalWidth(int originalWidth);
 
-	/**
-	 * Gets the height of the sprite.
-	 *
-	 * @return the height
-	 */
 	int getHeight();
-
```
JimGeersinga: this line shouldnt be removed

abextm: It is a paired interface method, so it should be removed

runelite-api/src/main/java/net/runelite/api/events/WidgetMenuOptionClicked.java
```diff
@@ -28,7 +28,7 @@
 import net.runelite.api.widgets.WidgetInfo;
 
 /**
- * An event where an option has been clicked in a {@link net.runelite.api.widgets.Widget}s menu.
+ * A MenuManager widget menu was clicked. This event is NOT fired for non-MenuManager menu options
```
deathbeam: Either add link to MenuManager here or do not mention it if it's not reachable from here (and as it is in client, it is not reachable).

abextm: I can't link to MenuManager because this event is in the wrong module

deathbeam: Makes sense. I guess we can move it in separate PR.

runelite-api/src/main/java/net/runelite/api/VarClientInt.java
```diff
@@ -28,7 +28,11 @@
 import lombok.Getter;
 
 /**
- * An enumeration of integer local variables.
+ * Client size only, content-developer integers
```
Adam-: ?

runelite-api/src/main/java/net/runelite/api/events/ScriptCallbackEvent.java
```diff
@@ -34,11 +34,12 @@
 public class ScriptCallbackEvent
 {
 	/**
-	 * The script being called.
+	 * The script that is currently being executed;
```
Adam-:
```suggestion
	 * The script that is currently being executed
```

PR #10038, opened 2019-10-11
Add faceguard to revertible items

### LINKED ISSUE #9943: Neitiznot faceguard GE value not shown in bank
Presently, the Neitiznot faceguard's value is not calculated in bank value calculator.

trelkins3:
Fixes #9943 - add faceguard to itemMapping.java under 'revertible items'; bank plugin should calculate and register faceguard value as intended now.


trelkins3:
@Nightfirecat - thanks for the review and recommendation, I'll take a look!

trelkins3:
Hold on... think I got the orders wrong again, the syntax is making more sense.

trelkins3:
@Nightfirecat - I have a couple questions:

1) It looks like for some items that have multiple significant components (ex. blessed saradomin sword), the current nomenclature is to map the untradeable item to the value of the most expensive component item. Is this correct? If so, I'll map the value of the faceguard to the basilisk jaw since the helm of neitiznot is negligible in comparison.

2) Should I just close this PR and open a new one with my final changes so I don't gum up the history of master if/when I get this figured out and merged?

3) Items are defined in item_variations.json, correct? It looks like neither the basilisk jaw or faceguard are defined there, so I'll get the item IDs and add them if that's the case.

Thank you so much for your help - I'm new to open source and have found that trying to contribute to RuneLite is a lot of fun!

Alexsuperfly:
How the ornament kits are handled currently is how you need to make this, thats is two separate entires for each peice

Dont open a new pr just fix this one

item_variations is for something else entirely

Nightfirecat:
1. BSS degrades to the Saradomin's Tear only. (the sword used to create it is lost) So in that case, and others (like the tentacle whip), that is properly set up.
2. The history can be squashed or rewritten by you, or by us when we merge it. You can keep your work here so it stays associated with the discussion we're having her.
3. Items are added in [ItemID](https://github.com/runelite/runelite/blob/master/runelite-api/src/main/java/net/runelite/api/ItemID.java), though it does seem to be missing the jaw. I guess we need @abextm or somebody else to run his cache tools to regenerate that file with the latest cache. (which will contain that item)

trelkins3:
OH I get what you mean by the ornament kits now and doing two entries per item! Took me a little bit - my apologies, thanks to both of you.

I'm not 100% confident in my squash abilities so I'll leave it to you. If you'd like me to try, I will.

Sounds like my CI builds are going to keep failing until ItemID is updated. Thank you for explaining how that works - I spent a solid 20 minutes digging around for a file that had all the IDs in it and then figured it was abstracted enough in the code where I've been working that it didn't matter.

TripleDeuce:
I have a suggestion, can the Blade of saeldor (c) have the same value as the Blade of saeldor (inactive) in the Bank value indicator. I understand why the avernic defender doesn't have any value, because once combined you can't get the avernic defender hilt back to potentially sell. The Blade of saeldor (c) is able to be uncharged and become the Blade of saeldor (inactive) which is still around 172M. I hate how my bank value drops about 172m by having the Blade of Saeldor with infinite charges.

PR #10042, opened 2019-10-11
Add documentation to enum and class files under runelite-api.

benjaminlgur:
Recreating PR #10040 with proper formatting.

I added some missing documentation I noticed while reading through it. I tried to keep added documentation in the same style of existing documentation.


benjaminlgur:
The purpose of this PR is to mainly to populate the [api documentation page](http://static.runelite.net/api/runelite-api/overview-summary.html) so that there will be less gaps and it will be more easily referenceable. If there is some interest in it would be willing to write documentation for every missing item in the documentation page.

PR #7401, opened 2019-01-17
Add Start-minigame in pyramid plunder to menu entry swapper

gregg1494:
Add Start-minigame in pyramid plunder to menu entry swapper

Closes #7138

PR #10046, opened 2019-10-12
Add menu swaps for pyramid plunder start-minigame and quick-leave

### LINKED ISSUE #7138: menu entry swapper at plunder
I think adding a left click "start-minigame" to the mummy at the beginning of the pyramid plunder would be a good idea. Menu swapper exists for a lot of things, so why not this one?

benjaminlgur:
This pull request Closes #7138 #7401. 

As opposed to other PRs this one closes both issues with a single config and I wrote the code such that it checks the mummy such that if other NPCs exist/are added with a start-minigame option this code will not have to change.

mummy swap
door swap
menu


Emil-Hammer:
Enabling both swaps with one option is not optimal. Pyramid plunder is done in a lot of ways. Some people rush first room and some go all the way to the last room. There's a huge risk of accidently clicking quick-leave.

benjaminlgur:
If having both swaps enabled is a problem it should probably be fixed with a drop down menu like many of the other menu swapper configurations. Splitting up the plug in options into 2 entry's would take up too much space and would be diffrent than the other existing elements of the plug in. Ill push a new commit as soon as I have the time, maybe today or tomorrow.

benjaminlgur:
image1

Adam-:
added in 5344cd6ddbd0e435bad7a088dbc266738319a5fc

PR #10052, opened 2019-10-13
Add option to manually broadcast raid layout message

### LINKED ISSUE #9541: Change party raid scout broadcast to be manual
Instead of spamming every single raid you scout, there should be a button on the screen when you enable the feature, so that once you've actually found a solid scout, you can choose to broadcast *that* raid to the discord party.

Right now, the feature is basically un-usable.

deathbeam:
Closes #9541

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
@@ -39,6 +39,7 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.InstanceTemplates;
+import net.runelite.api.MenuAction;
 import net.runelite.api.NullObjectID;
 import static net.runelite.api.Perspective.SCENE_SIZE;
 import net.runelite.api.Point;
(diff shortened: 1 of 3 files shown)


Adam-:
I don't see how this addresses the concern in #9541 of it just spamming uninteresting raids, since this still does that. I'd be okay with it just never broadcasting unless you click broadcast as suggested.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -252,6 +254,16 @@ public void onChatMessage(ChatMessage event)
 			}
 		}
 	}
+	@Subscribe
+	public void onOverlayMenuClicked(final OverlayMenuClicked event)
+	{
+		if (event.getEntry().getMenuAction() == MenuAction.RUNELITE_OVERLAY &&
+			event.getEntry().getOption().equals(RaidsOverlay.BROADCAST_ACTION) &&
+			event.getOverlay() instanceof RaidsOverlay)
```
Adam-: this is also just `event.getOverlay() == overlay`

deathbeam:
Because you can simply disable that, it's config option @Adam- . And this manual broadcast works always.

deathbeam:
Sure why not

PR #10063, opened 2019-10-14
Swap PvP "Attack" with "Walk here" in safe zones.

nateprat:
Swap "Attack" with "Walk here" in PvP Safe Zones 
https://github.com/runelite/runelite/issues/10064

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -311,4 +311,13 @@ public interface MenuEntrySwapperConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "swapPvpAttack",
+			name = "PvP Attack",
+			description = "Swap attack option when in PvP safe areas."
+	)
+	default boolean swapPvpAttack() {
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -313,9 +313,9 @@ public interface MenuEntrySwapperConfig extends Config
 	}
 
 	@ConfigItem(
-			keyName = "swapPvpAttack",
-			name = "PvP Attack",
-			description = "Swap Attack with Walk here when in PvP safe zones."
+		keyName = "swapPvpAttack",
+		name = "PvP Attack",
+		description = "Swap Attack with Walk here when in PvP safe zones."
 	)
 	default boolean swapPvpAttack()
 	{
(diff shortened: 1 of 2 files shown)


Adam-:
IMO this seems fine, but I want to wait for the upcoming Jagex statement about the MES before adding anymore stuff to it for now.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -33,19 +33,15 @@
 import javax.inject.Inject;
 import lombok.Getter;
 import lombok.Setter;
-import net.runelite.api.Client;
-import net.runelite.api.GameState;
-import net.runelite.api.ItemComposition;
-import net.runelite.api.MenuAction;
-import net.runelite.api.MenuEntry;
-import net.runelite.api.NPC;
+import net.runelite.api.*;
```
dekvall: No [wildcard imports](https://github.com/runelite/runelite/wiki/Code-Conventions).

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -604,13 +602,30 @@ else if (config.swapBones() && option.equals("bury"))
...
+	{
+		if (WorldType.isPvpWorld(client.getWorldType()))
+		{
+			Widget widget = client.getWidget(PVP_WORLD_SAFE_ZONE);
+			return widget != null && !widget.isSelfHidden();
+		}
+		else
+			{
+				return false;
+			}
```
dekvall: Dedent these lines.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -311,4 +311,14 @@ default boolean swapEnchant()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "swapPvpAttack",
+			name = "PvP Attack",
+			description = "Swap Attack with Walk here when in PvP safe zones."
```
dekvall: Dedent these lines.

nateprat:
With new update regarding MES, this will be deprecated

Adam-:
I'm not convinced the new update rules this out. Also, the safe zones used to behave this way before they added in the ~5 second timer that makes safe zones pvp when you enter them right? So this seems fairly harmless. Does this deal with that timer (eg. not hide the attack options then)?

deathbeam:
`Removes or deprioritises attack, cast (or similar) options from the minimenu in any scenario` I think its pretty clear?

Adam-:
Ah hm, looked over that. That's probably referring to whenever you actually *can* attack but okay, fair enough.

Nightfirecat:
[Per dekvall's comment](https://github.com/runelite/runelite/issues/10064#issuecomment-580506747) this feature is explicitly prohibited by Jagex now.

PR #10065, opened 2019-10-14
bank tags: add custom text overlay

Hydrox6:
Adds the capability to add arbitrary text to items in tag tabs. 

demo

`¬` is used as a splitter for the config due to it being a very rarely used character; originally, it was `:`, but I felt that people would be likely to use it and wonder why it didn't work.

`,` is replaced with space in the text due to its use in the CSV storage. I am able to change it if the desire to use commas is high, though it will require extra functions.

It does not work while searching if the tag doesn't have a tab due to the matching for the tag being fuzzy, though this could be added if there is a desire for it (though it would only be for when the user types the full tag name)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagOverlay.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2019 Hydrox6 <ikada@protonmail.ch>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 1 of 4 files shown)


deathbeam:
From Discord:

Imo this should be limited to just numbers instead of arbitrary text because there is no use case for arbitrary text. The upside of just numbers is that overlay space is limited, but for numbers we can use RS quant formatter to shorten it. Also, the actual use case for metadata like this would be user tagging inv, and it would automatically insert quant to field like this. But I dont really like inserting this number to arbitrary text field. For any other "tags in tags" use case, user can simply define multiple tags for item, so arbitrary text metada for it is pointless.

abextm:
There is no actual reason to limit it to numerics. If the user wants it to be stack formatted they can just type it stack formatted. A use case for arbitrary text is so I can put notes like `pray mage` on my prayer pots in my hespori tab, or `fill me` on a trident.

deathbeam:
The stack formatting should not be user's concern. And also the point was that this 'feature' would be used for stuff like inventory tagging, and do you want to stack format quantities before saving them when doing that? That simply loses information.

Also your examples do not make sense, how are they related to tag tabs? I don't think this PR was well-thought because it does not have any use-cases except quantity. Also, when exporting the tags, I don't want to bother with parsing some arbitrary garbage users can enter into it and then let them share it with each other. And the quantity use-case (what is still the only use-case for this) should be definitely exportable for stuff like loadouts for bosses.

abextm:
I have literally listed 2 use cases — that need to be associated with a tag
— for which I have almost wrote this for in the past. Just because you
don't want it doesn't mean it is useless. Additionally, there is no reason
to add MORE code for LESS functionality. Escaping characters or just
removing them from the input is trivial.

deathbeam:
It's not like I don't want this feature, I do, but for quantity. And as I already said before, implementing it like this just makes that use-case worse and more annoying to deal with. Its not more code for less functionality, its more code to sanitize the input so implementing the most useful part of it is nice and clean. And you ignored the part about formatting the number inputs before, as I said, pre-formatting them is bad as its losing data, and if I wanted to take just quantities from input (for exporting/importing) there is no way to distinguish them without ugly workarounds like adding some affix. And I definitely dont want to support exporting/importing text values. I guess the alternative would be if something is inputted as valid number, then pre-format it before displaying, that should be also easily distinguishable, but that is again another ugly workaround just because this feature wants to accept any input.

abextm:
Idk why you really want this data as a number somewhere. You can store digits in a string, and in the config, and in the export string. There is no reason for it have to be a defined format other than wasting my time removing it. If the user wants to see `18.5k` they can use their keyboard to type `18.5k`, and that way we don't have to have a number input that supports si prefixes.

deathbeam:
I already said multiple times why I want it as number, and why I dont want to export non-numbers and you totally skipped that part again so 🤷‍♂

abextm:
Its trivial to write something that can read a string that consists of things that are not digits out of another string, so I did completely ignore that made up limitation.

deathbeam:
Its pointless data that need to be escaped and urlencoded and then put to URL. Numbers would be simple, elegant and short. I never said its impossible, but arbitrary text is useless garbage that I dont plan to deal with on website.

abextm:
Its not useless or pointless, I have given some (imo more useful than quantity) uses. I don't see how changing a parsing regex to `[^,]` from `[0-9]` is difficult.

deathbeam:
Its useless for sharing. Your example use cases are useful only for the person who created them (as they are basically just notes without actual meaning).

abextm:
Then we can just not export the data if you want. Quantities are equally only useful for the person who created them, as consumable loadout can differ greatly between people depending on their skillset.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagOverlay.java
```diff
@@ -0,0 +1,78 @@
...
+import net.runelite.client.ui.overlay.WidgetItemOverlay;
+import net.runelite.client.ui.overlay.components.TextComponent;
+import javax.inject.Inject;
+import java.awt.Color;
+import java.awt.FontMetrics;
+import java.awt.Graphics2D;
+import java.awt.Point;
+import java.awt.Rectangle;
+
+public class BankTagOverlay extends WidgetItemOverlay
```
Nightfirecat: Package-private
```suggestion
class BankTagOverlay extends WidgetItemOverlay
```

PR #10068, opened 2019-10-15
Fix slayer infobox disappearing on '0' timeout

### LINKED ISSUE #10066: Slayer Infobox shouldn't disappear when timeout is 0
Currently, setting the Infobox timeout in the slayer plugin to 0 causes the infobox to instantly get removed, unlike every other plugin with an Infobox timeout, where 0 stops it from disappearing. The Slayer Plugin should be changed so that it works the same as other plugins.

trelkins3:
Resolve #10066 - actually fixed it using VCS in IDEA this time!

See #10067 for more history.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -390,7 +390,7 @@ public class SlayerPlugin extends Plugin
 			}
 		}
 
-		if (infoTimer != null)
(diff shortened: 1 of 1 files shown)


Alexsuperfly:
that already exists <https://github.com/runelite/runelite/blob/165994de08001ed7acc277f69ad3287603cc97d0/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java#L42>

deathbeam:
Oh, indeed. Then I am probably just confusing this scenario with different one.

trelkins3:
Previous review deemed invalid - this PR is ready for merge!

PR #10075, opened 2019-10-16
loot tracker: reset panel when logging in/out of RL account

### LINKED ISSUE #7896: Loot Tracker persists across accounts
**Describe the bug**
Whenever you use the persistent loot tracker it stays even if you switch Runelite accounts. (Maybe i'm strange and it's not much of an issue but I use different RL accounts for my main and ironman.)

**To Reproduce**
Kill something with the persistent loot tracker on and logged into a Runelite account.
Log out of that account and into a different one. 
Check the loot tracker. The loot remains. 

**Expected behavior**
The loot tracker shouldn't persist across different accounts. 

**Screenshots**

 - Windows 10
 - RuneLite version: 1.5.12
 - Launcher version: 1.6.0

**Additional context**

dekvall:
The workaround for resyncing when switching RL accounts was to switch the plugin off and on again. This uses that same behavior when switching accounts instead.

Closes #7896

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
@@ -398,6 +398,18 @@ class LootTrackerPanel extends PluginPanel
 	}
 
 	/**
+	 * Removes the records from then panel
+	 */
+	void removeRecords()
+	{
+		records.clear();
+		rebuild();
+		add(errorPanel);
+		actionsContainer.setVisible(false);
+		overallPanel.setVisible(false);
+	}
+
+	/**
 	 * Changes grouping mode of panel
 	 *
 	 * @param group if loot should be grouped or not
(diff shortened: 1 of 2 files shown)


Adam-:
`LoginResponse` seems wrong here instead of `SessionOpen`. They are two very different things. The session open event is required for the loot tracker to work just in general when starting the client I believe. Did you test this?

dekvall:
As i said in the commit message, on `SessionOpen` will fire every time you click the button to log in to your RL account, even during a login, and also upon a successful login. On every test i did, `SessionOpen` fired along `LoginResponse` although the latter only fired once. So it seemed appropriate to use to clear the loot as one wouldn't want to do that until actually logged in. I must admit that I am not 100 % certain on how `LoginResponse` works behind the scenes, but for this use-case it worked very well.

Adam-:
It is possible for the plugin to start prior to the user being logged in from the normal startup sequence (not clicking the login icon but simply being logged in due to having a saved session). There is no `LoginResponse` for that since it isn't a new login, but it is what loads the loot.

Nightfirecat:
Needs rebase.

Is there more work needed to complete this?

dekvall:
Yes, I still need to address Adam's comments. I will get to it when I find the time.

PR #10076, opened 2019-10-16
skill calc: add support for level aware actions

### LINKED ISSUE #2905: [Skilling Calc] Herbiboar
Unsure how to implement this myself but here is the xp given.

There is an avg of 2.56 inspections per Herbiboar.
Each inspection yields 50 xp so 128 xp per Herbiboar

Each Herbiboar harvested gives a base xp of 1950 and gives an additional 30 xp per hunter lvl until level 95. At level 96 you start to get an additional 19 per level.

so if you were level 85 you'd get a total of 2100 xp ( 1950 base + 5*30) 5 being the number of levels over 80

if you were lvl 97 you'd get 2438 xp ( 1950 + (5*15) + (2*19) ) the 2 * 19 comes from additional lvls over 95.

And ofc add 128 xp on top of that for the inspections

Apologies if it sounds a bit complex. The wiki probably explains it better lul

dekvall:
I also added all the relevant level aware actions i could think of
 - Herbiboar
 - Book of Arcane Knowledge
 - Wintertodt firemaking
 - Wintertodt construction

Closes #2905


makingcomment:
Would the agility pyramid completion bonus fit into this system?

dekvall:
Not at the moment, but it's probably not too hard to do the required modifications for it to be able to. I will look into it.

dekvall:
It should now be support for the agility pyramid as well.

Nightfirecat:
This is in some desperate need of unit testing. It looks like the code should cover cases where the goal XP is multiple levels beyond the current level, which would affect the number of iterations needed to reach intermediate levels, but having tests to know that's working as intended would be great.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -307,25 +310,89 @@ private void calculate()
...
+		}
+
+		int accumulatedActions = 0;
+		double accumulatedXp = 0;
+
+		while (accumulatedXp < neededXp)
+		{
+			final int accumulatedLvl = Experience.getLevelForXp((int) (startXP + accumulatedXp));
+			actionXp = calculateScaledActionXp(accumulatedLvl, action);
+			actionXp *= action.isIgnoreBonus() ? 1 : xpFactor;
```
Nightfirecat: Is this applying `xpFactor` a second time to `actionXp`? This exact calculation is already done at the top of this method.

runelite-client/src/test/java/net/runelite/client/plugins/skillcalculator/CalculatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import net.runelite.api.Experience;
+import net.runelite.client.plugins.skillcalculator.beans.SkillDataEntry;
+import net.runelite.client.plugins.skillcalculator.beans.XpScaling;
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class CalculatorTest
```
Nightfirecat: These tests should be added in the commit where the functions tested are added.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/beans/XpScaling.java
```diff
@@ -0,0 +1,36 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.skillcalculator.beans;
+
+import lombok.Getter;
+
+@Getter
+public class XpScaling
```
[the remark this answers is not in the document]
dekvall: I have just learned that the experience gained from the training manual which is obtained from organized crime scales quadratically with level. So I think that I will rework this representation.

PR #10088, opened 2019-10-18
clues: add Challenge Scrolls to their clue scrolls

### LINKED ISSUE #2740: Have Clue overlay stay up during challenge clues
Kind of annoying if you dont look at answer before talking to the npc as the answer goes away and once you get the challenge clue you cant bring the answer back

Hydrox6:
Closes #2740

This keeps the answer overlay visible if the user opens the Challenge Scroll itself, which is required when doing Elite clues.

There are a couple of clues I'm not sure about, specifically any with a number greater than 1,000 that doesn't have commas, but the information seems to be much more accurate than Sherlock's text was.

Also bundled are a few minor clue changes:
* Thanks to Cook from OSW, it has been discovered that the `HICK JET` clue never had a challenge scroll.
* The two Mandrith clues have been made distinct; now, one is for the question, and the other is for the Puzzle/Light box.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -504,17 +504,6 @@ public class ClueScrollPlugin extends Plugin
 			}
 		}
 
-		// (This|The) anagram reveals who to speak to next:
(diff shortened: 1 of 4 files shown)


Adam-:
`clues: make Anagram and Cryptic clues check all clue text` is not changing cryptic clues to check all clue text? Is there a reason it is all being reordered?

Hydrox6:
Oops, that should say `Cipher` not `Cryptic`. I'll go rename the commit

Issue #9808, opened 2019-09-04
Wrong achievement requirement (Karamja hard: Kill a metal dragon)


jachymb:
**Describe the bug**
The achievement diary requiement plugin says I need 12 Agility + 34 Woodcutting to kill a metal dragon in Brimhaven dungeon. This is not true, you can use the southern dungeon entrance which has no such requirement and is closer to the dragons anyway.

**To Reproduce**
Open the Karamja Achievement diary guide, scroll to hard.

**Expected behavior**
No skill requirement for this task.

Elpan1:
The southern entrance has a one time fee of 5000 trading sticks.

jachymb:
Yes, but getting these does not require any skill, imo even for ironmen. You can get them by exchanging jades/opals which can be obtained in various ways.

Elpan1:
I guess that's one way of seeing it. It's a very quick fix, however I don't know the others opinion about this matter.

Hydrox6:
The requirements are low, to the point where most accounts will have them well before doing the hard diary, and the 5k trading sticks have quite a bit of use. The accounts that don't want to or can't get those requirements will likely already know of the trading stick method.

trelkins3:
Is the best way to fix this simply to remove the level requirement text? Any update on this?

Elpan1:
The requirements for going trough the dungeon is **12 agility** and **34 woodcutting**. However, there exists an alternative method of paying **5k** trading sticks to bypass mentioned requirements. Which of these cases to choose and how to formulate this in the guide is merely a question of Runelite standards. How are stuff like this usually done, based on mechanical requirement or alt. methods?

dekvall:
Fixed in 9321d6e64e1fed4d52b8bee51b709c9fe8384e7a.

PR #10097, opened 2019-10-18
Wrong achievement requirement (Karamja hard: Kill a metal dragon)

Elpan1:
Closes#9808

The current skill requirements are not actual requirements for the diary but merely an optional route. In addition to this, both of the current so called skill requirements are already in possession of a player doing this part of the diary due to previous requirements. Finally, the current skill requirements can be eschewed by using tradingsticks.


Adam-:
I don't understand this "previous requirements" since you can do any of the diary tasks in any order you want, the tasks have no dependencies.

Elpan1:
Right, my mistake. However, the point still stands that the current skill requirements does not amount to the full truth in this matter (as they are only relevant at one optional route). For example, the official wiki does not present neither of these two skill requirements. I simply commited this to close #9808 and have no further judging in this matter, do as you feel best fit the projects guidelines.

PR #10102, opened 2019-10-19
xp tracker: enable xp progress tooltips for p2p skills on f2p worlds

dekvall:
The information is already available if you look it up on the hiscores and for me it's enough that the icon has some transparency to indicate that the skill is members only.
skill_hover


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerConfig.java
```diff
@@ -124,4 +124,15 @@ default OnScreenDisplayModeBottom onScreenDisplayModeBottom()
...
+
+	@ConfigItem(
+		position = 7,
+		keyName = "enableSkillTabTooltipForF2P",
+		name = "Enable skill tab tooltips for F2P",
+		description = "Configures showing xp progress in the tooltip for P2P skills on F2P worlds."
+	)
+	default boolean enableSkillTabTooltipForF2P()
+	{
+		return true;
```
deathbeam: Imo this should be disabled by default

dekvall: I know we have a policy of enabling as small amount of plugins as possible by default. But, in this case I think it makes sense to oppose that. It's such a minor annoyance and I know that I wouldn't stop to think; "Maybe RL has an option to override this members only tooltip text?" . This has been in vanilla for a long time, i think they changed it with mobile, and it's one of those things that should just work imo. "Members Only" gives no more info than just that, and the skill icon still has the changed opacity indicating it's different from the F2P ones, that it's a members only skill, with or without the option. It makes more sense to me that the F2P exclusive players disable the option if they want to.

In the end i'll do w/e you decide, because i know the option exists, but i still wanted give my thought process on the matter..

dekvall: From discord suggestion, I changed it to only override for players who have previously been members, but kept it enabled by default.

Nightfirecat:
Needs rebase

PR #10103, opened 2019-10-19
loottracker: add confirm dialog to reset all

dekvall:
Unfortunately, people lose their loot this way.

Logged in with sync warning:
confirm_loottracker_logged_in
Not logged in or sync turned off warning:
confirm_loottracker_logged_out

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
@@ -40,6 +40,7 @@ import javax.swing.BoxLayout;
 import javax.swing.ImageIcon;
 import javax.swing.JLabel;
 import javax.swing.JMenuItem;
+import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JPopupMenu;
 import javax.swing.border.EmptyBorder;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -312,6 +313,16 @@ public void mouseEntered(MouseEvent mouseEvent)
 		final JMenuItem reset = new JMenuItem("Reset All");
 		reset.addActionListener(e ->
 		{
+			final int result = JOptionPane.showOptionDialog(overallPanel,
+				"If you are logged in, this action will also permanently delete the current loot from the RuneLite website.",
```
deathbeam: This is incorrect. Remote loot is deleted only if sync is enabled

dekvall: Ah, indeed. Thanks!

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -74,6 +75,10 @@
 
 	private static final String HTML_LABEL_TEMPLATE =
 		"<html><body style='color:%s'>%s<span style='color:white'>%s</span></body></html>";
+	private static final String syncWarningText =
```
Adam-: By convention we use all caps for static/final fields.

PR #10110, opened 2019-10-20
Show notification when a "shown" impling enters the area

### LINKED ISSUE #10096: Notification when certain implings are "seen"
would be nice to get a windows notification when high end impling is "seen" if u happen to miss it

MattyJacques:
Closes  #10096

A notification is now displayed when a impling that the player has selected to be shown enters the players area


dekvall:
Wouldn't this be better implemented with an impliing specific enum config with `[Off, Highlight, Notify]`, instead of a `showNotification` boolean? I like having most implings highlighted, but i certainly wouldn't want to have to be notified for every eclectic impling i pass by in order to be able to receive notifications for dragon or lucky implings.

MattyJacques:
Ah, did not think of that. I'll implement the enum instead.

runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingsConfig.java
```diff
@@ -332,4 +332,15 @@ default boolean showName()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		position = 28,
+		keyName = "shownotification",
+		name = "Show notification when implings are seen ",
```
Adam-: I try to make the names fit on the config panel - maybe `Notify when seen`

runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingsPlugin.java
```diff
@@ -94,6 +98,11 @@ public void onNpcSpawned(NpcSpawned npcSpawned)
 
 		if (impling != null)
 		{
+			if (config.showNotification() && showImplingType(impling.getImplingType()))
```
Adam-: The `onNpcChanged` also needs to be checked too, since some implings spawn in that way.

Adam-:
I'd be okay with just the current method. Would `Notify` imply `Highlight`?

dekvall:
> Would `Notify` imply `Highlight`?

Yes, I think that it should. It makes sense for it to be highlighted if it's deemed important enough for a notification.

MattyJacques:
Dropdown does work better. Not quite sure if enum should be in config so let me know if it needs changed.

runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingsConfig.java
```diff
@@ -36,15 +36,22 @@
 @ConfigGroup("implings")
 public interface ImplingsConfig extends Config
 {
+	enum ImplingMode
```
Adam-: Does this work? I think it makes the enum class package private, and the config stuff needs to be able to access it.

MattyJacques: It does yeah, I wasn't sure whether to put this in the plugin class or in the config. What would you suggest?

Adam-: I must be mis-remembering then. I will check it later. This is probably fine as it is.

Adam-:
This shows a bunch of warnings on startup due to the key reuse:
```
java.lang.IllegalArgumentException: No enum constant net.runelite.client.plugins.implings.ImplingsConfig.ImplingMode.false
	at java.lang.Enum.valueOf(Enum.java:238)
	at net.runelite.client.config.ConfigManager.stringToObject(ConfigManager.java:585)
	at net.runelite.client.config.ConfigManager.getConfiguration(ConfigManager.java:376)
	at net.runelite.client.config.ConfigManager.setDefaultConfiguration(ConfigManager.java:512)
	at net.runelite.client.plugins.PluginManager.loadDefaultPluginConfiguration(PluginManager.java:192)
	at net.runelite.client.RuneLite.start(RuneLite.java:295)
	at net.runelite.client.RuneLite.main(RuneLite.java:245)
```
although I think this is fine. It reapplies the default value of NONE over top of everything, but everyone would lose their impling config regardless.

MattyJacques:
Is this during start? I didn't see it any my config saved my options so I thought it worked

Adam-:
Its from the existing config trying to be unmarshalled into the new enum, yes. It's fine.

PR #10114, opened 2019-10-21
Add enum constant for allotment/flower + change priff allotment patch

Elpan1:
Closes #10111
Added enum constant for herb/flower patch.
Changed priff previous(allotment/flower/herb) patch to new constant(allotment/flower).

PR #10115, opened 2019-10-21
worldmap: fix invalid description for prifddinas farming patch

### LINKED ISSUE #10111: World map tooltip lists herb patch at Prifddinas allotment
There is no herb patch there though.

Kamielvf:
This farming patch does not include a herb patch as the tooltip was describing.

image

Closes #10111


Kamielvf:
Duplicate of #10114 oops 😄

Issue #9632, opened 2019-08-10
Object marker colors remembered from time of placement


ChiSox:
Ground markers can remember the color they were at time of placement. Object markers should be updated to include this same feature. It would be useful to use for marking certain things different colors. For example: Deposit boxes one color while trees are another.

PR #10124, opened 2019-10-23
change ObjectMarker plugin to persist the current color at the time the object is marked

brettmoan2:
Change the attributes `ObjectIndicatorsPlugin.objects` and `ObjectIndicatorsPlugin.points` so that when an game object is marked, the _current_ `ObjectIndicatorsConfig.color` is tied to that object. This allows each game object to have a different indicator for their color.

Change serialization to accommodate the type change. 

Object individual color will also persist through 

- zone/area changes
- logout/login 
- and even closing/re-opening of the client.

**in a nut shell:** 
_In ObjectIndicatorsPlugin_ 

-  `List<TileObject> objects` becomes `Map<TileObject, Color> objects`
- `Map<Integer, Set<ObjectPoint>> points` becomes `Map<Integer, Map<ObjectPoint, Color>> points` 

_In ObjectIndicatorsOverlay_ 

- `config.markerColor()` becomes `entry.getValue()`


Nightfirecat:
Why is this not implemented as an added `Color` field on `ObjectPoint`?

brettmoan2:
main reason
- I didn't think of doing it that way.

secondary Reason:
- not sure how we would then re-implment the check
`objectPoints.containsKey(point)`

 https://github.com/runelite/runelite/blob/524f10bfe1fe68eaa6dd22baf781c9c7ed2df05b/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java#L428

As if the color is now different, the contains would always be false.

so if you think that's preferable, I'll give it a shot, to see how i could work around that check, (by implementing a comparison that excludes color) and see if that implementation still provides the same functionality

Nightfirecat:
Check how that's done in the ground markers plugin--you can exclude fields from the generated `@EqualsAndHashCode`. That way, we could maintain the comparison based only on the point's location, excluding the color.

brettmoan2:
@Nightfirecat I've made your suggested edits.

brettmoan2:
If anyone has any further feedback, its more thwn welcome. I tried to make sure to stick to the contribution guidelines. (Such as not using * imports)

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -466,11 +473,21 @@ private void savePoints(final int id, final Set<ObjectPoint> points)
...
 		// in most cases due to the specific object being identified being ambiguous, so remove
 		// them
+		// .map() necessary for added color to points
 		return points.stream()
 			.filter(point -> !point.getName().equals("null"))
+			.map(point ->
+			{
+				if (Objects.isNull(point.getColor()))
+				{
+					point.setColor(provideConfig(configManager).markerColor());
```
Nightfirecat: As mentioned above, inject the config class, then use `config.markerColor()`.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -421,7 +427,8 @@ private void markObject(String name, final TileObject object)
 			regionId,
 			worldPoint.getX() & (REGION_SIZE - 1),
 			worldPoint.getY() & (REGION_SIZE - 1),
-			client.getPlane());
+			client.getPlane(),
+			provideConfig(configManager).markerColor());
```
Nightfirecat: As mentioned above, inject the config class then use `config.markerColor()`.

brettmoan2:
@Nightfirecat 


I hope i made the changes to the way you were asking. I'm not all that confident in my understanding of java lingo, as i have limited experience, so i may have implemented your requested changes for config improperly.  



leveraging the provideConfig  function i added a ```	@Inject
	private ObjectIndicatorsConfig config;``` to the constructor of `ObjectIndicatorsPlugin` and then just used `config.markerColor()` where i was previously using `provideConfig(configManager).markerColor()`

this seem to work.

the other thing i tried to see if it would work, as i was sure was method injection, to add another parameter to the functions that need the config item. However when I tried that, it caused errors as `config.markerColor()` is needed inside of `markObject(String name, final TileObject object)` if i modified `markObject(String name, final TileObject object)` to `markObject(String name, final TileObject object, ObjectIndicatorsConfig config)` i would then need to pass a config object from where thats called, and one placec its called is in the plugin's `public void onMenuOptionClicked(MenuOptionClicked event)` if i make that `public void onMenuOptionClicked(MenuOptionClicked event, ObjectIndicatorsConfig config)` to handle that . the brakes the promised arguments for `onMenuOptionClicked` and prevents the plugin from loading.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
```diff
@@ -40,6 +42,7 @@
 import net.runelite.client.ui.overlay.OverlayPriority;
 import net.runelite.client.ui.overlay.OverlayUtil;
 
+
```
Nightfirecat: Now you've got an extra newline here for some reason?

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -466,11 +471,21 @@ private void savePoints(final int id, final Set<ObjectPoint> points)
...
 		{
 		}.getType());
+
 		// Prior to multiloc support the plugin would mark objects named "null", which breaks
 		// in most cases due to the specific object being identified being ambiguous, so remove
 		// them
+		// .map() necessary for added color to points
 		return points.stream()
 			.filter(point -> !point.getName().equals("null"))
+			.map(point ->
```
Nightfirecat: I assume this should be changed to match the behavior of the ground markers plugin. In fact, this whole PR should probably match the behavior which is present there. That is:

* A config for "remember marker colors" should be added
* When selected, the color used should be the color stored on the point
  * If that color is null (from existing config prior to this enhancement), use the current config color
* If deselected, all markers will use the current config color (current behavior)

As part of that change, this block will no longer be necessary--null color handling will then be done in the overlay.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsOverlay.java
```diff
@@ -95,15 +100,15 @@ else if (object instanceof GroundObject)
...
 			}
 
 			if (polygon2 != null)
 			{
-				OverlayUtil.renderPolygon(graphics, polygon2, config.markerColor());
+				OverlayUtil.renderPolygon(graphics, polygon2, color);
 			}
-		}
 
+		}
```
Nightfirecat: Remove the newline above this line, and re-add the one below it. (it looks strange on the diff)

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
brettmoan2: I like that. Good stuff man. I'll work on adding logic to overlay so that the null logic can be added there. by taking out this, then all "previously" marked points and any future marked points marked while config.("remember marker colors").equals(false), will have a null color for point. 

the save would then not over-ride this, and to get a specific color, a point would need to be un-marked and re-marked while the new config is true.

The new config would default to false, so no-one has any change in behavior unless they actively opt-in to using the new config.

Adam-:
Using a map to associate the color to the object is pretty hacky and makes it harder to extend the code in the future. The correct way to do this is to create a class for the highlighted objects which contains it (or maybe just a reference to the ObjectPoint).

dekvall:
This fixes #9632

PR #10126, opened 2019-10-23
xp globe: add skill icon scaling

### LINKED ISSUE #10116: Add scaling to the skill icons in XP Orbs
Currently, when scaling an XP orb below the default size of 35, the edges of the skill can clip out of the edges of the orb, which looks very bad. Conversely, making the orb larger than 35 results in a lot of empty space around the icon, which looks bad.

To fix this, I propose that the icon scale. Scaling smaller than default will have to cap somewhere, however. The user should be given an option as to whether they want the icon to scale or not. It might also be a good idea to add a lower cap to the size of the orbs, now that we can easily do this.

NathanielPather:
Closes #10116 

### Features
Skill icons scales with xpOrbSize.
xpOrbSize changed to have minimum value of 20.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
@@ -25,8 +25,10 @@
 package net.runelite.client.plugins.xpglobes;
 
 import java.awt.Color;
-
-import net.runelite.client.config.*;
+import net.runelite.client.config.Alpha;
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
 
 @ConfigGroup("xpglobes")
 public interface XpGlobesConfig extends Config
@@ -144,9 +146,6 @@ public interface XpGlobesConfig extends Config
 		return 2;
 	}
 
-	@Range(
-			min = 20
-	)
 	@ConfigItem(
 		keyName = "Orb size",
 		name = "Size of orbs",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
@@ -147,7 +147,7 @@ public interface XpGlobesConfig extends Config
 	}
 
 	@Range (
-			min = 20
+		min = 20
 	)
 	@ConfigItem(
 		keyName = "Orb size",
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -221,11 +223,13 @@ private void drawSkillImage(Graphics2D graphics, XpGlobe xpGlobe, int x, int y)
...
+
 		graphics.drawImage(
-			skillImage,
-			x + (config.xpOrbSize() / 2) - (skillImage.getWidth() / 2),
-			y + (config.xpOrbSize() / 2) - (skillImage.getHeight() / 2),
-			null
+				resized,
+				x + (config.xpOrbSize() / 2) - (resized.getWidth() / 2),
+				y + (config.xpOrbSize() / 2) - (resized.getHeight() / 2),
+				null
```
Hydrox6: Remove the excess indents

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
```diff
@@ -146,6 +146,9 @@ default int progressArcStrokeWidth()
 		return 2;
 	}
 
+	@Range (
+			min = 20
```
Hydrox6: Remove the excess indent

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -53,6 +53,8 @@
 import net.runelite.client.ui.overlay.OverlayUtil;
 import net.runelite.client.ui.overlay.components.LineComponent;
 import net.runelite.client.ui.overlay.components.PanelComponent;
+import net.runelite.client.util.ImageUtil;
+
```
Hydrox6: Remove the extra newline

NathanielPather: No idea how I added the extra indents. Thanks tho.

runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -221,10 +222,12 @@ private void drawSkillImage(Graphics2D graphics, XpGlobe xpGlobe, int x, int y)
 			return;
 		}
 
+		BufferedImage resized = ImageUtil.resizeImage(skillImage, config.xpOrbSize() / 2, config.xpOrbSize() / 2);
```
deathbeam: I don't think this should be resizing the image every single render cycle

Nightfirecat: To expand on this, you can generate the image once in the plugin, and update it on config change accordingly, then get that image from the overlay so it is not generated once per frame.

Hydrox6:
This was added in e7f50433a352ab996b7b5b6114b504194312000d

PR #10147, opened 2019-10-27
Add Fairy Ring to emote clue scrolls

### LINKED ISSUE #10143: Haunted Woods Panic emote clue (Hard) fairy ring
Would be nice to add fairy ring ALQ to this clue.

Darkkgreen:
Now it hints the closest fairy ring, if applied. One can also turn off the feature if wanted.

The only thing I'm not sure is the way I'm retrieving the configuration from the plugin.

Closes #10143


Adam-:
IMO a configuration option for this is not necessary. Also I think you forgot to `git add` a file

Darkkgreen:
I'll remove the config option and verify if I forgot to add a file later today

Darkkgreen:
Removed the config. And I didn't forget to `git add` a file, that was it

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -52,6 +52,8 @@
 import static net.runelite.client.plugins.cluescrolls.clues.emote.STASHUnit.SHANTAY_PASS;
 import net.runelite.client.plugins.cluescrolls.clues.item.ItemRequirement;
 import static net.runelite.client.plugins.cluescrolls.clues.item.ItemRequirements.*;
+
+import net.runelite.client.plugins.fairyring.FairyRings;
```
deathbeam: Fix your imports (check code style section in wiki)

Adam-:
I think that I prefer #10149 to this

deathbeam:
I like this one way more. Well, except the fairy ring arg should be optional, but imo it should be in separate section, it makes more sense than just appending it.

Adam-:
For the most part the clues aren't actually *at* the fairy rings, it is just a convenient way to travel. What if there is one near a spirit tree etc? Would we make an enum/member for them too?

deathbeam:
Well clue being near fairy ring is very common compared to it being near spirit tree, but I guess. Maybe separate "Travel:" section? That would apply to everything like this.

deathbeam:
Implemented in https://github.com/runelite/runelite/pull/10149

PR #10149, opened 2019-10-27
clues plugin: add more fairy rings to coord clues

trimbe:


Adam-:
can you also add #10143 ?

PR #10151, opened 2019-10-28
bank plugin: add ctrl+f hotkey to initiate bank search

### LINKED ISSUE #3994: CTRL-F to open search in bank
It would be nice to be able to press CTRL-F to open the search prompt in the bank. That's all.

trimbe:
Closes #3994


trimbe:
I was checking that in `initSearch` but I've added it to `keyPressed` as well so the event can be consumed.

PR #10153, opened 2019-10-28
Adding Zeah Runecraft Plugin

Pinibot:
This plugin will highlight the dense runestones with certain colours to provide context as to what actions can be taken.

When a dense runestone is depleted, it is highlighted red.
When the player is currently mining a dense runestone, the runestones are highlighted orange.
When the player is not currently mining, the runestones are highlighted green.
When a depleted runestone becomes available again, it will either be highlighted orange or green, depending on if the player is mining or not.

The highlighting can be diabled by the new config group "Zeah Runecraft".
I debated whether to combine it with the existing "Runecraft" config group, but this looks very Abyss specific.

The purpose of this plugin is to provide a better experience when AFKing as it can be easy to miss sound cues when a rock depletes.


runelite-client/src/main/java/net/runelite/client/plugins/zeahrunecraft/ZeahRunecraftPlugin.java
```diff
@@ -0,0 +1,170 @@
...
+		mining = MiningAnimations.isPlayerMining(playerAnimationId);
+	}
+
+	@Subscribe
+	public void onGameObjectSpawned(final GameObjectSpawned event)
+	{
+		GameObject gameObject = event.getGameObject();
+		int gameObjectId = gameObject.getId();
+
+		// Check to see if the GameObject is a Dense Runestone
```
Adam-: All of this just reduces to `if (gameObjectId == obj1 || gameObjectId == obj2) runestones.add(...`

Adam-:
Does this make sense as its own plugin or would it fit into the mining plugin?

deathbeam:
Imo it should go to mining plugin. Also I dont think changing colors of the dense runestones based on mining status is useful at all?

runelite-client/src/main/java/net/runelite/client/plugins/zeahrunecraft/ZeahRunecraftPlugin.java
Pinibot: Sort of, because each of the runestones has a separate varbit associated with it it would need to be something like:

```
if(gameobjectId == 8981)
{
     runestones.add(new DenseRunestone(false, gameObject,
						Varbits.NORTH_DENSE_RUNESTONE_DEPLETED));
}
else if (gameobjectId == 10796)
{
     runestones.add(new DenseRunestone(false, gameObject,
						Varbits.SOUTH_DENSE_RUNESTONE_DEPLETED));
}
```
Which I feel is less generic? Happy to change to the above though if you prefer.

I will override equals and hashcode on DenseRunestone though so that when adding it to the list it will only add if the gameobjectId is unique though. I need that logic in there though as during testing running from the stones to the altar and back would spawn new runestone gameobjects but not despawn the old ones so i ended up with far more than just two.

Adam-: Look at using `ObjectId` and/or `NullObjectId` for those constants. Are you sure duplicate spawn events happen? That shouldn't happen if you are handling both despawns and the game state change event. I think.

Pinibot: I had debug mode on and my breakpoints were never hit in onGameObjectDespawned and onGameObjectChanged during the normal laps. IIRC the despawn happened when i finally teled away but the runestones would keep spawning new gameobjects with different worldpoints? I removed the onGameObjectChange override altogether because it didn't seem to ever get invoked.

The runestones seem to work differently from normal rocks (again from my limited knowledge). I tested mining a normal rock showing objectIds and the depleted rock and the full rock have different ids. When a runestone depletes however the ID never changes, only the value of the associated varbit. Will these null object IDs still be present in this case?

deathbeam:
Well, that is my point. Mining plugin already has not mining/mining indicator.

Hydrox6:
Actually, the mining plugin only has the respawn timer stuff. Session tracking is currently in the MLM plugin for some reason.

deathbeam:
Oh, ffs. Well, then, that should be added instead of recoloring to mining plugin for the sake of consistency.

deathbeam:
Looks good. Also you probably want to look at this because currently your code style is a mess: https://github.com/runelite/runelite/wiki/Code-Conventions

runelite-client/src/main/java/net/runelite/client/plugins/mining/DenseRunestone.java
```diff
@@ -0,0 +1,99 @@
...
+	private boolean depleted;
+
+	@Getter
+	private final GameObject gameObject;
+
+	@Getter
+	private final Varbits depletionVarbit;
+
+	@Override
+	public int hashCode()
```
deathbeam: Just use `EqualsAndHashCode` from Lombok

runelite-client/src/main/java/net/runelite/client/plugins/mining/DenseRunestoneMineOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+		setLayer(OverlayLayer.ABOVE_SCENE);
+		this.client = client;
+		this.plugin = plugin;
+		this.config = config;
+	}
+
+	@Override
+	public Dimension render(final Graphics2D graphics)
+	{
+		if (config.showDenseRunestoneStatus())
```
deathbeam: Change this to early return instead of wrapping everything in it

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -42,12 +49,35 @@
 import static net.runelite.api.ObjectID.ORE_VEIN_26662;
 import static net.runelite.api.ObjectID.ORE_VEIN_26663;
 import static net.runelite.api.ObjectID.ORE_VEIN_26664;
+
+import java.time.Instant;
```
deathbeam: Your imports are still incorrectly formatted

runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
```diff
@@ -24,12 +25,46 @@
...
+import static net.runelite.api.ObjectID.ROCKS_36205;
+import static net.runelite.api.ObjectID.ROCKS_36206;
+import static net.runelite.api.ObjectID.ROCKS_36207;
+import static net.runelite.api.ObjectID.ROCKS_36208;
+import static net.runelite.api.ObjectID.ROCKS_36209;
+
 import java.time.Duration;
 import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
```
deathbeam: same here

runelite-client/src/main/java/net/runelite/client/plugins/mining/DenseRunestoneMineOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+						highlightColor, highlightColor.darker());
+				}
+			}
+
+		});
+
+		return null;
+	}
+
+	private Optional<Color> getDenseRunestoneHighlightColor(final boolean isDepleted, final boolean isMining)
```
deathbeam: You can simply return null here

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningOverlay.java
```diff
@@ -72,7 +72,7 @@ public Dimension render(Graphics2D graphics)
 		}
 
 		Instant now = Instant.now();
-		for (Iterator<RockRespawn> it = respawns.iterator(); it.hasNext();)
+		for (Iterator<RockRespawn> it = respawns.iterator(); it.hasNext(); )
```
deathbeam: revert changes in this file

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -57,12 +83,9 @@
 	name = "Mining",
 	description = "Show ore respawn timers",
 	tags = {"overlay", "skilling", "timers"},
-	enabledByDefault = false
-)
+	enabledByDefault = false)
```
deathbeam: revert this `)` change

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -174,4 +239,77 @@ public void onWallObjectSpawned(WallObjectSpawned event)
...
+
+	private void checkRunestoneDepletion()
+	{
+		if (client.getGameState() != GameState.LOGGED_IN || recentlyLoggedIn)
+		{
+			return;
+		}
+
+		final int region = client.getLocalPlayer().getWorldLocation().getRegionID();
+		runestones.forEach(r ->
```
deathbeam: It is probably not good idea to do this on every single varbit change

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -174,4 +239,77 @@ public void onWallObjectSpawned(WallObjectSpawned event)
...
+			case MINING_ADAMANT_PICKAXE:
+			case MINING_RUNE_PICKAXE:
+			case MINING_DRAGON_PICKAXE:
+			case MINING_DRAGON_PICKAXE_UPGRADED:
+			case MINING_DRAGON_PICKAXE_OR:
+			case MINING_INFERNAL_PICKAXE:
+			case MINING_3A_PICKAXE:
+			case MINING_CRYSTAL_PICKAXE:
+			case DENSE_ESSENCE_CHIPPING:
+				mining = true;
```
deathbeam: can't you simply make method that checks what is current player's animation instead of adding boolean and checking it in anim id? Would be simpler imo (e.g simple list.contains(playerAnimId) and make method isMining that does that)

Pinibot: Wouldn't that cause the logic to be performed unnecessarily a lot because isMining() is called from the Overlay class on the render thread meaning the animation is checked multiple times per second rather than just when it changes?

Happy to change this method to read from a valid list of animation ids instead of a switch, but think having the boolean is still the best idea unless i've misunderstood?

deathbeam: Changing it to boolean is so insignificat performance boost that it doesnt matter, and just doing contains on list is way cleaner and simpler.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -174,4 +246,75 @@ public void onWallObjectSpawned(WallObjectSpawned event)
...
+	}
+
+	private boolean addRockToRespawns(Rock rock, TileObject object, int region)
+	{
+		RockRespawn rockRespawn = new RockRespawn(rock, object.getWorldLocation(), Instant.now(),
+			(int) rock.getRespawnTime(region).toMillis(), rock.getZOffset());
+		return respawns.add(rockRespawn);
+	}
+
+	private void populateMiningAnimations()
```
deathbeam: just create static final immutable list

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -174,4 +246,75 @@ public void onWallObjectSpawned(WallObjectSpawned event)
 			}
 		}
 	}
+
+
+	protected boolean isMining()
```
deathbeam: you can also most likely just move this to overlay

runelite-client/src/main/java/net/runelite/client/plugins/mining/DenseRunestoneMineOverlay.java
```diff
@@ -43,6 +58,23 @@
 class DenseRunestoneMineOverlay extends Overlay
 {
 	private static final int MAX_DISTANCE_TO_DRAW_STONES = 2000;
+	private static final ImmutableList<Integer> miningAnimations = ImmutableList.of(
```
deathbeam: UPPER_CASE

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
 		recentlyLoggedIn = false;
 	}
 
+	@Subscribe
+	public void onGameObjectSpawned(GameObjectSpawned event)
+	{
+		GameObject object = event.getGameObject();
+
+		Rock rock = Rock.getRock(object.getId());
+		if (rock != null)
```
deathbeam: this null check is unnecessary

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
 		recentlyLoggedIn = false;
 	}
 
+	@Subscribe
+	public void onGameObjectSpawned(GameObjectSpawned event)
+	{
+		GameObject object = event.getGameObject();
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
...
+
+		Rock rock = Rock.getRock(object.getId());
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
...
+		if (rock != null)
+		{
+			if (rock == Rock.DENSE_RUNESTONE)
+			{
+				Varbits depletionVarbit = object.getId() == NULL_8981 ? Varbits.DENSE_RUNESTONE_NORTH_DEPLETED
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
...
+					: Varbits.DENSE_RUNESTONE_SOUTH_DEPLETED;
+				boolean isDepleted = client.getVar(depletionVarbit) == 1;
+
+				// Add the newly spawned Dense Runestone to the list
+				DenseRunestone denseRunestone = new DenseRunestone(isDepleted, object, depletionVarbit);
```
deathbeam: just do .add(new DenseRunestone)

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -174,4 +224,46 @@ public void onWallObjectSpawned(WallObjectSpawned event)
...
+		// Check to see if the player is in the same region as the dense runestones
+		final int region = client.getLocalPlayer().getWorldLocation().getRegionID();
+		if (region == DENSE_RUNESTONE_MINE_REGION_ID)
+		{
+			runestones.forEach(r ->
+			{
+				r.setDepleted(client.getVar(r.getDepletionVarbit()) == 1);
+				if (r.isDepleted())
+				{
+					Rock rock = Rock.getRock(r.getGameObject().getId());
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -127,8 +176,11 @@ public void onGameObjectDespawned(GameObjectDespawned event)
 		Rock rock = Rock.getRock(object.getId());
 		if (rock != null)
 		{
-			RockRespawn rockRespawn = new RockRespawn(rock, object.getWorldLocation(), Instant.now(), (int) rock.getRespawnTime(region).toMillis(), rock.getZOffset());
-			respawns.add(rockRespawn);
+			if (rock == Rock.DENSE_RUNESTONE)
+			{
+				runestones.removeIf(r -> r.getGameObject().getId() == object.getId());
+			}
+			addRockToRespawns(rock, object, region);
```
deathbeam: empty line above

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -113,6 +140,28 @@ public void onGameTick(GameTick gameTick)
...
+		GameObject object = event.getGameObject();
+
+		Rock rock = Rock.getRock(object.getId());
+		if (rock != null)
+		{
+			if (rock == Rock.DENSE_RUNESTONE)
+			{
+				Varbits depletionVarbit = object.getId() == NULL_8981 ? Varbits.DENSE_RUNESTONE_NORTH_DEPLETED
+					: Varbits.DENSE_RUNESTONE_SOUTH_DEPLETED;
+				boolean isDepleted = client.getVar(depletionVarbit) == 1;
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -123,13 +167,12 @@ public void onGameObjectDespawned(GameObjectDespawned event)
 
 		final GameObject object = event.getGameObject();
 		final int region = client.getLocalPlayer().getWorldLocation().getRegionID();
-
-		Rock rock = Rock.getRock(object.getId());
-		if (rock != null)
+		final Rock rock = Rock.getRock(object.getId());
+		if (rock == Rock.DENSE_RUNESTONE)
```
deathbeam: with null-check not being required I meant only for runestones, but for addRockToRespawns it is still required.

deathbeam:
`[ERROR] src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java:[175,19] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.`

dekvall:
I did some in-game testing with this and imo it feels weird that the clickboxes of both runestones are removed when mining either one of them. If you really want to remove clickboxes as an indication of not mining, i think that only the clickbox for the object the player is interacting with should be removed.

I would prefer if the mining session tracker from the motherlode plugin was moved to the mining plugin to get the same interface for mining like for fishing and woodcutting. Removing clickboxes feels inferior to the normal session tracker, but adding that is probably outside the scope of this PR. At this time, removing the clickbox indicator seems like the obvious solution. 

The respawn timer of the dense runestone is off as there seems to be some randomness to it. One solution is to handle it like the motherlode upper veins; use the worst case scenario time and remove the timer if the runestone respawns before that. Another is to use the average time, but i feel the former way is preferable. The maximum respawn time i encountered during testing was 19301 ms. I posted all respawn times below. My mining level is 80 but i don't find anything on the wiki which says mining level matters.
```
2019-11-06 19:47:35 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18780 ms
2019-11-06 19:48:19 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18099 ms
2019-11-06 19:48:50 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16725 ms
2019-11-06 19:51:14 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18757 ms
2019-11-06 19:51:40 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12699 ms
2019-11-06 19:52:08 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 13961 ms
2019-11-06 19:52:42 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12921 ms
2019-11-06 19:53:14 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 17405 ms
2019-11-06 19:54:04 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12641 ms
2019-11-06 19:55:13 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12641 ms
2019-11-06 19:58:19 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 17940 ms
2019-11-06 19:58:57 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12944 ms
2019-11-06 19:59:44 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18581 ms
2019-11-06 20:01:09 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 15888 ms
2019-11-06 20:02:56 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 19301 ms
2019-11-06 20:07:32 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 17078 ms
2019-11-06 20:13:36 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16901 ms
2019-11-06 20:14:41 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16841 ms
2019-11-06 20:16:05 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 13261 ms
2019-11-06 20:17:14 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 14116 ms
2019-11-06 20:18:21 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 17875 ms
2019-11-06 20:20:48 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16861 ms
2019-11-06 20:21:58 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16896 ms
2019-11-06 20:23:05 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18121 ms
2019-11-06 20:23:55 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 14953 ms
2019-11-06 20:25:29 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 14941 ms
2019-11-06 20:27:49 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 16288 ms
2019-11-06 20:29:00 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 12821 ms
2019-11-06 20:30:57 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 18640 ms
2019-11-06 20:32:20 [Client] INFO  n.r.c.plugins.mining.MiningPlugin - Respawn took 17518 ms
```

Adam-:
I agree it feels weird for it to not show the clickbox when mining, it isn't consistent with how the other resource overlays work. Also if the respawn times have up to 7 seconds of variance maybe it would be better to not show a timer at all?

dekvall:
I think timers would be nice and the motherlode vein respawn times seem to vary between 90 and 166 seconds so i think we can handle these similarily; by changing the color of the overlay when it is able to respawn.

Adam-:
I suppose thats fine. @Pinibot are you interested in making these changes?

Nightfirecat:
This PR needs rebasing and updates per the comments here. As it has not been updated in some months since the most recent feedback, I am closing it. Anyone else considering working on this feature should keep in mind the conversation which has been had here.

Pinibot, if you choose to continue working on this feature, please re-open this PR and push any needed changes.

PR #10155, opened 2019-10-29
Add layout chat command

Adam-:


Adam-:
The way the plugin is written makes it difficult to build a `Raid` in order to get the existing layout message

PR #10157, opened 2019-10-29
update ExaminePlugin to account for ground items

### LINKED ISSUE #6201: Examining ground items does not trigger the Examine plugin

Brodan:
Fixes #6201.

This PR updates a conditional in the `ExaminePlugin` so that price look-ups are triggered when examining items on the ground and a message is posted to chat, the same way it behaves when examining an inventory item.

The diff looks a lot bigger than it is because of indentation changes.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
@@ -126,6 +126,10 @@ public class ExaminePlugin extends Plugin
 				quantity = widgetItem != null && widgetItem.getId() >= 0 ? widgetItem.getQuantity() : 1;
 				break;
 			}
+			case EXAMINE_ITEM_GROUND:
+				type = ExamineType.ITEM;
+				id = event.getId();
+				break;
 			case EXAMINE_ITEM_BANK_EQ:
 			{
 				type = ExamineType.ITEM_BANK_EQ;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -252,67 +256,48 @@ public void onChatMessage(ChatMessage event)
 		int widgetChild = TO_CHILD(widgetId);
 		Widget widget = client.getWidget(widgetGroup, widgetChild);
 
-		if (widget == null)
-		{
-			return null;
-		}
+		if (widget != null) {
```
Nightfirecat: Why did you change the guard clause to a massive `if` block and un-do the formatting that was present here? The code here already conformed to our code style guidelines and should not be changed on a whim.

Nightfirecat: Revisiting this comment [from the original form of this PR](https://github.com/runelite/runelite/pull/6284/files#r229890170), `findItemFromWidget()` has only one entry point, and it is not reached when examining a ground item.

Brodan: Yea I was referencing my old PR so I could re-open this since so much has changed but it looks like all that needed to be added now is that new examine `case`. I'm trying to figure out what all this code is doing again, but this seems to be working locally just fine.

Brodan: Wow, yea, I think my old PR was just trying waaaay too hard to needlessly refactor code in addition to adding that feature. All this feature really needs is that extra case. Lemme know your thoughts!

PR #10169, opened 2019-10-31
Remove stretched mode plugin/api

Vilifi:
With Jagex's recent statement on third-party clients, they explicitly state the following:

`
Not Allowed - "Any movement or resizing of click zones for any interface or component under combat options, inventory, worn equipment, prayers or spell book"
`

Stretched mode, as its designed, resizes the entire UI to make everything significantly easier to click, as such it should be removed. I probably missed quite a few things here, apologies. This probably should be done by someone who has access to an injector, so they can remove the mixins as well.


dekvall:
Based on the amended [statement](https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1) from Jagex, all plugins that are currently in the client complies with the their 3pc rules. Stretched mode included.

Issue #10144, opened 2019-10-26
World Map: Split "Show quest names and status" option into two separate options


makingcomment:
Less selfish reason: Having this as single option is inconsistent with having split options for each of agility shortcuts and rare trees.
More selfish reason: I like the extra information provided by the quest names tooltips, but I'm not a fan of the quest status icons on the world map. Every other icon on the world map is circular, while the quest status icons aren't, which to me makes them a bit of an eyesore. And even beyond that, I like the default world map icons and don't like having to change them to gain the tooltips.

**Describe the solution you'd like**
Two options within the world map plugin: A "Show quest names" option to control the tooltip, and a "Show quest status" option to control the quest status icons.

**Describe alternatives you've considered**
- Change the quest status icons to be more in line with the all of the other icons' circular aesthetic (maybe do this regardless?).
- ~Move the quest status into the quest name tooltip.~ Upon further inspection, quest status information information is already in the tooltip, so this option would just be removing the quest status icons.

Darkkgreen:
I'll take a look into it later

Darkkgreen:
Would you take a look inside the pull request, please?

makingcomment:
Closed by game update

PR #10170, opened 2019-10-31
World Map - Split quest name and status from custom icons

Darkkgreen:
Transformed all the quest icons that weren't round, added a new config to turn off custom icons and made it work.

How it looks like:

QuestIcons

Should solve #10144


makingcomment:
In terms of function, this looks awesome. Thank you for working on this!

(no comment on the code/implementation)

raiyni:
I don't think it makes sense to have a toggle just to show the icons and tooltips separately. Plus these icons make it worse because you can't even see that it's a quest.

makingcomment:
> I don't think it makes sense to have a toggle just to show the icons and tooltips separately. Plus these icons make it worse because you can't even see that it's a quest.

Please see #10144

Darkkgreen:
TBF I guess I forgot to change the size, will do this later today.

@raiyni The overall idea is to discuss, that's why I made this pull request. Keeping all icons round made sense to me. IIRC this is the only icon that is overridden with those flags, making it easier to remember and you can toggle it off anyway.

I could try to make it more readable but I guess it will never be perfect since the icon is really small

Nightfirecat:
I don't have much of an opinion on splitting the tooltip toggle from the quest progress icons, but the proposed new icons are much worse, to the point of being unrecognizable as quest icons in some cases.

PR #10184, opened 2019-11-01
Make God Wars Dungeon killcount movable

### LINKED ISSUE #10162: Add the ability to hide or move the Godwars Killcount
Would it be possible to add the ability to hide or move the NPC killcount from the upper right? 

Thanks in advance!

PERSS0N:
closes #10162 
Add God Wars Dungeon killcount movable


deathbeam:
Does this really need all 4 widgets? And do you have screenshot of how the interface looks like when made movable compared to vanilla?

deathbeam:
Implemented in https://github.com/runelite/runelite/commit/4beb8766abc218152b78977bd5f72bb5de4216d8

PR #10196, opened 2019-11-04
Add document update event to search

VarezF:
https://github.com/runelite/runelite/issues/9545

Update on edit of search bar


deathbeam:
Maybe this could be moved to searchbar class and then maybe some custom method (like setOnSearchListener or something) and then update other search bar usages to use that method.

Issue #9974, opened 2019-10-04
Bind anti-drag to other keys


justhereforsuggestion:
I love the anti-drag feature, as it helped me a lot since I suffer from tremors in my hand with any press I make. However, the anti-drag is exclusively tied to shift at the moment.

I don't know how to make my own anti-drag with AHK, so I was wondering if you guys knew or if you guys could allow us to change the hotkey for anti-drag in RL.

**Additional context**
This will be very helpful to people who suffer from tremors and make switching gear much less stressful and in line with those who do not suffer from this.

dekvall:
This has been discussed extensively in #3200 and while i sympathize with your situation, this is currently not something we are looking to implement.

justhereforsuggestion:
Aww, that's a shame. I see that it has been more than a year ago since the controversy but if that's still your stance, so be it.

Kamielvf:
Closing, relevant discussion in #3200

Castro3halo:
Given that it has been nearing on half a decade since this was initially suggested and a lot of jagex policy has changed on this over the years. Any chance this could be reviewed ?

Issue #3200, opened 2018-05-24
Feature Request: Anti-Drag


anon199469:
Is it possible to have anti drag back to how it used to be? I’d rather not have to use AHK like everyone else who used anti drag without shift. We shouldn’t cater to streamers. If there is a big enough issue I’m sure JaGex will reach out. This is honestly just QoL

Adam-:
We are not comfortable pushing this any further than OSB has at this time.

jbaker5011:
@anon199469 you can still emulate how it used to work with Windows by enabling Sticky Keys and then pressing shift

anon199469:
I use shift to drop stuff. High level PvPers already use software for this and Jagex has said it is fine. I just don’t like the streamer favoritism. It wasn’t an issue until Skill Specs complained

matthewpetch:
> Jagex has said it is fine

Where?

forsco:
I'm not sure how I stand with this looking from both sides.

If people are just going to work around the shift anti-drag albeit by sticky keys or AHK is there much of a point changing it to this?

BigMemeYek:
Pkers already have their own version bring it back 😡

anon199469:
JaGex fixes the drag issue closer to EoC. They implemented it into their game because hey saw it was a QoL update. It should be the same here.

joshbz93:
Revert please.  Shift is used to drop items.  At least give an option to use a different key.

iguaan:
It's not about if we consider this balaned, I don't think anyone here at least is arguing that this is way too overpowered, it's about doing what has been already done for a while. In RL (and also OSB) you can just keep using this feature the same way as it was 'pre-nerf' if you disable shift-drop and use sticky keys to stuck a shift key. This is an easy way of covering RL in case someone starts whining about the feature being too overpowered.

SoyChai:
Following behind OSBuddy when it comes to potentially advantage-giving features is normal practice for us, since their company has much closer communication with Jagex than we do. It's unfortunate that we didn't make this decision *before* releasing the Anti-Drag plugin, that was obviously a mistake; but we have to stick to our guns now.

Big sorry to anyone who was starting to get comfortable with it! :(

joshbz93:
Feature that tell you what bosses are in a raid: OK (although literally cheating in my opinion)
Feature that guides you through boss fights: OK.  
Feature that guides you through clue scrolls: OK.  
Feature that keeps you from accidentally dragging an item: Over Powered. 
[...]

PR #10209, opened 2019-11-05
Allowing anti-drag key to be configurable

Pinibot:
Currently the plugin only supports the shift key for increasing the drag delay but this isn't ideal as shift is often bound to "shift to drop", meaning that items can be accidently dropped if this option is enabled.

I've added a new ModiferlessKeybind config option to allow the user to choose any key that they wish to be used in place of shift.


Nightfirecat:
Based on discussion in #3200 and #9974, I do not think we will be accepting this change.

deathbeam:
Yeah, we rejected this before so closing, sorry

PR #10219, opened 2019-11-07
Add adventure log plugin

Trevor159:
WIP PR for the adventure log plugin. I have completed everything I wanted to and now I just need feedback and reviews.

I have not implemented the raids data yet because #7981 and #8122 would make it easy and consistent.

<details>
<summary>Feedback that I have already received and have not implemented (That I remember)</summary>

`The validateObject doesn't need to be as complex as it is` -Adam

my response: hmm, adam why did you want the adventurelog json validator to be less complex than the config one? I looked through everything that is being checked and it all seems valid and could be made more strict if we wanted. I made the max value length way less because the longest data in the adventure log is probably the username hash which is 128 chars.

`for the json validation, why not simply do something similar like what we do atm with web socket messages` -DB

abex's response: I think adam didn't want the types in the http-api

</details>

This google doc has more details on the specific events and such
https://docs.google.com/document/d/1phIAZOyr5onOqGK170DjOF3ebb0q6YTcstvFOQBeKqg


http-api/src/main/java/net/runelite/http/api/adventurelog/LogType.java
```diff
@@ -0,0 +1,39 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.http.api.adventurelog;
+
+public enum LogType
```
deathbeam: Why? Can't you simply do same thing like websocket messages are doing atm and simply serialize classname to _type field or something on logdata

http-service/src/main/java/net/runelite/http/service/adventurelog/AdventureLogController.java
```diff
@@ -0,0 +1,89 @@
...
+
+		return adventureLogService.getLogs(session.getUser(), username);
+	}
+
+	@RequestMapping(method = POST)
+	public void addLog(
+		HttpServletRequest request,
+		HttpServletResponse response,
+		@RequestParam String username,
+		@RequestBody Object values
```
deathbeam: http service has dependency on http api, what is the point using Object as body when you can simply use AdventureLogRecord collection?

runelite-client/src/main/java/net/runelite/client/game/LevelUpManager.java
```diff
@@ -0,0 +1,90 @@
...
+import net.runelite.api.events.StatChanged;
+import net.runelite.client.eventbus.EventBus;
+import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.events.LevelUp;
+
+@Singleton
+public class LevelUpManager
+{
+	@Inject
+	private EventBus eventBus;
```
deathbeam: Constructor injection etc etc

runelite-client/src/main/java/net/runelite/client/plugins/adventurelog/AdventureLogConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+@ConfigGroup("adventurelog")
+public interface AdventureLogConfig extends Config
+{
+	@ConfigItem(
+		keyName = "raids",
+		name = "Raids",
+		description = "enable raids"
+	)
+	default boolean enableRaids()
```
deathbeam: I guess this is WIP but why is here checkbox only for raids and not other stuff

http-api/src/main/java/net/runelite/http/api/adventurelog/AdventureLogClient.java
```diff
@@ -0,0 +1,86 @@
...
+import okhttp3.Response;
+
+@Slf4j
+@AllArgsConstructor
+public class AdventureLogClient
+{
+	private static final MediaType JSON = MediaType.parse("application/json");
+	private static final Gson GSON = RuneLiteAPI.GSON;
+
+	private final UUID uuid;
```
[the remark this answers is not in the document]
Trevor159: This is how it is done on both the loottracker and config client's also. I previously stored the hashed username in here which is probably what you remember.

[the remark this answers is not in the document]
deathbeam: Oh, you are right

http-service/src/main/java/net/runelite/http/service/adventurelog/AdventureLogController.java
Trevor159: Mongo doesn't seem to know how to parse POJOs like logData. It parses general Objects fine so I just kept it as an Object.

deathbeam: Then just serialize it by yourself

http-api/src/main/java/net/runelite/http/api/adventurelog/AdventureLogRecord.java
```diff
@@ -0,0 +1,46 @@
...
+@NoArgsConstructor
+public class AdventureLogRecord
+{
+	private String logType;
+	private LogData logData;
+	private EnumSet<WorldType> worldTypes;
+
+	public AdventureLogRecord(LogData logData, EnumSet<WorldType> worldTypes)
+	{
+		this.logType = logData.getClass().getSimpleName();
```
deathbeam: I meant custom type adapter for GSON, but I guess this works too. logType should be final then

runelite-client/src/main/java/net/runelite/client/events/AdventureLogSubmission.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package net.runelite.client.events;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import net.runelite.http.api.adventurelog.LogData;
+
+@Data
+@AllArgsConstructor
+public class AdventureLogSubmission
```
deathbeam: Now there isnt really a point in have this class at all. or well, having both LogData and AdventureLogSubmission, I would just make this interface and implement it to all other stuff you are submitting (so basically delete LogData and use ALSubmission everywhere where you are using LogData atm)

Trevor159: I can't make it an interface due to the check here https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java#L131 but what I have done probably satisfies you.

runelite-client/src/main/java/net/runelite/client/plugins/adventurelog/AdventureLogPlugin.java
```diff
@@ -0,0 +1,729 @@
...
+		checkBarbAssault();
+	}
+
+	@Subscribe
+	public void onChatMessage(ChatMessage event)
+	{
+		String message = Text.removeTags(event.getMessage());
+		if (message.contains(PET_FOLLOWING_MESSAGE))
+		{
+			shouldCheckPetFollowing = true;
```
dekvall: You can probably just check the 16 msb in varp 447 for the follower npcId here instead which would let you skip the npcSpawned check later. Might be more relieable if the npcSpawned event would ever fire before this, but idk if it ever does. 
```
int petID = (client.getVar(VarPlayer.FOLLOWER) >> 16) & 0xffff;
```

Nightfirecat:
I think this would be better-suited as a hub plugin now?

Trevor159:
> I think this would be better-suited as a hub plugin now?

Probably, I also don't have any interest in this myself anymore so if anyone wants to yoink it feel free.

PR #3076, opened 2018-05-21
Custom notifications

Abextm:
This is enabled by default on windows, because windows' notifications are broken for a large amount of users.

Closes #2060
Closes #920
Supersedes #2230

PR #10237, opened 2019-11-11
Add custom notification sounds

### LINKED ISSUE #920: Allow custom sound notifcations
Different sound notifications other than windows notifiactions (runeloader does it perfect by adding almost a bell sound or a 'ting' that makes you always hear it during movies/games/shows)

Trevor159:
Yoinked #3076 and stripped it to only audio.

Closes #920


runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -154,14 +155,14 @@ default boolean requestFocusOnNotification()
 	}
 
 	@ConfigItem(
-		keyName = "notificationSound",
-		name = "Enable sound on notifications",
+		keyName = "notificationSound_",
```
Adam-: Please don't do this

PR #10238, opened 2019-11-11
Make GWD overlay movable

### LINKED ISSUE #10162: Add the ability to hide or move the Godwars Killcount

winterdaze:
Closes #10162 
Supersedes #10184 

https://gyazo.com/653f434ef9ea0b9578f65f8484910580


deathbeam:
This looks way better, I wonder why the original PR did what it did..

PR #10239, opened 2019-11-12
Add multiple FPS targets for multiple modes

### LINKED ISSUE #10226: FPS Control: Independent Mode FPS Settings
Right now, there is only a way to set one FPS target and choose when that target is activated during one of three modes:

1. Never
2. Unfocused
3. Always

This is annoying because there is no way to set an fps target for when the player is gaming and an independent fps target for when the client is in the background.

A separate input box for target FPS for each mode instead of a dropdown for the mode and 1 single box for target FPS.

al3x-huang:
closes #10226, I have made the following changes:

- Added separate input boxes for global FPS target and unfocused FPS target. 
- Added separate enable toggles for global FPS target and unfocused FPS target.
- Updated the logic in `FpsDrawListener` to take into account of the new settings. 

The global FPS target, when activated, will limit the applet's FPS while focused and unfocused. Players can choose to further limit unfocused FPS by enabling `Limit FPS unfocused` and setting the desired target. I don't enforce that this limit must be strictly lower than the global FPS even though I would assume this to be the common use case. 

The changes made above still preserves all the functionalities of the UI before, where a dropdown menu selected for a single FPS target enforcement in one of two modes (`always` and `unfocused`) or no target enforcement in `never` mode. Users can still get legacy behavior with the added benefit that they can now set two respective targets, one for always and one specifically for unfocused.


runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsDrawListener.java
```diff
@@ -79,18 +87,18 @@ void reloadConfig()
 	void onFocusChanged(FocusChanged event)
 	{
 		this.isFocused = event.isFocused();
+		reloadConfig();  //load new delay
```
deathbeam: Add space after `//`

runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsConfig.java
```diff
@@ -32,32 +32,58 @@
...
+		name = "Limit Global FPS",
+		description = "Global FPS limit in effect regardless of<br>" +
+			"whether window is in focus or not",
 		position = 1
 	)
-	default FpsLimitMode limitMode()
+	default boolean enableFps()
 	{
-		return FpsLimitMode.NEVER;
+		return true;
```
deathbeam: This should be disabled by default to follow previous behaviour

al3x-huang: It's interesting to me that the default behavior before is "never" by default since users are enabling the plugin to get an FPS limit. It is counterintuitive to need to enable it twice but I will make the change.

deathbeam: Well the plugin is also used for drawing FPS what is atm its only feature enabled by default

al3x-huang: Ah good point, makes sense.

runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsDrawListener.java
```diff
@@ -79,18 +87,18 @@ void reloadConfig()
 	void onFocusChanged(FocusChanged event)
 	{
 		this.isFocused = event.isFocused();
+		reloadConfig();  // load new delay
```
deathbeam: only 1 space before //

runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsDrawListener.java
```diff
@@ -67,7 +67,15 @@ private FpsDrawListener(FpsConfig config)
...
-		targetDelay = 1000 / Math.max(1, config.maxFps());
+
+		if (config.enableFpsUnfocused() && !isFocused)
+		{
+			targetDelay = 1000 / Math.max(1, config.maxFpsUnfocused());
+		}
+		else
+		{
+			targetDelay = 1000 / Math.max(1, config.maxFps());
+		}
```
deathbeam: add empty line below

runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsConfig.java
```diff
@@ -32,32 +32,58 @@
...
 	)
 	default int maxFps()
 	{
 		return 50;
 	}
 
+	@ConfigItem(
+		keyName = "limitFpsUnfocused",
+		name = "Limit FPS unfocused",
+		description = "FPS limit while window is out of focus<br>" +
```
Adam-: I think `FPS limit while window is out of focus` is sufficient here

runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsDrawListener.java
```diff
@@ -67,7 +67,16 @@ private FpsDrawListener(FpsConfig config)
 	void reloadConfig()
 	{
 		lastMillis = System.currentTimeMillis();
-		targetDelay = 1000 / Math.max(1, config.maxFps());
+
```
Adam-: This can just be:
```java
int fps = config.enableFpsUnfocused() && !isFocused
  ? config.maxFpsUnfocused()
  : config.maxFps();
targetDelay = 1000 / Math.max(1, fps);
```

al3x-huang:
Can someone merge this?

PR #10247, opened 2019-11-13
Motherlode Scaling & Pay-dirt Skill Calculator

AdamScoggins:
Added in a scaling option for the Mining Skill Icon in MLM
Rocks on other levels do not show anymore
Added Pay-dirt into the Skill Calculator


runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeRocksOverlay.java
```diff
@@ -26,12 +26,10 @@
  */
 package net.runelite.client.plugins.motherlode;
 
-import java.awt.Color;
-import java.awt.Dimension;
-import java.awt.Graphics2D;
-import java.awt.Polygon;
+import java.awt.*;
```
JZomerlei: Avoid wild char imports.

PR #10249, opened 2019-11-13
Paydirt skilltracker

Jimpix-svg:
Implement pay-dirt into the skill calculator.


Jimpix-svg:
Duplicate commits with #10247

PR #10251, opened 2019-11-13
Implement loot tracking for MLM

Jimpix-svg:


Adam-:
IMO this seems out of scope for the loottracker since this is essentially a skilling resource and isn't really "loot".

Jimpix-svg:
> 
> 
> IMO this seems out of scope for the loottracker since this is essentially a skilling resource and isn't really "loot".

There isn't currently a way to track the ores received in terms of currency. How about adding the profit into the MLM Plugin Overlay?

Adam-:
There already is an overlay which shows what ores you found. I guess you could add price to that.

Nightfirecat:
Per Adam's comment, this can be implemented as a price line added to the MLM ores overlay.

PR #10258, opened 2019-11-14
itemstats: Add 'Always Show Base Stats' & 'Show Stats In Bank' 

BenDol:
Items already equipped or items that are the same should still show their stats.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
@@ -90,9 +90,13 @@ public class ItemStatOverlay extends Overlay
 		final int child = WidgetInfo.TO_CHILD(entry.getParam1());
 		final Widget widget = client.getWidget(group, child);
 
-		if (widget == null || (group != WidgetInfo.INVENTORY.getGroupId() &&
-			group != WidgetInfo.EQUIPMENT.getGroupId() &&
-			group != WidgetInfo.EQUIPMENT_INVENTORY_ITEMS_CONTAINER.getGroupId()))
+		if (widget == null
+			|| !(group == WidgetInfo.INVENTORY.getGroupId()
+				|| group == WidgetInfo.EQUIPMENT.getGroupId()
+				|| group == WidgetInfo.EQUIPMENT_INVENTORY_ITEMS_CONTAINER.getGroupId()
+				|| (config.showStatsInBank()
+					&& (group == WidgetInfo.BANK_ITEM_CONTAINER.getGroupId()
+						|| group == WidgetInfo.BANK_INVENTORY_ITEMS_CONTAINER.getGroupId()))))
 		{
 			return null;
 		}
@@ -107,12 +111,18 @@ public class ItemStatOverlay extends Overlay
 				itemId = widgetItem.getItemId();
 			}
 		}
-		else if (group == WidgetInfo.EQUIPMENT_INVENTORY_ITEMS_CONTAINER.getGroupId())
+		else if (group == WidgetInfo.EQUIPMENT_INVENTORY_ITEMS_CONTAINER.getGroupId()
+			|| group == WidgetInfo.BANK_ITEM_CONTAINER.getGroupId()
+			|| group == WidgetInfo.BANK_INVENTORY_ITEMS_CONTAINER.getGroupId())
 		{
-			final Widget widgetItem = widget.getChild(entry.getParam0());
-			if (widgetItem != null)
+			int index = entry.getParam0();
+			if (index > -1)
 			{
-				itemId = widgetItem.getItemId();
+				final Widget widgetItem = widget.getChild(index);
+				if (widgetItem != null)
+				{
+					itemId = widgetItem.getItemId();
+				}
 			}
 		}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
@@ -42,6 +42,7 @@ import net.runelite.client.ui.overlay.Overlay;
 import net.runelite.client.ui.overlay.tooltip.Tooltip;
 import net.runelite.client.ui.overlay.tooltip.TooltipManager;
 import net.runelite.client.util.ColorUtil;
+import net.runelite.client.util.QuantityFormatter;
 import net.runelite.http.api.item.ItemEquipmentStats;
 import net.runelite.http.api.item.ItemStats;
 
@@ -182,7 +183,7 @@ public class ItemStatOverlay extends Overlay
 
 		final String prefix = value > 0 ? "+" : "";
 		final String suffix = showPercent ? "%" : "";
-		final String valueString = (int)value == value ? String.valueOf((int)value) : String.valueOf(value);
+		final String valueString = QuantityFormatter.formatNumber(value);
 		return label + ": " + ColorUtil.wrapWithColorTag(prefix + valueString + suffix, color) + "</br>";
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
@@ -167,7 +167,6 @@ public class ItemStatOverlay extends Overlay
 	}
 
 	private String getChangeString(
-		final String label,
 		final double value,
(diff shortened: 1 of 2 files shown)


deathbeam:
Showing original stats make no sense because item stats is showing difference, so when equipping same item correct value to show would be nothing because nothing changed.

And when unequipping item, correct value to show is stat change after going from that item to nothing and not its stats.

BenDol:
I want to be able to see the stats of an item weather or not it has a change. As for removing an item it seems somewhat obsolete to show the stats as a subtraction when all you'd want to see is the items stats. I have to unequip my items to see the items stats which is silly imo.

deathbeam:
Its not, its how it works, its diffs and not item stats. If you want to always see stats and not see diffs, then implement it as option.

BenDol:
It's literally called item stats plugin, not item stat difference plugin. It doesn't take away from the differentiation aspect of the plugin to see the actual stats on items.

To be honest it should show the stats of the item and then a +/- value next to it in brackets if it is suppose to be an item stat plugin (how every other game shows stats on items). Should this be implemented as a separate setting, then?

deathbeam:
I would prefer to have current behaviour stay as it is and if you want to add anything extra like what you just described, add option.

BenDol:
Updated now with the setting `Always Show Base Stats`

image

image

Works exactly as it did before with this setting off.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -183,25 +190,54 @@ private String getChangeString(
...
+		final boolean inverse,
+		final boolean showPercent,
+		final boolean dontShowBase)
+	{
+		StringBuilder b = new StringBuilder();
+		if (value != 0 || diffValue != 0)
 		{
-			b.append(getChangeString("Weight", s.getWeight(), true, false));
+			String changeStr = getChangeString(diffValue, inverse, showPercent);
+			if (config.alwaysShowBaseStats() && !dontShowBase)
```
deathbeam: Only use for dontShowBase is weight, and you are negating it there as well, wouldnt it make more sense to just change the param to showBase and remove the negations?

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -183,25 +190,54 @@ private String getChangeString(
...
+		final double diffValue,
+		final boolean inverse,
+		final boolean showPercent,
+		final boolean dontShowBase)
+	{
+		StringBuilder b = new StringBuilder();
+		if (value != 0 || diffValue != 0)
 		{
-			b.append(getChangeString("Weight", s.getWeight(), true, false));
+			String changeStr = getChangeString(diffValue, inverse, showPercent);
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
deathbeam: Also empty line above

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -222,32 +258,43 @@ private String buildStatBonusString(ItemStats s)
 		final ItemStats subtracted = s.subtract(other);
 		final ItemEquipmentStats e = subtracted.getEquipment();
 
-		if (subtracted.isEquipable() && e != null)
+		final StringBuilder b = new StringBuilder();
+		if (config.showWeight())
```
deathbeam: empty line above

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -222,32 +258,43 @@ private String buildStatBonusString(ItemStats s)
...
-			if (e.getAstab() != 0 || e.getAslash() != 0 || e.getAcrush() != 0 || e.getAmagic() != 0 || e.getArange() != 0)
+		if (subtracted.isEquipable() && e != null && ce != null)
+		{
+			b.append(buildStatRow("Prayer", ce.getPrayer(), e.getPrayer(), false, false));
+			b.append(buildStatRow("Speed", ce.getAspeed(), e.getAspeed(), true, false));
+			b.append(buildStatRow("Melee Str", ce.getStr(), e.getStr(), false, false));
+			b.append(buildStatRow("Range Str", ce.getRstr(), e.getRstr(), false, false));
+			b.append(buildStatRow("Magic Dmg", ce.getMdmg(), e.getMdmg(), false, true));
+
+			StringBuilder abb = new StringBuilder();
```
deathbeam: final

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -222,32 +258,43 @@ private String buildStatBonusString(ItemStats s)
...
-				b.append(getChangeString("Stab", e.getAstab(), false, false));
-				b.append(getChangeString("Slash", e.getAslash(), false, false));
-				b.append(getChangeString("Crush", e.getAcrush(), false, false));
-				b.append(getChangeString("Magic", e.getAmagic(), false, false));
-				b.append(getChangeString("Range", e.getArange(), false, false));
+				b.append(ColorUtil.wrapWithColorTag("Attack Bonus</br>", JagexColors.MENU_TARGET)).append(abb);
 			}
 
-			if (e.getDstab() != 0 || e.getDslash() != 0 || e.getDcrush() != 0 || e.getDmagic() != 0 || e.getDrange() != 0)
+			StringBuilder dbb = new StringBuilder();
```
deathbeam: final

http-api/src/main/java/net/runelite/http/api/item/ItemStats.java
```diff
@@ -42,10 +42,9 @@ public ItemStats subtract(ItemStats other)
 			return this;
 		}
 
-		final double newWeight = weight - other.weight;
+		final double newWeight = Math.round((weight - other.weight) * 10000) / 10000d;
```
deathbeam: Why did you changed this?

BenDol: so it doesn't show double stats as `3.24522222222222` as I was experiencing

Hydrox6: Why not format the number to the desired number of decimals when you convert it into a string, like how `QuantityFormatter#PRECISE_DECIMAL_FORMATTER` works

BenDol: to what benefit? The value needs to be a double so it would just need to be converted back to a double

deathbeam: Because you just want to display it in different format, and changing the value in core api is bad way to do it.

BenDol: to be clear it's only the new weight that has this format issue, when you subtract the other weight from it that is when the large double decimal format occurs. So it seems to me that this is the appropriate place for this given the format is from the api

BenDol: I will revert and update to use the QuantityFormatter, but not sure I see the issue here personally. I could be missing something though.

BenDol:
Was there anything else required for this change?

deathbeam:
Nope, looks good. but the build is failing for some reason

BenDol:
Yeah it didn't seem like it was because of these changes when I checked it. But I might have missed something in the logs.

```
test(net.runelite.client.rs.ClientConfigLoaderTest)  Time elapsed: 10.172 sec  <<< ERROR!
java.net.SocketTimeoutException: connect timed out
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
```

dekvall:
Try triggering a rebuild, that test was failing when Jagex was having connection issues, but the test is fixed now.

PR #10270, opened 2019-11-15
Reclassify teleport sfx as other players' area effects instead of environmental area effects

### LINKED ISSUE #10248: Music plugin: Add teleport sounds to the "other players' area sounds" filter
Judging by #10190 it will need some sort of special case since the source is `null` for teleport sounds.

GiantFrog:
resolves #10248

This will also mute teleports cast by the player, as there is no simple way to tell who created the sound without a source. Note that using a mounted amulet of glory still makes the teleport sounds, even if every possible selection is muted.

Perhaps you'd rather have teleports be environmental sfx anyway? This seemed like a better place to put them, though.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
@@ -564,7 +564,7 @@ public class MusicPlugin extends Plugin
 	public void onAreaSoundEffectPlayed(AreaSoundEffectPlayed areaSoundEffectPlayed)
 	{
 		Actor source = areaSoundEffectPlayed.getSource();
-		int soundID = areaSoundEffectPlayed.getSoundId();
+		int soundId = areaSoundEffectPlayed.getSoundId();
 		if (source == client.getLocalPlayer()
 			&& musicConfig.muteOwnAreaSounds())
 		{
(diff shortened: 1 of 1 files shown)


Adam-:
`200` should be added to `SoundEffectID`. Additionally I expect the list of sourceless player-sounds sounds to grow a lot, so maybe make an immutableset for them?

GiantFrog:
Alright, now there's an immutable set in MusicPlugin. We can easily slap some more sound effect IDs in there as they are discovered!

runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
```diff
@@ -558,13 +564,14 @@ public void onScriptCallbackEvent(ScriptCallbackEvent ev)
 	public void onAreaSoundEffectPlayed(AreaSoundEffectPlayed areaSoundEffectPlayed)
 	{
 		Actor source = areaSoundEffectPlayed.getSource();
+		int soundID = areaSoundEffectPlayed.getSoundId();
```
Adam-:
```suggestion
		int soundId = areaSoundEffectPlayed.getSoundId();
```

GiantFrog:
Very good point, hold on. I want to check that a source is actually null before muting everything in that list, not just assume.

PR #10271, opened 2019-11-15
skillcalc: Combined bow halves and added full bow value for Fletching calc.

jacobkeith:


dekvall:
Thanks for your pr!

Can you further elaborate some on why you prefer this over the current way where you can just shft+click select both actions? Personally i would argue that with 3 actions per bow, where the experience gained from one is the sum of the two others, will make the skillcalc ui even more bloated than it already is.

I also think that most people don't fletch complete bows, they usually either just string or cut them right? The only reason i can think of is if you want to find out how many actions it would take without clicking on the action, in which case i think that the two extra clicks in exchange for not having one extra action per bow, is worth it.

Adam-:
Why are you making this change? Also, I think we should keep the `(u)` naming consistent with the game.

jacobkeith:
Having a full bow value would be helpful instead of just summing up two values using the selection mode. Half can be changed back to u though.

PR #10286, opened 2019-11-17
inventorygrid: re-add a delay to showing the overlay

### LINKED ISSUE #10269: Re-add the 'inventory grid' delay option
There used to be an option to set the delay for the inventory grid. Specifically, it was how long you could move an item in ms before the inventory grid appears. 

I always accidentally move items just barely when shift-dropping large inventories or performing other quick inventory activities and the 'strobe effect' created by the inventory grid is extremely off putting. I loved the feature that allowed me to delay it's appearance.

dekvall:
If a player for example tries to drop an entire inventory quickly with the overlay turned on, the inventory grid will flash for each small mouse movement if the left mouse key is held down. This change mitigates the issue by adding a delay option that controls the amount of time the user has to wait after pressing it before enabling any overlay functionality. Showing the grid still requires the mouse to be moved ~5px away from the first clicked position.

This delay just delays the activation of the overlay, as the the delay countdown is started upon item press rather than the exit this ~5px range. Making it respect the exit is needlessly complex as desired functionality is reached with the current changeset.

Closes #10269

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.inventorygrid;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Units;
 
 @ConfigGroup("inventorygrid")
 public interface InventoryGridConfig extends Config
@@ -59,5 +60,16 @@ public interface InventoryGridConfig extends Config
 	default boolean showHighlight()
 	{
 		return true;
+	}
+
+	@ConfigItem(
+		keyName = "dragDelay",
+		name = "Drag delay",
+		description = "Time to wait after an item press before the overlay is enabled"
+	)
+	@Units(Units.MILLISECONDS)
+	default int dragDelay()
+	{
+		return 0;
 	}
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
```diff
@@ -60,4 +60,14 @@ default boolean showHighlight()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "overlayDelay",
```
Adam-: should this use the old config key?

dekvall: I don't have a strong opinion about it, i just figured that the behavior for this config was sufficiently different than that of the old one. If you want to we can just use the old one.

Adam-: It looks to me like the behavior isn't different though?

dekvall: I guess it depands on how you see it, for me the old delay was used by the plugin to determine if you were dragging an item. Now, the dragging check is done by checking the amount of pixels dragged. And the new delay is used to determine how much time has to pass before showing any overlay. But i'll have to rebase anyway, so i'll just change it back.

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
```diff
@@ -60,4 +60,14 @@ default boolean showHighlight()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "dragDelay",
+		name = "Drag delay",
+		description = "Time in ms to wait after item press before the overlay is enabled"
+	)
+	default int dragDelay()
```
Nightfirecat: Now that #10662 has been merged, this should use the "ms" units annotation.

PR #10292, opened 2019-11-18
Raids: rotations whitelist fix

Alexsuperfly:
fce1b9518d4665e0a24996fc30481730beff63db caused configs with a space between boss names to not evaluate as equal when comparing them to the scouted rotation.

This pr culls that whitespace and any quotation marks, which during testing appeared around the boss names for me after a client restart.

Also comes with some fancy new tests to attempt to not have it break like that again.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.raids;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.inject.Binder;
 import com.google.inject.Provides;
@@ -40,7 +41,9 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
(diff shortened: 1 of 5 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/raids/RaidsPluginTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RaidsPluginTest
+{
+	private static final String EMPTY = "";
```
Nightfirecat: Since this is such a simple value, I don't think it's worth using a constant to represent it.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -416,12 +422,7 @@ private void updateList(ArrayList<String> list, String input)
 			Matcher m = ROTATION_REGEX.matcher(input);
 			while (m.find())
 			{
-				String rotation = m.group(1).toLowerCase();
-
-				if (!list.contains(rotation))
-				{
-					list.add(rotation);
-				}
+					list.add(m.group(1).toLowerCase().replaceAll("\\s|\"", ""));
```
Nightfirecat: De-indent by one tab.
```suggestion
				list.add(m.group(1).toLowerCase().replaceAll("\\s|\"", ""));
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -407,7 +408,7 @@ private void updateLists()
 		updateList(layoutWhitelist, config.whitelistedLayouts());
 	}
 
-	private void updateList(ArrayList<String> list, String input)
+	private void updateList(Set<String> list, String input)
```
Nightfirecat: Is there any reason the passed `list` should be a list or a set? This would be better accepting a `Collection` instead.

Alexsuperfly:
added in 0e920bdd98f4b8141fea652deaa7bb7b1f726a6a

PR #10299, opened 2019-11-20
Update ChatHistoryPlugin welcome message string to include TL

### LINKED ISSUE #10297: Chat History plugin does not work in Twisted League worlds
The plugin currently works be hooking onto the message that appears when first logging in but on twisted league worlds its different.

The message is `Welcome to Old School Runescape: Twisted League!`

sethtroll:
Normal world: `Welcome to Old School RuneScape.`
Twisted league: `Welcome to Old School Runescape: Twisted League!`

fixes #10297


ddaltonzkd:
If this is still available, I would like to contribute.

Adam-:
There is not much more to address on this issue, since this PR fixes it.

PR #10300, opened 2019-11-20
Update cryptic clue message to include bringing a sled

### LINKED ISSUE #10287: Add "sled" to required item for Trollweiss elite clue step
Can't slide down this mountain without a sled :( 
image

sethtroll:
fixes #10287


dekvall:
I listed cryptic clues that could benefit from an item recommendation in https://github.com/runelite/runelite/pull/9628#issuecomment-520750625 but there's not a significant amount of them, even fewer which actually require a specific item to be able to be completed. So imo just a note in the description like this seems fine.

PR #10305, opened 2019-11-22
League Chat Icons

hsamoht:
This plugin changes the Ironman icon to a League icon on all Chat Messages sent from a League world. You are now able to tell who is playing on League without confusing them as a main game Ironman if they are talking in clan chat for example. The plugin will also change the Ironman icon in the input area of the Chat Widget if you are on a League world.

Jagex have [commented ](https://old.reddit.com/r/2007scape/comments/dzhony/twisted_league_stats_and_changes/f87jjk5/?st=k39l4qb1&sh=4ea1118c) on the issue saying they would like to do this. They are however unable / unlikely to do it anytime soon due to engine work.

The plugin has no configs and only requires a toggle on/off to change all icons.

- **Chatbox input:** Icon is replaced if you are on a League world.
- **Public Chat:** Icon is replaced if you are on a League world.
- **Clan Chat:** Icon is replaced if message is sent from a League world.
- **Private Chat:** Icon is replaced if message is sent from a League world. Requires either:
   1) Sender is in the same Clan Chat
   2) Sender is on your Friends list

Example of Public Chat on a League world:
public chat example

Example of Clan Chat from a League world:
clan chat example

Example of Private message from League world:
private message example

_Edited: Changed PR description to match the current state of the PR. View earlier edits to see starting point._


hsamoht:
**Player chatbox** icon is in direct conflict with **Key remapper** (wasd) plugin. Both tries to change / reset name and text in the input chatbox. Doesn't break the client, but one will win out in the race condition. Not sure how to fix without changing both or having a global name including icon.

No issues with the Clan Rank plugin. Both rank and league icon gets displayed as expected.

Adam-:
Additionally since this is now the 3rd plugin using WorldClient and the calls are relatively expensive - maybe that should be centralized and cached somewhere?

hsamoht:
> Additionally since this is now the 3rd plugin using WorldClient and the calls are relatively expensive - maybe that should be centralized and cached somewhere?

It does sound like a good idea based on my limited knowledge of the other plugins. Some of the data would be nice to have cached at least. For example this plugin is only interested in the "Activity" property of a world which would only (if ever) change on update days. It does seem overly expensive to end up performing two HTTP requests to get this data.

Someone with more design/photoshop skills than myself might want to take a look at the icon(s) as well. They have to be 13x13 to fit chat and I only shrinked the 18x18 version from Hiscore plugin which might appear unclear / lacking detail.

runelite-client/src/main/java/net/runelite/client/plugins/chatleagueicons/ChatLeagueIconsConfig.java
```diff
@@ -0,0 +1,88 @@
...
+@ConfigGroup("chatLeagueIcons")
+public interface ChatLeagueIconsConfig extends Config
+{
+	@ConfigItem(
+		keyName = "leagueIcon",
+		name = "Icon",
+		description = "The league icon to display in chat.",
+		position = 1
+	)
+	default LeagueIcon leagueIcon()
```
deathbeam: I dont think the look of the icon should be configurable

deathbeam: Honestly why has this so much configuration in general, just turn the plugin on or off.

hsamoht: It has already been changed to a no-config plugin. We have also fixed the other issues mentioned in this PR. The latest version can be seen here https://github.com/Adam-/runelite/tree/leagueicons

deathbeam: Ah, nice. @Adam- should really start making PRs with that stuff its kind of hard to track :/

Adam-:
@deathbeam i've updated this PR with my branch if you want to review it now.

runelite-client/src/main/java/net/runelite/client/game/WorldService.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	private void fetch()
+	{
+		log.debug("Fetching worlds");
+
+		try
+		{
+			WorldResult worldResult = worldClient.lookupWorlds();
+			worldResult.getWorlds().sort(Comparator.comparingInt(World::getId));
```
deathbeam: Maybe this should also convert WorldResult::getWorlds to api.Worlds as that is common thing we want to do anyway. And then maybe just add the utility findWorld method here as well

runelite-client/src/main/java/net/runelite/client/plugins/defaultworld/DefaultWorldPlugin.java
```diff
@@ -116,32 +115,32 @@ private void changeWorld(int newWorld)
 			return;
 		}
 
-		try
+		final WorldResult worldResult = worldService.getWorlds();
```
deathbeam: This happens right at startup so are we sure the worlds will be actually loaded at this point? Maybe the world client could do blocking call when its being created.

hsamoht:
Made an attempt to improve the icon. 



**Default** = Regular IM Icon for comparison and scale.
**Current** = The 18x18 Icon from Hiscore Plugin shrinked down to 13x13. Currently in the PR.
**Modified** = The 18x18 Icon from Hiscore Plugin centered in 13x13. Added black pixels where it goes out of bounds.
**Remade** = My attempt at remaking the 18x18 Icon from Hiscore Plugin in 13x13 while keeping the look, feel and sharpness of the icon.

If I have to say so myself the Remade one turned out pretty good. Should I update it in the PR?

Alexsuperfly:
remade 😍

deathbeam:
I think using CompletableFuture would be way cleaner than notify

Adam-:
I don't see a great way to work that in, can you elaborate?

deathbeam:
```java
        private final CompletableFuture<WorldResult> completableFuture = new CompletableFuture<>();

        // Fetch thread pseudo code
        new Thread() {
            if (!completableFuture.isDone()) {
                completableFuture.complete(worldResult)
            }
        }

        // getWorlds pseudo code
        if (!completableFuture.isDone()) {
            // Waits until fetch thread completes future, with timeout so we dont wait endlessly
            return completableFuture.get(5, TimeUnit.MINUTES);
        }

        // rest of getWorlds. completableFuture w
        return worldResult;
```

Something like this in ugly pseudo code

PR #10307, opened 2019-11-22
Change Bank plugin to allow users to enter bank pin with keyboard.

JBerben:
Made basic changes to the Bank plugin, API, and added a new rs2asm script to support keyboard functionality for users wanting to enter their bank pin via their keyboard.


Adam-:
I don't think this can be done without faking interface actions (whether the mouse presses or the script invocation due to them) - this is not something we ever do lightly due to the anti macro measures. I'd probably rather not add this just due to that.

Additionally I think the original reason why the PIN works this way is as an anti-keylogging measure. Which while primitive by todays standards is still very much a thing - just google "runelite" and you will find several dumb keyloggers on the first page.

Finally the majority of this PR appears to just be reformatting all of our code (~6k LOC changed) which is just not cool & makes it extra hard to read, so I have not really read this thoroughly.

PR #10308, opened 2019-11-22
Customize chat box clearing

### LINKED ISSUE #6929: Fix the ctrl+backspace discrepancies and the ctrl+w bugs.
Pressing ctrl+backspace in the text boxes in Runelite's sidetabs removes one word correctly
E.g.: "hello this is sentence" =(ctrl+backspace)> "hello this is "

But in game, it removes everything you've typed in the chat box. For some reason, ctrl+w does remove one word, but it also removes the leading space, and the "/" for clan chat text.

JZomerlei:
This closes #6929


deathbeam:
If you want to implement something like this, it should be configurable via the `Hotkey` configuration we have

JZomerlei:
Good suggestion. 

I switched to if statement instead of a switch because the the value had to be a constant and wasn't sure what to do about that.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatKeyboardListener.java
```diff
@@ -54,49 +55,47 @@ public void keyTyped(KeyEvent e)
...
 				clientThread.invoke(() ->
 				{
-					client.setVar(VarClientStr.CHATBOX_TYPED_TEXT, "");
+					client.setVar(VarClientStr.CHATBOX_TYPED_TEXT, replacement);
 					client.runScript(ScriptID.CHAT_PROMPT_INIT);
 				});
-				break;
+			}
+		}
+		else if (chatCommandsConfig.clearChatBox() == Keybind.NOT_SET && e.getKeyCode() == chatCommandsConfig.clearChatBox().getKeyCode())
```
deathbeam: clearChatBox != Keybind.NOT_SET

JZomerlei: Oh whoops lmao.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatKeyboardListener.java
```diff
@@ -54,49 +55,47 @@ public void keyTyped(KeyEvent e)
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (!e.isControlDown() || !chatCommandsConfig.clearShortcuts())
+		if (chatCommandsConfig.clearSingleWord() != Keybind.NOT_SET && e.getKeyCode() == chatCommandsConfig.clearSingleWord().getKeyCode())
```
deathbeam: I would still keep this as switch() and just compare against config, and then just add early return based on config. will make diff way smaller instead of what you did atm

Adam-: I don't think you can switch key binds can you? Also I am pretty sure for this to work correctly with the modifiers you need to be using `Keybind::matches` and not just comparing the key codes like this.

JZomerlei: Correct. Using switch didn’t work because the config() stuff isn’t a constant. The way around it would be on every key pressed even would be to assign them to a final or use onChangedConfig. 

Also I tried a few different key binds and they worked but I never tried to purposely do wrong ones to see if they’d match. I’ll see what I can do. Out on vacation currently.

abextm: You need to use matches to support modifiers correctly. You can never use a switch here because you can't switch on things that are not primitives, enums, or a string. You also don't need to manually check NOT_SET, just do `if (chatCommandsConfig.clearSingleWord().matches(e))`...

deathbeam: Yeah my suggestion about switch was mostly because how it was implemented atm, but it needs to use matches instead as adam and abex already said

JZomerlei: chatCommandsConfig.clearChatBox().matches(e) worked great. Good idea @abextm .

Issue #5609, opened 2018-09-24
Add a "Menu Entry Swapper" option for Arceuus librarians.


purebtw:
Should be pretty straight forward.

deathbeam:
Swap what option?

PR #10315, opened 2019-11-24
Improve Kourend Library User Experience

Hydrox6:
Fixes the overlay not resetting if the bookcase was not checked before the layout changed

Adds an optional tutorial overlay to help users understand how to use the plugin better.

no_data
incomplete
complete

~~Adds a Menu Entry Swap for helping the library staff (thanks @Blackberry0Pie)~~

Adds the option to hide Varlamore Envoy, since it's only used in the Depths of Despair quest, and is not asked for (thanks @Blackberry0Pie)

Makes it so the book that is currently being asked for is highlighted green in the scene overlay (thanks @Blackberry0Pie)


runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryTutorialOverlay.java
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		Player player = client.getLocalPlayer();
+		if (player == null)
+		{
+			return null;
+		}
+
+		WorldPoint playerLoc = player.getWorldLocation();
+		if (playerLoc.getRegionID() != KourendLibraryPlugin.REGION)
```
abextm: this should probably mirror the other overlay's hidden field

Hydrox6: I'm not sure why the current overlay even has a hidden flag, since Overlays that have been removed don't render. 
It also leads to a weird situation where If the user starts the client with the plugin enabled, they see the marks, but if they toggle the plugin even once before going to the library, they don't see the marks until they start trying to find the books.

abextm: idr exactly why we went with a flag instead of adding/removing the overlay, but the behavior was to only show the overlay when you actually start interacting with the minigame. I don't want this to say "interact with the white squares" when there are no white squares because you haven't interacted yet.

ShaunDreclin:
Resolves #5609
Supersedes #10321

Issue #9159, opened 2019-06-20
REQUEST: Different color skyboxes.


lugia00:
Dark purple, dark blue, dark grey.
Also themed once, for example: 'summer, autumn, winter and spring.'

thanks.

chiragbharadwaj:
I think it would be cool if we could use the light brown/beige colored skybox as the default instead of sky blue. This is what I'm talking about:

![o](https://user-images.githubusercontent.com/7849039/60031405-b1817400-9672-11e9-9c59-9b22352cb689.jpg)

IMO this looks a lot better than the current blue, which is too bright. Maybe we can have it as a toggle?

lugia00:
Any updates on this? :)

Mega-boy:
skybox plugin should have an option for "custom color" where you can just enter a hex value and it will override everything.

Issue #7299, opened 2019-01-12
Caves' Skybox Should Still Be Black/Dark


BongoBooty:
**Is your feature request related to a problem? Please describe.**
Upon entering caves like the asgarnian ice dungeon or the cave north of troll stronghold the skybox is white like snow. I understand why this might be, but the area is still underground.

**Describe the solution you'd like**
When entering areas like this, either add a toggle to make it dark or light. I'd prefer either black or dark grey.

**Describe alternatives you've considered**
Can't think of any.

**Additional context**
None

PR #10341, opened 2019-11-28
Skybox color customization options

logarrhythmic:
Current state of these options:
Adds a custom sky color picker, and a dropdown to replace the sky color in either the overworld or everywhere. I should probably change this to a toggle and custom color picker for overworld and caves separately, to accommodate both #7299 and #9159. 

Old text:

> Addresses #9159, although further options might still be nice and the current UX feels a bit strange.
> 
> Adds the following options to the skybox plugin:
> 
> *Custom sky color* - standard color picker
> 
> *Mode* - dropdown, "None", "Replace target sky color", "Overworld" and "Everywhere". *None* is the current behavior of using the sky map and making no substitutions. *Replace target sky color* uses the custom color in place of a specific color value, and blending between areas still works properly. *Overworld* uses the custom color when the player is not in an underground area, and *Everywhere* always uses the custom color.
> 
> *Set current sky as target* - checkbox, gives an "are you sure" window and then stores the sky color of the player's current chunk as the target color. Automatically resets to unchecked, but currently the config panel doesn't update to reflect this until it is reopened. Let me know if there's a way to force an update, and I'll fix this PR to do that. Alternatively, if there's a way to add a button to the config panel that triggers the onConfigChanged event or something similar, that would be exactly what I'm looking for here.
> 
> Target color - hidden option, used as storage and set by the above checkbox.
> 
> The idea with the target color feature is to allow for replacement of the most common sky color, which I personally find too gray, while still getting the special sky colors of various swamps, volcanoes, caves, and other areas which have received special attention in the skybox definition file.
> 
> I also considered a feature that would allow for several source -> replacement pairs to be set, but I couldn't think of a good way to make the UI for that. If there's interest, I'll consider revisiting that idea.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
@@ -25,13 +25,17 @@
 package net.runelite.client.plugins.skybox;
 
 import com.google.inject.Inject;
+import com.google.inject.Provides;
+import java.awt.Color;
 import java.io.IOException;
 import net.runelite.api.Client;
+import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.api.Player;
 import net.runelite.api.coords.LocalPoint;
 import net.runelite.api.events.BeforeRender;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
@@ -47,6 +51,9 @@ public class SkyboxPlugin extends Plugin
 	@Inject
 	private Client client;
 
+	@Inject
+	private SkyboxPluginConfig config;
+
 	private Skybox skybox;
 
 	@Override
@@ -60,6 +67,12 @@ public class SkyboxPlugin extends Plugin
 	{
 		client.setSkyboxColor(0);
 		skybox = null;
+	}
+
+	@Provides
+	SkyboxPluginConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(SkyboxPluginConfig.class);
 	}
 
 	private int mapChunk(int cx, int cy, int plane)
@@ -91,6 +104,14 @@ public class SkyboxPlugin extends Plugin
 			return;
 		}
 
+		Color overrideColor = player.getWorldLocation().getY() < Constants.OVERWORLD_MAX_Y
+			? config.customOverworldColor() : config.customOtherColor();
+		if (overrideColor != null)
+		{
+			client.setSkyboxColor(overrideColor.getRGB());
+			return;
+		}
+
 		int px, py;
 		if (client.getOculusOrbState() == 1)
 		{
@@ -99,7 +120,7 @@ public class SkyboxPlugin extends Plugin
 		}
 		else
 		{
-			LocalPoint p = client.getLocalPlayer().getLocalLocation();
+			LocalPoint p = player.getLocalLocation();
 			px = p.getX();
 			py = p.getY();
 		}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2019 logarrhytmic <https://github.com/logarrhythmic>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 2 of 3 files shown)


abextm:
I really don't like how this abuses the config

logarrhythmic:
same

logarrhythmic:
I see there's a framework for adding entire panels, like the library plugin does. Would it be okay to add one of those? It seems like a pretty good solution to me, and it would also let me add functionality for multiple replacement colors instead of just the 1.
I'm thinking I would replace the "target color" mode with an "advanced" option in the dropdown, which would enable the sky color replacement panel.
The panel would then have a more sophisticated UI with a list of color pairs, and a way to add new ones.

abextm:
There isn't a framework for adding custom things to the config panel, which is what this really needs. It would probably be best for this to just be a global override color that makes it not invoke `Skybox` entirely

logarrhythmic:
I could strip this PR down to just have the global overrides and then work on the replacement pairs to make another PR later, since I really want that specific feature myself.

abextm:
that would probably be best. I think the best thing for more granular overrides would be to be able to merge the config in our `skybox.txt` format

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
```diff
@@ -91,6 +103,17 @@ public void onBeforeRender(BeforeRender r)
 			return;
 		}
 
+		if
+		(
+			config.overrideMode() == SkyOverrideMode.ALL ||
+			(config.overrideMode() == SkyOverrideMode.OVERWORLD && client.getLocalPlayer().getWorldLocation().getY() < 4200)
```
Nightfirecat: It looks to me like `4160` marks the first tile beyond the northernmost water, and this should also be defined in some publicly-accessible location (maybe `Constants`?) for general use--in the past, other plugins/features have wanted a reference to this value.

runelite-client/src/main/java/net/runelite/client/plugins/skybox/Skybox.java
```diff
@@ -352,14 +352,17 @@ private int chunkData(int cx, int cy, int plane, ChunkMapper chunkMapper)
 		}
 
 		int cv = chunks[(stride * (cy - y1)) + (cx - x1)];
+
```
Nightfirecat: Now that there are no more code changes to Skybox.java, remove the added newlines in this file.

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
```diff
@@ -0,0 +1,59 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
Nightfirecat: Remove the newline between the copyright header and package definition.

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
```diff
@@ -0,0 +1,59 @@
...
+package net.runelite.client.plugins.skybox;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+import java.awt.Color;
+import net.runelite.client.plugins.skybox.config.SkyOverrideMode;
```
Nightfirecat: All imports should be in a single block with no newlines separating them.

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
```diff
@@ -91,6 +103,17 @@ public void onBeforeRender(BeforeRender r)
...
+		)
```
Nightfirecat: The formatting for this condition doesn't quite match what we use elsewhere in the project. I'd expect this to look like the following:

```java
if (config.overrideMode() == SkyOverrideMode.ALL
	|| (config.overrideMode() == SkyOverrideMode.OVERWORLD && client.getLocalPlayer().getWorldLocation.getY() < 4200))
{
	// ...
}
```

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
logarrhythmic: Yup, seems good, adding it as Constants.OVERWORLD_MAX_Y.

runelite-client/src/main/java/net/runelite/client/plugins/skybox/Skybox.java
logarrhythmic: Yep, will do

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
logarrhythmic: Alright

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPluginConfig.java
logarrhythmic: Okay

runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
logarrhythmic: Okay

logarrhythmic:
I'm going to change this to two toggles and custom color pickers for overworld and caves separately, to accommodate both #7299 and #9159.

PR #10342, opened 2019-11-28
GPU: smooth out fog corners by 1 tile, eliminating the diagonal line

logarrhythmic:
before and after video clips: https://imgur.com/a/5GM2t7D

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
@@ -81,8 +81,13 @@ void main()
   int fogSouth = max(FOG_SCENE_EDGE_MIN, cameraZ - drawDistance);
   int fogNorth = min(FOG_SCENE_EDGE_MAX, cameraZ + drawDistance - TILE_SIZE);
 
+
   // Calculate distance from the scene edge
-  float fogDistance = min(min(vertex.x - fogWest, fogEast - vertex.x), min(vertex.z - fogSouth, fogNorth - vertex.z));
+    int xDist = min(vertex.x - fogWest, fogEast - vertex.x);
+    int zDist = min(vertex.z - fogSouth, fogNorth - vertex.z);
+    float nearestEdgeDistance = min(xDist, zDist);
+    float secondNearestEdgeDistance = max(xDist, zDist);
+    float fogDistance = nearestEdgeDistance - 1.5 * TILE_SIZE * clamp((nearestEdgeDistance + 1) / (secondNearestEdgeDistance + 1), 0, 1);
 
   vFogAmount = fogFactorLinear(fogDistance, 0, fogDepth * TILE_SIZE) * useFog;
 }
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/resources/net/runelite/client/plugins/gpu/vert.glsl
```diff
@@ -81,8 +81,13 @@ void main()
...
   int fogNorth = min(FOG_SCENE_EDGE_MAX, cameraZ + drawDistance - TILE_SIZE);
 
+
   // Calculate distance from the scene edge
-  float fogDistance = min(min(vertex.x - fogWest, fogEast - vertex.x), min(vertex.z - fogSouth, fogNorth - vertex.z));
+    int xDist = min(vertex.x - fogWest, fogEast - vertex.x);
+    int zDist = min(vertex.z - fogSouth, fogNorth - vertex.z);
+    float nearestEdgeDistance = min(xDist, zDist);
+    float secondNearestEdgeDistance = max(xDist, zDist);
+    float fogDistance = nearestEdgeDistance - 1.5 * TILE_SIZE * clamp((nearestEdgeDistance + 1) / (secondNearestEdgeDistance + 1), 0, 1);
```
Adam-: Instead of offsetting `fogDistance` did you consider increasing the `fogDepth` by the same amount? that might make more sense.

logarrhythmic: Increasing `fogDepth` doesn't get rid of the diagonal line at all. Did you look at the first clip in the imgur link?

The idea here is that `nearestEdgeDistance / secondNearestEdgeDistance` is near 0 along most of the edges, but approaches 1 at the diagonal lines going between corners. This is then multiplied by `1.5 * TILE_SIZE` and substracted from the edge distance to make vertices at the corners of the scene get a fog amount which would correspond to them being up to 1.5 tiles closer to the edge. It needs to be a constant multiplier, and not based on the `fogDepth` option at all, to have the desired effect.

These will draw a graph of what's happening at each corner for you: 
[corner component](https://www.wolframalpha.com/input/?i=plot+%28min%28%7Cx%7C%2C%7Cy%7C%29%2B1%29+%2F+%28max%28%7Cx%7C%2C%7Cy%7C%29%2B1%29%2C+x%3D0..10%2C+y%3D0..10), [whole equation](https://www.wolframalpha.com/input/?i=plot+min%28x%2Cy%29+-+1.5+*+%28min%28x%2Cy%29%2B1%29+%2F+%28max%28x%2Cy%29%2B1%29%2C+x%3D0..10%2C+y%3D0..10)

1.5 is a value I ended up with through trial and error, when 1, 2 and 1.4142 didn't work quite as well. Playing with the values in WolframAlpha just now let me figure out what I was doing wrong with those, though, so I'll be making a commit to adjust the math. Basically, instead of `([second]nearestEdgeDistance + 1)` I should be doing `([second]nearestEdgeDistance + 1.5*1.5)` if I'm multiplying the whole thing by 1.5. This'll eliminate the square you see in the corner in the second WolframAlpha link, so I can get rid of the clamp function.

PR #10360, opened 2019-11-30
League Chat Icons: Fix conflict with Chat Colors plugin on usernames

### LINKED ISSUE #10358: Chat League Icons plugin prevents Chat Color > Usernames from displaying color
**Describe the bug**
The Chat League Icons plugins will prevent the Chat color plugin from displaying color on Usernames if you had it set up.

**To Reproduce**
1. Setup your Chat Color for Usernames
2. Activate the Chat Leagues Icons plugin
3. The Usernames color will disappear if you type in chat.

**Expected behavior**
I was expecting to see the League Icon next to my name + the Usernames color to be displayed.

**Screenshots**
CLI

 - OS: [Ubuntu]
 - RuneLite version: [1.5.42.1]
 - Launcher version: [2.0.0]

hsamoht:
Fixes #10358 

**Changes:** 
- League Chat Icons no longer overrides username color set from Chat Colors plugin. 

**Screenshots**
SS of regular chat

SS of transparent chat

SS showing Chat Color + Emoji + League together


runelite-client/src/main/java/net/runelite/client/plugins/leaguechaticons/LeagueChatIconsPlugin.java
```diff
@@ -58,13 +58,14 @@
...
 @PluginDescriptor(
-	name = "Chat League Icons",
+	name = "League Chat Icons",
 	description = "Changes the chat icon for players on league worlds",
 	enabledByDefault = false
 )
 @Slf4j
 public class LeagueChatIconsPlugin extends Plugin
 {
+
```
Nightfirecat: Remove this extra newline.
```suggestion
```

runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
```diff
@@ -107,7 +107,7 @@ public void onConfigChanged(ConfigChanged event)
 		}
 	}
 
-	@Subscribe
+	@Subscribe(priority = 1) // run after all other onChatMessage plugins
```
Nightfirecat: I don't think this is a good solution for this issue. The League Chat Icons plugin should be modified to not remove color tags instead imo.

Nightfirecat: If this is the solution that we want to add, we should add a test to ensure this priority will always be higher than that of other `ChatMessage` subscribers.

hsamoht: Not exactly sure how to go about writing this integration test. All other tests written for `ChatMessage` seem to be unit tests. Even the Emoji and Chat Color interaction is handled as a unit test rather than integration.

Seems to me the easiest way of checking the order would be to check the outcome of a "default" `ChatMessage` invocation (the one all subscribes to). I am not sure where that is located. That way we would know it is correct regardless of what other subscribers do as well as future proof for changes and new plugins. 

Personally I am in favour of my initial solution by using the information from `MessageNode` rather than `ChatMessage`. It is afterall the `MessageNode` that gets updated in all the plugins: Emoji, Chat Color and this. This would allow the plugin to change the icon without, indirectly, also changing the name.

Either way we should indeed have both unit and integration tests for this plugin.

hsamoht: Adam said integration testing this might not be doable.

Showing it working with Color + Emoji + League:
SS showing Chat Color + Emoji + League together

Don't think any other plugins modifies the `ChatMessage` or `MessageNode` objects.

PR #10364, opened 2019-12-01
Right-Click Camera, Rotate te camera with right-click

Wynadorn:
Added a small plugin that allows rotating of the camera through right-click when there are no menu entries.

I find this plugin very useful when there's lots of camera rotating.
Clicking anywhere on the floor, a wall or in the sky will allow you to rotate the camera.
This makes it so you dont have to keep swapping your finger to the scroll wheel.

Turning you camera with right-click realy shines in activities like agility, runecrafting, stealing artifacts, or bosses like Vorkath.

Options are available to ignore "Examine" entries and to open the menu though middle-click.
I've added an option to make middle-click open the menu, this so you can prepare a "Walk here", for example during bossing.

Give it a go, hope you'll like it!

config options

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
@@ -23,7 +23,7 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package net.runelite.client.plugins.zoom;
+package net.runelite.client.plugins.camera;
 
 import com.google.common.primitives.Ints;
 import com.google.inject.Inject;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -194,4 +218,165 @@ public void keyReleased(KeyEvent e)
...
+	private boolean hasMenuEntries(MenuEntry[] menuEntries)
+	{
+		for (MenuEntry menuEntry : menuEntries)
+		{
+			MenuAction action = MenuAction.of(menuEntry.getType());
+			switch (action)
+			{
+				case CANCEL:
+				case WALK:
+					continue;
```
deathbeam: Shouldn't this just be break

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -194,4 +218,165 @@ public void keyReleased(KeyEvent e)
...
+				case EXAMINE_OBJECT:
+				case EXAMINE_NPC:
+				case EXAMINE_ITEM_GROUND:
+				case EXAMINE_ITEM:
+				case EXAMINE_ITEM_BANK_EQ:
+					if (config.ignoreExamine())
+					{
+						continue;
```
deathbeam: Same here

PR #10369, opened 2019-12-02
runelite-api: Add priorities utility class

Trevor159:
I am not sure if this is wanted, but I think it is good to have all of the priorities in one place so there aren't random constants throughout the events.

I am not sure if this is the right place to put the class.


deathbeam:
This pretty much defeats the purpose of having modularity, would you just add also priorities from plugins here? Idk, I don't think this makes sense

Adam-:
I agree, this makes no sense.

PR #10372, opened 2019-12-02
league icons: add option to use green ironman helmet

### LINKED ISSUE #10352: Chat League Icons, Icon is ugly
Chat icon that replaces the default ironman helmet is too bright and colorfull.

Suggested: Normal ironman helmet with a green tint.

dekvall:
Adds an option to use league_ironman as a league icon.

Transparent chatbox:
transparent_chatbox
Normal chatbox:
normal_chatbox

Closes #10352


dekvall:
After Discord discussion it was concluded that this will not be added because: 1) It's confusing to have multiple icons and 2) Leagues just happens to be ironman mode for this season, it might not be in the future.

PR #10379, opened 2019-12-03
Make ToB party interface movable

deathbeam:


deathbeam:
Also added movable HP bars inside ToB. The tooltip for them moves when they are not in default position but it does not really matter.

PR #10383, opened 2019-12-04
mining: show respawn timers after GameState change

Alexsuperfly:
After making #10381 i noticed the mining plugin had this same fault so i decided to refactor it aswell.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
@@ -45,6 +45,7 @@ import static net.runelite.api.ObjectID.ORE_VEIN_26664;
 import net.runelite.api.WallObject;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.GameObjectDespawned;
+import net.runelite.api.events.GameObjectSpawned;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.WallObjectSpawned;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -132,6 +135,23 @@ public void onGameObjectDespawned(GameObjectDespawned event)
...
+	public void onGameObjectSpawned(GameObjectSpawned event)
+	{
+		if (client.getGameState() != GameState.LOGGED_IN)
+		{
+			return;
+		}
+		GameObject object = event.getGameObject();
+
+		// If the Lovakite ore respawns before the timer is up, remove it
+		if (object.getId() == ROCKS_28596 || object.getId() == ROCKS_28597)
```
Nightfirecat:
```suggestion
		if (Rock.getRock(object.getId()) == Rock.LOVAKITE)
```

Nightfirecat: Alternatively, refactor this method and `onWallObjectSpawned` to use a common helper method and add these IDs to the `switch` that already exists there, alongside the `ORE_VEIN` cases.

Alexsuperfly: i took your refactor suggestion, that seems more extendable in the future if other rocks end up behaving that way

runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
```diff
@@ -87,7 +88,7 @@ Duration getRespawnTime(int region)
 				return region == MINING_GUILD ? Duration.ofMinutes(6) : super.respawnTime;
 			}
 		},
-	ORE_VEIN(Duration.ofSeconds(MiningOverlay.ORE_VEIN_MAX_RESPAWN_TIME), 150),
+	ORE_VEIN(Duration.ofSeconds(MiningOverlay.ORE_VEIN_MAX_RESPAWN_TIME), 150, ORE_VEIN_26661, ORE_VEIN_26662, ORE_VEIN_26663, ORE_VEIN_26664),
```
Adam-: This doesn't seem right. Not sure why the existing code is as it is but currently it adds when the depleted ore veins spawn, and this would in addition add when the non-depleted ones despawn.. did you test this at mlm?

Alexsuperfly:
rebased to use the new GAME_TICKS system for timing

Adam-:
@Nightfirecat have you tested these new respawn timers?

Nightfirecat:
Yes, I checked both MLM veins (because those historically had issues) and Lovakite rocks with this patch, along with a handful of rocks in the mining guild, and they both seemed fine.

Issue #10317, opened 2019-11-24
Screenmarkers can create a null marker


abextm:
<details><summary> stack traces </summary>

```
java.lang.NullPointerException: null
	at net.runelite.client.plugins.screenmarkers.ScreenMarkerOverlay.getName(ScreenMarkerOverlay.java:55)
	at net.runelite.client.ui.overlay.OverlayManager.saveOverlayPosition(OverlayManager.java:345)
	at net.runelite.client.ui.overlay.OverlayManager.saveOverlay(OverlayManager.java:247)
	at net.runelite.client.plugins.screenmarkers.ScreenMarkerPlugin.finishCreation(ScreenMarkerPlugin.java:194)
	at net.runelite.client.plugins.screenmarkers.ui.ScreenMarkerCreationPanel$1.mousePressed(ScreenMarkerCreationPanel.java:88)
	at java.awt.AWTEventMulticaster.mousePressed(AWTEventMulticaster.java:280)
	at java.awt.Component.processMouseEvent(Component.java:6536)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6304)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4904)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4532)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4476)
	at java.awt.Container.dispatchEventImpl(Container.java:2283)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:760)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
[...]

PR #10388, opened 2019-12-04
Removes null values from the screenmarkers

sevvlor:
This should fix part of #10317. I haven't been able to replicate the double calling of `finishCreation`.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
@@ -35,8 +35,8 @@ import java.awt.Rectangle;
 import java.awt.image.BufferedImage;
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
```diff
@@ -250,10 +251,14 @@ public void updateConfig()
 		}
 
 		final Gson gson = new Gson();
+
 		final List<ScreenMarker> screenMarkerData = gson.fromJson(json, new TypeToken<ArrayList<ScreenMarker>>()
 		{
+			// Intentionally left empty
 		}.getType());
 
+		screenMarkerData.removeAll(Collections.singleton(null));
```
deathbeam: you can simply filter the null values below instead of using removeAll. Also the extra comment is unnecessary

Adam-:
I'd be interesting in knowing what causes this first before we try to fix it I think.

Adam-:
Couldn't figure this out, added as part of 5216b03090595a568debc735ae1f58ecc1491541

PR #10398, opened 2019-12-06
Strip out non-digits characters in skill calculator

### LINKED ISSUE #10386: [Skill Calc] Strip commas and periods
I frequently use window's built in calculator tool alongside the runelite skill calc. However, when attempting to copy and paste from the calculator into the skill calc commas will be placed separating the digits into groupings of three for legibility. 

Currently if I were to hit enter the skill calc would replace my number with 0. In order for the skill calc to properly read the input I must manual remove the commas every time. 

I believe an alternative to this manual adjustment could be to strip the commas out of the number instead of returning 0. I am aware that other regions use periods in place of commas, and would assuming that also stripping periods would be desirable.

pyonium:
Closes #10386

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
@@ -98,7 +98,7 @@ class UICalculatorInputArea extends JPanel
 	{
 		try
 		{
-			return Integer.parseInt(field.getText().replaceAll("[,.']", ""));
+			return Integer.parseInt(field.getText().replaceAll("\\D", ""));
 		}
 		catch (NumberFormatException e)
 		{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
@@ -98,7 +98,7 @@ class UICalculatorInputArea extends JPanel
 	{
 		try
 		{
-			return Integer.parseInt(field.getText().replaceAll("[,.]", ""));
+			return Integer.parseInt(field.getText().replaceAll("[,.']", ""));
 		}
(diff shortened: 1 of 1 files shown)


deathbeam:
Imo this should be removed when inputting the text and not when getting it

pyonium:
I tested it the way it's implemented now, and it seems fine (you can input whatever, it evaluates when you press enter). What would be the preferred behaviour?

JZomerlei:
I think DB is saying that `
.replaceAll("[,.']", "")
` should be on the setInput method like

```
	private void setInput(JTextField field, Object value)
	{
		field.setText(String.valueOf(value).replaceAll("[,.']", ""));
	}
```

So that the it's right when it's placed into the text field rather than needing to be corrected every time it's used in calculations in the getInput() method.

JZomerlei:
Additionally @Nightfirecat noted that you can strip out all non numeric values by using \D instead of specifically using [,.']

pyonium:
@JZomerlei I tried this change, however, this doesn't produce the desired result.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
```diff
@@ -98,7 +98,7 @@ private int getInput(JTextField field)
 	{
 		try
 		{
-			return Integer.parseInt(field.getText());
+			return Integer.parseInt(field.getText().replaceAll("\\D", ""));
```
Nightfirecat: Actually, this should use a pre-compiled `Pattern` (ie. `private static final Pattern NON_DIGIT_MATCHER = Pattern.compile("\\D");`) and use that to replace non-digit matches.

```suggestion
			return Integer.parseInt(NON_DIGIT_MATCHER.matcher(field.getText()).replaceAll(""));
```

Adam-: I'm fine with it as it is, regardless of the subtle resolve.

PR #10402, opened 2019-12-07
Chat Commands - Add boss hiscore support for kc command

TheStonedTurtle:
Updates the !kc command to utilize the jagex hiscores. Also updates the output to include their rank and how many they've killed during their current login session.

As of creating the PR the hiscores api is still broken and not including all boss kcs (missing callisto and cerb atm)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -114,6 +118,8 @@
 	private String lastBossKill;
 	private int lastPb = -1;
 
+	private Map<HiscoreSkill, Integer> kcMap = new HashMap<>();
```
Adam-: What is the purpose of this?

TheStonedTurtle: This stores the current KC for bosses when killed if they are a hiscore boss. Since the hiscore updates on logout and people like the real-time kc commands I'm using this map to store & send the actual kc. I figured since Jagex is hosting this KC data it's no longer necessary to persist it via the config. I believe the website is the only other place that uses this data and it could just query the hiscores as well.

TheStonedTurtle: Based on discord feedback I've reverted the commit that removed config storage for boss kcs.

PR #10404, opened 2019-12-07
Add plugin-level notification overrides

TheStonedTurtle:
Adds a toggleable system for plugin-level notification overrides. Goal was to allow specific plugins to send notifications while focused and not others even though both were enabled, such as chat alerts and wintertodt. Decided to add all config options to make this more customizable.

j4a9Uhp5Of


TheStonedTurtle:
I've updated the PR to have an annotation `@CustomNotifier` which handles the notifier options and updated the UI to have a bell icon when enabled which takes them to a view with just the notifier options.

This should be ready for an initial review now.

J3m1y7HG0Y

TheStonedTurtle:
The PR has been updated with a new UI design and other logic changes.

dwyVeRX3VT

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -132,32 +134,53 @@ private Notifier(
...
+		final boolean useCustom = customSettings != null && runeLiteConfig.enableCustomNotifier() && !customSettings.notificationState().equals(NotificationSettings.States.ON);
+		final NotificationSettings config = useCustom ? customSettings : getRuneLiteNotificationSettings();
+
+		if (config.notificationState().equals(NotificationSettings.States.OFF))
 		{
 			return;
 		}
 
-		if (runeLiteConfig.requestFocusOnNotification())
+		if (!config.sendNotificationsWhenFocused() && clientUI.isFocused())
```
Nightfirecat: This can be combined with the guard clause above as there are no assignments or operations between guards.

TheStonedTurtle:
@Nightfirecat Thanks for pointing the boolean background out, I've fixed it now 👍 
image

In general I think there needs to be an indicator inside the list view for which plugins have customizable notification settings.

Nightfirecat:
Adding the `[discussion]` tag as we should discuss our thoughts on the UI/UX of this implementation before proper code review takes place.

Marstead:
From some discussion in Discord, this might close https://github.com/runelite/runelite/issues/11679.

My two cents on the UI -- the RuneLite plugin should have a "Allow plugins to override notification settings" option, unchecked by default. If this is checked, all of the plugins with their own notifications (Ground Items, Chat Notifications, Idle Notifier etc) will gain new options for:

Send notifications when focused
Notification sound
Flash notification

Each of these will include "RuneLite Default" as the default option. For the notification sound, choosing a Custom sound instead of RuneLite Default will require the user to add an additional *.wav file to the Runelite folder named after the plugin (e.g. "GroundItems.wav", "IdleNotifier.wav", etc)

MoralFiber:
Why does this no longer exist

cdfisher:
> Why does this no longer exist

It was never added in the first place.

MoralFiber:
OH... :(

PR #10405, opened 2019-12-07
MES: Add config for Trade Crewmembers (Trade/Charter/Previous)

### LINKED ISSUE #4053: Charter option to Menu Entry Swapper.
I'd like to click straight charter ship when talking to the trademen. As it would be faster to run farm runs then :)

thaugrud:
**Changes:**
- Adds a config item for Trade Crewmembers to select which option  to swap

Image of new config item

Closes #4053

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -518,6 +518,14 @@ public class MenuEntrySwapperPlugin extends Plugin
 					case CHARTER:
 						swap("charter", option, target, index);
 						break;
+					case PREVIOUS_LOCATION:
+						swap("charter-to catherby", option, target, index);
+						swap("charter-to brimhaven", option, target, index);
+						swap("charter-to port khazard", option, target, index);
+						swap("charter-to corsair cove", option, target, index);
+						swap("charter-to musa point", option, target, index);
+						swap("charter-to port sarim", option, target, index);
+						break;
 				}
 			}
 		}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -454,7 +454,8 @@ public class MenuEntrySwapperPlugin extends Plugin
 				swap("assignment", option, target, index);
 			}
 
-			if (config.swapTrade())
+			// exclude crewmembers as they are handled separately
+			if (config.swapTrade() && !target.equals("trader crewmember"))
 			{
 				swap("trade", option, target, index);
 				swap("trade-with", option, target, index);
@@ -466,7 +467,8 @@ public class MenuEntrySwapperPlugin extends Plugin
 				swap("claim-slime", option, target, index);
 			}
 
-			if (config.swapTravel())
+			// exclude crewmembers as they are handled separately
+			if (config.swapTravel() && !target.equals("trader crewmember"))
 			{
 				swap("travel", option, target, index);
 				swap("pay-fare", option, target, index);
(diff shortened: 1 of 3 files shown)


hsamoht:
Added the new option "Previous location" to the config.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -429,7 +430,8 @@ private void swapMenuEntry(int index, MenuEntry menuEntry)
 				swap("claim-slime", option, target, index);
 			}
 
-			if (config.swapTravel())
+			// exclude crewmembers as they are handled separately
+			if (config.swapTravel() && !target.equals("trader crewmember"))
```
dekvall: `charter` should probably get taken off of this swap

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -468,6 +470,27 @@ private void swapMenuEntry(int index, MenuEntry menuEntry)
 			{
 				swap("start-minigame", option, target, index);
 			}
+
+			if (target.equals("trader crewmember"))
```
dekvall: Move this above `swapTrade` and `swapTravel` like the slayer assignment is so you don't need to do the `!target.equals("trader crewmemeber")` for those.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -468,6 +470,27 @@ private void swapMenuEntry(int index, MenuEntry menuEntry)
...
+						swap("charter", option, target, index);
+						break;
+					case PREVIOUS_LOCATION:
+						swap("charter-to catherby", option, target, index);
+						swap("charter-to brimhaven", option, target, index);
+						swap("charter-to port khazard", option, target, index);
+						swap("charter-to corsair cove", option, target, index);
+						swap("charter-to musa point", option, target, index);
+						swap("charter-to port sarim", option, target, index);
+						break;
```
dekvall: also add `case OFF:` to this imo

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
hsamoht: Doing this change we get a side effect if the option `OFF` is selected where it will swap the trade option depending on what `swapTrade` is set to. Is that a behaviour we want? Doesn't occur with `swapTravel` as `charter` gets removed from those options.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
hsamoht: Wouldn't this be more correct then? Seeing as neither will do anything.

``` 
case OFF:
case default:
    break;
```

dekvall: Sure, but i don't think it will ever hit `default`

Adam-:
Alternatively, we could just add a swap for `Previous location`, since trade is already covered?

hsamoht:
That seems like a good solution for that option specifically. Downside is it may not be obvious to the user which option takes priority over the other. Upside being it probably a lot cleaner in terms of interface.

The initial PR was addressing the issue of not being able to have `Charter` as first option when also having `Trade` selected. If we're not going with the dropdown solution I'd argue the priority order should be:

1) Previous location
2) Charter
3) Trade

Would that priority swap have any side effects for other NPCs?

jevonsp:
Is this gonna get merged into rl? can i put it in myself if so?

PR #10420, opened 2019-12-09
npcaggro: remove hint hiding option

Hydrox6:
Ever since this "feature" was merged, users have been getting confused when the plugin magically doesn't work. The hint isn't a tutorial overlay, it's an integral part of user feedback. Hiding it permanently should never have been merged in the first place; a user should either get the plugin to work by following the instructions, or turn the plugin off.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
@@ -94,21 +94,10 @@ public interface NpcAggroAreaConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "npcUnaggroShowNotWorkingOverlay",
-		name = "Hide overlay hint",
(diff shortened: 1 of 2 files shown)


Adam-:
If anything this should just be a revert of the commit adding it, but since the hint is on by default I don't understand how anyone could be confused by this?

Hydrox6:
Users have repeatedly shown that they will just enable random options without knowing what they do, then complain when it doesn't work; It happens with Item Prices' `Hide Tooltips on Inventory Items` a surprising amount. 

And as I mentioned above, the hint is integral to telling the user how to use the plugin and when it's not working. If they don't like that there's a constant overlay on the screen, they can disable the plugin, as it will only ever appear if they're not using the plugin.

Good call on reverting the commit directly though. I'll update it to do that.

Nightfirecat:
Needs rebase. Code still LGTM

Hydrox6:
Rebased, wasn't sure exactly what to do with the conflict so I made a second commit to fix the position.

Hydrox6:
Squashed the commits together after recommendation from this morning

PR #10421, opened 2019-12-10
Adventure log pbs

### LINKED ISSUE #9549: [chatcommands] Add shorthands for Gauntlet and Corrupted Gauntlet kc

Alexsuperfly:
Allows the use of the PoH adventure log to populate the players killcounts and pbs.
The log owner is tracked when it gets opened and only grabs the data if that name and local player name are the same.

fixes #9549


Nightfirecat:
FYI I force-pushed this to update the order the boss log and house adventurer log are checked in the first commit so there's less code reordering.

Issue #10351, opened 2019-11-29
sire timer after update


themozerath:
**Describe the bug**
The sire 'stun' timer dissapears if you stun sire whenever he is stunned already, since now this is possible in the new update. 
**To Reproduce**
1.stun sire
2. Stun sire again within the 30second stun. 
3. the timer will be dissapeared

*Expected behavior**
Instead of the timer to be dissapeared restart the timing 

**Screenshots**

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.5.42
 - Launcher version: 1.6.0

**Additional context**

[launcher.log](https://github.com/runelite/runelite/files/3905873/launcher.log)
[client_2019-11-25.0.log](https://github.com/runelite/runelite/files/3905874/client_2019-11-25.0.log)
[launcher_2019-11-25.0.log](https://github.com/runelite/runelite/files/3905875/launcher_2019-11-25.0.log)
[client.log](https://github.com/runelite/runelite/files/3905876/client.log)

PR #10422, opened 2019-12-10
added removeGameTimer before creating a new one for sire stun and added

btzc:
Addresses #10351 

A bug was introduced where the stun timer disappeared with the sire was stunned again while already stunned. The intended functionality was for the timer to refresh instead.

Added `stun_flag` to keep track of whether or not the sire is already stunned or not. If the flag is 0 then the sire isn't stunned, and if the flag is 1 then the sire is. Then checked on stun flag to see whether or not we needed to remove the current timer or not.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -726,12 +726,15 @@ public class TimersPlugin extends Plugin
 				// Show the countdown when the Sire enters the stunned state.
 				case NpcID.ABYSSAL_SIRE_5887:
 					// not stunned
-					if(stun_flag == 0) {
+					if(stun_flag == 0)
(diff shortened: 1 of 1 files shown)


btzc:
Oh I see that makes sense. So just to make sure I understand what you mean you're suggesting I try and look into something maybe a little similar to this? https://github.com/runelite/runelite/blob/d76a2670de9d51711163ff6a3deb699b36092ff4/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java#L470

Nightfirecat:
Something like that yeah, for whichever subscriber is appropriate. (like I said, idk if it is a chat message or an animation or something else)

Adam-:
Why is the timer going away when it is being re-stunned?

Nightfirecat:
This PR has not seen any changes since feedback was given, so I am closing it. Please re-open it and push your changes if you plan to update it further.

Issue #10418, opened 2019-12-09
[Feature Request] Highlight Library books you're missing in the right-hand list


Jademalo:
**Is your feature request related to a problem? Please describe.**
When collecting books for the library, it shows you where the one you currently need for the task is. However, it's difficult to see at a glance which books you're missing, and where you can go to efficiently restock.

**Describe the solution you'd like**
In the same way that the current book for the current task is highlighted in green, it would be useful to have books that you don't have in your inventory highlighted in red.
This way you can see the most efficient way to go up and down the floors to restock as many books with as little running as possible. 

This shouldn't be too bad to implement, since "Hide Duplicate Book" in the library options is already able to determine which books you posess.

**Additional context**
Current
![image](https://user-images.githubusercontent.com/386846/70422930-7b8b4d00-1a64-11ea-9aaf-46252c12be11.png)

Proposed
![image](https://user-images.githubusercontent.com/386846/70423206-f48aa480-1a64-11ea-8ed3-8fea5e398200.png)

vitorebom:
Aditionally, is it possible for the list to include the Dark Manuscript and it fetches the location when you talk to Biblia?

Hydrox6:
That's an entirely separate issue, #10296

PR #10425, opened 2019-12-11
Highlight missing Kourend Library books in side panel

h3half:
Adds feature request #10418.

Here is a screenshot of this code in action:
Screenshot_1

It highlights the requested book green, books in your inventory white, and books not in your inventory red.

I could easily add a togglable config setting for whether to enable the coloring or not, but I'm not sure if that's necessary. The coloring does make the panel look a little busier but it's not that bad IMO.


Jademalo:
Awesome, thank you for the quick work in implementing this!

abextm:
I don't like the red. It would probably be better to dim the color of books in your inventory and keep things that are missing full white.

PR #10453, opened 2019-12-16
woodcutting: fix inaccurate despawn events

### LINKED ISSUE #10449: Woodcutting: Tree respawn timers don't go away when changing floors at redwoods
**Screenshots**
Screenshot

**Environment:**
 - OS: Windows 10
 - RuneLite version: 1.5.43-SNAPSHOT

Alexsuperfly:
game objects only update when the player can see them on their current plane
track the previous onGameTick player plane to ignore any despawn event that does not occur on that same plane

the server only spawns and despawns objects on your current plane
so when you go down stairs you get hit with all the changes that happened while you were gone, this means any tree that has been chopped while you were upstairs gets a fresh new despawn event.  
this pr accounts for that by only allowing the plugin to act on those events if you have been on that plane for a tick already

fixes #10449


Adam-:
The despawn events aren't "inaccurate" - this is fixing the plugin showing despawn timers when changing planes.

Issue #10442, opened 2019-12-15
Ignore list notes


adambtwl0l:
**Is your feature request related to a problem? Please describe.**
It's rather annoying when reporting toxic individuals/bots and seeing your ignore list and that you don't realize why people are on there anymore. 

**Describe the solution you'd like**
Would it be possible to make it function like friends list where people can add tags? Also would it be possible to right click look them up to see if they've been banned? That way we could easily see they're banned and thus no real reason to keep them on there 

**Describe alternatives you've considered**
Typing the names into the hs plugin manually (very annoying)

Adam-:
added in #10455

PR #10455, opened 2019-12-17
add ignore list notes by modifying FriendNotes plugin

Rami-J:
Closes Issue: #10442 

Adds notes for players on the ignore list, by modifying the FriendNotes plugin.


runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -61,10 +62,13 @@
 public class FriendNotesPlugin extends Plugin
 {
 	private static final String CONFIG_GROUP = "friendNotes";
+	private static final String CONFIG_GROUP_IGNORE = "ignoreNotes";
```
Adam-: It would probably make more sense to reuse the config group for friendnotes, since a player cannot be on both your friend list and ignore list, and it would let the note migrate between lists.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -177,11 +177,12 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
 		if (groupId == WidgetInfo.FRIENDS_LIST.getGroupId() || groupId == WidgetInfo.CLAN_CHAT.getGroupId() ||
 				groupId == WidgetInfo.CHATBOX.getGroupId() && !KICK_OPTION.equals(option) || //prevent from adding for Kick option (interferes with the raiding party one)
-				groupId == WidgetInfo.RAIDING_PARTY.getGroupId() || groupId == WidgetInfo.PRIVATE_CHAT_MESSAGE.getGroupId())
+				groupId == WidgetInfo.RAIDING_PARTY.getGroupId() || groupId == WidgetInfo.PRIVATE_CHAT_MESSAGE.getGroupId() ||
+				groupId == WidgetInfo.IGNORE_LIST.getGroupId())
 		{
 			boolean after;
 
-			if (!AFTER_OPTIONS.contains(option))
+			if (!AFTER_OPTIONS.contains(option) || (option.equals("Delete") && groupId != WidgetInfo.IGNORE_LIST.getGroupId()))
```
Adam-: Is this for adding lookup to ignore list? This should probably be in a separate PR

Rami-J: Yes, I made a separate PR for it now

> Is this for adding lookup to ignore list? This should probably be in a separate PR

Adam-:
The approach this is using to detect when a player is unignored won't work if you eg. type in a player name to remove. I am unsure if removing the note is useful when unignored though, why is it being done?

Rami-J:
Oh, that is something I didn't realize. I can remove that check then. I wanted to mirror the behavior of removing a friend with a note.

Adam-:
Okay. I modified the event now for ignores too so this should just work: 608ec68ae65978988dc12301ece6d37ae7e33907

PR #10457, opened 2019-12-17
change arrow shaft xp rate to per unit xp for unit produced

Mekanation:
This is my first ever open source commit. Let me know if I need to improve anywhere.


Alexsuperfly:
This would cause the actions field to display a wildly inaccurate value since these can only be made in batches of 15.  I believe a better fix for the use case in which the issue of opened for would be to add additional fields that are the full exp value for fletching ammo from scratch.

Mekanation:
@Alexsuperfly From the thread the request came in on, the ask was for the actions to be basically a units count due to how other fletching items are calculated. 

But with my merge another issue would be the xp. This is only 0.33 to fit the formatting of the other json items, so the total units per 5 xp would only be 14.95 instead of 15. I don't know if the client rounds up units or allows for partials.

PR #10459, opened 2019-12-17
Add farming notifications

CreativeTechGuy:
Similar to the Birds house notifications, this adds similar notifications for each patch type. Once the last of that patch type has finished, a notification will be sent letting the player know what patch needs to be harvested. This follows similar logic to the summary panel but, for example, will separate trees from hardwood trees.

As part of this change, I needed to add names to PatchImplementation.java. I also added a new field, `pluralityWord` for each patch to keep the notification messages grammatically correct.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
@@ -71,10 +71,21 @@ public interface TimeTrackingConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "farmingNotification",
+		name = "Farming notification",
+		description = "Notify you when your crops are ready to harvest",
+		position = 4
+	)
+	default boolean farmingNotification()
+	{
+		return false;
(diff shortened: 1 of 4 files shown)


abextm:
I don't want notifications for this until #8381is closed and proven working

CreativeTechGuy:
@abextm Is the plan to block any new features because of a bug? Yes that would be good to fix, but I don't see how that blocks the addition of a new QOL feature. Worst case scenario, this notification will be delayed. But delayed is better than nothing.

abextm:
This is blocked on that feature because without it this feature will be horrible broken. Best case scenario is the notification is delayed, worst case scenario (and the scenario for all newish accounts) is that the notification fires too early, which is completely unacceptable for a notification.

KasparRosin:
> I don't want notifications for this until #8381is closed and proven working

#8381

Adam-:
Could we just delay the notification by the length of a growth tick? Even though not ideal it still would be a lot better than no notification.

abextm:
I would rather not delay the notification by >10 hours, that seems mostly pointless

PR #10463, opened 2019-12-17
FPS: Add range to the fps target config items

Owain94:


runelite-client/src/main/java/net/runelite/client/plugins/fps/FpsConfig.java
```diff
@@ -43,6 +44,9 @@ default boolean limitFps()
 		return false;
 	}
 
+	@Range(
+		max = 50
```
Adam-: Additionally the min should be set to `1`, as the plugin limits it to that.

Issue #10443, opened 2019-12-15
Examine plugin positive overflow


BegOsrs:
**Describe the bug**
Examine plugin is overflowing when the total price is over max integer value.

**To Reproduce**
1. Enable the examine plugin
2. Examine some item stack valued at over max integer value

**Expected behavior**
Should be displaying the real value instead of a negative value (overflowing)

**Screenshots**
https://github.com/runelite/runelite/blob/f8b801f6cdad414a66349b808f45f7eda78dd8b0/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java#L359
https://github.com/runelite/runelite/blob/f8b801f6cdad414a66349b808f45f7eda78dd8b0/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java#L379
**Environment (please complete the following information):**
 - OS: Ubuntu 18.04.3 LTS
 - RuneLite version: 1.5.42.3
 - Launcher version: 1.6.2

zacharydwaller:
I fixed this and put in a pull request. The same overflow occurs with the mouse-over price text as well

PR #10465, opened 2019-12-17
Fix integer overflow on Examine plugin price text

zacharydwaller:
Stack prices for the Examine plugin are calculated with 32-bit integers. If the stack price was over 2.147 bil (such as if you had multiple tbows or something) it would overflow and display a negative number. This fixes that by just casting the item price to a long int before calculating the stack price.

This issue is also present in the ItemPrices plugin, which I've created another pull request for.

This Fixes: #10443


zacharydwaller:
Updated description to mention which issue it fixes.

PR #10466, opened 2019-12-18
Fix integer overflow on ItemPrices price text

zacharydwaller:
Stack prices for the ItemPrices plugin are calculated with 32-bit integers. If the stack price was over 2.147 bil (such as if you had multiple tbows or something) it would overflow and display a negative number. This fixes that by just casting the item price to a long int before calculating the stack price.

This issue is also present in the Examine plugin, which I've created another pull request for.

This fixes a related issue pointed out in: #10443

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
@@ -229,12 +229,12 @@ class ItemPricesOverlay extends Overlay
 		return null;
 	}
 
-	private String stackValueText(long qty, long gePrice, long haValue, long haProfit)
+	private String stackValueText(int qty, int gePrice, int haValue, int haProfit)
 	{
 		if (gePrice > 0)
 		{
 			itemStringBuilder.append("EX: ")
-				.append(QuantityFormatter.quantityToStackSize(gePrice * qty))
+				.append(QuantityFormatter.quantityToStackSize((long) gePrice * qty))
 				.append(" gp");
 			if (config.showEA() && qty > 1)
 			{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -267,7 +271,7 @@ private String stackValueText(int qty, int gePrice, int haValue, int haProfit)
 
 			itemStringBuilder.append("</br>");
 			itemStringBuilder.append("HA Profit: ")
-				.append(ColorUtil.wrapWithColorTag(String.valueOf(haProfit * qty), haColor))
+				.append(ColorUtil.wrapWithColorTag(String.valueOf(haStackPrice), haColor))
```
makingcomment: You seem to be replacing the total alch profit by the total alch price here.

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
```diff
@@ -231,10 +231,13 @@ else if (id == ItemID.PLATINUM_TOKEN)
 
 	private String stackValueText(int qty, int gePrice, int haValue, int haProfit)
```
Adam-: You could just change geprice/havalue/haprofit to `long` and that would promote the multiplications later

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
zacharydwaller: You're right, I can fix that

runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesOverlay.java
zacharydwaller: I thought of that as well but didn't want to change the function header if I didn't have to for some reason. I'll make this change there isn't really any reason not to change it.

zacharydwaller:
Changed the method signature to use longs. This eliminates the need to calculate stack sizes separately and can just do the multiplication as before. HA Profit stack price is no longer replaced by HA stack price.

JZomerlei:
Can you edit your first box to say that it closes the issue that was opened? https://github.com/runelite/runelite/issues/10443

Thanks.

zacharydwaller:
Updated description to mention opened issue

Issue #8879, opened 2019-05-21
Pull Target Levels from Vanilla Goals for Skill Calculator


WLoumakis:
This is not related to a problem, per se, but rather a QoL-type issue. The current way the Skill Calculator works is a bit agitating (at least to me). You can click off of a skill and upon clicking back on it the target level gets wiped.

I feel as though it would be much more elegant to utilize the vanilla XP goals, similar to how the XP tracker works, to either extrapolate the target level or the target xp. Then, if the user wishes to see intermediate or post-goal levels, they can change the target level/xp.

Perhaps instead of making this the default, though I don't see why anyone wouldn't want this, an option could be added in Configuration->Skill Calculator to where you can add a clickbox to pull from vanilla goals instead of wiping each time or at least to remember the last input target level/xp. Doesn't have to be in config, but just a thought.

WLoumakis:
Since this hasn't gotten any attention and there has been a similar post (see issue [#10445](https://github.com/runelite/runelite/issues/10445)), I decided to go in and make the changes myself.  The only file changed is runelite-client/src/main/java/net.runelite.client/plugins/skillcalculator/SkillCalculator.java.

The changes can be found on [my forked version](https://github.com/WLoumakis/runelite)

Changes made:
1. Added imports for Skill and VarPlayer.
2. Changed the following two lines of code
`targetLevel = currentLevel + 1;
targetXP = Experience.getLevelForXp(targetLevel);`
to
`VarPlayer endGoal = endGoalVarpForSkill(calculatorType.getSkill());
targetXP = endGoal != null ? client.getVar(endGoal) : -1;
targetLevel = enforceSkillBounds(Experience.getLevelForXp(targetXP));`
3. Added the private static method `endGoalVarpForSkill` method from the XpTrackerPlugin.java file.

One thing to note is that I did not do any testing on this, specifically of concern would be the `targetXP = endGoal ? ...` line since it would potentially set `targetXP` to -1, although I am almost certain that would get caught in the `enforceSkillBounds` call.

All changes are currently live on my fork, but if someone can walk me through creating a pull request here then I can get that done.

WLoumakis:
Skill calculator now pulls from in-game goals, if present. See #10472.

PR #10472, opened 2019-12-18
Added target level functionality for skill calculator

WLoumakis:
See issue #8879 for a description.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -44,6 +44,8 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import net.runelite.api.Client;
 import net.runelite.api.Experience;
+import net.runelite.api.Skill;
+import net.runelite.api.VarPlayer;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.skillcalculator.beans.SkillData;
@@ -130,8 +132,18 @@ class SkillCalculator extends JPanel
 		// Update internal skill/XP values.
 		currentXP = client.getSkillExperience(calculatorType.getSkill());
 		currentLevel = Experience.getLevelForXp(currentXP);
-		targetLevel = enforceSkillBounds(currentLevel + 1);
-		targetXP = Experience.getXpForLevel(targetLevel);
+		VarPlayer endGoalVarp = endGoalVarpForSkill(calculatorType.getSkill());
+		int endGoal = client.getVar(endGoalVarp);
+		if (endGoal != -1)
+		{
+			targetLevel = Experience.getLevelForXp(endGoal);
+			targetXP = endGoal;
+		}
+		else
+		{
+			targetLevel = enforceSkillBounds(currentLevel + 1);
+			targetXP = Experience.getXpForLevel(targetLevel);
+		}
 
 		// Remove all components (action slots) from this panel.
 		removeAll();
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -121,8 +123,9 @@ void openCalculator(CalculatorType calculatorType)
 		// Update internal skill/XP values.
 		currentXP = client.getSkillExperience(calculatorType.getSkill());
 		currentLevel = Experience.getLevelForXp(currentXP);
-		targetLevel = enforceSkillBounds(currentLevel + 1);
-		targetXP = Experience.getXpForLevel(targetLevel);
+		VarPlayer endGoal = endGoalVarpForSkill(calculatorType.getSkill());
+		targetXP = endGoal != null ? client.getVar(endGoal) : -1;
+		targetLevel = enforceSkillBounds(Experience.getLevelForXp(targetXP));
```
Adam-: `Experience.getLevelForXp(-1)` throws an `IllegalArgumentException` and will crash the calculator.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -121,8 +123,9 @@ void openCalculator(CalculatorType calculatorType)
 		// Update internal skill/XP values.
 		currentXP = client.getSkillExperience(calculatorType.getSkill());
 		currentLevel = Experience.getLevelForXp(currentXP);
-		targetLevel = enforceSkillBounds(currentLevel + 1);
-		targetXP = Experience.getXpForLevel(targetLevel);
+		VarPlayer endGoal = endGoalVarpForSkill(calculatorType.getSkill());
+		targetXP = endGoal != null ? client.getVar(endGoal) : 0;
```
Adam-: This will always set the target to 0 xp/level 1 if there is no goal.

> One thing to note is that I did not do any testing on this

Can you test this?

WLoumakis: I did do a manual `targetXP = 0` before committing and it looks like it always just sets the target level to `currentLevel + 1` but I don't have any 99's so I can't test that it works always. Guess I can just throw a case in there for using the current method (i.e., `currentLevel + 1`) just to be sure. What do you think?

Adam-: If you have 99s or have no goal set the vars are just 0. I don't see why it would set the target level to currentLevel + 1 if you are removing the code which does that? It looks to me like if the target was `0` it would set `targetXP=0` and `targetLevel=1`, which would be wrong.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -121,8 +123,9 @@ void openCalculator(CalculatorType calculatorType)
...
+		targetLevel = enforceSkillBounds(Experience.getLevelForXp(targetXP));
```
Adam-: `getLevelForXp` always returns a level bound between the limits of `enforceSkillBounds`

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
WLoumakis: Yeah I was confused when I saw that too, but regardless I've gone ahead and made the necessary changes. I made sure to go in-game and test all options related to end goals and now it looks like everything works as intended. Hopefully you'll agree!

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
WLoumakis: Removed the `enforceSkillBounds` call. Note: I didn't add a `enforceXPBounds` call around the `targetXP = client.getVar(endGoal);` line since RuneScape automatically enforces the 200M XP cap within the XP tracker.

PR #10473, opened 2019-12-18
plugins: remove crystalmathlabs

### LINKED ISSUE #8619: The Crystal Math Labs plugin is broken and no longer updating the site
The crystal math labs plugin no longer updates after every logout. This can be re-created by logging in then logging out, and refreshing your user page on crystalmathlabs, and seeing that it wasn't updated.

Hydrox6:
If we'd have had external plugins back when this plugin was created, this would definitely have been one of them.

Closes #8619


abextm:
See runelite/plugin-hub#9

dekvall:
This fixes #8619

Adam-:
For now this has been combined with a more general `xp updater` plugin.

deathbeam:
I still think it should be removed as every reason why it should be removed still applies. E.g the API is still shitty, there is still huge RMT banner at the top of the website

Nightfirecat:
Superseded by 553e568.

PR #10478, opened 2019-12-19
Skill Calculator calculates XP on focusout

### LINKED ISSUE #10387: [Skill Calc] Field Entry on Unfocus
Currently in order to input a number for level or experience a user must press enter in order for the input to be registered. When attempting to change both current and target fields, it feels awkward for my hands to move over the enter key each time, instead of keeping one hand on the numbers / ctrl+v. Similarly after pasting or entering a value and then clicking into the search box I sometimes forget to hit enter to apply the changes.

There is no visual indication that a change has not been registered, which I believe is confusing design. It results me in attempting to use the calculator with incorrect values without realizing it immediately.

My preferred solution would be to attempt to enter the input when the field is unfocused. One alternative solution would be to instead attempt to enter the input upon focus of another field within the calculator, though I believe this to be less effective. A second alternative solution would be to shade the background of the field when it has changed from the currently applied value, perhaps a shade of yellow or red, to indicate that the field has been changed but that the change has not yet been applied.

zacharydwaller:
Closes #10387 

Currently in order to input a number for level or experience a user must press enter in order for the input to be registered. This change makes it so that the user only needs to unfocus the text input for it to calculate the XP.

This improves usability for users that prefer to use tab over Enter, and in cases where XP/Level values are pasted.


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -413,4 +423,44 @@ private boolean slotContainsText(UIActionSlot slot, String text)
...
+			@Override
+			public void focusGained(FocusEvent e)
+			{
+				// No action needed for focusGained
+			}
+
+			@Override
+			public void focusLost(FocusEvent e)
+			{
+				try
```
Adam-: This feels very hacky. I think it would make more sense to have one listener per field and make this just callback to a Consumer/Callable. That way it could work similar to the action listeners- `field.addFocusListener(buildListener(e ->onFieldTargetXPUpdated()))`.  Also if you only want to override one method you can use `FocusAdapter`.

zacharydwaller:
Back from vacation - I replaced the FocusListener with a FocusAdapter so I wouldn't have to override the focusGained. I also removed all the if/elses, the FocusAdapter is just passed a Consumer which is calls on focusLost. This lets adding a focus listener have the same usage as adding any other listener.

PR #10480, opened 2019-12-20
raids:  allow "unknown" to evaluate to both unknown room types

Alexsuperfly:
match the white/blacklisted room against a lax contains check on the room name itself which allows the old way of specifying a unknown room to still function for people as expected when highlighting it on the overlay


Adam-:
The full length is still used for !layout and I'd prefer for that to show (puzzle) etc instead as there is more room. How does (p)/(c) look like on the overlay?

Alexsuperfly:
the point of this change is to make it so that the overlay displays the same name that is tracked internally for denoting a whitelisted/blacklisted room.  The name was shorted because if it was much longer it would start overlapping with the text on the left side.

Alexsuperfly:
changed the implementation to preserve the full room name but still allow white/blacklisting with the name shown on the overlay.

Alexsuperfly:
changed the implementation again based on discussion on discord

PR #10481, opened 2019-12-21
Added color selector for grimy herbs option

zzznz27:
This addition gives the option to highlight herbs that have not been cleaned yet 

Speicifically useful for herbs with difficult to see grimy icons


Hydrox6:
This feature feels better served by the [Inventory Tags](https://github.com/runelite/runelite/wiki/Inventory-Tags) plugin than hacking something into Item Identification

zzznz27:
Maybe but would that not cause clashes with items that have been tagged already?

This plugin already serves the use case of identifying difficult to identify items so it seemed like a good idea to extend that to both grimy and clean versions?

Hydrox6:
I can't see a situation in which tagging Grimy Herbs with Inventory Tags would conflict with an already existing tag, since you'd either tag it a colour since you want to clean it, or tag it another colour because you want to drop it instantly (or tag the clean version another colour because you want to drop it instantly).

Having an exception for *just* grimy herbs in a plugin originally designed to add something to differentiate items that have the exact same sprite just seems silly imo, especially when another plugin already does this specific differentiation

zzznz27:
The way I imagined it would work is by using the item id to automatically identify herbs that are grimy. This is how I'd see a conflict happening 

It could be left up to the user to tag each herb but I think this would be more useful

I also wouldn't rule out adding more to this plugin because it's not the original intention

This feature still falls under the umbrella of item identification so it makes sense to use it this way a f it also extends to banked herbs so it cant just be inventory tags

runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
```diff
@@ -46,24 +46,39 @@
...
+	GRIMY_TOADFLAX(Type.GRIMYHERB, "Toad", "TOA", ItemID.GRIMY_TOADFLAX),
+	GRIMY_IRIT(Type.GRIMYHERB, "Irit", "I", ItemID.GRIMY_IRIT_LEAF),
+	GRIMY_AVANTOE(Type.GRIMYHERB, "Avan", "A", ItemID.GRIMY_AVANTOE),
+	GRIMY_KWUARM(Type.GRIMYHERB, "Kwuarm", "K", ItemID.GRIMY_KWUARM),
+	GRIMY_SNAPDRAGON(Type.GRIMYHERB, "Snap", "S", ItemID.GRIMY_SNAPDRAGON),
+	GRIMY_CADANTINE(Type.GRIMYHERB, "Cadan", "C", ItemID.GRIMY_CADANTINE),
+	GRIMY_LANTADYME(Type.GRIMYHERB, "Lanta", "L", ItemID.GRIMY_LANTADYME),
+	GRIMY_DWARF_WEED(Type.GRIMYHERB, "Dwarf", "D", ItemID.GRIMY_DWARF_WEED),
+	GRIMY_TORSTOL(Type.GRIMYHERB, "Torstol", "TOR", ItemID.GRIMY_TORSTOL),
 	//Saplings
```
KasparRosin: Add a line break above

zzznz27: line break added

Adam-:
This is definitely doable just with invtags. This feels kind of weird that *only* the grimy herbs have a configurable text color though.

zzznz27:
I can change it to work with inventory tag if you like but would that be a bit weird with banked herbs?

My understanding is that inventory tags don't apply to anything in the bank

Adam-:
It's not hard to make it show on the bank. I don't remember why we don't do that, I think because most of the time its not necessary. Maybe could make it an option.

zzznz27:
General consensus is that this should be in inventory tags 

Made that change but made the highlight apply to the outline rather than text to avoid clashes with existing item identification

Adam-:
I wouldn't make a special group for this, and just use the existing groups. That's what they're for. Instead I would just make an option to show the invtags in the bank.

Nightfirecat:
Per Adam's comment, we're not looking to have special highlighting specifically for grimy herbs. An option to show inventory tags in the bank would be fine instead though.

Issue #10309, opened 2019-11-22
Hide consumable stats tooltip when there is no effect


wizguin:
**Describe the solution you'd like**
Checkbox to hide consumable stats when there is no effect, e.g when food heals +0 hp

**Describe alternatives you've considered**
Turning off consumable stats altogether, but I play multiple accounts so have to toggle it depending on what I want to see (fishing on main so don't want to see +0 hp for roe and caviar, but doing combat on alt so want to see optimal healing for food).

**Additional context**
![image](https://user-images.githubusercontent.com/16276645/69429300-20025500-0d2b-11ea-92a7-28858332a11e.png)

SirGirion:
Hey I will be taking this feature request.

PR #10488, opened 2019-12-24
Add toggle to hide consumable stat boosts if they wouldn't result in a change

SirGirion:
Issue #10309

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
@@ -125,6 +125,10 @@ public class ItemStatOverlay extends Overlay
 
 				for (final StatChange c : statsChanges.getStatChanges())
 				{
+					if (config.hideNoConsumableChange() && c.getPositivity() == Positivity.NO_CHANGE)
+					{
+						continue;
+					}
 					b.append(buildStatChangeString(c));
 				}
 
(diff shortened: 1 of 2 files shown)


dekvall:
I don't think that this needs to be configurable. Imo the tooltip entry for the stat should just be omitted in `buildStatChangeString` if there's no relative change.

SirGirion:
Removed the configuration option, will just be omitted if no relative change.

dekvall:
This still removes information about theoretical and absolute stat changes from the tooltip if there is no relative change. At least the theoretical info should still show.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
```diff
@@ -259,14 +259,14 @@ private String buildStatChangeString(StatChange c)
 		StringBuilder b = new StringBuilder();
 		b.append(ColorUtil.colorTag(Positivity.getColor(config, c.getPositivity())));
 
-		if (config.relative())
+		if (config.relative() && c.getPositivity() != Positivity.NO_CHANGE)
```
dekvall: I would just replace `config.relative()` with a variable like `boolean showRelative = config.relative() && c.getPositivity() != Positivity.NO_CHANGE`, and use that in it's place everywhere in this method. So that the setting appears as completely off when there is no change. This will also make it not add the unnecessary parentheses around the absolute value if there is no change and only that and the relative option are selected.

Adam-:
I think all of the checks for `config.relative() ` would need to be replaced here. What you have currently can emit a string like `" (absolute) stat"` which seems wrong?

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatOverlay.java
SirGirion: I think I've gotten the gist of what you were suggesting. Sorry for the extreme delay; real life got in the way

abextm:
This should not just say ` Hitpoints` if you only have relative on

PR #10499, opened 2019-12-26
xp tracker: don't unpause on failed login

### LINKED ISSUE #10491: XP Tracker: Unpause During Failed Login Attempt
**Describe the bug**
The XP Tracker feature to Pause on Logout does not function properly. Attempting to login but using the wrong password will unpause.

**To Reproduce**
Easy to reproduce. 
Enable XP Tracker and Pause on Logout. 
Gain some xp. 
Logout
Attempt to log back in with the incorrect password. 
Watch the tracker and observe as the skills are unpaused between the attempted login and the error message stating that you entered the wrong password.

**Expected behavior**
One would expect the unpause to occur after a successful login, not after a login attempt, to avoid this.

**Screenshots**

 - OS: Windows
 - RuneLite version: Current (12/24/19)
 - Launcher version: Current (12/24/19)

**Additional context**
None

dekvall:
If the wrong password is entered, the xp tracker unpauses for a brief moment since the `GameState` is no longer `LOGIN_SCREEN`.

Closes #10491

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
@@ -670,7 +670,19 @@ public class XpTrackerPlugin extends Plugin
 			xpPauseState.tickXp(skill, skillExperience, xpTrackerConfig.pauseSkillAfter());
 		}
 
-		xpPauseState.tickLogout(xpTrackerConfig.pauseOnLogout(), !GameState.LOGIN_SCREEN.equals(client.getGameState()));
+		final boolean loggedIn;
+		switch (client.getGameState())
+		{
+			case LOGIN_SCREEN:
+			case LOGGING_IN:
+			case LOGIN_SCREEN_AUTHENTICATOR:
+				loggedIn = false;
+				break;
+			default:
+				loggedIn = true;
+				break;
+		}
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -670,7 +670,17 @@ public void tickSkillTimes()
...
-		xpPauseState.tickLogout(xpTrackerConfig.pauseOnLogout(), !GameState.LOGIN_SCREEN.equals(client.getGameState()));
+		switch (client.getGameState())
+		{
+			case LOGIN_SCREEN:
+			case LOGGING_IN:
+			case LOGIN_SCREEN_AUTHENTICATOR:
+				xpPauseState.tickLogout(xpTrackerConfig.pauseOnLogout(), false);
+				break;
+			default:
+				xpPauseState.tickLogout(xpTrackerConfig.pauseOnLogout(), true);
```
Nightfirecat: This duplication isn't great. Maybe this change could become:

```java
final boolean loggedIn;
switch(client.getGameState())
{
	case LOGIN_SCREEN:
	case LOGGING_IN:
	case LOGIN_SCREEN_AUTHENTICATOR:
		loggingIn = false;
		break;
	default:
		loggingIn = true;
		break;
}
xpPauseState.tickLogout(xpTrackerConfig.pauseOnLogout(), loggingIn);
```

PR #10508, opened 2019-12-28
tearsofguthix: add timer

hiwilliam36:
Closes https://github.com/runelite/runelite/issues/10349
Please tell me if there is anything I need to fix, as this is my first time doing something like this.


Adam-:
Is there a reason why you chose to use an overlay instead of a `Timer` infobox? They are more compact and the code which manages them largely already exists.

hiwilliam36:
I don't really have a good reason. I used them because the tutorial I watched used an overlay so I was more comfortable using that. If you want I can try to use a timer instead.

Adam-:
That would be better, and make it consistent with all of the other timers in the client.

hiwilliam36:
I have changed it to use a Timer infobox. Please tell me if there is anything else I need to change.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixPlugin.java
```diff
@@ -112,4 +142,68 @@ public void onDecorativeObjectDespawned(DecorativeObjectDespawned event)
...
+			timerRunning = false;
+		}
+	}
+
+	int getQp()
+	{
+		try
+		{
+			int qp = client.getVar(VarPlayer.QUEST_POINTS);
+			chatClient.submitQp(client.getLocalPlayer().getName(), qp);
```
Adam-: This code is for the `!qp` command. It doesn't make sense here.

Adam-:
varbit 5099 holds the timer used to draw the in-game progress bar. It is initialized to your QP and decreases by 4 each tick. You could use that to initialize and infobox time.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixPlugin.java
```diff
@@ -24,30 +24,38 @@
...
-import java.time.Instant;
-import java.util.HashMap;
-import java.util.Map;
-import javax.inject.Inject;
+import com.google.inject.Provides;
 import lombok.Getter;
-import net.runelite.api.Client;
-import net.runelite.api.DecorativeObject;
-import net.runelite.api.ObjectID;
+import net.runelite.api.*;
```
Nightfirecat: Don't star-import non-static imports. [See our code conventions for more info.](https://github.com/runelite/runelite/wiki/Code-Conventions)

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixPlugin.java
```diff
@@ -112,4 +142,68 @@ public void onDecorativeObjectDespawned(DecorativeObjectDespawned event)
 		DecorativeObject object = event.getDecorativeObject();
 		streams.remove(object);
 	}
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		if (!hasCreatedTimer)
+		{
+			if (client.getLocalPlayer().getWorldLocation().getRegionID() == 12948 && client.getLocalPlayer().getWorldLocation().getX() >= 3254 && client.getLocalPlayer().getWorldLocation().getX() <= 3262)
```
Nightfirecat: All of these worldpoint bounds should be made into a constant `WorldArea`.

PR #10513, opened 2019-12-28
NMZ: health points notification

### LINKED ISSUE #10502: NMZ Notification for HP threshold
It would be nice for the window to alert when HP is above a value you don't want to be above when using absorption potions. 

A notification for when HP is equal to a number character set 0,1 would disable or something.

JZomerlei:
Closes #10502 
Config:
image


Nightfirecat:
This is already available in the client by combining the Regen Meter plugin's `Hitpoint Regen Notification (seconds)` option with the Nightmare Zone plugin's `Overload notifications` option.

PR #10526, opened 2019-12-31
Remove duplicate Item Count in GroundItem notifs

### LINKED ISSUE #10319: Ground Item Notifications have duplicated quantities for no reason
Currently, the notification has the format
```
[{player}] received a highlighted drop: {name} x {quantity} ({quantity})
```
With the only difference being that the bracketed quantity gets passed through `QuantityFormatter#quantityToStackSize` first. This results in duplicated quantity for no discernible reason.

One of them should be removed, though the `Lots!` stuff should be kept. If anyone has any strong feelings about which one should be kept, leave a comment.

oneilljos:
Prevent duplicate quantities in Ground Item Notifications

This resolves #10319 

- The `" x {quantity}"` part of the string was removed, because the `"({quantity})"` style is what is used for text on items on the ground, so this aligns with that style.
- Closing `!` was also removed, because it looked a little odd in the notification text (Also duplicated in the `Lots!` case) - I'm fine with whatever consensus or general preference is here

## Before Changes
image

## After Changes
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -663,9 +663,6 @@ public class GroundItemsPlugin extends Plugin
 
 		if (item.getQuantity() > 1)
 		{
-			notificationStringBuilder.append(" x ").append(item.getQuantity());
-
-
 			if (item.getQuantity() > (int) Character.MAX_VALUE)
 			{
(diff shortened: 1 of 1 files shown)


dekvall:
Thanks for the PR!

I don't think that the amount in the notification text needs to be stack-formatted, we don't need to save space here like we have to in the overlay. I also prefer the  `x {quantity}!` format because it doesn't look as compact as the one with parentheses.
```
[{player}] received a highlighted drop: {name} x {quantity}!
```

PR #10532, opened 2020-01-02
gpu: Add GPU service for external plugins

slaytostay:
Add GPU service to set resource loader, init uniforms and set uniforms on draw call for custom shaders
This is an alternative to https://github.com/runelite/runelite/pull/10495


abextm:
This is basically the inverse of how services are supposed to work.
This makes every plugin duplicate every shader entirely that wants to use this, which it probably shouldn't. Also it disallows >1 plugin from inheriting this, and if any plugin breaks, it will cause the entire gpu plugin to break without any way of the user knowing it was an external plugin.

slaytostay:
I agree, I'll close my PRs and revert the original region locker PR on the plugin hub to the version with the duplicated gpu plugin. I don't think there's a clean way of doing this.

Issue #1123, opened 2018-03-26
Ground Items: Add support for stack counts


ShaunDreclin:
Would love to be able to add "Coins<100", "Feather<20", "Fire rune<50" etc to my hidden items.

Infinitay:
Hide GE/HA value if < 100?

ShaunDreclin:
That would hide items that I want to pick up that are worth less than 100gp on the GE. (Ironman btw)

Hiding everything below one price isn't as useful as hiding specific items by their stack counts

Nemo296:
I'd like to voice my support for a feature like this. I like to loot stackable items that are worth over X amount and having items that are stackable over X price highlighted or stackable items under that price hidden would be a super helpful feature!

Thanks.

PR #10533, opened 2020-01-02
Ground Item: show ammo if count is greater than X

Blackberry0Pie:
Adds a config option to now show piles of ammo with fewer than X ammo in them

Note: untradable ammo such as broad arrows will always show if "always show untradable items" is enabled.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -369,4 +369,15 @@ public interface GroundItemsConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "showAmmoCount",
+			name = "Show ammo count",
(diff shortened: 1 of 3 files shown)


dekvall:
I think that the problem should be approached in a more general sense, eg. support for stack counts when specifying highlighted items( #1123). Nowhere else does the plugin take into account what an item is used for (ammunition) when determining if it should be highlighted or not. But that issue has been open for a long time, so maybe this can be added in the meantime(?)

deathbeam:
I agree that this should be implemented in more generic sense, implementing it like this just do not makes sense

FrankFlitton:
Re @dekvall: if the item.contains checks are removed it should work in a more general sense. It would be applied to all items, not just ammunition, but it would be useful none the less. Is there a property that declares if the item is stackable? Would love to contribute to this new feature.

dekvall:
@FrankFlitton there is `ItemComposition#isStackable` you could use to check that. But imo adding support for `<` and `>` in the hidden and highlight list is superior to a config like this.

PR #10535, opened 2020-01-02
Wintertodt interrupt options

Blackberry0Pie:
Adds interrupt options for full inv and idle


xTheEc0:
I believe there is a general 'user is idle' interrupt so having a wintertodt one is pointless?

However, there is no interrupt when you finish fletching all roots so maybe add that? (been bothering me lately : P)

Adam-:
I don't understand this idle setting; it looks like you are making it so if you are already idle and get hit by cold/snowfall/brazier, it will notify you again telling you you are idle.

Adam-:
taken #10657 instead

PR #4914, opened 2018-08-15
Chambers of Xeric Thieving Bat Finder plugin

TimLehner:
This is a plugin for the Thieving room in Chambers of Xerics

When thieving, several chests will contain poison instead of grubs, and one chest will contain bats. This plugin keeps track of the chests as they are opened. Once a chest has been opened it will either contain grubs or be empty, if empty it either contained the bats and will no longer close, or it has poison and will close.

While the bats have not been found, this plugin marks all chests that have never been opened so teams can quickly and efficiently check all chests. 

image

If a poison chest is found, this is marked with a different colour.

image

After the bats have been found, chest markers disappear except for the poison chests.

image

Note: Your character must be in the same loaded area as any other player that opens the chest

PR #10540, opened 2020-01-03
raids plugin: add bats locator

### LINKED ISSUE #2592: Bats finder in raids
Would it be possible to add a plugin that helps you find bats? There is a website that it could be based off 
https://dikkenoob.github.io

### LINKED ISSUE #3115: Bat Finder for Raids Thieving Room
The bats inside the thieving room in raids are not random and there is a site where you can enter in what chests are poisoned and it will show you where the bats are.

There are 3 room layouts so the plugin would have to figure out the room layout and then take into account when you\someone else finds a poisoned chest and similar to the hot\cold plugin slowly figure out where the bats are.

https://twitter.com/EndlessDawnRS/status/998991141159751685
https://t.co/4EoAzdLvgW
https://dikkenoob.github.io/

chestnut1693:
This addition to the raids plugin will help locate bats in the thieving room. The chests are overlayed with transparent coloured dots. Unvisited chests are magenta, grubs chests are blue but these will be hidden, poison chests are green and the bats chest is white. Unvisited chests that are most likely to contain bats or poison have a bigger dot and are not transparent. Previously, indices of the chests were used instead of dots and an enlarged indices would represent the most likely chests. [Tim Lehner's](https://github.com/runelite/runelite/pull/4914) pull request has been used as a reference and the dots looked more promising. Tim's pull request was also used as a reference to make sure the functionality of this bats locator would surpass or atleast be equal to that of Tim's.

Initial state of the room with a left turn.
image

State of the room when a poison chest has been found.
image

State of the room when the bats chest has been found, in this case all three poison chests were found before the bats chest.
image

The bats locator preemptively hides chests with grubs from the overlay. Finding a poison chest will mark all other chests as grubs that are not in the same solution sets as the poison chest. This is all based on the solution sets given by [Not Noob](https://dikkenoob.github.io/) and [Endless Dawn](https://docs.google.com/spreadsheets/d/1eEf8yI_MG6cKuIXSqsMKO8YfuQSn5pxWtGWxSDaqUsA). It could happen that the chests in the thieving room are not yet supported by the solution sets. If this happens, the bats locator will fallback and hide the found chests containing grubs and show the found chests containing poison or bats. The poison and bats will not be overlayed with a dot but with a number. This number is the index of the chest starting from 0. The indices go from near to far and left to right when facing the entrance of the thieving room. This is a bit different from the existing online tools but it is consistent. Drawing numbers instead of dots will make it easier for people to report the invalid or missing solution set. It is advised to search all the chests before reporting. A few solution sets have been added or enhanced during development and the solution sets from the form of Endless Dawn's sheet have not been added. Some solution sets coming from Not Noob and Endless Dawn contain only two or three chests while most contain four and there might be a duplicate somewhere. Transforming the solution sets was a manual labor and mistakes could have been made.

A bats chest has been found in the room with a right turn. 
image

A chest is opened containing poison which was not expected by the bats locator so it fallsback. None of the other chests were opened otherwise they would not show a transparent magenta dot.
image

The bats locator can handle opened chests when starting the plugin. The bats locator can not differentiate a poison chest from a bats chest when both are open when starting the plugin. The bats locator can be enabled by using the checkbox in the raids plugin menu. The checkbox is only there to show or hide the overlay, the bats locator registers spawning chests regardless of the checkbox.

Chest numbers of [right turn](https://user-images.githubusercontent.com/47870624/71784595-49cebd00-2ff5-11ea-909f-f06fc2210fb9.png), [left turn](https://user-images.githubusercontent.com/47870624/71784591-3cb1ce00-2ff5-11ea-9802-b5257508edfe.png) and [straight](https://user-images.githubusercontent.com/47870624/71784556-f52b4200-2ff4-11ea-8822-dda26edcf2ae.png) room.

Credits to Tim Lehner, Not Noob, Endless Dawn and myself.

Closes #2592
Closes #3115


runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/WorldPointComparator.java
```diff
@@ -0,0 +1,120 @@
...
+		this.rotation = rotation;
+	}
+
+	@Override
+	public int compare(WorldPoint worldPointA, WorldPoint worldPointB)
+	{
+		switch (rotation)
+		{
+			case 0:
+				if (worldPointA.getY() < worldPointB.getY())
```
Adam-: You can use Java comparator chains instead of building your own: `Comparator.comparing(WorldPoint::getY).thenComparing(Comparator.comparing(WorldPoint::getX))`

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsOverlay.java
```diff
@@ -0,0 +1,100 @@
...
+			for (WorldPoint chestLocation : batsLocator.getChestLocations())
+			{
+				i++;
+				Chest chest = batsLocator.getChestMap().get(chestLocation);
+				if (chest.getState() == Chest.State.GRUBS)
+				{
+					continue;
+				}
+				Player player = client.getLocalPlayer();
+				if (player != null)
```
Adam-: This and the condition below do not change between iterations. Pull them out of the loop.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/Chest.java
```diff
@@ -0,0 +1,32 @@
+package net.runelite.client.plugins.raids.bats;
+
+import java.awt.Color;
+import lombok.Getter;
+import lombok.Setter;
+
+public class Chest
```
Adam-: Is there a reason why the chest location isn't being stored here? It looks like it would simplify things a bit.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/WorldPointComparator.java
chestnut1693: How would it work between case 0 and 2? They both compare Y first but case 0 returns -1 when Y of A is less than Y of B and case 2 returns -1 when Y of A is greater than Y of B.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/Chest.java
chestnut1693: It didn't seem necessary.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/WorldPointComparator.java
Adam-: Case 2 looks the same case 0  only .reverse(). I think you can replace this entire class with a method in BatsLocator that returns a `Comparator` based on the input `rotation`.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
```diff
@@ -0,0 +1,503 @@
...
+	public void poisonSplatEvent(WorldPoint worldPoint)
+	{
+		Chest chest = chests.get(worldPoint);
+		if (chest != null)
+		{
+			chest.setTickPoison(client.getTickCount());
+		}
+	}
+
+	//This does have a delay switching from poison chest to bats chest which can be observed by the color change at the chest, only when starting the plugin.
```
Adam-: It looks like you have the same comment here twice, though I don't really understand either of them.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
```diff
@@ -0,0 +1,503 @@
...
+	public BatsLocator(Client client)
+	{
+		this.client = client;
+	}
+
+	public void troughSpawnEvent(GameObject trough)
+	{
+		if (roomType == null && rotation == -1)
+		{
+			int plane = trough.getPlane();
```
Adam-: This is all just duplicated from `WorldPoint.fromLocalInstance`, use it instead.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/WorldPointComparator.java
chestnut1693: The list was changed from containing `WorldPoint` to `Chest`. Sorting had to be done with `Function` as `Chest::WorldPoint::getY` was not possible. The approach came from Stack Overflow and it seems to work.

`case 1: Comparator.comparing(((Function<Chest, WorldPoint>)Chest::getLocation).andThen(WorldPoint::getX)).thenComparing(Comparator.comparing(((Function<Chest, WorldPoint>)Chest::getLocation).andThen(WorldPoint::getY)).reversed());`

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
chestnut1693: I'm not interested in the world point but in the variable `instanceX` as this uniquely identifies the room. The world point of the trough can be different each time the raid chambers are entered but `instanceX` is always the same. If i remember correctly, the variable `instanceX` is the base of a chunk where the trough spawns in. I've seen this approach used elsewhere.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
chestnut1693: A color change from green (poison) to white (bats) is observerd when the raids plugin is started. The bats chest can't be differentiated from the poison chests so it is coloured green when it spawns. The next tick arrives and no poison splat has occured so the bats locator thinks it is a bats chest. It could happen that the bats chest and two poison chests are open when the raids plugin is started which results in white dots, even for the poison chests. It is ofcourse unlikely that the raids plugin is switched on during a raid but I thought it was worth mentioning.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
```diff
@@ -0,0 +1,503 @@
...
+		{
+			switch (chestObject.getId())
+			{
+				case OPENED_POISON_OR_BATS:
+					poisonBatsChests.add(chest);
+					openChest(chest, Chest.State.POISON);
+					break;
+				case OPENED_WITHOUT_GRUBS:
+				case OPENED_WITH_GRUBS:
+					grubsChests.add(chests.get(chestLocation));
```
Adam-:
```suggestion
					grubsChests.add(chest);
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/WorldPointComparator.java
Adam-:
```java
Comparator.comparing(((Function<Chest, WorldPoint>)Chest::getLocation).andThen(WorldPoint::getY))
							.thenComparing(((Function<Chest, WorldPoint>)Chest::getLocation).andThen(WorldPoint::getX));
```

Reduces to simply:

```java
Comparator.comparing(Chest::getLocation, Comparator.comparing(WorldPoint::getY).thenComparing(WorldPoint::getX));
```

Can maybe static import `Comparator.comparing` if you want.

chestnut1693:
I found that it is relatively easy to find not known solution sets but it is hard to detect solution sets that are partly incorrect. For example in the left room, (9, 34, 53, 62) comes from Not Noob but the solution I found in game was (13, 34, 53, 62). Both could be correct but it is hard to know if (9, 34, 53, 62) is incorrect.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
Adam-: What you call `instanceX` is simply the base of the template chunk in world coords. So it would just be `worldPoint.getX() & -CHUNK_SIZE`

Adam-:
If we don't know the correctness of completeness of this, maybe we just add it into external plugins?

chestnut1693:
> 
> 
> If we don't know the correctness of completeness of this, maybe we just add it into external plugins?

I like to see it with the main plugins but if that is not option then external plugins will have to do.

Adam-:
I'd be fine with it in the main client if it were well understood and the code was more organized. It is hard for me to handle bug reports when a large portion of the solver logic is a bunch of magic numbers that may or may not be complete or correct.

chestnut1693:
Would it be possible to upload the solution set once three chests with poison and one chest with bats is found? Solutions sets could be downloaded when runelite is started and the bats locator can load them in. However, this will only work when solution sets consist of four chests and Jagex doesn't change the solution sets.

runelite-client/src/main/java/net/runelite/client/plugins/raids/bats/BatsLocator.java
chestnut1693: I now see the difference between the world coord of the trough and the world coord returned from `WorldPoint.fromLocalInstance`. The variable I call `instanceX` is called `templateChunkX` in the function. I could remove all the duplicate stuff but I how do I get the rotation of the chunk then?

You find this more appropriate?
`LocalPoint localPoint = trough.getLocalLocation();`
`rotation = client.getInstanceTemplateChunks()[trough.getPlane()][localPoint.getSceneX() / CHUNK_SIZE][localPoint.getSceneY() / CHUNK_SIZE] >> 1 & 0x3;`
`switch (WorldPoint.fromLocalInstance(client, localPoint).getX() & -CHUNK_SIZE) {}`

Adam-:
@Trevor159 can you take this and refactor it?

Trevor159:
> @Trevor159 can you take this and refactor it?

I personally think this is better suited for external plugins, but if you want it in release idm.

Adam-:
That would be fine too, I just didn't want this to get lost.

Trevor159:
@chestnut1693 do you mind PRing this to the plugin-hub?

chestnut1693:
I do not mind but I first need to get familiar with the plugin-hub and figure out why downloading the plugin manifest keeps failing.

Trevor159:
> I do not mind but I first need to get familiar with the plugin-hub and figure out why downloading the plugin manifest keeps failing.

you should come in the discord if you need help with stuff like that, there is a developer channel to ask questions in. Also the readme here has a guide for the plugin-hub https://github.com/runelite/plugin-hub

chestnut1693:
added as external plugin

PR #10542, opened 2020-01-03
Cooking calculator: add missing foods

bytemunch:
add vscode to gitignore
add missing foods to skill_cooking.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ project.properties
 .project
 .settings/
 .classpath
+.vscode
+.factorypath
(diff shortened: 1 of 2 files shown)


dekvall:
The gitgnore addition is unrelated to the skill calculator and should be a separate commit.

In my opinion adding most of these will just add clutter to the UI, i don't think anyone does Fishcake or Roasted Chompies etc for xp. However, there is a number of unusual items in the calculator already so i guess one could make the point that they should be added for completeness.

I do appreciate you fixing the level requirement on anchovies and xp for spicy sauce though.

bytemunch:
Sorry, first time contributing to anything so I'm not up to date with the right way to do things. I was only going in to change the anchovy level but then saw some stuff missing; I'd argue for completeness based on different people playing different ways, I like to be as efficient as possible but some people enjoy the more obscure stuff you can do in the game :)

Issue #9574, opened 2019-08-04
Recolor Prifddinas Agility Course Portals


Mollium:
**While doing the new Prifddinas Agility Course, you’ll randomly get portals to click on that accelerate your laps. 

Runelite colors them in the same color that’s given to the other agility obstacles on the course. This makes it harder to spot them thus resulting in missclicking and skilling them.

**A simple solution would be to give them another color, like blue for example. Something to distinguish them from the standard obstacles.

**Another great way could be to make an arrow over the portals: just something to make it easier to spot a portal.

melkypie:
Added in #10543

PR #10543, opened 2020-01-03
Add prifddinas portal highlighting to agility plugin

melkypie:
Adds functionality described in #9574


Adam-:
It looks like this is just making the color configurable, the plugin already highlights the portals.

Issue #9736, opened 2019-08-25
Right-Click Save-as for the screenshot button


stsguitar23:
**Is your feature request related to a problem? Please describe.**

I would like to organize my screenshots better.

**Describe the solution you'd like**

Being able to right click, and save a screenshot to a specific folder, or save it with
the name "kurasks gear"
So next time i get kurasks as a slayer task I can just search my screenshots folder
for "kurasks gear"

PR #10544, opened 2020-01-03
Added save as option when right clicking the screenshot button

melkypie:
Adds #9736

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -26,7 +26,6 @@ package net.runelite.client.plugins.screenshot;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -212,25 +219,43 @@ protected void startUp() throws Exception
...
+			{
+				Desktop.getDesktop().open(SCREENSHOT_DIR);
+			}
+			catch (IOException ex)
+			{
+				log.warn("Error opening screenshot dir", ex);
+			}
+		});
+
+		if (config.uploadScreenshot() == UploadStyle.NEITHER)
```
Adam-: I don't see why this would need to require having upload style set to neither considering its a right click option. Also this wouldn't work anyway since the config is only checked on startup.

melkypie: Removed it

Adam-:
why has the popup map been pulled up to the plugin?

melkypie:
Was there because of the IF before but since that is not needed I have removed and put the popups in where they should be.

PR #10551, opened 2020-01-05
Show raids loot value at the end of the raid

melkypie:
Adds the functionality to view raids loot value when opening the chest
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
@@ -47,6 +47,8 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.InstanceTemplates;
+import net.runelite.api.InventoryID;
+import net.runelite.api.ItemContainer;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MessageNode;
 import net.runelite.api.NullObjectID;
@@ -58,6 +60,8 @@ import net.runelite.api.VarPlayer;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.VarbitChanged;
+import net.runelite.api.events.WidgetLoaded;
+import net.runelite.api.widgets.WidgetID;
 import net.runelite.client.callback.ClientThread;
 import net.runelite.client.chat.ChatColorType;
 import net.runelite.client.chat.ChatCommandManager;
@@ -69,6 +73,7 @@ import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.ChatInput;
 import net.runelite.client.events.ConfigChanged;
 import net.runelite.client.events.OverlayMenuClicked;
+import net.runelite.client.game.ItemManager;
 import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -215,6 +232,53 @@ public void onConfigChanged(ConfigChanged event)
 		clientThread.invokeLater(() -> checkRaidPresence(true));
 	}
 
+	@Subscribe
+	public void onWidgetLoaded(WidgetLoaded event)
+	{
+		if (!config.showLootValue() ||
+			event.getGroupId() != WidgetID.CHAMBERS_OF_XERIC_REWARD_GROUP_ID ||
```
Adam-: Check widget id first, it is cheaper.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -163,6 +177,8 @@
 	@Getter
 	private boolean inRaidChambers;
 
+	private boolean chestOpened = false;
```
Adam-:
```suggestion
	private boolean chestOpened;
```

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -215,6 +232,53 @@ public void onConfigChanged(ConfigChanged event)
...
+		final Collection<Item> rewardItems = Arrays.stream(rewardItemContainer.getItems())
+			.filter(item -> item.getId() > 0)
+			.map(item -> new Item(item.getId(), item.getQuantity()))
+			.collect(Collectors.toList());
+
+		int rewardValue = 0;
+
+		for (Item reweardItem : rewardItems)
+		{
+			if (UNTRADABLE_ITEMS.contains(reweardItem.getId()))
```
Adam-: Why is this here?

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -215,6 +232,53 @@ public void onConfigChanged(ConfigChanged event)
...
+		{
+			return;
+		}
+
+		final Collection<Item> rewardItems = Arrays.stream(rewardItemContainer.getItems())
+			.filter(item -> item.getId() > 0)
+			.map(item -> new Item(item.getId(), item.getQuantity()))
+			.collect(Collectors.toList());
+
+		int rewardValue = 0;
```
Adam-: This should maybe be a `long`.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
melkypie: Guess it is not needed, was thinking about it not checking item price if it is an untradable item but the item prive would be 0 anyways so it is useless, will take it out

Adam-: The item manager has some logic to apply prices to untradeables, but if you don't want it to there is a `ignoreUntradeableMap` flag.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
melkypie: QuantityFormatter.quantitotyToRsDecimalStack requires an int, that is why I left it as int

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
melkypie: Removed the check

PR #10557, opened 2020-01-06
hiscore api: add Callisto and Cerberus

Hydrox6:
Jagex finally added them to the `index_lite` format :tada:


Hydrox6:
Updated the test

Issue #6700, opened 2018-11-26
Supply usage tracker


CreamOSRS:
Keeps track of the supplies you use on task , bossing or pking so you know exactly how much you earn.  Used along side Loot tracker or an addition to it.

raiyni:
Duplicate #4126

raiyni:
Duplicate of #6700

PR #10560, opened 2020-01-06
Consumed items tracker

tgod5565:
I believe this would be a huge QOL addition if we could somehow have a consumed item tracker that shows how much food you've eating and the specific potions you've consumed.


Hydrox6:
First, this is not how you make an issue
Second, Duplicate of #6700
Third, there's a plugin in the plugin hub that attempts to do this.

PR #10565, opened 2020-01-08
Add script fired events and add script inspector devtool

Trevor159:
Relies on !96

Picture of the dev tool 

Swing sucks and I suck at swing.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/events/ScriptPreFired.java
+++ b/runelite-api/src/main/java/net/runelite/api/events/ScriptPreFired.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2020, Trevor <https://github.com/Trevor159>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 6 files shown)


runelite-api/src/main/java/net/runelite/api/events/ScriptPreFired.java
```diff
@@ -0,0 +1,45 @@
...
+{
+	/**
+	 * The script id of the invoked script
+	 */
+	private final int scriptId;
+
+	/**
+	 * The source of the script invoke, this will be null unless
+	 */
+	private Widget source = null;
```
abextm: this should probably hold the whole scriptevent

runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
```diff
@@ -0,0 +1,493 @@
...
+import net.runelite.client.ui.ClientUI;
+import net.runelite.client.ui.ColorScheme;
+import net.runelite.client.ui.DynamicGridLayout;
+import net.runelite.client.ui.FontManager;
+import net.runelite.client.util.Text;
+
+@Slf4j
+public class ScriptInspector extends JFrame
+{
+	private final static String DEFAULT_BLACKLIST = "3174,1004";
```
Nightfirecat: Name these in ScriptID so they're not magic numbers.

runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
```diff
@@ -0,0 +1,493 @@
...
+	private final static int MAX_LOG_ENTRIES = 10000;
+
+	private final Client client;
+	private final EventBus eventBus;
+	private final ConfigManager configManager;
+
+	private final JPanel tracker = new JPanel();
+	private ScriptTreeNode currentNode;
+	private int lastTick = 0;
```
Nightfirecat: 0 is the initial value for int fields.
```suggestion
	private int lastTick;
```

runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
abextm: imo its better not in scriptid so I don't have to see them when they change, it doesn't matter for this use case.

Nightfirecat: A comment explaining the IDs here would be sufficient then. It's just awkward to leave these as magic numbers.

PR #10584, opened 2020-01-10
chatfilter: Add collapse chat message option

XrioBtw:
Option to collapse duplicate chat messages and add a coloured count at the end of the message.

illustration

animation

The count formatting (e.g. "x 2") follows the format of the collapsed menu entries in the ground items plugin.

RegEx can probably benefit my implementation, but I would need some help to formulate an expression.

If someone has a better implementation feel free to PR and merge that instead.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
@@ -25,6 +25,7 @@
  */
 package net.runelite.client.plugins.chatfilter;
 
+import java.awt.Color;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
@@ -96,5 +97,27 @@ public interface ChatFilterConfig extends Config
 	default boolean filterLogin()
 	{
 		return false;
+	}
+	
+	@ConfigItem(
+		keyName = "collapseChatMessages",
+		name = "Collapse chat messages",
+		description = "Collapses chat messages together and appends count",
+		position = 7
+	)
+	default boolean collapseChatMessages()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "chatMessageCountColor",
+		name = "Collapsed chat message count color",
+		description = "Color of the appended count on collapsed chat messages",
+		position = 8
+	)
+	default Color chatMessageCountColor()
+	{
+		return Color.decode("#FF00FF");
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
@@ -297,7 +297,10 @@ public class ChatFilterPlugin extends Plugin
 		{
 			for (MessageNode m : b.getLines())
 			{
-				if (m == null) continue;
+				if (m == null)
+				{
+					continue;
+				}
 				if (isEqual(m, node))
 				{
 					return m;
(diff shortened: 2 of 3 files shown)


XrioBtw:
and this is why it is WIP ...

Known issues:
- [ ] Clan chat join / leave messages not stacking

XrioBtw:
I have no more to add to this as of now, hence WIP has been removed.
Feel free to suggest changes.

dekvall:
I don't like how this moves chat messages around, if i say the same thing when talking to two different people the first message will be moved away from the conversation and just show `Hey! x 2` in the new conversation, altering the chat history.

I would prefer for this to just check if the last messages can be merged with the previous messages.
```
Welcome to Old School RuneScape.
You swing your pick at the rock.
You manage to mine some copper.
```
would when attempting to mine the next rock become
```
Welcome to Old School RuneScape.
You swing your pick at the rock.
You manage to mine some copper.
You swing your pick at the rock.
```
and when mined it would become
```
Welcome to Old School RuneScape.
You swing your pick at the rock. x 2
You manage to mine some copper. x 2
```

XrioBtw:
@dekvall I understand your concern, but having it like you describe leads to an (in my opinion) unwanted behaviour:
- When skilling I often use the last received chat message as an indication that I am in fact skilling (as opposed to idling because I e.g. clicked too early / in the wrong place).
- If you stack up a few duplicate messages, then receive 8 unique messages and finally receive another message that stacks up with the very first stack, the result will be no visual indication in the chat that anything happened. I guess a solution to this is to start a new stack if the old stack is outside the visible portion of the chat box.

Your suggested functionality can undoubtedly be implemented, I am just unsure if it is any better than the current functionality. I will change it if more people prefer your suggestion.

A middle ground approach is to add an option to only stack messages if duplicate of the previous one, but this does nothing for repeated alternating messages, e.g. mining.

raiyni:
It should have a configurable `collapse last n` messages probably and not do what your image shows where it shifts messages around.

dekvall:
@XrioBtw I am not sure i understand, it will still show in the chat when something new happens.
 - The message that indicates you are mining will still display. They will jump when they get merged, and i think it's noticeable when the number increments if there is only a singular action being done.
 - If you stack up duplicate messages and get unique ones, you will have to create new stack of messages since the sequence no longer matches up. The new messages will show up as usual in the chat.
```
Welcome to Old School RuneScape.
You swing your pick at the rock. x 2
You manage to mine some copper. x 2
Player: Hello!
You swing your pick at the rock.
You manage to mine some copper.
```
My main concern is with the message history reordering, an personally i think this is the cleanest way to solve it.

XrioBtw:
The chat message collapse functionality has been updated to reflect the suggestions.

- Stacking duplicate messages will no longer reorder chat history.
- A new duplicate message stack will only be created if the previous stack is outside the visible portion of the chat box (the bottom 8 messages).

Adam-:
I don't particularly like the ambiguity of
```
You swing your pick at the rock. x 2
You manage to mine some copper. x 2
```
Since there are 2 different possible situations that can create it (aabb or abab). Maybe there is a way we could express a group of messages being grouped better?

XrioBtw:
@Adam- I can't think of a way to implement that and if I could I likely would overcomplicate the implementation. Feel free to PR a new and improved version of this.

Nightfirecat:
Have we come to a consensus how we would handle interleaved messages (abba, abab, etc.)

Some behavior which might alleviate that would be to move the line for the most recently received message to the bottom (along with incrementing its counter) so users will always know the bottom message was the most recently received one.

That is, for abba, you'd see:

> b x 2
> a x 2

And for abab, you'd see:

> a x 2
> b x 2

And for a more complicated example, say abcadebac, you'd see:

> d
> e
> b x 2
> a x 3
> c x 2

dekvall:
I would still prefer not moving messages around and think a grouping approach is the best way to go about it. So that your examples would be

abba:
```
a
b x 2
a
```

abab:
```
| a
| b x 2
```
abcadebac (I don't think that this will be a very common case):
```
a
b
c
a
d
e
b
a
c
```
abcabcabcdabcd:
```
| a
| b
| c x 2
| a
| b
| c
| d x 2
```
And maybe nested (but that might become confusing for longer messages, but i think it would be useful), ababcababcababc:
```
| | a
| | b x 2
| c x 3
```
I think that the feature should primarily focus on alleviating the spam of alternating messages you get when performing actions. Hence the no collapsing of your third example.

Adam-:
added in #11556

PR #10587, opened 2020-01-11
Update graceful recovery rate bonuses

### LINKED ISSUE #10577: Run Restoration with New Graceful Effect
As per today's update, the Graceful set effect is split among the pieces. From the update notes:

"Previously, run energy restored 30% faster while wearing all 6 pieces of the Graceful outfit. Now, the hood, cape, gloves, and boots give an individual 3% boost, and the top and legs give 4% each. This adds up to a 20% run energy bonus with an additional 10% bonus for wearing the whole outfit, which matches the previous 30% boost."

melkypie:
Updates the graceful recovery rate bonuses to match up with the latest OSRS update. Fixes #10577

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
@@ -25,7 +25,9 @@
 package net.runelite.client.plugins.runenergy;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import com.google.inject.Provides;
+import java.util.List;
 import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,49 +201,60 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
-		// Calculate the amount of energy recovered every second
-		double recoverRate = (48 + client.getBoostedSkillLevel(Skill.AGILITY)) / 360.0;
+		gracefulBonus += highRecoverItems.stream().filter(p -> p).count() * 0.04;
 
-		if (isLocalPlayerWearingFullGraceful())
+		// If the bonus is 20%, player has the full graceful set effect so we add 10% more
+		// Have to account for a floating point error so we floor it down to match up
+		if (Math.floor(gracefulBonus * 100) / 100 == 1.2)
 		{
-			recoverRate *= 1.3; // 30% recover rate increase from Graceful set effect
```
dekvall: Imo this should just be replaced with something like `recoverRate *= gracefulModifier()` instead of baking it into a `getRecoverRate`

melkypie: Moved the `recoverRate` and everything that pertained to calculating it to `getEstimatedRecoverTimeRemaining()` and renamed the method to `getGracefulModifier() ` which only calculates the effect the graceful has on recovery rate

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,49 +201,62 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
+			ALL_GRACEFUL_TOPS.contains(items[EquipmentInventorySlot.BODY.getSlotIdx()].getId()));
 
-		// Calculate the amount of energy recovered every second
-		double recoverRate = (48 + client.getBoostedSkillLevel(Skill.AGILITY)) / 360.0;
+		gracefulModifier += highRecoverItems.stream().filter(p -> p).count() * 0.04;
 
-		if (isLocalPlayerWearingFullGraceful())
+		// If the bonus is 20%, player has the full graceful set effect so we add 10% more
+		// Have to account for a floating point error so we floor it down to match up
+		if (Math.floor(gracefulModifier * 100) / 100 == 1.2)
```
dekvall: Keep `isLocalPlayerWearingFullGraceful()` and check it first so you don't need to do this.

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,49 +201,62 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
-		return (ALL_GRACEFUL_HOODS.contains(items[EquipmentInventorySlot.HEAD.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_TOPS.contains(items[EquipmentInventorySlot.BODY.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_LEGS.contains(items[EquipmentInventorySlot.LEGS.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_GLOVES.contains(items[EquipmentInventorySlot.GLOVES.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_BOOTS.contains(items[EquipmentInventorySlot.BOOTS.getSlotIdx()].getId()) &&
+		// Items that have 3% bonus recovery rate
+		final List<Boolean> lowRecoverItems = Lists.newArrayList(
+			ALL_GRACEFUL_HOODS.contains(items[EquipmentInventorySlot.HEAD.getSlotIdx()].getId()),
+			ALL_GRACEFUL_BOOTS.contains(items[EquipmentInventorySlot.BOOTS.getSlotIdx()].getId()),
+			ALL_GRACEFUL_GLOVES.contains(items[EquipmentInventorySlot.GLOVES.getSlotIdx()].getId()),
```
dekvall: Imo having separate `ifs` for these and the `highRecoverItems` would be more readable.

```
if (ALL_GRACEFUL_HOODS.contains(items[EquipmentInventorySlot.HEAD.getSlotIdx()].getId()))
{
    gracefulModifier += 0.03
}
```
etc.

abextm: also this will throw `ArrayOutOfBoundsException`s, so you need to do bounds checks

Issue #4377, opened 2018-07-18
Track and Record GE buy/sell history


douglasg14b:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
GE buy/sell history tracking. This would be an amazing feature to have for GE flippers, even if it was just a simple list of buy/sell orders and histories.

An expanded scope would be providing insights on that information such as margins, profits, losses, common items bought/sold...etc

**Describe alternatives you've considered**
OSBuddy keeps track of this information and displays a list of  historical purchases and sales on their site under your account.

Asymons:
It was meant to solve a somewhat separate issue, but was marked as duplicate. May work on extending to full list in future.

PR #10589, opened 2020-01-11
Remember bought items price/quantity - Runelite

Eziuxxx:
Dear All,

Not sure if this is the correct place to make a suggestions if I am wrong - I am sorry.

Would it be possible to make somehow Runelite remember price for which you bought item from GE? For example On January 1 I bought t bow for example for 1b and today is January 10 and price is 950. Lets say I dont remember when and for how much I bought the item so this would help a lot. I think this would really help to all the merchants and players who dont want to sell items for less than they bough it.

Thank you.


chestnut1693:
You created a pull request but these are for people requesting to add their code to runelite. Creating an issue on the other hand is the correct place to ask for new features. You can close this pull request and ask again with an issue.

Hydrox6:
1. This is not how to make an issue
2. This feature already exists on the website. Log in with a Google account with the button in the top right, and then future GE trades will be logged on there.
3. Duplicate of #4377

PR #10599, opened 2020-01-14
cache: Add hitsplat definition

hex-agon:
This PR simply adds hitsplat definition decoding capabilities to the cache project.

### DIFF
--- a/cache/src/main/java/net/runelite/cache/definitions/HitSplatDefinition.java
+++ b/cache/src/main/java/net/runelite/cache/definitions/HitSplatDefinition.java
@@ -29,7 +29,7 @@ import lombok.Data;
 @Data
 public class HitSplatDefinition
 {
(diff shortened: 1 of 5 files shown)


cache/src/main/java/net/runelite/cache/definitions/HitSplatDefinition.java
```diff
@@ -0,0 +1,48 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.cache.definitions;
+
+import lombok.Data;
+
+@Data
+public class HitSplatDefinition
+{
+	String stringFormat = "";
```
Adam-:
```suggestion
	private String stringFormat = "";
```
?

PR #10600, opened 2020-01-14
Add reanimated monsters to npc_health.json

### LINKED ISSUE #10594: Add reanimated monsters to npc_health.json
Add the monsters you can reanimate through the Arceuus spellbook to the [npc_health.json](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/resources/npc_health.json) file. 
This would allow the opponent info plugin to display their exact hp instead of a % value, and the XP Tracker to accurately calculate kills to next level while in combat with them.

wradstok:
Closes  #10594


Adam-:
This just doesn't work. The lookup method is

```java
	@Nullable
	public Integer getHealth(final String name, final int combatLevel)
	{
		return healthMap.get(name + "_" + combatLevel);
	}
```

so they must have a combat level. Maybe it is `-1`.

Joshua-F:
> 
> 
> This just doesn't work. The lookup method is
> 
> ```java
> 	@Nullable
> 	public Integer getHealth(final String name, final int combatLevel)
> 	{
> 		return healthMap.get(name + "_" + combatLevel);
> 	}
> ```
> 
> so they must have a combat level. Maybe it is `-1`.

They're set to `0`.

wradstok:
Sorry, somehow managed to miss that. Thanks @Joshua-F, I've added _0 to all reanimated monsters in the file.

PR #10603, opened 2020-01-15
added tracking for CoX and ToB pb

matcpn:
Resolves the remaining two steps of #8454

Let me know what we think of something like this - tracking pb with the ConfigManager instead of relying on jagex's chat printouts

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -335,7 +335,7 @@ public class ChatCommandsPlugin extends Plugin
 		matcher = RAIDS_DURATION_PATTERN.matcher(message);
 		if (matcher.find())
 		{
-		    matchPb(matcher);
+			matchPb(matcher);
 		}
 
 		lastBossKill = null;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -88,6 +90,7 @@
 {
 	private static final Pattern KILLCOUNT_PATTERN = Pattern.compile("Your (.+) (?:kill|harvest|lap|completion) count is: <col=ff0000>(\\d+)</col>");
 	private static final Pattern RAIDS_PATTERN = Pattern.compile("Your completed (.+) count is: <col=ff0000>(\\d+)</col>");
+	private static final Pattern COX_DURATION_PATTERN = Pattern.compile("Congratulations - your raid is complete! Duration <col=ff0000>([0-9:]+)</col>");
```
matcpn: This chat line shows up as pink - do I need to wrap this in <col> tags with the right hex value?

Adam-:
This looks like it is consistent with the way the other pbs work, so its seems fine.

Adam-:
added in 10e5a1f1814aa0c0a0c47e635e816318db2a86b4

PR #10625, opened 2020-01-20
Add Level/XP Difference to Skill Calc

zacharydwaller:
Adds level and XP difference calculations to Skill Calculator. Updating the Level/XP Difference will calculate and set the Target Level/XP based on `Target = Current + Difference`. Also, setting the Current or Target fields will update the Difference fields based on `Difference = Target - Current`.

Examples:

xpdiff

Calculating how much XP is required to level from Current to Target. In this case how much XP is required for 92 to 99.

xpdiff2

Calculating how many levels a quest reward would give. In this case if I were at 20 Strength, completing Waterfall Quest (13750 XP) would give me 12 levels.


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UICalculatorInputArea.java
```diff
@@ -44,14 +44,19 @@
 	private final JTextField uiFieldCurrentXP;
 	private final JTextField uiFieldTargetLevel;
 	private final JTextField uiFieldTargetXP;
+	private final JTextField uiFieldDeltaLevel;
+	private final JTextField uiFieldDeltaXP;
+
```
Nightfirecat: Remove this extra newline

Adam-:
The UI for this is already sort of crunched. Also, in what cases would you want to manually input level diff or xp diff?

PR #10631, opened 2020-01-22
Shift-Click to Withdraw-All/Deposit-All

### LINKED ISSUE #10612: Shift click option for 'withdraw all' in the bank with Menu Entry Swapper?
This would help with banking a lot as it would remove the need to right click or keep swapping the default mode when withdrawing a variety of items for farm runs etc.

An option added to the menu entry swapper plugin that allows holding shift for the 'Withdraw all' option in the bank

image

zacharydwaller:
Closes #10612

Ctrl-clicking will now swap the left click to Withdraw-All or Deposit-All. Ctrl-click was chosen instead of Shift-click because Shift is used for the Swap Bank Op option (shift-click for eating/wearing/etc).

This is super useful for when you want to alternate between Withdraw-1/5/etc and Withdraw-All quickly without having to select the All button or right click. An example would be if you had a stack of Ava's devices (which I do). I could just left click Ava's Accumulator to withdraw 1 like usual, and then ctrl-click my arrows to withdraw them all. Really simple and helpful QOL change.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -307,30 +307,49 @@ public class MenuEntrySwapperPlugin extends Plugin
 		// widget ticking and prior to our client tick event. This is because drag start
 		// is what builds the context menu row which is what the eventual click will use
 
+		// Swap to shift-click deposit behavior
 		// Deposit- op 2 is the current withdraw amount 1/5/10/x
-		if (shiftModifier && menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 2
-			&& config.swapBankOp() && menuEntryAdded.getOption().startsWith("Deposit-"))
+		if (shiftModifier && config.bankDepositShiftClick() != ShiftDepositMode.OFF
+			&& menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 2
+			&& menuEntryAdded.getOption().startsWith("Deposit-"))
 		{
-			MenuEntry[] menuEntries = client.getMenuEntries();
+			ShiftDepositMode shiftDepositMode = config.bankDepositShiftClick();
+			final int actionId = shiftDepositMode.getMenuAction().getId();
+			final int opId = shiftDepositMode.getIdentifier();
+			bankModeSwap(actionId, opId);
+		}
 
-			// Find the extra menu option; they don't have fixed names, so check
-			// based on the menu identifier
-			for (int i = menuEntries.length - 1; i >= 0; --i)
+		// Swap to shift-click withdraw behavior
+		// Deposit- op 1 is the current withdraw amount 1/5/10/x
+		if (shiftModifier && config.bankWithdrawShiftClick() != ShiftWithdrawMode.OFF
+			&& menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 1
+			&& menuEntryAdded.getOption().startsWith("Withdraw-"))
+		{
+			ShiftWithdrawMode shiftWithdrawMode = config.bankWithdrawShiftClick();
+			final int actionId = shiftWithdrawMode.getMenuAction().getId();
+			final int opId = shiftWithdrawMode.getIdentifier();
+			bankModeSwap(actionId, opId);
+		}
+	}
+
+	private void bankModeSwap(int entryTypeId, int entryIdentifier)
+	{
+		MenuEntry[] menuEntries = client.getMenuEntries();
+
+		for (int i = menuEntries.length - 1; i >= 0; --i)
+		{
+			MenuEntry entry = menuEntries[i];
+
+			if (entry.getType() == entryTypeId && entry.getIdentifier() == entryIdentifier)
 			{
-				MenuEntry entry = menuEntries[i];
+				// Raise the priority of the op so it doesn't get sorted later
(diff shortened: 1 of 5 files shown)


zacharydwaller:
Ctrl-Click wasn't liked much so I changed it to Shift-Click. Since Shift-Click was already being used by the Swap Bank Op option I created some dropdowns to customize shift-click behavior.

Shift-click for deposit is set to Extra Op by default to accommodate users who were previously using Swap Bank Op.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -307,34 +307,52 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 		// widget ticking and prior to our client tick event. This is because drag start
 		// is what builds the context menu row which is what the eventual click will use
 
-		// Deposit- op 2 is the current withdraw amount 1/5/10/x
-		if (shiftModifier && menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 2
-			&& config.swapBankOp() && menuEntryAdded.getOption().startsWith("Deposit-"))
+		// Swap to shift-click deposit behavior
+		// Deposit-/CC_OP/id 2 is the current withdraw amount 1/5/10/x
```
Adam-: I think you `'s/`'d these comments somehow?

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftDepositMode.java
```diff
@@ -0,0 +1,63 @@
...
+	private final MenuAction menuAction;
+	private final int identifier;
+
+	@Override
+	public String toString()
+	{
+		return name;
+	}
+
+	public MenuAction getMenuAction()
```
Adam-: With `@Getter` is it not necessary to make your own

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftDepositMode.java
```diff
@@ -0,0 +1,63 @@
...
+
+@Getter
+@RequiredArgsConstructor
+public enum ShiftDepositMode
+{
+	DEPOSIT_1("Deposit-1", MenuAction.CC_OP, 3),
+	DEPOSIT_5("Deposit-5", MenuAction.CC_OP, 4),
+	DEPOSIT_10("Deposit-10", MenuAction.CC_OP, 5),
+	DEPOSIT_X_CUSTOM("Deposit Custom X", MenuAction.CC_OP_LOW_PRIORITY, 6),
+	DEPOSIT_X("Deposit-X", MenuAction.CC_OP_LOW_PRIORITY, 7),
```
Adam-: This is the setter for x right? This probably isn't useful and would confuse people more than anything.

zacharydwaller:
Implemented the suggestions, removed Withdraw/Deposit-X, replaced methods with @Getter, etc

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -307,34 +307,52 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 		// widget ticking and prior to our client tick event. This is because drag start
 		// is what builds the context menu row which is what the eventual click will use
 
-		// Deposit- op 2 is the current withdraw amount 1/5/10/x
-		if (shiftModifier && menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 2
-			&& config.swapBankOp() && menuEntryAdded.getOption().startsWith("Deposit-"))
+		// Swap to shift-click deposit behavior
+		// Deposit-, CC_OP, and id 2 is the current withdraw amount 1/5/10/x
```
Adam-: The original comment here is correct. Deposit op 2 is the current withdraw amount. We just have the op called "identifier". The "op" also implies CC_OP.

zacharydwaller:
Changed back to original comment

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftDepositMode.java
```diff
@@ -0,0 +1,56 @@
...
+	DEPOSIT_5("Deposit-5", MenuAction.CC_OP, 4),
+	DEPOSIT_10("Deposit-10", MenuAction.CC_OP, 5),
+	DEPOSIT_X_CUSTOM("Deposit Custom X", MenuAction.CC_OP_LOW_PRIORITY, 6),
+	DEPOSIT_ALL("Deposit-All", MenuAction.CC_OP_LOW_PRIORITY, 8),
+	EXTRA_OP("Eat/Wield/Etc.", MenuAction.CC_OP_LOW_PRIORITY, 9),
+	OFF("Off", MenuAction.UNKNOWN, 0);
+
+	private final String name;
+
+	@Getter
```
raiyni: You only need `@Getter` on the enum

zacharydwaller:
Removed unnecessary @Getter annotations

Adam-:
Is there a reason why these do not default to OFF?

zacharydwaller:
No reason for Withdraw behavior to not default to Off, I can change it to default to off if needed. I default Deposit behavior to Extra-Op because my PR overwrites the old Swap Bank Op option - so if a user had that on previously they might be upset if it suddenly stopped working

Adam-:
The old swap option also defaulted to off, though

zacharydwaller:
Fair point, I'll set it so that it defaults OFF

PR #10639, opened 2020-01-23
Fixes cox personal best time tracking

melkypie:
Makes #10603 actually work. for CoX


Adam-:
Is the `Your completed Chambers of Xeric count is` also a friendmsg?

melkypie:
No it isn't, if it was the !kc command wouldn't work also

Issue #5033, opened 2018-08-21
Add support for thieving to loot tracker


Matyim:
I was thinking that with the new addition of the loot tracker for kills on mobs that it would also be good to have one tracking thieving loot to show the total price. I was thinking of instead of the way where it tracks items on the ground after a kill that it would just check each time you pickpocket for a new item going into your inventory. Or by looking in chat and then adding that to the list, so for a master farmer in chat it would say 4 potato seeds, then it could list potato seeds in the loot counter and say the total price.

deathbeam:
This makes no sense, would you display thousands of entries for each pickpocket? We have #4126 what is more skilling oriented, adding this to loot tracker would be dumb

PR #10640, opened 2020-01-24
Add pickpockets to loot tracker

danielcimento:
Suggested a few times in the past (e.g. https://github.com/runelite/runelite/issues/5033).

I thought this would be a good addition to the loot tracker. It was something I found myself wanting when thieving master farmers. It seems to have been a bit polarizing in the past, so I've put it behind a config flag by default.

Possibly extensible to cover loot from all types of thieving (stalls, chests, etc). Let me know if there are any improvements needed

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
@@ -53,6 +53,16 @@ enum LootEvent
 	KINGDOM_OF_MISCELLANIA("Kingdom of Miscellania"),
 	LARRANS_CHEST_BIG("Larran's big chest"),
 	LARRANS_CHEST_SMALL("Larran's small chest"),
+	// Pickpocket-able NPCs who give loot other than coin pouches
+	PICKPOCKET_BANDIT("Bandit (Pickpocket)"),
+	PICKPOCKET_CAVE_GOBLIN("Cave Goblin (Pickpocket)"),
+	PICKPOCKET_ELF("Elf (Pickpocket)"),
+	PICKPOCKET_GNOME("Gnome (Pickpocket)"),
+	PICKPOCKET_HAM_MEMBER("H.A.M. Member (Pickpocket)"),
+	PICKPOCKET_HERO("Hero (Pickpocket)"),
+	PICKPOCKET_MASTER_FARMER("Master Farmer (Pickpocket)"),
+	PICKPOCKET_ROGUE("Rogue (Pickpocket)"),
+	PICKPOCKET_TZHAAR_HUR("TzHaar-Hur (Pickpocket)"),
 	THEATRE_OF_BLOOD("Theatre of Blood");
 
 	@Getter
@@ -66,9 +76,23 @@ enum LootEvent
 		13151, LootEvent.ELVEN_CRYSTAL_CHEST
 	);
 
+	static final Map<String, LootEvent> PICKPOCKET_EVENT_LOOKUP_TABLE = ImmutableMap.<String, LootEvent>builder()
+		.put("Master Farmer", PICKPOCKET_MASTER_FARMER)
+		.put("Martin", PICKPOCKET_MASTER_FARMER)
+		.put("TzHaar-Hur", PICKPOCKET_TZHAAR_HUR)
+		.put("Elf", PICKPOCKET_ELF)
+		.put("Hero", PICKPOCKET_HERO)
+		.put("Gnome", PICKPOCKET_GNOME)
+		.put("Bandit", PICKPOCKET_BANDIT)
+		.put("Cave Goblin", PICKPOCKET_CAVE_GOBLIN)
+		.put("Rogue", PICKPOCKET_ROGUE)
+		.put("H.A.M. Member", PICKPOCKET_HAM_MEMBER)
+		.build();
+
 	static final List<LootEvent> DIRECT_TO_INVENTORY_EVENTS = ImmutableList.<LootEvent>builder()
 		.add(HERBIBOAR, HESPORI, GAUNTLET)
 		.addAll(CHEST_EVENT_LOOKUP_TABLE.values())
+		.addAll(PICKPOCKET_EVENT_LOOKUP_TABLE.values())
 		.build();
 
 	static final Map<String, LootEvent> CLUE_SCROLL_LOOKUP_TABLE = ImmutableMap.<String, LootEvent>builder()
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2018, Psikoi <https://github.com/psikoi>
+ * Copyright (c) 2018, Adam <Adam@sigterm.info>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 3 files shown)


deathbeam:
I dont think just putting it behind config flag solves anything, the original point still stands, loot tracker is tracking monster/boss loot and thats it.

dekvall:
To be fair, nowadays the loot tracker also tracks Kingdom rewards, Fishing Trawler and a variety of chests. It's not exclusively tracking monster and boss loot anymore. I don't think that we need to track pickpockets of npcs which only yield coins but it would be interesting to see the loot you get from master farmers, ham members and elves etc. I think that we have passed the old line already and I am not sure where we should draw the new one. Maybe this config is a nice compromise while we decide.

abextm:
I think if we want a config (I'm not sure we do) it should be very limited — just [Everything/PvM]. I don't want the config for it filled with options for every single non PvM activity it can do.

danielcimento:
@abextm

I'd be willing to help merge everything into a non-PvM option, if you need me to.

@deathbeam

I see where you're coming from, but I'm afraid I don't quite understand why only tracking PvM loot makes the loot tracker better as a feature.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -486,6 +489,15 @@ public void onChatMessage(ChatMessage event)
 			return;
 		}
 
+		final Matcher pickpocketMatcher = pickpocketingRegex.matcher(message);
+		if (config.trackPickpockets() && pickpocketMatcher.matches())
+		{
+			String pickpocketTarget = WordUtils.capitalize(pickpocketMatcher.group("target"));
+			eventType = pickpocketTarget + " (Pickpocket)";
```
XrioBtw: I would recommend adding e.g. `private static final String PICKPOCKET_EVENT = " (Pickpocket)"` and using this here and in line 542.

danielcimento: Mmm, that's a good idea. I had a const at first but I took it out because I hadn't thought to use it in both places

danielcimento:
I've refactored everything into non-combat vs combat loot entries and changed the toggle so it hides all non-combat loot. I tried to make the change as non-intrusive as possible, though I did move a lot of the event strings/constants ("Hespori", "Herbiboar", etc) into an enum class, just to reduce clutter a bit and maybe help with future extensibility.

Let me know if these changes are amenable, or if there's something else that needs to be changed.

danielcimento:
Hmm, I'll take a look at those build failures

Adam-:
I think this is fine to just track pickpocketing. Maybe we could extend the "ignored items" feature to also ignore events if we wanted. If you want to refactor the EventType stuff it should preferably be in a separate commit prior to adding any of the new pickpocketing stuff.

danielcimento:
> 
> 
> I think this is fine to just track pickpocketing. Maybe we could extend the "ignored items" feature to also ignore events if we wanted. If you want to refactor the EventType stuff it should preferably be in a separate commit prior to adding any of the new pickpocketing stuff.

That makes sense. I'll move the event type stuff to a separate PR

Adam-:
It can be in the same PR, just separate commits. Basically just rebase and reorder them since you sort of have done it already but in reverse.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -474,62 +467,68 @@ public void onChatMessage(ChatMessage event)
...
+		}
+	}
+
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked event)
+	{
+		// H.A.M. members show up differently in the chat box, so we need to specially track them
+		if (event.getMenuOption().equals("Pickpocket"))
+		{
+			pickpocketingHamMembers = Text.removeTags(event.getMenuTarget()).equals("H.A.M. Member");
```
Adam-: You could probably just store the full name here and always use it to avoid this issue in the future with other things.

danielcimento: Store the name of the last pickpocket target? That was my first implementation, but I think it may have some issues on its own:

* there's a very small (though not non-existent) risk of race conditions, since players can select different menu options while mid-pickpocket, so the loot might potentially be recorded under the wrong name.
* some pickpocket-able NPCs have randomized names, though their pickpocketing message is the same. e.g. I believe all the Prifddinas elves have different names, but still display as "you pick the elf's pocket", so there'd be a lot of book-keeping required to map all those names.

I do agree that a specific boolean just for ham members is a bit hacky, so I'll consider more extensible alternatives before pushing the newest changes

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
```diff
@@ -0,0 +1,64 @@
+package net.runelite.client.plugins.loottracker;
```
Adam-: The license header should be added to all new files.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public enum LootEvent
```
Adam-:
```suggestion
enum LootEvent
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
```diff
@@ -0,0 +1,64 @@
...
+	FISHING_TRAWLER("Fishing Trawler"),
+	GAUNTLET("The Gauntlet"),
+	HESPORI("Hespori"),
+	HERBIBOAR("Herbiboar"),
+	KINGDOM_OF_MISCELLANIA("Kingdom of Miscellania"),
+	LARRANS_CHEST_BIG("Larran's big chest"),
+	LARRANS_CHEST_SMALL("Larran's small chest"),
+	THEATRE_OF_BLOOD("Theatre of Blood");
+
+	final String asText;
```
Adam-:
```suggestion
	@Getter
	private final String name;
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootEvent.java
```diff
@@ -0,0 +1,64 @@
...
+	static final Map<String, LootEvent> CLUE_SCROLL_LOOKUP_TABLE = ImmutableMap.<String, LootEvent>builder()
+		.put("beginner", CLUE_SCROLL_BEGINNER)
+		.put("easy", CLUE_SCROLL_EASY)
+		.put("medium", CLUE_SCROLL_MEDIUM)
+		.put("hard", CLUE_SCROLL_HARD)
+		.put("elite", CLUE_SCROLL_ELITE)
+		.put("master", CLUE_SCROLL_MASTER)
+		.build();
+
+	LootEvent(String asText)
```
Adam-: Use `@RequiredArgsConstructor`

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -125,6 +130,17 @@
 	// Last man standing map regions
 	private static final Set<Integer> LAST_MAN_STANDING_REGIONS = ImmutableSet.of(13658, 13659, 13914, 13915, 13916);
 
+	private static final Pattern pickpocketingRegex = Pattern.compile("You pick (the )?(?<target>.+)'s? pocket.*");
```
Adam-:
```suggestion
	private static final Pattern PICKPOCKET_REGEX = Pattern.compile("You pick (the )?(?<target>.+)'s? pocket.*");
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -125,6 +130,17 @@
...
+	/*
+	 * This map is used when a pickpocket target has a different name in the chat message than their in-game name,
+	 * *and* the name in the chat box conflicts with other targets (otherwise just use that name in the look-up map)
+	 * Note that if the two NPCs can be found in the same place, there is a chance of race conditions
+	 * occurring when changing targets mid-pickpocket, in which case a different solution may need to be considered.
+	 */
+	private static final Map<String, List<String>> PICKPOCKET_DISAMBIGUATION_MAP = ImmutableMap.of(
```
Adam-: This can probably just be a multimap

dabolink:
> I think this is fine to just track pickpocketing. Maybe we could extend the "ignored items" feature to also ignore events if we wanted. If you want to refactor the EventType stuff it should preferably be in a separate commit prior to adding any of the new pickpocketing stuff.

I did create a PR a while ago #9752 which allows the user to hide specific events. Haven't checked on it but if you want I could rebase it.

Adam-:
I applied the changes I left a few days ago, and started a test. I noticed that Paladins aren't on this. Is that because the loot from them isn't random? I don't think it makes much sense to only track random loot.

danielcimento:
Yeah I originally left off Paladins because the loot wasn't random. It might be better to leave off enemies whose loot is just coin pouches instead (i.e. add paladins). It's mostly a question of whether people want it tracked. It's not too much of a hassle to add in after the fact, anyhow

Adam-:
can you add anything which is pickpocket-able?

danielcimento:
Sure

Adam-:
Most of this was added in #10759. I kept the `LootEvent` out as it makes the diff a lot smaller. If you want to revisit that we could, though I am not sure I want to use it for pickpocketing since it requires having to hardcode in all of the possible pickpocket targets which is not so great. The other usages seemed good, though.

Nightfirecat:
I am closing this PR for now, as the main objective (tracking pickpocket loots) has been implemented and merged separately. This PR, or a new PR, can be opened for the refactoring portion of this feature.

PR #10650, opened 2020-01-25
Add configurability to the Idle Notifier plugin.

### LINKED ISSUE #10761: Flash Notifications
Within the "RuneLite"setting there is "flash notifications" can we have a sub catagory for that. My idea with it is to let us choose how bright and what color we want the screen to flash.

Paveldin:
This commit offers the user to customize the color of their flashing screen notification depending the context in which it is triggered.
image

Fixes #10761

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -96,7 +96,8 @@ public class Notifier
 		.build();
 
 	// Notifier properties
-	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
+
+	public static Color FLASH_COLOR = new Color(255, 0, 0, 70);
 	private static final int MINIMUM_FLASH_DURATION_MILLIS = 2000;
 	private static final int MINIMUM_FLASH_DURATION_TICKS = MINIMUM_FLASH_DURATION_MILLIS / Constants.CLIENT_TICK_LENGTH;
 
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -96,8 +96,7 @@ public class Notifier
 		.build();
 
 	// Notifier properties
-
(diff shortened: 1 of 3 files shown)


Adam-:
This is incorrect because other plugins using `notifier.notify()` would inherit the flash color from the last thing which set it.

Paveldin:
I have updated the code so FLASH_COLOR is updated everywhere. I have remove the ability to change the color depending the context (One color for Hitpoints, one for prayer, etc).

This will keep the code clean and be sure we get the same color everywhere.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -96,7 +96,7 @@ public String toString()
 		.build();
 
 	// Notifier properties
-	private static final Color FLASH_COLOR = new Color(255, 0, 0, 70);
+	public static  Color FLASH_COLOR = new Color(255, 0, 0, 70);
```
Nightfirecat: This should not be static if it is changed to be an instance field.
```suggestion
	public Color FLASH_COLOR = new Color(255, 0, 0, 70);
```

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -24,10 +24,15 @@
...
 package net.runelite.client.plugins.idlenotifier;
 
+import java.awt.Color;
+import net.runelite.client.Notifier;
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
+import java.awt.*;
```
Nightfirecat: Don't use star imports for non-static imports. Please [review our Code Contribution guidelines](https://github.com/runelite/runelite/wiki/Code-Conventions).

Nightfirecat:
This is basically covered by #12133

PR #10662, opened 2020-01-26
Add Unit support to Config

Hydrox6:
Adds an annotation (like `@Range`) to specify the units to be displayed in the box. This should lessen confusion on certain config items that have their units hidden by length or by being in the tooltip.
image
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/Units.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/Units.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, Adam <Adam@sigterm.info>
+ * Copyright (c) 2020, Hydrox6 <ikada@protonmail.ch>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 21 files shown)


runelite-client/src/main/java/net/runelite/client/config/Units.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Used with ConfigItem, defines what units are shown to the side of the box.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@Documented
+public @interface Units
+{
+	String name() default "";
```
Adam-: 1) why is there a default
2) if you make this `String value()` then you can just use `@Unit("ms")`

Hydrox6: Ah, didn't know it worked like that.

Hydrox6:
Updated to use a `UnitType` enum

abextm:
I don't like the enum — It prevents external plugins from using this. It would probably be better to have some static final strings on the interface if we want to centralize parts of this.

Hydrox6:
Now updated to use static strings (apparently the `public static final` was unneeded, according to IDEA)

PR #10638, opened 2020-01-23
Sidebar toggle hotkey changes

loldudester:
(New pr because I am a dingus with git)

Adds a config option for the existing Ctrl+F11 hotkey which hides the sidebar entirely.

Also adds a new configurable hotkey that toggles the current/last open plugin panel (defaults to Ctrl+F12).

Both configs are at the bottom of the Runelite config panel.

java_tyLDQy8Uuk

PR #10664, opened 2020-01-26
Sidebar toggle hotkey changes

loldudester:
Re-do of #10638 because I'm an idiot. Last time I *promise*.

Adds a config option for the existing Ctrl+F11 hotkey which hides the sidebar entirely.

Adds a new configurable hotkey that toggles the current/last open plugin panel (defaults to Ctrl+F12).

Both configs are at the bottom of the Runelite config panel.

java_tyLDQy8Uuk

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -39,8 +39,6 @@ import java.awt.LayoutManager;
 import java.awt.Rectangle;
 import java.awt.Toolkit;
 import java.awt.TrayIcon;
-import java.awt.event.InputEvent;
-import java.awt.event.KeyEvent;
 import java.awt.event.MouseEvent;
 import java.awt.image.BufferedImage;
 import javax.annotation.Nullable;
@@ -73,7 +71,6 @@ import net.runelite.client.RuneLiteProperties;
 import net.runelite.client.callback.ClientThread;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.config.ExpandResizeType;
-import net.runelite.client.config.Keybind;
 import net.runelite.client.config.RuneLiteConfig;
 import net.runelite.client.config.WarningOnExit;
 import net.runelite.client.eventbus.Subscribe;
@@ -345,8 +342,7 @@ public class ClientUI
 			frame.add(container);
 
 			// Add key listener
-			final HotkeyListener sidebarListener = new HotkeyListener(() ->
-				new Keybind(KeyEvent.VK_F11, InputEvent.CTRL_DOWN_MASK))
+			final HotkeyListener sidebarListener = new HotkeyListener(config::sidebarToggleKey)
 			{
 				@Override
 				public void hotkeyPressed()
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -353,6 +353,17 @@ public class ClientUI
 
 			keyManager.registerKeyListener(sidebarListener);
 
+			final HotkeyListener pluginPanelListener = new HotkeyListener(config::panelToggleKey)
+			{
+				@Override
+				public void hotkeyPressed()
+				{
+					togglePluginPanel();
+				}
+			};
+
+			keyManager.registerKeyListener(pluginPanelListener);
+
 			// Add mouse listener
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -588,6 +600,34 @@ public void requestFocus()
 		{
 			OSXUtil.requestFocus();
 		}
+		// The workaround for Windows is to minimise and then un-minimise the client to bring
+		// it to the front because java.awt.Window#toFront doesn't work reliably.
+		// See https://stackoverflow.com/questions/309023/how-to-bring-a-window-to-the-front/7435722#7435722
+		else if (OSType.getOSType() == OSType.Windows && !frame.isFocused())
```
Nightfirecat: This shouldn't be in this PR

loldudester: 👀 how'd that get in there

loldudester: Fixed
probably
maybe
git is hard

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -346,7 +343,7 @@ public void init(final RuneLite runelite) throws Exception
 
 			// Add key listener
 			final HotkeyListener sidebarListener = new HotkeyListener(() ->
```
Nightfirecat: This can be shortened to `final HotkeyListener sidebarListener = new HotkeyListener(config::sidebarToggleKey);`

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -354,6 +354,18 @@ public void hotkeyPressed()
 
 			keyManager.registerKeyListener(sidebarListener);
 
+			final HotkeyListener pluginPanelListener = new HotkeyListener(() ->
+				config.panelToggleKey())
```
Nightfirecat: This can be shortened to `final HotkeyListener pluginPanelListener = new HotkeyListener(config::panelToggleKey);`

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -739,6 +751,30 @@ private void toggleSidebar()
...
+		if (isPanelOpen)
+		{
+			contract();
+		}
+		else
+		{
+			// Try to restore last panel
+			expand(currentNavButton);
+
+			//Checks if the toolbar was previously closed by toggleSidebar
```
Nightfirecat:
```suggestion
			// Add the plugin toolbar if it was previously closed by toggleSidebar
```

Nightfirecat:
@Adam- I've tested this in the client already fwiw

Adam-:
if I use ctrl+f12 to close the panel, then ctrl+f11 to expand it it erroneously adds the width of the (already existing) toolbar to the client frame.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -742,6 +749,30 @@ private void toggleSidebar()
...
+			// Try to restore last panel
+			expand(currentNavButton);
+
+			// Add the plugin toolbar if it was previously closed by toggleSidebar
+			if (!container.isAncestorOf(pluginToolbar))
+			{
+				container.add(pluginToolbar);
+				frame.expandBy(pluginToolbar.getWidth());
+			}
+		}
```
Nightfirecat: Per Adam's comment, this causes the frame to expand by the plugin toolbar's width if you collapse the panel (F+12) then expand the sidebar (F+11). From what I can tell, that's partially because of the `sidebarOpen = !sidebarOpen;` state mutation which happens here.

Anyway, I think this whole method body can safely be replaced by the following:

```java
final boolean pluginPanelOpen = pluginPanel != null;

currentButton.setSelected(!pluginPanelOpen);

if (pluginPanelOpen)
{
	contract();
}
else
{
	expand(currentNavButton);
}
```

Nightfirecat: I neglected to suggest a null check for `currentButton`, since that can be nulled when clicking a plugin panel open and closed again. I've amended your commit to add this.

Nightfirecat:
Needs rebase after the side mouse buttons config change

PR #10665, opened 2020-01-26
Inventory Grid Changes

Graviton1647:
Changes to the Grid adding Color options and a new option to show the grid all the time

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
@@ -24,17 +24,31 @@
  */
 package net.runelite.client.plugins.inventorygrid;
 
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
+import java.awt.*;
+
 @ConfigGroup("inventorygrid")
 public interface InventoryGridConfig extends Config
 {
+
+
+	@ConfigItem(
+		keyName = "showindividualGrid",
+		name = "Toggle individual grid",
+		description = "Shows boxes where items go",
+		position = 5
+	)
+	default boolean showIndividualGrid() { return false; }
+
 	@ConfigItem(
 		keyName = "showItem",
 		name = "Show item",
-		description = "Show a preview of the item in the new slot"
+		description = "Show a preview of the item in the new slot",
+		position = 4
 	)
 	default boolean showItem()
 	{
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
```diff
@@ -24,17 +25,34 @@
...
 @ConfigGroup("inventorygrid")
 public interface InventoryGridConfig extends Config
 {
+
+
+	@ConfigItem(
+		keyName = "showindividualGrid",
+		name = "Toggle individual grid",
+		description = "Shows boxes where items go",
+		position = 5
```
Adam-: Whats up with all of the config positions being backwards?

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
```diff
@@ -74,6 +72,22 @@ public Dimension render(Graphics2D graphics)
...
+		{
+
+			WidgetItem targetWidgetItem = inventoryWidget.getWidgetItem(i);
+
+			final Rectangle bounds = targetWidgetItem.getCanvasBounds();
+
+			if (config.showIndividualGrid())
+			{
+				graphics.setColor(config.colorGrid());
+				graphics.fill(bounds);
```
Adam-: This is incorrect because if both showIndividualGrid (which maybe should be renamed to something with a better name?) and showGrid is on, it will double-fill the slot with the configured color, which if using alpha will make the end color incorrect.

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
Graviton1647: They was rendering backwards for some reason when I did it the other way

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
Graviton1647: 'it will double-fill the slot with the configured color, which if using alpha will make the end color incorrect.' ye I kept that as I thought it looked nice

runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridConfig.java
Adam-: Items with a lower priority are higher on the list, and we traditionally keep the config items ordered in source the same way their priorities are.

Issue #10370, opened 2019-12-02
Hot Cold Solver: persisted arrow hint


Alexsuperfly:
When teleporting to a location where a hint arrow can be displayed then activating the device to solve another step, the hint arrow location is not cleared when its location is no longer in the list of possible solutions.

![](https://i.imgur.com/dGFMYme.png)

An image from a friend of mine who discovered this, they teleported to the duel arena, saw the displayed hint arrow, activated the device which determined that displayed location to be an incorrect one but the arrow still existed.

PR #10670, opened 2020-01-27
cluescroll plugin: Clear stale location clue hint arrows 

Nightfirecat:
Location clue scrolls added a hint arrow when the player was nearby one
of the clue's locations. It did not, however clear that arrow if the
locations were changed and the previously-set hint arrow location was no
longer marked. This commit updates this behavior to clear the active
hint arrow should that be the case. (and if no NPC hint arrows are set)

Fixes runelite/runelite#10370

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -396,14 +396,17 @@ public class ClueScrollPlugin extends Plugin
 		if (clue instanceof LocationClueScroll)
 		{
 			final WorldPoint[] locations = ((LocationClueScroll) clue).getLocations();
+			final boolean npcHintArrowMarked = client.getHintArrowNpc() != null && npcsToMark.contains(client.getHintArrowNpc());
+
+			if (!npcHintArrowMarked)
+			{
+				client.clearHintArrow();
+			}
 
 			for (WorldPoint location : locations)
 			{
 				// Only set the location hint arrow if we do not already have more accurate location
-				if (location.isInScene(client)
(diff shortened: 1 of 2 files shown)


Nightfirecat:
I've added a test to ensure the hint arrow is correctly being cleared after the clue locations update.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -398,6 +398,11 @@ public void onGameTick(final GameTick event)
 					}
 				}
 			}
+
+			if (!npcHintArrowMarked && !addedHintArrow)
+			{
+				client.clearHintArrow();
```
Adam-: can't you just move this up above L380 and avoid the whole "addedHintArrow" logic?

Nightfirecat: I guess there's no harm in clearing the hint arrow each tick before resetting it. :+1:

runelite-client/src/test/java/net/runelite/client/plugins/cluescrolls/ClueScrollPluginTest.java
```diff
@@ -61,4 +99,53 @@ public void getGetMirrorPoint()
 		converted = ClueScrollPlugin.getMirrorPoint(point, false);
 		assertEquals(point, converted);
 	}
-}
\ No newline at end of file
+
+	@Test
+	public void testLocationHintArrowCleared()
+	{
+		Guice.createInjector(BoundFieldModule.of(this)).injectMembers(this);
```
Adam-: We generally place this in `@Before`

Issue #10646, opened 2020-01-25
slayer item overlay broken on lizaards


botanicvelious:
**Describe the bug**
lizard, small lizard, and desert lizard do not show the ice above them anymore

**To Reproduce**
1. Go to lizards east of shanty pass
2. get them to 4 hp
3. dont see the ice item on top of them

**Expected behavior**
show the slayer item you need to use to kill them at 4hp

**Environment (please complete the following information):**
 - OS: Windows
 - RuneLite version: 1.6.4.1

**Additional context**
6c56880989a72143a8e670729d3e6f2fbd318682 removed the target weakness display for lizards.

pystic-mizza:
hey. just a heads up that this is still occurring, overlay still not showing.

PR #10681, opened 2020-01-28
Update Task.java: Fix the Lizard Slayer overlay

Paveldin:
Added the ice cooler item overlay to the lizards weakness. Sulphur lizard has been remove from the list because it is already present in the task list (further down) and is NOT vulnerable to ice cooler.

Related to Issue [#10646](https://github.com/runelite/runelite/issues/10646)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
@@ -123,7 +123,7 @@ enum Task
 	KURASK("Kurask", ItemID.KURASK),
 	ROGUES("Rogues", ItemID.ROGUE_MASK, "Rogue"),
 	LESSER_DEMONS("Lesser demons", ItemID.LESSER_DEMON_MASK),
-	LIZARDS("Lizards", ItemID.DESERT_LIZARD, "Desert lizard", "Sulphur lizard", "Small lizard", "Lizard"),
+	LIZARDS("Lizards", ItemID.DESERT_LIZARD, 4, ItemID.ICE_COOLER, "Desert lizard", "Small lizard", "Lizard"),
 	LIZARDMEN("Lizardmen", ItemID.LIZARDMAN_FANG, "Lizardman"),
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
```diff
@@ -123,7 +123,7 @@
 	KURASK("Kurask", ItemID.KURASK),
 	ROGUES("Rogues", ItemID.ROGUE_MASK, "Rogue"),
 	LESSER_DEMONS("Lesser demons", ItemID.LESSER_DEMON_MASK),
-	LIZARDS("Lizards", ItemID.DESERT_LIZARD, "Desert lizard", "Sulphur lizard", "Small lizard", "Lizard"),
+	LIZARDS("Lizards", ItemID.DESERT_LIZARD, 4, ItemID.ICE_COOLER, "Desert lizard", "Small lizard", "Lizard"),
```
Alexsuperfly: sulphur lizards are a valid lizard task, why are you removing them

Paveldin: I added the ItemID.ICE_COOLER because lizards requires that item to be kill, not the Sulphur lizard.  I noticed: `Players can kill these as an alternative to Desert Lizards, provided they have the higher Slayer level. However, sulphur lizards do not require ice coolers to finish off. `

We could keep the sulphur lizard in the tast, but add an ice cooler.

Alexsuperfly: from discussion in discord, just adding back the sulphur lizard would undesirably show the weakness overlay on them when it doesnt need to be.  A suggestion was proposed that another list of targets that dont have a weakness could be added, something like `Task(String name, int itemSpriteId, String[] targetNames, int weaknessThreshold, int weaknessItem, String... affectedTargetNames)`,  and just use that to tell the overlay to only draw the weakness on the appropriate monsters.

Adam-:
I think the correct way to fix this would be to associate the weaknesses to the npcs and not to the tasks.

PR #10683, opened 2020-01-28
Add support for left-click gem mine with Karamja Gloves with Menu Swapper

hjdarnel:
Right now Menu Entry Swapper "Teleport Item" doesn't allow left-clicking on the Karamja gloves for quick trips to the Gem Mine. This change will add this function to the gloves.


hjdarnel:
Closing this. Realizing that gloves have two possible teleports leave this ambiguous, and it's probably best to have an explicit dropdown for the option and to add it to the external menu entry swapper plugin instead.

PR #10690, opened 2020-01-29
Add ability and config to swap Lookup and Search in wiki plugin

hjdarnel:
Let me know if this code style is too ugly, I felt dumb using a switch for a boolean. I also removed a switch with a single case and instead used if.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2020, Henry Darnell <hjdarnel@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.wiki;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("wiki")
+public interface WikiConfig extends Config
+{
+
+	@ConfigItem(
+		keyName = "swapLookupSearch",
+		name = "Swap Lookup and Search",
+		description = "Swap Lookup to left-click Search on the Wiki button",
+		position = 1
+	)
+	default boolean swapLookupSearch()
+	{
+		return false;
+	}
+
+}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
@@ -28,16 +28,16 @@ import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
-@ConfigGroup("wiki")
+@ConfigGroup(WikiPlugin.CONFIG_GROUP_KEY)
 public interface WikiConfig extends Config
 {
 	@ConfigItem(
-		keyName = "swapLookupSearch",
-		name = "Swap Lookup and Search",
-		description = "Swap Lookup to left-click Search on the Wiki button",
+		keyName = "leftClickSearch",
+		name = "Left Click Search",
+		description = "Swap left-click on the Wiki button to Search",
 		position = 1
 	)
-	default boolean swapLookupSearch()
+	default boolean leftClickSearch()
 	{
 		return false;
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
@@ -31,7 +31,6 @@ import net.runelite.client.config.ConfigItem;
 @ConfigGroup("wiki")
 public interface WikiConfig extends Config
 {
-
 	@ConfigItem(
 		keyName = "swapLookupSearch",
 		name = "Swap Lookup and Search",
@@ -42,5 +41,4 @@ public interface WikiConfig extends Config
 	{
 		return false;
 	}
-
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
@@ -192,29 +192,17 @@ public class WikiPlugin extends Plugin
 			client.setAllWidgetsAreOpTargetable(true);
 		});
 
-		if (config.swapLookupSearch())
+		final int position = config.swapLookupSearch() ? 4 : 5;
+		icon.setAction(position, "Search"); // Start at option 4 so the target op is below the Search menuentry
+		icon.setOnOpListener((JavaScriptCallback) ev ->
 		{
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,46 @@
...
+package net.runelite.client.plugins.wiki;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("wiki")
+public interface WikiConfig extends Config
+{
+
```
TheStonedTurtle: Remove empty line

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,46 @@
...
+		keyName = "swapLookupSearch",
+		name = "Swap Lookup and Search",
+		description = "Swap Lookup to left-click Search on the Wiki button",
+		position = 1
+	)
+	default boolean swapLookupSearch()
+	{
+		return false;
+	}
+
```
TheStonedTurtle: Remove empty line

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -373,8 +399,8 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 				MenuEntry[] oldEntries = menuEntries;
 				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length - 1);
 				for (int ourEntry = oldEntries.length - 1;
-					ourEntry >= 2 && oldEntries[oldEntries.length - 1].getType() != MenuAction.SPELL_CAST_ON_WIDGET.getId();
-					ourEntry--)
+					 ourEntry >= 2 && oldEntries[oldEntries.length - 1].getType() != MenuAction.SPELL_CAST_ON_WIDGET.getId();
+					 ourEntry--)
```
TheStonedTurtle: Remove added whitespace

hjdarnel:
Thanks for the feedback, I've pushed some updates.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -204,6 +220,17 @@ private void onWidgetHiddenChanged(WidgetHiddenChanged ev)
...
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
+		{
+			// Since the Widget menu is configured on startup, it's easiest to just rebuild it when the config changes
+			shutDown();
+			startUp();
```
dekvall: You should probably use `addWidgets` instead of `startUp` and move the code in `shutDown` to a new method instead and call that here.

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -104,6 +110,14 @@
 
 	private boolean wikiSelected = false;
 
+	private static final String CONFIG_GROUP_KEY = "wiki";
```
dekvall:
```suggestion
	static final String CONFIG_GROUP_KEY = "wiki";
```

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,44 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.wiki;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("wiki")
```
dekvall:
```suggestion
@ConfigGroup(WikiPlugin.CONFIG_GROUP_KEY)
```

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -180,16 +194,18 @@ private void addWidgets()
...
 		});
-		icon.setAction(5, "Search"); // Start at option 5 so the target op is ontop
+
+		// set position to 5 if we want Lookup on left-click, or set position to 4 if we want Search
+		final int position = config.swapLookupSearch() ? 4 : 5;
+		icon.setAction(position, "Search");
 		icon.setOnOpListener((JavaScriptCallback) ev ->
 		{
-			switch (ev.getOp())
+			if (ev.getOp() == (position + 1))
```
dekvall:
```suggestion
			if (ev.getOp() == position + 1)
```

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiConfig.java
```diff
@@ -0,0 +1,44 @@
...
+public interface WikiConfig extends Config
+{
+	@ConfigItem(
+		keyName = "swapLookupSearch",
+		name = "Swap Lookup and Search",
```
dekvall: I think this name is a bit ambiguous, it's not clear to me what will happen when i toggle the option. Something like `Leftclick Search` would be better imo.

hjdarnel: Does it make more sense now?

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
hjdarnel: How does this look now?

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -112,6 +126,11 @@ public void startUp()
 
 	@Override
 	public void shutDown()
+	{
+		removeWidgets();
+	}
+
+	private void removeWidgets()
```
dekvall: Remove the `clientThread` from here invocation and invoke it similarily to `addWidgets` with `clientThread.invokeLater(this::removeWidgets)

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -204,6 +223,17 @@ private void onWidgetHiddenChanged(WidgetHiddenChanged ev)
 		}
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
+		{
+			// Since the Widget menu is configured on startup, it's easiest to just rebuild it when the config changes
```
dekvall: This comment feels outdated

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -180,16 +197,18 @@ private void addWidgets()
 			icon.setSpriteId(SpriteID.WIKI_SELECTED);
 			client.setAllWidgetsAreOpTargetable(true);
 		});
-		icon.setAction(5, "Search"); // Start at option 5 so the target op is ontop
+
+		// set position to 5 if we want Lookup on left-click, or set position to 4 if we want Search
```
dekvall: This comment feels redundant, the code right below it tells us the same thing

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -180,16 +197,18 @@ private void addWidgets()
...
+		final int position = config.leftClickSearch() ? 4 : 5;
```
dekvall: `position` feels a bit broad and could have a better name. Isn't it actually the index of the search op?

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -180,16 +197,18 @@ private void addWidgets()
...
+		icon.setAction(position, "Search");
 		icon.setOnOpListener((JavaScriptCallback) ev ->
 		{
-			switch (ev.getOp())
+			if (ev.getOp() == position + 1)
```
dekvall: I don't fully understand why this is `position + 1` but it was `case 6` earlier so i guess it's fine

abextm: javascriptcallback lives in cs2 land with 1 based indexes, setAction is in engine land with 0 based indexes

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
hjdarnel: It's the index but I don't personally understand the order (if there's only 3 elements in the array... why is position 4 on top 😄)

abextm: there are 10 slots in the array, and the optarget is as if it is in slot 4½, so you have to use something <=4 to put it before, and >=5 to put it after

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -204,6 +222,16 @@ private void onWidgetHiddenChanged(WidgetHiddenChanged ev)
 		}
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
+		{
+			clientThread.invokeLater(this::removeWidgets);
```
abextm: use a single invokeLater here

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
```diff
@@ -180,16 +197,17 @@ private void addWidgets()
 			icon.setSpriteId(SpriteID.WIKI_SELECTED);
 			client.setAllWidgetsAreOpTargetable(true);
 		});
-		icon.setAction(5, "Search"); // Start at option 5 so the target op is ontop
+
+		final int index = config.leftClickSearch() ? 4 : 5;
```
abextm:
```suggestion
		final int searchIndex = config.leftClickSearch() ? 4 : 5;
```

runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java
hjdarnel: That makes sense. Thanks!

PR #10699, opened 2020-01-31
loottracker - Add ignore events config option

### LINKED ISSUE #7482: Add monsters to ignore / hide in loot tracker
Whenever I use the loot tracker, and I kill irrelevant monsters (eg. Scavenger beasts in raids), all the loot of those monsters shows up in the loot tracker. When I reset the loot tracker and I kill another, it pops up again.

I'd like an option in the loot tracker settings menu, where we can enter names of monsters that we do not want the loot to appear from, in the loot tracker.

You could right click a monster name in the loot tracker, and select "Add to ignore monster list of loot tracker" to add it to that list.

Both solutions could also be implemented.

You could also hide the monsters instead of putting them on ignore, like you can do with items.

TheStonedTurtle:
Adds a config option similar to the IgnoreItems but that works on the event level.

Closes #7482

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
@@ -82,9 +82,9 @@ public interface LootTrackerConfig extends Config
 	@ConfigItem(
 		keyName = "syncPanel",
 		name = "Synchronize panel contents",
-		description = "Synchronize your local loot tracker with your online (requires being logged in). This means" +
-			" that panel is filled with portion of your remote data on startup and deleting data in panel deletes them" +
-			" also on server."
+		description = "Synchronize your local loot tracker with your server data (requires being logged in).<br/>" +
+			" This means the panel is filled with portions of your remote data on startup<br/>" +
+			" and deleting data in the panel also deletes it on the server."
 	)
 	default boolean syncPanel()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
@@ -90,4 +90,21 @@ public interface LootTrackerConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "ignoredEvents",
+		name = "Ignored groups",
+		description = "Configures which loot groups should be excluded from the panel UI"
+	)
+	default String getIgnoredEvents()
+	{
+		return "";
+	}
+
+	@ConfigItem(
+		keyName = "ignoredEvents",
+		name = "",
+		description = ""
+	)
+	void setIgnoredEvents(String key);
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
@@ -35,6 +35,7 @@ import java.awt.event.MouseEvent;
 import java.awt.image.BufferedImage;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Predicate;
 import javax.swing.BorderFactory;
@@ -355,6 +356,11 @@ class LootTrackerPanel extends PluginPanel
 		final LootTrackerRecord record = new LootTrackerRecord(eventName, subTitle, type, items, 1);
 		sessionRecords.add(record);
 
+		if (hideIgnoredItems && plugin.isEventIgnored(eventName))
+		{
+			return;
+		}
+
 		LootTrackerBox box = buildBox(record);
 		if (box != null)
 		{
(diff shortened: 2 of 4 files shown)


Adam-:
I'd probably rather see the solution described here: https://github.com/runelite/runelite/pull/10640#issuecomment-578461305

TheStonedTurtle:
The PR has been rebased and updated to be more like the solution linked in the previous comment.

tW4DQvx5ZI

TheStonedTurtle:
Changed UI instances of `event` to `group` as suggested.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -90,4 +90,21 @@ default boolean syncPanel()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "ignoredEvents",
+		name = "Ignored groups",
+		description = "Configures which loot groups should be excluded from the panel UI"
+	)
+	default String getIgnoredEvents()
```
Adam-, line 99: shouldn't this be hidden?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -734,6 +737,28 @@ boolean isIgnored(String name)
...
+		if (ignore)
+		{
+			ignoredSet.add(name);
+		}
+		else
+		{
+			ignoredSet.remove(name);
+		}
+
+		config.setIgnoredEvents(Text.toCSV(ignoredSet));
```
Adam-, line 753: If the config item for this is not actually going to be hidden, this will cause it to reorder itself randomly which seems not great

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -450,14 +456,19 @@ private void rebuild()
 		}
 		else
 		{
+			// Filter here instead of when building box for more accurate limiting
+			final List<LootTrackerRecord> session = sessionRecords.stream()
+				.filter(r -> !hideIgnoredItems || !plugin.isEventIgnored(r.getTitle()))
+				.collect(Collectors.toList());
```
Adam-, line 462: You could probably just make this reverse() + limit + forEach(buildBox)

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
TheStonedTurtle: I didn't hide this as the config option for hiding items is also not hidden.

I'm pretty sure having it hidden could cause issues with the limiting that would prevent someone from re-including an event.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
TheStonedTurtle: How would I prevent this re-ordering itself? 

This method is a copy/paste of the existing `toggleItem` method which is also not hidden in the config panel, so it must have the same issue.

TheStonedTurtle:
The review comments have been addressed and this PR is ready for re-review

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -450,15 +456,11 @@ private void rebuild()
...
-			{
-				start = sessionRecords.size() - MAX_LOOT_BOXES;
-			}
-			for (int i = start; i < sessionRecords.size(); i++)
-			{
-				buildBox(sessionRecords.get(i));
-			}
+			sessionRecords.stream()
+				.sorted(Collections.reverseOrder())
+				.filter(r -> !hideIgnoredItems || !plugin.isEventIgnored(r.getTitle()))
```
Adam-, line 461: I would maybe add a comment indicating why this is checking ignored (since buildBox also checks which looks unnecessary) as it is subtle.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -500,13 +507,17 @@ private LootTrackerBox buildBox(LootTrackerRecord record)
 
 		// Create box
 		final LootTrackerBox box = new LootTrackerBox(itemManager, record.getTitle(), record.getType(), record.getSubTitle(),
-			hideIgnoredItems, config.priceType(), config.showPriceType(), plugin::toggleItem);
+			hideIgnoredItems, config.priceType(), config.showPriceType(), plugin::toggleItem, plugin::toggleEvent, plugin.isEventIgnored(record.getTitle()));
```
Adam-, line 510: Just call `isEventIgnored` once and store the result

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -734,6 +738,28 @@ boolean isIgnored(String name)
 		return ignoredItems.contains(name);
 	}
 
+	void toggleEvent(String name, boolean ignore)
+	{
+		final Set<String> ignoredSet = new LinkedHashSet<>(ignoredEvents);
```
Adam-, line 743: additionally fix `toggleItem` to not randomly reorder things too

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
TheStonedTurtle: I don't understand what you are asking for me to change exactly.

I did notice `LootTrackerBox` could just use the `ignoredEvent` flag locally instead of storing the result, did you mean this?

Adam-: `isEventIgnored` is called on L485 and again here on L510. You can just call it once and store the result in a variable.

TheStonedTurtle:
The PR has been rebased and the comments addressed.

PR #10700, opened 2020-01-31
item prices plugin: added hover to show inventory/equipment total GE value

alex-zoltowski:
When hovering over either the inventory or equipment icon, it will display a tooltip with their respective total GE value.

A bit of redundancy with my function - makeTotalValueTooltip()
and the existing function - getItemStackValueText()

Not sure if it should be handled differently or not^^^

Let me know of your thoughts.

Alex

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
@@ -174,7 +174,7 @@ public class BankPlugin extends Plugin
 		switch (event.getEventName())
 		{
 			case "setBankTitle":
-				final ContainerPrices prices = bankCalculation.calculate(getBankTabItems());
+				final ContainerPrices prices = bankCalculation.calculate(getBankTabItems(), false);
 				if (prices == null)
 				{
 					return;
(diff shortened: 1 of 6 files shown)


jplsek:
Would utilizing ContainerCalculation from the bank plugin work? That would help reduce some duplicate code.

alex-zoltowski:
@jplsek Thanks for pointing that out, I think that would work - but the ContainerCalculation class is private... not sure how that should be handled.

jplsek:
Just make it public. It's possible the maintainers will want that moved outside the bank plugin if you use it, but I'll let them decide what should happen.

alex-zoltowski:
@jplsek  When implementing the ContainerCalculation class I realized it didn't account for noted items (because items in bank obviously aren't noted). My changes might be a stretch - but I added a boolean parameter to its function that will check for noted items if it is true.

Adam-:
Generally we don't import things cross-plugin like this. The bool for notes seems fine, you could maybe just not make it a flag since it generally won't get hit since you can't put notes into the bank.

alex-zoltowski:
@Adam- appreciate the feedback. How do you feel about my original change that adds the feature to show inventory/equipment total price when hovering over their respective icons? I can take out my recent cross-plugin changes and just manually calculate it in the item prices plugin. If you think this feature doesn't make much sense, just lemme know and I'll close this request.

Adam-:
I'm not sure if it makes sense or not. You can price check your whole inv already ingame via 2 clicks. Though it doesn't work in combat. I guess if the implementation was clean enough I don't mind.

PR #10701, opened 2020-01-31
Optimize loottracker UI

abextm:
This also fixes some bugs with the visible/hide icon getting out of sync with the actual state.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
@@ -26,11 +26,16 @@ package net.runelite.client.util;
 
 import java.awt.AWTException;
 import java.awt.Color;
+import java.awt.Component;
+import java.awt.Container;
+import java.awt.EventQueue;
 import java.awt.Font;
 import java.awt.Frame;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/util/SwingUtil.java
```diff
@@ -293,4 +299,54 @@ public static void addModalTooltip(AbstractButton button, String on, String off)
...
+
+	/**
+	 * Removes all of a component's children faster than calling removeAll() on it in many cases
+	 */
+	public static void fastRemoveAll(Container c)
+	{
+		// If we are not on the EDT this will deadlock, in addition to being totally unsafe
+		assert SwingUtilities.isEventDispatchThread();
+
+		// when a component is removed is has to be resize for some reason, but only if its valid
```
Nightfirecat: This sentence is kind of spaghetti. Is it meant to say something along the lines of `valid components must be resized when they are removed for some reason`?

abextm: it is supposed to say
`when a component is removed it has to be resized for some reason, but only if it's valid`

PR #10707, opened 2020-02-01
PluginManager: require start/stop to be invoked on the EDT

### LINKED ISSUE #10677: Plugins can get stuck being constantly toggled
If a plugin is toggled twice while the executor is blocked it will get in a loop of turning itself off and back on again because updating the pluginlistitem's view fires an event, which makes the executor switch state.

abextm:
The current way of invoking it on the executor would let the client have
a start and a stop for a plugin sitting in the executor's queue, making
it toggle endlessly. This only was because we were using invokeAndWait
in startPlugin, which can't be invoked on the EDT because it would
deadlock with itself.

Fixes #10677

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
@@ -33,6 +33,7 @@ import com.google.common.hash.HashingInputStream;
 import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -44,6 +45,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Function;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.swing.SwingUtilities;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.client.RuneLite;
 import net.runelite.client.RuneLiteProperties;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
```diff
@@ -215,12 +217,22 @@ public void startPlugins()
...
+						plugins.remove(plugin);
+					}
+				});
 			}
-			catch (PluginInstantiationException ex)
+			catch (InterruptedException | InvocationTargetException e)
 			{
-				log.warn("Unable to start plugin {}", plugin.getClass().getSimpleName(), ex);
-				plugins.remove(plugin);
+				throw new RuntimeException(e);
```
Adam-: Are you sure this is the correct way to handle this?

runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
```diff
@@ -242,9 +244,19 @@ private void refreshPlugins()
...
 				try
 				{
-					pluginManager.stopPlugin(p);
+					SwingUtilities.invokeAndWait(() ->
+					{
+						try
+						{
+							pluginManager.stopPlugin(p);
+						}
+						catch (Exception e2)
```
Adam-:
```suggestion
						catch (Exception e)
```

runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
abextm: this should never happen unless the thread gets interrupted (which should be unchecked anyway imo)

PR #10713, opened 2020-02-02
plugin hub: always wrap description text

jplsek:
To avoid the odd behavior where the jlabel only sometimes wraps,
the text is now wrapped in html which always wraps. The tooltip
can now also be removed.

The other way I've read about is to use  a jtextarea, but this approach should be fine.


jplsek:
Apparently, based on the discord, the ideal situation was to never wrap text.

The problem is that some third party plugins are using \<html\> tags in their descriptions already.
Since the desire is to not wrap text, then making sure that third party plugins are not using \<html\> tags needs to be enforced (if it isn't already for newer plugins).

PR #10715, opened 2020-02-02
PluginHubPanel: Limit height of description to 2 lines, enable word wrap

abextm:
This maintains the current behavior of only supporting html if the description starts with `<html>`

Supersedes/Closes #10713

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
@@ -28,6 +28,7 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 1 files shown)


jplsek:
The only minor style issue this has is that there is no ellipsis for long descriptions.

abextm:
I spend a while looking for a way to do that, and there basically is not unless I implement my own renderer.

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
```diff
@@ -98,6 +100,11 @@
 	private static final ImageIcon CONFIGURE_ICON_HOVER;
 	private static final Pattern SPACES = Pattern.compile(" +");
 	private static final JaroWinklerDistance DISTANCE = new JaroWinklerDistance();
+	private static final Escaper HTML_ESCAPER = Escapers.builder()
```
Adam-: There is an `HtmlEscapers.htmlEscaper()` if that helps.

PR #10739, opened 2020-02-06
Plugins: create XpUpdater plugin

Alexsuperfly:
moves the CML plugin to a new XpUpdater plugin
and includes TempleOSRS as another site to
send xp updates to


Adam-:
added in 553e56862927f8591fd37ad3e86425e640b23eaf

PR #10748, opened 2020-02-08
add new slepe anagram clue

### LINKED ISSUE #10747: New Hard Clue step missing from Clue Scroll plugin
**Issue:**
A new hard clue step was added in the Nightmare of Ashihama update, To solve you need to talk to Daer Krand in the Sisterhood Sanctuary under Slepe and solve his puzzle box.

**Solution:**
Add the step to the Clue Scroll plugin

**Screenshots:**
image

image

 - OS: Windows 10 Build 1909
 - RuneLite version: 1.6.5
 - Launcher version: 2.1.0

kmutchnick:
first contribution and was unsure of how exactly to get a worldpoint... any tips much appreciated!

Fixes #10747


Adam-:
You should use the developer tools to get the world location. Where did this current worldpoint come from you are using now?

See https://github.com/runelite/runelite/wiki/Using-the-client-developer-tools

kmutchnick:
Thanks @Adam- that one was from some transportation code I found from the rowboat to Slepe. I updated the WorldPoint using the developer tools to be right in front of the staircase you need to use to access the dungeon at WorldPoint(3728, 3302, 0) whereas the real worldpoint is WorldPoint(3821, 9777, 2).

I took a look at the rest of the clues, and it seems like in some cases, the ones not on the surface are placed on the surface, presumably so the world map will show the clue as expected, so that's why I made that choice. But we could update it to be the true one if that's preferred. 

One question I have is how the code knows to highlight the correct NPC without giving it an ID, and just the NPC's name? 

Thanks!

Adam-:
The code just matches based on NPC name and id, which is generally more robust when you don't need to identify specific types of NPCs.

PR #10752, opened 2020-02-08
client: add notification fired event

Trevor159:
For external plugins to do custom notification stuff.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/events/NotificationFired.java
+++ b/runelite-client/src/main/java/net/runelite/client/events/NotificationFired.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2020, Trevor <https://github.com/Trevor159>
+ * All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/events/NotificationFired.java
```diff
@@ -0,0 +1,35 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.events;
+
+import java.awt.TrayIcon;
+import lombok.Data;
+
+@Data
```
Nightfirecat: Unless this has reason to be mutable, `@Value` is more appropriate here.

PR #10773, opened 2020-02-11
Add RS-specific time units

Nightfirecat:
This PR adds a list of time units used in some duration calculations, specifically client ticks and game ticks, and applies them to plugins using `Duration` calculations where they would be applicable.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/RSTimeUnit.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/RSTimeUnit.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2020, Jordan Atwood <jordan.atwood423@gmail.com>
+ * All rights reserved.
+ *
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/util/RSTimeUnit.java
```diff
@@ -0,0 +1,83 @@
...
+	private final String name;
+	private final Duration duration;
+
+	RSTimeUnit(String name, Duration estimatedDuration)
+	{
+		this.name = name;
+		duration = estimatedDuration;
+	}
+
+	public boolean isDurationEstimated()
```
Adam-: What is all of this? are they missing `@Override`?

Nightfirecat: Yes. I've added `@Override`s to these methods now.

PR #10780, opened 2020-02-12
Add ability to output emojis to chatbox

hjdarnel:
This adds a Menu Entry to the All tab right click, when Emojis configuration setting is true. 
image
image
image


dekvall:
I don't think it makes much sense to add this in it's current state. Without being able to click on an emoji to append its tag to the input, this is more or less a built in wiki page. 

However, if we do want to add something like this, I think that it's preferable to have the tag displayed under the icon instead of as a hover text since having to move the mouse in order to see what to type would get annoying pretty quickly.

hjdarnel:
We discussed in Discord making this a right click on chat buttons and would instead just emit the icons and triggers into chat with `Client::addChatMessage`, probably with a timeout and be removed after a couple of seconds like the log in/out messages. I will update this to reflect

hjdarnel:
To continue this conversation here: which Chat tab should have the right click option? I'm preferential to `All`, `Game`, or each tab. I'm currently considering something like "Show Emojis" for the menu option verbiage?

Nightfirecat:
I think adding it to the "All" tab makes the most sense to me, and maybe also to the active tab? I'd start with just adding it to the "All" tab, I think. And `Show emojis` sounds like a good menu action name. :+1:

hjdarnel:
ok, I removed the navbar and panel in favor of leaving in the "emitting into chat" implementation. I don't like the config name very much. Any feedback on the formatting? It's hard to make it look too good because the font isn't monospaced. We could do something with `StringUtils.leftPad()` to make the output formatted similar char lengths but it still won't look like a table.

I updated PR description and title appropriately

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
+		return configManager.getConfig(EmojiConfig.class);
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event) throws Exception
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
+		{
+			shutDown();
+			startUp();
```
dekvall: I don't think you need to do a full shutDown/startUp, isn't it enough to just remove the potentially emitted message when the config for emitting the message is set to false?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
+	private static final Pattern TAG_REGEXP = Pattern.compile("<[^>]*>");
+	private static final Pattern WHITESPACE_REGEXP = Pattern.compile("[\\s\\u00A0]");
+	private static final String EMIT = "Show";
+	private static final String MENU_TARGET = "Emoji Key";
+	private static final int EMOJI_EMIT_DURATION = 20;
+	private static final int EMOJI_PER_LINE = 8;
+	/**
+	 * queue of temporary messages added to the client
+	 */
+	private final Deque<EmojiEmitMessage> emojiEmitMessages = new ArrayDeque<>();
```
dekvall: I don't think we need to have multiples of these, if the player chooses to emit emojis in quick succession imo it's fine to just remove the current emitted message and send a new one.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -166,6 +222,134 @@ public void onOverheadTextChanged(final OverheadTextChanged event)
...
+		}
+
+		MenuEntry[] entries = client.getMenuEntries();
+		for (MenuEntry entry : entries)
+		{
+			if (entry.getOption().equals(EMIT) && entry.getTarget().equals(MENU_TARGET))
+			{
+				return;
+			}
+		}
```
dekvall: Is this required? I think it should be enough to just check `event.getOption()`

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -166,6 +222,134 @@ public void onOverheadTextChanged(final OverheadTextChanged event)
...
+	{
+		ChatMessageBuilder builder = new ChatMessageBuilder();
+		Emoji[] allEmoji = getAllEmoji();
+
+		for (int i = 0; i < allEmoji.length; i++)
+		{
+			Emoji emoji = allEmoji[i];
+			builder
+				.img(modIconsStart + emoji.ordinal())
+				.append("  " + Text.unescapeTags(emoji.trigger) + "  ");
```
dekvall: Maybe something like `FontTypeFace#getTextWidth` can be used when aligning these(?)
The current formatting unfortunately looks really messy.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
Nightfirecat: Furthermore, it is very bad practice to directly call `startUp` and `shutDown`. If the code they contain should be accessed from multiple sources, put it into a separate function.

runelite-api/src/main/java/net/runelite/api/MenuAction.java
```diff
@@ -272,6 +272,10 @@
 	 * Menu action for configuring runelite overlays.
 	 */
 	RUNELITE_OVERLAY_CONFIG(1502),
+	/**
+	 * Menu action for triggering Emoji output to chat.
+	 */
+	RUNELITE_EMOJI_OUTPUT(1503),
```
Nightfirecat: Don't add a new menu action type. Use `RUNELITE` instead and check the menu entry's text when clicked.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2019, Lotto <https://github.com/devLotto>
+ * Copyright (c) 2020, Lotto <https://github.com/devLotto>
```
Nightfirecat: Don't update copyright date of other entries.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
+	@Provides
+	EmojiConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(EmojiConfig.class);
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event) throws Exception
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
```
Nightfirecat: This should also check the config key so future configurations (if any) will not enter this block.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
+		{
+			shutDown();
+			startUp();
+		}
+	}
+
 	@Override
-	protected void startUp()
+	protected void startUp() throws Exception
```
Nightfirecat: Does `startUp` throw an exception now? Why is this added?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
 	private int modIconsStart = -1;
 
+	@Provides
+	EmojiConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(EmojiConfig.class);
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event) throws Exception
```
Nightfirecat: What exception is this method expected to throw?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -54,19 +77,52 @@
...
+	{
+		if (event.getGroup().equals(CONFIG_GROUP_KEY))
+		{
+			shutDown();
```
Nightfirecat: Where is `shutDown` defined?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -166,6 +222,134 @@ public void onOverheadTextChanged(final OverheadTextChanged event)
 		event.getActor().setOverheadText(updatedMessage);
 	}
 
+	@Subscribe
+	public void onMenuEntryAdded(MenuEntryAdded event)
+	{
+		int childId = WidgetInfo.TO_CHILD(event.getActionParam1());
+		if (!config.showEmitEmojiEntry() || childId != WidgetInfo.RESIZABLE_VIEWPORT_ALL_CHAT_TAB.getChildId())
```
Nightfirecat: Why is this only comparing child IDs? And why is it comparing only to `RESIZABLE_VIEWPORT_ALL_CHAT_TAB`? What happened to checking `FIXED_VIEWPORT_ALL_CHAT_TAB`?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: is it enough to check `		if (event.getGroup().equals(CONFIG_GROUP_KEY) && event.getKey().equals("showEmitEmojiEntry"))
` or do we need to point to a field of the EmojiConfig class?

Nightfirecat: That's fine.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: The child Id of the All tab on fixed and resize is the same, so I didn't catch that. I'll check both verbosely.

Do we need to check the Group Id of the `MenuEntryAdded` too?

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: Yes, we don't need to do anything on config toggle on, and on toggle off we can just remove the message. Good call. Thanks for the explanation too.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: @Nightfirecat how do we check the exact id of the widget instead of the child id? Can we do something like this? 

```
int widgetId = event.getActionParam1();
if (!config.showEmitEmojiEntry()
	|| widgetId != WidgetInfo.RESIZABLE_VIEWPORT_ALL_CHAT_TAB.getId()
	|| widgetId != WidgetInfo.FIXED_VIEWPORT_ALL_CHAT_TAB.getId())
{
	return;
}
```

I've tried this but I'm not sure if it's correct or not, it's not working but it's possible the `WidgetInfo.RESIZABLE_VIEWPORT_ALL_CHAT_TAB` and `FIXED_VIEWPORT_ALL_CHAT_TAB` are just wrong (since I've added them). Not sure which id in the widget inspector to use

Nightfirecat: That code looks correct to me. The widget inspector will show the group and child IDs in the tree on the left, and should display the `ID` as a row when selecting the widget.

On second thought, it will likely be easier to use the MenuManager class to manage adding these custom menu entries to these widgets.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/Emoji.java
```diff
@@ -90,7 +90,7 @@
 
 	private static final Map<String, Emoji> emojiMap;
 
-	private final String trigger;
```
dekvall: Use `@Getter` instead

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: Yup, I completely agree. I updated the PR description with the output as it stands with the latest changes.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: I've refactored to use the MenuManager helper. Let me know how it looks.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/Emoji.java
hjdarnel: Thanks :)

runelite-client/src/test/java/net/runelite/client/util/TextTest.java
```diff
@@ -42,4 +42,12 @@ public void removeTags()
 		assertEquals("Remove no tags", Text.removeTags("Remove no tags"));
 	}
 
+	@Test
+	public void unescapeTags()
```
Nightfirecat: This should include cases where tags are removed, since that is included as part of the behavior

runelite-client/src/main/java/net/runelite/client/util/Text.java
```diff
@@ -150,6 +151,38 @@ else if (c != '\r')
 		return out.toString();
 	}
 
+	/**
+	 * Unescape a string for widgets, replacing &lt;lt&gt; and &lt;gt&gt; with their unescaped counterparts
+	 */
+	public static String unescapeTags(String str)
+	{
+		StringBuffer out = new StringBuffer(str.length());
```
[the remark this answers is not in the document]
hjdarnel: I copy pasted this from escapeTags and flipped some logic. I'll use StringBuilder instead.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiConfig.java
```diff
@@ -0,0 +1,59 @@
...
+		return false;
+	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "emojiEmitDuration",
+		name = "Message duration",
+		description = "Configure how long the emoji hint message stays in chat"
+	)
+	@Units(Units.TICKS)
```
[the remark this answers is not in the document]
hjdarnel: Yeah, I agree. I changed this.

hjdarnel:
> Also, why is `icon.png` added?

I originally added this to be uniform with other nav bar icons, where `icon.png` refers to displayed icon. Since we're not doing that way anymore I removed it.

runelite-client/src/main/java/net/runelite/client/util/Text.java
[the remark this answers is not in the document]
hjdarnel: I stand corrected, we need to use StringBuffer because `Matcher#appendReplacement(StringBuilder sb, String replacement)` only exists since Java 9

runelite-client/src/test/java/net/runelite/client/util/TextTest.java
```diff
@@ -42,4 +42,16 @@ public void removeTags()
...
+	{
+		assertEquals("Test", Text.unescapeTags("Test"));
+		assertEquals("<:)", Text.unescapeTags("<lt>:)"));
+		assertEquals(">:)", Text.unescapeTags("<gt>:)"));
+		assertEquals("\n", Text.unescapeTags("<br>"));
+		assertEquals("<col=FFFFFF This is a very special message.</col> >:)", Text.unescapeTags("<col=FFFFFF This is a very special message.</col> <gt>:)"));
+		assertEquals("<col=FFFFFF This is a very special message.</col> <:)", Text.unescapeTags("<col=FFFFFF This is a very special message.</col> <lt>:)"));
+		assertEquals("<col=FFFFFF This is a very special message.</col> \n:)", Text.unescapeTags("<col=FFFFFF This is a very special message.</col> <br>:)"));
+		assertEquals("<col=FFFFFF This is a very special message.</col>Not so much.", Text.unescapeTags("<col=FFFFFF This is a very special message.</col>Not so much."));
```
Nightfirecat: I meant test a string with some tag present and observe that the tag would be preserved in the result. eg. `<col=FFFFFF>Colored text</col> and some other text`

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -63,12 +82,77 @@
...
+	{
+		return configManager.getConfig(EmojiConfig.class);
+	}
+
+	/**
+	 * Remove tags, except for &lt;lt&gt; and &lt;gt&gt;
+	 *
+	 * @return String with tags removed
+	 */
+	private static String removeTags(String str)
```
Nightfirecat: This method has still been moved compared to where it was originally. This should be appearing on what is currently L333 of this file. Look at the diff view of this PR to see what I mean. (down where you've added `timeoutEmojiMessages()`)

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -63,12 +82,77 @@
...
+
+	static final String CONFIG_GROUP_KEY = "emoji";
+	private static final String EMIT = "Show";
+	private static final String MENU_TARGET = "Emoji Key";
+	private static final WidgetMenuOption FIXED_INVENTORY_ALL_TAB = new WidgetMenuOption(EMIT,
+		MENU_TARGET, WidgetInfo.FIXED_VIEWPORT_ALL_CHAT_TAB);
+	private static final WidgetMenuOption RESIZABLE_INVENTORY_ALL_TAB = new WidgetMenuOption(EMIT,
+		MENU_TARGET, WidgetInfo.RESIZABLE_VIEWPORT_ALL_CHAT_TAB);
+	private static final int EMOJI_PER_LINE = 7;
+	//	 Holds current emoji message, if any
```
Nightfirecat:
```suggestion
	//	Holds current emoji message, if any
```

runelite-client/src/test/java/net/runelite/client/util/TextTest.java
hjdarnel: My bad, I didn't realize what I had added there.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
hjdarnel: Yeah, sorry, the diff was confusing me. _Best as I can tell_ this is put back now.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -63,12 +82,52 @@
 	@Inject
 	private ChatMessageManager chatMessageManager;
 
+	@Inject
+	private Provider<MenuManager> menuManager;
+
+	@Inject
+	private ClientToolbar clientToolbar;
```
Nightfirecat: This is unused.

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiConfig.java
```diff
@@ -0,0 +1,59 @@
...
+import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Units;
+
+@ConfigGroup(EmojiPlugin.CONFIG_GROUP_KEY)
+public interface EmojiConfig extends Config
+{
+	@ConfigItem(
+		position = 1,
+		keyName = "showEmitEmojiEntry",
+		name = "Add trigger to All chat tab",
```
Nightfirecat:
```suggestion
		name = "Add Emoji key menu option to All chat tab",
```

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiConfig.java
```diff
@@ -0,0 +1,59 @@
...
+		description = "Add trigger to All tab right-click to display possible emojis"
```
Nightfirecat:
```suggestion
		description = "Add menu option to All tab right-click to display a key of available emojis and their trigger text"
```

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiConfig.java
```diff
@@ -0,0 +1,59 @@
...
+	)
+	default boolean showEmitEmojiEntry()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "emojiEmitDuration",
+		name = "Message duration",
```
Nightfirecat:
```suggestion
		name = "Emoji key duration",
```

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiConfig.java
```diff
@@ -0,0 +1,59 @@
...
+		description = "Configure how long the emoji hint message stays in chat"
```
Nightfirecat:
```suggestion
		description = "Configure how long the emoji key message is displayed in chat"
```

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -222,4 +329,63 @@ private static String removeTags(String str)
...
+		if (removed && !force)
+		{
+			clientThread.invoke(() -> client.runScript(ScriptID.BUILD_CHATBOX));
+		}
+	}
+
+	private String getEmojiMessage()
+	{
+		ChatMessageBuilder builder = new ChatMessageBuilder();
+		ChatMessageBuilder spacer = new ChatMessageBuilder();
```
Nightfirecat: Move this inside the loop and remove the `spacer = new ChatMessageBuilder();` line at the end of the loop

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -166,6 +225,54 @@ public void onOverheadTextChanged(final OverheadTextChanged event)
...
 
+
+	@Subscribe
+	public void onGameTick(GameTick gameTick)
+	{
+		timeoutEmojiMessages(false);
+	}
+
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked event)
```
Nightfirecat: This should subscribe to `WidgetMenuOptionClicked` now that MenuManager is being used to add the menu options. Similarly, this should guard on both the menu option text (like it currently does) and that the widget is one of the all tabs. I would reference other plugins making use of MenuManager such as the inventory tags and instance map plugins to see similar usages there.

Ex:

```java
private static Set<WidgetInfo> ALL_TABS = ImmutableSet.of(
	FIXED_INVENTORY_ALL_TAB,
	RESIZABLE_INVENTORY_ALL_TAB
);

@Subscribe
public void onWidgetMenuOptionClicked(WidgetMenuOptionClicked event)
{
	if (ALL_TABS.contains(event.getWidget())
		&& event.getMenuOption().equals(EMIT)
		&& event.getMenuTarget().equals(MENU_TARGET))
	{
		timeoutEmojiMessages(true);
		addEmojiPaletteMessage();
	}
}
```

Creating an `ALL_TABS` collection would have the benefit of de-duplicating the menu manager calls later on as well.

raiyni:
This needs rebased

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -198,28 +330,62 @@ String updateMessage(final String message)
 		return Strings.join(messageWords, " ");
 	}
 
-	/**
-	 * Remove tags, except for &lt;lt&gt; and &lt;gt&gt;
-	 *
-	 * @return
-	 */
-	private static String removeTags(String str)
+	private void timeoutEmojiMessages(boolean force)
```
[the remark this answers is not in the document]
hjdarnel: We still need to remove the current outputted message if the user triggers the menu option again, so I think we still need `currentMessage` field to exist in the class so we can reference it without passing it by params.

If you want I can still use `executorService` to call `timeoutEmojiMessage` after x seconds/ticks but we do need an instance of that class I think.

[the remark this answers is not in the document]
devLotto: Yes. The field would still need to exist if you want to do that, but it would just be a MessageNode instead.

hjdarnel:
I rebased and addressed most of the feed. I was having an unescaped while loop in my `getEmojiMessage()` when I used `new StringBuilder` inside my loop, so I need to take another crack at it tomorrow or this weekend. Thanks for your feedback!

runelite-client/src/main/java/net/runelite/client/plugins/emojis/EmojiPlugin.java
```diff
@@ -198,28 +330,62 @@ String updateMessage(final String message)
...
+			{
+				spacer.append(" ");
+			}
+
+			if (i % EMOJI_PER_LINE == 0)
+			{
+				spacer.append("\n");
+			}
+
+			builder.append(Text.unescapeTags(spacer.build()));
```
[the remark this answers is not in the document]
hjdarnel: I'm not sure what's going on here. You can see I moved the instantiation to inside the loop. When I try to swap to a `StringBuilder` I get a continual loop here.
 ```
while (fontFace.getTextWidth(spacer.build()) < 55)
{
	spacer.append(" ");
}
```

and I can't figure out why. When I compare the value of `fontFace.getTextWidth(spacer.build()` before and after appending some spaces, the width is indeed increasing. Why would this loop forever then?

Otherwise, I've addressed all feedback so far except using `StringBuilder` instead of `ChatMessageBuilder`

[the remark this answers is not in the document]
hjdarnel: Yep, not sure what's going on here. When I use a `StringBuilder#append(String)` to add spaces in this loop, the `FontTypeFace fontFace.getTextWidth(spacer.build())` call isn't increasing in value. Even when testing against a string with length > 100, `getTextWidth()` is returning the same value as though the string is getting trimmed or something. Not sure what to do here.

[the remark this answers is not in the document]
Nightfirecat: You should set a breakpoint here and figure out why it only works before making that change.

hjdarnel:
I moved this out into an external plugin to fit what I originally had in mind. https://github.com/runelite/plugin-hub/pull/145

Thank you guys for all your feedback, hate to close this after all that but I like that implementation more.

PR #10787, opened 2020-02-13
chat commands: add Bounty Hunter, Bounty Hunter Rogue, and Last Man Standing commands

### LINKED ISSUE #9855: Add a new chat command for LMS
Similar to other chat commands like !kc !pb and !clues, can there be a chat command added for LMS and possibly BH?

An example would be: 
!lms
Status: 1679 (Legend)  Wins: 55  Kills: 458  Games Played: 168  Rank: 187

Or along those lines, OSRS uses the word Rank already to describe what I called Status, so maybe another word/phrase could be used like "In Game Rank" or "Grade".

sadegr:
Closes #9855

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -99,7 +99,10 @@ public class ChatCommandsPlugin extends Plugin
 	private static final String TOTAL_LEVEL_COMMAND_STRING = "!total";
 	private static final String PRICE_COMMAND_STRING = "!price";
 	private static final String LEVEL_COMMAND_STRING = "!lvl";
+	private static final String BOUNTY_HUNTER_HUNTER_COMMAND = "!bh";
+	private static final String BOUNTY_HUNTER_ROGUE_COMMAND = "!bhrogue";
 	private static final String CLUES_COMMAND_STRING = "!clues";
+	private static final String LAST_MAN_STANDING_COMMAND = "!lms";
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -1055,6 +1061,101 @@ private void combatLevelLookup(ChatMessage chatMessage, String message)
...
+					hiscoreSkill = result.getBountyHunterHunter();
+					break;
+				case BOUNTY_HUNTER_ROGUE:
+					hiscoreSkill = result.getBountyHunterRogue();
+					break;
+				case LAST_MAN_STANDING:
+					hiscoreSkill = result.getLastManStanding();
+					break;
+				default:
+					return;
```
Nightfirecat: This should log some warning if called with a non-implemented minigame.

sadegr:
It hadn't actually occurred to me until you mentioned it that BH is no longer in the game at this time. It appears on the hiscores so I don't see any harm in adding it... the only catch is that if Jagex were to remove it from the hiscores (which admittedly seems unlikely, but you never know), any time someone types one of the BH commands, any user with those commands enabled would get an unhandled `IndexOutOfBoundsException` in `HiscoreResultBuilder.build()` that will bubble up and may even crash the entire application (I didn't look very far but didn't see it being handled).

Let me know what you think... I'm not exactly sure how hypothetical this scenario is, but wanted to point it out in case someone thinks we should handle it ahead of time.

Adam-:
We generally keep things around until removed from hiscores (TL, the old DMM seasonal, etc). So this is fine.

PR #10792, opened 2020-02-13
timers: Don't add imbued heart timer on magic debuff

Nightfirecat:
This commit prevents combinations of Chambers of Xeric potions
(Overload, Xeric's Aid) amounting to the same boost amount as the imbued
heart would provide from adding an imbued heart timer.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -30,6 +30,8 @@ import com.google.inject.Provides;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
+import lombok.AccessLevel;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
According to some issue reports from Discord, the timer is also restarting while at the maximum boost amount. (Example: A player was seeing their timer restart sometimes while killing Kraken when using their imbued heart)
I'll write another test tomorrow to ensure that duplicate `StatChanged` events for a given level do not trigger multiple timer additions, but I believe the code as written does not need to change to support that.

soulsword285:
There was a report somewhere in the discord where someone was getting the imbued heart timer during a raid.

I wasn't following it too closely but I did notice the same thing in my raid this morning.

Just for some extra info - during raids, it is possible that you are overloaded and then get burned. Getting burned reduces all stats by 5 which may be the cause of it happening within raids. The -5 stat is then increase after the 15 seconds it takes for the overload to refresh.

Nightfirecat:
This doesn't solve the issue since overloads at raids can cause you to boost to the same amount (when boosting from some de-buffed stage).

PR #10797, opened 2020-02-14
client: add sessionfile and config arguments

Trevor159:
budget profiles

The behavior of this is as follows:
input file is relative to the RL dir unless an absolute path or a path that starts with "./" or ".\"
The --config arg does nothing if the user is logged into runelite (it keeps the current behavior of putting the config in the rl/profile dir)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -177,10 +177,12 @@ public class RuneLite
 		final OptionParser parser = new OptionParser();
 		parser.accepts("developer-mode", "Enable developer tools");
 		parser.accepts("debug", "Show extra debugging output");
-		parser.accepts("sessionfile", "Use a specified session file")
+
+		final ArgumentAcceptingOptionSpec<File> sessionfile = parser.accepts("sessionfile", "Use a specified session file")
 			.withRequiredArg().defaultsTo(DEFAULT_SESSION_FILE)
 			.withValuesConvertedBy(new FileConverter());
-		parser.accepts("config", "Use a specified config file")
+
+		final ArgumentAcceptingOptionSpec<File> configfile = parser.accepts("config", "Use a specified config file")
 			.withRequiredArg().defaultsTo(DEFAULT_CONFIG_FILE)
 			.withValuesConvertedBy(new FileConverter());
 
@@ -251,9 +253,6 @@ public class RuneLite
 				}
 			}
 
-			File sessionfile = (File) options.valueOf("sessionfile");
-			File configfile = (File) options.valueOf("config");
-
 			PROFILES_DIR.mkdirs();
 
 			final long start = System.currentTimeMillis();
@@ -261,8 +260,8 @@ public class RuneLite
 			injector = Guice.createInjector(new RuneLiteModule(
 				clientLoader,
 				developerMode,
-				sessionfile,
-				configfile));
+				options.valueOf(sessionfile),
+				options.valueOf(configfile)));
 
 			injector.getInstance(RuneLite.class).start();
 
@@ -392,13 +391,11 @@ public class RuneLite
 			}
 		}
 
-		@Override
 		public Class<? extends File> valueType()
 		{
 			return null;
 		}
 
-		@Override
 		public String valuePattern()
 		{
 			return null;
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -33,7 +33,6 @@ import com.google.inject.Injector;
 import java.io.File;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Locale;
 import javax.annotation.Nullable;
@@ -43,6 +42,7 @@ import javax.swing.SwingUtilities;
 import joptsimple.ArgumentAcceptingOptionSpec;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
+import joptsimple.ValueConversionException;
 import joptsimple.ValueConverter;
 import joptsimple.util.EnumConverter;
 import lombok.Getter;
@@ -55,12 +55,12 @@ import net.runelite.client.chat.CommandManager;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.discord.DiscordService;
 import net.runelite.client.eventbus.EventBus;
+import net.runelite.client.externalplugins.ExternalPluginManager;
 import net.runelite.client.game.ClanManager;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.game.LootManager;
 import net.runelite.client.game.chatbox.ChatboxPanelManager;
 import net.runelite.client.menus.MenuManager;
-import net.runelite.client.externalplugins.ExternalPluginManager;
 import net.runelite.client.plugins.PluginManager;
 import net.runelite.client.rs.ClientLoader;
 import net.runelite.client.rs.ClientUpdateCheckMode;
@@ -180,11 +180,11 @@ public class RuneLite
 
 		final ArgumentAcceptingOptionSpec<File> sessionfile = parser.accepts("sessionfile", "Use a specified session file")
 			.withRequiredArg().defaultsTo(DEFAULT_SESSION_FILE)
-			.withValuesConvertedBy(new FileConverter());
+			.withValuesConvertedBy(new ConfigFileConverter());
 
 		final ArgumentAcceptingOptionSpec<File> configfile = parser.accepts("config", "Use a specified config file")
 			.withRequiredArg().defaultsTo(DEFAULT_CONFIG_FILE)
-			.withValuesConvertedBy(new FileConverter());
+			.withValuesConvertedBy(new ConfigFileConverter());
 
 		final ArgumentAcceptingOptionSpec<ClientUpdateCheckMode> updateMode = parser
 			.accepts("rs", "Select client type")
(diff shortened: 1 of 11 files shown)


Trevor159:
> I would not limit this to files in RL dir and just require full path because its kind of confusing limitation

I disagree because the usecase for most users will be just to add a arg to the launcher exe and I do not trust them if they get to do more than just rename the file. I would prefer for it to be consistent with where we currently place files.

deathbeam:
> > I would not limit this to files in RL dir and just require full path because its kind of confusing limitation
> 
> I disagree because the usecase for most users will be just to add a arg to the launcher exe and I do not trust them if they get to do more than just rename the file. I would prefer for it to be consistent with where we currently place files.

Then at least document it because atm its not documented at all, and also not obvious for people that actually want to use it. But also, you are allowing users to pass custom properties, so it just makes 0 sense to me to limit the path to runelite directory.

abextm:
imo it should be relative to the rldir by default, If they pass an absolute path use that.

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -172,6 +178,8 @@ public static void main(String[] args) throws Exception
 		final OptionParser parser = new OptionParser();
 		parser.accepts("developer-mode", "Enable developer tools");
 		parser.accepts("debug", "Show extra debugging output");
+		parser.accepts("sessionfile", "Use a specified session file").withRequiredArg();
```
abextm:
```suggestion
		ArgumentAcceptingOptionSpec<String>  sessionFileOp = parser.accepts("sessionfile", "Use a specified session file")
		.withRequiredArg()
		.defaultsTo("session");
```
then use you can just do
```
String sessionFile = options.valueOf(sessionFileOp)
```

runelite-client/src/main/java/net/runelite/client/RuneLiteModule.java
```diff
@@ -61,17 +61,23 @@
 
 	private final Supplier<Applet> clientLoader;
 	private final boolean developerMode;
+	private final String sessionfile;
```
abextm: also imo these should be `File`s

Trevor159: Maybe I was doing something wrong but bindConstant did not like Files

deathbeam: Yeah because File is not constant

runelite-client/src/main/java/net/runelite/client/account/SessionManager.java
```diff
@@ -52,35 +55,49 @@
...
+		ConfigManager configManager,
+		EventBus eventBus,
+		WSClient wsClient)
 	{
 		this.configManager = configManager;
 		this.eventBus = eventBus;
 		this.wsClient = wsClient;
+
+		Path path = Paths.get(sessionfile);
+		if (path.isAbsolute())
```
deathbeam: What if someone is doing `./my-session-file` what is relative path and completely valid use-case? Pretty sure then this check will be incorrect

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -172,6 +176,10 @@ public static void main(String[] args) throws Exception
 		final OptionParser parser = new OptionParser();
 		parser.accepts("developer-mode", "Enable developer tools");
 		parser.accepts("debug", "Show extra debugging output");
+		parser.accepts("sessionfile", "Use a specified session file")
+			.withRequiredArg().defaultsTo(DEFAULT_SESSION_FILE);
+		parser.accepts("config", "Use a specified config file")
```
Adam-: Did you try `ofType(File.class);`? Does that work?

Trevor159: I wanted the input as a string so I can change the directory to the RL directory. The most common usecase will be args in the launcher with users that will not understand that the file is going into the cwd. My current implementation also lets users use the cwd or an absolute path if they want.

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -177,9 +178,11 @@ public static void main(String[] args) throws Exception
 		parser.accepts("developer-mode", "Enable developer tools");
 		parser.accepts("debug", "Show extra debugging output");
 		parser.accepts("sessionfile", "Use a specified session file")
```
deathbeam: You can just store this as local variable so you can get File from it without casting (see updateMode)

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -358,6 +374,37 @@ public void shutdown()
...
+				return new File(fileName);
+			}
+			else
+			{
+				return new File(RuneLite.RUNELITE_DIR, fileName);
+			}
+		}
+
+		@Override
+		public Class<? extends File> valueType()
```
deathbeam: I dont think you need to actually override this and valuePattern

deathbeam: By not overriding it I meant those methods should not need to be there

deathbeam:
Oooh i see, the other one is using EnumConverter that already overrides those methods. Alright then, then you should keep the methods and also override annotation

deathbeam:
Cleaned this up and added some checks for validity of the input file

runelite-client/src/main/java/net/runelite/client/util/JoptFileConverter.java
```diff
@@ -0,0 +1,74 @@
...
+			file = new File(fileName);
+		}
+		else
+		{
+			file = new File(RuneLite.RUNELITE_DIR, fileName);
+		}
+
+		final Path path = Paths.get(file.getAbsolutePath());
+
+		if (!Files.exists(path) || !Files.isRegularFile(path) || !Files.isWritable(path)) {
```
Trevor159: Should the file have to exist? With the current config/session implementation it will generate the file if it does not exist.

deathbeam: Oh, right. Yeah i guess, I will change the check

runelite-client/src/main/java/net/runelite/client/util/JoptFileConverter.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import joptsimple.ValueConversionException;
+import joptsimple.ValueConverter;
+import net.runelite.client.RuneLite;
+
+public class JoptFileConverter implements ValueConverter<File>
```
abextm: This should have a name that describes what it is doing

deathbeam:
Updated

runelite-client/src/main/java/net/runelite/client/util/JoptValueToFileConverter.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import joptsimple.ValueConversionException;
+import joptsimple.ValueConverter;
+import net.runelite.client.RuneLite;
+
+public class JoptValueToFileConverter implements ValueConverter<File>
```
abextm: thats even worse

deathbeam: Well its literally what it is doing

abextm: The classname should tell my why its not JOptSimple's FileConverter

deathbeam: Hmm? jopt does not have FileConverter

abextm: It creates a `ValueConverter<File>` internally, which is what I was referencing, which is different than this. Also don't prefix it with `Jopt`

deathbeam: Its not prefix its simply converting jopt value to file. If you have any ideas for name instead of just not liking current name then that would be a bit more helpful :d

abextm: it should probably be `ConfigFileConverter` or something and be private to `n.r.c.RuneLite` since nothing else has a use for it

deathbeam: Alright updated

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -373,38 +375,50 @@ public void shutdown()
...
 			}
 			else
 			{
-				return new File(RuneLite.RUNELITE_DIR, fileName);
+				file = new File(RuneLite.RUNELITE_DIR, fileName);
 			}
+
+			final Path path = Paths.get(file.getAbsolutePath());
+
+			if (Files.exists(path) && (!Files.isRegularFile(path) || !Files.isWritable(path)))
```
abextm: Why do we convert to nio just for this?
```suggestion
			if (file.exists() && (!file.isFile() || !file.canWrite()))
```

Adam-:
What is the reason for this to be added to all of these tests?

runelite-client/src/main/java/net/runelite/client/RuneLite.java
deathbeam: Because I was trying file.isWrite and it wasnt there, lol. Updated

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -331,7 +335,7 @@ private void saveToFile(final File propertiesFile) throws IOException
 
 		parent.mkdirs();
 
-		File tempFile = new File(parent, SETTINGS_FILE_NAME + ".tmp");
+		File tempFile = new File(parent, RuneLite.DEFAULT_CONFIG_FILE + ".tmp");
```
Adam-: This is probably not less-correct than before, but I think the config writing code still has races since it does an open+truncate and then lock. So, I think it could truncate a file while another RL is holding a lock and writing to it.

PR #10798, opened 2020-02-14
Implement LinkBrowser.open() and change instances of Desktop.Open()

MrCarroll:
This should fix up most errors opening folders. Primarily this effects Linux where the Desktop.open() functionality isn't as robust as other platforms.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -28,11 +28,9 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Provides;
-import java.awt.Desktop;
 import java.awt.Graphics;
 import java.awt.Image;
 import java.awt.image.BufferedImage;
-import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.time.LocalDate;
 import java.util.concurrent.ScheduledExecutorService;
@@ -82,6 +80,7 @@ import net.runelite.client.util.HotkeyListener;
 import net.runelite.client.util.ImageCapture;
 import net.runelite.client.util.ImageUtil;
 import net.runelite.client.util.Text;
+import net.runelite.client.util.LinkBrowser;
 
 @PluginDescriptor(
 	name = "Screenshot",
@@ -188,14 +187,7 @@ public class ScreenshotPlugin extends Plugin
 				.<String, Runnable>builder()
 				.put("Open screenshot folder...", () ->
 				{
-					try
-					{
-						Desktop.getDesktop().open(SCREENSHOT_DIR);
-					}
-					catch (IOException ex)
-					{
-						log.warn("Error opening screenshot dir", ex);
-					}
+					LinkBrowser.open(SCREENSHOT_DIR.toString());
 				})
 				.build())
 			.build();
--- a/runelite-client/src/main/java/net/runelite/client/ui/FatalErrorDialog.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/FatalErrorDialog.java
@@ -28,7 +28,6 @@ import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Container;
-import java.awt.Desktop;
 import java.awt.Dimension;
 import java.awt.Font;
 import java.awt.event.WindowAdapter;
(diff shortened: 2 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -44,6 +45,7 @@
 public class LinkBrowser
 {
 	private static boolean shouldAttemptXdg = OSType.getOSType() == OSType.Linux;
+	private static boolean isSnapClient = ! Strings.isNullOrEmpty(System.getenv("SNAP"));
```
deathbeam: This is not specific to snap only so this check do not makes sense. If you are worried about blocking threads, then just send it to different thread

MrCarroll: After considering everything, from making the entire class threaded or seperating this class into two, and having looked at the actual use of the existing and new methods, it does appear the best thing to do is just throw the method in an anonymous thread if the threading is a concern. There isn't enough usage of this to begin with to overcomplicate it.

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -74,6 +75,36 @@ public static boolean browse(final String url)
 		return false;
 	}
 
+	/**
+	 * Tries to open a directory in the OS native file manager.
+	 * @param directory directory to open
+	 * @return @return true if operation was successful
+	 * THIS METHOD MAY BLOCK FOR A SUBSTANTIAL AMOUNT OF TIME.
```
abextm: I would rather this just handle threading on it's own and not return anything; the return value is useless.

runelite-client/src/main/java/net/runelite/client/ui/FatalErrorDialog.java
```diff
@@ -141,14 +140,13 @@ public void windowClosing(WindowEvent e)
 
 		addButton("Open logs folder", () ->
 		{
-			try
+			new Thread()
```
abextm: prefer passing a lambda to subclassing Thread

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -48,37 +49,67 @@
...
+			}
 
-		showMessageBox("Unable to open link. Press 'OK' and link will be copied to your clipboard.", url);
-		return false;
+			if (shouldAttemptXdg && attemptXdgOpen(directory))
+			{
+				log.debug("Opened directory through xdg-open to {}", directory);
+				return;
+			}
+			log.warn("LinkBrowser.open() could not open {}", directory);
```
deathbeam: Maybe this should show messagebox as well to let the user know, and maybe copy target path to clipboard too

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -48,37 +49,68 @@
...
+				log.debug("Opened directory through Desktop#browse to {}", url);
+				return;
+			}
+
+			if (shouldAttemptXdg && attemptXdgOpen(url))
+			{
+				log.debug("Opened directory through xdg-open to {}", url);
+				return;
+			}
+			log.warn("LinkBrowser.browse() could not open {}", url);
```
Nightfirecat: Add newline above

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
```diff
@@ -48,37 +49,68 @@
 	/**
```
Nightfirecat: Change class javadoc to read `Utility class used for web and file browser navigation`

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -82,6 +80,8 @@
 import net.runelite.client.util.ImageCapture;
 import net.runelite.client.util.ImageUtil;
 import net.runelite.client.util.Text;
+import net.runelite.client.util.LinkBrowser;
+
```
Nightfirecat: Remove this extra newline

runelite-client/src/main/java/net/runelite/client/util/LinkBrowser.java
Nightfirecat: Add newline above

MrCarroll:
I've made the changes requested. Thanks for the feedback about it being on the master branch making other peoples PR's impossible, I didn't consider that limitation and next time I'll avoid it.

Issue #10714, opened 2020-02-02
XP globe infoboxes cut off


thebinaryeagle:
If you have your xp globes towards the side of the window, the infoboxes get cut off rather than moving to the side to stay within it.

![RuneLite_NsTM0LPUbL](https://user-images.githubusercontent.com/60583962/73615353-1c196b80-45d5-11ea-8c6a-a9fded29980d.png)

JDushane:
I've got it working so that the infobox is moved in case it gets cut off on the left side. I don't have it working for the right, but I don't see that being a problem as pretty much all of the regular osrs UI is crammed over there anyways.

thebinaryeagle:
It’s better than nothing but I did test putting them under the minimap at one point and I could see that being a comfortable position for some people to place them in. Ideally, it would be nice to have that freedom since it’s still technically a bug that it would get cut off.

mrboots123:
#10982

PR #10804, opened 2020-02-15
Got it working so that the tooltip doesnt get cut off left side

JDushane:
I've got it working so that the infobox is moved in case it gets cut off on the left side. I don't have it working for the right, but I don't see that being a problem as pretty much all of the regular osrs UI is crammed over there anyways.


JDushane:
Forgot to mention explicitly that this was to address issue #10714

Nightfirecat:
For what it's worth, I don't feel a solution which fixes this bug on only one side will be merged. I recommend looking into solving it for overflowing on the right side of the game frame as well.

PR #10815, opened 2020-02-16
Fixed slayer kill counter and added 1 click teleport for Gem mine

ThendBend:
slayer kill counter was bugged when you kill 2 slayer monsters at the same time in multi combat. For example when you barrage Nechs.


Nightfirecat:
1. Features/bugfixes should be PRed separately, one per functional change, not combined into a single PR
2. Your code change for fixing barrages is not sufficient. To do this "correctly", some type of knapsack algorithm will need to be implemented. See discussion on #1046 as to why your solution is not sufficient.
3. We are not adding left-click defaults for teleport items which have non-"Teleport"/non-"Rub" teleport options. Please use the shift-click configuration to set the default shift option to your desired menu entry.

PR #10822, opened 2020-02-18
fix so that head api is the correct value

botanicvelious:


abextm:
added in cb52f0fa5f6e9cca816f25b971047ac0d99a8d1d

PR #10830, opened 2020-02-18
config: fix entering values directly into options with units not working

Hydrox6:
Currently, trying to enter a value directly into an option with a unit will fail unless the user also inputs the unit.
I was trying to avoid writing a whole AbstractFormatter initially, since it seemed overkill. Evidently, it was the only way to do it properly.

Putting both the Formatter and the Factory in the same file is probably bad. Please let me know how you want it doing

Thanks to Trevor for notifying me of this issue.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2020, Hydrox6 <ikada@protonmail.ch>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
```diff
@@ -0,0 +1,94 @@
...
+			return Integer.valueOf(trimmedText);
+		}
+		catch (NumberFormatException e)
+		{
+			throw new ParseException(trimmedText + " is not an integer.", 0);
+		}
+	}
+
+	@Override
+	public String valueToString(final Object value) throws ParseException
```
Nightfirecat: This doesn't throw any exceptions.
```suggestion
	public String valueToString(final Object value)
```

runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
```diff
@@ -0,0 +1,94 @@
...
+
+	UnitFormatter(String units)
+	{
+		this.units = units;
+	}
+
+	@Override
+	public Object stringToValue(final String text) throws ParseException
+	{
+		final String trimmedText;
```
Nightfirecat: Newline below this
```suggestion
		final String trimmedText;

```

runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
```diff
@@ -0,0 +1,94 @@
...
+		// Using the spinner controls causes the value to have the unit on the end, so remove it
+		if (text.substring(text.length() - units.length()).equals(units))
```
Nightfirecat: This will throw a `StringIndexOutOfBoundsException` if you enter a string which is just a number for a field which expects to have units added. Why doesn't this simply check `text.endsWith(units)`?

runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
```diff
@@ -0,0 +1,94 @@
...
+		formatters = new HashMap<>();
+		this.units = units;
+	}
+
+	@Override
+	public JFormattedTextField.AbstractFormatter getFormatter(final JFormattedTextField tf)
+	{
+		if (!formatters.containsKey(tf))
+		{
+			formatters.put(tf, new UnitFormatter(units.value()));
```
Nightfirecat: Why pass `units.value()` rather than `units` itself? I'd rather this factory handle as little as possible and pass most of its logic to the formatter class.

runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
Hydrox6: Honestly didn't know that existed. Wow that makes it a lot cleaner

runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
```diff
@@ -0,0 +1,95 @@
...
+		formatters = new HashMap<>();
+		this.units = units;
+	}
+
+	@Override
+	public JFormattedTextField.AbstractFormatter getFormatter(final JFormattedTextField tf)
+	{
+		if (!formatters.containsKey(tf))
+		{
+			formatters.put(tf, new UnitFormatter(units));
```
Adam-: Whats with all this logic if there is only ever one formatter factory per text field?

PR #10832, opened 2020-02-19
worldmap: add ore information to mining site tooltips

dekvall:
Adds an option to show the available ores of a mining site in its world map tooltip.
mining_tooltips


Nightfirecat:
Shouldn't the `worldmap: support multiline tooltips` commit come first? (maybe GitHub's UI is showing them out of order?)

dekvall:
I think that it's a GitHub UI issue, it looks correct on my end.

PR #10850, opened 2020-02-22
mining: add urt, efh, te and basalt respawn timers

### LINKED ISSUE #10823: Mining: Show timers on Weiss basalt
When mining other things with the mining plugin enabled, you get a nice timer showing you when the ore will become mineable again. It would be nice if this was enabled for the basalt rocks under Weiss too.

melkypie:
fixes #10823

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
@@ -30,20 +30,21 @@ import java.util.Map;
 import lombok.AccessLevel;
 import lombok.Getter;
 import static net.runelite.api.ObjectID.*;
+import static net.runelite.client.util.RSTimeUnit.*;
 
 enum Rock
 {
-	TIN(Duration.ofMillis(2400), 0, ROCKS_11360, ROCKS_11361),
-	COPPER(Duration.ofMillis(2400), 0, ROCKS_10943, ROCKS_11161),
-	IRON(Duration.ofMillis(5400), 0, ROCKS_11364, ROCKS_11365, ROCKS_36203)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mining/Rock.java
```diff
@@ -90,7 +90,11 @@ Duration getRespawnTime(int region)
 	ORE_VEIN(Duration.ofSeconds(MiningOverlay.ORE_VEIN_MAX_RESPAWN_TIME), 150),
 	AMETHYST(Duration.ofSeconds(75), 120),
 	ASH_VEIN(Duration.ofSeconds(30), 0, ASH_PILE),
-	GEM_ROCK(Duration.ofMinutes(1), 0, ROCKS_11380, ROCKS_11381);
+	GEM_ROCK(Duration.ofMinutes(1), 0, ROCKS_11380, ROCKS_11381),
+	URT_SALT(Duration.ofMillis(5400), 0, ROCKS_33254),
+	EFH_SALT(Duration.ofMillis(5400), 0, ROCKS_33255),
+	TE_SALT(Duration.ofMillis(5400), 0, ROCKS_33256),
+	BASALT(Duration.ofMillis(5400), 0, ROCKS_33257);
```
Nightfirecat: Since #10773 has been merged, these can be changed to use `Duration.of(9, GAME_TICKS)` so long as `RSTimeUnit.GAME_TICKS` is imported.

melkypie: Changed, let me know if it is also better to update the rest of the enum to match with the GAME_TICKS also.

melkypie:
Changed every duration value to use `GAME_TICKS` in the enum in a separate commit before the actual commit as per dekvall's suggestion

PR #10863, opened 2020-02-23
inventorytags: add the ability to have infinite color groups

melkypie:
Adds the ability to not be restricted to groups for colors, making it possible to have infinite amount of color groups
runelite_inventory_tags


dekvall:
So I think that a list of currently active custom groups would need to be available somewhere for this to be useful. Say you mark items A, B, C with color `#aabbcc` and then mark items D, E, F with color `#ddeeff`, then you remember that you also need to add item G to the first group. Without a list of custom groups, or a way to extract the color from an existing tag, you will most likely not be able to remember that those items have the color `#aabbcc` and subsequently not be able to add G to the group.

Also, I personally believe that 6 groups is more than enough for any use case, but i understand that there are a number of ways to play this game.

PR #10868, opened 2020-02-24
GroundItems: Keep item lists in input order

loldudester:
Highlighted and Hidden item lists currently become unordered when the lists are updated. This changes HashSets to ArrayLists to avoid this.

Duplicates in the lists will no longer be removed but since they can only happen by manually adding something to the list twice, it's not a huge issue.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -36,12 +36,10 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
-import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


Adam-:
It already checks for duplicates so it is not necessary to use a set.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -548,8 +546,8 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 
 	void updateList(String item, boolean hiddenList)
 	{
-		final Set<String> hiddenItemSet = new HashSet<>(hiddenItemList);
-		final Set<String> highlightedItemSet = new HashSet<>(highlightedItemsList);
+		final ArrayList<String> hiddenItemSet = new ArrayList<>(hiddenItemList);
+		final ArrayList<String> highlightedItemSet = new ArrayList<>(highlightedItemsList);
```
Nightfirecat:
```suggestion
		final List<String> hiddenItemSet = new ArrayList<>(hiddenItemList);
		final List<String> highlightedItemSet = new ArrayList<>(highlightedItemsList);
```

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -560,7 +558,7 @@ void updateList(String item, boolean hiddenList)
 			hiddenItemSet.removeIf(item::equalsIgnoreCase);
 		}
 
-		final Set<String> items = hiddenList ? hiddenItemSet : highlightedItemSet;
+		final ArrayList<String> items = hiddenList ? hiddenItemSet : highlightedItemSet;
```
Nightfirecat:
```suggestion
		final List<String> items = hiddenList ? hiddenItemSet : highlightedItemSet;
```

PR #10869, opened 2020-02-24
Prayer: Add prayer threshold notification.

Sween-anator:
Add feature to send notification when prayer points drop below a set threshold. Reworked code from Nightmare Zone absorption potion notification.


Alexsuperfly:
idle notifier already does this

PR #10873, opened 2020-02-24
Make Highlight value take priority over other value thresholds

### LINKED ISSUE #10840: Ground Item value threshold (Low/medium/high/insane) overrides highlight > value threshold
When using the ground items plugin, the thresholds for the item values for low/medium/high/insane seem to always take priority over the threshold for highlight > value. That is, items that meet one of the four value thresholds will not be highlighted (or trigger a notification enabled), but instead only colored based on value tier.

For example, if the highlight > value threshold is set to 5,000 and the low item value threshold is set to 20,000, and I drop an item of value 6,000, it will highlight and notify me. However, if I drop an item of value 21,000 or greater, it will color based on item value alone and will _not_ notify me. The following screenshots demonstrate this.

Ideally, at least in my opinion, it would make more sense for the highlight > value threshold to override the item value thresholds, especially considering the highlight is the only way to receive a notification on loot drop.

zacharydwaller:
Closes #10840

Previously the Low/Medium/High/Insane threshold highlighting would take priority over the Highlight threshold highlighting. This simply puts Highlight at the top of the list, giving it priority.

Example:
My High value is 250k and Insane value is 1000k: An Abyssal Whip (2400k) it'll be highlighted as Insane, a Blessed D'hide Body (600k) will be highlighted as High.
If I set my Highlight threshold to 500k, the Blessed D'hide Body will now be highlighted with the Highlight color.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -433,6 +433,11 @@ public class GroundItemsPlugin extends Plugin
 		// Cache colors
 		priceChecks.clear();
 
+		if (config.getHighlightOverValue() > 0)
+		{
+			priceChecks.put(config.getHighlightOverValue(), config.highlightedColor());
+		}
+
 		if (config.insaneValuePrice() > 0)
 		{
 			priceChecks.put(config.insaneValuePrice(), config.insaneValueColor());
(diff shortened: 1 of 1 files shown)


raiyni:
I think the user in #10840 is moreso talking about the notification not firing.

If I want to be notified of anything greater than 5k, this will change everything to that highlight color now.

Adam-:
> 
> 
> I think the user in #10840 is moreso talking about the notification not firing.
> 
> If I want to be notified of anything greater than 5k, this will change everything to that highlight color now.

On release if you want to be notified of anything >5k, you have to set notify value at 5k and all of the thresholds <5k, which would be no different since either with this or without this, you would have the items all highlighted.

PR #10876, opened 2020-02-24
Ignore price thresholds when checking for highlight notify

### LINKED ISSUE #10840: Ground Item value threshold (Low/medium/high/insane) overrides highlight > value threshold

deathbeam:
Closes #10840
Supersedes/Closes #10873


deathbeam:
Resolved in #10873

PR #10877, opened 2020-02-24
Added GPU plugin option to hide the UI

### LINKED ISSUE #10803: Hide all UI
Yes, You can already hide the minimap.
But I'd like a plugin that would hide ALL UI elements of the stretched display mode, preferably with a keyboard shortcut.
Would be great for taking screenshots and stuff.
begone, ui
(at some point I'd like to get rid of the top Runelite bar as well, but hey one step at a time)

zacharydwaller:
Closes #10803

Added a feature to hide the UI, which is very nice for taking screenshots or recording video footage. There's currently ways to hide the minimap and minimize UI panels, but not any way that I know of to hide the entire UI.

This updates the GPU plugin and adds an option to Hide UI. The user suggested that a keyboard shortcut - I'll wait for feedback too see whether or not that should be implemented. World of Warcraft uses Alt-Z, this shortcut isn't used elsewhere in Runelite but Alt isn't used for many non-development shortcuts if I'm not mistaken.

Config Option successfully hides UI in-game:

Also, it doesn't hide the UI if the user isn't logged in or the Message of the Day screen is present - so the user can actually log into the game if the option is enabled.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -67,6 +67,7 @@ import net.runelite.api.Texture;
 import net.runelite.api.TextureProvider;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.hooks.DrawCallbacks;
+import net.runelite.api.widgets.WidgetInfo;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
This doesn't close #10803 as it only works with the GPU plugin enabled. I'd rather see changes made to hide UI regardless of whether the GPU plugin is used.

Issue #10731, opened 2020-02-05
Camera only north option is available


Locomojitoo:
Please check if your issue is not a duplicate by [searching existing issues](https://github.com/runelite/runelite/search?type=Issues)

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. Windows, Ubuntu, macOS]
 - RuneLite version: [e.g 1.4.6]
 - Launcher version: [e.g 1.5.2]

**Additional context**
Add any other context about the problem here, e.g. logs. Your client logs can usually be found in
`C:\Users\<your_user_name>\.runelite\logs` on Windows and `~/.runelite/logs` on Linux and macOS.

Locomojitoo:
It should be possible to look west east south when you click on compass but it only gives north as option even when the camera option is turned on

ThePharros:
it should work if you right click on the compass

Locomojitoo:
Ahh shit you are right. Is it by chance possible to make it possible to
change angles with a simple click? This would be great if you want to play
on windows touchscreen devices

Op wo 5 feb. 2020 om 05:06 schreef ThePharros <notifications@github.com>

hjdarnel:
What if the Camera plugin was extended so clicking the compass quickly in succession (within a configurable reset time) would cycle between N S E and W, so players could double tap compass to look south, and after timeout tapping compass would again return to north?

I have this partially implemented. Could also do a simple config option dropdown for default compass direction, but I find the multiple tapping would be more useful. Touch players could make the timeout as long as they need.

PR #10878, opened 2020-02-25
Add Camera Cycle to camera plugin

hjdarnel:
The idea with this functionality is that multiple repeated clicks would cycle the camera through the cardinal directions. Two quick clicks would reset to Look South, three to Look East, etc. The timeout that determines how quickly these clicks need to be is configurable (mouse vs tablet users could enjoy different values)

This is off by default. It shouldn't conflict with the Compass options `compassLook` functionality, but just extend it. The order will continue to be `North, South, East, West` however if East is the current top menu entry the menu will show `East, West, North, South` to prevent duplicate menu entries.

Closes https://github.com/runelite/runelite/issues/10731 particularly this comment https://github.com/runelite/runelite/issues/10731#issuecomment-582265261

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraConfig.java
@@ -28,6 +28,7 @@ import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 import net.runelite.client.config.Range;
+import net.runelite.client.config.Units;
 
 @ConfigGroup("zoom") // using the old plugin's group name
 public interface CameraConfig extends Config
@@ -201,5 +202,28 @@ public interface CameraConfig extends Config
 	default boolean preserveYaw()
 	{
 		return false;
+	}
+
+	@ConfigItem(
+		keyName = "compassMultipleClickCycle",
+		name = "Multiple clicks on compass ",
+		description = "Cycle between Look North, South, East, and West by clicking on compass in rapid succession",
+		position = 15
+	)
+	default boolean compassMultipleClickCycle()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "compassMultipleClickTimeout",
+		name = "Multiple click timeout",
+		description = "Configure how quickly the compass resets to \"Look North\" after the last click",
+		position = 16
+	)
+	@Units(Units.SECONDS)
+	default int compassMultipleClickTimeout()
+	{
+		return 2;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
@@ -34,6 +34,7 @@ import java.util.Arrays;
 import javax.inject.Inject;
 import javax.swing.SwingUtilities;
 import net.runelite.api.Client;
+import net.runelite.api.Constants;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
 import net.runelite.api.ScriptID;
@@ -44,7 +45,9 @@ import net.runelite.api.events.BeforeRender;
 import net.runelite.api.events.ClientTick;
 import net.runelite.api.events.FocusChanged;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.api.events.MenuEntryAdded;
+import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.ScriptCallbackEvent;
 import net.runelite.api.events.ScriptPreFired;
 import net.runelite.api.events.WidgetLoaded;
@@ -64,6 +67,7 @@ import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.tooltip.Tooltip;
 import net.runelite.client.ui.overlay.tooltip.TooltipManager;
+import org.apache.commons.lang3.ArrayUtils;
 
 @PluginDescriptor(
 	name = "Camera",
@@ -81,6 +85,11 @@ public class CameraPlugin extends Plugin implements KeyListener, MouseListener
 	private static final String LOOK_SOUTH = "Look South";
 	private static final String LOOK_EAST = "Look East";
 	private static final String LOOK_WEST = "Look West";
+
+	private static final String DEFAULT_LEFT_CLICK_OPTION = LOOK_NORTH;
+	private int lastUpdatedTick = 0;
+	private String currentLeftClickOption;
+	private String[] options = new String[]{LOOK_NORTH, LOOK_SOUTH, LOOK_EAST, LOOK_WEST};
 
 	private boolean controlDown;
 	// flags used to store the mousedown states
@@ -124,6 +133,7 @@ public class CameraPlugin extends Plugin implements KeyListener, MouseListener
 		rightClick = false;
 		middleClick = false;
 		menuHasEntries = false;
+		currentLeftClickOption = DEFAULT_LEFT_CLICK_OPTION;
 		copyConfigs();
 		keyManager.registerKeyListener(this);
 		mouseManager.registerMouseListener(this);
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +127,83 @@ protected void shutDown()
...
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 3);
+
+				// fill the menu from top to bottom
+				// but iterate through options cyclically from 0 -> 3 -> 0
+				int entryIndex = menuEntries.length - 1;
+
+				// The handling for these entries is done in ToplevelCompassOp.rs2asm`
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+				entryIndex -= 1;
+				currentOptionIndex += 1;
```
hjdarnel: Lines 164-165 could be removed with something like `menuEntries[entryIndex--]` and `currentOptionIndex++` but I found it hard to follow when written that way

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraConfig.java
```diff
@@ -158,4 +159,27 @@ default boolean compassLook()
...
+	}
+
+	@ConfigItem(
+		keyName = "compassMultipleClickTimeout",
+		name = "Multiple click timeout",
+		description = "Configure how quickly the compass resets to \"Look North\" after the last click",
+		position = 12
+	)
+	@Units(Units.SECONDS)
+	default int compassMultipleClickTimeout()
```
Nightfirecat: This currently isn't used. There's a `RESET_DIRECTION_TIMEOUT` constant which should have its usages replaced by this.

hjdarnel: My bad, you're right. Fixed.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -66,6 +71,11 @@
 	private static final String LOOK_SOUTH = "Look South";
 	private static final String LOOK_EAST = "Look East";
 	private static final String LOOK_WEST = "Look West";
+	private static final int RESET_DIRECTION_TIMEOUT = 5;
+	private int lastUpdatedTick = 0;
+	private static final String defaultLeftClickOption = LOOK_NORTH;
```
Nightfirecat: This should be moved with the other constants and renamed to use the same naming scheme.
```suggestion
	private static final String DEFAULT_LEFT_CLICK_OPTION = LOOK_NORTH;
```

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -66,6 +71,11 @@
...
+	private String currentLeftClickOption;
+	private String[] options = new String[]{LOOK_NORTH, LOOK_SOUTH, LOOK_EAST, LOOK_WEST};
```
Nightfirecat: This can be made `static final` and moved with the other constants.
```suggestion
	private static final String[] OPTIONS = new String[]{LOOK_NORTH, LOOK_SOUTH, LOOK_EAST, LOOK_WEST};
```

Nightfirecat: Also NSEW is a strange order, I would expect to see this in NESW order.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +128,84 @@ protected void shutDown()
...
+			int currentOptionIndex = ArrayUtils.indexOf(options, currentLeftClickOption);
 
-			// The handling for these entries is done in ToplevelCompassOp.rs2asm
-			menuEntries[--len] = createCameraLookEntry(menuEntryAdded, 4, LOOK_WEST);
-			menuEntries[++len] = createCameraLookEntry(menuEntryAdded, 3, LOOK_EAST);
-			menuEntries[++len] = createCameraLookEntry(menuEntryAdded, 2, LOOK_SOUTH);
-			menuEntries[++len] = north;
+			if (config.compassLook())
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 3);
```
Nightfirecat: `3` should instead be a variable or reference to a variable, such as `OPTIONS.length - 1` since it makes no sense to be a magic number.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +128,84 @@ protected void shutDown()
...
+				currentOptionIndex += 1;
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+				entryIndex -= 1;
+				currentOptionIndex += 1;
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+			}
+			else
+			{
+				// set top entry to the current left click
+				menuEntries[menuEntries.length - 1] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
```
Nightfirecat: I don't understand... Why is `onMenuEntryAdded` running any code when `config.compassLook()` is false? We shouldn't be adding or changing any menu entries when that's the case.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +128,84 @@ protected void shutDown()
...
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+				entryIndex -= 1;
+				currentOptionIndex += 1;
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+				entryIndex -= 1;
+				currentOptionIndex += 1;
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
+				entryIndex -= 1;
+				currentOptionIndex += 1;
+				menuEntries[entryIndex] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + 1, options[currentOptionIndex % count]);
```
Nightfirecat: This code is completely duplicated and should be replaced by a loop, such as:

```java
for (int i = 0; i < count; i++)
{
	menuEntries[menuEntries.length - i] = createCameraLookEntry(menuEntryAdded, currentOptionIndex + i, OPTIONS[currentOptionIndex % count]);
}
```

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +128,84 @@ protected void shutDown()
...
 			client.setMenuEntries(menuEntries);
 		}
 	}
 
-	private MenuEntry createCameraLookEntry(MenuEntryAdded lookNorth, int identifier, String option)
+	@Subscribe
+	private void onMenuOptionClicked(MenuOptionClicked event)
+	{
+		// if the option is disabled, don't process the click any differently
+		if ((event.getMenuAction() == MenuAction.CC_OP) && config.compassMultipleClickCycle() && event.getMenuOption().equals(currentLeftClickOption))
```
Nightfirecat:
```suggestion
		if (event.getMenuAction() == MenuAction.CC_OP && config.compassMultipleClickCycle() && event.getMenuOption().equals(currentLeftClickOption))
```

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -117,36 +128,84 @@ protected void shutDown()
...
 
 			client.setMenuEntries(menuEntries);
 		}
 	}
 
-	private MenuEntry createCameraLookEntry(MenuEntryAdded lookNorth, int identifier, String option)
+	@Subscribe
+	private void onMenuOptionClicked(MenuOptionClicked event)
+	{
+		// if the option is disabled, don't process the click any differently
```
Nightfirecat: Do we have a way to tell whether this menu option was clicked via left-clicking the map or clicking a right-click menu? (regardless of if it was the top menu option) I would think that the cycling behavior would not be wanted if clicking a right-click menu option.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
hjdarnel: That's the order that the current Camera Options plugin shows options, so I didn't want to do different than that. I can change this to NESW though.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
hjdarnel: if `config.compassLook() == true` then we need to build out the menu options for that plugin, and rearrange them depending on the current left click option. If `config.compassLook() == false` then we still need to modify the top option to Look South, East, West, etc depending on the current left click value. 

If `config.compassMultipleClickTimeout() == false` (which is the config option I'm adding) then we just don't do anything special on a compass click, instead `currentLeftClickOption` will always be the default "Look North"

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
Nightfirecat: Hm, that's fine then. Let's leave it be.

hjdarnel:
Should this just be an external plugin?

Nightfirecat:
Based on discussion in Discord, since the "Look East/South/West" feature already exists in the main client, this seems reasonable to go into the main client as well.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
hjdarnel: this is covered by line 233 `event.getMenuOption().equals(currentLeftClickOption)` I think.

hjdarnel:
After a long hiatus... this PR can be reviewed again

Nightfirecat:
Needs rebase now that Look NESW is now a vanilla feature and it was removed from the plugin.

hjdarnel:
I'm having trouble rebasing this. I don't really understand how the menu entry class works. If someone else wanted to pick this up and put their own PR up they can feel free.

PR #10879, opened 2020-02-25
Ground Items: Add notification by value config option

TheStonedTurtle:
Pretty much does the title

image

This also fixes a bug in the notification where it should have displayed `Lots` but was instead display 65535

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -232,14 +232,19 @@ public class GroundItemsPlugin extends Plugin
 			// The spawn time remains set at the oldest spawn
 		}
 
-		boolean shouldNotify = !config.onlyShowLoot() && config.highlightedColor().equals(getHighlighted(
+		final Color highlightedColor = getHighlighted(
 			groundItem.getName(),
 			groundItem.getGePrice(),
-			groundItem.getHaPrice()));
+			groundItem.getHaPrice());
+		boolean shouldNotify = !config.onlyShowLoot() && config.highlightedColor().equals(highlightedColor);
 
 		if (config.notifyHighlightedDrops() && shouldNotify)
 		{
 			notifyHighlightedItem(groundItem);
+		}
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -234,15 +234,20 @@ public void onItemSpawned(ItemSpawned itemSpawned)
...
 			groundItem.getGePrice(),
-			groundItem.getHaPrice()));
+			groundItem.getHaPrice());
+		boolean shouldNotify = !config.onlyShowLoot() && config.highlightedColor().equals(highlightedColor);
 
 		if (config.notifyHighlightedDrops() && shouldNotify)
 		{
 			notifyHighlightedItem(groundItem);
 		}
+		else if (highlightedColor != null && (!config.onlyShowLoot() || groundItem.isMine()) && getGroundItemConfigValue(groundItem) > config.getNotifyOverValue())
```
Nightfirecat: Why does this require `highlightedColor != null`?

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -651,21 +670,15 @@ public void onFocusChanged(FocusChanged focusChanged)
...
-			.append(item.getName());
+		final StringBuilder notificationStringBuilder = new StringBuilder(item.getName());
 
 		if (item.getQuantity() > 1)
 		{
 			notificationStringBuilder.append(" x ").append(item.getQuantity());
 
-
-			if (item.getQuantity() > (int) Character.MAX_VALUE)
+			if (item.getQuantity() >= (int) Character.MAX_VALUE)
```
Nightfirecat: This should be in a separate commit.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
TheStonedTurtle: I added this to ensure that items that are explicitly hidden would not trigger valuable notifications, but now I see this would prevent notifications from working if their price checkpoints are lower than their notification threshold.

Should I still prevent items that are explicitly hidden from triggering this notification?

PR #10881, opened 2020-02-25
Tooltip: Fix for off screen and under cursor rendering

### LINKED ISSUE #10791: Tooltips are rendered in an incorrect location with tooltip position Above Cursor
Tooltips are rendered incorrectly when `Tooltip Position` is set to `Above Cursor` resulting in the tooltip being rendered outside of the client bounds or partially under the mouse cursor:

image
image

DSerp:
Closes #10791 

Without a way to get the cursor dimensions, I just used an offset similar to the way it is used for the y-axis.


Nightfirecat:
Why is this modifying the x offset? The bug has nothing to do with horizontal positioning.

DSerp:
The bug has to be moved horizontally to not be under the cursor

Nightfirecat:
I see, that code is for addressing the second part of the linked bug. (tooltips can be rendered under the cursor when tiling downwards)

1. Cursors are generally 32x32 px on Windows and similar sizes on other OSes, so a constant for that size should be created and referenced for this translation
2. Please split this into two commits--one which fixes *only* tooltips going beyond the canvas edges, and the other which *only* translates tooltips to avoid rendering under the cursor

DSerp:
The offset is now at 32 and I split the fixes into two different commits

image

runelite-client/src/main/java/net/runelite/client/ui/overlay/tooltip/TooltipOverlay.java
```diff
@@ -122,6 +130,11 @@ private Dimension renderTooltips(Graphics2D graphics, List<Tooltip> tooltips)
 				mousePosition.move(mouseCanvasPosition.getX(), mouseCanvasPosition.getY() + offset + newBounds.height);
 			}
 
+			if (tooltips.size() > 1 && this.runeLiteConfig.tooltipPosition() == TooltipPositionType.ABOVE_CURSOR)
+			{
+				mousePosition.translate(CURSOR_OFFSET, 0);
```
Nightfirecat: After testing this ingame, I can say confidently this is not the right behavior. We should not be translating tooltips to the right of the cursor when there are multiple present--they should instead tile upwards like so:

```
--------------
|  tooltip 1 |
--------------
---------------------
| multiline tooltip |
| another line      |
| foobar            |
---------------------
|\
|  \
| ___\
|/ \\
    \\
```

DSerp: Alright that sounds better to me. I appreciate the art you made as well 😅

DSerp: image

runelite-client/src/main/java/net/runelite/client/ui/overlay/tooltip/TooltipOverlay.java
```diff
@@ -122,6 +129,11 @@ private Dimension renderTooltips(Graphics2D graphics, List<Tooltip> tooltips)
 				mousePosition.move(mouseCanvasPosition.getX(), mouseCanvasPosition.getY() + offset + newBounds.height);
 			}
 
+			if (tooltips.size() > 1 && this.runeLiteConfig.tooltipPosition() == TooltipPositionType.ABOVE_CURSOR)
+			{
+				mousePosition.translate(0, (int) -bounds.getHeight() - offset);
```
Adam-: So this is moving the tooltips back down if they would extend over the top? We don't do this for the normal tooltip layout do we?

DSerp: That is exactly what happens when the tooltip extends under the bottom of the screen, so I replicated it for the top. Although if the tooltip were to extend beyond the right side of the screen it would just stop at the edge.

DSerp: So the original code had different functionalities when going out of the screen if it was either horizontally or vertically

Adam-:
fixed in 080f3133c5c94db601159477741bdc6fe8502d80

PR #10893, opened 2020-02-26
Fix Shift Anti Drag default bank drag delay

### LINKED ISSUE #10891: Shift Anti Drag Plugin sets bank drag delay to 0, not default value on focus change or shift release.
**Describe the bug**
A focus change or release of shift causes the Shift Anti Drag plugin to set the bank drag delay to 0ms, which is not the default value, resulting in players being unable to withdraw items from the bank similar to before. 

**To Reproduce**
Change focus or release shift while bank is open.

**Expected behavior**
Delay is not set to 0, rather to the default runescape drag delay.

**Screenshots**
https://www.youtube.com/watch?v=RyzSckNSStc
https://streamable.com/6f78m

Issue is pretty clearly found in AntiDragPlugin.java
`@Override
	public void keyReleased(KeyEvent e)
	{
		if (e.getKeyCode() == KeyEvent.VK_SHIFT)
		{
			client.setInventoryDragDelay(DEFAULT_DELAY);
			// In this case, 0 is the default for bank item widgets.
			setBankDragDelay(0);
		}
	}

	@Subscribe
	public void onFocusChanged(FocusChanged focusChanged)
	{
		if (!focusChanged.isFocused())
		{
			client.setInventoryDragDelay(DEFAULT_DELAY);
			setBankDragDelay(0);
		}
	}`

brennanwill:
Fixes #10891
Changes default bank drag delay from 0 to 5, in line with official client behavior.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
@@ -98,8 +98,7 @@ public class AntiDragPlugin extends Plugin implements KeyListener
 		if (e.getKeyCode() == KeyEvent.VK_SHIFT)
 		{
 			client.setInventoryDragDelay(DEFAULT_DELAY);
-			// In this case, 0 is the default for bank item widgets.
-			setBankDragDelay(0);
+			setBankDragDelay(DEFAULT_DELAY);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Adam-:
I find it odd this has a comment specifically saying 0 is right. @JBerben where did this `0` value come from?

Adam-:
Regardless this does seem obviously wrong because if you look at the inspector it is definitely 5.

PR #10894, opened 2020-02-27
Plugin Hub: Reorder plugin panel and rework the search function

loldudester:
1) Changed the panel to sort by display name, not internal name

2) The existing search logic gives basically useless results.

This makes it more in line with how the main plugin panel's search works.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
@@ -37,7 +37,9 @@ import java.awt.event.ActionEvent;
 import java.awt.event.KeyEvent;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
```diff
@@ -140,6 +137,18 @@
 			manifest = newManifest == null ? loaded : newManifest;
 			this.installed = installed;
 
+			if (manifest != null)
+			{
+				Collections.addAll(keywords, SPACES.split(manifest.getDisplayName().toLowerCase()));
+				Collections.addAll(keywords, SPACES.split(manifest.getDescription().toLowerCase()));
```
abextm: this can throw npe, and I'm not sure its even a good idea

deathbeam: Updated (and marked description/tags as nullable). And at least for now imo it is fine (as long as external plugin descriptions make sense, and atm most of them do).

PR #10895, opened 2020-02-27
Skill Calculator: Fetch only the XP when opening the same skill

### LINKED ISSUE #10445: Skill Calculator: Do not reset text inputs when refreshing already selected skill
When figuring out how many actions you have left until next XP goal, you need to retype all fields if you were to refresh the current skill you are on the entire calculator resets. 

Have XP/Level Goal and typed in boxes for actions to be done not reset when refreshing box. 

**Additional context**
This should only happen if you're refreshing the current stat and not when you do a skill calc on another skill and return. 
Example: I calculate the number of battle staffs until 84 crafting when I am at 75 crafting. I craft a bit and I want to know the new total, I should be able to click 'Crafting' and get an updated count of what ever I was previously interested in with the goal of 84 crafting still.

DSerp:
Closes #10445 

When clicking the same selected skill it will now only fetch the current XP

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -128,7 +128,7 @@ class SkillCalculator extends JPanel
 		currentXP = client.getSkillExperience(calculatorType.getSkill());
 		currentLevel = Experience.getLevelForXp(currentXP);
 
-		if (currentSkill == null || !currentSkill.equals(calculatorType.getSkill()))
+		if (currentSkill != calculatorType.getSkill())
 		{
 			currentSkill = calculatorType.getSkill();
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -123,48 +124,54 @@
...
-		VarPlayer endGoalVarp = endGoalVarpForSkill(calculatorType.getSkill());
-		int endGoal = client.getVar(endGoalVarp);
-		if (endGoal != -1)
-		{
-			targetLevel = Experience.getLevelForXp(endGoal);
-			targetXP = endGoal;
-		}
-		else
+
+		if (currentSkill == null || !currentSkill.equals(calculatorType.getSkill()))
```
Adam-:
```suggestion
		if (currentSkill != calculatorType.getSkill())
```

PR #10908, opened 2020-02-28
Check player animation to update fishing status

### LINKED ISSUE #10900: Fishing Tooltip does not update on level-up
**Bug**
The fishing tooltip (Status, Caught, Per hour) does not update the status when character levels up. and pauses their action.

**Expected behavior**
When leveling up, fishing becomes interrupted. The tooltip should update status to "NOT Fishing" until the level-up screen is cleared, since the screen is not interrupted and must be cleared by the user.

Screenshots (.gif):
https://i.gyazo.com/afb59824d2dc18dc09223d327e74dac7.gif

**Environment:**
 - OS: Windows 10
 - RuneLite version: 1.6.8
 - Launcher version: 2.1.0

BrandtHill:
closes #10900 

In render method, provide additional check to see if `WidgetInfo.LEVEL_UP` widget is loaded.

------
This solution works in all cases whereupon leveling up Fishing does NOT unlock new fish.
In the cases where you do unlock fish, there is a second dialogue describing which new fish you can catch. If you click through the level up dialogue, the fishing status will resolve to "Fishing" as if the second dialogue wasn't there. For other skills (Cooking, Mining) when you unlock something upon leveling up, a widget with groupId `DIALOG_SPRITE_GROUP_ID`, 193, is loaded, which would enable checking against presence of `WidgetInfo.DIALOG_SPRITE` to determine if player was in 'unlock new stuff' dialogue. However, when unlocking Fishing, the loaded groupId is 11, which is not assigned to anything in `WidgetID.java`, and a WidgetInfo.DIALOG_SPRITE widget is never present while leveling up Fishing.

If this solution isn't satisfactory, I can provide an additional check against the player's animationId versus an immutable set of all fishing animations, which would work, but it's just a little heavy-handed.

Brandt

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -103,6 +103,14 @@ public final class AnimationID
 	public static final int FISHING_CRUSHING_INFERNAL_EELS = 7553;
 	public static final int FISHING_CUTTING_SACRED_EELS = 7151;
 	public static final int FISHING_BAREHAND = 6709;
+	public static final int FISHING_BAREHAND_WINDUP_1 = 6703;
+	public static final int FISHING_BAREHAND_WINDUP_2 = 6704;
+	public static final int FISHING_BAREHAND_CAUGHT_SHARK_1 = 6705;
+	public static final int FISHING_BAREHAND_CAUGHT_SHARK_2 = 6706;
+	public static final int FISHING_BAREHAND_CAUGHT_SWORDFISH_1 = 6707;
+	public static final int FISHING_BAREHAND_CAUGHT_SWORDFISH_2 = 6708;
+	public static final int FISHING_BAREHAND_CAUGHT_TUNA_1 = 6710;
+	public static final int FISHING_BAREHAND_CAUGHT_TUNA_2 = 6711;
 	public static final int MINING_BRONZE_PICKAXE = 625;
 	public static final int MINING_IRON_PICKAXE = 626;
 	public static final int MINING_STEEL_PICKAXE = 627;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
@@ -27,7 +27,10 @@ package net.runelite.client.plugins.fishing;
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.util.Set;
 import javax.inject.Inject;
+import com.google.common.collect.ImmutableSet;
+import net.runelite.api.AnimationID;
 import net.runelite.api.Client;
 import net.runelite.api.GraphicID;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY;
@@ -46,6 +49,28 @@ class FishingOverlay extends Overlay
 {
 	private static final String FISHING_SPOT = "Fishing spot";
 	static final String FISHING_RESET = "Reset";
+
+	private static final Set<Integer> FISHING_ANIMATIONS = ImmutableSet.of(
+		AnimationID.FISHING_BARBTAIL_HARPOON,
+		AnimationID.FISHING_BAREHAND,
+		AnimationID.FISHING_BAREHAND_CAUGHT_SHARK_1,
+		AnimationID.FISHING_BAREHAND_CAUGHT_SHARK_2,
+		AnimationID.FISHING_BAREHAND_CAUGHT_SWORDFISH_1,
+		AnimationID.FISHING_BAREHAND_CAUGHT_SWORDFISH_2,
+		AnimationID.FISHING_BAREHAND_CAUGHT_TUNA_1,
+		AnimationID.FISHING_BAREHAND_CAUGHT_TUNA_2,
+		AnimationID.FISHING_BAREHAND_WINDUP_1,
+		AnimationID.FISHING_BAREHAND_WINDUP_2,
+		AnimationID.FISHING_BIG_NET,
+		AnimationID.FISHING_CAGE,
+		AnimationID.FISHING_CRYSTAL_HARPOON,
+		AnimationID.FISHING_DRAGON_HARPOON,
+		AnimationID.FISHING_HARPOON,
+		AnimationID.FISHING_INFERNAL_HARPOON,
+		AnimationID.FISHING_KARAMBWAN,
+		AnimationID.FISHING_NET,
+		AnimationID.FISHING_OILY_ROD,
+		AnimationID.FISHING_POLE_CAST);
 
 	private final Client client;
 	private final FishingPlugin plugin;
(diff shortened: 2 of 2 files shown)


Adam-:
This is probably the wrong way to approach the problem since any interface will stall the fishing, not just level up. I would either not fix it or make it also timeout if the fishing anim hasn't played recently.

BrandtHill:
Updated to check if current animation is fishing

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -54,6 +57,23 @@
 
 	private final PanelComponent panelComponent = new PanelComponent();
 
+	private final ImmutableSet<Integer> fishingAnimations = new ImmutableSet.Builder<Integer>()
+			.add(AnimationID.FISHING_BARBTAIL_HARPOON)
+			.add(AnimationID.FISHING_BAREHAND)
+			.add(AnimationID.FISHING_BIG_NET)
+			.add(AnimationID.FISHING_CAGE)
+			.add(AnimationID.FISHING_CRUSHING_INFERNAL_EELS)
```
Nightfirecat: This isn't a fishing animation

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -54,6 +57,23 @@
...
+			.add(AnimationID.FISHING_CRYSTAL_HARPOON)
+			.add(AnimationID.FISHING_CUTTING_SACRED_EELS)
```
Nightfirecat: This isn't a fishing animation

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -78,7 +98,9 @@ public Dimension render(Graphics2D graphics)
 		panelComponent.getChildren().clear();
 		if (client.getLocalPlayer().getInteracting() != null
 			&& client.getLocalPlayer().getInteracting().getName().contains(FISHING_SPOT)
-			&& client.getLocalPlayer().getInteracting().getGraphic() != GraphicID.FLYING_FISH)
+			&& client.getLocalPlayer().getInteracting().getGraphic() != GraphicID.FLYING_FISH
+			&& client.getWidget(WidgetInfo.LEVEL_UP) == null
```
Nightfirecat: I don't see a need to specifically check this case if we're adding an animation check.

BrandtHill: After level up dialogue the fishing animation continues for some time, and it was unclear to me if fishing was actually taking place, but I agree it's unneeded and it has been removed along with non-fishing animations.

Adam-:
Has someone tested with this all of the fishing spots? I know when we tried something similar with MLM the mining anim there actually stops for a time when you are still mining

Adam-:
Also the set of anims the idle notifier acts on is different, does that matter?
```
			case FISHING_CRUSHING_INFERNAL_EELS:
			case FISHING_CUTTING_SACRED_EELS:
			case FISHING_BIG_NET:
			case FISHING_NET:
			case FISHING_POLE_CAST:
			case FISHING_CAGE:
			case FISHING_HARPOON:
			case FISHING_BARBTAIL_HARPOON:
			case FISHING_DRAGON_HARPOON:
			case FISHING_INFERNAL_HARPOON:
			case FISHING_CRYSTAL_HARPOON:
			case FISHING_OILY_ROD:
			case FISHING_KARAMBWAN:
			case FISHING_BAREHAND:
```

Nightfirecat:
The idle notifier has some animations which aren't actually fishing animations. (crushing infernal eels, cutting sacred eels)

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -54,6 +56,21 @@
 
 	private final PanelComponent panelComponent = new PanelComponent();
 
+	private final ImmutableSet<Integer> fishingAnimations = new ImmutableSet.Builder<Integer>()
+			.add(AnimationID.FISHING_BARBTAIL_HARPOON)
+			.add(AnimationID.FISHING_BAREHAND)
```
Nightfirecat: This animation is uninterrupted, however a different animation plays when a player has caught a fish bare-handed, and that animation is missing here.

BrandtHill: I found four new undocumented animations (6703, 6704 upon start, 6710, 6711 upon catch) whilst barehanded fishing for Tuna. My Fishing level is 63. Can anyone with a higher fishing level let me know what the animation IDs are for Swordfish and Sharks?
To find I'd just throw `if (client.getLocalPlayer().getAnimation() != -1) log.info("Animation: {}", client.getLocalPlayer().getAnimation());` at the top of `onGameTick()`

runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -103,6 +103,10 @@
 	public static final int FISHING_CRUSHING_INFERNAL_EELS = 7553;
 	public static final int FISHING_CUTTING_SACRED_EELS = 7151;
 	public static final int FISHING_BAREHAND = 6709;
+	public static final int FISHING_BAREHAND_WINDUP_1 = 6703;
+	public static final int FISHING_BAREHAND_WINDUP_2 = 6704;
+	public static final int FISHING_BAREHAND_CAUGHT_1 = 6710;
+	public static final int FISHING_BAREHAND_CAUGHT_2 = 6711;
```
BrandtHill: These Barehanded caught animations (and maybe windup ones too) are specific to Tuna, so names are subject to change once we figure out what Shark/Swordfish ones are

Nightfirecat: These are not named correctly. Since these apply only to tuna, they should be named appropriately to reflect that.
```suggestion
	public static final int FISHING_BAREHAND_CAUGHT_TUNA_1 = 6710;
	public static final int FISHING_BAREHAND_CAUGHT_TUNA_2 = 6711;
```

BrandtHill: Thank you for finding the rest of these animations.

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -54,6 +56,29 @@
 
 	private final PanelComponent panelComponent = new PanelComponent();
 
+	private final ImmutableSet<Integer> fishingAnimations = new ImmutableSet.Builder<Integer>()
```
Adam-:
```suggestion
	private final ImmutableSet<Integer> fishingAnimations = new ImmutableSet.Builder<>()
```

BrandtHill: Type inference doesn't work here

Adam-: use https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.html#of(E,%20E,%20E,%20E,%20E,%20E,%20E...)

BrandtHill: You got it

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingOverlay.java
```diff
@@ -54,6 +56,28 @@
 
 	private final PanelComponent panelComponent = new PanelComponent();
 
+	private final ImmutableSet<Integer> fishingAnimations = ImmutableSet.of(
```
Nightfirecat: This field can be made `static`, be defined as a `Set` (or even a `Collection` since we're not calling any set-specific methods of it), and renamed to match the constant naming schema:
```suggestion
	private static final Set<Integer> FISHING_ANIMATIONS = ImmutableSet.of(
```

PR #10913, opened 2020-02-29
Add an option to show the average amount for a loot source in the loot tracker

mikesena:
I've found myself manually calculating the average anyway out of curiousity.

This adds a tooltip to the number of kills label so that the average can be seen.

For those who'd prefer, it also allows the total GP to be replaced with the average value.
The tooltip will show the option not selected.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
@@ -222,6 +222,7 @@ class LootTrackerBox extends JPanel
 		if (kills > 1)
 		{
 			subTitleLabel.setText("x " + kills);
+			subTitleLabel.setToolTipText(QuantityFormatter.formatNumber(totalPrice / kills) + " gp (average)");
 		}
 
 		validate();


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -69,6 +69,16 @@ default boolean showPriceType()
 		return false;
 	}
 
+	@ConfigItem(
+		keyName = "showAverage",
+		name = "Show Average Price",
+		description = "Whether to show the total value or the average"
```
Nightfirecat:
```suggestion
		description = "Show the average value of each loot instead of the sum total value"
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -215,13 +218,24 @@ void rebuild()
...
+		if (showAverage)
+		{
+			priceLabel.setText(averageString);
+			priceLabel.setToolTipText(totalString + " (total)");
+		}
+		else
+		{
+			priceLabel.setText(priceTypeString + QuantityFormatter.quantityToStackSize(totalPrice));
+			priceLabel.setToolTipText("<html>" + QuantityFormatter.formatNumber(totalPrice) + " gp<br>" + averageString + " (average)</html>");			
+		}
```
Nightfirecat: Why does this change add the average string to the tooltip even when the configuration option is disabled?

mikesena: @Nightfirecat  - I saw the setting as the displayed value, and the tooltip as a means to show the opposite.  2 reasons for doing that:
1. I assume it won't be the option that people will want displayed by default.  The number of kills / values (i.e. "x 135") currently doesn't display anything.  So even though it's disabled, I think it's reasonable to add that as a hover item.
2. Both figures are valuable to know - and I think there'd be interest from many in seeing both.  The hover text is simply the one you don't normally want to see.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -215,13 +218,24 @@ void rebuild()
...
+		String averageString = kills > 1 ? QuantityFormatter.formatNumber(totalPrice / kills) + " gp" : "0";
+		String totalString = priceTypeString + QuantityFormatter.formatNumber(totalPrice) + " gp";
+		if (showAverage)
+		{
+			priceLabel.setText(averageString);
+			priceLabel.setToolTipText(totalString + " (total)");
+		}
+		else
+		{
+			priceLabel.setText(priceTypeString + QuantityFormatter.quantityToStackSize(totalPrice));
```
Nightfirecat: Compared to the price label text prior to this change, this is missing the `+ " gp"` suffix. `totalString` should be used instead
```suggestion
			priceLabel.setText(totalString);
```

mikesena: I can't believe I missed that sorry.  I've pushed up a change that removes the totalString, in favour of making it exactly the same as before.

mikesena: @Nightfirecat  - I think this conversation can be resolved

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -215,13 +218,23 @@ void rebuild()
...
+		String averageString = kills > 1 ? QuantityFormatter.formatNumber(totalPrice / kills) + " gp" : "0";
+		if (showAverage)
+		{
+			priceLabel.setText(averageString);
+			priceLabel.setToolTipText(priceTypeString + QuantityFormatter.formatNumber(totalPrice) + " gp (total)");
+		}
+		else
+		{
+			priceLabel.setText(priceTypeString + QuantityFormatter.quantityToStackSize(totalPrice) + " gp");
+			priceLabel.setToolTipText("<html>" + QuantityFormatter.formatNumber(totalPrice) + " gp<br>" + averageString + " (average)</html>");
```
Nightfirecat: There's loads of trailing space here
```suggestion
			priceLabel.setToolTipText("<html>" + QuantityFormatter.formatNumber(totalPrice) + " gp<br>" + averageString + " (average)</html>");
```

mikesena: Sigh, sorry - missed that too.  Fixed.

Adam-:
I don't see why this can't just be added to the tooltip? The current impl also doesn't look like it has anything on the UI to signify it is an average which might be confusing.

mikesena:
@Adam- That was my first implementation locally too.
The tooltip is the casual, curious user - what I'm doing most of the time.

After running it a bit locally, I think I'll end up having it as my default.  I'm more interested in the average long-term to decide whether a slayer task was worth it.  Leaving it up while doing gargoyles was informative to me... and I figure it may be for others too.

I'm happy to remove it and just have it as a tooltip if you prefer.  I didn't think there was harm in being able to have it shown.


My reasoning for the lack of a '(avg)' suffix, or something like that, was the character limit.  Some of the longer tasks are already pretty close to the shortened gp format (755k).  Players would know they've changed it too, so it's less likely to be confusing.

All that said - I'm also very happy to go with your suggestion and add a suffix if that's preferable.

mikesena:
@Adam- gave it more thought.. Just committed a change that shows an icon to indicate and toggle the total / average next to the 'show ignored items' button.  Let me know what you think.  I think it's a better spot for the setting anyway.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
```diff
@@ -90,4 +90,4 @@ default boolean syncPanel()
 	{
 		return true;
 	}
-}
\ No newline at end of file
+}
```
Nightfirecat: Remove this eof line addition now that there are no other config changes

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -215,13 +218,23 @@ void rebuild()
 			priceTypeString = priceType == LootTrackerPriceType.HIGH_ALCHEMY ? "HA: " : "GE: ";
 		}
 
-		priceLabel.setText(priceTypeString + QuantityFormatter.quantityToStackSize(totalPrice) + " gp");
-		priceLabel.setToolTipText(QuantityFormatter.formatNumber(totalPrice) + " gp");
+		String averageString = kills > 1 ? QuantityFormatter.formatNumber(totalPrice / kills) + " gp" : "0";
```
Nightfirecat: This makes the average value of a single kill 0, which is not correct.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -215,13 +218,23 @@ void rebuild()
...
+		{
+			priceLabel.setText(priceTypeString + QuantityFormatter.quantityToStackSize(totalPrice) + " gp");
+			priceLabel.setToolTipText("<html>" + QuantityFormatter.formatNumber(totalPrice) + " gp<br>" + averageString + " (average)</html>");
+		}
 
 		final long kills = getTotalKills();
 		if (kills > 1)
 		{
 			subTitleLabel.setText("x " + kills);
+			subTitleLabel.setToolTipText(averageString + " (average)");
```
Nightfirecat: Why is this tooltip added here? Given that this is price information, it should only be handled with the price data on the right, which is already covered in the code above.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
mikesena: Done.  I squashed my 2 commits together - assuming that the visual button is a better approach.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
mikesena: Correct, great pick-up.  I've made this > 0, as it was just meant to be a safety check for the division by zero.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
mikesena: This was my original way of doing it.  I hovered over the kill count, hoping it would show me the average.  Maybe others have the same approach.

If you're okay for it to be left it, I think it's a good location to always be able to see the average, without clicking the button.  I can remove it if you'd prefer though.

Nightfirecat: I don't think it belongs here, but I'll let others comment with their feelings on this.

Adam-:
I'd rather this just be a tooltip since I think it is too niche to have this much change.

mikesena:
> I'd rather this just be a tooltip since I think it is too niche to have this much change.

I'm obviously new here - you guys make the call.  I think it's a helpful thing - having the button to display the average feels like a useful thing to me... but, maybe it is too niche.  Happy to just go with the tooltip if you'd prefer.  Need either you or @Nightfirecat to make a call :)

My personal vote is the button toggle displaying average on all at once, rather than just the tooltip.

Nightfirecat:
I'll defer to Adam's call here, let's stick with a tooltip for now.

mikesena:
> I'll defer to Adam's call here, let's stick with a tooltip for now.

Too easy - updated the commit to just be the tooltip.

PR #10915, opened 2020-02-29
Add shift-modifier to bank deposit box

### LINKED ISSUE #10914: Deposit shift-modifier does not work on bank deposit boxes
title

haakonrp:
Uses same option as for bank deposit if possible.

Fixes #10914


Adam-:
I don't see why this needs its own config option? Can't you just reuse the bank deposit one?

haakonrp:
True, will do that.

haakonrp:
The only problem is that the identifiers and types of the menu options when using the deposit box is different than the bank's menu options.

Adam-:
I would add the identifiers for the bank deposit boxes to the ShiftDepositMode enum, then remove the menu action from it since it can be computed from the identifier (`identifier >= 6 ? low : normal)`

haakonrp:
How would you switch between the bank and deposit box identifiers? Check if deposit box interface is open? Have to be able to decide which identifier to pass to the bankModeSwap-method.

haakonrp:
Or maybe something like this works?
`final int opId = menuEntryAdded.getActionParam0() == 4 ? shiftDepositMode.getIdentifier() : shiftDepositMode.getIdentifierDepositBox();`

Adam-:
It looks like your PR already has code to detect whether or not it is a deposit box or not, so I dont see why this would be an issue?

haakonrp:
The check is not specific to deposit boxes, will be true for both bank depositing and deposit boxes. Can't find a smart way to check if it's a deposit box tho, got any ideas? :)

Adam-:
I guess the confusion here is I'd assumed your initial PR worked, but it looks like it doesn't work. One of the parameters to the menu will be the component id and you should be able to extract an interface id from that, could just do that?

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -314,8 +314,8 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 			&& menuEntryAdded.getOption().startsWith("Deposit-"))
 		{
 			ShiftDepositMode shiftDepositMode = config.bankDepositShiftClick();
-			final int actionId = shiftDepositMode.getMenuAction().getId();
-			final int opId = shiftDepositMode.getIdentifier();
+			final int opId = menuEntryAdded.getActionParam1() == 983043 ? shiftDepositMode.getIdentifier() : shiftDepositMode.getIdentifierDepositBox();
```
Adam-:
```suggestion
			final int opId = TO_GROUP(menuEntryAdded.getActionParam1()) == BANK_INVENTORY_GROUP_ID ? shiftDepositMode.getIdentifier() : shiftDepositMode.getIdentifierDepositBox();
```

Adam-: It also may be worth inverting this logic so it doesn't break any non-bank cases now where it happens to work, though I am not aware of any.

PR #10919, opened 2020-03-01
!kc chat command - Pull hiscore values defaulting to RL config

### LINKED ISSUE #10897: make new command !rank similar to !lvl (which works for bosses), to show boss kc+rank
Currently _!kc Vorkath_ would show this:
`Vorkath kill count: 195`
and _!lvl Vorkath_ would show this:
`Level Vorkath: 195 Experience: -1 Rank 83,429`

This is cool, however the experience part doesn't really fit, using Level Vorkath as the prefix sounds weird, and using !lvl to show kc is unintuitive.
### Suggestion
make !kc work like !lvl and use the highscores, or introduce new command !rank which works analogous to how !lvl Vorkath would work, while removing the "Experience" part, making _!rank vorkath_ show:
`Vorkath kill count: 195 Rank 83,429`
### Pros
+ Flex your KC rank

### Cons
- Highscores not up-to-date until you relog
- Highscores unreliable recently, 
=> add new "!rank"  command instead of changing old "!kc" command

TheStonedTurtle:
If the activity is not supported or the hiscore request times out the RL config value will be used.
If the request succeeds it will show the rank and the highest level between the hiscore and RL values.

image

I've also made it so that if the stored config value is lower than the hiscore result for the local player then the config is updated to reflect the hiscore result.

Closes #10897


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -1509,12 +1553,12 @@ private static String longBossName(String boss)
 			// The Gauntlet
 			case "gaunt":
 			case "gauntlet":
-				return "Gauntlet";
```
Alexsuperfly: youll need to add these as cases for each gauntlet because thats what the kc message has

TheStonedTurtle: I'm not sure i follow what you are saying alex, I need the `The ` prefixed onto the long boss name result for the gauntlet values. This is because the HiscoreSkill has `THE_GAUNTLET` and `THE_CORRUPTED_GAUNTLET` and we use `HiscoreSkill.valueOf()` to grab the correct enum value.

I can remove `THE_` from that enum but it'll require changing multiple things, including renaming two images, so this seemed like the cleanest solution.

Alexsuperfly: the the will be fine if thats what it needs to return, you just also need to have a case for what they currently are because the kc message that the game gives you when you complete one doesnt have a `the` in it.

Adam-:
I'd like to wait for the hiscores to be fixed first before adding more dependencies on it.

Jerkin42:
Isn't this to be able to merge now?

TheStonedTurtle:
The `!lvl` command can accomplish this task so I'm closing the PR

PR #10927, opened 2020-03-02
emote clue: Add missing ornament items

### LINKED ISSUE #10899: Add item variations to emote clues
[From today's news post:](https://secure.runescape.com/m=news/game-improvements-and-email-linking-for-mobile-accounts?oldschool=1)

> Items with more than one variant (e.g. Godswords with ornament kits or dragon items with a gold trim) can now be used and stored in STASH units when doing Treasure Trails. If you find other items that should be included, let us know in your feedback!

I don't have a list of items on hand, but I would recommend [checking the list of items which can be decorated](https://oldschool.runescape.wiki/w/Ornament_kit) and checking for usages of the decorated items to see if they are used.

Dava96:
closes #10899


Nightfirecat:
Looks like this covers all the cases we hadn't already covered before. :+1:

PR #10933, opened 2020-03-03
Feature - Kingdom of Miscellania notification message

### LINKED ISSUE #10939: Feature - Kingdom of Miscellania notification message
Picture this - you're logging on to Runescape after not having played for a few weeks. You afk NMZ for a few hours while watching Vampire Diaries with the Misses. You log off. Alas, all this while you've had money in the dwindling coffer of your vassal state, Miscellania, but your approval rating is so low that your once-loyal subjects aren't bringing in nearly as much coal and mahogany logs as you'd prefer. If only there was a way to be reminded of your diplomatic duties...

Introducing additional functionality to the Kingdom Plugin. Wouldn't it be great to be shown the current status of you coffer and favor/approval ratings? Check out the screenshot below:
image

*But how does it work? I thought the Varbits associated with Managing Miscellania were only sent to your client by the server when you're in the bounds of Miscellania.*
Astute observation. This plugin works by saving the state of your coffer and favor when you visit Miscallenia, and when when you haven't visited Miscellania in more than a day, the displayed values will be calculated based on how many days have passed since you last knew about your favor/coffer status.

Luckily I've already implemented this feature in Pull Request #10933 

~~If anyone has any requests, like making this configurable with thresholds, let me know.~~
I made it configurable with thresholds

BrandtHill:
I dreamt of an additional feature for the KingdomPlugin several months and finally decided to implement it. Upon login, if the user is managing the kingdom, their coffer and favor will be sent as a chat notification. The values are calculated using the last known coffer and favor values the same way the server calculates them.

image

~~Let me know if I should create a corresponding issue for this feature and amend this PR to close it.~~
closes #10939 

Brandt

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Infinitay <https://github.com/Infinitay>
+ * Copyright (c) 2020, Brandt Hill <https://github.com/BrandtHill>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -24,30 +24,54 @@
  */
 package net.runelite.client.plugins.kingdomofmiscellania;
 
-import java.awt.image.BufferedImage;
-import net.runelite.client.ui.overlay.infobox.Counter;
-import net.runelite.client.util.QuantityFormatter;
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
 
-public class KingdomCounter extends Counter
+@ConfigGroup(KingdomConfig.CONFIG_GROUP_NAME)
+public interface KingdomConfig extends Config
 {
-	private final KingdomPlugin plugin;
+	String CONFIG_GROUP_NAME = "kingdomofmiscellania";
+	int MAX_COFFER = 7_500_000;
+	int MAX_APPROVAL_PERCENT = 100;
 
-	KingdomCounter(BufferedImage image, KingdomPlugin plugin)
+	@ConfigItem(
+		position = 1,
+		keyName = "sendNotifications",
+		name = "Send Notifications",
+		description = "Send chat notifications upon login showing current estimated coffer and approval"
+	)
+	default boolean shouldSendNotifications()
 	{
-		super(image, plugin, plugin.getApproval());
-		this.plugin = plugin;
+		return false;
 	}
 
-	@Override
-	public String getText()
+	@Range(
+		max = MAX_COFFER
+	)
+	@ConfigItem(
+		position = 2,
+		keyName = "cofferThreshold",
+		name = "Coffer Threshold",
+		description = "Send notifications if coffer is below this value"
+	)
+	default int getCofferThreshold()
 	{
-		return KingdomPlugin.getApprovalPercent(plugin.getApproval()) + "%";
+		return MAX_COFFER;
 	}
 
-	@Override
-	public String getTooltip()
+	@Range(
+		max = MAX_APPROVAL_PERCENT
+	)
+	@ConfigItem(
+		position = 3,
+		keyName = "approvalThreshold",
+		name = "Approval Threshold",
+		description = "Send notifications if approval percentage is below this value"
+	)
+	default int getApprovalThreshold()
 	{
-		return "Approval: " + plugin.getApproval() + "/" + KingdomPlugin.MAX_APPROVAL + "</br>"
-			+ "Coffer: " + QuantityFormatter.quantityToStackSize(plugin.getCoffer());
+		return MAX_APPROVAL_PERCENT;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
@@ -25,32 +25,55 @@
 package net.runelite.client.plugins.kingdomofmiscellania;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.inject.Provides;
+import java.time.Duration;
+import java.time.Instant;
 import javax.inject.Inject;
-import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import static net.runelite.api.ItemID.TEAK_CHEST;
+import net.runelite.api.Quest;
+import net.runelite.api.QuestState;
 import net.runelite.api.VarPlayer;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.api.events.VarbitChanged;
+import net.runelite.client.chat.ChatColorType;
+import net.runelite.client.chat.ChatMessageBuilder;
+import net.runelite.client.chat.ChatMessageManager;
+import net.runelite.client.chat.QueuedMessage;
+import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
+import net.runelite.client.util.QuantityFormatter;
 
 @PluginDescriptor(
 	name = "Kingdom of Miscellania",
-	description = "Show amount of favor when inside Miscellania",
-	tags = {"favor", "favour", "managing", "overlay"},
+	description = "Show amount of approval when inside Miscellania",
+	tags = {"favor", "favour", "managing", "overlay", "approval", "coffer"},
 	enabledByDefault = false
 )
 @Slf4j
 public class KingdomPlugin extends Plugin
 {
 	private static final ImmutableSet<Integer> KINGDOM_REGION = ImmutableSet.of(10044, 10300);
+	private static final String CONFIG_LAST_CHANGED_KEY = "lastChanged";
+	private static final String CONFIG_COFFER_KEY = "coffer";
+	private static final String CONFIG_APPROVAL_KEY = "approval";
+	private static final String CHAT_MESSAGE_FORMAT = "Your Kingdom of Miscellania approval is %d%%, and your coffer has %s coins.";
+	private static final int MAX_WITHDRAWAL_BASE = 50_000;
+	private static final int MAX_WITHDRAWAL_ROYAL_TROUBLE = 75_000;
+	private static final float APPROVAL_DECREMENT_BASE = 0.025f;
+	private static final float APPROVAL_DECREMENT_ROYAL_TROUBLE = 0.010f;
+	static final int MAX_APPROVAL = 127;
+
+	private boolean loggingIn;
 
 	@Inject
 	private Client client;
@@ -61,8 +84,14 @@ public class KingdomPlugin extends Plugin
 	@Inject
 	private ItemManager itemManager;
 
-	@Getter
-	private int favor = 0, coffer = 0;
+	@Inject
+	private KingdomConfig config;
+
+	@Inject
+	private ConfigManager configManager;
+
+	@Inject
+	private ChatMessageManager chatMessageManager;
 
 	private KingdomCounter counter;
 
@@ -72,11 +101,28 @@ public class KingdomPlugin extends Plugin
 		removeKingdomInfobox();
 	}
 
+	@Provides
+	KingdomConfig getConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(KingdomConfig.class);
+	}
+
 	@Subscribe
 	public void onVarbitChanged(VarbitChanged event)
 	{
-		favor = client.getVar(Varbits.KINGDOM_FAVOR);
-		coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		final int coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		final int approval = client.getVar(Varbits.KINGDOM_APPROVAL);
+
+		if (client.getGameState() == GameState.LOGGED_IN
+			&& isThroneOfMiscellaniaCompleted()
+			&& (isInKingdom() || coffer > 0 && approval > 0)
+			&& (getCoffer() != coffer || getApproval() != approval))
+		{
+			setLastChanged(Instant.now());
+			setCoffer(coffer);
+			setApproval(approval);
+		}
+
 		processInfobox();
 	}
 
@@ -87,11 +133,25 @@ public class KingdomPlugin extends Plugin
 		{
 			processInfobox();
 		}
+		else if (event.getGameState() == GameState.LOGGING_IN)
+		{
+			loggingIn = true;
+		}
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick gameTick)
+	{
+		if (loggingIn)
+		{
+			loggingIn = false;
+			createNotification();
+		}
 	}
 
 	private void processInfobox()
 	{
-		if (client.getGameState() == GameState.LOGGED_IN && hasCompletedQuest() && isInKingdom())
+		if (client.getGameState() == GameState.LOGGED_IN && isThroneOfMiscellaniaCompleted() && isInKingdom())
 		{
 			addKingdomInfobox();
 		}
@@ -99,7 +159,33 @@ public class KingdomPlugin extends Plugin
 		{
 			removeKingdomInfobox();
 		}
+	}
 
+	private void createNotification()
+	{
+		if (!config.shouldSendNotifications() || !isThroneOfMiscellaniaCompleted())
+		{
+			return;
+		}
+
+		if (getLastChanged() == null)
+		{
+			log.debug("Kingdom Of Miscellania values not yet set. Visit Miscellania to automatically set values.");
+			return;
+		}
+
+		Instant lastChanged = getLastChanged();
+		int lastCoffer = getCoffer();
+		int lastApproval = getApproval();
+		int estimatedCoffer = estimateCoffer(lastChanged, lastCoffer);
+		int estimatedApproval = estimateApproval(lastChanged, lastApproval);
+		if (estimatedCoffer < config.getCofferThreshold() || getApprovalPercent(estimatedApproval) < config.getApprovalThreshold())
+		{
+			sendChatMessage(String.format(
+				CHAT_MESSAGE_FORMAT,
+				getApprovalPercent(estimatedApproval),
+				QuantityFormatter.quantityToStackSize(estimatedCoffer)));
+		}
 	}
 
 	private void addKingdomInfobox()
@@ -122,20 +208,92 @@ public class KingdomPlugin extends Plugin
 		}
 	}
 
+	private int estimateCoffer(Instant lastChanged, int lastCoffer)
+	{
+		int daysSince = (int) Duration.between(lastChanged, Instant.now()).toDays();
+		int maxDailyWithdrawal = isRoyalTroubleCompleted() ? MAX_WITHDRAWAL_ROYAL_TROUBLE : MAX_WITHDRAWAL_BASE;
+		int maxDailyThreshold = maxDailyWithdrawal * 10;
+
+		for (int i = 0; i < daysSince; i++)
+		{
+			lastCoffer -= (lastCoffer > maxDailyThreshold) ? maxDailyWithdrawal : lastCoffer / 10;
+		}
+		return lastCoffer;
+	}
+
+	private int estimateApproval(Instant lastChanged, int lastApproval)
+	{
+		int daysSince = (int) Duration.between(lastChanged, Instant.now()).toDays();
+		float dailyPercentage = isRoyalTroubleCompleted() ? APPROVAL_DECREMENT_ROYAL_TROUBLE : APPROVAL_DECREMENT_BASE;
+
+		lastApproval -= (int) (daysSince * dailyPercentage * MAX_APPROVAL);
+		return Math.max(lastApproval, 0);
+	}
+
 	private boolean isInKingdom()
 	{
 		return client.getLocalPlayer() != null
 			&& KINGDOM_REGION.contains(client.getLocalPlayer().getWorldLocation().getRegionID());
 	}
 
-	private boolean hasCompletedQuest()
+	private boolean isThroneOfMiscellaniaCompleted()
 	{
 		return client.getVar(VarPlayer.THRONE_OF_MISCELLANIA) > 0;
 	}
 
-	static int getFavorPercent(int favor)
+	private boolean isRoyalTroubleCompleted()
 	{
-		return (favor * 100) / 127;
+		return Quest.ROYAL_TROUBLE.getState(client) == QuestState.FINISHED;
 	}
 
+	static int getApprovalPercent(int approval)
+	{
+		return (approval * 100) / MAX_APPROVAL;
+	}
+
+	private void sendChatMessage(String chatMessage)
+	{
+		final String message = new ChatMessageBuilder()
+			.append(ChatColorType.HIGHLIGHT)
+			.append(chatMessage)
+			.build();
+
+		chatMessageManager.queue(
+			QueuedMessage.builder()
+				.type(ChatMessageType.CONSOLE)
+				.runeLiteFormattedMessage(message)
+				.build());
+	}
+
+	private Instant getLastChanged()
+	{
+		return configManager.getRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED_KEY, Instant.class);
+	}
+
+	private void setLastChanged(Instant lastChanged)
+	{
+		configManager.setRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED_KEY, lastChanged);
+	}
+
+	int getCoffer()
+	{
+		Integer coffer = configManager.getRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_COFFER_KEY, int.class);
+		return coffer == null ? 0 : coffer;
+	}
+
+	private void setCoffer(int coffer)
+	{
+		configManager.setRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_COFFER_KEY, coffer);
+	}
+
+	int getApproval()
+	{
+		Integer approval = configManager.getRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_APPROVAL_KEY, int.class);
+		return approval == null ? 0 : approval;
+	}
+
+	private void setApproval(int approval)
+	{
+		configManager.setRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_APPROVAL_KEY, approval);
+	}
 }
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -387,9 +387,10 @@ public enum Varbits
 	MULTICOMBAT_AREA(4605),
 
 	/**
-	 * Kingdom Management
+	 * Kingdom of Miscellania Management
+	 * Kingdom Approval is represented as a 7-bit unsigned integer; 127 corresponds to 100% approval
 	 */
-	KINGDOM_FAVOR(72),
+	KINGDOM_APPROVAL(72),
 	KINGDOM_COFFER(74),
 
 	/**


runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -99,7 +138,40 @@ private void processInfobox()
 		{
 			removeKingdomInfobox();
 		}
+	}
+
+	private void setKnownValues()
+	{
+		if (client.getGameState() == GameState.LOGGED_IN && hasCompletedQuest() && isInKingdom())
+		{
+			configManager.setConfiguration(CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED, Instant.now().getEpochSecond());
```
raiyni: These should be added as hidden config options that can be then be used by the overlay and your code

BrandtHill: Do you mind pointing me towards an example of this?

Trevor159: https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java#L131

BrandtHill: Thank you. Added KingdomConfig.
Is it a known issue that ConfigManager can't store longs?

raiyni: store the Instant see `ConfigManager::stringToObject`

BrandtHill: Changed to store lastChanged as java.time.Instant instead of int representation of decaseconds epoch

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -52,9 +52,8 @@ default int getFavorThreshold()
 	}
 
 	/** Stored data **/
-	// Epoch in *Decaseconds of last time coffer or favor values changed.
-	// Default is -1. Must return positive value to display notification message
-	// *Stored as integer in Decaseconds because ConfigManager unable to store longs, and 2038 problem
+	// Instant of last coffer or favor change.
+	// Default is -1 epoch. Must return positive epoch value to display notification message
```
raiyni: Comments in this file aren't needed

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,109 @@
+package net.runelite.client.plugins.kingdomofmiscellania;
```
raiyni: This needs the license header

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,109 @@
...
+			keyName = "favor",
+			name = "",
+			description = "",
+			hidden = true
+	)
+	default int getFavor()
+	{
+		return -1;
+	}
+	@ConfigItem(
```
raiyni: space above this line

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,109 @@
...
+			keyName = "coffer",
+			name = "",
+			description = "",
+			hidden = true
+	)
+	default int getCoffer()
+	{
+		return -1;
+	}
+	@ConfigItem(
```
raiyni: space above this line

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,109 @@
...
+			keyName = "lastChanged",
+			name = "",
+			description = "",
+			hidden = true
+	)
+	default Instant getLastChanged()
+	{
+		return Instant.ofEpochSecond(-1);
+	}
+	@ConfigItem(
```
raiyni: space above this line

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -99,7 +142,43 @@ private void processInfobox()
 		{
 			removeKingdomInfobox();
 		}
+	}
 
+	private void setKnownValues()
```
raiyni: this doesn't need to be a function

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -133,9 +234,27 @@ private boolean hasCompletedQuest()
 		return client.getVar(VarPlayer.THRONE_OF_MISCELLANIA) > 0;
 	}
 
+	private boolean hasCompletedNextQuest()
```
raiyni:
```suggestion
	private boolean royalTroubleCompleted()
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
BrandtHill: Added license header

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
BrandtHill: Inlined function

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
BrandtHill: Renamed

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
BrandtHill: Removed comments

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -61,6 +77,12 @@
...
 	private ItemManager itemManager;
 
+	@Inject
+	private KingdomConfig config;
+
+	@Inject
+	private ChatMessageManager chatMessageManager;
+
 	@Getter
 	private int favor = 0, coffer = 0;
```
raiyni: These can be removed in favor of using the config

BrandtHill: True. Changed in latest commit,

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +189,60 @@ private void removeKingdomInfobox()
 		}
 	}
 
+	private int estimateCoffer(long lastChangedEpoch, int lastCoffer)
+	{
+		int daysSince = Math.max((int) (Instant.now().getEpochSecond() - lastChangedEpoch) / 86_400, 0);
```
Nightfirecat: Don't use a magic number to represent one day's worth of seconds. Make a `private static final int` constant instead and represent it with `(int) Duration.of(1, ChronoUnit.DAYS).get(ChronoUnit.SECONDS)`

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +189,60 @@ private void removeKingdomInfobox()
...
+		int maxDailyWithdrawl = royalTroubleCompleted() ? 75_000 : 50_000;
```
Nightfirecat: 50k and 75k should also be defined as constants.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +189,60 @@ private void removeKingdomInfobox()
...
+		}
+		return lastCoffer;
+	}
+
+	private int estimateFavor(long lastChangedEpoch, int lastFavor)
+	{
+		int daysSince = Math.max((int) (Instant.now().getEpochSecond() - lastChangedEpoch) / 86_400, 0);
+		float dailyPercentage = royalTroubleCompleted() ? 0.01f : 0.025f;
+
+		lastFavor -= (int) (daysSince * dailyPercentage * 127);
```
Nightfirecat: `127` should also be made into a constant and should get a healthy dose of javadoc to explain that approval rating is stored as an unsigned 7-bit int. (Perhaps this should be added to the varbit entry for kingdom favor)

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -51,6 +64,15 @@
...
 	private static final ImmutableSet<Integer> KINGDOM_REGION = ImmutableSet.of(10044, 10300);
+	private static final String CHAT_MESSAGE_FORMAT = "Your Kingdom of Miscellania favor is %d%%, and your coffer has %s coins.";
+	private static final int SECONDS_IN_DAY = (int) Duration.of(1, ChronoUnit.DAYS).getSeconds();
+	private static final int MAX_WITHDRAWAL_BASE = 50_000;
+	private static final int MAX_WITHDRAWAL_ROYAL_TROUBLE = 75_000;
+	private static final float FAVOR_DECREMENT_BASE = 0.025f;
+	private static final float FAVOR_DECREMENT_ROYAL_TROUBLE = 0.01f;
+	protected static final int MAX_FAVOR = 127;
+
+	private boolean loggingIn, shouldSendNotification;
```
Nightfirecat: Is `shouldSendNotification` needed? It seems like sending notifications is already fully guarded by `config.shouldSendNotifications()`, `loggingIn`,and `throneOfMiscellaniaCompleted()`.

BrandtHill: True, createNotification() is only called in one place after checking loggingIn. I’ll change when I get home.

BrandtHill: I think I originally used `shouldSendNotification` as the only flag, then I created `loggingIn` since I had to wait for the first game tick for `royalTroubleCompleted()` to return accurately, making it obsolete. Anyway, I've removed it.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,136 @@
...
+	/**
+	 * Stored data
+	 **/
+	@ConfigItem(
+		keyName = "lastChanged",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default Instant getLastChanged()
```
Adam-: you can just use `null` for this

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -51,6 +64,15 @@
 public class KingdomPlugin extends Plugin
 {
 	private static final ImmutableSet<Integer> KINGDOM_REGION = ImmutableSet.of(10044, 10300);
+	private static final String CHAT_MESSAGE_FORMAT = "Your Kingdom of Miscellania favor is %d%%, and your coffer has %s coins.";
+	private static final int SECONDS_IN_DAY = (int) Duration.of(1, ChronoUnit.DAYS).getSeconds();
+	private static final int MAX_WITHDRAWAL_BASE = 50_000;
+	private static final int MAX_WITHDRAWAL_ROYAL_TROUBLE = 75_000;
+	private static final float FAVOR_DECREMENT_BASE = 0.025f;
+	private static final float FAVOR_DECREMENT_ROYAL_TROUBLE = 0.010f;
+	protected static final int MAX_FAVOR = 127;
```
Adam-:
```suggestion
	static final int MAX_FAVOR = 127;
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -72,11 +97,21 @@ protected void shutDown() throws Exception
...
+	}
+
 	@Subscribe
 	public void onVarbitChanged(VarbitChanged event)
 	{
-		favor = client.getVar(Varbits.KINGDOM_FAVOR);
-		coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		if (client.getGameState() == GameState.LOGGED_IN && throneOfMiscellaniaCompleted() && isInKingdom())
+		{
+			config.setLastChanged(Instant.now());
```
Adam-: this isn't always necessarily the kingdom vars changing here, is that fine?

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -87,26 +122,65 @@ public void onGameStateChanged(GameStateChanged event)
...
+				{
+					sendChatMessage(String.format(
+						CHAT_MESSAGE_FORMAT,
+						getFavorPercent(estimatedFavor),
+						QuantityFormatter.quantityToStackSize(estimatedCoffer)));
+				}
+			}
+			else
+			{
+				log.info("Kingdom Of Miscellania values not yet set. Visit Miscellania to automatically set values.");
```
Adam-: this should be debug level

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
BrandtHill: When the `isInKingdom()` guard returns true, the majority of `VarbitChanged` events changing are those of the kingdom values, and the few extraneous events don't impact the kingdom values.

Adam-: Right, I just mean it would bump the time. That's fine?

BrandtHill: The value really is `lastTimeKnown`. For most cases it shouldn't matter, but I'll add a check to make the value make accurate to its purpose.

BrandtHill:
Let me know if there's anything that needs to be changed.

BrandtHill:
Happy to make any requested changes for approval.

BrandtHill:
Let me know if anything needs to be changed.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +197,60 @@ private void removeKingdomInfobox()
...
+	}
+
 	private boolean isInKingdom()
 	{
 		return client.getLocalPlayer() != null
 			&& KINGDOM_REGION.contains(client.getLocalPlayer().getWorldLocation().getRegionID());
 	}
 
-	private boolean hasCompletedQuest()
+	private boolean throneOfMiscellaniaCompleted()
```
Nightfirecat, line 228:
```suggestion
	private boolean isThroneOfMiscellaniaCompleted()
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +197,60 @@ private void removeKingdomInfobox()
...
 	{
 		return client.getVar(VarPlayer.THRONE_OF_MISCELLANIA) > 0;
 	}
 
+	private boolean royalTroubleCompleted()
```
Nightfirecat, line 233:
```suggestion
	private boolean isRoyalTroubleCompleted()
```

Nightfirecat (commented):
Mostly small fixups. This PR generally looks good

BrandtHill:
Thanks for re-reviewing. Suggested changes have been committed.

Nightfirecat (approved):
LGTM. Thanks for keeping up with this PR all this time!

BrandtHill:
> LGTM. Thanks for keeping up with this PR all this time!

I never lost hope!

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+package net.runelite.client.plugins.kingdomofmiscellania;
+
+import java.time.Instant;
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
+
+@ConfigGroup("kingdomOfMiscellania")
```
Adam-, line 33:
```suggestion
@ConfigGroup("kingdomofmiscellania")
```

Generally we keep these all lowercase.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,134 @@
...
+
+	@ConfigItem(
+		position = 1,
+		keyName = "sendNotifications",
+		name = "Send Notifications",
+		description = "Send chat notifications upon login showing current estimated coffer and favor"
+	)
+	default boolean shouldSendNotifications()
+	{
+		return true;
```
Adam-, line 47: I think the daily task plugin already spams people enough, I don't want to add more spam to the login by default

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,134 @@
...
+	)
+	default int getFavorThreshold()
+	{
+		return MAX_FAVOR_PERCENT;
+	}
+
+	/**
+	 * Stored data
+	 **/
+	@ConfigItem(
```
Adam-, line 81: This and everything below here is per-profile, and so should use the set/get rsprofile functions in config manager, so that it works correctly when using more than one account/profile.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +198,60 @@ private void removeKingdomInfobox()
 		}
 	}
 
+	private int estimateCoffer(long lastChangedEpoch, int lastCoffer)
+	{
+		int daysSince = Math.max((int) (Instant.now().getEpochSecond() - lastChangedEpoch) / SECONDS_IN_DAY, 0);
```
Adam-, line 203: You can just use the java time apis for conversion between time units, 
```java
(int) Duration.between(Instant.now(), lastChanged).toDays()
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +203,90 @@ private void removeKingdomInfobox()
...
+			.build();
+
+		chatMessageManager.queue(
+			QueuedMessage.builder()
+				.type(ChatMessageType.CONSOLE)
+				.runeLiteFormattedMessage(message)
+				.build());
 	}
 
+	protected Instant getLastChanged()
```
Nightfirecat, line 263: These methods can all be `private` or package-private instead of `protected` as we will not subclass the plugin.
```suggestion
	private Instant getLastChanged()
```

Specifically, it looks like each of these added methods can be `private` besides `getCoffer()` and `getFavor()` as they are referenced from the counter.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +203,90 @@ private void removeKingdomInfobox()
...
+	{
+		return configManager.getRSProfileConfiguration(CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED_KEY, Instant.class);
+	}
+
+	protected void setLastChanged(Instant lastChanged)
+	{
+		configManager.setConfiguration(CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED_KEY, lastChanged);
+	}
+
+	protected int getCoffer()
```
Nightfirecat, line 273: As mentioned above, package-private.
```suggestion
	int getCoffer()
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -51,6 +63,18 @@
 public class KingdomPlugin extends Plugin
 {
 	private static final ImmutableSet<Integer> KINGDOM_REGION = ImmutableSet.of(10044, 10300);
+	private static final String CONFIG_GROUP_NAME = "kingdomofmiscellania";
```
Nightfirecat, line 66: Define this as a constant in the config interface, and reference it both in the `@ConfigGroup` definition and here in the plugin.

Nightfirecat (changes requested):

BrandtHill:
Regrettably, I no longer have a membership. 😔 😔 Would someone be so kind as to test this? Simply visit Miscellania, enable the notification message, and relog. Ty 😗

Nightfirecat:
The changes look fine to me. I'll put some money in miscellania and play with this feature enabled for a few days and see how it looks.

Nightfirecat:
This doesn't seem to be working currently. I turned on notifications with a coffer threshold of 0 and a favor threshold of 100, but got no notification a day and a half later after depositing some money into my kingdom coffers. I'll look at this more later to see if I can't figure out why it's not working.

BrandtHill:
Thanks for testing. I'll get a membership and test this out.

BrandtHill:
I've fixed bugs relating to the new config functions.

I've simulated the passage of time by calling the following function with different values and logging on and verifying the message is sent with approximations.
```java
setLastChanged(Instant.ofEpochSecond(1613522615));
setCoffer(1_000_000);
setFavor(127);
```

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +202,92 @@ private void removeKingdomInfobox()
...
+	private void setLastChanged(Instant lastChanged)
+	{
+		configManager.setRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_LAST_CHANGED_KEY, lastChanged);
+	}
+
+	int getCoffer()
+	{
+		Integer coffer = configManager.getRSProfileConfiguration(KingdomConfig.CONFIG_GROUP_NAME, CONFIG_COFFER_KEY, int.class);
+		return coffer == null ? 0 : coffer;
+	}
```
Nightfirecat, line 276: It strikes me that these should be nullable as we would not want to send notifications to players for whom we don't know their coffer/favor values. They would obviously be set once they return to Miscellania, but if a player were to turn these notifications on prior to going back to the island, they could get notifications for "0" coffer/favor values being returned here rather than `null` which would indicate they're not yet known.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
BrandtHill: This is accounted for, because `getLastChanged()` will return `null` until you’ve visited Miscellania, and no message will be sent.

Nightfirecat (approved):

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomConfig.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	@Range(
+		max = MAX_COFFER
+	)
+	@ConfigItem(
+		position = 2,
+		keyName = "cofferThreshold",
+		name = "Coffer Threshold",
+		description = "Send notifications if coffer is at or below this value"
```
Nightfirecat, line 57: After having tested this on a dev build for a few days, I think this would work better if it tested for the coffer/favor being strictly less than the config. That way, players can set the values to 0 to disable a specific check, or leave them at their maximum values to get notifications whenever they're not maximum.

BrandtHill: Good thinking. I've made that change.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -72,11 +101,22 @@ protected void shutDown() throws Exception
...
+		return configManager.getConfig(KingdomConfig.class);
+	}
+
 	@Subscribe
 	public void onVarbitChanged(VarbitChanged event)
 	{
-		favor = client.getVar(Varbits.KINGDOM_FAVOR);
-		coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		if (client.getGameState() == GameState.LOGGED_IN && isThroneOfMiscellaniaCompleted() && isInKingdom()
+			&& (getCoffer() != client.getVar(Varbits.KINGDOM_COFFER) || getFavor() != client.getVar(Varbits.KINGDOM_FAVOR)))
```
Nightfirecat, line 114: One annoying thing I found about this code is that this will not automatically update the values when visiting Etceteria until some varbit change happens after you're there. For instance, if you travel to Etceteria via spirit tree then leave right away, these conditions will never be met because `isInKingdom()` is `false` during the varbit changes while you're changing game states.

Because of the `isKingdom()` false negative, I propose this subscriber be changed somewhat to handle this case.

* If `inKingdom()` is false and either coffer or favor values are 0, don't set any new values (this will be the case in some varbit changed events as the indexes are transmitted)
* If `inKingdom()` is false, coffer and favor are both non-zero, **and** either of their values are different from the stored values, set lastChanged and the values (the server seems to send the last known values from visiting misc/etc--if we didn't get a varbitchanged on the island itself, we should update the values from varbit changes elsewhere in the world)
* If `inKingdom()` is true and either value is different from the stored values, set lastChanged and the values

ie.

```java
public void onVarbitChanged(VarbitChanged event)
{
	final int coffer = client.getVar(Varbits.KINGDOM_COFFER);
	final int favor = client.getVar(Varbits.KINGDOM_FAVOR);

	if (client.getGameState() == GameState.LOGGED_IN
		&& isThroneOfMiscellaniaCompleted()
		&& (isInKingdom() || coffer != 0 && favor != 0)
		&& (coffer != getCoffer() && favor != getFavor()))
	{
		setLastChanged(Instant.now());
		setCoffer(coffer);
		setFavor(favor);
	}

	processInfobox();
}
```

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -51,6 +63,17 @@
 public class KingdomPlugin extends Plugin
 {
 	private static final ImmutableSet<Integer> KINGDOM_REGION = ImmutableSet.of(10044, 10300);
+	private static final String CONFIG_LAST_CHANGED_KEY = "lastChanged";
+	private static final String CONFIG_COFFER_KEY = "coffer";
+	private static final String CONFIG_FAVOR_KEY = "favor";
+	private static final String CHAT_MESSAGE_FORMAT = "Your Kingdom of Miscellania favor is %d%%, and your coffer has %s coins.";
```
Adam-, line 69: why isn't this `%d coins`?

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -122,20 +207,92 @@ private void removeKingdomInfobox()
 		}
 	}
 
+	private int estimateCoffer(long lastChangedEpoch, int lastCoffer)
```
Adam-, line 210: You already have the last changed as an `Instant` so there's no reason to convert to epoch second and then back like you are doing here and in `estimateFavor`.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
BrandtHill: It prints as a string like `1.16M`, not an int like`1160395`

Adam-: Oh you're right. This seems fine then, but I think "favor" is actually "approval rating" (I cannot find any use of favor/favour to describe it anywhere). Might be best to change this here and in the configs etc.

runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -72,11 +101,27 @@ protected void shutDown() throws Exception
...
 	{
-		favor = client.getVar(Varbits.KINGDOM_FAVOR);
-		coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		final int coffer = client.getVar(Varbits.KINGDOM_COFFER);
+		final int favor = client.getVar(Varbits.KINGDOM_FAVOR);
+
+		if (client.getGameState() == GameState.LOGGED_IN
+			&& isThroneOfMiscellaniaCompleted()
+			&& (isInKingdom() || coffer > 0 && favor > 0)
+			&& (getCoffer() != coffer && getFavor() != favor))
```
Adam-, line 119: This should be `||` right?

Adam-: I think so, I've changed it

BrandtHill:
I renamed everything from `favor` to `approval` and changed the guard to `||`

BrandtHill:
Just here to say happy birthday to this PR. You would have been one year old today. Merged before your time.

PR #10949, opened 2020-03-04
Change Varbits.java to include bolt pouch

Kalakoi:
Add varbits for bolt pouch quantities and bolt types.


Trevor159:
Do you need this for an external plugin?

Kalakoi:
Yes, it is for an external plugin

Trevor159:
I think we would not mind having a bolt pouch overlay in master. If you are set on making an external plugin you can use client.getVarbitValue that we made so external plugins do not need to touch the varbits file.

Kalakoi:
If you would like the whole bolt pouch overlay plugin in master I can modify this pull request to include the entirety of it instead of just the varbits or make another pull request. The plugin is working and based on the rune pouch plugin.

Trevor159:
I am unsure whether we would like this as its own plugin or combined with the rune pouch plugin but I think it would be fine in master.

Nightfirecat:
This definition can be hard-coded into the external plugin or introduced with a main repo feature which uses it. Since there's no real reason to have this added in isolation, I am closing this PR.

PR #10952, opened 2020-03-04
ground items plugin: turn double tap to hide off by default

Trevor159:
Too many people trigger this on accident without knowing it is a feature. I think it is reasonable for people to have to set a delay to turn on the feature.


deathbeam:
Other clients have/had this on by default, so I personally dont think we should disable it.

Nightfirecat:
Revisiting this; I don't often see this being an issue, so I don't think it should be disabled by default.

PR #10962, opened 2020-03-06
Add Dps plugin

TheStonedTurtle:
Now that tinted hitsplats are a thing we can easily track damage dealt by the local player.

5TB3mlpR6u


makingcomment:
A couple questions/comments:

- I'm not sure the "Show total damage instead of DPS" option should be defaulted to true. It seems a little counter intuitive to have the plugin go against its name by default.
- Can the overlay title change based on whether total damage or DPS is being shown? Having it say "DPS Counter" when it's actually counting total damage could be misleading/confusing.
- When the counter is paused, can the "pause" right click option be changed "unpause" in order to reflect that? If a user pauses, forgets they paused, and attempts to pause again, they'll accidentally unpause.
- When the counter is paused, does it make sense that the counter automatically unpauses the next time something is attacked? I guess whether this would be preferred behavior depends on the reason for pausing, so I don't really know what to think about it.

PR #10965, opened 2020-03-06
Move Emoji out of plugin and set public access

hjdarnel:
This is in preparation for an Emoji external plugin I want to put together. This moves the Emoji class to the ui package `net.runelite.client.ui` and sets some stuff to public so external plugins can use it.


abextm:
This is definitely the wrong package, and I can't see any way this is useful. What are you trying to do?

hjdarnel:
@abextm we talked in Discord about making this class public so that external plugins could reference that list. I'm going to make an external plugin to replace what I originally wanted from #10780, but I don't want to keep up a list of emojis that the client has. It was suggested that I move this away from the existing `EmojiPlugin` since we like to keep those all private (?)

Does that make sense? I'm happy to rearrange in any way 😄

hjdarnel:
Sounds like we can do this with just reflection, so I'll close this.

PR #10966, opened 2020-03-06
cache: Create temporary foldeer when specified

Nightfirecat:
Prior to this commit, specifying a cache tmpdir via `-Dcache.tmpdir`
would fail if the given path did not already exist. This commit passes
that dir to the constructed TemporaryFolder, which creates that path
even when absent on the filesystem prior to executing
`getTemporaryFolder()`.


Nightfirecat:
Apparently I didn't check this thoroughly enough, it didn't work after another trial. I'll re-open if I get it right :upside_down_face:

PR #10970, opened 2020-03-06
client: fix default sessionfile/config args

Trevor159:


runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -179,12 +179,14 @@ public static void main(String[] args) throws Exception
 		parser.accepts("debug", "Show extra debugging output");
 
 		final ArgumentAcceptingOptionSpec<File> sessionfile = parser.accepts("sessionfile", "Use a specified session file")
-			.withRequiredArg().defaultsTo(DEFAULT_SESSION_FILE)
-			.withValuesConvertedBy(new ConfigFileConverter());
+			.withRequiredArg()
+			.withValuesConvertedBy(new ConfigFileConverter())
+			.defaultsTo(new File(RUNELITE_DIR, DEFAULT_SESSION_FILE));
```
deathbeam: The DEFAULT_SESSION_FILE should be File in first place, same with config file

deathbeam: Oh also you should use ofType

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -88,8 +88,9 @@
 	public static final File PROFILES_DIR = new File(RUNELITE_DIR, "profiles");
 	public static final File SCREENSHOT_DIR = new File(RUNELITE_DIR, "screenshots");
 	public static final File LOGS_DIR = new File(RUNELITE_DIR, "logs");
-	public static final String DEFAULT_SESSION_FILE = "session";
-	public static final String DEFAULT_CONFIG_FILE = "settings.properties";
+	public static final File DEFAULT_SESSION_FILE = new File(RUNELITE_DIR, "session");
+	public static final String DEFAULT_CONFIG_FILE_NAME = "settings.properties";
```
deathbeam: I would just remove this completely, as you can use DEFAULT_CONFIG_FILE.getName()

PR #10973, opened 2020-03-07
friendicons: add note icon in friends/ignore list

### LINKED ISSUE #10959: Show note icon next to friends/ignored players with a note
In the friends and ignore lists, you can add a note to player names. Would be cool to show an icon next to the name if there is a note on their name. Could be a toggleable setting in the Friend List RuneLite plugin.

ThatGamerBlue:
Closes #10959 

https://i.imgur.com/Q0kAszB.png


runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +100,94 @@
...
+		if (iconIdx != -1 || modIcons == null)
+		{
+			return;
+		}
+
+		final IndexedSprite[] newIcons = Arrays.copyOf(modIcons, modIcons.length + 1);
+		iconIdx = modIcons.length;
+		try
+		{
+			final BufferedImage iconImg = ImageIO.read(NotesPlugin.class.getResourceAsStream("notes_icon.png"));
```
Nightfirecat: This should use its own icon. (whether that's a copy of the notes plugin's icon or something else is fine) Maybe there's an ingame sprite which might work well for this?

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -68,6 +82,8 @@
 	private static final String EDIT_NOTE = "Edit Note";
 	private static final String NOTE_PROMPT_FORMAT = "%s's Notes<br>" +
 		ColorUtil.prependColorTag("(Limit %s Characters)", new Color(0, 0, 170));
+	private static final int ICON_WIDTH = 14;
+	private static final int ICON_HEIGHT = 12;
```
Nightfirecat: How do you come to 14x12 as the dimensions? The scripts seem to indicate that the "Previous name" icon is 14x14

runelite-api/src/main/java/net/runelite/api/ScriptID.java
```diff
@@ -233,4 +233,16 @@
 	 */
 	@ScriptArguments(integer = 2)
 	public static final int TOPLEVEL_RESIZE = 909;
+
+	/**
+	 * Called when the friends list is updated
```
Nightfirecat: Since the meaning the parameters is known, they should be listed in the javadoc. See some of the earlier ScriptID entries for examples.

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +100,94 @@
...
 		overlayManager.add(overlay);
+		loadIcon();
+		if (client.getGameState() == GameState.LOGGED_IN)
+		{
+			rebuildFriendsList();
+			rebuildIgnoreList();
+		}
+	}
+
+	@Subscribe
```
Nightfirecat: While this file doesn't seem to maintain the usual definitions order of `getConfig()`, `startUp()`, `shutDown()`, `@Subscribe`rs, then instance/static methods (it has some instance methods mixed with the subscribers), `shutDown()` should definitely be defined here before defining any more subscribers.

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +100,94 @@
...
+		if (event.getGameState() == GameState.LOGGED_IN)
+		{
+			loadIcon();
+		}
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (!event.getGroup().equals(CONFIG_GROUP))
```
Nightfirecat: Check both config group and config key.

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesConfig.java
```diff
@@ -0,0 +1,20 @@
+package net.runelite.client.plugins.friendnotes;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+
+@ConfigGroup("friendNotes")
```
Nightfirecat: Make the plugin's config group package-private and reference it here.

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -258,4 +349,76 @@ public void onRemovedFriend(RemovedFriend event)
 		log.debug("Remove friend: '{}'", displayName);
 		setFriendNote(displayName, null);
 	}
+
+	@Subscribe
+	public void onScriptCallbackEvent(ScriptCallbackEvent event)
+	{
+		if (config.showIcons())
```
Nightfirecat: Use a guard clause instead:
```java
if (!config.showIcons())
{
	return;
}

// function code continues here
```

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -258,4 +349,76 @@ public void onRemovedFriend(RemovedFriend event)
...
+		{
+			if (event.getEventName().equals("friend_cc_settext") || event.getEventName().equals("ignore_cc_settext"))
```
Nightfirecat: A `switch` statement would make this a bit cleaner:

```java
switch (event.getName())
{
	case "friend_cc_settext":
	case "ignore_cc_settext":
	{
		// ...
	}
	case "friend_cc_setposition":
	case "ignore_cc_setposition":
	{
		// ...
	}
}
```

Nightfirecat:
Also, we typically like to have commits which add scripts separate from commits which modify scripts. Can you add the scripts and their hashes in one commit, then the rest afterward?

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
ThatGamerBlue: we need to rebuild the list when a note changes too, imo the easiest way is to do it here

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
ThatGamerBlue: this is the dimensions of the icon that we add

ThatGamerBlue:
as far as i can tell, theyre run once on login, when someone logs in/out and when you add/remove someone

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
ThatGamerBlue: couldnt find a good in game sprite (has to be a modicon) so copypasted the notes icon

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
Nightfirecat: List rebuilding can be explicitly added to the tail end of `setFriendNote()` instead of implicitly called here after the config change event dispatches. Alternately, you can check that the key matches `"showIcons"` or starts with `KEY_PREFIX`.

Nightfirecat:
Also the first commit should include the unmodified scripts and hashes, then the second commit should add the `runelite_callback` entries and any other changes needed for the scripts.

ThatGamerBlue:
done

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +99,74 @@
...
+		{
+			return;
+		}
+
+		if (!event.getKey().equals("showIcons"))
+		{
+			return;
+		}
+
+		if (client.getGameState() == GameState.LOGGED_IN)
```
Nightfirecat: These conditions can all be combined.
```suggestion
		if (event.getGroup().equals(CONFIG_GROUP) && event.getKey().equals("showIcons") && client.getGameState() == GameState.LOGGED_IN)
```

ThatGamerBlue: easier to extend when its like this imo

ThatGamerBlue: replaced with a switch statement for future expansion if wanted

ThatGamerBlue:
rebased

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +99,74 @@
...
+	private ClientThread clientThread;
+
+	@Inject
+	private FriendNotesConfig config;
+
 	@Getter
 	private HoveredFriend hoveredFriend = null;
 
+	private int iconIdx = -1;
+	private String currentlyLayouting = null;
```
Nightfirecat: This is the default initialized value of this field, it does not need to be explicitly defined.
```suggestion
	private String currentlyLayouting;
```

ThatGamerBlue: changed

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +99,74 @@
...
+	private FriendNotesConfig getConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(FriendNotesConfig.class);
+	}
+
 	@Override
 	protected void startUp() throws Exception
 	{
 		overlayManager.add(overlay);
+		loadIcon();
```
Adam-: This is not thread-safe.

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -258,4 +333,107 @@ public void onRemovedFriend(RemovedFriend event)
...
+			case "friend_cc_settext":
+			case "ignore_cc_settext":
+				String[] stringStack = client.getStringStack();
+				int stringStackSize = client.getStringStackSize();
+				final String rsn = stringStack[stringStackSize - 1];
+				final String sanitized = Text.toJagexName(Text.removeTags(rsn));
+				currentlyLayouting = sanitized;
+				if (getFriendNote(sanitized) != null)
+				{
+					stringStack[stringStackSize - 1] = rsn + " <img=" + iconIdx + ">";
```
Adam-: I would maybe also check iconidx somehow isn't set here

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
```diff
@@ -84,19 +99,74 @@
...
+			rebuildIgnoreList();
+		}
+	}
+
+	@Subscribe
+	public void onGameStateChanged(GameStateChanged event)
+	{
+		if (event.getGameState() == GameState.LOGGED_IN)
+		{
+			loadIcon();
```
Adam-: What is this for? can this happen?

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesConfig.java
```diff
@@ -0,0 +1,20 @@
+package net.runelite.client.plugins.friendnotes;
```
Adam-: The license header should be added to all new files.

ThatGamerBlue: added

runelite-client/src/main/java/net/runelite/client/plugins/friendnotes/FriendNotesPlugin.java
ThatGamerBlue: not sure if thats what you meant but i added a guard clause if iconidx == -1 return

ThatGamerBlue:
synchronized imageio read

PR #10976, opened 2020-03-07
Automatically enable NPC aggression timers after multiple kills

mikesena:
This allows the NPC aggression timers to activate after multiple kills are made by the player on a particular NPC.

This means you don't have to manually add each of the NPCs being tracked on a slayer task, and also don't have to leave the aggression timer on all the time, with lines being redrawn constantly.


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -470,6 +503,8 @@ public void onGameStateChanged(GameStateChanged event)
 					onLogin();
 				}
 
+				npcLootCount = 0;
+				previousNPCName = null;
```
Nightfirecat: This would reset the counter if a player hopped worlds or disconnected, which seems like the wrong time to do that.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
```diff
@@ -49,11 +49,22 @@ default boolean alwaysActive()
 		return false;
 	}
 
+	@ConfigItem(
+			keyName = "npcUnaggroAutoActivation",
+			name = "Auto activate",
+			description = "Automatically activate aggression timers after a kill count has been reached",
+			position = 2
```
Nightfirecat:
```suggestion
		keyName = "npcUnaggroAutoActivation",
		name = "Auto activate",
		description = "Automatically activate aggression timers after a kill count has been reached. Set to 0 to disable",
			position = 2
```

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
mikesena: You're correct - doesn't make sense to clear on a world hop.  Probably me still learning how the framework works; I misunderstood when that would be called.

mikesena: I've pushed the change up.

Nightfirecat: It doesn't look like you have

mikesena: Done

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -321,6 +331,28 @@ public void onNpcSpawned(NpcSpawned event)
 		checkAreaNpcs(event.getNpc());
 	}
 
+	@Subscribe
+	public void onNpcLootReceived(final NpcLootReceived npcLootReceived)
```
Nightfirecat: This isn't the right event to use to check whether an NPC you killed died, since not all monsters drop loot, and even from monsters which do drop loot, there are situations where you would not be the recipient of it. You would be better off using a check similar to that used in the opponent info plugin.

Adam-:
This might be a good thing to use the new hitsplat stuff for, since we can reliably tell if you are in combat with it and with what.

mikesena:
> This might be a good thing to use the new hitsplat stuff for, since we can reliably tell if you are in combat with it and with what.

@Adam, are you thinking instead of counting NPC deaths, count hitsplats applied?

That may end up working better than the current approach - despawning works if noone else is around, but there's the potential for interference.

I'll give that a try now.  Can revert back if that's not the intent.

mikesena:
@Adam-  I've applied the hitsplat approach.  The wording of the plugin config I think could be tweaked.. I found it hard to word well.

Adam-:
Do we even still need the manual enter an NPC by name thing? We could just detect the hitsplats, turn it on then, and have it timeout after a bit?

mikesena:
No we don’t yeah - I had the same thought.  I got rid of that in the last commit. The activation is just based on hit splats and a timeout to disable.

I did leave the check if the actor is a NPC in because we pass that into the check area func.

Adam-:
It doesn't look to me like you got rid of it, because it is still there?

mikesena:
> It doesn't look to me like you got rid of it, because it is still there?

I honestly may be doing something dumb here with git but the latest commit looks like its 1c5d62183ab8fcbd4a4535c6211d25df9a8bf79e, which doesn't have the npcName stuff I used before?

There's existing functionality in the plugin to do name matching - is that what you're picking up maybe?

Adam-:
Yes, I meant the existing name matching feature.

Adam-:
Per the Discord discussion I think what we settled on was

1) update the wiki to include the aggression type (y/n/always)?
2) modify the plugin to automatically enable when fighting yes-aggressive monsters

PR #10982, opened 2020-03-08
fix(xptoolTipCutOff): fixes xp tool cut off.

mrboots123:
When xp globes are moved to corners and hovered, tool tip is clipped.
east
southWest
west
southEast


Adam-:
I'm curious what this would do if we made this render using the existing tooltip system. This existing code is fairly old and predates our tooltips. The tooltip system would probably have to be extended to support arbitrary panel tooltips, but I think that might work.

Adam-:
@mrboots123 can you test #10983 ?

PR #10985, opened 2020-03-09
Allow Loot Tracker to track Herbiboar loot with an Open Herb Sack

### LINKED ISSUE #10655: Herbi kills + loot not being added to loot tracker when using an open herb sack
When harvesting Herbiboar with an open herb sack, your accumulated herbs and kc will not be added to your loot tracker.

Suggested fix: The game filter, when disabled, includes each herb placed into your sack. This includes duplicates, which are added line-by-line. I've included a screenshot of what collecting multiple herbs, along with a duplicate, looks like. herbi

Thanks
Adoptulous

### LINKED ISSUE #10718: Heribiboar Loot Tracker Issue
For some reason, if you are tracking herbiboar and sip/decant a potion it will add the new potion in your inventory to the loot tracker for herbiboar.  I am not sure what the exact fix is for this since I am not a programmer.
Screen Shot 2020-02-02 at 7 06 17 PM

deon9718:
Closes #10655
Closes #10718

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -63,6 +63,7 @@ import net.runelite.api.GameState;
 import net.runelite.api.InventoryID;
 import net.runelite.api.ItemComposition;
 import net.runelite.api.ItemContainer;
+import net.runelite.api.MessageNode;
 import net.runelite.api.NPC;
 import net.runelite.api.Player;
 import net.runelite.api.SpriteID;
@@ -116,8 +117,10 @@ public class LootTrackerPlugin extends Plugin
 	private static final int THEATRE_OF_BLOOD_REGION = 12867;
 
 	// Herbiboar loot handling
-	private static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
+	@VisibleForTesting
+	static final String HERBIBOAR_LOOTED_MESSAGE = "You harvest herbs from the herbiboar, whereupon it escapes.";
 	private static final String HERBIBOAR_EVENT = "Herbiboar";
+	private static final Pattern HERBIBOAR_HERB_SACK_PATTERN = Pattern.compile(".+(Grimy .+?) herb.+");
 
 	// Hespori loot handling
 	private static final String HESPORI_LOOTED_MESSAGE = "You have successfully cleared this patch for new crops.";
@@ -500,10 +503,14 @@ public class LootTrackerPlugin extends Plugin
 
 		if (message.equals(HERBIBOAR_LOOTED_MESSAGE))
 		{
+			if (processHerbiboarHerbSackLoot(event.getTimestamp()))
+			{
+				return;
+			}
+
 			eventType = HERBIBOAR_EVENT;
 			lootRecordType = LootRecordType.EVENT;
 			takeInventorySnapshot();
-
 			return;
 		}
 
@@ -675,6 +682,34 @@ public class LootTrackerPlugin extends Plugin
 
 			inventorySnapshot = null;
 		}
+	}
+
+	private boolean processHerbiboarHerbSackLoot(int timestamp)
+	{
+		List<ItemStack> herbs = new ArrayList<>();
+
+		for (MessageNode messageNode : client.getMessages())
+		{
+			if (messageNode.getTimestamp() != timestamp
+				|| messageNode.getType() != ChatMessageType.SPAM)
+			{
+				continue;
+			}
+
+			Matcher matcher = HERBIBOAR_HERB_SACK_PATTERN.matcher(messageNode.getValue());
+			if (matcher.matches())
+			{
+				herbs.add(new ItemStack(itemManager.search(matcher.group(1)).get(0).getId(), 1, client.getLocalPlayer().getLocalLocation()));
+			}
+		}
+
+		if (herbs.isEmpty())
+		{
+			return false;
+		}
+
+		addLoot(HERBIBOAR_EVENT, -1, LootRecordType.EVENT, herbs);
+		return true;
 	}
 
 	void toggleItem(String name, boolean ignore)
--- a/runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
@@ -24,32 +24,66 @@
  */
 package net.runelite.client.plugins.loottracker;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.inject.Guice;
 import com.google.inject.testing.fieldbinder.Bind;
 import com.google.inject.testing.fieldbinder.BoundFieldModule;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -677,6 +680,40 @@ private void processInventoryLoot(String event, LootRecordType lootRecordType, I
 		}
 	}
 
+	private boolean processHerbSackLoot(int timestamp, String event, LootRecordType lootRecordType)
```
Nightfirecat: Since this is specific to the herbiboar, this should be named as such.
```suggestion
	private boolean processHerbiboarHerbSackLoot(int timestamp, String event, LootRecordType lootRecordType)
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -677,6 +680,40 @@ private void processInventoryLoot(String event, LootRecordType lootRecordType, I
...
+	{
+		List<ItemStack> herbs = new ArrayList<>();
+		for (Object messageObject : client.getMessages())
+		{
+			MessageNode messageNode = (MessageNode) messageObject;
+			if (messageNode.getTimestamp() == timestamp && messageNode.getType() == ChatMessageType.SPAM)
```
Nightfirecat: This would be cleaner as a guard clause.
```suggestion
			if (messageNode.getTimestamp() != timestamp || messageNode.getType() != ChatMessageType.SPAM)
			{
				continue;
			}
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -677,6 +680,40 @@ private void processInventoryLoot(String event, LootRecordType lootRecordType, I
...
+			{
+				String herbName;
```
Nightfirecat: As I understand from the code below, there are two possible chat messages which can appear:

* You put the Grimy Guam leaf herb into your herb sack.
* Your herb sack is too full to hold the Grimy Guam leaf herb.

Wouldn't a pattern matching `.+(Grimy .+) herb.+` match the herb name fully without any of the extra logic below? Also, the pattern matching/replacing should be replaced with a compiled `Pattern` object and a matcher on the messageNode string.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -677,6 +680,40 @@ private void processInventoryLoot(String event, LootRecordType lootRecordType, I
...
+					herbs.add(new ItemStack(itemManager.search(herbName).get(0).getId(), 1, client.getLocalPlayer().getLocalLocation()));
+				}
+			}
+		}
+
+		if (herbs.isEmpty())
+		{
+			return false;
+		}
+		else
```
Nightfirecat: Since the `if` branch returns from this method, the rest of this code does not need to be in an `else` block. That is, this should be replaced by the following:

```java
if (herbs.isEmpty())
{
	return false;
}

addLoot(event, -1, lootRecordType, herbs);
return true;
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
deon9718: There is also another message when you manually add a herb to your sack which is:

- You add the Grimy Guam leaf to your sack.

But I don't think its possible to manually add herbs in the same timestamp as harvesting Herbiboar.

deon9718: Pattern ended up being `.+(Grimy .+?) herb.+`

deon9718:
@Nightfirecat Numulites or fossils were never tracked to begin with, the harvesting of Herbiboar is what triggers LootTracker to track loot in the first place, and Herbiboar will only drop herbs on harvest. Numulites and fossils are received when inspecting Herbiboar's trail. Although I believe those inspect events do also generate chat messages and can be incorporated into the LootTracker plugin if desired.

Nightfirecat:
You're right. I just checked and that does appear to be the case. :+1:

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -677,6 +684,36 @@ private void processInventoryLoot(String event, LootRecordType lootRecordType, I
 		}
 	}
 
+	private boolean processHerbiboarHerbSackLoot(int timestamp)
+	{
+		List<ItemStack> herbs = new ArrayList<>();
+
+		for (Object messageObject : client.getMessages())
```
Adam-: I've made `IterableHashTable` generic now to make this less terrifying

runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
```diff
@@ -106,4 +158,90 @@ public void testFirstClue()
...
+			when(node.getType()).thenReturn(ChatMessageType.SPAM);
+			when(node.getValue()).thenReturn(herbMessage);
+
+			MessageNode nodeFull = mock(MessageNode.class);
+			when(nodeFull.getTimestamp()).thenReturn(0);
+			when(nodeFull.getType()).thenReturn(ChatMessageType.SPAM);
+			when(nodeFull.getValue()).thenReturn(herbFullMessage);
+
+			IterableHashTable messageTable = mock(IterableHashTable.class);
+			Iterator<Object> mockIterator = mock(Iterator.class);
```
Nightfirecat:
```suggestion
			IterableHashTable<MessageNode> messageTable = mock(IterableHashTable.class);
			Iterator<MessageNode> mockIterator = mock(Iterator.class);
```

runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
```diff
@@ -106,4 +158,90 @@ public void testFirstClue()
...
+		herbPrice.setName("Grimy guam leaf");
+		when(itemManager.search("Grimy guam leaf")).thenReturn(Collections.singletonList(herbPrice));
+
+		MessageNode node = mock(MessageNode.class);
+		when(node.getTimestamp()).thenReturn(0);
+		when(node.getType()).thenReturn(ChatMessageType.SPAM);
+		when(node.getValue()).thenReturn("You put the Grimy guam leaf herb into your herb sack.");
+
+		IterableHashTable messageTable = mock(IterableHashTable.class);
+		Iterator<Object> mockIterator = mock(Iterator.class);
```
Nightfirecat:
```suggestion
		IterableHashTable<MessageNode> messageTable = mock(IterableHashTable.class);
		Iterator<MessageNode> mockIterator = mock(Iterator.class);
```

Adam-:
I am confused what this `testHerbiboarPotionBug` test is for, since it doesn't look like this fixes #10718?

Adam-:
I've modified the first test to just check the event type is null, which is the root of the problem the 2nd test was trying to catch, so I've removed it.

PR #10990, opened 2020-03-09
Use bank title for determining if tag tab menu is active

### LINKED ISSUE #10954: When you change an icon in the bank tag tag tab, you can no longer reorder them until you go in and out of it
Go in the bank tag tag tab, change an icon, try to reorder tag tabs.

I do love the update btw.

raiyni:
Closes #10954

* Check the bank title when in tag tab menu because the chatbox input can be reset 
* Reset the bank title when leaving the tag tab menu because the script hook relies on bank title to stop the normal bank layout
* Remove the tag options for inventory and item drag because it doesn't make sense to tag items as `tagtabs`
* Fix script that broke from an auto merge

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
@@ -688,7 +688,7 @@ public class TabInterface
 		{
 			entries = createMenuEntry(event, TAG_INVENTORY, event.getTarget(), entries);
 
-			if (activeTab != null)
+			if (activeTab != null && !isTabMenuActive())
 			{
 				entries = createMenuEntry(event, TAG_INVENTORY, ColorUtil.wrapWithColorTag(activeTab.getTag(), HILIGHT_COLOR), entries);
 			}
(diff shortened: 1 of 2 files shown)


Nightfirecat:
This should be broken into smaller commits since there's multiple changes happening here.

Nightfirecat:
I'm not sure if something changed between before and now, but the first commit (cfa843d4761971d193c04fc79765bc6c95b24fca) seems to have no functional change. On neither master nor that commit can I get a "Deposit inventory/equipment tagtabs" menu option to be added.

raiyni:
> This PR fixes the linked bug, but causes another bug which can be reproduced as follows:
> 
> 1. Open the bank
> 2. Open the tag tabs tab via right-clicking the "New tag tab" button and clicking the "View tag tabs" menu entry
> 3. Attempt to close this view by clicking the "View all items" bank tab
> 
> Expected: The tag tabs tab would close and display your bank items
> Actual: The view does not close, and cannot be exited without either (a) closing the bank entirely or (b) switching to another tag view and repeating step (3)

I see you tested this early in the morning. Was it by chance after the update or something? I rebased on master as of  7:30 CST and it works fine.

Nightfirecat:
After rebasing, the first of these commits still seems to have no change that I can see. I'm not seeing tag inventory/equipment for the tag tab before or after applying it.

Nightfirecat:
I can't reproduce the behavior I saw in that comment now either. :shrug: 
I'm no expert in the bank plugin, so maybe the last two commits' messages make more sense to the people who are more familiar with it, but they don't really help me understand the actual behavior change or make clear to me the need for the change. That said, based solely off trying to reproduce the bug this PR aims to address, they seem to work. :+1:

PR #7361, opened 2019-01-15
timetracking: add timer warning colors and ability to sort timers.

jakewilson:
Adds feature for #7240.

Sorting timers: https://media.giphy.com/media/1ymqqfEFRTfWv0Vi5u/giphy.gif (not displaying in github for some reason)

Warnings:

PR #10996, opened 2020-03-11
timetracking: add timer warning colors and ability to sort timers.

### LINKED ISSUE #7240: Sorting timers in `Time Tracking` by time left and timer text color change at last seconds.
There would be much more ease of use if you could **sort timers by how much time is left**. _Again,_ I use timers for rune mining. Each timer represents a world and is started when rune depletes, since they are not arranged this way it requires a lot of effort to find which timer is about to end counting. 

Also the **text color of a timer should change when there is little time left** (say orange 5-10 sec). And of course it would be more convenient if you could modify the color and time at which to alert in settings.

Thank you!

melkypie:
Originally made by @jakewilson 
Rebased, added proposed changes and tested. 
From my testing I fixed a couple of bugs, for example, timer going to 0 and then starting it over would show in warning color even if it was above the warning threshold.
Also when the timer is stopped in a warning state it shows the same color rather than a darker color.

Lets the user sort by ascending or descending time.
Lets the user choose a time for showing a timer warning in orange.
Closes #7240
Supersedes #7361

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/SortOrder.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/SortOrder.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2020, Jake Wilson <https://github.com/jakewilson>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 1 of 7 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/Timer.java
```diff
@@ -96,4 +103,9 @@ void reset()
 		remaining = duration;
 		lastUpdate = Instant.now().getEpochSecond();
 	}
+
+	boolean isWarning()
+	{
+		return (remaining > 0) && warning;
```
dekvall:
```suggestion
		return warning && remaining > 0;
```

melkypie:
Made the requested changes

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -83,6 +83,29 @@ default int defaultTimerMinutes()
...
+	}
+
+	@ConfigItem(
+		keyName = "timerWarningSeconds",
+		name = "Timer Warning (Seconds)",
+		description = "The time at which to change the timer color to the warning color",
+		position = 6
+	)
+	@Units(Units.SECONDS)
+	default int timerWarningSeconds()
```
dekvall: Now that it's using `Units` the name `Timer Warning (Seconds)` feels redundant, consider renaming the option and corresponding fields to something like `Timer Warning Threshold` or similar

Nightfirecat:
This should be split into separate commits for each of the features added.

melkypie:
Put them in 2 seperate commits.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/ClockManager.java
```diff
@@ -131,6 +134,36 @@ public boolean checkCompletion()
...
+			Comparator<Timer> comparator = Comparator.comparingLong(Timer::getDisplayTime);
+			if (sortOrder == SortOrder.DESC)
+			{
+				comparator = comparator.reversed();
+			}
+
+			if (!Comparators.isInOrder(timers, comparator))
+			{
+				timers.sort(comparator);
+				saveTimers();
```
Adam-: I think this can just be removed?

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/clocks/Timer.java
```diff
@@ -96,4 +103,9 @@ void reset()
 		remaining = duration;
 		lastUpdate = Instant.now().getEpochSecond();
 	}
+
+	boolean isWarning()
+	{
+		return warning && (remaining > 0) ;
```
Adam-:
```suggestion
		return warning && (remaining > 0);
```

melkypie: No idea how I missed that.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/SortOrder.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package net.runelite.client.plugins.timetracking;
+
+public enum SortOrder
+{
+	NONE("NONE"),
+	ASC("ASC"),
+	DESC("DESC");
+
+	private final String name;
```
Adam-: I think this is unused?

PR #10998, opened 2020-03-12
Swap Talk-To with Teleport on Wizard Distentor and Aubury (Essense Mine)

c0dybrown:


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -411,4 +411,14 @@ default ShiftDepositMode bankDepositShiftClick()
 	{
 		return ShiftDepositMode.OFF;
 	}
+
+	@ConfigItem(
+		keyName = "swapEssenceMineTeleport",
+		name = "Essence Mine Teleport",
+		description = "Swaps Talk-To with Teleport for Wizard Distentor and Aubury."
```
Nightfirecat: Why does this not swap teleport [for all of the essence mine teleport NPCs](https://oldschool.runescape.wiki/w/Rune_essence_mine#Teleportation_and_accessing_the_mine)?

c0dybrown: All others are impractical to use as a they are so far from a bank. Those are the only two anyone uses as per the second paragraph here: https://oldschool.runescape.wiki/w/Pure_essence#Obtaining_pure_essence

dekvall: I don't think that it matters that they are impractical to use, it doesn't make sense not to include all essence mine teleports in a config named `Essence Mine Teleport`.

c0dybrown: It makes sense to not have to maintain a massive list of NPC names statically defined in your codebase until the end of time, especially when virtually all of them are unused. Look at how it affects the style in MenuEntrySwapperPlugin.java. Feel free to close the request if you don't like it, I'll just keep using it locally myself.

Nightfirecat: This could be solved by creating a constant `List` or `Set` with the names of these NPCs and checking `ESSENCE_TELEPORT_NPCS.contains(target)` instead of adding a new condition for each npc.

Adam-:
There are only 5 npcs which would need this swap. It makes sense to me to make it work on all of them, as there are undoubtedly people which do use them.

PR #11004, opened 2020-03-13
Update CipherClue.java

aimuari:
Include instructions on how to access the fairy hideout via "Fairy Ring code AIR, then DLR, then DJQ, then AJS" as per instructions here: 

https://oldschool.runescape.wiki/w/Fairy_Resistance_Hideout

I just did this clue twice and had to look it up both times. Save a bit of time looking it up.


aimuari:
Have a nice day.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CipherClue.java
```diff
@@ -55,7 +55,7 @@
 		new CipherClue("USBJCPSO", "Traiborn", new WorldPoint(3112, 3162, 0), "First floor of Wizards Tower", "How many air runes would I need to cast 630 wind waves?", "3150"),
 		new CipherClue("HCKTA IQFHCVJGT", "Fairy Godfather", new WorldPoint(2446, 4428, 0), "Zanaris throne room", "There are 3 inputs and 4 letters on each ring How many total individual fairy ring codes are possible?", "64"),
 		new CipherClue("ZSBKDO ZODO", "Pirate Pete", new WorldPoint(3680, 3537, 0), "Dock northeast of the Ectofunctus"),
-		new CipherClue("GBJSZ RVFFO", "Fairy Queen", new WorldPoint(2347, 4435, 0), "Fairy Resistance Hideout"),
+		new CipherClue("GBJSZ RVFFO", "Fairy Queen", new WorldPoint(2347, 4435, 0), "Fairy Resistance Hideout", "Fairy Ring code AIR, then DLR, then DJQ, then AJS"),
```
Nightfirecat: There is no `CipherClue(String, String, WorldPoint, String, String)` constructor for this class. The fairy ring order should be added to the "Fairy Resistance Hideout" string, and should be made shorter so it is not so wordy.

abextm:
This information in already shown in the fairy ring interface, I don't think we need to duplicate it here

PR #11006, opened 2020-03-13
hiscorepanel: implement rsn validation for hiscore lookup

ypperlig:
Implements RSN validation for hiscorepanel, not wasting Jagex's precious bandwith by avoiding sending of unnecessary bad requests.


runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -78,6 +79,11 @@
 	/* The maximum allowed username length in RuneScape accounts */
 	private static final int MAX_USERNAME_LENGTH = 12;
 
+	/**
+	 * RSN pattern based on the following limitations: https://runescape.wiki/w/Character_name#Limitations
+	 */
+	private static final Pattern RSN_PATTERN = Pattern.compile(String.format("^(?!\\s)[a-zA-Z0-9-\\s]{1,%s}$", MAX_USERNAME_LENGTH));
```
Nightfirecat: `\\s` isn't used in usernames, `[ _-]` is. Hence, it should be explicitly written out. `\\w` can be used for most of the actual username characters so I've inserted that there. Also, javadoc can be formatted as such to have a "see also"-style link.
```suggestion
	/**
	 * Pattern matching valid RSNs based on ingame limitations
	 * {@see https://runescape.wiki/w/Character_name#Limitations}
	 */
	private static final Pattern RSN_PATTERN = Pattern.compile(String.format("^(?![ _-])[\\w -]{1,%s}$", MAX_USERNAME_LENGTH));
```

Nightfirecat: Also, are "space" characters permitted at the end of a player's display name? eg. `foobar_`

ypperlig: \s was in fact meant to be whitespace at the beginning of the string. 

As for beginning and end of RSN's, it seems that Jagex strips certain characters when you set the display name, as well as do a hiscore lookup.

So say if you had __Foo__ as your RSN, it would in fact be "Foo". Same goes for any other character except a-zA-Z0-9.

ypperlig: Updated with requested changes, added a negative lookbehind for "invalid" characters.
Not sure if this would be the best solution though, as RuneScape does strip certain characters from the player lookup value. If you do a search for your in-game name and add a couple of underscores at the end, the call will still go through.

For example, this is a perfectly valid lookup value for the API:
\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_Helmaks\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_

Nightfirecat: In that case, perhaps it would be better to replace `_-` with ` ` and strip the string before matching the username. That would then reduce the regex complexity by removing the need for the lookbehind in the first place, as well as the need to match all whitespace variants.

Adam-:
It might be good to get some tests for this for when it inevitably ends up blocking someones actual rsn.

Issue #10771, opened 2020-02-11
Nightmare of Ashihama update region tracking


Alexsuperfly:
- [x] Add theme-appropriate skybox colors to the new regions, iron and main account versions of the boss room as well
- [x] Add the new regions to discord plugin

Nightfirecat:
The dungeon region IDs reported by the "World Map Location" dev tool are accurate. That is, the following regions correspond to the dungeon:

* 14999
* 15000
* 15001
* 15255
* 15256
* 15257
* 15511
* 15512
* 15513

PR #11010, opened 2020-03-14
Added nightmare dungeon area to skybox.txt

AustinGarrettR:
Added nightmare dungeon area to skybox colors.
Previously (Line 218 originally in skybox.txt) encompassed a tiny room. That line was removed, and the new region encompasses the previously removed region.

The color chosen is #0a0a0a (RGB 10,10,10).

Resolves 1 of 2 in #10771 

image


Nightfirecat:
When this is merged, it will close #10771, as the discord status PR which addressed the other portion of it has already been merged.

PR #11013, opened 2020-03-14
ground items: remove Highlight > value, add Notify >= Tier

### LINKED ISSUE #9241: Valuable ground item color priority over highlighted ones
Ground items plugin works great, however i would really like the value tier colors to take priority over the one highlight color if it falls into these categories. The only choice is to either have all items above a value threshold become highlighted, meaning they will have the same (highlighted item) color OR you have to refrain from highlighting them to get the tier colors, but pass on all loot notifications by doing so.

I want the item value tier colors to take priority over the highlighted color, so that if a highlighted item drops, it still gets colored (or to stand out maybe multiplied) according to the Low,Medium,High,Insane tier setups while still throwing its highlight drop notification.

I have considered using the "Highlight > value" option and set it to my Low/Medium value setting so that i can still get notifications on drops that i want to pay attention to, but it removes the fun "rarity" colors by making them all the same highlighted items color.

Alternative non-existent solutions would be to separate item notifications from highlighted items, then adding two settings to toggle notifications on valuable and highlighted drops.

**Additional context**
I really love the existing value tier color feature and it just bugs me that i have to choose between it and receiving notifications. Thanks for taking the time to read my suggestion!

runelite suggestion

### LINKED ISSUE #10961: Please revert ground items change or give us an option
It is illogical to have it work this way, where "value >" takes priority over anything and everything else. That defeats the whole purpose of the other options in the plugin, and essentially downgrades the plugin to a more barebones variant.

Asking for this change to be undone might be too big of an ask, but at least give us an option to choose the behavior please.

I've considered giving up on the plugin.

**Additional context**
-

Hydrox6:
Closes #9241
Closes #10961

Highlighted items should serve a specific purpose imo: for items you're specifically looking for. With `Highlight > value`, it turned the Highlighted items into another value tier, with the only difference between it and the low->insane values was that it provided notifications.

With the recent change to how the tiers are ordered, and the confusion around it, there seems to be a big desire for notifications by an item's value (a fixed version of the old functionality)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -431,10 +431,6 @@ public class GroundItemsPlugin extends Plugin
 
 		// Cache colors
 		ImmutableList.Builder<PriceHighlight> priceCheckBuilder = ImmutableList.builder();
-		if (config.getHighlightOverValue() > 0)
-		{
-			priceCheckBuilder.add(new PriceHighlight(config.getHighlightOverValue(), config.highlightedColor()));
-		}
 
 		if (config.insaneValuePrice() > 0)
 		{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -242,12 +242,7 @@ public class GroundItemsPlugin extends Plugin
 			// The spawn time remains set at the oldest spawn
 		}
 
-		boolean shouldNotify = !config.onlyShowLoot() && config.highlightedColor().equals(getHighlighted(
-			new NamedQuantity(groundItem),
-			groundItem.getGePrice(),
-			groundItem.getHaPrice()));
-
-		if (config.notifyHighlightedDrops() && shouldNotify)
+		if (!config.onlyShowLoot())
 		{
 			notifyHighlightedItem(groundItem);
 		}
@@ -370,12 +365,7 @@ public class GroundItemsPlugin extends Plugin
 			{
 				groundItem.setLootType(lootType);
 
-				boolean shouldNotify = config.onlyShowLoot() && config.highlightedColor().equals(getHighlighted(
-					new NamedQuantity(groundItem),
-					groundItem.getGePrice(),
-					groundItem.getHaPrice()));
-
-				if (config.notifyHighlightedDrops() && shouldNotify)
+				if (config.onlyShowLoot())
 				{
 					notifyHighlightedItem(groundItem);
 				}
@@ -645,6 +635,16 @@ public class GroundItemsPlugin extends Plugin
 
 	private void notifyHighlightedItem(GroundItem item)
 	{
+		boolean shouldNotifyHighlighted = config.highlightedColor().equals(getHighlighted(
+			new NamedQuantity(item),
+			item.getGePrice(),
+			item.getHaPrice()));
+
+		if (!config.notifyHighlightedDrops() || !shouldNotifyHighlighted)
+		{
+			return;
+		}
+
 		final Player local = client.getLocalPlayer();
 		final StringBuilder notificationStringBuilder = new StringBuilder()
 			.append("[")
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -32,6 +32,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.inject.Provides;
 import java.awt.Color;
 import java.awt.Rectangle;
+import static java.lang.Boolean.FALSE;
 import static java.lang.Boolean.TRUE;
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


Adam-:
```
		ValueCalculationMode mode = config.valueCalculationMode();
		for (Map.Entry<Integer, Color> entry : priceChecks.entrySet())
		{
			switch (mode)
			{
				case GE:
					if (gePrice > entry.getKey())
					{
						return entry.getValue();
					}
					break;
				case HA:
					if (haPrice > entry.getKey())
					{
						return entry.getValue();
					}
					break;
				default: // case HIGHEST
					if (gePrice > entry.getKey() || haPrice > entry.getKey())
					{
						return entry.getValue();
					}
					break;
			}
		}
```
		
Reduces to:

```
final int price = getValueByMode(gePrice, haPrice);
for (Map.Entry<Integer, Color> entry : priceChecks.entrySet())
{
  if (price > entry.getKey()) 
  {
    return entry.getValue();
  }
}
```

deathbeam:
Closing #10961 by completely removing highlight > value option is really dumb way to do it. I don't think that this option should be removed. The whole description of why you want to change this can be fixed simply with something like this: <https://github.com/runelite/runelite/pull/10876>

Hydrox6:
Gonna copy what was said in Discord.

> People obviously want notifications by value
> I'm not fixing 10961 by removing it btw, I'm removing a feature that's only real use is to abuse the highlight system and replacing it with something that actually works. 
> As long as `Highlight > value` is in the client, Highlights are just another tier of price value, presented as something separate from low->insane, when in reality it just sits above insane

Alexsuperfly:
this would also close #9241

loldudester:
Needs rebase, but otherwise 100% a great idea that nobody in discord disagrees with at all

Nightfirecat:
The only part of this PR I dislike is the addition of `Highlight > Value`. I would prefer to see an enum for notification threshold, with values corresponding to our value tiers (Low/Medium/High/Insane). It seems to me there would be no reason to notify for items which weren't already highlighted in some way.

Hydrox6:
I was on the fence about how this should be done tbh, I'll update it sometime this morning

Hydrox6:
I've updated the PR to use the named tiers. Notification checking is still done off price, since otherwise setting it to `Medium` wouldn't notify if you got an `Insane` drop, which is pretty silly. 

I'm not sure on the naming of the option (`Notify >= Tier`), or the placement of the conversion function (currently it's inside the Enum, but it could also be inside GroundItemsPlugin)

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -631,12 +615,26 @@ public void onFocusChanged(FocusChanged focusChanged)
...
-		if (!config.notifyHighlightedDrops() || !shouldNotifyHighlighted)
+		boolean shouldNotifyTier = getValueByMode(item.getGePrice(), item.getHaPrice()) > config.notifyTier().getValueFromTier(config) &&
+			FALSE.equals(hiddenItems.getUnchecked(new NamedQuantity(item)));
+
+		String dropType;
+		if (config.notifyHighlightedDrops() && shouldNotifyHighlighted)
+		{
+			dropType = "highlighted";
+		}
+		else if (config.notifyTier() != HighlightTier.OFF && shouldNotifyTier)
```
Nightfirecat, line 633: Isn't `config.notifyTier() == HighlightTier.OFF` already implicitly covered by the `getValueFromTier()` method call in `shouldNotifyTier`?

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -669,6 +669,19 @@ private void notifyHighlightedItem(GroundItem item)
 		notifier.notify(notificationStringBuilder.toString());
 	}
 
+	private int getValueByMode(int gePrice, int haPrice)
```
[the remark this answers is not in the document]
Hydrox6: It's in this commit since before, the code wasn't duplicated.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/config/HighlightTier.java
```diff
@@ -0,0 +1,53 @@
...
+
+public enum HighlightTier
+{
+	OFF,
+	LOW,
+	MEDIUM,
+	HIGH,
+	INSANE;
+
+	public long getValueFromTier(GroundItemsConfig config)
```
deathbeam, line 37: I would change this to return Long and then null-check and mark the method nullable, Long.MAX_VALUE is pretty ugly way to determine if highlight tier is off (or simply check if highlight tier is OFF and then skip)

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -645,11 +614,36 @@ public void onFocusChanged(FocusChanged focusChanged)
 
 	private void notifyHighlightedItem(GroundItem item)
 	{
+		boolean shouldNotifyHighlighted = config.notifyHighlightedDrops() &&
```
deathbeam, line 617: final

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -645,11 +614,36 @@ public void onFocusChanged(FocusChanged focusChanged)
...
+			config.highlightedColor().equals(getHighlighted(
+				new NamedQuantity(item),
+				item.getGePrice(),
+				item.getHaPrice()));
+
+		boolean shouldNotifyTier = getValueByMode(item.getGePrice(), item.getHaPrice()) > config.notifyTier().getValueFromTier(config) &&
```
deathbeam, line 623: final

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -645,11 +614,36 @@ public void onFocusChanged(FocusChanged focusChanged)
...
+			FALSE.equals(hiddenItems.getUnchecked(new NamedQuantity(item)));
+
+		String dropType;
```
deathbeam, line 626: final

PR #11016, opened 2020-03-15
friendslist: update titles using onScriptPostFired rather than onGameTick

melkypie:


runelite-client/src/main/java/net/runelite/client/plugins/friendlist/FriendListPlugin.java
```diff
@@ -62,44 +63,52 @@ protected void shutDown()
...
+					"/" +
+					limit +
+					")";
 
-			setIgnoreListTitle(title);
+				setIgnoreListTitle(title);
+			}
 		}
+
+
```
Nightfirecat: There's already extra whitespace at the end of the method before, we definitely don't need more of it. Remove these empty lines.
```suggestion
```

PR #11017, opened 2020-03-15
chatcommands: Don't remove spaces before cleared word

JZomerlei:
I can only check Ubuntu and Windows which both act the same where
`word1 word2 word3   word4   ` becomes `word1 word2 word3   ` 
(Keeps the trailing spaces after the word)

Using [this stack overflow thread]( https://stackoverflow.com/questions/228477/how-do-i-programmatically-determine-operating-system-in-java) as how to determine the OS


JZomerlei:
Forgot to add this addresses the reopening of https://github.com/runelite/runelite/issues/6929

JZomerlei:
Removed the OS specific stuff, just closer matches windows/unix default behavior which is the one in the first comment about this PR

PR #11022, opened 2020-03-16
WIP: Move screen capturing to ImageCapture class

Nightfirecat:
This (currently WIP) PR will move all code relating to screen capturing to `ImageCapture`, drastically simplifying the process of capturing screenshots by offloading the burden of canvas composition. The intended use case is accept methods which are as similar to an `Overlay` render method as possible to remain familiar to users while still offering a wide array of functionality for screenshot-specific overlays/rendering.

The reasons this is still in a "WIP" state are as follows:

1. Figuring out the threading is causing me serious headaches. The code as written very nearly deadlocks because of its nested thread calls. In addition to that, this code *should* call `overlay.accept(graphics)` on the client thread, but doesn't currently since there is no `ClientThread#invokeAndWait()` method available to us, so changing it to `clientThread(() -> overlay.accept(graphics))` would cut short the execution of the overlay rendering and leave us with incomplete overlays drawn to the screenshot.
2. I don't know if this is the right approach to implement this feature. If there is some consensus we should be looking to write this differently, I'd like to cut my losses on this change early to change my approach.
3. The history is still messy. the new additions to `ImageCapture` along with the added `ImageCaptureOptions` class should be in a fully separate commit along with individual commits for changing the raids and screenshot plugins' code paths to call it. For the sake of getting code in front of reviewers I've opted to forego that step for now. (I'll have that cleaned up in a few days' time)
4. I haven't written the necessary javadoc for this new code yet

The code as written is able to correctly generate screenshots with multiple overlays: (actor hull & report button date overlays)
- [2020-03-15_23-23-40](https://user-images.githubusercontent.com/2199511/76730521-55580600-6753-11ea-9fec-9318c99a5345.png)

However, some amount of the time, these overlays will not be correct: (note the distorted actor hull)
- [2020-03-15_23-23-40(1)](https://user-images.githubusercontent.com/2199511/76730582-77ea1f00-6753-11ea-9f21-75d9aa84c02f.png)
- [2020-03-15_23-23-40(2)](https://user-images.githubusercontent.com/2199511/76730638-8e907600-6753-11ea-83d4-2b1a6df2eba7.png)
- [2020-03-15_23-23-41(1)](https://user-images.githubusercontent.com/2199511/76730651-96e8b100-6753-11ea-801e-a2d99b7fdb9d.png)
- [2020-03-15_23-23-46](https://user-images.githubusercontent.com/2199511/76730665-a10aaf80-6753-11ea-9a2a-3fc52eeba08c.png)


Nightfirecat:
I've split this into commits which, once completed, should resemble what this PR would ideally look like.

Nightfirecat:
This PR is now ready for review.

Nightfirecat:
I've rebased this PR and updated the ImageCapture change to not remove the old `takeScreenshot` methods, but instead deprecate them so plugin-hub authors will have time to adapt the new API.

Issue #10620, opened 2020-01-19
Boss respawn timers persist across world hops


Trevor159:
Related to #7196 but not a duplicate

Boss respawn timers should be cleared when a player logs out or hops worlds.

Nightfirecat:
If some implementation is added to resolve this issue, it should also aim to have some configuration so it can resolve #7196 as well.

Issue #7196, opened 2019-01-05
Preserve boss timers from other worlds


CONNELD:
**Is your feature request related to a problem? Please describe.**
When doing the same boss on multiple worlds, the timer for the last boss is lost then the boss is killed on another world.

**Describe the solution you'd like**
Display the timer and world when hovering over the boss timer icon.

Example:
![runelite_2019-01-04_20-54-03](https://user-images.githubusercontent.com/42870988/50719371-26cb9f00-1069-11e9-9c79-41b89d19698f.png)

PR #11031, opened 2020-03-17
Clear boss timer info box on log/hop + added configuration

DiscretePython:
Addresses issues #10620 and #7196.

Respawn timers are cleared from the infoBoxManager on log out or hopping.

Configuration was added to display an overlay instead of infoboxes. The overlay uses a session to track kills of bosses across different worlds. The session tracks kills for NPCs with a hashmap from the name to an ArrayList of RespawnTimers. An additional integer was added to the respawn timer class to store the world in which the timer is counting.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersConfig.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2017, Adam <Adam@sigterm.info>
+ * Copyright (c) 2018, Joris K <kjorisje@gmail.com>
+ * Copyright (c) 2018, Lasse <cronick@zytex.dk>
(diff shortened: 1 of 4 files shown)


dekvall:
I'm thinking that this would be better implemented as a tooltip on the infobox rather than a new overlay. Hide the infobox when no timer is active, and display the respawn of the current world normally. If the boss on the current world hasn't been killed yet, just display the infobox without any countdown so the player can access the tooltip in order to show the info for the rest of the worlds.

DiscretePython:
The last commit changes it to the system you describe. If a timer exists and the user is in a different world, the infobox is shown without the timer and the world where the boss hasn't respawned is shown in the infobox.

Adam-:
I don't see why this needs to be any more complex than just the one infobox with a tooltip showing the other worlds. I think you can store the world with the timer, don't show the text if the world mismatches, and then store a reference to the previous timer to use for making the tooltip.

PR #11035, opened 2020-03-18
Kourend library features

### LINKED ISSUE #10418: [Feature Request] Highlight Library books you're missing in the right-hand list

rfick:
Adds a visual indicator on screen pointing to the location of the book which the player is currently being asked to retrieve. This is the same indicator used by the cluescrolls plugin.

minimap

map

Additionally, this changes the colors of the books in the sidebar to the following:
- Green: requested book, book not in inventory
- Orange: not requested book, book not in inventory
- White: book in inventory

sidebar

Closes #10418

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPanel.java
@@ -52,6 +52,7 @@ class KourendLibraryPanel extends PluginPanel
 	private static final ImageIcon RESET_ICON;
 	private static final ImageIcon RESET_HOVER_ICON;
 
+	private final KourendLibraryPlugin plugin;
 	private final KourendLibraryConfig config;
 	private final Library library;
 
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPanel.java
```diff
@@ -152,6 +164,29 @@ void update()
 					e.getValue().setLocation("<html>" + locs.stream().collect(Collectors.joining("<br>")) + "</html>");
 				}
 			}
+
+			final ItemContainer itemContainer = client.getItemContainer(InventoryID.INVENTORY);
```
Nightfirecat: This should be run on the client thread. Also, the plugin already keeps track of books the player is holding--this doesn't need to be calculated again.

rnoennig:
Works for me as well, thanks for your work @rfick

PR #11043, opened 2020-03-19
chatcommands: add !session command

### LINKED ISSUE #11037: [QoL]Add a !online or !timeonline command, allowing for your total playtime for your current session to be broadcasted to other runelite users
Hello!

I'm sure a lot of players who use runelite also have pures, or other builds that they are frequently creating and spending many hours training to perfection. I know I do!

I'm asking if it's possible to create a command that would show your total playtime for your current session to other runelite users that would help you combat situations where you are going to either reset your aggression timers, or bank, and coming back to someone being in your spot, who have recently or just logged into your world.

Jagex changed how drops function a while back, and if you log into a world, you will not be able to see any drops for x amount of minutes before having been logged into said world, so a lot of people who log into worlds are having a hard time believing people who claim to have been there for many hours already, and this causes a lot of conflicts and arguments between players that is unnecessary and shouldn't be taking place to begin with.

I feel like this is very much a quality of life update, and I am sure a lot of people would appreciate an update like this coming to the client. The client is already great, and has many awesome features readily available, and that's awesome!

Thank you for reading this.

P.S.

I realize that you can't really hold jeopardy on a spot, or a world at all, but we have played this game for many years together, and there has always been a mutual respect for one another, and I think that adding this feature would just add another layer to that.

melkypie:
Adds a simple !session commmand that displays how long you have been logged in for.
Closes #11037


dekvall:
The purpose of #11037 is to broadcast your logged in time to other players. This will display the logged in time of the local player regardless of who sent the message, and that info is already available on the report button. You'd need to submit the logged in time like e.g. the duel arena command does for it to be available to others.

melkypie:
Now other players can also see the session time.

deathbeam:
Needs rebase

PR #11055, opened 2020-03-22
Add toggle between two keys to Key Remapping plugin

midasvo:
Toggling means to use a single key to active one selected key after the other.

There are two configurable keys in total, the key to activate the toggle is also configurable. 

Seemed like a good addition to the Key Remapping plugin. For example, in a large majority of use cases you are switching between two KeyEvents. If you could specify those and toggle between them using a single key it would save some effort. It's disabled by default but I think it would provide some QoL.

### DIFF
--- a/http-api/src/main/java/META-INF/MANIFEST.MF
+++ b/http-api/src/main/java/META-INF/MANIFEST.MF
@@ -1,0 +1,3 @@
+Manifest-Version: 1.0
+Main-Class: net.runelite.client.RuneLite
+
--- a/http-api/src/main/java/META-INF/MANIFEST.MF
+++ b/http-api/src/main/java/META-INF/MANIFEST.MF
@@ -1,3 +1,0 @@
-Manifest-Version: 1.0
-Main-Class: net.runelite.client.RuneLite
-
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -254,10 +254,10 @@ public interface KeyRemappingConfig extends Config
 	}
 
 	@ConfigItem(
-			position = 20,
-			keyName = "toggleBetween",
-			name = "Toggle between",
-			description = "Toggle between two keys"
+		position = 20,
+		keyName = "toggleBetween",
+		name = "Toggle between",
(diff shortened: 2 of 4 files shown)


midasvo:
Here's a pic of the configuration:
[https://i.imgur.com/QDef1oW.png](config)

Adam-:
It doesn't really make sense to remap two arbitrary keys with the key remapping plugin. There is a lot of native tooling and things like AHK which can do key remaps better than we can, and also support remapping >1 key.

The reason that this plugin exists is for context-aware key remaps, where it can conditionally remap based on the game state. This isn't doing that.

midasvo:
Thanks for responding to my PR, the first one so this is just something I thought of that I would personally use and might fit this plugin. 

I do have some comments on what you've said

> It doesn't really make sense to remap two arbitrary keys with the key remapping plugin. 

Being able to easily configure switching menu two menu items, it's usually prayer and inventory for me (ex demonic gorillas)

> There is a lot of native tooling and things like AHK which can do key remaps better than we can, and also support remapping >1 key.

That's fair, you could write a script to activate a KeyEvent or use ahk to macro to the other menu item / KeyEvent though that seems like a lot more effort. 

> The reason that this plugin exists is for context-aware key remaps, where it can conditionally remap based on the game state. This isn't doing that.

Then I totally missed that, I read the source code and it accounts for when you are chatting or using like a world map and stuff but it's not the focus. The goal seems to be to remap keys. 

Do you think this shouldn't even be in Runelites domain? Would a more specific external plugin be better? Like menu-item-toggler - and it would do just that, assign a key to togle between certain menu items? I'd prefer not to use other 3rd party stuff if my RL can do it so just figuring out what to do with this.

Adam-:
If you aren't aware, there is an in-game settings panel for remapping fkeys to different game tabs. I guess you could make a pluginhub plugin if you wanted.

midasvo:
> If you aren't aware, there is an in-game settings panel for remapping fkeys to different game tabs. I guess you could make a pluginhub plugin if you wanted.

You mean the keybinding menu right? blob:https://imgur.com/69a41bdf-7714-4e7c-a4e5-8c70760c4c54

I'm not aware it does the same as what my changes are doing. My changes were made because I'm lazy and I can't be bothered to click F1 for invent and then F2 when I need my prayer menu, I just want to click ` to switch between specifically those two. That's new because I genuinely can't find it.

PR #11058, opened 2020-03-23
Revert "opponentinfo: add opponent's opponent for npcs"

TheStonedTurtle:
This seems to be against the rules again
> Indicates which player an NPC is focused on

https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1

reverts commit 34f94e973c18627a3d7f20faf8160647b87c09c9.


Adam-:
I think this actually has been there the whole time, we just missed it before....

PR #11061, opened 2020-03-24
Make all labels and tooltips have a maximum width

TheStonedTurtle:
Labels default to 215px (via `PluginPanel.PANEL_WIDTH;`) and tooltips default to 500px.

This needs more testing as i could only think of so many places to look

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/FixedWidthLabelUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/FixedWidthLabelUI.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2020, TheStonedTurtle <https://github.com/TheStonedTurtle>
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Can you provide some example before/after screenshots?

TheStonedTurtle:
Here are some examples where the tooltips were over 500px before and are now being auto-wrapped.

image
image

I couldn't find any cases were a label is currently > 215px that would be effected by the LabelUI. I included it as it would allow for more complex dialog confirmation in the future, but it can be removed if deemed unnecessary.

The LabelUI would fix having a JLabel inside a JPanel when calling the confirmation dialog as seen in the below image.
image
image

Adam-:
This seems like a good idea for tooltips, and it looks like we manually wrapped at least one of these ourselves:

image

But I'm not sure we want labels to automatically wrap since they are usually manually built and we can make them wrap text already using html tags etc.

PR #11063, opened 2020-03-24
music, camera: volume & zoom slider tooltips

### LINKED ISSUE #11056: [QoL] Add numbers to the sliding scales in the settings.
The sliding scales include the Zoom, Brightness, Music, Sound Effect and Area Sounds and adding numbers to get those perfect levels would be nice.

seandewar:
This PR adds the following tooltips to the option sliders for the music and camera plugins:

music slider tooltips camera zoom slider tooltip

~~This PR also adds transparency to the music slider buttons when dragging, which mimics the behaviour of the zoom slider in the official client.~~
**Edit:** This feature is no longer a part of this PR.

Closes #11056 (minus the request for a tooltip for the brightness slider, which only has 4 states anyway).


runelite-client/src/main/java/net/runelite/client/plugins/music/MusicOverlay.java
```diff
@@ -0,0 +1,90 @@
...
+		for (MusicPlugin.MusicSlider slider : MusicPlugin.MusicSlider.values())
+		{
+			if (slider.getHandle() == null || slider.getHandle().isHidden())
+			{
+				return null;
+			}
+
+			final Rectangle bounds = slider.getHandle().getBounds();
+
+			if (bounds.getX() <= 0)
```
Nightfirecat: In what case is the slider handle nonnull and visible with x <= 0?

runelite-api/src/main/java/net/runelite/api/VarClientInt.java
```diff
@@ -52,6 +52,9 @@
 	 */
 	INPUT_TYPE(5),
 
+	CAMERA_ZOOM_1(73),
+	CAMERA_ZOOM_2(74),
```
Nightfirecat: What are these? Why is only one of them used? What are the expected values?

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -523,6 +523,7 @@
 	SEED_VAULT_ITEM_TEXT(WidgetID.SEED_VAULT_GROUP_ID, WidgetID.SeedVault.ITEM_TEXT),
 	SEED_VAULT_INVENTORY_ITEMS_CONTAINER(WidgetID.SEED_VAULT_INVENTORY_GROUP_ID, WidgetID.SeedVault.INVENTORY_ITEM_CONTAINER),
 
+	OPTIONS_CAMERA_ZOOM_SLIDER_BUTTON(WidgetID.OPTIONS_GROUP_ID, WidgetID.Options.CAMERA_ZOOM_SLIDER_BUTTON),
```
Nightfirecat: `OPTIONS_CAMERA_ZOOM_SLIDER_HANDLE` would be more in line with how we describe slider handles in, say, the music plugin.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraOverlay.java
```diff
@@ -0,0 +1,93 @@
...
+		final Widget slider = client.getWidget(WidgetInfo.OPTIONS_CAMERA_ZOOM_SLIDER_BUTTON);
+
+		if (slider == null || slider.isHidden())
+		{
+			return null;
+		}
+
+		final Rectangle bounds = slider.getBounds();
+
+		if (bounds.getX() <= 0)
```
Nightfirecat: Same question here: when is this condition ever true?

runelite-api/src/main/java/net/runelite/api/VarClientInt.java
seandewar: They seem to be the internal camera zoom level. I'm unsure what the difference between both vars are, though - both always seem to be equal, even with the vanilla behaviour.

ScriptID.java:163 also just refers to both of these vars as being the zoom level.
CameraPlugin.keyReleased() doesn't give much idea either, as it simply calls the script with the same value for its two arguments (which presumably map to vars 73 & 74).

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraOverlay.java
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		final Point mousePosition = client.getMouseCanvasPosition();
+
+		if (bounds.contains(mousePosition.getX(), mousePosition.getY()))
+		{
+			final int value = client.getVar(VarClientInt.CAMERA_ZOOM_1);
+			final int max = config.innerLimit() ? config.INNER_ZOOM_LIMIT : Constants.CLIENT_DEFAULT_INNER_ZOOM_LIMIT;
+
+			final StringBuilder sb = new StringBuilder("Camera Zoom: ");
```
abextm: You can just use string concatenation here. it will generate the same code in this case

seandewar:
I've managed to make the Music plugin's custom sliders work well with the suggested `setOnMouseRepeatListener()`-based approach, but I can't quite get the vanilla client's zoom slider to work the same way.

With the vanilla zoom slider, the script set using `setOnMouseRepeatListener()` stops being run when dragging the handle, even when the it is set on another widget such as its track instead. I can't quite figure out why this is happening (Jagex does something special here?), so for now I might settle with the overlay-based approach for the Camera plugin.

Any ideas?

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraOverlay.java
```diff
@@ -0,0 +1,83 @@
...
+
+		if (slider == null || slider.isHidden())
+		{
+			return null;
+		}
+
+		final Point mousePosition = client.getMouseCanvasPosition();
+
+		if (slider.getBounds().contains(mousePosition.getX(), mousePosition.getY()))
+		{
```
Nightfirecat: Extend this guard clause to include the bounds check.

```suggestion
		if (slider == null || slider.isHidden()
			|| !slider.getBounds().contains(client.getMouseCanvasPosition()))
		{
			return null;
		}
```

Then the rest of this tooltip generation can happen without being contained in any further nested blocks.

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraOverlay.java
```diff
@@ -0,0 +1,83 @@
...
+			final Widget fixedViewport = client.getWidget(WidgetInfo.FIXED_VIEWPORT);
+			final boolean isFixedMode = fixedViewport != null && !fixedViewport.isHidden();
+
+			final int value = client.getVar(isFixedMode ? VarClientInt.CAMERA_ZOOM_FIXED_VIEWPORT : VarClientInt.CAMERA_ZOOM_RESIZABLE_VIEWPORT);
```
abextm: this should either just chose one value and stick with it, or use the same logic the game uses <https://github.com/RuneStar/cs2-scripts/blob/master/scripts/%5Bproc,zoom_position_slider%5D.cs2#L8-L10>

runelite-api/src/main/java/net/runelite/api/Constants.java
```diff
@@ -52,11 +52,31 @@
...
 
+	/**
+	 * The width of the game viewport when running in fixed mode.
+	 */
+	public static final int GAME_FIXED_VIEWPORT_WIDTH = 512;
+
+	/**
+	 * The height of the game viewport when running in fixed mode.
+	 */
+	public static final int GAME_FIXED_VIEWPORT_HEIGHT = 334;
```
seandewar: I've prefixed these constants with `GAME_` to be consistent with the stuff already here for `GAME_FIXED_WIDTH` and such, but is the `CLIENT_` prefix be better for these instead? (such as in `CLIENT_DEFAULT_ZOOM`)

runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraOverlay.java
```diff
@@ -0,0 +1,78 @@
...
+		final Widget slider = client.getWidget(WidgetInfo.OPTIONS_CAMERA_ZOOM_SLIDER_HANDLE);
+		final Point mousePos = client.getMouseCanvasPosition();
+
+		if (slider == null || slider.isHidden() || !slider.getBounds().contains(mousePos.getX(), mousePos.getY()))
+		{
+			return null;
+		}
+
+		// Same logic used in scripts to determine if the game is in resizable mode or not
+		final boolean resizableMode = client.getViewportHeight() > Constants.GAME_FIXED_VIEWPORT_WIDTH;
```
abextm: you transcribed this wrong, its should be width > height, but that makes this dead code anyway (the cs2 has a bug apparently, as its different from what the engine uses). Just make it always use resizable always because it makes no difference to the calculated value as these vars are always in sync, and I don't want to copy a bug here and have it change later.

runelite-api/src/main/java/net/runelite/api/Constants.java
```diff
@@ -52,11 +52,31 @@
...
+
 	/**
 	 * The default camera zoom value.
 	 */
 	public static final int CLIENT_DEFAULT_ZOOM = 512;
 
+	/**
+	 * The default camera inner zoom limit.
+	 */
+	public static final int CLIENT_DEFAULT_INNER_ZOOM_LIMIT = 896;
```
abextm: these should be in the plugin because no other plugin needs this info

PR #11066, opened 2020-03-25
runenergy: update graceful recovery rate logic to reflect 9 Jan game update

### LINKED ISSUE #10577: Run Restoration with New Graceful Effect

seandewar:
Updates the energy recovery rate calculation to reflect the changes made
in the [9th January 2020 game update](https://secure.runescape.com/m=news/eating-in-the-bank?oldschool=1).

Each individual piece of graceful now boosts your recovery rate by 3%,
with an extra 1% for the top and legs. With the full set, an extra boost
of 10% is added, totalling 30%. Previously, there was no boost for
individual pieces, instead the full set was required for a flat boost of
30%.

**Edit:** Closes #10577

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
@@ -26,7 +26,10 @@ package net.runelite.client.plugins.runenergy;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
+import java.util.Arrays;
 import javax.inject.Inject;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.EquipmentInventorySlot;
@@ -93,6 +96,27 @@ public class RunEnergyPlugin extends Plugin
 		GRACEFUL_CAPE_13630, GRACEFUL_CAPE_13669, GRACEFUL_CAPE_13670, GRACEFUL_CAPE_21064, GRACEFUL_CAPE_21066,
 		AGILITY_CAPE, AGILITY_CAPET, MAX_CAPE
 	);
+
+	@RequiredArgsConstructor
+	@Getter
+	private enum GracefulEquipmentSlot
+	{
+		HEAD(EquipmentInventorySlot.HEAD.getSlotIdx(), ALL_GRACEFUL_HOODS, 3),
+		BODY(EquipmentInventorySlot.BODY.getSlotIdx(), ALL_GRACEFUL_TOPS, 4),
+		LEGS(EquipmentInventorySlot.LEGS.getSlotIdx(), ALL_GRACEFUL_LEGS, 4),
+		GLOVES(EquipmentInventorySlot.GLOVES.getSlotIdx(), ALL_GRACEFUL_GLOVES, 3),
+		BOOTS(EquipmentInventorySlot.BOOTS.getSlotIdx(), ALL_GRACEFUL_BOOTS, 3),
+		CAPE(EquipmentInventorySlot.CAPE.getSlotIdx(), ALL_GRACEFUL_CAPES, 3);
+
+		private final int index;
+		private final ImmutableSet<Integer> items;
+		private final int boost;
+
+		private static final int TOTAL_BOOSTS = Arrays.stream(values()).mapToInt(GracefulEquipmentSlot::getBoost).sum();
+	}
+
+	// Full set grants an extra 10% boost to recovery rate
+	private static final int GRACEFUL_FULL_SET_BOOST_BONUS = 10;
 
 	@Inject
 	private Client client;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,30 +210,42 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
+		int boost = 0;
+
+		for (int i = 0; i < items.length; ++i)
+		{
+			final ImmutableSet<Integer> gracefulForSlot = GRACEFUL_EQUIPMENT_SLOT_MAP.get(i);
+
+			if (gracefulForSlot != null && gracefulForSlot.contains(items[i].getId()))
+			{
+				// 4% boost for the top and legs; 3% for the others
+				boost += i == EquipmentInventorySlot.BODY.getSlotIdx() || i == EquipmentInventorySlot.LEGS.getSlotIdx() ? 4 : 3;
```
Nightfirecat: Instead of having hardcoded values here and below for the full set effect check, an enum with the slot index, set of graceful item IDs, and slot bonus could be defined which would give better definition to this data. This also would allow creation of some static `FULL_SET_BONUS` value that can be referenced below.

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,30 +223,40 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
-			ALL_GRACEFUL_TOPS.contains(items[EquipmentInventorySlot.BODY.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_LEGS.contains(items[EquipmentInventorySlot.LEGS.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_GLOVES.contains(items[EquipmentInventorySlot.GLOVES.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_BOOTS.contains(items[EquipmentInventorySlot.BOOTS.getSlotIdx()].getId()) &&
-			ALL_GRACEFUL_CAPES.contains(items[EquipmentInventorySlot.CAPE.getSlotIdx()].getId()));
+		int boost = 0;
+
+		for (final GracefulEquipmentSlot slot : GracefulEquipmentSlot.values())
+		{
+			final int wornItemId = items[slot.getIndex()].getId();
```
Nightfirecat: This can cause `ArrayIndexOutOfBoundsException` as `equipment.getItems()` is not guaranteed to have a length out to the highest equipment slot type's ID. Moving the `items.length > slot.getIndex()` check which is currently part of the if statement below this to the start of the block should correct that.

```
if (items.length <= slot.getIndex())
{
	continue;
}

// rest of method
```

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -199,30 +223,45 @@ String getEstimatedRunTimeRemaining(boolean inSeconds)
...
+
+		for (final GracefulEquipmentSlot slot : GracefulEquipmentSlot.values())
 		{
-			return false;
+			if (items.length <= slot.getIndex())
+			{
+				continue;
+			}
+
+			final int wornItemId = items[slot.getIndex()].getId();
```
Nightfirecat: Entries are not guaranteed to be nonnull. Change this to `final Item wornItem` and null-check it along with checking `slot.getItems().contains(wornItem.getId())` below.

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -94,6 +97,27 @@
...
+		LEGS(EquipmentInventorySlot.LEGS.getSlotIdx(), ALL_GRACEFUL_LEGS, 4),
+		GLOVES(EquipmentInventorySlot.GLOVES.getSlotIdx(), ALL_GRACEFUL_GLOVES, 3),
+		BOOTS(EquipmentInventorySlot.BOOTS.getSlotIdx(), ALL_GRACEFUL_BOOTS, 3),
+		CAPE(EquipmentInventorySlot.CAPE.getSlotIdx(), ALL_GRACEFUL_CAPES, 3);
+
+		private final int index;
+		private final ImmutableSet<Integer> items;
+		private final int boost;
+
+		private static final int TOTAL_BOOSTS = Arrays.stream(values()).mapToInt(x -> x.getBoost()).sum();
```
Adam-, line 115:
```suggestion
		private static final int TOTAL_BOOSTS = Arrays.stream(values()).mapToInt(GracefulEquipmentSlot::getBoost).sum();
```

PR #11094, opened 2020-03-26
grounditems: manually match item thresholds

abextm:
The regex used was explosive and wouldn't complete with large inputs that can appear when the user has a mildly corrupted config


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -163,7 +164,7 @@
 
 	@Getter
 	private final Map<GroundItem.GroundItemKey, GroundItem> collectedGroundItems = new LinkedHashMap<>();
-	private final Map<Integer, Color> priceChecks = new LinkedHashMap<>();
+	private Map<Integer, Color> priceChecks = ImmutableMap.of();
```
Adam-: Is order guaranteed for this? It requires entrySet() to be in-order of insertion

PR #11099, opened 2020-03-26
Added a menu entry swapper option for the Portal Nexus

HughRaynor:
Added a menu entry swapper option for the Portal Nexus, enabling left-click teleport menu, alongside shift / right-click saved destination teleport, which is not currently possible.


HughRaynor:
I understand that 'Teleport Menu' is the default left click if you don't have a favourite selected, but I feel it's preferable to have the best of both worlds here, being able to flexibly left quick to choose your destination, whilst maintaining quick and easy access to your favourite destination via shift/right-click.

Adam-:
This doesn't make any sense to me, who would want to swap to teleport menu if they have something selected as left click? Why not just configure the nexus to be left click teleport menu in-game?

Nightfirecat:
I believe this mostly applies when entering other player's POHs, as the default left click option is up to the POH owner.

raiyni:

PR #11108, opened 2020-03-27
Added population check before quick-hopping.

### LINKED ISSUE #11102: World Hopper Should Skip Full Worlds
When using the quick-hop functionality of the World Hopper plugin, full worlds are not skipped. The client tries to login to the full world, and if unsuccessful, the client logs-out the player. This is inconvenient.

Please add a toggle which will skip full, or near-full worlds while quick-hopping. This is similar to the already implemented 'Quick-hop out of dangerous worlds' toggle, but will require an additional check to the player-count when attempting to hop worlds. 

Alternatively, this functionality may be made default. When the login to a full world fails, perhaps the plugin could just try the next world, instead of the player being logged-out. I don't think this will work, though, since the failure directly results in being logged-out.

HughRaynor:
Fixes #11102

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -586,7 +586,7 @@ public class WorldHopperPlugin extends Plugin
 				}
 			}
 
-			//Avoid switching to near-max population worlds to prevent failure
+			// Avoid switching to near-max population worlds to prevent failure
 			if (world.getPlayers() >= (MAX_PLAYER_COUNT * MAX_HOP_TO_PERCENTAGE))
 			{
 				continue;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -102,7 +102,8 @@ public class WorldHopperPlugin extends Plugin
 {
 	private static final int REFRESH_THROTTLE = 60_000; // ms
 	private static final int TICK_THROTTLE = (int) Duration.ofMinutes(10).toMillis();
-	private static final int MAX_PLAYER_COUNT = 1950;
+	private static final int MAX_PLAYER_COUNT = 2000;
+	private static final double MAX_HOP_TO_PERCENTAGE = 0.975;
 
 	private static final int DISPLAY_SWITCHER_MAX_ATTEMPTS = 3;
 
@@ -586,7 +587,7 @@ public class WorldHopperPlugin extends Plugin
 			}
 
 			//Avoid switching to near-max population worlds to prevent failure
-			if (world.getPlayers() >= MAX_PLAYER_COUNT)
+			if (world.getPlayers() >= (MAX_PLAYER_COUNT * MAX_HOP_TO_PERCENTAGE))
 			{
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -584,6 +584,12 @@ private void hop(boolean previous)
 				}
 			}
 
+			//Avoid switching to near-max population worlds to prevent failure
+			if (world.getPlayers() >= 1950)
```
ypperlig: This should probably not be hardcoded as such.
Make it a final int and add it to the top of the class:
```java
private static final int MAX_PLAYER_COUNT = 1950;
```

```suggestion
			if (world.getPlayers() >= MAX_PLAYER_COUNT)
```

HughRaynor: Amended in the latest commit. Thanks.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -102,6 +102,7 @@
 {
 	private static final int REFRESH_THROTTLE = 60_000; // ms
 	private static final int TICK_THROTTLE = (int) Duration.ofMinutes(10).toMillis();
+	private static final int MAX_PLAYER_COUNT = 1950;
```
Nightfirecat: I would rather this name not be inaccurate. Instead, define this as `2000` since that is the real max value, and define another constant, `MAX_HOP_TO_PERCENTAGE = 0.975` and use that as the basis for comparison.

HughRaynor: Amended to match your preferences.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -584,6 +586,12 @@ private void hop(boolean previous)
 				}
 			}
 
+			//Avoid switching to near-max population worlds to prevent failure
```
Nightfirecat:
```suggestion
			// Avoid switching to near-max population worlds to prevent failure
```

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -102,6 +102,8 @@
 {
 	private static final int REFRESH_THROTTLE = 60_000; // ms
 	private static final int TICK_THROTTLE = (int) Duration.ofMinutes(10).toMillis();
+	private static final int MAX_PLAYER_COUNT = 2000;
```
Adam-, line 105: If you really want this to be expressed as a percentage of 2k you could just set it to `2000 * 0.975` (I don't see any reason to not do this calculation at compile time) but I would be find just with setting this to `1950`

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -584,6 +586,12 @@ private void hop(boolean previous)
 				}
 			}
 
+			// Avoid switching to near-max population worlds to prevent failure
+			if (world.getPlayers() >= (MAX_PLAYER_COUNT * MAX_HOP_TO_PERCENTAGE))
```
cman85, line 590: FWIW, this would entirely prevent users from hopping to W2 or W330. I don't think that's desirable behavior.

Adam-: This code is in the hop up/down stuff. It has nothing to do with general hopping?

cman85: ahh nice. Looks like I need to check more context before commenting, heh

Adam-:
added in 15b5dd02902b5f86440441009aa82e656458643f

PR #11126, opened 2020-03-28
OverlayRenderer: prevent moving DYNAMIC and TOOLTIP overlays

Alexsuperfly:
prevent moving DYNAMIC and TOOLTIP overlays as these are not intended to be able to be moved.  

This fixes an issue where the Mouse Tooltips plugins tooltip can be accidentally moved and cannot be returned to its normal position 

[video showing them being moved on release](https://streamable.com/v4xav)
getting your cursor into a position where its rendering and you could be dragging another overlay at the same time seemed to be the best way to "start it",  from there the tooltips bounding box would be visible in one of the snap corners and you could then drag it anywhere you wanted


deathbeam:
I cleaned up this a bit (reused the logic we use for getting correct overlay position from render method). Can you retest this @Alexsuperfly ?

Alexsuperfly:
works good DB

PR #11127, opened 2020-03-28
Add Plugin with timer and readiness indicator for nature rune chest thieving

### LINKED ISSUE #3924: Thieving Chest timers
When a thieving chest is looted a timer starts (time depends on chest), when the timer has expired the chest can be looted again. It's kind of annoying you can't actively see when the chest is available for thieving again.

It would be nice to have an overlay on the chest which shows the cooldown period visually.

There is a similar timer for hunter traps. This is what I think would be the best.

**Additional context**
N/A

travisfe:
Closes #3924

 Creates a new UI element that count down how long until the [nature rune theiving chests](https://oldschool.runescape.wiki/w/Chest_(nature_runes) can be looted again. Also shows a solid color when the chest is ready to be looted.

Both colors are customizable in the config


Nightfirecat:
This was added to the plugin hub in runelite/plugin-hub#181

Issue #11104, opened 2020-03-27
Jormungand's prison task location


Alexsuperfly:
user in discord claims that his `!task` command does not work on his basilisk task, I believe this is because the location is not whitelisted [here](https://github.com/runelite/runelite/blob/18a9a60b7d3b9bcfa55e89f364986628aefd020a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java#L184).
I was also able to get a [client log](https://github.com/runelite/runelite/files/4390994/client.4.log) from them

![](https://cdn.discordapp.com/attachments/442492468307427330/692900000459522068/9bbefdcbf426fb2e95cf16c0c03ed697.png)

![](https://cdn.discordapp.com/attachments/442492468307427330/692906932381417532/0b62ff6859294114d935e6e8ad607bc5.png)

thomaschow:
@Adam-  I'd like to take this task. I started looking at the cleint logs and I'm a bit skeptical the problem is in the whitelist as you mentioned: 

> I believe this is because the location is not whitelisted [here](https://github.com/runelite/runelite/blob/18a9a60b7d3b9bcfa55e89f364986628aefd020a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java#L184).

The client logs point to a socket timeout, and a subsequent NPE [here](https://github.com/runelite/runelite/blob/18a9a60b7d3b9bcfa55e89f364986628aefd020a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java#L784-L786). 

I wonder if [`task`](https://github.com/runelite/runelite/blob/18a9a60b7d3b9bcfa55e89f364986628aefd020a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java#L776) is set to null here, as a result of a null return value of the underlying GSON parser. 

Is it possible to get more information from the user on what the [http params](https://github.com/runelite/runelite/blob/18a9a60b7d3b9bcfa55e89f364986628aefd020a/http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java#L148) were submitted to get this Task? Alternatively, would you know how to reproduce this locally?

Alexsuperfly:
im not too sure the log is actually related, i kinda just included it for completeness. 
Looking back it, the NPE is logged after the timestamp in the picture showing him attempting to `!task`, so its likely some other one off that did that.  

[...]

PR #11128, opened 2020-03-28
Fix Jormungand's prison task location for the Slayer Plugin

thomaschow:
### Description

- Fix https://github.com/runelite/runelite/issues/11104. The solution is to add an entry to the `SlayerTaskPlugin`'s `Task` whitelist.

### Testing

- [x] Added unit tests.
- [x] I verified the behavior locally with the debugger.

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
@@ -708,6 +708,50 @@ public class SlayerPluginTest
 	}
 
 	@Test
+	public void testTaskLookup() throws IOException
+	{
+		net.runelite.http.api.chat.Task task = new net.runelite.http.api.chat.Task();
+		task.setTask("Abyssal demons");
+		task.setLocation("Abyss");
+		task.setAmount(42);
+		task.setInitialAmount(42);
+
+		when(slayerConfig.taskCommand()).thenReturn(true);
+		when(chatClient.getTask(anyString())).thenReturn(task);
+
+		ChatMessage setMessage = new ChatMessage();
+		setMessage.setType(ChatMessageType.PUBLICCHAT);
+		setMessage.setName("Adam");
+		setMessage.setMessageNode(mock(MessageNode.class));
+
+		slayerPlugin.taskLookup(setMessage, "!task");
+
+		verify(chatMessageManager).update(any(MessageNode.class));
+	}
+
+	@Test
+	public void testTaskLookupInvalid() throws IOException
+	{
+		net.runelite.http.api.chat.Task task = new net.runelite.http.api.chat.Task();
+		task.setTask("task<");
+		task.setLocation("loc");
+		task.setAmount(42);
+		task.setInitialAmount(42);
+
+		when(slayerConfig.taskCommand()).thenReturn(true);
+		when(chatClient.getTask(anyString())).thenReturn(task);
+
+		ChatMessage chatMessage = new ChatMessage();
+		chatMessage.setType(ChatMessageType.PUBLICCHAT);
+		chatMessage.setName("Adam");
+		chatMessage.setMessageNode(mock(MessageNode.class));
+
+		slayerPlugin.taskLookup(chatMessage, "!task");
+
+		verify(chatMessageManager, never()).update(any(MessageNode.class));
+	}
+
+	@Test
 	public void testTaskLookupFromNPCChat() throws IOException
 	{
 		Widget npcDialog = mock(Widget.class);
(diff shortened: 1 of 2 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -86,6 +86,8 @@
 	private static final String TASK_KONAR_BOSS = "You're now assigned to bring balance to the Alchemical<br>Hydra 35 times. Your reward point tally is 724.";
 
 	private static final String TASK_EXISTING = "You're still hunting suqahs; you have 222 to go. Come<br>back when you've finished your task.";
+	private static final String TASK_EXISTING_2 = "You're still bringing balance to basilisks in the Jormungand's Prison, with 136 to go. Come back when you're finished.";
```
Nightfirecat: This should include the `<br>`s which exist in the actual text.
```suggestion
	private static final String TASK_EXISTING_2 = "You're still bringing balance to basilisks in the<br>Jormungand's Prison, with 136 to go. Come back when<br>you're finished.";
```

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -86,6 +86,8 @@
...
+	private static final String TASK_EXISTING_INVALID_LOC = "You're still bringing balance to basilisks in the Prison, with 136 to go. Come back when you're finished.";
```
Nightfirecat: Same here.
```suggestion
	private static final String TASK_EXISTING_INVALID_LOC = "You're still bringing balance to basilisks in the<br>Prison, with 136 to go. Come back when<br>you're finished.";
```

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -705,6 +707,60 @@ public void testCombatBraceletUpdate()
 		assertEquals(30, slayerPlugin.getAmount());
 	}
 
+	@Test
+	public void testTaskLookupFromNPCChat() throws IOException
```
Nightfirecat: Is this meant to throw IOException?

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
Nightfirecat: This var name should be changed as well. It should become `TASK_CHECKSLAYERGEM_KONAR_2` and move next to that line earlier in the file.

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -705,6 +707,60 @@ public void testCombatBraceletUpdate()
...
+		when(chatClient.getTask(anyString())).thenReturn(task);
+
+		ChatMessage setMessage = new ChatMessage();
+		setMessage.setType(ChatMessageType.PUBLICCHAT);
+		setMessage.setName("Adam");
+		setMessage.setMessageNode(mock(MessageNode.class));
+
+		slayerPlugin.taskLookup(setMessage, "!task");
+
+		verify(chatMessageManager).update(any(MessageNode.class));
```
Nightfirecat: Why is this validating that the chat message manager is making changes here? This is not testing the right behavior. Instead, this should be similar to `testCheckSlayerGemKonarTask()`.

thomaschow: I intended for this test to be similar to [`testTaskLookup`](https://github.com/runelite/runelite/pull/11128/files#diff-e1297a0ae6bdf66b5be1b855ea54e07bR765) in that it tests `taskLookup` does what's expected (update the chat message if we have a whitelisted task location). In this case I'm doing slightly more end-to-end testing by starting the test from the the NPC dialog, but I can remove the earlier "setup" code if we want to contain the test to just that one method.

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
thomaschow: This is to be consistent with https://github.com/runelite/runelite/pull/11128/files#diff-e1297a0ae6bdf66b5be1b855ea54e07bR765 since they are testing the same code. 

I can reorder the tests to make this a bit more clear.

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
thomaschow: @Nightfirecat would this be an acceptable test given my earlier comment?

Nightfirecat: That test is specififally verifying lookup. This one does not need to.

thomaschow: I'm happy to remove these tests altogether then; I'm simply adding an entry to the locations whitelist which is used in taskLookup

Nightfirecat: That seems like a good solution. We've got pretty sufficient tests for this code path in the plugin and just need to add the new location. :+1:

thomaschow:
ptal @Nightfirecat

PR #11132, opened 2020-03-29
Fixed typo in account plugin

TylerADavis:
There's a typo in the tooltip text for the login button from the account plugin. Namely it uses `login` instead of `log in`, and `login` is only a noun, not a verb. It reads to me like the verb is the intended usage here. This PR switches `login` and `logout` to `log in` and `log out`.

Article describing difference: https://grammarist.com/spelling/log-in-login/


TylerADavis:
Done

TylerADavis:
Does anything else need to be done before this is merged? Or who should I talk to about merging?

PR #11140, opened 2020-03-29
Toggleable prayer sounds

c4fe:
Allows prayer activation and deactivation sounds to be disabled in the Music plugin tab.


runelite-api/src/main/java/net/runelite/api/SoundEffectID.java
```diff
@@ -62,7 +62,33 @@
 	public final static int ZERO_DAMAGE_SPLAT = 511;
 	public final static int TAKE_DAMAGE_SPLAT = 510;
 	public final static int ATTACK_HIT = 2498;
-	public final static int PRAYER_ACTIVATE_VROOM = 2690;
+
+	public final static int PRAYER_ACTIVATE_2690 = 2690; //Thick Skin
```
Nightfirecat: Instead of naming each of these after the ID, name them after the prayer, eg. `PRAYER_ACTIVATE_THICK_SKIN`

runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
```diff
@@ -590,4 +621,15 @@ else if (source == null
 			areaSoundEffectPlayed.consume();
 		}
 	}
+
+	@Subscribe
+	public void onSoundEffectPlayed(SoundEffectPlayed soundEffectPlayed)
+	{
+		int soundId = soundEffectPlayed.getSoundId();
```
Nightfirecat: This value is used only once, and is clear without a variable name attached to it, so this should be passed directly to the `contains()` check instead.

PR #11142, opened 2020-03-29
Menu Entry Swapper: using bones on Phials or the altar is always at the top of the menu

pkhermouch:
When using bones on Phials to un-note bones, or using bones on an altar, always put these options at the top of the menu if the "Bury" plugin configuration is on. I was training Prayer this weekend, and was getting annoyed when I would try to perform these actions and players were getting in the way (the house party worlds are generally very busy).

Using bones on Phials with the Bury configuration disabled:
phials-bury-off

Using bones on Phials with the Bury configuration enabled:
phials-bury-on

Using bones on an altar with the Bury configuration disabled:
altar-bury-off

Using bones on an altar with the Bury configuration enabled:
altar-bury-on

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -750,9 +750,18 @@ public class MenuEntrySwapperPlugin extends Plugin
 				swap("teleport", option, target, index);
 			}
 		}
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I believe this would not be allowed [per Jagex's statement on third-party clients](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) as it would stray afoul the `Reorders or removes player-based options, such as ‘Trade with’` rule. This should also be unnecessary when using the Entity Hider plugin to hide other players.

Issue #11156, opened 2020-03-31
Loot Tracker - Remember collapse state when rebuilding panel


TheStonedTurtle:
When toggling between showing/hiding ignored items or ignoring a specific item the side panel is rebuilt and all previously collapsed boxes are re-expanded.

It would be nice if rebuilding in these scenarios remembered the collapse state of boxes.

![XOwHXCFEZs](https://user-images.githubusercontent.com/29030969/78055321-3b641900-7338-11ea-917c-0a1251a62cb1.gif)

PR #11163, opened 2020-04-01
Fix: Loot tracker boxes now remember their collapsed state after being rebuilt

kayvan61:
Fix for issue #11156 Edited the LootTrackerRecord class. Hopefully that is not outside the scope for that class.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
@@ -83,6 +83,7 @@ class LootTrackerBox extends JPanel
 	private long totalPrice;
 	private boolean hideIgnoredItems;
 	private BiConsumer<String, Boolean> onItemToggle;
+	private boolean isGroupBox;
 
 	LootTrackerBox(
 		final ItemManager itemManager,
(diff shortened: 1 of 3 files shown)


kayvan61:
Sorry about that. I'll fix the style issues.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -227,6 +229,13 @@ void rebuild()
 
 		validate();
 		repaint();
+		for (LootTrackerRecord rec: records)
+		{
+			if (rec.isShouldCollapseBox())
+			{
+				this.collapse();
```
Adam-, line 236: this can't be correct due to collapse re-looping records

kayvan61: Would adding a break here fix that issue? since once we find a collapsed record all the records for the box are updated to collapse so we can stop checking the remaining records.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
```diff
@@ -186,6 +190,7 @@ void addKill(final LootTrackerRecord record)
 		}
 
 		kills += record.getKills();
+		records.add(record);
```
Adam-, line 193: Is it possible to do this in a way which doesn't require storing all of the records on each box? Usually they get combined with exiting records which uses less memory, which I like.

kayvan61:
I think this has been resolved. regardless its super stale. Sorry for the waste of time.

PR #11170, opened 2020-04-02
Update Plugin Hub disclaimer label

### LINKED ISSUE #10997: Clarify Plugin Hub risk label
No, simply a misleading label in my opinion, a sentiment I've seen shared by others. Many people's first reaction when they see the big bold "**Use at your own risk!**" label is something like, "oh, this a cheat client now? I won't use these because I don't want to get banned". Heard this first hand when my friend discovered the feature, and I had to explain to him that everything goes through the same checks to comply with Jagex's rules. I've seen the same thing asked about on reddit a few times. I know it may seem obvious from a developer's standpoint who knows RL's firm stance on this, but that bold label even caught me off guard and I had to check out the documentation to confirm these were still complying.

I'm not sure what's ideal, but some ideas:
- Replace bold "Use at your own risk" label with "Stability not guaranteed", or something along those lines. The current label makes it sound more seriously risky than it is.
- Add a sentence to the initial label saying "These plugins are equally reviewed to comply with Jagex's rules."

Every runelite user going to the plugin hub github/runelite github wiki and getting to the part that explains these plugins are also reviewed to comply. We know that isn't going to happen, and I bet some people are still refraining from using these plugins out of fear that they don't comply with Jagex's rules. I've also noticed people have asked this a few times in discord, so it would help reduce that as well.

**Additional context**
Talking about this panel in the plugin hub:
image

Matsyir:
Slight update to the plugin hub/external plugin disclaimer to clarify that the plugins are still verified not to be malicious or rule-breaking, while keeping the point that main RL developers don't maintain them, and only adding one line. As discussed in #10997, using Dekvall's suggested label. The issue hasn't had any updates since then, and his suggestion seemed to be well received, so I figure it was good.

Updated label screenshot
Closes #10997


oscarelgenio2:
como hago para usar el plugin hub? me sale los mismo

PR #11181, opened 2020-04-03
DPS counter change

QGoh:
Changes/Additions:
- show damage now shows damage + DPS as opposed to showing damage only
- reset option to reset DPS counter at the start of each new boss fight
- boss only option to only track DPS/damage on boss targets
- display tracker option to choose when to display tracker (always, in-combat, never)
- DPS counter pauses when player dies
- added some missing bosses
- changed the fight timer to always show on the overlay instead of being a tool-tip


Adam-:
This should probably be in at least 7 different commits, as it is hard to review mashups. Can you split it up?

QGoh:
> This should probably be in at least 7 different commits, as it is hard to review mashups. Can you split it up?

Sure, that shouldn't be a problem.

dekvall:
The changes have been split up in #11254

PR #11192, opened 2020-04-04
Screenshot plugin: fix screenshots being taken of player when setting not selected

### LINKED ISSUE #11184: Own deaths will be screenshotted if 'Screenshot Friend Deaths' is on and 'Screenshot Deaths' off
**Describe the bug**
Runelite will screenshot your deaths eventhough 'Screenshot Deaths' is off

**To Reproduce**
1. Go to 'Screenshot Settings'
2. Toggle 'Screenshot Friend Deaths' on and 'Screenshot Deaths' off

**Expected behavior**
Runelite will screenshot your deaths eventhough 'Screenshot Deaths' is off

**Environment:**
 - OS: Windows
 - RuneLite version: 1.6.10.1
 - Launcher version: 1.6.0

akhtar-u:
closes #11184

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -249,7 +249,7 @@ public class ScreenshotPlugin extends Plugin
 		{
 			takeScreenshot("Death", "Deaths");
 		}
-		else if ((player.isClanMember() || player.isFriend()) && config.screenshotFriendDeath() && player.getCanvasTilePoly() != null)
+		else if ((player != client.getLocalPlayer() || player.isClanMember() || player.isFriend()) && config.screenshotFriendDeath() && player.getCanvasTilePoly() != null)
 		{
 			takeScreenshot("Death " + player.getName(), "Deaths");
 		}


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -249,7 +249,7 @@ public void onPlayerDeath(PlayerDeath playerDeath)
 		{
 			takeScreenshot("Death", "Deaths");
 		}
-		else if ((player.isClanMember() || player.isFriend()) && config.screenshotFriendDeath() && player.getCanvasTilePoly() != null)
+		else if ((player != client.getLocalPlayer() || player.isClanMember() || player.isFriend()) && config.screenshotFriendDeath() && player.getCanvasTilePoly() != null)
```
Adam-, line 252: This logic is incorrect; the reason the bug happens is due to `isFriend()` returning `true` for the local player. This is a behavior of the underlying Jagex client (I think so that you can see your own chat with chat set to friends?). It may be worth adding a test for the issue described into `ScreenshotPluginTest`.

akhtar-u: Would comparing names work?

Nightfirecat: You don't need to compare names. Right now, the issue with your code is that `player != client.getLocalPlayer() || ... || player.isFriend()` is always true.

Adam-:
fixed in edcfeb25fdf99d47d6fce8b9b7c4eaa30f85b687

PR #11036, opened 2020-03-18
Revert Shift Anti-Drag and Disable in PvP Scenarios

jcwhisman:
The changes made reflect the exact same sentiments stated by @dekvall  [in PR #10870](https://github.com/runelite/runelite/pull/10870).

The following features have been reverted/added to the anti-drag plugin:
- **Shift Anti-Drag** plugin has been reverted back to original **Anti Drag**
- _Shift Only_ anti-drag still remains an option.
- Any form of anti-drag is disabled in PvP scenarios. (Detected by the spec orb enabling/disabling). Including in PvP worlds and wilderness.
- Anti drag is enabled for banking.

With the addition of removing all forms of anti-drag for PvP scenarios, I believe this sufficiently solves the issue/uproar in the PvP community that caused the plugin to get reverted in the first place.

Example on a PvP world:
rDJcg9x

PR #11193, opened 2020-04-04
Add toggle for player indicator in PvP areas

jdkleuver:
Add a toggle to show player indicators in the wilderness and other PvP areas.

Friends and clan members will take precedence over this, and they will appear with the appropriate colour. This will take precedence over non-clan members (if you have both enabled then it will use the pvp colour in pvp areas and the non clan member colour everywhere else)

The method used for detecting whether a player is in a PvP area is taken from PR #11036


Adam-:
Why is this useful?

jdkleuver:
The reason I made it was because I wanted to have it on in the wilderness so that it is easier to see when pkers log in around me. The problem was that whenever I leave the wilderness and go to a populated area like the GE my whole screen and minimal gets filled with player indicators. So it saves me from having to turn it off every time I leave the wilderness.

Nightfirecat:
This is kind of like an implicit PVP player warning feature, as it would allow players to know that player indicators are potential PKers. I'm against this feature, personally.

PR #11215, opened 2020-04-07
Add Menu Entry Swapper option for "Pet" stray dog instead of "Shoo-away"

bhoffman20:
Add option to menu entry swapper plugin to swap "Shoo-away" with "Pet" for the default left-click on Stray dogs.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -431,4 +431,14 @@ public interface MenuEntrySwapperConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "swapPetStray",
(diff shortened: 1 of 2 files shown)


Spedwards:
First off, why? This has to be the most pointless thing I've seen.

You're going to have better luck implementing this as an external plugin on the plugin hub.

bhoffman20:
As for why, I've seen people complain about the default option being "Shoo-away", and the change was very minor, and easy to add to an existing plugin. I can't argue that it isn't pointless, but I think people would like it. It's not game-breaking, and has no impact on anyone except the people that care to see the option changed. 

An external plugin wouldn't seem worth it for a single menu entry change.

Nightfirecat:
I think this is a candidate to add to [the Menu Swapper Extended plugin-hub plugin](https://github.com/raiyni/menu-swapper-extended) rather than the main client.

bhoffman20:
I didn't know there was already a plugin for extending the Menu Entry Swapper. I agree that makes more sense for this option, so I will close this PR and open one there.

PR #11218, opened 2020-04-07
runelite-client: add ClientShutdown event

abextm:
Requires !107

This should hopefully make the client not corrupt it's cache randomly, and prevents the config sets from racing shutdown

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
+++ b/runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
@@ -106,7 +106,7 @@ public class EventBus
 			builder.putAll(subscribers);
 		}
 
-		builder.orderValuesBy(Comparator.comparing(Subscriber::getPriority)
+		builder.orderValuesBy(Comparator.comparing(Subscriber::getPriority).reversed()
 			.thenComparing(s -> s.object.getClass().getName()));
 
 		for (Class<?> clazz = object.getClass(); clazz != null; clazz = clazz.getSuperclass())
--- a/runelite-client/src/main/java/net/runelite/client/events/ClientShutdown.java
+++ b/runelite-client/src/main/java/net/runelite/client/events/ClientShutdown.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2020 Abex
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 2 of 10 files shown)


runelite-client/src/main/java/net/runelite/client/events/ClientShutdown.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Value
+@Slf4j
+public class ClientShutdown
```
deathbeam, line 37: Can't this smply be singleton? You are creating it only once ever during the whole lifecycle of the app

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -317,14 +326,76 @@ public void init(final RuneLite runelite) throws Exception
...
+							log.warn("Unexpected exception occurred while check for confirm required", e);
+						}
+					}
+
+					if (result == JOptionPane.OK_OPTION)
+					{
+						saveClientBoundsConfig();
+						ClientShutdown csev = new ClientShutdown();
+						eventBus.post(csev);
+						new Thread(() ->
```
deathbeam, line 358: I dont really like the complexity of this, cant you at least move it to separate method? Its very annoying to read.

Also the method do not needs Runelite as param anymore right

runelite-client/src/main/java/net/runelite/client/events/ClientShutdown.java
abextm: I could but I don't see a reason to have it sitting on the heap for the entire life of the application when it can be created on demand

deathbeam: Well you can make singleton initialization lazy :d but i guess.

abextm: That sounds like you have just invented what I already have, but with a race condition

runelite-client/src/main/java/net/runelite/client/eventbus/EventBus.java
```diff
@@ -106,7 +106,7 @@ public synchronized void register(@Nonnull final Object object)
 			builder.putAll(subscribers);
 		}
 
-		builder.orderValuesBy(Comparator.comparing(Subscriber::getPriority)
+		builder.orderValuesBy(Comparator.comparing(Subscriber::getPriority).reversed()
```
Adam-, line 109: This probably breaks https://github.com/runelite/runelite/commit/9a2872eacfd929cbed22bac1a6bc8835bde7175f

runelite-client/src/main/java/net/runelite/client/events/ClientShutdown.java
```diff
@@ -0,0 +1,71 @@
...
+
+	public void waitForAllConsumers(Duration totalTimeout)
+	{
+		long deadline = System.nanoTime() + totalTimeout.toNanos();
+		for (Future<?> task; (task = tasks.poll()) != null; )
+		{
+			long timeout = deadline - System.nanoTime();
+			if (timeout < 0)
+			{
+				log.warn("Timed out waiting for task completion");
```
Adam-, line 54: is this supposed to exit here?

PR #11227, opened 2020-04-08
Add level and XP difference fields in skill calculator

Xmaxer:
I often have to calculate the difference in XP manually when trying to perform more complex, situational calculations myself. I noticed there were similar PR's a while back such as this one: #9283 
 which was a response to #9260 but from reading the replies it was never merged and didn't have functionality to edit the fields as you can the other fields.

Difference changes update the target, and vice versa. Maybe a more appropriate name is due rather than 'Level difference' and 'Experience difference' but I just called it as is. Others I've seen were 'Remaining levels' and 'Remaining Experience' which may also be suitable.

ezgif-7-caf93fff8a5c

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -79,6 +79,9 @@ class SkillCalculator extends JPanel
 	private int currentXP = Experience.getXpForLevel(currentLevel);
 	private int targetLevel = currentLevel + 1;
 	private int targetXP = Experience.getXpForLevel(targetLevel);
+	private int levelDifference = targetLevel - currentLevel;
+	private int xpDifference = Experience.getXpForLevel(currentLevel + levelDifference) - Experience.getXpForLevel(currentLevel);
+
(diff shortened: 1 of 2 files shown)


Adam-:
I've never understood why this is useful. If you know you want +x levels or +y experience, just add it to the current experience? It doesn't seem to me like it would be worth the loss in screen space since the panel is already fairly packed as it is.

Xmaxer:
For me it's more to do with the automatic difference calculation that happens between the current and target XP. Which I sometimes do need for other manual calculations, and for me personally just feels like one of those things I should have automatically. I've asked a few others in my cc about it, and there are some in similar scenarios and of course some who would never have had use for this. Also the only reason it's possible to input into the difference fields is to complement how the other fields work, so it feels more 'complete'.

All in all I can understand why some perceive it as a useless feature, and I am open to opinions. It's a small feature in the end that could maybe be implemented another way or not at all. 

Thanks for the response.

Nightfirecat:
If we are to merge it (which I don't know that we're interested in adding this feature--some discussion may be needed to get a consensus on this?) we would want the remaining XP/level fields to be disabled per discussion in #9283.

Nightfirecat:
Remaining XP was added in #11509

Issue #11154, opened 2020-03-31
Skill calculator - price for the quantity required


tvaidotas:
If possible it would be nice to see what is the total price for the amount of items required to reach the selected level. 

For example if going from 80 prayer to 99 prayer you select the altar and then click on dragon bones, it requires 23k bones you then need to go and check how much are the bones going for and then calculate it manually. 

Where if you select the dragon bones it would then expand or dispplay in some way what is the total price against the current price times the amount required.

[https://imgur.com/LnHyUfC](potential placement for the total price)

jonathanjyun:
@Adam- you can assign this issue to me.

iawhd:
Buy limits might also be useful here.

PR #11245, opened 2020-04-10
Merge for feature request issue #11154

jonathanjyun:
Added functionality to automatically calculate price for required quantity of items in Skill Calculator.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
@@ -42,10 +42,8 @@ import javax.swing.Box;
 import javax.swing.JCheckBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
-import net.runelite.api.Client;
-import net.runelite.api.Experience;
-import net.runelite.api.Skill;
-import net.runelite.api.VarPlayer;
+
+import net.runelite.api.*;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.skillcalculator.beans.SkillData;
@@ -56,6 +54,9 @@ import net.runelite.client.ui.DynamicGridLayout;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.ui.PluginPanel;
 import net.runelite.client.ui.components.IconTextField;
+import net.runelite.client.util.AsyncBufferedImage;
+import net.runelite.http.api.item.ItemPrice;
+import net.runelite.http.api.item.ItemStats;
 
 class SkillCalculator extends JPanel
 {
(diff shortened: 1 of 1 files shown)


Adam-:
This is making some fairly large assumptions of 1) the skill is buyable (most aren't - some even are gathering skills in which case this really doesn't apply at all) 2) you are buying these items on only 1 account and only in the GE and only for the duration of the buy limit. I'm not sure assuming either of these really makes sense for the skillcalc?

jonathanjyun:
1. For the skills that aren't buyable I didn't display any costs, it only applies to those that are buyable on the GE.
2. How would I address this concern? I'm not too sure about how to go about it.

Overall, are you saying that this feature request is unnecessary?

Adam-:
I don't see anything in the code which differentiates based on buyable skills or not?

jonathanjyun:
On line 379, I check whether an item is buyable or not. From my understanding, skills that have no buyable items are un-buyable skills, so that line wouldn't change the existing behavior of un-buyable items (and thus un-buyable skills if they only require un-buyable items).

jonathanjyun:
Screen Shot 2020-04-13 at 12 41 17 PM
Adam, here's a screenshot for reference as to how my updates handle buyable items (i.e. sandstone) vs non-buyable items (i.e. gem rocks). As you can see, non-buyable items have the original behavior.

Adam-:
As a counter-example, woodcutting has "buyable" items (logs), yet woodcutting is not a buyable skill.

jonathanjyun:
I see what you mean now. I'll include the relevant updates and commit for further feedback. Thanks!

jonathanjyun:
Before I commit my changes, I want to confirm the categorization of my skills are correct, since some of them are a little more complicated?

For buyable skills I have: magic, prayer, herblore, firemaking, construction, hunter, cooking, farming, crafting, smithing, fletching. 

For unbuyable skills I have: attack, mining, woodcutting, defence, ranged, hitpoints, agility, strength, slayer, fishing, runecraft, thieving.

Belieal:
I don't think hunter can be considered a buyable skill.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -42,10 +42,8 @@
 import javax.swing.JCheckBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
-import net.runelite.api.Client;
-import net.runelite.api.Experience;
-import net.runelite.api.Skill;
-import net.runelite.api.VarPlayer;
+
+import net.runelite.api.*;
```
Nightfirecat, line 46: [Our code conventions are to avoid star imports.](https://github.com/runelite/runelite/wiki/Code-Conventions) Revert this change.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -453,6 +487,61 @@ public void focusLost(FocusEvent e)
 		};
 	}
 
+	private boolean isBuyableSkill(final Skill skill)
+	{
+		switch (skill)
```
Nightfirecat, line 492: Don't have a separate body for each case, stack cases which have the same body. Eg:

```java
switch (skill)
{
	case ATTACK:
	case STRENGTH:
	case DEFENCE:
		return false;
	case CONSTRUCTION:
	case HERBLORE:
	case SMITHING:
		return true;
}
```

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -345,7 +344,16 @@ private void calculate()
...
+			if (isBuyableSkill(currentSkill))
+			{
+				slot.setText("<html>Lvl. " + action.getLevel() + " (" + formatXPActionString(xp, actionCount, "exp) - ") + formatXPActionCostString(actionCount, action.getName()));
+
+			}
+			else
+			{
+				slot.setText("Lvl. " + action.getLevel() + " (" + formatXPActionString(xp, actionCount, "exp) - "));
+
+			}
```
Nightfirecat, line 356: This can use some de-duplication.
```suggestion
			String slotText = "Lvl. " + action.getLevel() + " (" + formatXPActionString(xp, actionCount, "exp) - "));

			if (isBuyableSkill(currentSkill))
			{
				slotText = "<html>" + slotText + formatXPActionCostString(actionCount, action.getName());
			}

			slot.setText(slotText);
```

PR #11252, opened 2020-04-11
hitsplats: add missing multihitsplat varients

Trevor159:
I have not tested much, will test more but this should be fairly straightforward. The hitsplats were taken from the cache.

I didn't think this was worth putting in the javadoc but

The cyan hitsplats are used at the nightmare and verzik p1

The yellow hitsplats are use on the nightmare pillars

I have no idea where the white/orange hitsplats are used or if there are any usages I missed for the above two.


Trevor159:
I tested this at verzik and zalcano and it seems to work fine.

PR #11254, opened 2020-04-12
DPS Counter Changes

QGoh:
Changes/Additions:
- show damage now shows damage + DPS as opposed to showing damage only
- reset option to reset DPS counter at the start of each new boss fight
- boss only option to only track DPS/damage on boss targets (not sure how useful this will be because if you're in a party it only works if everyone has this option ticked on)
- choose when to display DPS tracker (always, in-combat, never)
- DPS counter resets when player dies
- added some missing bosses
- changed the fight timer to always show on the overlay instead of only on mouseover
- track damage dealt to each unique target


Spedwards:
So you're just abandoning #11181?

QGoh:
Ya, sorry, new to version control and I didn't know how to split up a commit like Adam asked so I just made a new fork.

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -181,14 +220,21 @@ public void onHitsplatApplied(HitsplatApplied hitsplatApplied)
 					specialCounterUpdate.setMemberId(localMember.getMemberId());
 					wsClient.send(specialCounterUpdate);
 				}
+				// damage to specific targets
+				if (dpsConfig.targetDamage()) {
+					if (!targets.containsKey(npcName)) {
+						targets.put(npcName, hitsplat.getAmount());
+					} else {
```
Spedwards, line 227: Need to change these to meet the style guide.

```suggestion
				if (dpsConfig.targetDamage())
				{
					if (!targets.containsKey(npcName))
					{
						targets.put(npcName, hitsplat.getAmount());
					}
					else
					{
```

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsOverlay.java
```diff
@@ -24,22 +24,23 @@
  */
 package net.runelite.client.plugins.dpscounter;
 
-import java.awt.Dimension;
-import java.awt.FontMetrics;
-import java.awt.Graphics2D;
+import java.awt.*;
```
Spedwards, line 27: Should probably change these back to individual imports as well.

Adam-:
You shouldn't be creating new PRs if you have to rebase your changes, but instead just force push to the original PR. This is super annoying because then I lose the context and original comments from the original PR. The history of this PR is not much better than the last one; over half of the commits are just trying to fix the previous commits. I recommend you do a rebase and try and consolidate them logically into the original commits.

It would also be useful if you wrote commit messages describing what is being changed and perhaps why. For example a commit called `timer change` with no other description or context has no meaning to me. What timer is changing? How is it changing? Why is it changing? 

If you are unwilling to learn git enough to send a well-formed patch set I would instead recommend just identifying one change you want to make and send only that and then I can look at that individually.

TheStonedTurtle:
Here's a blog post about git messages that is passed around the discord pretty commonly, I would recommend reading it.

https://chris.beams.io/posts/git-commit/

Issue #11092, opened 2020-03-26
Motherlode Plugin widget extension:


Pingvinszar2:
1) Ability to replace Ore names with Ore icons and Gem names with Gem icons.
2) Ability to add/remove ores.
3) Ability to add/remove gems.
4) Ability to monitor ores/hour by type (eg. Coal/hour, Gold/Hour, etc.)
5) Ability to monitor gems/hour.

PR #11261, opened 2020-04-13
Pull request for feature-request #11092

Tyler-Tran:
Decided to tackle requests 4 and 5 in #11092 but decided that gems/hr is usually going to be very low and not as useful. It seemed as though it might be more cluttered to add gems, so I only added ores. I plan on tackling the other parts of the request but wanted to separate the pull requests for them. Attached are screenshots of the feature:
ss1
ss2

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
@@ -201,7 +201,6 @@ public class MotherlodeSession
 	public void resetRecent()
 	{
 		recentPayDirtMined = null;
-
 		recentMined = 0;
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
@@ -38,7 +38,7 @@ public class MotherlodeSession
 {
 	private static final Duration HOUR = Duration.ofHours(1);
 
-	private int perHour;
+	private int payDirtPerHour;
 
 	private Instant lastPayDirtMined;
 	private int totalMined;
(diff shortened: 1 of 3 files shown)


Hydrox6:
In future, please give PRs a descriptive name that says what it does; "Pull request for feature" is not descriptive in and of itself.
Also, put `Closes #` with the issue number that is closed by the PR, so GitHub will automatically close the issue

Tyler-Tran:
Thanks, will keep that in mind for the future.

makingcomment:
A config option for whether to display this extra information would be nice.

Tyler-Tran:
The original request wanted to do configure for each ore. Would it be better to do 2 configs for ores for total and hourly, or have configs for each ore? I think that having a toggle for each ore might make the config cluttered.

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
```diff
@@ -108,25 +114,64 @@ void incrementGemFound(int gemID)
...
 	{
+		Instant now = Instant.now();
+		Duration timeSinceStart;
+
 		switch (item)
 		{
 			case ItemID.GOLDEN_NUGGET:
 				nuggetsFound += count;
+
+				timeSinceStart = Duration.between(recentPayDirtMined, now);
```
Adam-, line 125: Can't `recentPayDirtMined` be null here?

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
```diff
@@ -108,25 +114,64 @@ void incrementGemFound(int gemID)
...
+				if (!timeSinceStart.isZero())
+				{
+					nuggetsPerHour = (int) ((double) nuggetsFound * (double) HOUR.toMillis() / (double) timeSinceStart.toMillis());
```
Adam-, line 128: Maybe move this calculation to a common method since it is duplicated 6 times

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
```diff
@@ -159,9 +204,39 @@ public void resetRecent()
 		recentMined = 0;
 	}
 
-	public int getPerHour()
+	public int getPayDirtPerHour()
```
Adam-, line 207: You can just use lombok Getters instead of manually writing our accessors like this

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
Tyler-Tran: I believe it can potentially be null, but it would also be better for the time since start to be at the time of mining for better numbers. However its probably necessary to make a check in each case to prevent the null case

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
Tyler-Tran: Yea I can probably make a function for this

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSession.java
Tyler-Tran: Will look into this

ivanfeli:
Wouldn't it look better as: 

> Coal: 15 (245/hr)

 instead of separate rows?

PR #11268, opened 2020-04-14
Randomize player color

siddhu33:
Provide random colours for any player seen ( up to a defined limit of 2500 to avoid overloading any caches )

Tested in local build
g2

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
@@ -71,7 +71,7 @@ public class PlayerIndicatorsService
 	public void forEachPlayer(final BiConsumer<Player, Color> consumer)
 	{
 		if (!config.highlightOwnPlayer() && !config.drawClanMemberNames()
-				&& !config.highlightFriends() && !config.highlightNonClanMembers())
+			&& !config.highlightFriends() && !config.highlightNonClanMembers())
 		{
 			return;
 		}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
@@ -30,12 +30,36 @@ import javax.inject.Inject;
 import javax.inject.Singleton;
 import net.runelite.api.Client;
 import net.runelite.api.Player;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.Random;
 
 @Singleton
 public class PlayerIndicatorsService
 {
 	private final Client client;
 	private final PlayerIndicatorsConfig config;
+
+	private final int MAX_SIZE = 2500;
+	private final ConcurrentHashMap<String, Color> colorMap = new ConcurrentHashMap<>();
+	private final Random random = new Random();
+
+	private Color randomColor( Player player )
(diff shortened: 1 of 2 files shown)


deathbeam:
Why would you randomize the color? That do not makes sense to me, at least funtionality wise. I would use same thing as party pugin is using, e.g create the color from hash from name

siddhu33:
I am using the ColorUtil class now to hash the color from the name given. I think its just a fun little option tbh, you could maybe use it for an actual reason, but the idea was just because it would look cool.

runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
```diff
@@ -30,13 +30,37 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 
 @Singleton
 public class PlayerIndicatorsService
 {
 	private final Client client;
 	private final PlayerIndicatorsConfig config;
 
+	private final int MAX_SIZE = 2500;
+	private final ConcurrentHashMap<String, Color> colorMap = new ConcurrentHashMap<>();
```
Nightfirecat, line 44: A `LoadingCache` seems like a better structure to be used here, as you can set a `maximumSize` on that, rather than manually removing elements once a map reaches max size.

deathbeam: This do not needs to use cache at all anymore since its now coloring based on name hash, so this should just be removed

siddhu33: Alright then, will remove it, checked the framerate, didn't seem to budge much with / without it, so cheers

siddhu33:
@deathbeam Removed the caching now.

runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
```diff
@@ -24,19 +24,26 @@
...
-import java.util.function.BiConsumer;
-import javax.inject.Inject;
-import javax.inject.Singleton;
 import net.runelite.api.Client;
 import net.runelite.api.Player;
+import net.runelite.client.util.ColorUtil;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.awt.*;
```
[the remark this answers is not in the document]
Nightfirecat: Beyond that, the imports block should look like it did before, with all imports in a single block with no line breaks between any lines.

runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsService.java
```diff
@@ -86,6 +93,10 @@ else if (config.highlightNonClanMembers() && !isClanMember)
 			{
 				consumer.accept(player, config.getNonClanMemberColor());
 			}
+			else if ( config.randomizeColor() )
```
Nightfirecat, line 96:
```suggestion
			else if (config.randomizeColor())
```

Nightfirecat: Also, this behavior seems very weird. If a player were to have only friends selected to be highlighted, enabling this config would cause all non-friend players to be highlighted using a random color, which seems wrong.

PR #11276, opened 2020-04-15
Change Farming growing progress colour

pretsell:
I think that green Complete and green In-progress colours can be hard to distinguish at a glance what has finished growing.
So cheap dirty win, swap Diseased to red and set Growing to yellow.

Could add additional In-progress colour into runelite-client/src/main/java/net/runelite/client/ui/ColorScheme.java to achieve same end result.

Thanks.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CropState.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CropState.java
@@ -34,8 +34,8 @@ import net.runelite.client.ui.ColorScheme;
 public enum CropState
 {
 	HARVESTABLE(ColorScheme.PROGRESS_COMPLETE_COLOR),
-	GROWING(ColorScheme.PROGRESS_COMPLETE_COLOR),
-	DISEASED(ColorScheme.PROGRESS_INPROGRESS_COLOR),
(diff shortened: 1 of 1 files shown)


raiyni:
Diseased and dead aren't the same thing, thus they shouldn't be the same color.

pretsell:
agreed; but then why should Doing and Done be the same colour?

raiyni:
It's not, it's orange? `GROWING(ColorScheme.PROGRESS_INPROGRESS_COLOR)`

`public static final Color PROGRESS_INPROGRESS_COLOR = new Color(230, 150, 30);`

pretsell:
Where did you find that? my branch/PR?
Or in master runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CropState.java?

raiyni:
Meh, I read the diff wrong. The diseased should maybe be yellow instead. But I don't think it matters because you are looking at the progress???

pretsell:
I just think traffic light colouring system would be more fitting.
green = done/success
yellow = in progress/pending
red = stopped/error

What colour do you think would fit diseased best, to distinguish it from the existing? bright orange or yellow / super bright green? I did have a bit of play around before posting but nothing grabbed me.

PR #11282, opened 2020-04-16
Fix shift-modifier for deposit box after new interface update

haakonrp:
Deposit box interface update broke shift-modifiers for deposit boxes.


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -319,7 +319,7 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 		// Swap to shift-click deposit behavior
 		// Deposit- op 2 is the current withdraw amount 1/5/10/x
```
Nightfirecat, line 320: This comment must be updated with this code.

PR #11304, opened 2020-04-17
fixes mouse tooltip item value color not matching ground item value color

### LINKED ISSUE #11277: Mouse tooltip item value color does not match ground item value color
**Describe the bug**
When hovering over an item on the ground, the color of the mouse tooltip name of the item does not match the color of the text of the ground item. This only seems to happen on untradable items from my small amount of testing with feathers and cannonballs, though it is a small sample size.

**Screenshots**
NQcjIRbL5J
CyuvFuUbOP
My ground item settings for reference:
RuneLite_Zp40O7J4IQ

**Environment:**
 - OS: Windows 10
 - RuneLite version: 1.6.10.1
 - Launcher version: 1.6.0

rnoennig:
Fixes #11277

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -54,12 +54,9 @@ import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.api.ItemComposition;
 import net.runelite.api.ItemID;
-import net.runelite.api.ItemLayer;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
-import net.runelite.api.Node;
 import net.runelite.api.Player;
-import net.runelite.api.Scene;
 import net.runelite.api.Tile;
 import net.runelite.api.TileItem;
 import net.runelite.api.coords.WorldPoint;
@@ -406,6 +403,7 @@ public class GroundItemsPlugin extends Plugin
 			.tradeable(itemComposition.isTradeable())
 			.lootType(dropped ? LootType.DROPPED : LootType.UNKNOWN)
 			.spawnTime(Instant.now())
+			.stackable(itemComposition.isStackable())
 			.build();
 
 
@@ -482,40 +480,22 @@ public class GroundItemsPlugin extends Plugin
 				return;
 			}
 
-			int itemId = event.getIdentifier();
-			Scene scene = client.getScene();
-			Tile tile = scene.getTiles()[client.getPlane()][event.getActionParam0()][event.getActionParam1()];
-			ItemLayer itemLayer = tile.getItemLayer();
-
-			if (itemLayer == null)
-			{
-				return;
-			}
+			final int itemId = event.getIdentifier();
+			final int sceneX = event.getActionParam0();
+			final int sceneY = event.getActionParam1();
 
 			MenuEntry[] menuEntries = client.getMenuEntries();
 			MenuEntry lastEntry = menuEntries[menuEntries.length - 1];
 
-			int quantity = 1;
-			Node current = itemLayer.getBottom();
+			final WorldPoint worldPoint = WorldPoint.fromScene(client, sceneX, sceneY, client.getPlane());
+			GroundItem.GroundItemKey groundItemKey = new GroundItem.GroundItemKey(itemId, worldPoint);
+			GroundItem groundItem = collectedGroundItems.get(groundItemKey);
+			int quantity = groundItem.getQuantity();
 
-			while (current instanceof TileItem)
-			{
-				TileItem item = (TileItem) current;
-				if (item.getId() == itemId)
-				{
-					quantity = item.getQuantity();
-				}
-				current = current.getNext();
-			}
-
-			final ItemComposition itemComposition = itemManager.getItemComposition(itemId);
-			final int realItemId = itemComposition.getNote() != -1 ? itemComposition.getLinkedNoteId() : itemComposition.getId();
-			final int itemPrice = itemManager.getItemPrice(realItemId);
-			final int price = itemPrice <= 0 ? itemComposition.getPrice() : itemPrice;
-			final int haPrice = Math.round(itemComposition.getPrice() * Constants.HIGH_ALCHEMY_MULTIPLIER) * quantity;
-			final int gePrice = quantity * price;
-			final Color hidden = getHidden(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice, itemComposition.isTradeable());
-			final Color highlighted = getHighlighted(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice);
+			final int gePrice = groundItem.getGePrice();
+			final int haPrice = groundItem.getHaPrice();
+			final Color hidden = getHidden(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice, groundItem.isTradeable());
+			final Color highlighted = getHighlighted(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice);
 			final Color color = getItemColor(highlighted, hidden);
 			final boolean canBeRecolored = highlighted != null || (hidden != null && config.recolorMenuHiddenItems());
 
@@ -549,7 +529,7 @@ public class GroundItemsPlugin extends Plugin
 				}
 			}
 
-			if (config.showMenuItemQuantities() && itemComposition.isStackable() && quantity > 1)
+			if (config.showMenuItemQuantities() && groundItem.isStackable() && quantity > 1)
 			{
 				lastEntry.setTarget(lastEntry.getTarget() + " (" + quantity + ")");
 			}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
@@ -50,6 +50,7 @@ class GroundItem
 	private LootType lootType;
 	@Nullable
 	private Instant spawnTime;
+	private boolean stackable;
 
 	int getHaPrice()
 	{


raiyni:
You need to fix your imports according to our style guide. https://github.com/runelite/runelite/wiki/Code-Conventions

raiyni:
My previous comment still applies.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -495,31 +495,18 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
-			final int itemPrice = itemManager.getItemPrice(realItemId);
-			final int price = itemPrice <= 0 ? itemComposition.getPrice() : itemPrice;
-			final int haPrice = Math.round(itemComposition.getPrice() * Constants.HIGH_ALCHEMY_MULTIPLIER) * quantity;
-			final int gePrice = quantity * price;
-			final Color hidden = getHidden(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice, itemComposition.isTradeable());
-			final Color highlighted = getHighlighted(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice);
+			final int itemPrice = groundItem.getGePrice();
+			final int haPrice = groundItem.getHaPrice();
+			final int gePrice = itemPrice <= 0 ? haPrice : itemPrice;
+			final Color hidden = getHidden(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice, groundItem.isTradeable());
```
Adam-, line 505: I don't understand why you have removed quantity from the color calculation here? Also, If we want to make this the same as the overlay color, can't we just copy the logic from the overlay? Which is:
```
			final Color highlighted = plugin.getHighlighted(new NamedQuantity(item), item.getGePrice(), item.getHaPrice());
			final Color hidden = plugin.getHidden(new NamedQuantity(item), item.getGePrice(), item.getHaPrice(), item.isTradeable());
```

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -495,31 +495,18 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
-			final int realItemId = itemComposition.getNote() != -1 ? itemComposition.getLinkedNoteId() : itemComposition.getId();
-			final int itemPrice = itemManager.getItemPrice(realItemId);
-			final int price = itemPrice <= 0 ? itemComposition.getPrice() : itemPrice;
-			final int haPrice = Math.round(itemComposition.getPrice() * Constants.HIGH_ALCHEMY_MULTIPLIER) * quantity;
-			final int gePrice = quantity * price;
-			final Color hidden = getHidden(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice, itemComposition.isTradeable());
-			final Color highlighted = getHighlighted(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice);
+			final int itemPrice = groundItem.getGePrice();
+			final int haPrice = groundItem.getHaPrice();
+			final int gePrice = itemPrice <= 0 ? haPrice : itemPrice;
```
Adam-, line 504: Why is this here? The overlay doesn't have that:

```java
		for (GroundItem item : groundItemList)
		{
			final LocalPoint groundPoint = LocalPoint.fromWorld(client, item.getLocation());

			if (groundPoint == null || localLocation.distanceTo(groundPoint) > MAX_DISTANCE
				|| (onlyShowLoot && !item.isMine()))
			{
				continue;
			}

			final Color highlighted = plugin.getHighlighted(new NamedQuantity(item), item.getGePrice(), item.getHaPrice());
```

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -495,31 +495,18 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			final Color hidden = getHidden(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice, groundItem.isTradeable());
+			final Color highlighted = getHighlighted(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice);
 			final Color color = getItemColor(highlighted, hidden);
-			final boolean canBeRecolored = highlighted != null || (hidden != null && config.recolorMenuHiddenItems());
```
Adam-, line 520: Why are you deleting this feature? It looks unrelated.

rnoennig: Thanks for looking into this. This condition prevents that the computed color is actually applied if the value doesn't match any configured value and the default color should be applied.
Can you clarify which feature is actually removed by this edit? Thanks.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
rnoennig: I assume you mean the ternary operator that uses the HA price when GE is 0? If not please clarify what you mean by "this".
I tried to preserve the existing logic as best as I could. It's possible that this can be removed, I'll test it and report back.

Please compare with the version from branch master:
https://github.com/runelite/runelite/blob/1c9c71ef7c93ccf4c479b7137635e5e266d07788/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java#L514

rnoennig: This price defaulting is indeed not necessary. I removed it.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
rnoennig: I didn't notice the feature "recolorMenuHiddenItems". I brought it back and edited this pull request.

Nightfirecat:
It would help a lot for reviewing to squash this into two commits, one which fixes each of the issues you described in the issue.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
Adam-: Why are you still removing this highlighted check? This will make `recolorMenuHiddenItems` apply to highlighted items, and also break items which are highlighted and hidden (which should show it as highlighted, as the overlay does)

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -495,31 +495,18 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
-			final int gePrice = quantity * price;
-			final Color hidden = getHidden(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice, itemComposition.isTradeable());
-			final Color highlighted = getHighlighted(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice);
+			final int gePrice = groundItem.getGePrice();
+			final int haPrice = groundItem.getHaPrice();
+			final Color hidden = getHidden(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice, groundItem.isTradeable());
+			final Color highlighted = getHighlighted(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice);
 			final Color color = getItemColor(highlighted, hidden);
-			final boolean canBeRecolored = highlighted != null || (hidden != null && config.recolorMenuHiddenItems());
+			final boolean shouldNotRecolorHiddenMenuItem = hidden != null && !config.recolorMenuHiddenItems();
```
Adam-, line 507: In general avoid double-negatives as it is hard to follow for me (eg `!shouldNot...)`

Adam-:
Unrelated but it may be worth also writing a test for some of this highlight/hide/coloring behavior since it is not so straight forward

rnoennig:
> Unrelated but it may be worth also writing a test for some of this highlight/hide/coloring behavior since it is not so straight forward

Good idea, I'll write a test for the features you mentioned and what was described in the original bug report.

rnoennig:
> It would help a lot for reviewing to squash this into two commits, one which fixes each of the issues you described in the issue.

@Nightfirecat can I still do this within this PR or do I need to create a new one?

dekvall:
You should do it within this PR.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -495,31 +495,18 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
-			final Color highlighted = getHighlighted(new NamedQuantity(itemComposition.getName(), quantity), gePrice, haPrice);
+			final int gePrice = groundItem.getGePrice();
+			final int haPrice = groundItem.getHaPrice();
+			final Color hidden = getHidden(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice, groundItem.isTradeable());
+			final Color highlighted = getHighlighted(new NamedQuantity(groundItem.getName(), quantity), gePrice, haPrice);
 			final Color color = getItemColor(highlighted, hidden);
 			final boolean canBeRecolored = highlighted != null || (hidden != null && config.recolorMenuHiddenItems());
 
-			if (color != null && canBeRecolored && !color.equals(config.defaultColor()))
+			if (color != null && canBeRecolored)
```
Adam-, line 509: Why is this being removed? It would cause the plugin to apply the default overlay color to the menus which it currently doesn't do. It seems unrelated to this bug at least. I think the reason this exists is because the overlay needs a color but the menus already have their own default color which is white.

runelite-client/src/test/java/net/runelite/client/plugins/grounditems/GroundItemsPluginTest.java
```diff
@@ -0,0 +1,239 @@
...
+	@Bind
+	private Notifier notifier;
+
+	@Mock
+	@Bind
+	private ScheduledExecutorService executor;
+
+	@Mock
+	@Bind
+	private RuneLiteConfig runeLiteConfig;
```
Adam-, line 113: This isn't injected into the plugin, so it probably doesn't need to be mock/bound?

runelite-client/src/test/java/net/runelite/client/plugins/grounditems/GroundItemsPluginTest.java
```diff
@@ -0,0 +1,239 @@
...
+		when(tileItem.getId()).thenReturn(itemId);
+		when(tile.getItemLayer()).thenReturn(itemLayer);
+		when(itemManager.getItemComposition(itemId)).thenReturn(itemComposition);
+		when(client.getLocalPlayer()).thenReturn(player);
+		when(player.getWorldLocation()).thenReturn(worldPoint);
+		when(tile.getWorldLocation()).thenReturn(worldPoint);
+		when(itemComposition.getNote()).thenReturn(-1);
+		when(client.getMenuEntries()).thenReturn(new MenuEntry[]{menuEntry});
+		when(client.getScene()).thenReturn(scene);
+		when(scene.getTiles()).thenReturn(new Tile[][][]{new Tile[][]{new Tile[]{tile}}});
```
Adam-, line 219: Instead of mocking all of this code, you can just remove it from the plugin, since it is unused

Adam-:
I ended up just fixing the last part myself. The test wasn't really what I had envisioned so I removed it.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
rnoennig: If I understood your comment correctly, the default color should only be used for the overlay and not for the menu entry? I recommend changing it so both match the configured colors(highlight, default, hidden). Otherwise I'd recommend changing the description of the color so that's it's clear that the default color doesn't change the default color of menu entries.

runelite-client/src/test/java/net/runelite/client/plugins/grounditems/GroundItemsPluginTest.java
rnoennig: I'm not sure what you mean. The 4 lines your highlighted are non-lenient stubs, if they aren't called during execution an error would be thrown. I assure you those lines are needed.

rnoennig:
> I ended up just fixing the last part myself. The test wasn't really what I had envisioned so I removed it.

@Adam- What had you envisioned?

runelite-client/src/test/java/net/runelite/client/plugins/grounditems/GroundItemsPluginTest.java
Adam-: If you look at the final commit I pushed, it removed the code which used this from the plugin, which makes it unnecessary.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
Adam-: This would cause a rather user visible behavior change, which generally if we do those (especially without a config option toggle) it upsets people. At the very least it is not related to the two bugs this is trying to fix though.

PR #11310, opened 2020-04-18
Add wise old man tracking to xp updater plugin

psikoi:
Adding automatic tracking on logout to my new app [Wise Old Man](https://wiseoldman.net/).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterPlugin.java
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018, Adam <Adam@sigterm.info>
  * Copyright (c) 2020, Alexsuperfly <alexsuperfly@users.noreply.github.com>
+ * Copyright (c) 2020, Psikoi <https://github.com/psikoi>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -43,7 +44,6 @@ import net.runelite.http.api.RuneLiteAPI;
 import okhttp3.Call;
 import okhttp3.Callback;
 import okhttp3.HttpUrl;
-import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 
@@ -108,7 +108,7 @@ public class XpUpdaterPlugin extends Plugin
 			if (Math.abs(totalXp - lastXp) > XP_THRESHOLD)
 			{
 				log.debug("Submitting update for {}", local.getName());
-				sendUpdateRequest(local.getName());
+				update(local.getName());
 				lastXp = totalXp;
 			}
 		}
@@ -124,10 +124,9 @@ public class XpUpdaterPlugin extends Plugin
 		}
 	}
 
-	private void sendUpdateRequest(String username)
+	private void update(String username)
 	{
 		String reformedUsername = username.replace(" ", "_");
-		OkHttpClient httpClient = RuneLiteAPI.CLIENT;
 
 		if (config.cml())
 		{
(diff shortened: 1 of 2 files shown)


psikoi:
@Adam- are you open to increasing the exp limit, like I mentioned on the description above?

while-loop:
Or have the threshold as a config w/ 10k as the lower limit? I like my stats frequently 😁.

psikoi:
Then everyone would set it to the minimum possible, as noone would have a reason to want a lower frequency.

psikoi:
Does anyone oppose me increasing the minimum exp limit to 30k or 50k?

makingcomment:
I oppose that. I like having all/most of my sessions logged, not just the long ones, but 30-50k xp could be more than an hour's worth of xp depending on the skill or the amount of effort applied.

If it's a change you want to make, I suggest putting it in a separate PR. The wise old man addition seems pretty uncontroversial, so hopefully it can be merged without issue, but the xp minimum could use some discussion.

psikoi:
> I oppose that. I like having all/most of my sessions logged, not just the long ones, but 30-50k xp could be more than an hour's worth of xp depending on the skill or the amount of effort applied.
> 
> If it's a change you want to make, I suggest putting it in a separate PR. The wise old man addition seems pretty uncontroversial, so hopefully it can be merged without issue, but the xp minimum could use some discussion.

Yeah of course you do, there's 0 downsides to players like yourself, but if it highly affects the server costs of a free project, then it's worth discussing. 

When I asked the question, it wasn't meant for runelite users, it was meant for the developers. Otherwise it's like asking the general population if they want their taxes raised.

makingcomment:
> Yeah of course you do, there's 0 downsides to players like yourself, but if it highly affects the server costs of a free project, then it's worth discussing.

I'm aware of the reason you might want to raise the threshold - those are why it's currently 10k xp instead of 1 xp. I'm sure the impact to server costs would be greatly reduced if the threshold were raised all the way to 1 million xp, but at that point there's little to no point to using the plugin. You have to find a balance where the plugin remains useful without overloading trackers' servers, and I think raising the threshold from where it is now moves away from that balance.

I think 10k is a happy medium - it's high enough that stats aren't submitted every time a users log in for 2 minutes but low enough that it doesn't limit tracking for slower skills like I mentioned in my last post.

Additionally, I think writing off the player perspective as you did is not productive. Players are ultimately the ones who use the plugin, so the fact that what you're proposing is something players might not like is worth noting and taking into consideration.

psikoi:
I'll leave the exp limit as it is, but will have to increase it in the future (for my website) if it causes issues.

Adam-:
We should probably de-duplicate this code now it is duplicated twice..

psikoi:
Do you mean the request part?

psikoi:
@Adam- I changed it, there's now a request builder for each platform, and only one generic "sendRequest" method. Is that what you were looking for?

psikoi:
Gotcha

psikoi:
@Nightfirecat can you check now? should be as you asked

psikoi:
Added "wom" abbreviation to the plugin tags

runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterPlugin.java
```diff
@@ -124,7 +125,7 @@ public void onGameTick(GameTick gameTick)
 		}
 	}
 
-	private void sendUpdateRequest(String username)
+	private void update(String username)
 	{
 		String reformedUsername = username.replace(" ", "_");
 		OkHttpClient httpClient = RuneLiteAPI.CLIENT;
```
Nightfirecat, line 133: This appears to be unused after applying the first commit.

runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterPlugin.java
```diff
@@ -166,6 +168,29 @@ private void update(String username)
...
+
+		if (config.wiseoldman())
+		{
+			HttpUrl url = new HttpUrl.Builder()
+					.scheme("https")
+					.host("wiseoldman.net")
+					.addPathSegment("api")
+					.addPathSegment("players")
+					.addPathSegment("track")
+					.build();
```
Nightfirecat, line 180: All of your chained calls here are indented one level too far.
```suggestion
				.scheme("https")
				.host("wiseoldman.net")
				.addPathSegment("api")
				.addPathSegment("players")
				.addPathSegment("track")
				.build();
```

PR #11317, opened 2020-04-19
Properly address Grand Exchange price as 'GE' abbreviation instead of 'EX'

Flightkick:
Grand Exchange is addressed as GE everywhere, EX is confusing and can be mistaken for EXperience.

This changes the label placed over items on the ground.


Flightkick:
> 
> 
> This should be also changed in ItemPricesOverlay and BankPlugin

Good point, missed those. Thanks!

Issue #9228, opened 2019-06-27
Improvements to Barbarian Assault timer


Chicagolight:
**Is your feature request related to a problem? Please describe.**
Barbarian assault requires juggling a number of conditions. It is easy to miss the timer and cost your team points. This problem stems from a lack of audio cues and an inconvenient display location.

**Describe the solution you'd like**
Support an audio cue for the last few seconds of the call timer similar to the sound that plays before potion expiration. Support an option to relocate the display closer to the inventory/combat style pane.

PR #11325, opened 2020-04-20
Barbarian Assault Audio Cue

davidhamilton532:
This adds an optional audio cue to BA that would trigger when the call to your teammates changes.
Additionally adds a configurable delay for how long after the call changes that the audio cue should trigger.

This partially covers #9228.


runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultConfig.java
```diff
@@ -35,17 +35,30 @@
...
 	default boolean showTimer()
 	{
 		return true;
 	}
 
+	@ConfigItem(
+		keyName = "callChangeAudioCue",
+		name = "Call Change Audio Cue",
+		description = "Play an audio cue when the call changes",
+		position = 2
```
Nightfirecat, line 50: I don't see a need to add `position`s just for this.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/Round.java
```diff
@@ -63,6 +63,6 @@ public long getRoundTime()
 
 	public long getTimeToChange()
 	{
-		return 30 + (Duration.between(Instant.now(), roundStartTime).getSeconds() % 30);
+		return 30 - (Duration.between(roundStartTime, Instant.now()).getSeconds() % 30);
```
Nightfirecat, line 66: Please don't change unrelated code.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -156,6 +159,25 @@ public void onWidgetLoaded(WidgetLoaded event)
 		}
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick tick)
+	{
+		if (config.callChangeAudioCue() && currentRound != null)
```
Nightfirecat, line 165: Performing a null check on `currentRound` is cheaper than a configuration check, so flip the order of these.
```suggestion
		if (currentRound != null && config.callChangeAudioCue())
```

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -156,6 +159,25 @@ public void onWidgetLoaded(WidgetLoaded event)
...
+		{
+			long timeToChange = currentRound.getTimeToChange();
+
+			if (timeToChange == 30 && !playingCue)
```
Nightfirecat, line 169: Since `30` is a constant that should be shared between this class and Round, make it a package-private constant and reference it throughout both classes.

Adam-:
We already have a notification system in the client which can play custom sounds amongst other things. Would it make sense to just use that instead?

davidhamilton532:
I've implemented changes based on @Nightfirecat 's feedback.

@Adam- I had first looked into doing this as a notification but was running into issues where it wouldn't trigger when I wanted it to based on my notification settings (i.e. it would only trigger when the window wasn't focused). I also wasn't aware that notifications could be set to play custom sounds. Though I agree doing it as a notification or event/listener might be better. Do you know of an example I can look at for a notification like you mentioned?

Nightfirecat:
You would need to turn on the "Send notifications when focused" setting under `RuneLite`.

davidhamilton532:
Right. I was trying to make it so the audio played no matter what, whether the window was focused or not.

davidhamilton532:
Missed changing the operator back to `+` in reverting the change in `Round`.

Nightfirecat:
That should be up to the user ideally. `Notify on Call Change` would be a perfectly acceptable feature which would tie in with our existing notification system that users already know to enable with the client focused when they want that.

davidhamilton532:
I've reworked it to a notification and actually like it this way better than I remember the first time I tried it as a notification.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -54,19 +56,21 @@
...
 	private static final String START_WAVE = "1";
 	private static final String ENDGAME_REWARD_NEEDLE_TEXT = "<br>5";
+	static final int CALL_DURATION = 30;
 
 	private Font font;
 	private Image clockImage;
 	private int inGameBit = 0;
 	private String currentWave = START_WAVE;
 	private GameTimer gameTime;
+	private boolean playingCue = false;
```
Nightfirecat, line 73: `false` is the default value for boolean fields, so it can be omitted.
```suggestion
	private boolean playingCue;
```

Nightfirecat: Also, this value should be reset to `false` if either the plugin or `config.notifyCallChange()` are disabled.

davidhamilton532:
Implemented changes from @Nightfirecat 's latest feedback.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -60,6 +61,7 @@
 )
 public class BarbarianAssaultPlugin extends Plugin
 {
+	static final String CONFIG_GROUP = "barbarianAssault";
```
Nightfirecat, line 64: Reference this constant from the config class as well instead of having it hard-coded there also.

runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
```diff
@@ -156,6 +180,25 @@ public void onWidgetLoaded(WidgetLoaded event)
 		}
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick tick)
+	{
+		if (currentRound != null && config.notifyCallChange() && currentRound.getRoundTime() >= 10)
```
Nightfirecat, line 186: What is `roundTime`'s value when you start a round? This code looks like it would fire a notification immediately when starting a round if the clock shows `0:30` when you begin, which would not be ideal.

PR #11337, opened 2020-04-21
Add option to use custom list of worlds to hop through.

1Defence:
Allows you to cycle through a user-defined list such as; 353,354,378 circularly instead of the entire list of worlds.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -36,6 +36,8 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -161,6 +163,10 @@ public class WorldHopperPlugin extends Plugin
 
 	private final Map<Integer, Integer> storedPings = new HashMap<>();
 
+	private List<World> worldCycleList = new ArrayList<>();
+
+	private final String NUMBER_COMMA_ONLY_REGEX = "((?<!^,)\\d+(,(?!$)|$))+";;
+
 	private final HotkeyListener previousKeyListener = new HotkeyListener(() -> config.previousKey())
 	{
 		@Override
@@ -227,6 +233,9 @@ public class WorldHopperPlugin extends Plugin
 
 		// populate initial world list
 		updateList();
+
+		//set initial world cycle if present
+		setCustomWorldCycle(config.customWorldCycle());
 	}
 
 	@Override
@@ -280,6 +289,25 @@ public class WorldHopperPlugin extends Plugin
 					panel.setFilterMode(config.subscriptionFilter());
 					updateList();
 					break;
+				case "customWorldCycle":
+					setCustomWorldCycle(event.getNewValue());
+					break;
+			}
+		}
+	}
+
+	private void setCustomWorldCycle(String worldList)
+	{
+		WorldResult worldResult = worldService.getWorlds();
+		worldCycleList.clear();
+		if(worldResult != null || worldList.isEmpty())
+		{
+			if (worldList.matches(NUMBER_COMMA_ONLY_REGEX))
+			{
+				int[] worldCycleListInt = Arrays.stream(config.customWorldCycle().split(",")).mapToInt(Integer::parseInt).filter((x -> (worldResult.findWorld(x) != null))).toArray();
+				//convert valid worlds from string to int
+				Arrays.stream(worldCycleListInt).forEach(x -> worldCycleList.add(worldResult.findWorld(x)));
+				//convert int array to world list
 			}
 		}
 	}
@@ -525,7 +553,8 @@ public class WorldHopperPlugin extends Plugin
 		currentWorldTypes.remove(WorldType.SKILL_TOTAL);
 		currentWorldTypes.remove(WorldType.LAST_MAN_STANDING);
 
-		List<World> worlds = worldResult.getWorlds();
+		//use world cycle list if present
+		List<World> worlds = !worldCycleList.isEmpty() ? worldCycleList : worldResult.getWorlds();
 
 		int worldIdx = worlds.indexOf(currentWorld);
 		int totalLevel = client.getTotalLevel();
@@ -541,18 +570,14 @@ public class WorldHopperPlugin extends Plugin
 			 */
 			if (previous)
 			{
-				worldIdx--;
-
-				if (worldIdx < 0)
+				if (--worldIdx < 0)
 				{
 					worldIdx = worlds.size() - 1;
 				}
 			}
 			else
 			{
-				worldIdx++;
-
-				if (worldIdx >= worlds.size())
+				if (++worldIdx >= worlds.size())
 				{
 					worldIdx = 0;
 				}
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -541,18 +570,14 @@ private void hop(boolean previous)
 			 */
 			if (previous)
 			{
-				worldIdx--;
-
-				if (worldIdx < 0)
+				if (--worldIdx < 0)
```
Nightfirecat, line 573: Please don't change unrelated code, especially if it only makes it harder to read.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -161,6 +163,10 @@
 
 	private final Map<Integer, Integer> storedPings = new HashMap<>();
 
+	private List<World> worldCycleList = new ArrayList<>();
+
+	private final String NUMBER_COMMA_ONLY_REGEX = "((?<!^,)\\d+(,(?!$)|$))+";;
```
Nightfirecat, line 168: Rather than using a new regex for splitting commas and checking numbers, use `Text.fromCSV` and filter any entries which aren't numbers.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,6 +289,25 @@ public void onConfigChanged(final ConfigChanged event)
...
+		WorldResult worldResult = worldService.getWorlds();
+		worldCycleList.clear();
+		if (worldResult != null || worldList.isEmpty())
+		{
+			if (worldList.matches(NUMBER_COMMA_ONLY_REGEX))
+			{
+				int[] worldCycleListInt = Arrays.stream(config.customWorldCycle().split(",")).mapToInt(Integer::parseInt).filter((x -> (worldResult.findWorld(x) != null))).toArray();
+				//convert valid worlds from string to int
+				Arrays.stream(worldCycleListInt).forEach(x -> worldCycleList.add(worldResult.findWorld(x)));
+				//convert int array to world list
```
Nightfirecat, line 310: Don't construct multiple streams to achieve this, join them and put each operation on a new line.

```
Arrays.stream(Text.fromCSV(config.customWorldCycle()))
	.mapToInt(s ->
	{
		try
		{
			return Integer.parseInt(s);
		}
		catch (NumberFormatException e)
		{
			return 0;
		}
	})
	.map(worldResult::findWorld)
	.filter(world -> world != null)
	.forEach(world -> worldCycleList.add(world));
```

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,6 +289,25 @@ public void onConfigChanged(final ConfigChanged event)
...
 					updateList();
 					break;
+				case "customWorldCycle":
+					setCustomWorldCycle(event.getNewValue());
+					break;
+			}
+		}
+	}
+
+	private void setCustomWorldCycle(String worldList)
```
Nightfirecat, line 299: What is `worldList` used for? It seems to me this function always receives `config.customWorldCycle()` as this value and it's not needed.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,6 +289,25 @@ public void onConfigChanged(final ConfigChanged event)
...
+	{
+		WorldResult worldResult = worldService.getWorlds();
+		worldCycleList.clear();
+		if (worldResult != null || worldList.isEmpty())
```
Nightfirecat, line 303: `worldList` being empty should result in some kind of guard clause instead this obtuse way of setting `worldCycleList` to an empty list by parsing an empty string to a list to be added.

1Defence: Fixed, this was supposed to be an AND operator checking both was said guard.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
1Defence: The parameter was passed under the assumption that the event fired prior to config receiving the change not after; As this isn't the case removed as redundant.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,10 +286,37 @@ public void onConfigChanged(final ConfigChanged event)
...
+						{
+							return Integer.parseInt(s);
+						}
+						catch (NumberFormatException e)
+						{
+							return 0;
+						}
+					})
+					.filter(world -> worldResult.findWorld(world) != null)
+					.forEach(world -> worldCycleList.add(worldResult.findWorld(world)));
```
Nightfirecat, line 316: You have one too many levels of indent here.
```suggestion
				.mapToInt(s ->
				{
					try
					{
						return Integer.parseInt(s);
					}
					catch (NumberFormatException e)
					{
						return 0;
					}
				})
				.filter(world -> worldResult.findWorld(world) != null)
				.forEach(world -> worldCycleList.add(worldResult.findWorld(world)));
```

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,10 +286,37 @@ public void onConfigChanged(final ConfigChanged event)
...
+					break;
 			}
 		}
 	}
 
+	private void setCustomWorldCycle()
+	{
+		String worldList = config.customWorldCycle();
+		WorldResult worldResult = worldService.getWorlds();
+		worldCycleList.clear();
```
Nightfirecat, line 300: Run as little code as you can early on. In this case, I would move the `!worldList.isEmpty()` check from the following `if` statement up here to become a guard clause.
```suggestion
		worldCycleList.clear();

		String worldList = config.customWorldCycle();
		if (worldList.isEmpty())
		{
			return;
		}
		
		WorldResult worldResult = worldService.getWorlds();
```

1Defence: For consistency, should I also have an early return on the following worldResult null check? its effectively the same as nothing else is checked past said point.

Nightfirecat: For the last block of code in a method, it matters less, but the way you've changed it is fine.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -280,10 +286,46 @@ public void onConfigChanged(final ConfigChanged event)
...
+					return Integer.parseInt(s);
+				}
+				catch (NumberFormatException e)
+				{
+					return 0;
+				}
+			})
+			.filter(world -> worldResult.findWorld(world) != null)
+			.forEach(world -> worldCycleList.add(worldResult.findWorld(world)));
+
```
Nightfirecat, line 326:
```suggestion
```

PR #11339, opened 2020-04-22
Gauntlet Loot automatic screenshots

### LINKED ISSUE #11335: Automatic Screenshot of Gauntlet loot
Gauntlet loot doesn't automatically screenshot with the Screenshot plugin, it isn't a generic Boss Kill format in the chatbox and there is no existing custom handling for Gauntlet loot scenarios

We'd need to add custom handling to the Screenshot plugin to capture screenshots when viewing Gauntlet loot

Based on [ScreenshotPlugin.java](https://github.com/runelite/runelite/blob/19412ac51c61d4904022f548cd165c35db86875d/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java) 
When grabbing gauntlet loot, the `BOSSKILL_MESSAGE_PATTERN `checks for `"Your (.+) kill count is: <col=ff0000>(\\d+)</col>."`

The Gauntlet specifically names completion as `Your Gauntlet completion count is: x` with various timestamps above it, but that isn't your loot! The loot comes when you open the chest, which shows a generic chest message

I'm looking into how we can automatically screenshot these screens

This may also differ based on Corrupted, but I'm sure we can mask the text differences... if only I could complete Corrupted rn to check ;)

**Additional context**
... workaround atm is to screenshot Untradeable drops and delete the 500 screenshots taken inside the Gauntlet

~Yes loot tracker reads Gauntlet loot. Might look into how they track it later and see if I can draw inspiration~

Gotem https://github.com/runelite/runelite/blob/fae69d4fd267f7b4000ee7c25cd137c7336923a1/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L134 Thanks LootTrackerPlugin

I'll probably throw something together locally tomorrow night EST, kinda late atm

georgegaytan:
Closes #11335 

Gauntlet is SPECIAL when it gives you loot

How Gauntlet provides loot:
1. First you open the chest, it reads "You open the chest."
2. Then you have a brief delay while I assume OSRS rolls your drop
3. Then you actually get the loot with a generic chest message "You find some treasure in the chest!", shared by other chests in game

To get it to capture, we have to first await the message _"You open the chest."_
We then flag `shouldTakeGauntletScreenshot` to true to note the Gauntlet Loot chest has opened
Once items are received with `shouldTakeGauntletScreenshot == true`, we take the screenshot, `shouldTakeGauntletScreenshot`, and we're done

Example with latest:
Gauntlet Loot 2020-04-22_21-28-33


georgegaytan:
@Nightfirecat No worries about that, this is just a draft. This current PR is more investigation until it is marked as ready for review, not meant to be a candidate for merging

Note above - "This will actually put the client to sleep for .75s which is not acceptable
...
Looking into tracking multiple messages in a row or adding an event inspired by the loot tracker, but this is tricky bc existing Auto Screenshots are based on Strings/Text Patterns on 1-single chatMessage"

I would be curious to see if anyone had any suggestions for the latter bit, given the existing pattern for Screenshots

benjamin-musil:
I think we should use something else besides `Thread.sleep()`.

Double thumbs down, no discussions allowed here, also reading the existing thread before commenting isn't allowed either.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -346,6 +347,19 @@ public void onChatMessage(ChatMessage event)
 			}
 		}
 
+		if (config.screenshotGauntletLoot() && chatMessage.equals(GAUNTLET_LOOTED_MESSAGE))
+		{
+			try
```
TheStonedTurtle, line 352: Nightfire was saying to set a flag here (i.e. `gauntletScreenshot`) and then use the `ItemContainerChanged` event to take the screenshot.

For Example:
```java
	@Subscribe
	public void onItemContainerChanged(ItemContainerChanged c)
	{
		if (!gauntletScreenshot || c.getContainerId() != InventoryID.INVENTORY.getId())
		{
			return;
		}
		
		gauntletScreenshot = false;
		takeScreenshot("Gauntlet loot", "Gauntlet loot");
	}
```

This is how the loot tracker, and other plugins, handles similar situations.

georgegaytan: Awesome. I actually have something similar on my local, but time got away from me

When I'm off from work tonight I'll apply this to what I had brewing

georgegaytan:
@Nightfirecat @TheStonedTurtle Updated, changes documented in Summary
Setting this to Ready for Review

Let me know if anything else should be added or tweaked

Alexsuperfly:
whats preventing this from just taking a picture on the message you actually get the loot with

georgegaytan:
The unique message for Gauntlet reads `You open the chest.` which is followed by `You find some treasure in the chest!` -> the latter pops up when you actually get the loot

Example of chest opening:
image

image

I suppose we could extend this to work as the LootTracker does with global chests, if we wanted to just track `You find some treasure in the chest!`

Either of these approaches preferable?

[LootTrackerPlugin L139](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L139)
```
// Chest loot handling
private static final String CHEST_LOOTED_MESSAGE = "You find some treasure in the chest!";
private static final Pattern LARRAN_LOOTED_PATTERN = Pattern.compile("You have opened Larran's (big|small) chest .*");
private static final Map<Integer, String> CHEST_EVENT_TYPES = ImmutableMap.of(
	5179, "Brimstone Chest",
	11573, "Crystal Chest",
	12093, "Larran's big chest",
	13113, "Larran's small chest",
	13151, "Elven Crystal Chest"
);
```

[LootTrackerPlugin L503](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L503)
```
if (message.equals(CHEST_LOOTED_MESSAGE) || LARRAN_LOOTED_PATTERN.matcher(message).matches())
{
	final int regionID = client.getLocalPlayer().getWorldLocation().getRegionID();
	if (!CHEST_EVENT_TYPES.containsKey(regionID))
	{
		return;
	}

	eventType = CHEST_EVENT_TYPES.get(regionID);
	lootRecordType = LootRecordType.EVENT;
	takeInventorySnapshot();
	return;
}
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -42,16 +42,8 @@
...
-import net.runelite.api.Client;
-import net.runelite.api.GameState;
-import net.runelite.api.Player;
-import net.runelite.api.Point;
-import net.runelite.api.SpriteID;
-import net.runelite.api.events.ChatMessage;
-import net.runelite.api.events.GameTick;
-import net.runelite.api.events.PlayerDeath;
-import net.runelite.api.events.WidgetLoaded;
+import net.runelite.api.*;
```
Nightfirecat, line 45: Our code conventions are to avoid star imports. [Please read up on them and ensure your PR follows them.](https://github.com/runelite/runelite/wiki/Code-Conventions)

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
```diff
@@ -143,11 +143,22 @@ default boolean screenshotBossKills()
 		return false;
 	}
 
+	@ConfigItem(
+		keyName = "gauntletLoot",
+		name = "Screenshot Gauntlet Loot",
+		description = "Configures whether or not screenshots are automatically taken of Gauntlet Loot",
+		position = 10
+	)
+	default boolean screenshotGauntletLoot()
```
Nightfirecat, line 152: I think this config is not needed since we can use the `screenshotRewards()` config that already exists.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -346,6 +354,11 @@ public void onChatMessage(ChatMessage event)
 			}
 		}
 
+		if (config.screenshotGauntletLoot() && chatMessage.equals(GAUNTLET_LOOTED_MESSAGE))
```
Nightfirecat, line 357: Since the chest loot message is pretty ambiguous, this should also check that the player is next to the gauntlet chest by way of checking that the player is in the gauntlet chest region.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
georgegaytan: Gotcha thanks. Didn't even notice that happened 

Fixed adcb9e662a4dd734341b4a4cec2564dd8fdd4165

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
georgegaytan: adcb9e662a4dd734341b4a4cec2564dd8fdd4165

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
georgegaytan: Thanks, added it to the rewards config

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
georgegaytan: Got ya. I pulled the location from the existing loot tracker regionId if that's alright

https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L134-L137

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -267,6 +273,19 @@ public void onPlayerLootReceived(final PlayerLootReceived playerLootReceived)
...
+	@Subscribe
+	public void onItemContainerChanged(ItemContainerChanged event)
+	{
+		if (!shouldTakeGauntletScreenshot || event.getContainerId() != InventoryID.INVENTORY.getId())
+		{
+			return;
+		}
+
+		shouldTakeGauntletScreenshot = false;
+		takeScreenshot("Gauntlet Loot", "Gauntlet Loot");
```
Nightfirecat, line 286: I'd put this in the `"Boss Kills"` folder with other similar loot. (think Chambers of Xeric, Theatre of Blood)

Also, this currently doesn't record the actual gauntlet completion kill count. We should get this from the completion message like is done for other similar activities. (this could also use different filenames for regular vs. corrupted gauntlet)

georgegaytan: Happy to move to `Boss Kills` putting it in the same vein as raids. But I have a question about using KC in the filename


_"Completion of The Gauntlet will reward players with various loot, including crystal armour seeds for crystal armour, the Blade of Saeldor, and crystal shards to charge the previous items. Failing to finish, either by death or leaving early, will give a small token reward based upon how well the player performed, but won't grant any chance of unique items."_ - OSRS Wiki

A player can die in a Gauntlet run and receive loot. In these scenarios, the KC is neither incremented nor referenced

Thinking of 3 potential options

- We use KC in the filename and ignore screenshotting bad Gauntlet runs (my preference, though there's potential for confusion when loot is in loot tracker but not screenshotted)
- We use KC in the filename and set a special name or timestamp only for bad Gauntlet runs (eh)
- We don't use KC in the filename and use timestamp only for Gauntlet runs (similar to pet)

Adam-: If you can get gauntlet loot without kc, then this is probably fine. Maybe just name the folder `Gauntlet`?

georgegaytan: Done 337c0677acde349f8e625dca7ad16b15a8d872bf

Alexsuperfly:
Could you perhaps try this [commit](https://github.com/Alexsuperfly/runelite/commit/75a082bb71e9c80f95d4088383a8cb71f1d537f6) out.  I believe it does what you want and also refactors the loot tracker to prevent confusion on how the gauntlet works in the future.

georgegaytan:
> Could you perhaps try this [commit](https://github.com/Alexsuperfly/runelite/commit/75a082bb71e9c80f95d4088383a8cb71f1d537f6) out. I believe it does what you want and also refactors the loot tracker to prevent confusion on how the gauntlet works in the future.

Looks good. I cloned your branch and fetched upstream, but unfortunately couldn't do a Gauntlet test run. Plugins were not available due to 
image

Same deal on your end?

Alexsuperfly:
I don't think its possible to have that if you've directly pulled my branch since its only 9 commits behind upstream, so perhaps hard reset to my branch or rebase whatever you're currently on to upstream/master.

georgegaytan:
@Alexsuperfly Got it. Did a test run and it works 👍

Alexsuperfly:
You're welcome to take and author that yourself into this pr, can change the screenshot folder name to something else too, `Chest Loot` was just the first thing i thought of.

georgegaytan:
@Alexsuperfly merged in your commit's changes. In this version, I opted to just use the Event name as the directory name. Having a named folder with `The Gauntlet` makes it a lot easier to find. Seems appropriate for this scenario since, at least as of today, it's a pretty unique boss&loot system

Nightfirecat:
It seems like you've forgotten to push your changes. This PR hasn't been updated since last week.

Adam-:
I have taken 1444ddcb677baf74839bea0752896309391ffee3 instead

georgegaytan:
@Nightfirecat my b, last week was all late nights at work. Didn't get my patch finalized till last night when I could verify the commit and the final merge of the commit w/ my work

Alas, post-midnight finished all that and forgot to push when I went to bed lol

Issue #11354, opened 2020-04-23
Spam clicking client close button causes exceptions


Juuldamen:
**Describe the bug**
Each click on the client close button after the first will throw exceptions.

**To Reproduce**
1. Open the client and wait for it to load.
2. Click on the close button and spam it after the first click.
3. See error

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.6.12
 - Launcher version: Built from master

**Additional context**
Logs of the error:
```
2020-04-23 22:46:38 [pool-2-thread-1] WARN  n.r.c.util.RunnableExceptionLogger - Uncaught exception in runnable net.runelite.client.ClientSessionManager$$Lambda$1069/1208109353@453563c5
java.lang.NullPointerException: null
	at net.runelite.client.SessionClient.delete(SessionClient.java:89)
	at net.runelite.client.ClientSessionManager.lambda$onClientShutdown$0(ClientSessionManager.java:84)
	at net.runelite.client.util.RunnableExceptionLogger.run(RunnableExceptionLogger.java:41)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
2020-04-23 22:46:38 [RuneLite Shutdown] WARN  n.r.client.events.ClientShutdown - Error during shutdown: 
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:206)
	at net.runelite.client.events.ClientShutdown.waitForAllConsumers(ClientShutdown.java:60)
	at net.runelite.client.ui.ClientUI.lambda$shutdownClient$9(ClientUI.java:581)
	at java.lang.Thread.run(Thread.java:748)
[...]

PR #11358, opened 2020-04-24
Preventing ClientSessionManager from calling onClinetShutdown multiple times 

Doxterpepper:
Saw this as a good easy bug to fix to get introduced to runelite. Is this a good solution?
#11354

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ClientSessionManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ClientSessionManager.java
@@ -81,7 +81,11 @@ public class ClientSessionManager
 		{
 			try
 			{
-				sessionClient.delete(sessionId);
+				UUID localUuid = sessionId;
+				if (localUuid != null)
+				{
+					sessionClient.delete(localUuid);
+				}
 			}
 			catch (IOException ex)
 			{


Adam-:
`sessionId` can also be null if a ping fails, so maybe that should also be fixed (as it would also throw without this bug). Maybe the correct way to fix this is to prevent the shutdown from happening twice?

Doxterpepper:
So abort shutdown if the session ID is already null?

Adam-:
The root of this problem is `shutdownClient` is being called >1 times I think. Not sure the best way to make it not do that. Completely unrelated to that, is L125 can set `sessionId = null` which means that it can be null in `onClientShutdown`, so I would just null check it there.

deathbeam:
Simple isShuttingDown variable in ClientUI would suffice imo

Doxterpepper:
> Simple isShuttingDown variable in ClientUI would suffice imo

That's what I was originally thinking too, but I felt like adding a new variable just for this added clutter. Especially when checking if sessionId is null does the same thing and prevents any other scenario where we'd call delete with a null session ID.

runelite-client/src/main/java/net/runelite/client/ClientSessionManager.java
```diff
@@ -75,6 +75,11 @@ public void start()
 	@Subscribe
 	private void onClientShutdown(ClientShutdown e)
 	{
+		if (sessionId == null)
```
Adam-, line 78: this null check should be down next to the `delete` call

runelite-client/src/main/java/net/runelite/client/ClientSessionManager.java
```diff
@@ -81,7 +81,10 @@ private void onClientShutdown(ClientShutdown e)
 		{
 			try
 			{
-				sessionClient.delete(sessionId);
+				if (sessionId != null)
```
Adam-, line 84: There are potentially ~3 threads here (event, which comes in on EDT?), this thread, and the thread executing `ping`, so this is not thread safe. You should make a local reference to sessionId and then use that.

Doxterpepper: Good catch, I hadn't thought of the threaded nature of this.

runelite-client/src/main/java/net/runelite/client/ClientSessionManager.java
```diff
@@ -81,7 +81,11 @@ private void onClientShutdown(ClientShutdown e)
 		{
 			try
 			{
-				sessionClient.delete(sessionId);
+				UUID localUuid = sessionId;
+				if (localUuid != null)
+				{
+					sessionClient.delete(sessionId);
```
Adam-, line 87: this should be `localUuid`

PR #11366, opened 2020-04-25
api: correct pearl fly fishing and barbarian rod animations

Hydrox6:
So, each of these rods have 2 animation IDs, with one of them showing the casting of the rod. This casting animation is used by the normal and oily rods. **However**, for whatever reason, the Fly fishing and Barbarian rods both use the non-casting animation whilst active.

Thanks to `Kaan` on the Discord for reporting this.


Nightfirecat:
Can you please add the PR body to the commit message? As-is, the commit message itself doesn't explain why that fixes it.

Hydrox6:
Good idea. Added

PR #11368, opened 2020-04-25
Migrate to Java JDK 11 (LTS)

Flightkick:
I would like to use some JDK 11 API features which are currently not available since the project is still on JDK 8.

Test succeeds, project builds, client still seems to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ cache:
   directories:
     - $HOME/.m2
 jdk:
-- openjdk8
 - openjdk11
 install: true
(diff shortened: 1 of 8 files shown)


Adam-:
About 45% of our users use a launcher that ships Java 8, as we only updated the launcher to Java 11 last year. I do not want to force that many users to upgrade their launchers at this time.

Also, the version of Spring we run doesn't work with Java 11.

raiyni:
added in https://github.com/runelite/runelite/commit/e49b9ea172c6f77218acd801750898e3591ed3d9

PR #11374, opened 2020-04-26
feat(grounditems): add item pile (menu entries) sorting

RanadeepPolavarapu:
Adds a new feature to the *Ground Items* plugin to sort the item pile menu entries by **GE**, **HA**, **HIGHEST** (default), and **OFF** options. The highlighting and everything else works as-is as that is untouched.

Behaviour: 
image

Config:
image

This behaviour is very useful for PvP'ing and PvM'ing.


abextm:
This is a [rejected feature](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features)

PR #11394, opened 2020-04-28
Add disable hotkey to anti-drag plugin.

Unholypanda:
This pull request adds the control key as a hotkey to disable the anti-dragging functionality when the onShiftOnly setting is disabled which makes re-organizing the inventory while the plugin is enabled less cumbersome with a long delay.


runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
```diff
@@ -112,7 +112,13 @@ public void keyTyped(KeyEvent e)
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (e.getKeyCode() == KeyEvent.VK_SHIFT && config.onShiftOnly())
+		System.out.println(e.getKeyCode());
```
Adam-, line 115: delete this.

runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragConfig.java
```diff
@@ -53,4 +53,15 @@ default boolean onShiftOnly()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "disableOnCtrl",
```
Adam-, line 58: Fix the indenting here

Nightfirecat:
Upon re-reviewing this, as written, this does not work properly when holding <kbd>Ctrl</kbd> and opening the bank. Also it sets `held` when pressing and releasing <kbd>Ctrl</kbd> which isn't correct. I've authored <https://github.com/Nightfirecat/runelite/commit/3a9cec27de818623a096486762d870a3e202f04a> which splits <kbd>Ctrl</kbd> holds into a separate field and references that when opening the bank. @Unholypanda Can you confirm that this linked commit works as you would expect?

Unholypanda:
Tested your version and it seems to work properly in the bank. Sorry if I didn't do the testing process properly I'm kinda new to using Git collaboratively. I just copied your changes into my repository, tested, and pushed the updated code.

Nightfirecat:
Added in cc2ed2311bec808193fbae382cadf4a973f64b20

PR #11420, opened 2020-05-01
loot-tracker: make clue scroll drops stackable

### LINKED ISSUE #11416: clue scroll drops in loot tracker
was doing clues while in f2p (beginner) and the loot tracker stacks the clue scrolls dropped by dark wizards (as i expect it would) when i went to p2p for medium clues, the mediums didn't stack in loot tracker and clogs up the loot (killing guards in fally) currently at ~15 clues, but when i am going for 100 it will definitely become more of an issue. dont know which one is meant to be implemented, but i would assume them stacking as if 1 drop (like every other drop that isn't stackable).

begref

medissue

while-loop:
closes #11416 

Fixes the bug where clue scroll variations (excluding beginner & master) were not being stacked in loot tracker.


raiyni:
You don't need to include the issue number in your commits. It can get spammy for the issue author everytime you push.

PR #11423, opened 2020-05-01
Screenshot: Untradeable Drops Screenshots for the Gauntlet

### LINKED ISSUE #11362: Untradable drops screenshots for The Gauntlet
would it be possible to filter out the untradeable drops from the screenshot folder when getting things within The Gauntlet? (ie weapon frames, corrupted shards, corrupted teleports, etc) the folder is getting quite quite heavy with screenshots that are not really important.

basicDamen:
Closes #11362 

Removes the screenshotting of untradeable items when within the Gauntlet as it spams screenshots on every untradeable drop.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -48,6 +48,7 @@ import net.runelite.api.GameState;
 import net.runelite.api.Player;
 import net.runelite.api.Point;
 import net.runelite.api.SpriteID;
+import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.PlayerDeath;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -248,6 +250,10 @@ public void onPlayerDeath(PlayerDeath playerDeath)
 		if (player == client.getLocalPlayer() && config.screenshotPlayerDeath())
 		{
 			takeScreenshot("Death", "Deaths");
+			if (insideGauntlet)
```
dekvall, line 253: I don't think that you need to check this

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -357,14 +363,15 @@ public void onChatMessage(ChatMessage event)
 			}
 		}
 
-		if (config.screenshotUntradeableDrop())
-		{
-			Matcher m = UNTRADEABLE_DROP_PATTERN.matcher(chatMessage);
-			if (m.matches())
+		if (config.screenshotUntradeableDrop()) {
+			if (!insideGauntlet)
```
dekvall, line 367: change this to `if (config.screenshotUntradeableDrop() && !insideGauntlet)`

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -379,6 +386,16 @@ public void onChatMessage(ChatMessage event)
...
+
+		if (chatMessage.equals("You enter the Gauntlet."))
+		{
+			insideGauntlet = true;
+		}
+
+		if (chatMessage.equals("You leave the Gauntlet."))
+		{
+			insideGauntlet = false;
+		}
```
dekvall, line 398: These should check for type. Also, is it possible to leave in any way so that the exit message isn't triggered? Logging out, teleporting etc? Maybe it's better to use a region check?

PR #9116, opened 2019-06-15
Reset Discord status after a while of being in menu

DSerp:
Closes #667

PR #11444, opened 2020-05-03
Add setting for Discord to not show RuneLite activity when not logged in to runescape

RicardovanDijke:
kinda niche functionality, but I leave RuneLite running when not playing to save the loot tab and gained xp. This causes Discord to display i'm playing RuneLite all day long.


Adam-:
This should probably just be in `checkForTimeout` similar to #9116

PR #11452, opened 2020-05-03
agility: separate Prifddinas portals overlay

machkr:
Allows for the portals in the Prifddinas agility course to be toggled/configured separately from traditional obstacles.


melkypie:
I believe there is already an approved pr which is waiting to be merged #10543

machkr:
You are right! I should've searched first - closing.

Issue #11427, opened 2020-05-01
Add buy/sell 1/5/10/50 menu entry swaps for shops, like you can do on mobile.


geheur:
**Is your feature request related to a problem? Please describe.**
When buying a lot of an item from shops, having to right click to select the buy/sell option for the item takes some effort. The same is true for selling large amounts an item to shops.

**Describe the solution you'd like**
I'd like to be able to swap the buy/sell 1/5/10/50 options to the top of the menu so I can buy or sell lots of items with just a left-click.

**Describe alternatives you've considered**
Mousekeys/ahk - takes more effort than a single click and I'd have to do some research to learn how to set this up.
For selling stackable items, putting the item in the bottom of the inventory to make the right click menu open up with "Sell 50" already under the mouse - only works for selling.

**Additional context**
This swap was once rejected but the wording on this page https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#menu-entry-swapping makes it sound like it would now be ok to have this swap due to mobile having something similar.

Adam-:
added in #11453

PR #11453, opened 2020-05-03
Adds menu entry swaps for Buy 1/5/10/50 and Sell 1/5/10/50.

geheur:
Pull request for issue #11427.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2020, Sam <dasistkeinnamen@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.menuentryswapper;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum BuyMode
+{
+	OFF("Off"),
+	BUY_1("Buy 1"),
+	BUY_5("Buy 5"),
+	BUY_10("Buy 10"),
+	BUY_50("Buy 50");
+
+	private final String name;
+
+	@Override
+	public String toString()
+	{
+		return name;
+	}
+}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
@@ -31,17 +31,11 @@ import lombok.RequiredArgsConstructor;
 @RequiredArgsConstructor
 public enum BuyMode
 {
-	OFF("Off"),
+	OFF(null),
 	BUY_1("Buy 1"),
 	BUY_5("Buy 5"),
 	BUY_10("Buy 10"),
 	BUY_50("Buy 50");
 
-	private final String name;
-
-	@Override
-	public String toString()
-	{
-		return name;
-	}
+	private final String option;
 }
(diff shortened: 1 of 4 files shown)


dekvall:
This seems dangerous, especially the default sell setting which some users will undoubtedly forget to reset their setting after selling something in bulk and then end up selling their gear to the general store. Some specific items have had their buy prices [artificially inflated](https://www.reddit.com/r/2007scape/comments/31gus9/dfs_costs_2b_in_general_store/) and might be hard to get back once sold.

I'm thinking a better solution would be to have an option to remember the last used Sell option and reset it every time you open a shop interface. I can see how the use of a configurable buy option is beneficial to ironmen who needs to run back and forth from shops a lot though.

I don't know of many situations where hopping and selling items are done except for the karamja store, and even then it would only be one extra click per hop in comparison to a lost dfs.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
```diff
@@ -0,0 +1,47 @@
...
+	BUY_10("Buy 10"),
+	BUY_50("Buy 50");
+
+	private final String name;
+
+	@Override
+	public String toString()
+	{
+		return name;
+	}
```
Nightfirecat, line 46: All of this can be removed. [Enum names are converted to Title-Cased Entries in the resulting combobox.](https://github.com/runelite/runelite/blob/8eb57c5a92ee05df47dbdc9d530f3c2da77f0bd4/runelite-client/src/main/java/net/runelite/client/ui/components/ComboBoxListRenderer.java#L62-L65)
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
```diff
@@ -0,0 +1,47 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.menuentryswapper;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
```
Nightfirecat, line 31: Similarly, these are not needed.
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/SellMode.java
```diff
@@ -0,0 +1,47 @@
...
+	SELL_10("Sell 10"),
+	SELL_50("Sell 50");
+
+	private final String name;
+
+	@Override
+	public String toString()
+	{
+		return name;
+	}
```
Nightfirecat, line 46: See my comments on `BuyMode`.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -773,6 +773,46 @@ else if (option.equals("collect to inventory") || option.startsWith("collect-not
...
+					case BUY_5:
+						swap("buy 5", option, target, index);
+						break;
+					case BUY_10:
+						swap("buy 10", option, target, index);
+						break;
+					case BUY_50:
+						swap("buy 50", option, target, index);
+						break;
+				}
```
Nightfirecat, line 794: Instead of creating a switch case for each enum value, the menu options can be defined in the enum and set much more concisely here.
```suggestion
				swap(config.shopBuy().getOption(), option, target, index);
```

geheur:
I didn't consider how dangerous this could be if you forgot you left it enabled.

I think the safest way to do it would be to set a buy/sell quantity on a per-item basis and have it activated through some kind of shift-rightclick menu. If you're buying or selling so much of something that you'd go through the effort of setting a sell/buy amount for an item, it's unlikely that buying or selling 1/5/10/50 of that item on accident in the future would cripple you financially.

I don't know if individual buy/sell quantities is considered "overpowered" by jagex but if it is then you could enable the swap for individual items and have the same quantity for each item where this is enabled, with the quantity being set in the plugin config.

Resetting the sell setting every time a shop interface would address a few problematic situations, but if you forget you have this enabled and sell an item, and then try to "value" another item, you'd still sell it accidentally.

Nightfirecat:
I think it's fine to leave in. In my experience, players are buying from shops much more often than selling to them, and if they're selling to shops, it's usually only very specific items to specific shops--usually to those that offer high-alch prices.

geheur:
@Nightfirecat what exactly are you referring to as being safe to leave in? The way it currently is on my branch (option in config to set buy/sell 1/5/10/50 for all items), dekvall's suggestion (reset this setting every time a shop interface is opened), or my suggestion (remember sell/buy quantity on a per-item basis)?

Nightfirecat:
I mean it should be fine to leave as-is currently PR'd.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/BuyMode.java
geheur: This should not be removed if the option needs to be accessible to make code in MenuEntrySwapperPlugin more concise.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
geheur: Added this for other similar menu entry swaps.

geheur:
I think the risk of accidentally buying or selling something you don't want it still really high - I wouldn't trust myself to use this as it is now even knowing the dangers.

I plan on making this setting unique to each item, but feel free to accept this as it is now if y'all prefer that.

geheur:
Alright, I'm happy with the way this works now. The swap is now specific to the item and whether it's a Buy or a Sell.

Adam-:
I don't see anything wrong with having a sell-bulk option and a buy-bulk option and just defaulting them to 1?

geheur:
Reverted all of the item-specific stuff, and changed the swap to only occur when shift is held, to prevent accidental buying or selling.

Nightfirecat:
Why is this PR `+557 −462`? Please revert everything except the feature you're adding.

geheur:
Alright, took out everything not strictly related to this change.

The reason this PR looked big earlier was because I:

Took carriage returns out of a file (MenuEntrySwapperConfig.java), since either intellij or git was automatically removing them and that was causing me some problems, and it seemed like most other files in the project had unix-style line endings.

Changed some other unrelated menu entry swaps to be more concise in the same way you suggested for the buy/sell swap in this PR (getting the option string from the Enum instead of a switch statement.

Sorry if doing any of that was a faux pas, or if it should have been done differently.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -884,7 +896,7 @@ private int findIndex(MenuEntry[] entries, int limit, String option, String targ
 	{
 		if (strict)
 		{
-			List<Integer> indexes = optionIndexes.get(option);
+			List<Integer> indexes = optionIndexes.get(option.toLowerCase());
```
Nightfirecat, line 899: Why is this changed?

geheur: I changed this so that I could have the option strings in BuyMode and Sellmode have the same capitalization as they do in-game, rather than already being in lower case. Should I change BuyMode and Sellmode to just have the option be lowercase to start with?

Nightfirecat: That would be best. :+1:

PR #11463, opened 2020-05-04
add raids & trade widget IDs

while-loop:
Screen Shot 2020-05-04 at 2 56 24 PM
Screen Shot 2020-05-04 at 2 56 57 PM
Screen Shot 2020-05-04 at 3 25 18 PM
Screen Shot 2020-05-04 at 3 25 36 PM


while-loop:
> @while-loop We usually don't add widget IDs unless they're being used. ...
> We've had other plugin hub plugins hardcode them instead of keeping them "unused" in the main client

PR #11480, opened 2020-05-06
Add duplicate/collapse chat plugin

CoreyForsyth:
I'm making this PR mostly for feedback and discussion. 
It might be a little rough around the edges but after using it for a while, I think it is very nice to have.

A few things to note:

1. I decided to use a cache for keeping count of the messages

- This allows it to count more than just the chat buffer length, so if you are logged in for many hours, message counts can go into the hundreds or thousands with no problem
- This also prevents modification of the chat buffer itself, so it will retain all of the chat history. This means if the plugin is turned off after consolidating the chats, the chat will go back to normal
- To prevent this cache from growing unbounded, I added a configurable max size. I defaulted to 500 because it was the first number that popped into my head, but I believe it should be able to handle much more than that. This size corresponds to unique messages, not total messages.

2. I used the same callback that the chatfilter plugin uses, "chatFilterCheck". 

- This could very well belong as a part of that plugin instead of as a separate one, although to me the functionality is different enough to warrant a separate plugin

3. Public chat consolidation is configurable

- Only the "spammy" message types are consolidated by default: ENGINE, GAMEMESSAGE, NPC_EXAMINE, OBJECT_EXAMINE, SPAM.
- PUBLICCHAT is configurable on/off

4. Public chat can be hidden after X amount of duplicates

- If the player so chooses, they can filter out messages if they are repeated a bunch. I tested this out in world 1 and had great success. After a bit of standing around the GE, the duplicate messages were filtered out and only regular conversation was left


melkypie:
I still think this should be in chat filters, as it basically offers the same functionality as chat filters do - filter out messages but this time only duplicate messages. I don't believe there is enough reason to add one more plugin to the already extensive list of plugins.
Also I suggest you take a look at https://github.com/runelite/runelite/wiki/Code-Conventions to fix all those Travis errors

CoreyForsyth:
> I still think this should be in chat filters, as it basically offers the same functionality as chat filters do - filter out messages but this time only duplicate messages. I don't believe there is enough reason to add one more plugin to the already extensive list of plugins.
> Also I suggest you take a look at https://github.com/runelite/runelite/wiki/Code-Conventions to fix all those Travis errors

Yeah, I can definitely agree with that. I can refactor this into the chat filter plugin no problem. I am interested to hear feedback about this implementation as well. And thank you for the suggestion, I will do that now

runelite-client/src/main/java/net/runelite/client/plugins/duplicatechat/DuplicateChatPlugin.java
```diff
@@ -0,0 +1,214 @@
...
+	DuplicateChatConfig provideConfig(ConfigManager configManager)
+	{
+		return configManager.getConfig(DuplicateChatConfig.class);
+	}
+
+	@Override
+	protected void startUp() throws Exception
+	{
+		chatIndex = new HashMap<>();
+		chatCounts = HashMultiset.create();
```
Nightfirecat, line 77: These can be initialized with the plugin rather than on startUp. They should also be cleared on shutDown.

runelite-client/src/main/java/net/runelite/client/plugins/duplicatechat/DuplicateChatPlugin.java
```diff
@@ -0,0 +1,214 @@
...
+		description = "Collapse duplicate game chats into a single message",
+		enabledByDefault = false
+)
+@Slf4j
+public class DuplicateChatPlugin extends Plugin
+{
+
+	private static final String COUNT_FORMAT = " (%d)";
+	private Multiset<String> chatCounts;
+	private Map<String, Integer> chatIndex;
```
Nightfirecat, line 59: Can't these both be replaced by a single `MultiMap<String, Integer>`?

CoreyForsyth: No, I don't think so. The chatCounts keeps track of the amount of duplicate chat messages with the key being the message, and the chatIndex keeps track of the newest index of the message in question, so that the chatfilter can tell if it should be blocked or not

raiyni:
All of the refactoring looks unneeded?

CoreyForsyth:
> All of the refactoring looks unneeded?

I'd say so, it was fairly minor, like pulling some things out to a separate function. Since this is using the same filtering pattern as the regular chat filter there was bound to be a bit of change. Do you see anything in particular you would want me to change?

runelite-client/src/main/java/net/runelite/client/plugins/duplicatechat/DuplicateChatPlugin.java
Nightfirecat: Isn't that achieved by checking the following in a `ListMultiMap`?

- Number of duplicated chat messages: `multiMap.get(str).size()`
- Newest index of message in question: `Iterables.getLast(multiMap.get(str))`

CoreyForsyth: I see what you are saying now, that makes a lot of sense. I made the change to use a multimap, it definitely looks cleaner now. Thank you for the suggestion!

Issue #11482, opened 2020-05-06
Add option to toggle activity overlay in the Wintertodt Plugin


anonymouscnidaria:
**Is your feature request related to a problem? Please describe.**
The wintertodt plugin shows the number of roots and kindlings in the inverntory

**Describe the solution you'd like**
I would like it to be toggle-able so I can turn it off as it's not something i need to see on screen

**Describe alternatives you've considered**
n/a

**Screenshot**
![image](https://user-images.githubusercontent.com/58953980/81161053-87505080-8fbd-11ea-81ed-eab3163f8c54.png)

abextm:
added in eae4e0b

PR #11483, opened 2020-05-06
Add Overlay Toggle to Wintertodt Plugin

JonnoUK:
Requested feature
https://github.com/runelite/runelite/issues/11482

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
@@ -128,4 +128,14 @@ public interface WintertodtConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 9,
+			keyName = "Overlay",
+			name = "Client Overlay Toggle",
+			description = "Toggle the overlay for this plugin"
+	)
+	default boolean clientOverlayToggle() {
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
@@ -135,7 +135,8 @@ public interface WintertodtConfig extends Config
 			name = "Client Overlay Toggle",
 			description = "Toggle the overlay for this plugin"
 	)
-	default boolean clientOverlayToggle() {
+	default boolean clientOverlayToggle()
+	{
 		return true;
 	}
 }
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
```diff
@@ -128,4 +128,15 @@ default boolean notifyBrazierOut()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 9,
+			keyName = "Overlay",
+			name = "Client Overlay Toggle",
+			description = "Toggle the overlay for this plugin"
```
Nightfirecat, line 136: These lines are all indented one level too far.
```suggestion
		position = 9,
		keyName = "Overlay",
		name = "Client Overlay Toggle",
		description = "Toggle the overlay for this plugin"
```

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtOverlay.java
```diff
@@ -53,7 +53,7 @@ private WintertodtOverlay(WintertodtPlugin plugin)
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		if (!plugin.isInWintertodt())
+		if (!plugin.isInWintertodt() || plugin.clientOverlayToggle == false)
```
Nightfirecat, line 56: `boolean == false` is the same as `!boolean`, use the latter form.

Also, don't store the config state in plugin, just inject the config and directly call the config method here instead.

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtConfig.java
Nightfirecat: Also, the names given to the added config fields/method are not good. I would recommend naming the method and keyName `displayOverlay`, the name `Display Overlay`, and the description `Display an overlay which shows your activity status and number of points and kindling held in your inventory.`

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtOverlay.java
```diff
@@ -50,10 +50,13 @@ private WintertodtOverlay(WintertodtPlugin plugin)
 		getMenuEntries().add(new OverlayMenuEntry(RUNELITE_OVERLAY_CONFIG, OPTION_CONFIGURE, "Wintertodt overlay"));
 	}
 
+	@Inject
+	private WintertodtConfig config;
```
Nightfirecat, line 54: Inject this in the constructor in the same way that `plugin` is injected in this class.

runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -126,7 +126,7 @@
 	private Instant lastActionTime;
 
 	private int previousTimerValue;
-
+
```
Nightfirecat, line 129: Remove this change
```suggestion

```

Adam-:
added in eae4e0b6a06618aec9f9df5d4be8bdadddd26f53

Issue #10566, opened 2020-01-08
[Suggestion] Include current world with !layout chat command, toggle-able. 


Jawnb64:

SieBrum:
What is the point of this if you can just right-click the !layout message and use the "hop-to" feature? Saves clicks and won't add extra settings / useless information

PR #11487, opened 2020-05-06
Add Player World to end of !Layout command (Toggle)

JonnoUK:
Resolves https://github.com/runelite/runelite/issues/10566

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
@@ -738,9 +738,6 @@ public class RaidsPlugin extends Plugin
 		if (config.worldDisplay())
 		{
 			worldString = " on W" + client.getWorld();
-		} else
-		{
-			worldString = "";
 		}
 
 		String response = new ChatMessageBuilder()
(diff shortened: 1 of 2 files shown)


melkypie:
Currently this will only show the world the player receiving/sending the message is on not the world the actual raid is on.

JonnoUK:
> Currently this will only show the world the player receiving/sending the message is on not the world the actual raid is on.

From my tests you can only do the !layout command once a member of your party has entered the raid.
Would a check on whether you are scouting the raid or not help?:

`if (config.worldDisplay() && inRaidChambers)`

Hydrox6:
The world needs to be sent along with the data to actually be valid, since it would only ever show the world that the receiving player is on.

melkypie:
That wouldn't do anything, the process of a chat command is it has to be submitted to the server and then the other clients pull it down from the server and the display it.
You will need to have http-service set up to be able to test this properly

Adam-:
If you assume the player is always on your FL/CC you could probably just grab the player world from that, and not show it otherwise.

runelite-client/src/main/java/net/runelite/client/plugins/raids/RaidsPlugin.java
```diff
@@ -193,6 +193,8 @@
 
 	private RaidsTimer timer;
 
+	private String worldString;
```
Hydrox6, line 196: This should be local to the function that uses it

PR #11493, opened 2020-05-07
Start/Stop plugins via keybind

gp-it-be:


Nightfirecat:
I don't think we will want to add such a feature, as the most frequent uses for it that players might want would be quite controversial, such as toggling the entity hider plugin.

PR #11502, opened 2020-05-08
loot tracker: Use "an" prefix for monster names starting with vowels

Nightfirecat:
Amazingly none of the various util libraries we pull in have a "startsWithVowel" method or anything of the like.


makingcomment:
The way you've done this technically isn't correct. The indefinite article for a word is "an" if and only if the word starts with a vowel _sound_ (otherwise it's "a").

Of course, words usually only start with vowel sounds when they start with a vowel, so this implementation will work for almost all cases. In fact, the only in-game exception I can think of is "Unicorn" (but there might be more).

PR #2444, opened 2018-05-07
Add category system to plugin configuration UI. Fixes #2411

AdamBallinger:
001447b626d2d6324d0fbc55c2c883f5
Adds category field to PluginDescriptor annotation.
New categories can be added in the PluginCategory enum.
New categories are automatically added to the config UI.

PR #11507, opened 2020-05-08
Plugin Categories

### LINKED ISSUE #11417: Plugin organizer
I am suggesting a plugin that will let users choose to organize all plugins into categories to make it much easier to find plugins without having to remember the names of plugins.

This plugin would put all plugins into categories like combat, skills, items etc to make it much easier to find plugins related to a specific activity. I would also suggest having a favorite category at the very top where all of a users favorite plugins will be. These plugin categories should be drop-down menus to show the plugins for each category. This will make it much easier to find the plugins and categories without having to scroll a lot to find what you want.

This plugin have been suggested before, but has never been released, and i think that's a shame, because i think that a lot of people would like to have the option to organize plugins by categories, and i think a lot of people would greatly benefit from it.

This is a screenshot from https://github.com/runelite/runelite/pull/2444, which was never released, but is similar to something that i have in mind for this plugin, except like i said i think that the categories should be drop down menus to make categories and plugins easier to find without a lot of scrolling.

Spedwards:
Closes #11417

This is a revival of #2444. Plugin categorisation will not be enabled by default for those that do not want it.

~Because of the Plugin Hub, the first commit would need to be merged in advance to allow for external plugins to add a category.~

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListPanel.java
@@ -31,16 +31,21 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
+import javax.swing.BorderFactory;
 import javax.swing.JButton;
+import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.ScrollPaneConstants;
+import javax.swing.SwingConstants;
 import javax.swing.SwingUtilities;
 import javax.swing.border.EmptyBorder;
 import javax.swing.event.DocumentEvent;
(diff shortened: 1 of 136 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListPanel.java
```diff
@@ -252,35 +298,90 @@ private void onSearchBarChanged()
...
+		categoryLabels.values().forEach(mainPanel::remove);
 
 		showMatchingPlugins(true, text);
 		showMatchingPlugins(false, text);
 
+		repaint();
 		revalidate();
 	}
 
 	private void showMatchingPlugins(boolean pinned, String text)
```
Spedwards, line 311: I am not particularly happy with this method. It requires some serious review.

deathbeam:
Can you maybe use less obnoxious headers? They are *huge*

Spedwards:
> Can you maybe use less obnoxious headers? They are _huge_

I have decreased the font size for the category headers down to 24 points from the previous 32. Hopefully that's better.

Spedwards:
I've squashed all the commits and added `PluginCategory.HUB` which is now the default category. This is specifically for external plugins.

Whether they can change their category or they must remain in the Hub category can be left to admins.

psikoi:
Why is the font on those category titles so big?

psikoi:
You got any images of how it looks currently?

Spedwards:
> You got any images of how it looks currently?

It currently looks like this.

Plugin List

Spedwards:
> this is ugly and will produce more confusion than it solves; see #11606

I'm in favour of your PR. I was asked how it currently looked and I answered. And thanks for calling it ugly :/

makingcomment:
I don't think this is ugly. I also don't see why this and #11606 can't coexist.

abextm:
This is basically just a more constrained tag system, which we already have. I don't see any use in duplicating all that information we already have in a way that will confuse people more.

PR #11509, opened 2020-05-08
Skill calculator change to show needed xp and xp number formatting

RobertCurll:
Skill calculator changes to improve formatting of numbers and to include the required amount of experience between current xp and target xp.


Adam-:
This has been pr'd a few times in the past, but I really don't want to use so much screen real estate for this. Maybe a tooltip on `Target Experience` would work instead.

RobertCurll:
Updated to have required experience to level up as a tooltip on Target Experience instead of a new row on the panel.

RobertCurll:
How the skill calculator looks like now
image

Nightfirecat:
Because you've opened this PR from your master branch, I'm unable to make updates to it. Please update your branch to https://github.com/Nightfirecat/runelite/tree/needed-xp as it cleans up the code and splits the two changes into separate commits for each of them.

PR #11522, opened 2020-05-09
discord: Update activity message while playing PVP or DMM

### LINKED ISSUE #11430: Discord plugin leaks ingame activity wile playing Deadman Mode
Given the competitive PVP nature of DMM/SDMM/DMMT, we should not display the current activity of players on DMM. I think it would be safe to add a "Playing Deaman Mode" activity line rather than omitting it like we do for wilderness activities.

rangvaldr:
Closes #11430.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/WorldType.java
+++ b/runelite-api/src/main/java/net/runelite/api/WorldType.java
@@ -26,7 +26,6 @@ package net.runelite.api;
 
 import java.util.Collection;
 import java.util.EnumSet;
-import java.util.function.Predicate;
 
 /**
  * An enumeration of possible world types.
@@ -82,11 +81,8 @@ public enum WorldType
 	}
 
 	private static final EnumSet<WorldType> PVP_WORLD_TYPES = EnumSet.of(
+		DEADMAN, // dmmt worlds are also flaged as DEADMAN
 		PVP
-	);
-
-	private static final EnumSet<WorldType> DEADMAN_WORLD_TYPES = EnumSet.of(
-		DEADMAN
 	);
 
 	/**
@@ -138,17 +134,5 @@ public enum WorldType
 	public static boolean isPvpWorld(final Collection<WorldType> worldTypes)
 	{
 		return worldTypes.stream().anyMatch(PVP_WORLD_TYPES::contains);
-	}
-
-	/**
-	 * Checks whether a world having a {@link Collection} of {@link WorldType}s is a Deadman world.
-	 *
-	 * @param worldTypes A {@link Collection} of {@link WorldType}s describing the given world.
-	 * @return           True if the given worldtypes of the world are a Deadman world, false otherwise.
-	 * @see Client#getWorldType()
-	 */
-	public static boolean isDeadmanWorld(final Collection<WorldType> worldTypes)
-	{
-		return worldTypes.stream().anyMatch(DEADMAN_WORLD_TYPES::contains);
 	}
 }
--- a/runelite-api/src/main/java/net/runelite/api/WorldType.java
+++ b/runelite-api/src/main/java/net/runelite/api/WorldType.java
@@ -26,6 +26,7 @@ package net.runelite.api;
 
 import java.util.Collection;
 import java.util.EnumSet;
+import java.util.function.Predicate;
 
 /**
  * An enumeration of possible world types.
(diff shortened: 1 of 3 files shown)


runelite-api/src/main/java/net/runelite/api/WorldType.java
```diff
@@ -77,10 +78,13 @@
 	}
 
 	private static final EnumSet<WorldType> PVP_WORLD_TYPES = EnumSet.of(
-		DEADMAN,
```
Nightfirecat, line 80: Deadman is still a PVP world type. Why is this removed?

runelite-api/src/main/java/net/runelite/api/WorldType.java
```diff
@@ -131,4 +135,16 @@ public static boolean isPvpWorld(final Collection<WorldType> worldTypes)
...
 	}
+
+	/**
+	 * Checks whether a world having a {@link Collection} of {@link WorldType}s is a Deadman world.
+	 *
+	 * @param worldTypes A {@link Collection} of {@link WorldType}s describing the given world.
+	 * @return           True if the given worldtypes of the world are a Deadman world, false otherwise.
+	 * @see Client#getWorldType()
+	 */
+	public static boolean isDeadmanWorld(final Collection<WorldType> worldTypes)
```
Nightfirecat, line 146: This can be replaced by `Client.getWorldType().contains(WorldType.DEADMAN)` since there is only one deadman world type.

runelite-api/src/main/java/net/runelite/api/WorldType.java
rangvaldr: Well it can be left, but the deadman check comes before the PVP world check. So if they're in a deadman world, `PVP_WORLD_TYPES` will never be checked.

Nightfirecat: `isPvpWorld` is used outside of the discord plugin, so this would cause adverse effects for other plugins.

rangvaldr: Fixed.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordPlugin.java
```diff
@@ -381,6 +381,12 @@ private void checkForAreaUpdate()
 			return;
 		}
 
+		if (client.getWorldType().contains(WorldType.DEADMAN))
```
deathbeam, line 384: Why are you implementing this differently than current PVP world check? Just use the same logic and in same method

deathbeam: Actually nvm, that is already contained in PVP check. I don't understand why anything extra needs to be implemented

rangvaldr: Well if you were okay with just hiding the activity completely, similar to PVP, no changes would be necessary. But in #11430 you had mentioned adding the activity line "Playing Deadman Mode", which is what's implemented here and does require the additional changes. If you just want to hide the activity, feel free to close the PR.

deathbeam: Imo it is fine to display playing DMM and I would be happy merging that, my only requirement is for it to be implemented for all cases we are skipping atm (e.g all types of PVP worlds)

rangvaldr: Gotcha.

rangvaldr: Resolved with 30eab1a.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordPlugin.java
```diff
@@ -381,6 +381,19 @@ private void checkForAreaUpdate()
 			return;
 		}
 
+		final EnumSet<WorldType> worldType = client.getWorldType();
+
+		if (worldType.contains(WorldType.DEADMAN))
```
deathbeam, line 386: this should return playing DMM even for tournament DMM worlds imo

rangvaldr: There was a comment in the code that tournament worlds were also flagged deadman.

deathbeam: Nvm, I just read the comment in PVP_WORLD_TYPES, this is fine

PR #11526, opened 2020-05-10
added Dragonfruit Pie to item stats

### LINKED ISSUE #11525: Add dragonfruit pie to item stats
Dragonfruit pie currently is not added in item stats, so it doesn't show that it heals +10 hitpoints per bite and gives a +4 boost to fletching.

ThePharros:
Closes #11525


melkypie:
This only adds it as a food item, I suggest you take a look at how other pies are added

ThePharros:
forgot Fletching stat boost, one sec

ThePharros:
> This only adds it as a food item, I suggest you take a look at how other pies are added

I only saw `.add(food(diff))` for the usage of other pie IDs

PR #11527, opened 2020-05-10
Fix check for menu entry type for copy to clipboard not working on friend messages

MarbleTurtle:
Fixes bug with menu option not appearing on friend's messages
Allows copying of the "<" and ">" characters


MarbleTurtle:
Removed modifications to tags and left in the check for action

PR #11533, opened 2020-05-10
KeyRemapping: Add space as an option to remap.

Tinusf:
Hi I wanted to be able to remap space in the KeyMapping plugin when you are talking to NPCs. So I added this as an option in the KeyMapping plugin. Also added an option for remapping shift for dropping items. I did this because I am using a splitted keyboard and my space button is actually on the right side of the keyboard which makes it awkward to hold in space whilst talking to NPCs.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -241,4 +241,15 @@ public interface KeyRemappingConfig extends Config
 	{
 		return new ModifierlessKeybind(KeyEvent.VK_ESCAPE, 0);
 	}
+
+	@ConfigItem(
+		position = 20,
+		keyName = "space",
+		name = "Space",
+		description = "The key which will replace {Space}."
+	)
+	default ModifierlessKeybind space()
+	{
+		return new ModifierlessKeybind(KeyEvent.VK_SPACE, 0);
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -246,7 +246,7 @@ public interface KeyRemappingConfig extends Config
 		position = 20,
 		keyName = "space",
 		name = "Space",
-		description = "The key which will replace {Space}."
+		description = "The key which will replace {Space} when dialogs are open."
 	)
 	default ModifierlessKeybind space()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -156,6 +156,11 @@ class KeyRemappingListener implements KeyListener
 				}
 			}
 
+			if (plugin.isDialogOpen() && config.space().matches(e))
+			{
+				mappedKeyCode = KeyEvent.VK_SPACE;
+			}
+
 			if (mappedKeyCode != KeyEvent.VK_UNDEFINED && mappedKeyCode != e.getKeyCode())
 			{
 				final char keyChar = e.getKeyChar();


Nightfirecat:
Shift is a modifier key, and should not be remappable

Tinusf:
> 
> 
> Shift is a modifier key, and should not be remappable

Oh okay, I can revert that commit, however it seemed it work fine for me actually

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
```diff
@@ -241,4 +241,15 @@ default ModifierlessKeybind esc()
 	{
 		return new ModifierlessKeybind(KeyEvent.VK_ESCAPE, 0);
 	}
+
+	@ConfigItem(
+			position = 20,
+			keyName = "space",
+			name = "Space",
+			description = "The key which will replace {Space}."
```
Nightfirecat, line 249: Your indents here are one level too far.
```suggestion
		position = 20,
		keyName = "space",
		name = "Space",
		description = "The key which will replace {Space}."
```

Tinusf:
Ah yeah, good catch

Tinusf:
Is it possible to get this merged in?

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -156,6 +156,11 @@ else if (config.esc().matches(e))
 				}
 			}
 
+			if (config.fkeyRemap() && plugin.isDialogOpen() && config.space().matches(e))
```
Adam-, line 159: Why is this tied to fkeys?

Tinusf: Ah, yea I did that because the escape key was also tied into the fkeys, and because all the F-keys and escape key is underneath the "Remap F Keys" checkbox.

Tinusf: But it is indeed a little bit confusing, maybe there should be an additional checkbox created for escape and space?

Tinusf:
Ok I added a new checkbox for other and moved both escape and space into underneath this. Take a look here:  https://i.imgur.com/FqJqY0F.png.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
Adam-: The escape key is part of the fkeys as far as the game is concerned (you can f-key map a tab to escape)

Tinusf: Ok so you'd prefer having the escape key under f-keys and just space underneath other?

Adam-:
I don't think this actually needs its own toggle, since the default is to map space->space it doesn't actually do anything. We could just add an additional check here `https://github.com/runelite/runelite/pull/11533/files#diff-81d3da003d4146f8fcf87ba2e430df45R167` that the mapped key code isn't the same as the existing code perhaps.

Tinusf:
> 
> 
> I don't think this actually needs its own toggle, since the default is to map space->space it doesn't actually do anything. We could just add an additional check here `https://github.com/runelite/runelite/pull/11533/files#diff-81d3da003d4146f8fcf87ba2e430df45R167` that the mapped key code isn't the same as the existing code perhaps.

Yea that's true, pushed a new commit now.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -156,7 +156,12 @@ else if (config.esc().matches(e))
 				}
 			}
 
-			if (mappedKeyCode != KeyEvent.VK_UNDEFINED)
+			if (plugin.isDialogOpen() && config.space().matches(e))
+			{
+				mappedKeyCode = KeyEvent.VK_SPACE;
+			}
+
+			if (mappedKeyCode != KeyEvent.VK_UNDEFINED && e.getKeyCode() != mappedKeyCode)
```
Adam-, line 164: You could add this to the code down in keyRelease too

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -308,6 +313,10 @@ else if (config.esc().matches(e))
 				{
 					mappedKeyCode = KeyEvent.VK_ESCAPE;
 				}
+				else if (config.space().matches(e))
```
Adam-, line 316: It might be good to write a test for this, since it would show you this is incorrect as it only remaps when fkeys is on.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
Tinusf: True

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
Tinusf: 👀

Adam-:
I've refactored the plugin to not require having to duplicate the reverse-logic in the release handler and then updated this. However the keybind config options seem to not work correctly when trying to set them to `space` which we probably need to look at first.

PR #11536, opened 2020-05-10
Key-remapping: Don't remap f-keys when the bank pin interface is open

loldudester:


runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingPlugin.java
```diff
@@ -138,7 +138,9 @@ boolean isDialogOpen()
 		// Most chat dialogs with numerical input are added without the chatbox or its key listener being removed,
 		// so chatboxFocused() is true. The chatbox onkey script uses the following logic to ignore key presses,
 		// so we will use it too to not remap F-keys.
-		return isHidden(WidgetInfo.CHATBOX_MESSAGES) || isHidden(WidgetInfo.CHATBOX_TRANSPARENT_LINES);
+		return isHidden(WidgetInfo.CHATBOX_MESSAGES) || isHidden(WidgetInfo.CHATBOX_TRANSPARENT_LINES)
+			// We want to block F-key remapping in the bank pin interface too, so it does not interfere with the Keyboard Bankpin feature of the Bank plugin
```
Nightfirecat, line 142: Wrap at 120 cols to be in line with the comment lines above
```suggestion
			// We want to block F-key remapping in the bank pin interface too, so it does not interfere with the
			// Keyboard Bankpin feature of the Bank plugin
```

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -918,4 +919,9 @@
 	{
 		static final int CONTAINER = 2;
 	}
+
+	static class BankPin
+	{
+		static final int CONTAINER = 1;
```
Nightfirecat, line 925: `213.1` seems to be the background where `213.0` is the container (since it contains the buttons, text, border elements, etc.)

PR #11540, opened 2020-05-10
Revises when dps overlay is shown

MarbleTurtle:
The dps counters total will change if other players are attacking a boss without showing the overlay, change to show total dmg done if it has started tracking dmg by other players. This change was in favor of having it start tracking when the local player attacked as cases where grouping there would be cases of other players dealing first damage, which would want to be tracked.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsOverlay.java
@@ -95,15 +95,14 @@ class DpsOverlay extends OverlayPanel
 	public Dimension render(Graphics2D graphics)
 	{
 		Map<String, DpsMember> dpsMembers = dpsCounterPlugin.getMembers();
-		if (dpsMembers.isEmpty())
-		{
-			return null;
-		}
-
 		boolean inParty = !partyService.getMembers().isEmpty();
 		boolean showDamage = dpsConfig.showDamage();
 		DpsMember total = dpsCounterPlugin.getTotal();
 		boolean paused = total.isPaused();
+		if (dpsMembers.isEmpty() && total.getStart() == total.getEnd())
+		{
+			return null;
+		}
 
 		final String title = (inParty ? "Party " : "") + (showDamage ? "Damage" : "DPS") + (paused ? " (paused)" : "");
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsOverlay.java
```diff
@@ -95,15 +95,14 @@ public void onMouseOver()
...
-		if (dpsMembers.isEmpty())
-		{
-			return null;
-		}
-
 		boolean inParty = !partyService.getMembers().isEmpty();
 		boolean showDamage = dpsConfig.showDamage();
 		DpsMember total = dpsCounterPlugin.getTotal();
 		boolean paused = total.isPaused();
+		if (dpsMembers.isEmpty() && total.getStart() == total.getEnd())
```
Adam-, line 102: This is probably wrong as its a reference comparison of Instants. What are you trying to do here? Maybe just check && total.getDamage() == 0 instead?

MarbleTurtle: If 0 dmg is dealt that would still start total's timer which would mean that the timer is still started but not visible to the player which is what I was aiming to fix.

Adam-: This seems fine then I guess. Maybe make an `isReset` in DpsMember then because this is currently relying on an implementation detail of reset() using the same instant for start/end which is very subtle. And please update the commit message to describe what behavior you are trying to change.

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsMember.java
```diff
@@ -90,8 +96,11 @@ void reset()
 	{
 		damage = 0;
 		start = end = Instant.now();
+		reset = true;
 	}
 
+	boolean isReset() { return reset; }
```
Nightfirecat, line 102: This class is already annotated `@Getter`, you don't need to create an explicit getter.

Adam-: this can just be `Objects.equals(start, end)`

PR #11541, opened 2020-05-10
Remove bounty hunter emblem tier from player name for lookup

### LINKED ISSUE #11440: hiscore: Bounty Hunter emblem indicators cause hiscore lookups to use wrong name
Searched for this issue, didn't seem to find one for this. Feel free to close if this is a duplicate.

**Describe the bug**
In the new version of Bounty Hunter, players who have emblems have their name strings modified to have the tier of emblem they are currently carrying appended to their username. This causes the 'lookup' option on players to use the wrong name and fail.

**To Reproduce**
1. Go to a Bounty Hunter world
2. Right click someone with an emblem skull and click 'lookup'

**Expected behavior**
The correct player name should be used in the lookup.

**Screenshots**
Example image

In this instance, the player's actual name is 'lite tur', but the lookup option sends 'lite tur4' to the hiscore tab.

Linux Mint 19, Java 8
RL 1.6.13
Launcher 1.6.1

bfmoatbio:
A player in the wilderness of a bounty hunter world with a bounty hunter
emblem in their inventory will have their name decorated in the player
menu with an emblem icon and the tier of emblem they are carrying. Before
removing any tags, that name will look like this:

<col=ffffff>bfmoatbio10<col=00ff>  (level-126)

where "10" represents the emblem icon () and tier (10).
removeTags will remove the icon tag but leave the emblem tier, resulting
in an incorrect username returned by getMenuTarget (in the case of the
example, bfmoatbio10).

Before calling removeTags on a player menu target, perform a special
remove of the bounty hunter emblem icon and its tier, if present.

Closes #11440

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.menus;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
@@ -33,10 +34,12 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
+import net.runelite.api.IconID;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
 import net.runelite.api.NPCComposition;
@@ -61,6 +64,8 @@ public class MenuManager
 	private static final int IDX_LOWER = 4;
 	private static final int IDX_UPPER = 8;
 
+	private static final Pattern BOUNTY_EMBLEM_TAG_AND_TIER_REGEXP = Pattern.compile(String.format("%s[1-9]0?", IconID.BOUNTY_HUNTER_EMBLEM.toString()));
+
 	private final Client client;
 	private final EventBus eventBus;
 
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
```diff
@@ -282,6 +282,14 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 		eventBus.post(playerMenuOptionClicked);
 	}
 
+	private String removeTagsFromPlayerMenuTarget(String target)
```
melkypie, line 285: I don't think it warrants a seperate function since you are only adding a single line. I think a comment would be better.

runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
```diff
@@ -269,11 +273,12 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 			}
 		}
 
-		String target = event.getMenuTarget();
+		// removes bounty hunter emblem tag and tier from player name, e.g. username<img=20>5 -> username
```
Nightfirecat, line 276:
```suggestion
		// removes bounty hunter emblem tag and tier from player name, e.g. username<img=20>5<col=40ff00> (level-42) -> username<col=40ff00> (level-42)
```

runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
```diff
@@ -269,11 +273,12 @@ public void onMenuOptionClicked(MenuOptionClicked event)
...
+		String targetWithoutBHEmblemAndTier = BOUNTY_EMBLEM_TAG_AND_TIER_REGEXP.matcher(event.getMenuTarget()).replaceAll("");
```
Nightfirecat, line 277: I think leaving this named `target` is fine.
```suggestion
		String target = BOUNTY_EMBLEM_TAG_AND_TIER_REGEXP.matcher(event.getMenuTarget()).replaceAll("");
```

PR #11547, opened 2020-05-11
Add potion dosages to Item Identification plugin

ThePharros:
Adds dosage text to potion icons to better distinguish between visually similar doses. Some people may find it useful, most notably for differentiating between (3) and (4) doses.

https://imgur.com/a/DWdxe2r

Please review code thoroughly as I feel the method I went with is rather janky, and something is telling me there is a better solution lol.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
@@ -25,7 +25,6 @@
 package net.runelite.client.plugins.itemidentification;
 
 import com.google.common.collect.ImmutableMap;
-
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


ThePharros:
> This is very janky. As written, it seems to think that unfinished potions are 4-dose potions and is overall very clunky. I would recommend removing all added fields/static structures in `ItemIdentification` and using itemManager to examine the item's composition in the overlay to determine whether it's a dosed potion.

Alright, I've canonicalized via ItemManager. What method would you suggest for determining the dosages? I have the itemID for each potion variation. Should I instead make an enum for each numerical dosage and return based on that?

ThePharros:
The regex/mapping/enum alternatives seem to be beyond my scope, just going to close the PR. If someone wishes to create the proper method down the road, then by all means feel free to do so.

PR #11556, opened 2020-05-12
Add duplicate chat options to chatfilter

CoreyForsyth:
A few things to note:

1. I decided to use a cache for keeping count of the messages
- This allows it to count more than just the chat buffer length, so if you are logged in for many hours, message counts can go into the hundreds or thousands with no problem
- This also prevents modification of the chat buffer itself, so it will retain all of the chat history. This means if the plugin is turned off after consolidating the chats, the chat will go back to normal
- To prevent this cache from growing unbounded, I added a configurable max size. I defaulted to 500, however this is left up to the user. This size corresponds to unique messages, not total messages.

2. Public chat consolidation is configurable
- Only the "spammy" message types are consolidated by default: ENGINE, GAMEMESSAGE, NPC_EXAMINE, OBJECT_EXAMINE, SPAM.
- PUBLICCHAT is configurable on/off, default off

3. Public chat can be hidden after X amount of duplicates
- If the player so chooses, they can filter out messages if they are repeated a bunch. I tested this out in world 1 and had great success. After a bit of standing around the GE, the duplicate messages were filtered out and only regular conversation was left

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
@@ -108,4 +108,26 @@ public interface ChatFilterConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "collapseGameChat",
+		name = "Collapse Game Chat",
+		description = "Collapse duplicate game chat messages into a single line",
+		position = 9
+	)
+	default boolean collapseGameChat()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "collapsePlayerChat",
+		name = "Collapse Player Chat",
+		description = "Collapse duplicate player chat messages into a single line",
+		position = 10
+	)
+	default boolean collapsePlayerChat()
+	{
+		return false;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
@@ -130,4 +130,15 @@ public interface ChatFilterConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "maxRepeatedPublicChats",
+		name = "Max repeated public chats",
+		description = "Block player chat message if repeated this many times. 0 is off",
+		position = 11
+	)
+	default int maxRepeatedPublicChats()
+	{
+		return 0;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
@@ -28,23 +28,36 @@ package net.runelite.client.plugins.chatfilter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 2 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
```diff
@@ -292,4 +295,92 @@ boolean shouldFilterByName(final String playerName)
...
+	private boolean shouldCollapseMessageType(ChatMessageType chatMessageType)
+	{
+		switch (chatMessageType)
+		{
+			case ENGINE:
+			case GAMEMESSAGE:
+			case NPC_EXAMINE:
+			case OBJECT_EXAMINE:
+			case SPAM:
+				return true;
```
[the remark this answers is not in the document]
CoreyForsyth: What other chat types would you want to collapse? I believe the above are all of the "regular" game chats that are repetitive. Public chat is the only other kind of chat I would think would make sense to collapse.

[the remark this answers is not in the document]
raiyni: There is no way to turn off the collapsing for these types other than turning the plugin off.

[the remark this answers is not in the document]
CoreyForsyth: Ahh, my apologies. I just pushed a change that adds that as a configuration. I left that out because I initially made this as a separate plugin but I ended up merging it into the chatfilter plugin instead.

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -108,4 +110,60 @@ default boolean filterLogin()
...
+	default Color chatCountColor()
+	{
+		return new Color(0xA00080);
+	}
+
+	@ConfigItem(
+			keyName = "collapseGameChat",
+			name = "Collapse Game Chat",
+			description = "Configures if duplicate game chat can be collapsed",
+			position = 9
```
Nightfirecat, line 129: Your indents here are one level too far.
```suggestion
		keyName = "collapseGameChat",
		name = "Collapse Game Chat",
		description = "Configures if duplicate game chat can be collapsed",
		position = 9
```

Nightfirecat: Also the description for the added configs should be written into more active statements, eg. "Collapse duplicate game chat messages into a single line"

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -108,4 +110,60 @@ default boolean filterLogin()
...
+	}
+
+	@ConfigItem(
+		keyName = "maxTrackedChats",
+		name = "Max Tracked Chats",
+		description = "Configures the maximum number of chats that are tracked. Suggested 500",
+		position = 12
+	)
+	@Range(min = 100)
+	default int maxTrackedChats()
```
Nightfirecat, line 165: Will users care about this?

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -108,4 +109,37 @@ default boolean filterLogin()
...
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "chatCountColor",
+		name = "Collapsed Count Color",
+		description = "Configures the color for the number of collapsed messages",
+		position = 8
+	)
+	default Color chatCountColor()
```
Adam-, line 119: IMO this seems kind of useless? If you just append the count it would take the color of whatever the message was, probably

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -108,4 +109,37 @@ default boolean filterLogin()
...
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "collapsePlayerChat",
+		name = "Collapse Player Chat",
+		description = "Collapse duplicate player chat messages into a single line",
+		position = 10
+	)
+	default boolean collapsePlayerChat()
```
Adam-, line 141: This doesn't belong in this commit

PR #11557, opened 2020-05-12
Float loot tracker groups to top of list when obtaining drop

### LINKED ISSUE #8519: Display grouped loot tracker entries by most recent loot received
Currently, when in grouped view, the loot tracker will sort by the group's earliest loot received. It would be more useful to change the sorting so that the most recent loot received from each group determined the sort order.

Vidyogamasta:
Loot tracker panel now attempts to push the related group box to the top of the panel upon obtaining a drop

Closes #8519

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
@@ -502,7 +502,7 @@ class LootTrackerPanel extends PluginPanel
 			{
 				if (box.matches(record))
 				{
-					//float the matched box to the top of the UI list if it's not already first
+					// float the matched box to the top of the UI list if it's not already first
 					int idx = logsContainer.getComponentZOrder(box);
 					if (idx > 0)
 					{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPanel.java
```diff
@@ -502,6 +502,10 @@ private LootTrackerBox buildBox(LootTrackerRecord record)
 			{
 				if (box.matches(record))
 				{
+					//float the matched box to the top of the UI list
+					logsContainer.remove(box);
```
Adam-, line 506: Since the common case will not require reordering you can optimize this via
```java
int idx = logsContainer.getComponentZOrder(box);
if (idx > 0) {
  logsContainer.remove(idx);
  logsContainer.add(box, 0);
}
```

Vidyogamasta: Good point! Implemented the change as recommended and verified it's still functioning as expected.

Nightfirecat:
Does this close #8519?

Vidyogamasta:
> Does this close #8519?

Yeah, if merged this change would make it to where when a loot source is updated, it's pushed to the top. This gives it the effect of always being sorted by most recent.

Issue #11570, opened 2020-05-13
King Black Dragon, Corp, and maybe other respawn timers do not work in instances.


jesse1412:
The npc indicators plugin doesn't show respawn times for the instanced variants of KBD and corp (however it does for the non-instanced version of KBD). I would assume most instanced bosses have the same issue (Krakan and Alchemical Hydra, notably).

To reproduce: Tag the bosses in an instance with respawn timers enabled, then kill the boss twice. The spawn tile should be highlighted with a timer, but it isn't. Additionally, if manually tagging with the menu option, the NPC becomes untagged after being killed.

Note, the boss timers plugin uses manually set timers for these bosses which gets around the issue, but the npc indicators plugin doesn't work.

PR #11571, opened 2020-05-13
remove respawn timers for non-whitelisted NPCs in instances

jesse1412:
Add a whitelist for NPCs that need respawn timers and are instanced.

This list includes instanced corp, KBD and krakan, as well as hydra, although it should be noted that of these bosses, some (all?) do not work correctly with the current npc indicators plugin anyway. See #11570 

The whitelist is currently redundant until #11570 is resolved, as the whitelisted bosses do not currently work with the plugin anyway and are possibly the only instanced NPCs where a respawn timer would be useful. It's still useful to stop respawn timers appearing in instances though, as they're almost always incorrectly created, notably in the inferno when tagging NPCs.

If fixing #11570 is infeasible, it may be best to disable all spawn timers in instances, unless there are notable use-cases that currently work with the plugin in instances.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
@@ -50,6 +50,7 @@ import net.runelite.api.MenuAction;
 import static net.runelite.api.MenuAction.MENU_ACTION_DEPRIORITIZE_OFFSET;
 import net.runelite.api.MenuEntry;
 import net.runelite.api.NPC;
+import net.runelite.api.NpcID;
 import net.runelite.api.coords.WorldPoint;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
As was discussed on discord, I don't think this is a feasible solution to this problem? iirc 11570 is not fixable on the client side

jesse1412:
> As was discussed on discord, I don't think this is a feasible solution to this problem? iirc 11570 is not fixable on the client side

Yeah this seems to be the case. I can't think of a clean solution, although a messy solution could be to track the NPCs in the whitelist by their NPC IDs rather than index (as there should only ever be one of those NPCs in any instance). Implementing it would be a bit messy for my liking, and it wouldn't be very robust (could only handle instanced spawn timers for NPCs that can only exist one at a time).

I'm closing this PR for now as there doesn't seem to be a nice solution.

PR #11574, opened 2020-05-13
player-indicators: don't decorate non-player custom menu options

### LINKED ISSUE #11448: The "edit-tag" option is formatted the same as your name if you are in a cc
If you aren't in a cc, your edit-tags option looks normal: 

If you are in a cc, your edit-tags option looks like your name in cc (different colour, purple in my case), and has your rank icon in front of it:

devLotto:
Fixes #11448 which appeared because of https://github.com/runelite/runelite/commit/087b2ac9695778464962f6d28b02db7c2e3ae9cd. The issue is that custom menu entries of type RUNELITE are decorated even if they aren't related to a player.

To fix the issue, the MenuAction.RUNELITE menu type is split into RUNELITE and RUNELITE_PLAYER. With RUNELITE_PLAYER being assigned to player options only, the Player Indicators plugin should no longer try to decorate general menu entries such as Edit-tags, and also not the right-click Lookup in chat or other menu entries that aren't player options.

Why this issue wasn't observed before https://github.com/runelite/runelite/commit/087b2ac9695778464962f6d28b02db7c2e3ae9cd: 
The plugin was changed in that commit to use the ClientTick event instead of MenuEntryAdded for going through the menu entries. Custom menu entries don't post a corresponding MenuEntryAdded event when they are added to the menu, so the plugin would not decorate them.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
@@ -253,9 +253,10 @@ public class MenuManager
 	@Subscribe
 	public void onMenuOptionClicked(MenuOptionClicked event)
 	{
-		if (event.getMenuAction() != MenuAction.RUNELITE)
+		if (event.getMenuAction() != MenuAction.RUNELITE
+			&& event.getMenuAction() != MenuAction.RUNELITE_PLAYER)
 		{
-			return; // not a player menu
+			return; // not a managed widget option or custom player option
 		}
 
 		int widgetId = event.getWidgetId();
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/menus/MenuManager.java
```diff
@@ -247,7 +247,8 @@ private void removeNpcOption(NPCComposition composition, String npcOption)
 	@Subscribe
 	public void onMenuOptionClicked(MenuOptionClicked event)
 	{
-		if (event.getMenuAction() != MenuAction.RUNELITE)
+		if (event.getMenuAction() != MenuAction.RUNELITE
+			&& event.getMenuAction() != MenuAction.RUNELITE_PLAYER)
 		{
 			return; // not a player menu
```
Nightfirecat, line 253: Adam commented in discord and I agree somewhat that it's odd that this comment isn't updated to reflect the code that this method actually runs. This should mention that player and managed widget menus are handled here.

devLotto:
Updated

PR #11577, opened 2020-05-13
Add all other pearl rod animations

### LINKED ISSUE #11575: Fishing Session with Pearl Fishing Rod does not track Angler fishing activity
This issue was brought up about a month ago with #11283  and was supposedly resolved with #11292 

I don't have access to RuneLite developer mode, but I suspect that there might be a different animation ID associated with Angler fishing(or maybe other types of bait fishing) vs. Eel fishing.

Hydrox6:
Closes #11575

There are 2 animations for each pearl rod (except Oily). Usually it
seems that only one is used when actually fishing, which were the ones
present before this commit. However, it seems that Jagex is not
consistent about what animations are used, as Anglerfish fishing uses
the other animation of the normal pearl rod. Given that the other two
exist, and they're technically usable (and can be used as the starting
animation), it's only a matter of time before Jagex accidentally gets
them the wrong way round again, so this is a pre-emptive measure. It
also means that the UI won't show "Not Fishing" for a split second when
casting out.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -205,6 +205,9 @@ public class IdleNotifierPlugin extends Plugin
 			case FISHING_PEARL_ROD:
 			case FISHING_PEARL_FLY_ROD:
 			case FISHING_PEARL_BARBARIAN_ROD:
+			case FISHING_PEARL_ROD_2:
+			case FISHING_PEARL_FLY_ROD_2:
+			case FISHING_PEARL_BARBARIAN_ROD_2:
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -205,6 +205,9 @@ public void onAnimationChanged(AnimationChanged event)
 			case FISHING_PEARL_ROD:
 			case FISHING_PEARL_FLY_ROD:
 			case FISHING_PEARL_BARBARIAN_ROD:
+			case FISHING_PEARL_ROD:
```
raiyni, line 208: I think these are typos?

Hydrox6: Whoops, good catch

PR #11580, opened 2020-05-14
screenshot: fix method for naming quest complete screenshots

### LINKED ISSUE #5353: Screenshots for finishing quests have incorrect filenames
Screenshot gets saved as "ds with My Arm.png"

ds with My Arm

johannfrias:
Closes #5353. See my original post, since I make references to it here.

I wrote an example solution that applies Option 2 to exception #*1*, #*2* and #*5*, Option 2a to exception #*4*, and Option 1 on the other exceptions. This means there are 2 constants and 4 if blocks worth of extra code from the base solution. When I talk simplicity vs accuracy, I made the file 2 constants and 4 if blocks less simple to make 18 names more accurate, but that begs the question: was that effort worth implementing? I chose these exceptions as a proof of concept of what "extra" code means, and they seemed the most simple to handle. (Some exceptions, using option 2, would use ~~dicts~~ maps, which hinders simplicity quite a bit; I wouldn't prefer to use maps.)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -598,7 +598,7 @@ public class ScreenshotPlugin extends Plugin
 		Matcher quest_match_final = quest_match_1.matches() ? quest_match_1 : quest_match_2;
 		if (!quest_match_final.matches())
 		{
-			return null;
+			return "Quest(quest not found)";
 		}
 
 		String quest = quest_match_final.group("quest");
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -582,6 +582,7 @@ public class ScreenshotPlugin extends Plugin
 	 *
 	 * @return Shortened string in the format "Quest(The Corsair Curse)"
 	 */
+	@VisibleForTesting
 	String parseQuestCompletedWidget()
 	{
 		Widget questChild = client.getWidget(WidgetInfo.QUEST_COMPLETED_NAME_TEXT);
@@ -606,11 +607,11 @@ public class ScreenshotPlugin extends Plugin
 
 		if (verb.contains("kind of"))
 		{
-			quest = quest + " partial completion";
+			quest += " partial completion";
 		}
 		else if (verb.contains("completely"))
 		{
-			quest = quest + " II";
+			quest += " II";
 		}
 
 		if (RFD_TAGS.stream().anyMatch((quest + verb)::contains))
@@ -620,7 +621,7 @@ public class ScreenshotPlugin extends Plugin
 
 		if (WORD_QUEST_IN_NAME_TAGS.stream().anyMatch(quest::contains))
 		{
-			quest = quest + " Quest";
+			quest += " Quest";
 		}
 
 		return "Quest(" + quest + ")";
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -242,7 +242,7 @@ public class ScreenshotPlugin extends Plugin
 		}
 		else if (client.getWidget(WidgetInfo.QUEST_COMPLETED_NAME_TEXT) != null)
 		{
-			fileName = parseQuestCompletedWidget();
+			fileName = parseQuestCompletedWidget(client.getWidget(WidgetInfo.QUEST_COMPLETED_NAME_TEXT).getText());
 			screenshotSubDir = "Quests";
 		}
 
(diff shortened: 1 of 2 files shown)


raiyni:
This needs tests.

johannfrias:
I'll be adding tests shortly, just trying to understand how the test file works. How many tests suffice, since there's a lot of cases to consider?

raiyni:
You are adding at least 4 specials cases, so that many?

makingcomment:
It's very possible that I'm misreading or misunderstanding the code, but it looks to me like these changes result in a screenshot not being taken in the [probably unlikely] event a new (or existing, if jagex makes changes) quest is ever added where the name doesn't match either of the two regexes. The current method still results in a screenshot, albeit one with an improper name, and that behavior may be worth keeping.

johannfrias:
@makingcomment Good point. In my private testing, I tried nonexistent quest texts that could possibly break the regexes. `QUEST_PATTERN_1` texts, as long as "have" or " 've " is near the beginning, seems to always export something. `QUEST_PATTERN_2` texts break when an irregular past tense verb is used. I propose an easy fix: when I check for `!quest_match_final.matches()`, I return some default string like `"Quest(Unknown quest)"` instead of `null`. Thoughts?

runelite-client/src/test/java/net/runelite/client/plugins/screenshot/ScreenshotPluginTest.java
```diff
@@ -245,4 +248,144 @@ public void testHunterLevel2()
...
+		assertEquals("Quest(The Corsair Curse)", screenshotPlugin.parseQuestCompletedWidget());
+
+		WidgetLoaded event = new WidgetLoaded();
+		event.setGroupId(QUEST_COMPLETED_GROUP_ID);
+		screenshotPlugin.onWidgetLoaded(event);
+
+		GameTick tick = new GameTick();
+		screenshotPlugin.onGameTick(tick);
+
+		verify(drawManager).requestNextFrameListener(any(Consumer.class));
```
Nightfirecat, line 269: All of this code is basically only testing `parseQuestCompletedWidget() != null` in the context of the calling code, so it can be removed.
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -574,6 +576,56 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
+
+		String text = questChild.getText();
+		// "You have completed The Corsair Curse!"
+		Matcher quest_match_1 = QUEST_PATTERN_1.matcher(text);
+		// "'One Small Favour' completed!"
+		Matcher quest_match_2 = QUEST_PATTERN_2.matcher(text);
+		Matcher quest_match_final = quest_match_1.matches() ? quest_match_1 : quest_match_2;
+		if (!quest_match_final.matches())
+		{
+			return "Quest(quest not found)";
```
Nightfirecat, line 601: Write a test for this code path.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -574,6 +576,56 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
+		}
+
+		String quest = quest_match_final.group("quest");
+		String verb = quest_match_final.group("verb") != null ? quest_match_final.group("verb") : "";
+
+		if (verb.contains("kind of"))
+		{
+			quest = quest + " partial completion";
```
Nightfirecat, line 609:
```suggestion
			quest += " partial completion";
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -574,6 +576,56 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
+		}
+		else if (verb.contains("completely"))
+		{
+			quest = quest + " II";
```
Nightfirecat, line 613:
```suggestion
			quest += " II";
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -574,6 +576,56 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
+		}
+
+		if (RFD_TAGS.stream().anyMatch((quest + verb)::contains))
+		{
+			quest = "Recipe for Disaster - " + quest;
+		}
+
+		if (WORD_QUEST_IN_NAME_TAGS.stream().anyMatch(quest::contains))
+		{
+			quest = quest + " Quest";
```
Nightfirecat, line 623:
```suggestion
			quest += " Quest";
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -574,6 +576,57 @@ String parseLevelUpWidget(WidgetInfo levelUpLevel)
...
 	}
 
+	/**
+	 * Parses a WidgetInfo pointing to the second widget of the quest-completed dialog
+	 * into a shortened string for filename usage.
+	 *
+	 * @return Shortened string in the format "Quest(The Corsair Curse)"
+	 */
+	@VisibleForTesting
+	String parseQuestCompletedWidget()
```
Adam-, line 586: This should probably just `String parseQuestCompletedText(String text)` which would make it easier to test due to not having the dep on the widget

PR #11583, opened 2020-05-14
Remove respawn indicators in instances

jesse1412:
Removes respawn timer for NPCs in instances. 

As discussed in #11570 respawn timers for NPCs in instances don't currently work with the plugin, and additionally lead to index collisions (which 01f134795d2b1834d4eb4d720a83ef5be6c75ed6 resolved). Removing indicators for NPCs in instances should resolve both the index collisions, and invalid NPC respawn indicators appearing in instances (notable example being mass-tagged inferno NPCs).

If a clean solution is ever found for making death timers work in instances, it may be worth re-opening #11571


jesse1412:
Seems to have been fixed in ba73d704338c87dde5591abbf881c8b91cd4aed1

PR #11594, opened 2020-05-14
Improve drift net plugin

bartpeeters:
Changes to drift net plugin:
- Only untag fish shoal when tick timeout reached or fish shoal is adjacent to a drift net that is not catching fish (= full or unset). 
You can still sometimes have a shoal that is shown as tagged while it is actually not, this rarely happens so I'm not able to determine how this happens. After the tick timeout it will still become untagged.
- Add option to highlight Annette when no more drift nets are left in inventory.

I don't have a lot of experience with Java so any feedback is appreciated!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
@@ -25,10 +25,13 @@
  */
 package net.runelite.client.plugins.driftnet;
 
+import java.util.Set;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
+import lombok.Setter;
 import net.runelite.api.GameObject;
 import net.runelite.api.Varbits;
+import net.runelite.api.coords.WorldPoint;
 
 @Data
 @RequiredArgsConstructor
@@ -37,10 +40,23 @@ class DriftNet
 	private final int objectId;
 	private final Varbits statusVarbit;
 	private final Varbits countVarbit;
+	private final Set<WorldPoint> adjacentTiles;
 
 	private GameObject net;
 	private DriftNetStatus status;
 	private int count;
+	@Setter
+	private DriftNetStatus prevTickStatus;
+
+	// Nets that are not accepting fish are those currently not accepting, or those which were not
+	// accepting in the previous tick. (When a fish shoal is 2 tiles adjacent to a drift net and is
+	// moving to a net that is just being setup it will be denied even though the net is currently
+	// in the CATCHING status)
+	boolean isNotAcceptingFish()
+	{
+		return (status != DriftNetStatus.CATCH && status != DriftNetStatus.SET) ||
+			(prevTickStatus != DriftNetStatus.CATCH && prevTickStatus != DriftNetStatus.SET);
+	}
 
 	String getFormattedCountText()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
@@ -76,12 +76,12 @@ public interface DriftNetConfig extends Config
 	)
 	@Range(
 		min = 1,
-		max = 60
+		max = 100
 	)
 	@Units(Units.TICKS)
 	default int timeoutDelay()
 	{
-		return 10;
+		return 60;
 	}
 
 	@ConfigItem(
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
@@ -94,4 +94,26 @@ public interface DriftNetConfig extends Config
 	{
 		return Color.CYAN;
 	}
+
+	@ConfigItem(
+		keyName = "tagAnnette",
+		name = "Tag Annette when no nets in inventory",
+		description = "Tag Annette when no nets in inventory",
+		position = 6
+	)
+	default boolean tagAnnetteWhenNoNets()
+	{
+		return true;
+	}
+
+	@ConfigItem(
+		keyName = "annetteTagColor",
+		name = "Annette tag color",
+		description = "Color of Annette tag",
+		position = 7
+	)
+	default Color annetteTagColor()
+	{
+		return Color.RED;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
@@ -26,12 +26,15 @@
 package net.runelite.client.plugins.driftnet;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 import lombok.Getter;
 import net.runelite.api.Actor;
@@ -99,8 +102,19 @@ public class DriftNetPlugin extends Plugin
 	private Map<NPC, Integer> taggedFish = new HashMap<>();
 	@Getter
 	private final List<DriftNet> NETS = ImmutableList.of(
-		new DriftNet(NullObjectID.NULL_31433, Varbits.NORTH_NET_STATUS, Varbits.NORTH_NET_CATCH_COUNT),
-		new DriftNet(NullObjectID.NULL_31434, Varbits.SOUTH_NET_STATUS, Varbits.SOUTH_NET_CATCH_COUNT));
+		new DriftNet(NullObjectID.NULL_31433, Varbits.NORTH_NET_STATUS, Varbits.NORTH_NET_CATCH_COUNT, ImmutableSet.of(
+			new WorldPoint(3746, 10297, 1),
+			new WorldPoint(3747, 10297, 1),
+			new WorldPoint(3748, 10297, 1),
+			new WorldPoint(3749, 10297, 1)
+		)),
+		new DriftNet(NullObjectID.NULL_31434, Varbits.SOUTH_NET_STATUS, Varbits.SOUTH_NET_CATCH_COUNT, ImmutableSet.of(
+			new WorldPoint(3742, 10288, 1),
+			new WorldPoint(3742, 10289, 1),
+			new WorldPoint(3742, 10290, 1),
+			new WorldPoint(3742, 10291, 1),
+			new WorldPoint(3742, 10292, 1)
+		)));
 
 	@Getter
 	private boolean inDriftNetArea;
@@ -190,12 +204,6 @@ public class DriftNetPlugin extends Plugin
 			net.setStatus(status);
 			net.setCount(count);
 		}
-
-		// When you collect any loot, all tags become invalidated
-		if (client.getVar(Varbits.DRIFT_NET_COLLECT) != 0)
-		{
-			taggedFish.clear();
-		}
 	}
 
 	@Subscribe
@@ -211,6 +219,17 @@ public class DriftNetPlugin extends Plugin
 		}
 	}
 
+	private boolean isFishNextToNet(NPC fish, Collection<DriftNet> nets)
+	{
+		final WorldPoint fishTile = WorldPoint.fromLocalInstance(client, fish.getLocalLocation());
+		return nets.stream().anyMatch(net -> net.getAdjacentTiles().contains(fishTile));
+	}
+
+	private boolean isTagExpired(Integer tick)
+	{
+		return tick + config.timeoutDelay() < client.getTickCount();
+	}
+
 	@Subscribe
 	public void onGameTick(GameTick tick)
 	{
@@ -219,8 +238,16 @@ public class DriftNetPlugin extends Plugin
 			return;
 		}
 
-		final int currentTickCount = client.getTickCount();
-		taggedFish.entrySet().removeIf(entry -> entry.getValue() + config.timeoutDelay() < currentTickCount);
+		List<DriftNet> closedNets = NETS.stream()
+			.filter(DriftNet::isNotAcceptingFish)
+			.collect(Collectors.toList());
+
+		taggedFish.entrySet().removeIf(entry ->
+			isTagExpired(entry.getValue()) ||
+			isFishNextToNet(entry.getKey(), closedNets)
+		);
+
+		NETS.forEach(net -> net.setPrevTickStatus(net.getStatus()));
 
 		armInteraction = false;
 	}
@@ -271,6 +298,7 @@ public class DriftNetPlugin extends Plugin
 	{
 		final NPC npc = event.getNpc();
 		fish.remove(npc);
+		taggedFish.remove(npc);
 	}
 
 	@Subscribe
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
@@ -39,9 +39,13 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameObject;
 import net.runelite.api.GameState;
+import net.runelite.api.InventoryID;
+import net.runelite.api.ItemContainer;
+import net.runelite.api.ItemID;
 import net.runelite.api.NPC;
 import net.runelite.api.NpcID;
 import net.runelite.api.NullObjectID;
+import net.runelite.api.ObjectID;
 import net.runelite.api.Player;
 import net.runelite.api.Varbits;
 import net.runelite.api.coords.WorldPoint;
@@ -51,6 +55,7 @@ import net.runelite.api.events.GameObjectSpawned;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.InteractingChanged;
+import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.NpcDespawned;
 import net.runelite.api.events.NpcSpawned;
 import net.runelite.api.events.VarbitChanged;
@@ -101,6 +106,12 @@ public class DriftNetPlugin extends Plugin
 	private boolean inDriftNetArea;
 	private boolean armInteraction;
 
+	@Getter
+	private boolean driftNetsInInventory;
+
+	@Getter
+	private GameObject annette;
+
 	@Provides
 	DriftNetConfig provideConfig(ConfigManager configManager)
 	{
@@ -132,6 +143,10 @@ public class DriftNetPlugin extends Plugin
 	@Subscribe
 	public void onGameStateChanged(GameStateChanged event)
 	{
+		if (event.getGameState() != GameState.LOGGED_IN)
+		{
+			annette = null;
+		}
 		switch (event.getGameState())
 		{
 			case LOGIN_SCREEN:
@@ -262,6 +277,11 @@ public class DriftNetPlugin extends Plugin
 	public void onGameObjectSpawned(GameObjectSpawned event)
 	{
 		GameObject object = event.getGameObject();
+		if (object.getId() == ObjectID.ANNETTE)
+		{
+			annette = object;
+		}
+
 		for (DriftNet net : NETS)
 		{
 			if (net.getObjectId() == object.getId())
@@ -275,6 +295,11 @@ public class DriftNetPlugin extends Plugin
 	public void onGameObjectDespawned(GameObjectDespawned event)
 	{
 		GameObject object = event.getGameObject();
+		if (object == annette)
+		{
+			annette = null;
+		}
+
 		for (DriftNet net : NETS)
 		{
 			if (net.getObjectId() == object.getId())
@@ -282,6 +307,18 @@ public class DriftNetPlugin extends Plugin
 				net.setNet(null);
 			}
 		}
+	}
+
+	@Subscribe
+	public void onItemContainerChanged(final ItemContainerChanged event)
+	{
+		final ItemContainer itemContainer = event.getItemContainer();
+		if (itemContainer != client.getItemContainer(InventoryID.INVENTORY))
+		{
+			return;
+		}
+
+		driftNetsInInventory = itemContainer.contains(ItemID.DRIFT_NET);
 	}
 
 	private boolean checkArea()
--- a/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetOverlay.java
@@ -28,6 +28,7 @@ import java.awt.Dimension;
 import java.awt.Graphics2D;
 import java.awt.Shape;
 import javax.inject.Inject;
+import net.runelite.api.GameObject;
 import net.runelite.api.NPC;
 import net.runelite.api.Point;
 import net.runelite.client.ui.overlay.Overlay;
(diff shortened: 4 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -94,13 +100,30 @@
...
 
 	@Getter
 	private boolean inDriftNetArea;
 	private boolean armInteraction;
 
+	@Getter
+	private boolean driftNetsInInventory = false;
+
+	@Getter
+	private GameObject annetteObject;
```
dekvall, line 125: You can just name this annette

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -175,12 +198,6 @@ private void updateDriftNetVarbits()
 			net.setStatus(status);
 			net.setCount(count);
 		}
-
-		// When you collect any loot, all tags become invalidated
-		if (client.getVar(Varbits.DRIFT_NET_COLLECT) != 0)
-		{
-			taggedFish.clear();
-		}
```
dekvall, line 183: Are you sure this wrong? It seemed correct when i tested it, though it's possible i was mistaken.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -241,6 +282,11 @@ private void tagFish(Actor fish)
 		taggedFish.put(fishTarget, client.getTickCount());
 	}
 
+	private boolean isAcceptingFish(DriftNetStatus status)
+	{
+		return status != DriftNetStatus.CATCH && status != DriftNetStatus.SET;
+	}
```
dekvall, line 288: This can just be a method of `DriftNet`

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -284,6 +336,20 @@ public void onGameObjectDespawned(GameObjectDespawned event)
...
+	public void onItemContainerChanged(final ItemContainerChanged event)
+	{
+
+		if (event.getItemContainer() != client.getItemContainer(InventoryID.INVENTORY))
+		{
+			return;
+		}
+
+		ItemContainer itemContainer = event.getItemContainer();
+		driftNetsInInventory = itemContainer.contains(ItemID.DRIFT_NET);
```
dekvall, line 349:
```suggestion
		final ItemContainer itemContainer = event.getItemContainer();
		if (itemContainer != client.getItemContainer(InventoryID.INVENTORY))
		{
			return;
		}
		driftNetsInInventory = itemContainer.contains(ItemID.DRIFT_NET);
```

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
```diff
@@ -76,12 +76,12 @@ default boolean highlightUntaggedFish()
...
 	@Range(
 		min = 1,
-		max = 60
+		max = 100
 	)
 	@Units(Units.TICKS)
 	default int timeoutDelay()
 	{
-		return 10;
+		return 100;
```
dekvall, line 84: 60 seconds is a very long time, i might have undershot the correct time with 6 seconds but what's the motivation for a tenfold increase?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,6 +221,30 @@ public void onGameTick(GameTick tick)
 			return;
 		}
 
+		for (DriftNet net : NETS)
+		{
+			DriftNetStatus netStatus = net.getStatus();
+			DriftNetStatus prevTickNetStatus = net.getPrevTickStatus();
```
dekvall, line 227: Why do we care about the net status of the previous tick?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,6 +221,30 @@ public void onGameTick(GameTick tick)
...
+			if (isAcceptingFish(netStatus) || isAcceptingFish(prevTickNetStatus))
+			{
+				Iterator<NPC> itr = taggedFish.keySet().iterator();
+				// untag fish which are next to a net that is not accepting fish
+				while (itr.hasNext())
+				{
+					NPC singleTaggedFish = itr.next();
+					WorldPoint fishPoint = WorldPoint.fromLocalInstance(client, singleTaggedFish.getLocalLocation());
+					for (WorldPoint point : net.getAdjacentTiles())
```
dekvall, line 236: I think it would be better if `adjacentTiles` was just a set, which would let you do `adjacentTiles.contains(fishpoint)`

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,6 +221,30 @@ public void onGameTick(GameTick tick)
 			return;
 		}
 
+		for (DriftNet net : NETS)
```
dekvall, line 224: It feels like this entire thing can be reduced to a method that takes an npc which would let you utilize the `removeIf` below, and not have to create additional iterators.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -256,12 +302,18 @@ public void onNpcDespawned(NpcDespawned event)
...
+		taggedFish.remove(npc);
 	}
 
 	@Subscribe
 	public void onGameObjectSpawned(GameObjectSpawned event)
 	{
 		GameObject object = event.getGameObject();
+		if (object.getId() == ObjectID.ANNETTE)
+		{
+			annetteObject = object;
```
dekvall, line 314: remove this on game object despawn too

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
```diff
@@ -37,10 +38,12 @@
 	private final int objectId;
 	private final Varbits statusVarbit;
 	private final Varbits countVarbit;
+	private final WorldPoint[] adjacentTiles;
```
dekvall, line 41: As mentioned below, i would make this a Set

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
bartpeeters: It's not wrong, it's just that the fish shoal stay tagged even when you collect the loot. The fish shoal only become untagged when they are adjacent to a drift net that is not actively catching (= full, unset) or when they have been tagged for a certain amount of ticks, which I was unable to determine but would seem to be somewhere randomly between 40 and 100.

dekvall: Alright, how did you time the ticks?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
bartpeeters: When a fish shoal is 2 tiles adjacent to a drift net and is moving to a net that is just being setup it will be denied even though the net is currently in the CATCHING status. The easiest solution seemed to be to just look to the previous tick.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
bartpeeters: I added an overlay to the fish shoal which showed the tick it was tagged (value of taggedFish).

bartpeeters: I agree with all the other comments, do I add these changes in a new commit(s) or do I amend the current commits?

dekvall: I get that, but i was more interested in how you determined when a shoal was no longer tagged? Was it when it moved away from the net, or just didn't move towards it for example?

dekvall: I'd prefer it if you amend your current commits :)

bartpeeters: Visually when a shoal was 2 or less tiles away from a drift net but did not go in it. Which means it was probably untagged for a longer time, but just never moved close enough to a net. That's why it's hard to determine how many ticks exactly it takes for a shoal to become untagged. But in this implementation it seldomly happens that a tagged shoal is actually not tagged.

Maybe if you have some time to test this implementation you can determine what causes a shoal which seems to be tagged to not actually be. It does not happen very often so I was unable to determine this case. It could be just a tick timeout or something else.

dekvall: No i trust you, i was mostly interested in how you did it because i had the same problem, it's very hard to determine. Sounds good.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
```diff
@@ -37,10 +39,32 @@
...
+	{
+		return isNotAcceptingFish(status);
+	}
+
+	boolean wasNotAcceptingFishPrevTick()
+	{
+		return isNotAcceptingFish(prevTickStatus);
+	}
+
+	void setPrevTickStatus()
```
dekvall, line 64: Use lombok `@Setter` instead, also a setter without an argument feels weird?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
```diff
@@ -37,10 +39,32 @@
...
 
 	private GameObject net;
 	private DriftNetStatus status;
 	private int count;
+	private DriftNetStatus prevTickStatus;
+
+	private boolean isNotAcceptingFish(DriftNetStatus status)
+	{
+		return status != DriftNetStatus.CATCH && status != DriftNetStatus.SET;
+	}
```
dekvall, line 52: I don't think this method is necessary

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
```diff
@@ -76,12 +76,12 @@ default boolean highlightUntaggedFish()
 	)
 	@Range(
 		min = 1,
-		max = 60
+		max = 100
```
dekvall, line 79: Maybe we can increase this some if we are certain it's around 40-60 secs, 120 maybe?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
```diff
@@ -94,4 +94,26 @@ default Color untaggedFishColor()
 	{
 		return Color.CYAN;
 	}
+
+	@ConfigItem(
+		keyName = "tagAnnetteWhenNoNets",
```
dekvall, line 99: can just be `tagAnnette` imo

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -94,13 +99,30 @@
 	private Map<NPC, Integer> taggedFish = new HashMap<>();
 	@Getter
 	private final List<DriftNet> NETS = ImmutableList.of(
-		new DriftNet(NullObjectID.NULL_31433, Varbits.NORTH_NET_STATUS, Varbits.NORTH_NET_CATCH_COUNT),
-		new DriftNet(NullObjectID.NULL_31434, Varbits.SOUTH_NET_STATUS, Varbits.SOUTH_NET_CATCH_COUNT));
+		new DriftNet(NullObjectID.NULL_31433, Varbits.NORTH_NET_STATUS, Varbits.NORTH_NET_CATCH_COUNT, new HashSet<WorldPoint>(ImmutableList.of(
```
dekvall, line 102: Could use `ImmutableSet`

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -284,6 +330,20 @@ public void onGameObjectDespawned(GameObjectDespawned event)
 		}
 	}
 
+	@Subscribe
+	public void onItemContainerChanged(final ItemContainerChanged event)
+	{
+
```
dekvall, line 336: remove this newline

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,6 +226,19 @@ public void onGameTick(GameTick tick)
...
 
+		for (DriftNet net : NETS)
+		{
+			// untag fish which are next to a net that is not accepting fish right now
+			// or was not accepting fish previous tick (When a fish shoal is 2 tiles adjacent
+			// to a drift net and is moving to a net that is just being setup it will be denied
+			// even though the net is currently in the CATCHING status)
+			if (net.isNotAcceptingFish() || net.wasNotAcceptingFishPrevTick())
+			{
+				taggedFish.entrySet().removeIf(entry -> checkFishOnTiles(entry.getKey(), net.getAdjacentTiles()));
```
dekvall, line 237: i was thinking that this could go in a method too so you could use the `removeIf` below in a `removeIf(expired || closeToUnsetNet)` (naming idk)

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
```diff
@@ -37,10 +39,32 @@
...
+
+	boolean isNotAcceptingFish()
+	{
+		return isNotAcceptingFish(status);
+	}
+
+	boolean wasNotAcceptingFishPrevTick()
```
dekvall, line 59: could probably be just `wasNotAcceptingFish` and a comment

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,6 +226,19 @@ public void onGameTick(GameTick tick)
...
+			}
+			net.setPrevTickStatus();
```
dekvall, line 239: This will be the same as the current status unless there has been a change this tick, so it might be worth considering setting this in onVarbitChanged instead? not sure.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetConfig.java
bartpeeters: After getting another fishing level yesterday I feel like 60 is ideal. Almost never had a fish which was tagged which didn't actually was and seldomly had a fish which was untagged still go in a net. Which is better in my opinion than a fish who is tagged but actually isn't. At 100 this happens pretty often so I wouldn't go higher than this.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
bartpeeters: Don't really know how that would work. We would still need to make sure that the next tick after a VarbitChanged, the prevTickStatus would be the same as the current status.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -196,6 +214,22 @@ public void onInteractingChanged(InteractingChanged event)
...
+	private boolean checkFishOnTiles(NPC fish, Set<WorldPoint> adjacentTiles)
+	{
+		final WorldPoint fishTile = WorldPoint.fromLocalInstance(client, fish.getLocalLocation());
+		return adjacentTiles.contains(fishTile);
+	}
+
+	private boolean checkFishNextToNet(NPC fish, Iterable<DriftNet> nets)
+	{
+		return Iterables.size(Iterables.filter(nets, net -> checkFishOnTiles(fish, net.getAdjacentTiles()))) > 0;
+	}
```
Nightfirecat, line 226: Instead of accepting an `Iterable`, a Collection or even a List would be preferable. That way, the Java stream API can be used instead of guava's `Iterables`:

```suggestion
	private boolean checkFishNextToNet(NPC fish, Collection<DriftNet> nets)
	{
		return nets.stream().anyMatch(net ->
		{
			final WorldPoint fishTile = WorldPoint.fromLocalInstance(client, fish.getLocalLocation());
			return net.getAdjacentTiles().contains(fishTile);
		});
	}
```

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -196,6 +214,22 @@ public void onInteractingChanged(InteractingChanged event)
 		}
 	}
 
+	private boolean checkFishOnTiles(NPC fish, Set<WorldPoint> adjacentTiles)
```
Nightfirecat, line 217: This can be inlined into `checkFishNextToNet` since it is so short. See my comment on that method

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,8 +238,15 @@ public void onGameTick(GameTick tick)
 			return;
 		}
 
-		final int currentTickCount = client.getTickCount();
-		taggedFish.entrySet().removeIf(entry -> entry.getValue() + config.timeoutDelay() < currentTickCount);
+		// nets that are not accepting fish are those currently not accepting or those who were
+		// not accepting in the previous tick (When a fish shoal is 2 tiles adjacent
+		// to a drift net and is moving to a net that is just being setup it will be denied
+		// even though the net is currently in the CATCHING status)
+		Iterable<DriftNet> notAcceptingNets = Iterables.filter(NETS, net -> net.isNotAcceptingFish() || net.wasNotAcceptingFish());
```
Nightfirecat, line 245: Per my comment on `checkFishNextToNet()`, this can become a `List` instead:
```suggestion
		List<DriftNet> notAcceptingNets = NETS.stream()
			.filter(net -> net.isNotAcceptingFish() || net.wasNotAcceptingFish())
			.collect(Collectors.toList());
```

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -94,13 +101,30 @@
...
 
 	@Getter
 	private boolean inDriftNetArea;
 	private boolean armInteraction;
 
+	@Getter
+	private boolean driftNetsInInventory = false;
+
+	@Getter
+	private GameObject annette;
```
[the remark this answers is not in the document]
bartpeeters: Are the npc's also garbage collected on scene changes? If so I could change this:
```	
@Subscribe
public void onGameStateChanged(GameStateChanged event)
{
  switch (event.getGameState())
  {
    case LOGIN_SCREEN:
    case HOPPING:
    case LOADING:
      reset();
      break;
    case LOGGED_IN:
      inDriftNetArea = checkArea();
      updateDriftNetVarbits();
      break;
  }
}
```
to
```
@Subscribe
public void onGameStateChanged(GameStateChanged event)
{
  if (event.getGameState() == LOGGED_IN) {
    inDriftNetArea = checkArea();
    updateDriftNetVarbits();
  }else {
    reset();
  }
}
```
And add `annette = null` to the reset function.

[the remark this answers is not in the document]
Nightfirecat: NPCs are despawned via `NpcDespawned` events on game state changes.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,8 +241,17 @@ public void onGameTick(GameTick tick)
...
-		taggedFish.entrySet().removeIf(entry -> entry.getValue() + config.timeoutDelay() < currentTickCount);
+		// nets that are not accepting fish are those currently not accepting or those who were
+		// not accepting in the previous tick (When a fish shoal is 2 tiles adjacent
+		// to a drift net and is moving to a net that is just being setup it will be denied
+		// even though the net is currently in the CATCHING status)
+		List<DriftNet> notAcceptingNets = NETS.stream()
+			.filter(net -> net.isNotAcceptingFish() || net.wasNotAcceptingFish())
+			.collect(Collectors.toList());
+
+		taggedFish.entrySet().removeIf(entry -> checkTagExpired(entry.getValue()) || checkFishNextToNet(entry.getKey(), notAcceptingNets));
```
dekvall, line 252: This line is a little long, can we break it off somewhere? Also i think we should `s/check/is/` in for both of these methods.

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNet.java
```diff
@@ -37,10 +40,23 @@
...
+
+	boolean isNotAcceptingFish()
+	{
+		return status != DriftNetStatus.CATCH && status != DriftNetStatus.SET;
+	}
+
+	boolean wasNotAcceptingFish()
+	{
+		return prevTickStatus != DriftNetStatus.CATCH && prevTickStatus != DriftNetStatus.SET;
+	}
```
dekvall, line 59: Since we don't actually care if we were or are accepting fish (`.filter(net -> net.isNotAcceptingFish() || net.wasNotAcceptingFish())`) maybe we should just combine this into one `notAcceptingFish` and put the comment here?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -196,6 +219,20 @@ public void onInteractingChanged(InteractingChanged event)
 		}
 	}
 
+	private boolean checkFishNextToNet(NPC fish, Collection<DriftNet> nets)
+	{
+		return nets.stream().anyMatch(net ->
+		{
+			final WorldPoint fishTile = WorldPoint.fromLocalInstance(client, fish.getLocalLocation());
```
dekvall, line 226: I think this can go outside?

runelite-client/src/main/java/net/runelite/client/plugins/driftnet/DriftNetPlugin.java
```diff
@@ -204,8 +241,17 @@ public void onGameTick(GameTick tick)
 			return;
 		}
 
-		final int currentTickCount = client.getTickCount();
-		taggedFish.entrySet().removeIf(entry -> entry.getValue() + config.timeoutDelay() < currentTickCount);
+		// nets that are not accepting fish are those currently not accepting or those who were
+		// not accepting in the previous tick (When a fish shoal is 2 tiles adjacent
+		// to a drift net and is moving to a net that is just being setup it will be denied
+		// even though the net is currently in the CATCHING status)
+		List<DriftNet> notAcceptingNets = NETS.stream()
```
dekvall, line 248: the `notAcceptingNets` wording is a little funky since we have `*notAcceptingFish` which in this case essentially is the same thing in this case, also it sounds like a boolean. Maybe something like `closedNets`(?) is better

bartpeeters:
Just tested that everything still works after the code changes. This is the case, so these changes could be merged. Thanks for the reviews @dekvall @Nightfirecat

PR #11596, opened 2020-05-14
DPS Counter: Boss only mode

MarbleTurtle:
Adds an option to only track damage dealt to a boss, mainly aimed at cases such as GWD when damaging the minions the counter would unpause but not pause when they're dead. 

Boss list has also been tweaked to avoid confusion on the plugin not working while attacking enemies that would be clearly bosses.


Adam-:
Can you split this into one commit for the new bosses and another for the boss only option?

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -162,7 +162,12 @@ public void onHitsplatApplied(HitsplatApplied hitsplatApplied)
 		}
 
 		Hitsplat hitsplat = hitsplatApplied.getHitsplat();
-
+		final int npcId = ((NPC) actor).getId();
+		boolean isBoss = BOSSES.contains(npcId);
+		if (dpsConfig.bossOnly() && !isBoss)
+		{
+			return;
+		}
```
Nightfirecat, line 193: Add some spacing between variable declarations and statement blocks. Also flip the conditions to avoid some config checks.
```suggestion
		final int npcId = ((NPC) actor).getId();
		boolean isBoss = BOSSES.contains(npcId);

		if (!isBoss && dpsConfig.bossOnly())
		{
			return;
		}

```

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsConfig.java
```diff
@@ -52,4 +52,14 @@ default boolean autopause()
 	{
 		return false;
 	}
+	@ConfigItem(
```
Nightfirecat, line 55: Config items are always separated by a blank line.
```suggestion

	@ConfigItem(
```

Adam-:
This requires rebase.

PR #11598, opened 2020-05-14
Add a visual cue to the metronome

JtSangerman:


dekvall:
This doesn't respect the configured `tickCount` which it probably should. Also, there is already a prayer flick indicator in the prayer bar, and on the prayer orb. For skilling there are fake "tick" xp drops that can be used as visual cues. I guess I don't really see an use case for this but I'm thinking a better cue would be a pie chart like we do in other places, while it looks like this would just flash every tick like a broken traffic light. If you prefer it this way it might be better as an external plugin?

JtSangerman:
@dekvall That is a good point about it not conforming to `tickCount`. This has been updated. An example of a use case would be for inferno, or any other PvM content, really. It has helped me to be able to keep your eyes on you character, rather than glued to your prayer orb/bar or elsewhere inconvenient to the content.

Adam-:
I wouldn't mind a visual cue, but drawing over the player seems sort of obnoxious? Is there anything better?

PR #11600, opened 2020-05-15
Add Time Remaining to XP Globe / Info Box

tomlegodais:
Add "Time Remaining" to the XP Globe Overlay and XP Info Overlay:

xp_orb

xp_tracker

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesConfig.java
@@ -171,4 +171,15 @@ public interface XpGlobesConfig extends Config
 	{
 		return 10;
 	}
+
+	@ConfigItem(
+		keyName = "xpGlobeTimeRemaining",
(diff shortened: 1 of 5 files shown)


while-loop:
I believe you only need the changes to the xpglobe plugin. 
Xptracker already has a config to show TTL for the on-screen canvas.

received_1327654210754048

tomlegodais:
I realized this after I made the change, my bad. Although I would like to rationalize it with this reasoning: being able to have XP Gained / XP Left / Percentage / Remaining Time all on the same XP tracker.

i.e. this is how I currently play with my in-game XP tracker overlay, would be nice to have the time show too.

RuneLite_miUUAoVXuL

Nightfirecat:
Since the xp orbs tooltip is already fairly large, I don't mind the idea of adding another line to it. The xp tracker overlay, however, is meant to be compact. We don't have the guarantee of plentiful screen real estate, especially for players using the fixed viewport, so I don't know that we'd want to add more lines to the overlay. Maybe it'd be fine if additional lines could be disabled? Just an idea.

Issue #11393, opened 2020-04-28
Confusing grammar on "Import local settings to remote RuneLite account"


Atroxide:
**Describe the bug**
On the info side panel, there is a button that says

> Import local settings to remote RuneLite account

I don't know what that button does? 
Does it **export** local settings **to** remote RuneLite account
Or does it **import** local settings **from** remote RuneLite account

The words **import** and **to** contradict one another.

I have no idea which direction the sync is when you click this button, and I guessed wrong and lost all of my plugin configuration.

**To Reproduce**
1. Go to 'info' side-panel
2. Read the first button

**Expected behavior**
Either 
**Export** local settings **to** remote RuneLite account
Or
**Import** local settings **from** remote RuneLite account

**Environment (please complete the following information):**
 - OS: Windows 10
 - RuneLite version: 1.6.12
 - Launcher version: 1.6.0

dekvall:
I don't think that they contradict each other. In order to get the `Import..` option you have to be logged in to your RL account, and hence the local settings file will be the external source that we are importing from. Namely, we are importing the current *local settings* and where are these settings imported to? Well, the RL account that you are logged in to, so i think it's more of a clarification than a contradiction.

A backup of your remote settings is also stored in `~/.runelite/profiles` when you import your local settings so your original remote RL settings can be restored by moving that file and importing again.

Atroxide:
So if I understand correctly- my client isn't actually using my local settings, its using 100% strictly the remote settings? Where exactly do the local settings come from? from before I logged into my account?

From a UI perspective, its extremely confusing and isn't user-friendly. I assumed clicking the button would sync my local settings (how my client is right now) to the cloud (remote account) so if I log in else where, my settings would be the same as they are now. Instead it ended up reverting my settings to when I first downloaded runelite before I ever set any plugins to be how I want them to be.

Maybe it's just me though. What is the actual intended use of this button if it isn't to sync your current setup to the 'cloud'?

dekvall:
If you are logged in to an RL account it's strictly using the settings stored in that account yes. The local settings file is used to save your current configuration when you are not logged in.

If you are logged in on the same account on multiple devices, the settings are automatically synced between them. Having to re-import your settings every time you made a change in a config would kind of defeat the purpose of being logged in in the first place.

The purpose of the button is if you have been playing an extensive amount of time without being logged in, the process of transferring your setting will be smooth if you ever choose to make an account.

claudiodekker:
[...]

PR #11607, opened 2020-05-15
#11393 - Confusing grammar on "Import local settings to remote RuneLite account"

claudiodekker:
(Potentially) Solves #11393

Please note I did not test this change, as I currently don't have access to an IDE.


dekvall:
I disagree with this proposed change and think that if anything, it makes it more confusing. You are still logged in, and i think it's wrong to use `Export` just because the server is spatially further away from you than your local settings file is.

`Synchronizing` in this context (you have just logged in and lost all your settings) would infer losing all your settings. Which is strictly not true, since you replace your logged in settings with settings from a local settings file.

I also think the amended `[...] as well as to any other RuneLite clients you might be logged in to` is unnecessary here, since preserving settings over devices is an integral part of the account system, and what else would replacing your settings do?

I understand that the issue you are referring to is tagged as `discussion` and i personally believe that the text is best kept as is, it's what most correctly describes what's happening.

PR #11619, opened 2020-05-16
Make background color of panel components configurable

### LINKED ISSUE #11435: Infobox Opacity
Screenshot (138)

It would be nice to have the option to increase the opacity of infoboxes to be solid so that they are easier to see when placing them at various different places on the client.

TheStonedTurtle:
Closes #11435 

w28Qor6T6e

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
@@ -283,13 +283,13 @@ public interface RuneLiteConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "infoBoxColor",
-		name = "Infobox Color",
-		description = "Configures the background color of infoboxes",
+		keyName = "overlayBackgroundColor",
+		name = "Overlay Color",
+		description = "Configures the background color of infoboxes and overlays",
 		position = 43
 	)
 	@Alpha
-	default Color infoBoxColor()
+	default Color overlayBackgroundColor()
 	{
 		return ComponentConstants.STANDARD_BACKGROUND_COLOR;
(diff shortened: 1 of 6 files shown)


TheStonedTurtle:
I have removed the Units (`/255`) from this option per discord feedback.

6sQDsmMPgb

deathbeam:
Imo it would be better if it would be simply color picker, and apply to all panel overlays. Adding configuration option specifically for infoboxes just because someone wants to see them better over their inventory is pretty dumb reason imo.

TheStonedTurtle:
I've updated this PR to use a Color picker for the infobox background colors

image

I'm not sure how to apply the RuneLite config color to all PanelComponent background colors (that aren't being manually set by a plugin like MLM sack) so i added this just for info boxes.

makingcomment:
I don't understand why the opacity configuration had to be _replaced_ with the background color configuration in this PR. Why not add both?

TheStonedTurtle:
The color picked includes an Alpha (opacity) option.

deathbeam:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayPanel.java#L65

TheStonedTurtle:
> 
> 
> https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayPanel.java#L65

How would I access the RuneLite config inside this method?

TheStonedTurtle:
PR has been updated so OverlayPanels get the color from the RuneliteConfig.

image

I've left this as a separate commit so its easier to review

deathbeam:
Pushed small change, can you retest @TheStonedTurtle

PR #11624, opened 2020-05-16
ClientUI: Remember sidebar state

### LINKED ISSUE #2312: Remember sidebar open or closed after exiting the client
The sidebar always shows (in it's minimised form=icons only) when starting the client even after closing (hiding) it the previous session.

Please allow for the client to remember whether the sidebar was set to open or closed (hidden).

Xortrox:
This has been a minor annoyance when playing multiple accounts.

I had a pull request not based on a separate branch earlier, I closed that one so I could make this.

Closes #2312

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -107,6 +107,7 @@ public class ClientUI
 	private static final String CONFIG_GROUP = "runelite";
 	private static final String CONFIG_CLIENT_BOUNDS = "clientBounds";
 	private static final String CONFIG_CLIENT_MAXIMIZED = "clientMaximized";
+	private static final String CONFIG_CLIENT_SIDEBAR_CLOSED = "clientSidebarClosed";
 	private static final int CLIENT_WELL_HIDDEN_MARGIN = 160;
 	private static final int CLIENT_WELL_HIDDEN_MARGIN_TOP = 10;
 	public static final BufferedImage ICON = ImageUtil.getResourceStreamFromClass(ClientUI.class, "/runelite.png");
@@ -477,7 +478,8 @@ public class ClientUI
 			sidebarNavigationButton = NavigationButton
 				.builder()
 				.priority(100)
-				.icon(sidebarClosedIcon)
+				.icon(sidebarOpenIcon)
+				.tooltip("Open SideBar")
 				.onClick(this::toggleSidebar)
 				.build();
 
@@ -487,7 +489,12 @@ public class ClientUI
 				null);
 
 			titleToolbar.addComponent(sidebarNavigationButton, sidebarNavigationJButton);
-			toggleSidebar();
+
+			// Open sidebar if the config closed state is unset
+			if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED) == null)
+			{
+				toggleSidebar();
+			}
 		});
 	}
 
@@ -875,6 +882,7 @@ public class ClientUI
 		{
 			sidebarNavigationJButton.setIcon(new ImageIcon(sidebarOpenIcon));
 			sidebarNavigationJButton.setToolTipText("Open SideBar");
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED, true);
 
 			contract();
 
@@ -885,6 +893,7 @@ public class ClientUI
 		{
 			sidebarNavigationJButton.setIcon(new ImageIcon(sidebarClosedIcon));
 			sidebarNavigationJButton.setToolTipText("Close SideBar");
+			configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED);
 
 			// Try to restore last panel
 			expand(currentNavButton);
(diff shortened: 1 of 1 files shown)


Xortrox:
Fixed the style issues, my main language isn't really Java so I'm used to code formatting from JS/Node.JS.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -552,6 +553,15 @@ public void show()
 			requestFocus();
 			giveClientFocus();
 			log.info("Showing frame {}", frame);
+
+			// Close sidebar if config is set for it.
+			if (config.rememberSidebarState())
+			{
+				if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED) != null)
+				{
+					toggleSidebar();
```
Adam-, line 562: Instead you can just not open the sidebar in the first place by removing the `toggleSidebar` call in `init`.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -552,6 +553,15 @@ public void show()
 			requestFocus();
 			giveClientFocus();
 			log.info("Showing frame {}", frame);
+
+			// Close sidebar if config is set for it.
+			if (config.rememberSidebarState())
```
Adam-, line 558: It probably isn't necessary to have a config option for this

Xortrox: Sorry for the double push, I only saw the first comment. But this should also be solved now.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -842,6 +851,7 @@ private void toggleSidebar()
 		{
 			sidebarNavigationJButton.setIcon(new ImageIcon(sidebarOpenIcon));
 			sidebarNavigationJButton.setToolTipText("Open SideBar");
+			configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED);
```
Adam-, line 854: this is backwards

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -485,7 +486,15 @@ public void layoutContainer(Container parent)
 				null);
 
 			titleToolbar.addComponent(sidebarNavigationButton, sidebarNavigationJButton);
-			toggleSidebar();
+
+			// Close sidebar if config is set for it.
+
+			if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED) == null)
```
Adam-, line 492: this is just wrong

Xortrox:
> It would be good if you test your code before submitting it.

I had not set up my project to build yet, I will do that and actually test from now on, looking at this shortly.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
Xortrox: In hindsight seems maybe you meant to remove the config related code entirely, even the remembered state? Anyway seems the travis build failed due to some env variable issue now, which seems to be out of my control.

Xortrox: Realized the build failed due to another checkstyle issue, I got confused because travis said "no environment variables set"

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -852,6 +859,7 @@ private void toggleSidebar()
 		{
 			sidebarNavigationJButton.setIcon(new ImageIcon(sidebarClosedIcon));
 			sidebarNavigationJButton.setToolTipText("Close SideBar");
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_CLIENT_SIDEBAR_CLOSED, true);
```
Adam-, line 862: This is still backwards. When the tooltip text is `Close Sidebar` it means the sidebar is open, because the option on it is to *close* it. So sidebar closed shouldn't be true.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
Xortrox: After doing this I realized that it causes a few other problems which I've also resolved, such as saveClientBoundsConfig() defaulting to 809 width to account for an always open sidebar.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -484,8 +485,15 @@ public void layoutContainer(Container parent)
 				0,
 				null);
 
+			sidebarNavigationJButton.setToolTipText("Open SideBar");
```
Adam-, line 488: This should just be `.tooltip("Open SideBar")` in the above navbutton builder.

Xortrox:
I assume I was meant to resolve the merge conflict on my own.

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -1097,11 +1106,7 @@ private void saveClientBoundsConfig()
 		}
 		else
 		{
-			// Try to expand sidebar
-			if (!sidebarOpen)
-			{
-				bounds.width += pluginToolbar.getWidth();
-			}
+			final Rectangle bounds = frame.getBounds();
```
Adam-, line 1109: this was merged incorrectly and now no longer compiles. I assume the removal of the above code was due to that it now remembers sidebar state so it doesn't have to offset the bounds by that anymore right?

Xortrox: Correct

PR #11626, opened 2020-05-16
Added comments that map out weapon types and added support for Accurate Ranging, Blocking, and Aim and Fire attack styles

ThePharros:


runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/WeaponType.java
```diff
@@ -27,45 +27,47 @@
...
-	TYPE_19(RANGING, RANGING, null, LONGRANGE),
-	TYPE_20(ACCURATE, CONTROLLED, null, DEFENSIVE),
-	TYPE_21(ACCURATE, AGGRESSIVE, null, DEFENSIVE, CASTING, DEFENSIVE_CASTING),
-	TYPE_22(ACCURATE, AGGRESSIVE, AGGRESSIVE, DEFENSIVE),
-	TYPE_23(CASTING, CASTING, null, DEFENSIVE_CASTING),
-	TYPE_24(ACCURATE, AGGRESSIVE, CONTROLLED, DEFENSIVE),
-	TYPE_25(CONTROLLED, AGGRESSIVE, null, DEFENSIVE),
-	TYPE_26(AGGRESSIVE, AGGRESSIVE, null, AGGRESSIVE),
-	TYPE_27(ACCURATE, null, null, OTHER);
+	TYPE_0(ACCURATE, AGGRESSIVE, null, DEFENSIVE), //Unarmed
```
Nightfirecat, line 43: It's not important to maintain the `TYPE_0` naming scheme. Instead, rename the enums themselves.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/WeaponType.java
```diff
@@ -27,45 +27,47 @@
...
+	TYPE_1(ACCURATE, AGGRESSIVE, AGGRESSIVE, DEFENSIVE), //Axe
+	TYPE_2(ACCURATE, AGGRESSIVE, null, DEFENSIVE), //Blunt
+	TYPE_3(ACCURATE_RANGING, RANGING, null, LONGRANGE), //Bow
+	TYPE_4(ACCURATE, AGGRESSIVE, CONTROLLED, DEFENSIVE), //Claws
+	TYPE_5(ACCURATE_RANGING, RANGING, null, LONGRANGE), //Crossbow
+	TYPE_6(AGGRESSIVE, RANGING, CASTING, null), //Salamander
+	TYPE_7(ACCURATE_RANGING, RANGING, null, LONGRANGE), //Chinchompa
+	TYPE_8(AIM_AND_FIRE, AGGRESSIVE, null, null), //Gun
```
Nightfirecat, line 51: In this case, a comment explaining that this is found on the Fixed device (since "gun" isn't a common weapon type) would be helpful.

ThePharros:
> The enum renaming and attack style changes should be done in separate commits:
> 
> 1. Give names to weapon type enum entries
> 2. Add missing attack styles

What's the best way to revert my current branch to master so I can

ThePharros:
closed via misclick >.>

ThePharros:
> The enum renaming and attack style changes should be done in separate commits:
> 
> 1. Give names to weapon type enum entries
> 2. Add missing attack styles

Alright. I've pushed the changes as separate commits.

makingcomment:
The "added support for Accurate Ranging, Blocking, and Aim and Fire attack styles" seems to have been lost in the latest changes. Is that intentional?

ThePharros:
> The "added support for Accurate Ranging, Blocking, and Aim and Fire attack styles" seems to have been lost in the latest changes. Is that intentional?

They are separate commits now as requested. I figured they would be looked at individually and then merged upon push.

ThePharros:
> None of these enum changes look correct

In terms of what? The enums were renamed to their corresponding weapon categories. Can be tested in-game onVarbitChanged(). Then the custom attack style enums are based on the corresponding in-game attack style.

Spedwards:
> They are separate commits now as requested. I figured they would be looked at individually and then merged upon push.

No they aren't. They're all in one commit: 070130774fb77e34e6cff71658d43476fa0aa6c9

ThePharros:
> > They are separate commits now as requested. I figured they would be looked at individually and then merged upon push.
> 
> No they aren't. They're all in one commit: [0701307](https://github.com/runelite/runelite/commit/070130774fb77e34e6cff71658d43476fa0aa6c9)

That commit only has the enum changes with the attack styles reverted back to master?

Adam-:
The plugin uses the enum ordinal values to lookup the styles though, so I don't see how this could possibly work at all? Does it work ingame?

Spedwards:
> > > They are separate commits now as requested. I figured they would be looked at individually and then merged upon push.
> > 
> > 
> > No they aren't. They're all in one commit: [0701307](https://github.com/runelite/runelite/commit/070130774fb77e34e6cff71658d43476fa0aa6c9)
> 
> That commit only has the enum changes with the attack styles reverted back to master?

No it isn't... You've renamed and added missing styles all in one commit which is what you were asked to split. Your split was also done incorrectly.

ThePharros:
> The plugin uses the enum ordinal values to lookup the styles though, so I don't see how this could possibly work at all? Does it work ingame?

Oof, you're right. I must've missed the ordinal logic so figured I could alphabetize the enum, I can revert the order.

ThePharros:
> > > > They are separate commits now as requested. I figured they would be looked at individually and then merged upon push.
> > > 
> > > 
> > > No they aren't. They're all in one commit: [0701307](https://github.com/runelite/runelite/commit/070130774fb77e34e6cff71658d43476fa0aa6c9)
> > 
> > 
> > That commit only has the enum changes with the attack styles reverted back to master?
> 
> No it isn't... You've renamed and added missing styles all in one commit which is what you were asked to split. Your split was also done incorrectly.

I'm confused. The first commit has both changes. The next commit removes the enum comments so when 1 and 2 are merged, it is only adding the new attack styles. Then the commit after that is using the master's vanilla attack styles with the enum renaming only (which I just learned needs its order to be fixed)

PR #11630, opened 2020-05-16
Swap boss icons

psikoi:
I made two boss icon sets for wiseoldman (small and medium sizes) and thought I'd offer **the small versions** to replace the current hiscore boss icons.

These new icons have been tweaked to have more consistent sizes, and some have been actually painted to be readable.

They also have a thinner black outline, which I personally prefer.

Some examples of icons tweaked for visual clarity:

image

If these are too small, I could try using the medium versions (around 24x24)

image


Nightfirecat:
Of these icon swaps, I prefer the following of ours to the proposed swaps:

* Abyssal Sire (why use the abyssal head model?)
* Bryophyta (nature rune isn't the clearest indicator)
* Chaos Fanatic (I much prefer the look of the full ancient staff to just the ancient symbol)
* Corporeal Beast (prefer corp beast model to ely sigil)
* Zulrah (prefer zulrah pet model to magic fang)
* I'm on the fence about the DK icons, not sure one way or another with these ones
* Also on the fence about the Wintertodt icon. Is the phoenix pet or the tome of fire the more recognizable drop from the boss?

Aside from these, the new icons look vastly better, especially CoX, Scorpia, the Gauntlet, and Zalcano. Appreciate the contribution!

makingcomment:
My thoughts on the icons mentioned by Nightfirecat:
- I completely agree that the current Abyssal Sire, Bryophyta, Chaos Fanatic, Corporeal Beast, and Zulrah icons were better.
- I very slightly prefer the current DK icons, but it's pretty close so I don't really care either way.
- I think the pheonix pet is a much better icon than the tome for Wintertodt.

Some additional points:
- I don't like the current Inferno icon _or_ the proposed new one, but I like the new one less than I like the current one.
- I prefer the current thick borders to the proposed thin ones. Nonetheless, if the thin ones are used, any current icons that are kept if/when swaps are made should have their borders updated for consistency.

psikoi:
I feel this is going to be very hard to have a consensus for which ones to keep/update.

And yes, if I were to keep any old ones, I'd update them to be consistent with the others

Nightfirecat:
The new images likely need `optipng` or something to reduce their size, but that's optimization which can happen once we're all in agreement on which icons are used. Also, I disagree regarding the border thickness, as a thinner ~1px border better matches the other hiscore skill/minigame icons.

psikoi:
Is there interest in this? I can tweak if need be.

Broooklyn:
There is definitely still interest. Personally I'm in agreement with NFC's comment above, as well as makingcomment's note about Zuk—not sure the best approach to that one.

Broooklyn:
I've been using these icons for a little while, so here are my thoughts:

- Bryophyta, Corp, and Chaos Fanatic aren't the clearest in what they are. Corp is I guess, but I think the pet icon is more fitting, although the ely icon is really clean.
- I think I'm of the camp where phoenix > tome for wintertodt
- Zuk icon has grown on me a lot, I quite like it now that I've gotten used to it.
- KBD and Scorpia are kinda hard to see against the dark background, not sure the best approach to that. May just be me I dunno.
- Something about the magic fang icon for zulrah kinda throws me off, not sure if it's the resolution or what, but it doesn't look as crisp as many of the others. Snakeling would probably be better imo.

These are obviously just the opinions of one man, so take them as you wish. Overall I quite like the icons, though, and would love to see these implemented.

dekvall:
Looking through the responses to this it looks like we should keep the following icons
 - Abyssal Sire
 - Bryophyta
 - Chaos Fanatic
 - Corporeal Beast
 - Zulrah
 - Wintertodt
and replace the remaining ones. 
As you said, it's hard to reach a consensus but this seems like the consensus of those who have weighed in, and the new icons really do look fantastic.

psikoi:
Sounds good. I can keep the current icons for those bosses, but i'll need to edit/clean them to match the others.

Nightfirecat:
I've rebased this and touched up the short list of icons we wanted to keep. Behold, the new new boss panel listing!

new new boss hiscore panel

PR #11638, opened 2020-05-18
itemcharges: add number of super strength doses

WLoumakis:


makingcomment:
#11634 is better

Nightfirecat:
Added in #11634

PR #11639, opened 2020-05-18
Swap the "Talk-to" option for "Buy-plank" when interacting with sawmill NPCs

h0dgep0dge:
I was surprised this wasn't already in there, so I just whipped this together. I thought perhaps it was excluded for balance reasons, in which case we can disregard this pull request.


makingcomment:
This is already in the Menu Entry Swapper Extended plugin on the Plugin Hub. That's doesn't necessarily mean that it shouldn't be in the base Menu Entry Swapper plugin instead, but it's worth noting.

h0dgep0dge:
I don't know anything about the plugin hub, other than a vague awareness of it's existence. If there's parallel development happening between the plugin hub and mainline Runelite perhaps there needs to be some better integration between them. In any case, somewhat beyond the scope of this thread lol

Nightfirecat:
We're not looking to incorporate into the client most of the more niche/less maintainable features (this being the former) which are already available in the plugin hub.

PR #11645, opened 2020-05-18
npcunaggroarea: split color config depending on timer status

abextm:
This allows you to hide the overlay entirely while NPCs are aggressive, e.g. if you are slaying.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
@@ -25,6 +25,7 @@
 package net.runelite.client.plugins.npcunaggroarea;
 
 import java.awt.Color;
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
```diff
@@ -83,21 +84,34 @@ default boolean showAreaLines()
...
+	@ConfigItem(
+		keyName = "npcUnaggroAreaColor",
+		name = "Unaggressive colour",
+		description = "Choose colour to use for marking NPC unaggressive area after NPCs have lost aggression",
+		position = 6
+	)
+	@Alpha
+	default Color unaggroAreaColor()
+	{
+		return new Color(0xFFFF00);
```
deathbeam, line 107: Shouldn't this just be Color.YELLOW

abextm: Imo its more clear for it to match the other config

abextm:
Its more clear because 1) its just a hexcode, and 2) we have to use the same hexcode somewhere else where we have alpha included. Its alot easier to see that `0x64FFFF00` is the same RGB as `0xFFFF00` when you don't hide the alphaless one behind a reference.

PR #11650, opened 2020-05-19
Add a check for if chat is focused to consume key typed events

### LINKED ISSUE #11578: Key Remapping conflicts with menu option presses
**Describe the bug**
When the Key Remapping plugin is toggled on, and an in-game menu is opened that accepts key presses for options (Spirit Tree, or Jewelry Box) is opened, you can't pre-select your option by holding the key down before the menu pops up.

**To Reproduce**
1. Go to a spirit tree or jewelry box in-game.
1. Toggle the Key Remap Plugin on, and remap F1 to 1 (can be any remap, 1 is safe to test with as its on every menu type)
1. Hold down the number 1 on your keyboard
1. Click on the spirit tree menu
1. Your client won't select the 1 menu option immediately, without re-pressing the number 1.

**Expected behavior**
In every in-game menu that can be passed by keyboard commands (holding space to continue through prompts, numbers for choosing menu options) if you're holding the key before the menu pops up, it will select the option you're holding down on the next tick after the menu is open, near instantly, and without you having to press the key again.

**Screenshots**
I can do a screen recording, but you won't be able to see me pressing the key on my keyboard if I'm just recording the screen. I don't know how to record with voice over, but if this is needed I can figure it out.

 - OS: macOS
 - RuneLite version: 1.6.14
 - Launcher version: 2.1.1 (I did "About RuneLite" for this value, no idea if that's the right place)

**Additional context**
[These are my logs](https://gist.github.com/alfred/d22717e7c1c376706580a87411972f3b) I didn't see anything pop up while I was doing the steps to reproduce.

This isn't game-breaking, but it _is_ a regression that only started happening on the most recent release, based on the currently open issues for the last few days, I'd guess it has something to do with keyboard bank pin?

alfred:
Fixes #11578 

This PR makes it so that when keys are remapped and pressed. They still fire the unmapped key press when the chatbox is not focused.

This allows players to "pre-press" menu options by holding them down before the menu is opened.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -59,7 +59,7 @@ class KeyRemappingListener implements KeyListener
 	public void keyTyped(KeyEvent e)
 	{
 		char keyChar = e.getKeyChar();
-		if (keyChar != KeyEvent.CHAR_UNDEFINED && blockedChars.contains(keyChar))
+		if (keyChar != KeyEvent.CHAR_UNDEFINED && blockedChars.contains(keyChar) && plugin.chatboxFocused())
 		{
 			e.consume();
 		}


alfred:
I'm trying to think of a scenario where you might die because of that and I can only think of Fight Caves where the NPC announces Jad is spawning. But in that case you need to move/hit space, because (someone can correct me if I'm wrong) Jad doesn't attack until after that dialogue is gone.

So I think its fine too, just means you have to press space before using the remapped keys right?

Adam-:
Jad definitely will attack you before the dialog is gone.

alfred:
Oof, I don't really have a good solution for the spacebar thing because it's not even a remapped key? Would it be acceptable to special-case spacebar so the key typed isn't consumed while a menu is open?

Adam-:
> 
> 
> Seems to work fine. If you end up in a chatbox which accepts Space ("Click here to continue" type chat), it ends up blocking fkey remaps, but I think that's fine?

If we could detect the keypress was remapped as an fkey we could check `!plugin.isDialogOpen()` in addition I think.

alfred:
> If we could detect the keypress was remapped as an fkey we could check `!plugin.isDialogOpen()` in addition I think.

Isn't what you're describing what this block does?

https://github.com/runelite/runelite/blob/8d4d974dc55ca95d388e96d73080943386a75df9/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java#L159-L162

From this conversation I'm thinking that iff these conditions are all met:
- `config.fkeyRemap()`
- `!plugin.isTyping()`
- `plugin.isDialogOpen()`
- the dialogue **does not** have number options then to do the remapped fkey action on keypress.

then you'd want to respect the remapping and fire the remapped events

The last bullet being the blocker there, I can't see anything in this plugin that reads menu options, are there any that do that I could take a look at and copy the functionality into here?

alfred:
> Seems to work fine. If you end up in a chatbox which accepts Space ("Click here to continue" type chat), it ends up blocking fkey remaps, but I think that's fine?

Actually I'm getting ahead of myself, this behavior is the same as the official release so, it doesn't feel like this PR breaks anything?

Adam-:
I think this is probably fine as-is. I would maybe elaborate in the commit message for what you are trying to fix (keys being held while the chatbox loses focus)

alfred:
For sure, let me know if what I just updated it to is okay, or if I should add more in the expanded commit message

PR #11652, opened 2020-05-20
Add notification for burners going into random timer phase

### LINKED ISSUE #11229: Add POH Incense burner RuneLite notification
There is currently a really well made POH plugin that includes an Incense burner timer. The timer turns green when a burner is first lit, and then orange when the burner enters the random timer zone.

While this is helpful, it can be frustrating to have to look at the screen the entire time during hosting, since the most important thing is that the burners stay lit non-stop.

I would like the plugin to trigger a RuneLite notification when either of the burners enters the orange zone, so that I can direct my attention elsewhere while the burners are lit. Much like how trade notifications, stat reductions and similar events can push the information through to a RuneLite notification.

AFurryCat:
Closes #11229 

Added a notification that fires for each burner 5 seconds before they go into the random timer phase.
Positions added to all config items so that I could group the burner related bits together.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
@@ -25,20 +25,46 @@
 package net.runelite.client.plugins.poh;
 
 import java.time.Instant;
-import lombok.AllArgsConstructor;
+import javax.inject.Inject;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 
 @Getter
 @Setter
-@RequiredArgsConstructor
-@AllArgsConstructor
 class IncenseBurner
 {
+	@Inject
+	private PohConfig config;
+
 	private final Instant start = Instant.now();
+	private final int notificationOffset = 5;
 	private final int id;
+	private boolean notificationFired;
 	private double countdownTimer;
 	private double randomTimer;
+	private Instant stableEnd;
 	private Instant end;
+
+	public IncenseBurner(int id, double countdownTimer, double randomTimer, Instant end)
+	{
+		this.id = id;
+		this.countdownTimer = countdownTimer;
+		this.randomTimer = randomTimer;
+		this.end = end;
+
+		notificationFired = false;
+
+		// Convert seconds as double to milliseconds as long to avoid losing decimal fraction timings
+		long countdownMilliseconds = (long)((countdownTimer - notificationOffset) * 1000);
+		stableEnd = start.plusMillis(countdownMilliseconds);
+	}
+
+	public boolean checkIfShouldNotify()
+	{
+		if (notificationFired || stableEnd.compareTo(Instant.now()) >= 0)
+			return false;
+
+		notificationFired = true;
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
@@ -37,7 +37,6 @@ class IncenseBurner
 	private PohConfig config;
 
 	private final Instant start = Instant.now();
-	private final int notificationOffset = 5;
 	private final int id;
 	private boolean notificationFired;
 	private double countdownTimer;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
@@ -55,14 +55,16 @@ class IncenseBurner
 		notificationFired = false;
 
 		// Convert seconds as double to milliseconds as long to avoid losing decimal fraction timings
-		long countdownMilliseconds = (long)((countdownTimer - notificationOffset) * 1000);
+		long countdownMilliseconds = (long)((countdownTimer - config.notifyBurnerLeadTime()) * 1000);
 		stableEnd = start.plusMillis(countdownMilliseconds);
 	}
 
 	public boolean checkIfShouldNotify()
 	{
 		if (notificationFired || stableEnd.compareTo(Instant.now()) >= 0)
+		{
 			return false;
+		}
 
 		notificationFired = true;
 		return true;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
@@ -27,24 +27,17 @@ package net.runelite.client.plugins.poh;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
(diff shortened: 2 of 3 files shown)


AFurryCat:
This is finally ready :)

runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
```diff
@@ -25,20 +25,46 @@
...
+
+		// Convert seconds as double to milliseconds as long to avoid losing decimal fraction timings
+		long countdownMilliseconds = (long)((countdownTimer - notificationOffset) * 1000);
+		stableEnd = start.plusMillis(countdownMilliseconds);
+	}
+
+	public boolean checkIfShouldNotify()
+	{
+		if (notificationFired || stableEnd.compareTo(Instant.now()) >= 0)
+			return false;
```
melkypie, line 65: This should be surrounded by {}

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
```diff
@@ -27,146 +27,186 @@
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
+import net.runelite.client.config.Units;
 
 @ConfigGroup("poh")
 public interface PohConfig extends Config
 {
```
melkypie, line 35: What's the reason for all the reordering happening in config?

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
```diff
@@ -27,146 +27,186 @@
...
 		return true;
 	}
 
 	@ConfigItem(
-		keyName = "showSpellbook",
-		name = "Show Spellbook altar",
-		description = "Configures whether or not the Spellbook altar is displayed"
+		keyName = "notifyBurner",
+		name = "Burner Notifications",
+		description = "Configures whether or not to send a notification 5 seconds before burners enter the random timer phase",
```
melkypie, line 105: This shouldn't say that it sends a notification after 5 seconds as it seems to be configurable in the next option

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohPlugin.java
```diff
@@ -217,6 +228,37 @@ else if (actorName != null)
...
+	public void onGameTick(GameTick gameTick)
+	{
+		if (!config.notifyBurner() || incenseBurners.size() == 0)
+			return;
+
+		getIncenseBurners().forEach((tile, burner) ->
+		{
+			if (BURNER_LIT.contains(burner.getId()) && burner.checkIfShouldNotify())
+			{
+				if (!recentlySentNotification)
```
melkypie, line 241: This can be moved in the upper if statement

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohPlugin.java
```diff
@@ -217,6 +228,37 @@ else if (actorName != null)
 
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick gameTick)
+	{
+		if (!config.notifyBurner() || incenseBurners.size() == 0)
+			return;
```
melkypie, line 235: This also needs to be surrounded by {}

runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
```diff
@@ -25,20 +25,46 @@
...
 
 @Getter
 @Setter
-@RequiredArgsConstructor
-@AllArgsConstructor
 class IncenseBurner
 {
+	@Inject
+	private PohConfig config;
+
```
melkypie, line 38: This is unused

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
AFurryCat: I added positioning to the entire config menu to allow the burner notifications to be grouped.
It made sense to re-order the config props into numeric position order for easier maintainability moving forwards.

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
AFurryCat: Fixed

runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
AFurryCat: Oops, it is not used correctly as it should have been.

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohPlugin.java
AFurryCat: Its in its own if statement because if burner.checkIfShouldNotify() returns true, then the burner class will also have marked itself as notified on its internal variable and won't return true going forwards.

As such making sure burner.checkIfShouldNotify() fires before I suppress the notification means it won't just wait until after the suppression ends and then fire again.

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
melkypie: I don't see the reason what this has to do with what this pr's point is.

runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
```diff
@@ -25,20 +25,48 @@
...
 	private final Instant start = Instant.now();
+	private final int notificationOffset = 5;
```
melkypie, line 40: This is now unused

runelite-client/src/main/java/net/runelite/client/plugins/poh/PohConfig.java
AFurryCat: They retain the same order in the in-game menu.
They were simply re-ordered in the file to match. Had to add position to all elements for the grouping to work correctly.
As I added in the configurable time limit option after the toggle notification one, it was a bit of a pain to scroll up and down the file finding the next in the numeric sequence.

runelite-client/src/main/java/net/runelite/client/plugins/poh/IncenseBurner.java
```diff
@@ -25,20 +25,47 @@
...
+	{
+		this.id = id;
+		this.countdownTimer = countdownTimer;
+		this.randomTimer = randomTimer;
+		this.end = end;
+
+		notificationFired = false;
+
+		// Convert seconds as double to milliseconds as long to avoid losing decimal fraction timings
+		long countdownMilliseconds = (long)((countdownTimer - config.notifyBurnerLeadTime()) * 1000);
```
Adam-, line 57: This will just crash due to `config` being null.

PR #11658, opened 2020-05-21
Use multiplicative color offsets for BackgroundComponent border

deathbeam:
Instead of fixed color offsets for outer and inner border colors switch
to multiplicative offsets. This also makes overlay color alpha
recoloring work properly with background component border.


Nightfirecat:
What's the visual difference? Are there particular background colors for which this change works notably better or worse?

TheStonedTurtle:
This seems to be most notable on bright colors with low opacity.

left = PR  | right = Master
image

This is slightly noticeable with the default Overlay/Infobox background coloring as well
left = PR  | right = Master
image

PR #11660, opened 2020-05-21
banktags: save last opened tab when opening worn items

### LINKED ISSUE #10824: Keep Bank Tags open after viewing "Worn Items"
I use bank tag tabs and the new "worn items" all the time, however the 2 don't work well together. Currently, when a bank tag tab is open and you open and close "worn items", the tag tab closes. 

After closing your worn items the tag tab should stay open.

N/A

**Additional context**
https://i.imgur.com/VXGgS9k.gif

johannfrias:
Closes #10824. This request is in two parts.

Part 1/Commit 1: Addressing the cited issue directly

Diagnosing the problem, I found 3 behaviors:
1. Opening the "worn items" doesn't change the hidden state of tab interface
2. Saving the last tag tab happens *on the tick*
3. Switching from main bank interface to gear interface and back can be done on the same tick

Therefore the issue is present in these two scenarios:
1. You click "worn items", a tick happens and you check for hidden: tabs are not hidden, so don't save the last tab; at this point the last tab is forgotten
2. You click "worn items" and quickly return to main screen before a tick happens: the hidden check doesn't even happen so there's no chance to save the last tab; last tab is forgotten

My proposed fix modifies the definition of the tabs hidden state, and moves the code to save the last tab outside of the tick-controlled function `update` and into its own function, controlled by `onWidgetHiddenChanged`.

Part 2/Commit 2: Addressing a related issue found during my testing

*Independent* of the above work (i.e. reproducible on both master and my first commit), I found that in the following scenario, the saved tag tab gets overwritten:
After selecting a bank tag tab, exit the bank, (log out) and *fully close* RL client. Relaunch RL, log in and open bank.

My proposed fix adds a boolean to make sure that the code responsible for the error in behavior cannot possibly be run until the bank has been opened once after RL startup. (Short reason: the variable `activeTab` is not properly assigned (instead it's null) until bank is opened once after startup, which breaks some of the logic. Only after that does the logic with `activeTab` work properly.)


raiyni:
You can just set the config for the remembered tab on menu entry clicked

johannfrias:
If I'm understanding that suggestion right, then I would add an if condition where I check that I click the "Show worn items" button. But just to be clear, it's not just that case that causes the issue, it's clicking on anything that keeps you inside the bank but leaves the main interface that doesn't set the remembered tab correctly (example: clicking the "Show menu" button on the top right section). Which means, to rely on menu option clicked, I have to account for all actions that cause the wrong behavior. (It could very well be just those two buttons, so the fix seems not too extensive.)

And for the record, I did come across a minor bug in the following scenario, using my second commit, and I'll see if adding the suggestion fixes it:
Start with your bank without the incinerator, select a tag tab, then enable incinerator and return to main interface; the incinerator is rendered incorrectly in that there exists a second scroll down button right on top of it (exiting and reentering the bank fixes the render issue)
Edit: the deeper issue is that I seem to be rendering an extra copy of "Scroll up" and "Scroll down" buttons every time I leave and reenter the main bank screen while staying inside the bank. So it's more than a minor bug and I'll deal with that as well.

raiyni:
That was intended.

johannfrias:
Latest commits address the points most recently brought up, along with a newly found issue (the incinerator rendering) relating to (and diagnosed because of) my example scenario but not being a side effect of my initial work.

It was a bit hard to diagnose the widget duplication issue; that said, my belief is that, based on my testing, the if condition I chose of checking widget children differentiates between opening the bank (none of upButton, etc. have been created yet - no children), and clicking on "worn items" (per se) and returning to main interface without leaving the bank (upButton, etc. already created but not destroyed due to not leaving bank - yes children). Let me know if there's a scenario that breaks that logic.

raiyni:
Just remove your init you added.

johannfrias:
Latest commit removes the init logic from BankTagsPlugin, though I still need to run init somewhere in the case of switching bank interfaces, which leaves menu option clicked, as suggested before. So I added an if block similar to the save mechanism from an earlier commit, with an important addition of an invokeLater in order to make sure the hidden state is given time to change and the init runs as expected (b/c without invokeLater, init returns too early and the remembered tab is forgotten).

raiyni:
This is more what I was thinking, utilizing remembered search. You can just cherry pick this.

https://github.com/raiyni/runelite/commit/e989535899db069798406f503683dfc235f694f0

johannfrias:
That patch is an interesting idea, I got a better understanding of remembered search from that, and some logic is improved. After merging with my work, with the independent changes, it practically works well. I did decide to add a couple minor fixes:

* To resolve the flickering of panels that sometimes happens if "Hide worn items"/"Hide menu" is clicked, I trigger update upon click; also, invoked later so the panel actually closes first
* To resolve a stress test by rapid clicking equipment/menu button, I use *two* update() calls in that if block; the 1st update() restores the tag tab but leaves activeTab null; at this point, clicking the button again before the next tick forgets the saved tab, so an explicit 2nd update() is needed to accurately set activeTab

raiyni:
I don't see why you would want to `invokeLater` N times you click with double updates. The case you present isn't something I would consider a valid use case. It seems like you would be creating a race between you clicking, the tick event, and the invokeLater.

johannfrias:
If you say the rapid click is not a valid use case, then I won't account for it in code and will take out the second update.

I'd still like to force a single update upon click, since switching panels takes less than a tick and it'd update the tag tab right away. Else, you see the All tab for a split second before the tag tab. If that latter behavior is preferred, I'll take out the force update entirely.

raiyni:
Understandable what you are trying to do. I wonder if what you are trying to achieve is possible without having to invokeLater the update method.

johannfrias:
I'm going to say it's not possible without invokeLater because, at the exact moment of click, the interface has not yet changed (you can verify by querying the relevant widgets that the main interface is still hidden). An update at this point just returns early. Ergo, you must wait for the interfaces to switch, hence, invokeLater.

raiyni:
See https://github.com/raiyni/runelite/commit/83c37e414aa0c0a99c4079dd1a5f156297c8be33

johannfrias:
I stand corrected. This resolves both points I raised.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -150,6 +155,7 @@
 	private Instant startScroll = Instant.now();
 	private String rememberedSearch;
 	private boolean waitSearchTick;
+	private boolean initializedOnce;
```
raiyni, line 158: remove this

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -202,6 +208,7 @@ public void init()
 
 		currentTabIndex = config.position();
 		parent = client.getWidget(WidgetInfo.BANK_CONTENT_CONTAINER);
+		initializedOnce = true;
```
raiyni, line 211: remove this

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -1092,6 +1123,7 @@ private void updateBounds()
 			child.setType(WidgetType.GRAPHIC);
 			child.setSpriteId(TabSprites.INCINERATOR.getSpriteId());
 			incinerator.revalidate();
+			child.revalidate();
```
raiyni, line 1126: remove this

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -753,6 +771,19 @@ else if (event.getMenuAction() == MenuAction.RUNELITE
...
+			|| (event.getWidgetId() == WidgetInfo.BANK_TUTORIAL_BUTTON.getId() && event.getMenuOption().equals(SHOW_TUTORIAL))))
+		{
+			saveTab();
+			rememberedSearch = TAG_SEARCH + activeTab.getTag();
+		}
+		else if (!Strings.isNullOrEmpty(rememberedSearch) && ((event.getWidgetId() == WidgetInfo.BANK_EQUIPMENT_BUTTON.getId() && event.getMenuOption().equals(HIDE_WORN))
+				|| (event.getWidgetId() == WidgetInfo.BANK_SETTINGS_BUTTON.getId() && event.getMenuOption().equals(HIDE_SETTINGS))))
+		{
+			bankSearch.reset(true);
+			bankSearch.search(InputType.NONE, rememberedSearch, true);
```
raiyni, line 785: rememberedSearch should probably be reset after this

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
johannfrias: Disagree, but I will revert this only on the basis that it doesn't directly relate to this thread, but it makes improvements which I documented in 4333553, and this should be added in the future if not now.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
johannfrias: See comment "Disagree, but..." 4333553

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
johannfrias: See comment "Disagree, but..." 4ba0fc7

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
raiyni: The listed code is no longer relevant and is unused.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -578,6 +567,32 @@ else if (waitSearchTick)
 		scrollTab(0);
 	}
 
+	public void saveTab()
```
Adam-, line 570:
```suggestion
	private void saveTab()
```

PR #11667, opened 2020-05-22
Reset metronome tick count when toggled on.

Rustyspewns:
When the metronome is set to a high tick count and toggled off, the counter is left paused. When the plugin is resumed, the metronome will 'tick' after the remaining number of game ticks. 

This change resets the counter when it is toggled back on to allow the full # of game ticks set before another 'tick'.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
@@ -59,7 +59,7 @@ public class MetronomePlugin extends Plugin
 	}
 
 	@Override
-	protected void startUp()
+	protected void shutDown()
 	{
 		tickCounter = 0;
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
@@ -57,6 +57,7 @@ public class MetronomePlugin extends Plugin
 	{
 		return configManager.getConfig(MetronomePluginConfiguration.class);
 	}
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
```diff
@@ -57,6 +57,11 @@ MetronomePluginConfiguration provideConfig(ConfigManager configManager)
 	{
 		return configManager.getConfig(MetronomePluginConfiguration.class);
 	}
+	@Override
```
Nightfirecat, line 60: Separate these methods with a blank line
```suggestion

	@Override
```

runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
```diff
@@ -57,6 +57,11 @@ MetronomePluginConfiguration provideConfig(ConfigManager configManager)
...
+	protected void startUp()
```
Nightfirecat, line 61: Generally, resetting plugin state should happen when plugins shut down.
```suggestion
	protected void shutDown()
```

runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePlugin.java
```diff
@@ -57,6 +57,11 @@ MetronomePluginConfiguration provideConfig(ConfigManager configManager)
...
+	{
+		tickCounter = 0;
```
Nightfirecat, line 63: This is good. We should also reset `shouldTock`.
```suggestion
		tickCounter = 0;
		shouldTock = false;
```

Issue #11029, opened 2020-03-17
Linux GPU screenshots wrong dimensions HiDPI


Noiseko:
**Describe the bug**
Screenshot dimensions are wrong when GPU plugin is enabled. **regardless** of stretched mode or UI scaling type.

**To Reproduce**
GDK_SCALE=2 ./Runelite.AppImage

**Screenshots**
GPU on | Stretched mode off
![broken](https://user-images.githubusercontent.com/33142989/76842254-dede0580-6831-11ea-80f8-75a40dd5551a.jpg)
GPU off | Stretched mode off
![gpuoff](https://user-images.githubusercontent.com/33142989/76842261-e0a7c900-6831-11ea-8926-b0a7b62cc5c5.jpg)

**Environment (please complete the following information):**
OS: Arch Linux
KERNEL: 5.5.1-arch1-1
GPU DRIVER: 4.5 Mesa 19.3.3
Runelite version: 1.6.9
Launcher Version: 2.1.0

Adam-:
duplicate of #12463 and/or fixed in #12594

shuhaowu:
I'm still having this problem after the fix above. Specifically, I have GDK_SCALE=2 and GPU enabled (stretched mode is disabled), which gives me this:
![2020-11-12_03-34-51](https://user-images.githubusercontent.com/338100/99021652-2044c300-252f-11eb-8d5a-24a9ef3a0407.png)

With GDK_SCALE=2 and GPU disabled (stretch disabled), I get 
![2020-11-12_03-35-01](https://user-images.githubusercontent.com/338100/99021672-2d61b200-252f-11eb-9312-62564d6997f8.png)

GDK_SCALE=1 with GPU enabled or disabled gets the correct dimensions, albeit the runelite dialog is too small to view.

PR #11682, opened 2020-05-23
check GDK_SCALE for screenshots

Noiseko:
fix for https://github.com/runelite/runelite/issues/11029

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -1300,6 +1300,13 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 			height = dim.height;
 		}
 
+		int gdkScale = Integer.parseInt(System.getenv("GDK_SCALE"));
+		if (gdkScale > 1)
+		{
+			width = width * gdkScale;
+			height = height * gdkScale;
+		}
+
 		ByteBuffer buffer = ByteBuffer.allocateDirect(width * height * 4)
 			.order(ByteOrder.nativeOrder());
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -1300,11 +1300,15 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 			height = dim.height;
 		}
 
-		int gdkScale = Integer.parseInt(System.getenv("GDK_SCALE"));
-		if (gdkScale > 1)
+		String gdkScale_env = System.getenv("GDK_SCALE");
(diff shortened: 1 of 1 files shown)


Adam-:
1) is #11029 truly *only* with gpu enabled?
2) this will crash on any system where GDK_SCALE isn't defined

Noiseko:
1. Yes this is completely only when GPU is enabled.
2. Ah, I guess you have a solution of that then?

Adam-:
Just add a nullcheck to `getenv`?

Noiseko:
okay think it's addressed now?

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1300,6 +1300,24 @@ private Image screenshot()
...
+				int gdkScale = Integer.parseInt(gdkScaleEnv);
+				if (gdkScale > 1)
+				{
+					width = width * gdkScale;
+					height = height * gdkScale;
+				}
+			}
+			catch (NumberFormatException e)
+			{
+				log.info("GDK_SCALE is not valid");
```
Adam-, line 1317: this should probably be warn or debug. Also have you tested this with stretched mode?

Noiseko: Yeah it works fine with stretched mode

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1300,6 +1300,24 @@ private Image screenshot()
...
+				int gdkScale = Integer.parseInt(gdkScaleEnv);
+				if (gdkScale > 1)
+				{
+					width = width * gdkScale;
+					height = height * gdkScale;
+				}
+			}
+			catch (NumberFormatException e)
+			{
+				log.warn("GDK_SCALE is not valid");
```
Nightfirecat, line 1317: Log `gdkScaleEnv` so it's clear what it was actually set to

Adam-:
Does it also work if you use the same logic the glViewport code uses?

```
			final AffineTransform t = ((Graphics2D) canvas.getGraphics()).getTransform();
			gl.glViewport(
				getScaledValue(t.getScaleX(), x),
				getScaledValue(t.getScaleY(), y),
				getScaledValue(t.getScaleX(), width),
				getScaledValue(t.getScaleY(), height));
```

Nightfirecat:
Adam is asking if replacing the env var fetching with his code works the same way.

Adam-:
> 
> 
> I don't know what you are suggesting. I appreciate my implementation might not be the best way to fix this, I'll happily test a more suitable fix?

eg. Use `getScaledValue(t.getScaleX(), width)` / `getScaledValue(t.getScaleY(), height)` as the width/height instead.

Adam-:
fixed in #12594

PR #11689, opened 2020-05-24
Notification focus label

dylanstanfield:
Making this suggestion because I had a hard time finding this setting. I've changed it to `Notification focus` to match the setting below it `Notification sound`.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
@@ -151,7 +151,7 @@ public interface RuneLiteConfig extends Config
 
 	@ConfigItem(
 		keyName = "notificationRequestFocus",
-		name = "Request focus on notification",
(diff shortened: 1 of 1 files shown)


Nightfirecat:
"Notification focus" doesn't really explain it. "Focus on notification" would be better imo.

dylanstanfield:
@Nightfirecat I agree, I think what the runelite settings menu is missing is subsection labels. I'd still push for `Notification focus` because it pairs it nicely with the sound label and your suggestion would still hide part of the key word "notification" behind an ellipsis. And the description on hover does a good job of explaining the entire setting.

When I have some extra time i'll look into adding section labels for menus.

Broooklyn:
I took a stab at this a couple weeks ago: https://github.com/Adam-/runelite/pull/6/commits/633d00a3fba8b9c1a50a8752eb3d38e204981f3e

Didn't come to fruition but you're welcome to look for some inspiration.

PR #11695, opened 2020-05-25
Loot Tracker: Removes CoX Duplicate Tracking Events

### LINKED ISSUE #11114: Loot Tracker Registers CoX Completions on Opening Rewards Chest
**Description**
Loot tracker counts Chambers of Xeric loot based on the number of times the rewards chest is opened rather than actual completions. Opening the chest multiple times at the end of the raid raises the completion count and numbers of each item looted despite only a single real completion.

**To Reproduce**
1. Open loot tracker
2. Finish CoX
3. Open rewards chest, close it without withdrawing items
4. Open again - CoX completions and number of items in loot tracker will increase

**Expected behavior**
CoX loot tracker entry should increase based on actual completions and not chests being opened. Might be a tricky implementation but maybe there are events that could gate the logic? ex. saving the loot from the raid and then clearing that info once the raid is exited?

**Environment**
Windows 7/10, RL v1.6.10.1

basicDamen:
Closes #11114 .

Seeing as this is something only one individual seems to come across, it's very hard to test whether or not this FULLY works to resolve the duplicate problem, though I don't believe this additional check is going to hurt any.

Logistics behind the reasoning on why this should fix it:
- CoX and ToB are the only two chest-related tracking events that get scanned through, currently.
- CoX and ToB are both 'Instanced Regions'
- Whenever you leave CoX or ToB, you are no longer in an instanced region and it should flag the loading gamestate event, which is what the prior check solely relied on to reset the chestLooted variable.

I doubt you anyone _should_ be able to get a duplicate entry past this.. but.. here this is.

I HAVE tested that it works in CoX by doing a completion and it does successfully still track the loot there.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -368,7 +368,7 @@ public class LootTrackerPlugin extends Plugin
 	@Subscribe
 	public void onGameStateChanged(final GameStateChanged event)
 	{
-		if (event.getGameState() == GameState.LOADING)
+		if (event.getGameState() == GameState.LOADING && !client.isInInstancedRegion())
 		{
 			chestLooted = false;
 		}


Nightfirecat:
Especially since this is a bug which nobody else can reproduce, I'd like to see a test case added to show that this code change works where master currently doesn't.

basicDamen:
> Especially since this is a bug which nobody else can reproduce, I'd like to see a test case added to show that this code change works where master currently doesn't.

The only guy that can get the bug to happen, because he’s actually looking for it now.. is the player who submitted that issue. - I didn’t test if the bug actually exists, due to the fact that I don’t have the time nor the interest to do ‘x’ number of CoX to consistently test it.

I did a CoX run to ensure that the Loot Tracker plugin still worked as intended with my new check and it does. Therefore, it would not hurt to add and should realistically fix it. (Hence my initial information.) The only conclusion I could even somewhat gather would be somehow something loads for him right there and spits out the event to set the chestLooted variable back to false. 🤷🏼‍♂️ If he can recreate it without leaving an instance, then he’s just Houdini.

Nightfirecat:
@melkypie You had been doing some testing of this. Does this change look like it would address the issue? Based on what was discovered in Discord (loaded chunks can end on the tile with the chest loaded, therefore gamestate changes can be produced with fairly normal gameplay after opening chest) this approach seems fine to me.

basicDamen:
> @melkypie You had been doing some testing of this. Does this change look like it would address the issue? Based on what was discovered in Discord (loaded chunks can end on the tile with the chest loaded, therefore gamestate changes can be produced with fairly normal gameplay after opening chest) this approach seems fine to me.

Yeah, Adam, Melky, and I were all discussing it and we came to the conclusion that this fix would suffice.

PR #11699, opened 2020-05-25
Made TabManager Class Public

MitchBarnett:
Made the TabManager Class Public so it can be used within other plugins.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabManager.java
@@ -43,7 +43,7 @@ import net.runelite.client.util.Text;
 import org.apache.commons.lang3.math.NumberUtils;
 
 @Singleton
-class TabManager
+Public class TabManager
 {
 	@Getter
 	private final List<TagTab> tabs = new ArrayList<>();
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabManager.java
@@ -43,7 +43,7 @@ import net.runelite.client.util.Text;
 import org.apache.commons.lang3.math.NumberUtils;
 
 @Singleton
-Public class TabManager
+public class TabManager
 {
 	@Getter
 	private final List<TagTab> tabs = new ArrayList<>();


Nightfirecat:
Why? If you need this code for a hub plugin, just copy the class into your plugin.

MitchBarnett:
@Nightfirecat To me it seems better for code readability to import a class than copy it over I see no reason for the class to be private the precedent by other files is for the class to be public.

The class is a singleton so I thought copying the code and creating my own instance would be bad practice but I am new to this project so I don't know what is normal.

raiyni:
Currently there isn't a reason to have this exposed to other plugins. What would you even want that for? If you are generating tag tabs based on some other data, you can just insert/delete with config manager. 

What nightfirecat mentioned is how other external plugins function when they need to extend functionality of existing plugins.

MitchBarnett:
> What would you even want that for?

I would like my plugin to tag items and create a bank tab for them. Basically the functionality of  "Import tag tab". Tagging items is easy as TagManager is public, creating the tabs should be just as easy.

> If you are generating tag tabs based on some other data, you can just insert/delete with config manager.

Sure I can do that and will do that if necessary but that can lead to problems if the bank-tag plugin is updated and the format of the config data is changed and could lead to the data being corrupted. Using the existing class and it's fuctions stops this being an issue. The abstraction layer is nice for new plugin developers like me.

> What nightfirecat mentioned is how other external plugins function when they need to extend functionality of existing plugins.

OK I didn't know that was the standard practice here as I am new. That has the drawback that this is a singleton manager class which there is only designed to be one instance of and I would have to think how best to manage this, as you said above I could just edit the config data but this is a bigger task to make sure there are no issues with the bank-tabs plugin.

raiyni:
>I would like my plugin to tag items and create a bank tab for them. Basically the functionality of "Import tag tab". Tagging items is easy as TagManager is public, creating the tabs should be just as easy.

Great, then you can use config manager, tag tabs are just a list in 1 config key. TagManager is public because tags can be searched in the GE.


> Sure I can do that and will do that if necessary but that can lead to problems if the 
bank-tag plugin is updated and the format of the config data is changed and could lead to the data being corrupted. Using the existing class and its fuctions stops this being an issue. The abstraction layer is nice for new plugin developers like me.
> 

The format is only going to change if profiles are added. 

>That has the drawback that this is a singleton manager class which there is only designed to be one instance of and I would have to think how best to manage this, as you said above I could just edit the config data but this is a bigger task to make sure there are no issues with the bank-tabs plugin.

You don't need to copy anything? All of the added functions are pretty irrelevant to what you want to do. You also shouldn't be trying to manage the TagTab list outside of the bank anyway.

Nightfirecat:
As a rule of thumb, we do not make client changes for hub plugins, particularly not increasing the visibility of plugin classes. I would recommend following raiyni's advice for this case.

PR #11716, opened 2020-05-27
chat: Add chat notification for Jagex broadcast

Broooklyn:
Adds a chat notification option for Jagex broadcasts. 

There are times when you might not be paying attention to your client and/or chat, and Jagex sends out a broadcast. This ensures you don't miss any potentially important information, official Twitch streams, etc.
Screen Shot 2020-06-04 at 9 49 46 AM
2020-06-04_19-37-42 copy

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsConfig.java
@@ -97,4 +97,15 @@ public interface ChatNotificationsConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		position = 6,
+		keyName = "notifyOnBroadcast",
+		name = "Notify on broadcast",
+		description = "Notifies you whenever you receive a broadcast message"
+	)
+	default boolean notifyOnBroadcast()
+	{
+		return false;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
@@ -144,6 +144,22 @@ public class ChatNotificationsPlugin extends Plugin
 					notifier.notify(chatMessage.getMessage());
 				}
 				break;
+			case BROADCAST:
+				if (config.notifyOnBroadcast())
+				{
+					// Some broadcasts have links attached, notated by `|` followed by a number, while others contain color tags.
+					// We don't want to see either in the printed notification.
+					String broadcast = chatMessage.getMessage();
+
+					int urlTokenIndex = broadcast.lastIndexOf('|');
+					if (urlTokenIndex != -1)
+					{
+						broadcast = broadcast.substring(0, urlTokenIndex);
+					}
+
+					notifier.notify(Text.removeFormattingTags(broadcast));
+				}
+				break;
 			case CONSOLE:
 				// Don't notify for notification messages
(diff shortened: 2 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsConfig.java
```diff
@@ -97,4 +97,12 @@ default boolean notifyOnDuel()
...
 		return false;
 	}
+
+	@ConfigItem(
+			position = 6,
+			keyName = "notifyOnBroadcast",
+			name = "Notify on broadcast",
+			description = "Notifies you whenever Jagex sends a broadcast"
+	)
+	default boolean notifyOnBroadcast() { return false; }
```
Spedwards, line 107:
```suggestion
	default boolean notifyOnBroadcast()
	{
		return false;
	}
```

runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsConfig.java
```diff
@@ -97,4 +97,15 @@ default boolean notifyOnDuel()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			position = 6,
+			keyName = "notifyOnBroadcast",
+			name = "Notify on broadcast",
+			description = "Notifies you whenever Jagex sends a broadcast"
```
Nightfirecat, line 105: You need to fix your auto-formatter.
```suggestion
		position = 6,
		keyName = "notifyOnBroadcast",
		name = "Notify on broadcast",
		description = "Notifies you whenever Jagex sends a broadcast"
```

Broooklyn:
Screen Shot 2020-05-30 at 8 31 58 PM
Now works properly in places like LMS where broadcasts don't have links attached but have colored text instead.

runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
```diff
@@ -124,6 +126,30 @@ private void updateHighlights()
 		}
 	}
 
+	/**
+	 * Remove tags, except for &lt;lt&gt; and &lt;gt&gt;
+	 *
+	 * @return
+	 */
+	private static String removeTags(String str)
```
Nightfirecat, line 134: Instead of copying this from EmojiPlugin, move the method from there to `Text` (naming it `removeFormattingTags`) and make it public in a commit prior to the current changes. (and add tests for it--see `TextTest`'s removeTags test cases)

runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
```diff
@@ -144,6 +170,20 @@ public void onChatMessage(ChatMessage chatMessage)
...
+					// Broadcasts can accompanied by a clickable link that is notated by `|` followed by a number that determines which website is linked.
+					// This prevents it from printing in broadcast notifications.
+					String broadcast = removeTags(chatMessage.getMessage());
+					notifier.notify(broadcast.substring(0, broadcast.lastIndexOf('|')));
+				}
+				if (config.notifyOnBroadcast())
+				{
+					String broadcast = removeTags(chatMessage.getMessage());
+					notifier.notify(broadcast);
+				}
```
Nightfirecat, line 185: This would notify twice on broadcasts that have links. I recommend combining the bodies of these two conditional blocks by doing the following: make a guard clause for `config.notifyOnBroadcast()`, then a check for `broadcast.lastIndexOf('|')` being `-1` (if it isn't, get the substring prior to that vertical bar). Finally, remove the formatting tags, then notify that string.

Issue #1983, opened 2018-04-26
Pyramid Plunder - Add door highlights


Caia-OSRS:
It would be nice to have the 4 doors in each room highlighted similar to agility click boxes.

Nightfirecat:
Changing this to request only door highlights as #2377 requests a precise timer.

Issue #2377, opened 2018-05-05
Exact timer overlay for pyramid plunder


Stickaplex:
can be hard to tell how much time is left so this would be handy

RebornFX:
The normal timer also tends to get in the way of the XP globes which isn't terrible but not ideal, so yeah would love this addition.

Frankkz:
+1 Would be nice to have a countdown timer, each session is 5 minutes.

jacksonbob10199:
doing pp now and thought of this, would be nice

dekvall:
Added in #11718

PR #11718, opened 2020-05-27
Pyramid plunder

Mitchell-Kovacs:
Adds a pyramid plunder plugin that can:
- replace the default large widget with an overlay (77f5540)
- highlight unopened doors and speartraps (09aa3ea)
- highlight unopened urns, chest and sarcophagus based on floor (5b685a4)

Here are the options:
image

This is what the custom overlay looks like. I went with a overlay instead of an infobox because I wanted to get rid of the widget as it was getting in my way while playing in fixed mode and still wanting to see information such as floor. You will also get an exact timer instead of the bar. When override overlay is on you will get a new overlay and the widget will be hidden, you can also toggle thieving level required as it's not very important.
image

Here is what the highlight doors and highlight spears turned on look like. Note that once a door is opened it will no longer remain highlighted.
image

You can also chose to highlight the urns, chest and sarcophagous if you prefer to click on overlays instead of game objects. It will only highlight them if they are closed. This setting works by selecting a floor. It will then highlight those containers on that floor and up. Setting to 9 effectively disables  highlighting for that object. While I can see how this is a colourful mess, I actually found it useful even though I did not think I would. It's also toggleable so I don't see the harm.

image

After searching I found that a pyramid plunder plugin was removed at some point in time as it was deemed no longer necessary. I find when playing on fixed mode that the widget is still overly large so I would prefer a way to have it removed.

I'm not the most familiar with runelites code, so if you have any suggestions I would be open to them.

**Note:** I'm a pleb that does not have 91 thieving so I could not check that last floor. My experience with the other floors though is that it should not be a problem.

Here are the related issues I could find:
Closes: https://github.com/runelite/runelite/issues/1983, https://github.com/runelite/runelite/issues/2377


Mitchell-Kovacs:
I've updated after receiving some feedback on discord.
- Use an infobox instead of an overlay
- Use convexhull instead of clickboxes for the containers

Here is what the updated setting menu looks like:
image

Hide default timer configures if you want to see the pyramid plunder timer.
Show exact timer adds the timer infobox:
image

I now use convex hulls to highlight containers instead of clickboxes:
image

Additionally, I now remove the speartrap clickboxes once you have moved pass them:

hide speartrap clickbox

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderObjects.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Set;
+import static net.runelite.api.ObjectID.GRAND_GOLD_CHEST;
+import static net.runelite.api.ObjectID.SARCOPHAGUS_21255;
+import static net.runelite.api.ObjectID.SPEARTRAP_21280;
+import static net.runelite.api.ObjectID.TOMB_DOOR_20948;
+import static net.runelite.api.ObjectID.URN_21261;
+import static net.runelite.api.ObjectID.URN_21262;
+import static net.runelite.api.ObjectID.URN_21263;
+
+public class PyramidPlunderObjects
```
Adam-, line 37:
```suggestion
class PyramidPlunderObjects
```

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderObjects.java
```diff
@@ -0,0 +1,53 @@
...
+{
+	public static final Set<Integer> TOMB_DOOR_WALL_IDS = ImmutableSet.of(26618, 26619, 26620, 26621);
```
Adam-, line 39: All of the objects in this class should be referencing ObjectID/NullObjectID

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -0,0 +1,172 @@
...
+		setLayer(OverlayLayer.ABOVE_SCENE);
+		this.client = client;
+		this.plugin = plugin;
+		this.config = config;
+	}
+
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		if (client.getLocalPlayer() == null)
```
Adam-, line 75: This can't happen

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderTimer.java
```diff
@@ -0,0 +1,44 @@
...
+import java.awt.Color;
+import java.awt.image.BufferedImage;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import net.runelite.api.Client;
+import net.runelite.api.Varbits;
+import net.runelite.client.ui.overlay.infobox.Timer;
+
+public class PyramidPlunderTimer extends Timer
```
Adam-, line 12:
```suggestion
class PyramidPlunderTimer extends Timer
```

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderTimer.java
```diff
@@ -0,0 +1,44 @@
+package net.runelite.client.plugins.pyramidplunder;
```
Adam-, line 1: Add license header to each new file.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+			tilesToHighlight.put(object, event.getTile());
+		}
+	}
+
+	@Subscribe
+	public void onGameObjectSpawned(GameObjectSpawned event)
+	{
+		GameObject object = event.getGameObject();
+
+		if (object == null)
```
Adam-, line 180: this can't happen

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+		{
+			objectsToHighlight.add(object);
+		}
+	}
+
+	public boolean isInPyramidPlunder()
+	{
+		int timer = client.getVar(Varbits.PYRAMID_PLUNDER_TIMER);
+		return client.getLocalPlayer() != null
+			&& timer > 0
```
Adam-, line 201: This region check is much faster, so check it first.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+				int ppTimer = client.getVar(Varbits.PYRAMID_PLUNDER_TIMER);
+				long timeLeft = (PYRAMID_PLUNDER_DURATION - ppTimer) * Constants.GAME_TICK_LENGTH;
+				timer = new PyramidPlunderTimer(timeLeft, itemManager.getImage(PHARAOHS_SCEPTRE), this,
+					config, client);
+				infoBoxManager.addInfoBox(timer);
+			}
+		}
+		else if (timer != null)
+		{
+			infoBoxManager.removeIf(infoBox -> infoBox instanceof PyramidPlunderTimer);
```
Adam-, line 154:
```suggestion
			infoBoxManager.removeInfoBox(timer);
```

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+	@Override
+	protected void shutDown() throws Exception
+	{
+		tilesToHighlight.clear();
+		objectsToHighlight.clear();
+		overlayManager.remove(overlay);
+		timer = null;
+		infoBoxManager.removeIf(infoBox -> infoBox instanceof PyramidPlunderTimer);
+
+		Widget ppWidget = client.getWidget(WidgetInfo.PYRAMID_PLUNDER_DATA);
```
Adam-, line 121: this should be in an `clientThread.invoke()`

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+		{
+			tilesToHighlight.clear();
+			objectsToHighlight.clear();
+		}
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick tick)
+	{
+		if (config.showExactTimer() && isInPyramidPlunder())
```
Adam-, line 141: Instead of all of this logic it may be better to just override Infobox `render()`

```java
@Override
public boolean render() {
    return config.showExactTimer();
}
```

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+			timer = null;
+		}
+	}
+
+	@Subscribe
+	public void onWallObjectSpawned(WallObjectSpawned event)
+	{
+		WallObject object = event.getWallObject();
+
+		if (object == null)
```
Adam-, line 164: this can't happejn

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
```diff
@@ -0,0 +1,204 @@
...
+@PluginDescriptor(
+	name = "Pyramid Plunder",
+	description = "Show custom overlay for Pyramid Plunder",
+	tags = {"minigame"},
+	enabledByDefault = false
+)
+public class PyramidPlunderPlugin extends Plugin
+{
+	// Total number of game ticks until the pyramid plunder round ends (5 minutes)
+	private static final int PYRAMID_PLUNDER_DURATION = 500;
```
Adam-, line 71: We have an GAME_TICKS temporal unit in `RSTimeUnit` if you want to use that.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -0,0 +1,172 @@
...
+		if (ppWidget == null)
+		{
+			return null;
+		}
+		ppWidget.setHidden(config.hideTimer());
+
+
+		// Highlight convex hulls of urns, chests, and sarcophagus
+		int currentFloor = client.getVar(Varbits.PYRAMID_PLUNDER_ROOM);
+		plugin.getObjectsToHighlight().forEach(object ->
```
Adam-, line 90: Prefer loops to `forEach` if it doesn't decrease readability.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
Mitchell-Kovacs: Just for my understanding, why should this be in `clientThread.invoke()`?

Mitchell-Kovacs:
Thanks for the review, I've updated to address your feedback. Sorry in advance if I misunderstood any of it/did not do the expected change. Also I figure that the commits will all be collapsed on merge but if not I can clean them up to however you would like. 
Ready for re-review.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderPlugin.java
Adam-: shutDown runs on the EDT and it is mostly not safe to mutate client state off of the client thread.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -0,0 +1,168 @@
...
+		if (ppWidget == null)
+		{
+			return null;
+		}
+
+		ppWidget.setHidden(config.hideTimer());
+
+		// Highlight convex hulls of urns, chests, and sarcophagus
+		int currentFloor = client.getVar(Varbits.PYRAMID_PLUNDER_ROOM);
+		for (GameObject object : plugin.getObjectsToHighlight())
```
raiyni, line 86: These need distance checks probably. It highlights the objects in all of the rooms.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderTimer.java
```diff
@@ -0,0 +1,74 @@
...
+import java.awt.Color;
+import java.awt.image.BufferedImage;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import net.runelite.api.Client;
+import net.runelite.api.Varbits;
+import net.runelite.client.ui.overlay.infobox.Timer;
+
+class PyramidPlunderTimer extends Timer
```
raiyni, line 36: The time displayed left seems to end a few seconds early in the infobox.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -0,0 +1,168 @@
...
+	@Override
+	public Dimension render(Graphics2D graphics)
+	{
+		Widget ppWidget = client.getWidget(WidgetInfo.PYRAMID_PLUNDER_DATA);
+		if (ppWidget == null)
+		{
+			return null;
+		}
+
+		ppWidget.setHidden(config.hideTimer());
```
raiyni, line 82: Do this in `onConfigChanged` and startup

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderConfig.java
```diff
@@ -0,0 +1,154 @@
...
+
+	@ConfigItem(
+		position = 8,
+		keyName = "highlightUrnsFloor",
+		name = "Highlight urns floor",
+		description = "Highlight the urns starting at selected floor and up"
+	)
+	default int highlightUrnsFloor()
+	{
+		return 9;
```
raiyni, line 130: Defaulting these to 9 seems too restricted for the standard player?

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
Mitchell-Kovacs: As in remove this logic entirely from the overlay and put it in those places or also just include this logic in those places?

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderConfig.java
Mitchell-Kovacs: My thinking is that it fills the screen up with the overlay so the standard player would not want it on, it also had the added benefit of not trying to figure out how the average player plays. I think efficient starts at being able to loot floor 8 and uses the following:
- urns starting floor 7
- chest starting floor 6
- sarcophagus starting never (9)

What numbers do you think I should use?

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
raiyni: Move it

Adam-: This won't work if done on startup/config changed

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderConfig.java
raiyni: Oh, right. There is only 8 floors. That seems fine to use 9 as off then by default.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderTimer.java
Adam-: This is probably just tick drift, all of the long running timers have it.

Mitchell-Kovacs: It's actually just me being dumb and using `duration.get(ChronoUnit.SECONDS)` thinking it will return the seconds remaining and not just the seconds portion of the duration. It also looks like you get kicked out one game tick after the timer hits 500.

Broooklyn:
Might just be me but I feel the timer is a bit hard to read due to it overlapping with the Sceptre icon 
Screen Shot 2020-06-22 at 8 14 55 PM

Mitchell-Kovacs:
> Might just be me but I feel the timer is a bit hard to read due to it overlapping with the Sceptre icon
> Screen Shot 2020-06-22 at 8 14 55 PM

I didn't really have a hard time reading it in game but from your screenshot it does look like it could be a bit difficult. If you are anyone else have a suggestion for a better image it's fine by me, I don't really care what it is.

Broooklyn:
`GOLDEN_SCARAB` looks the best to me (probably because the icon goes the opposite direction of the other items, but I can definitely appreciate the fact that `PHARAOHS_SCEPTRE` is the most thematically fitting. 

Screen Shot 2020-06-22 at 9 14 24 PM

Issue #11703, opened 2020-05-26
Toggle on/off clan chat join/leave message dissappearing


Maho70:
The fact that the message of the person who joined the clan chat dissappears after a couple of seconds and this is annoying. 

I wish this can be toggled. So you could choose if the message should dissappear or not.

PR #11724, opened 2020-05-28
Add timeout config option to clanchat join/leave message dissappearing.

ScanlonJ:
Adds a new config option to the clan chat plugin. 
Allows the user to set the tick delay before a Join/Leave message is removed. 
Setting the delay to 0 stops the messages from being removed. 

closes [#11703](https://github.com/runelite/runelite/issues/11703)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/friendschat/FriendsChatConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/friendschat/FriendsChatConfig.java
@@ -107,10 +107,21 @@ public interface FriendsChatConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "joinLeaveTimeout",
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/clanchat/ClanChatPlugin.java
```diff
@@ -128,6 +127,7 @@
 	 */
 	private final Deque<ClanJoinMessage> clanJoinMessages = new ArrayDeque<>();
 	private Map<String, ClanMemberActivity> activityBuffer = new HashMap<>();
+	private int join_leave_duration;
```
Adam-, line 130: It is not necessary store this here, you can just call `config.joinLeaveTimeout()` in `timeoutClanMessages`

PR #11736, opened 2020-05-28
Add random functionality to login screen override

### LINKED ISSUE #11731: Add login screen background option "Random"
Relating to this commit which added custom login screen backgrounds, it would be great if there was an additional option for randomly selecting between the background options (excluding "Custom" IMO) 7cc36acd9ad549978ce804eb789d655ab6f07084

**Additional context**
I think on runelite client startup it could randomly select a menu option to make active. Doesn't have to be on client logout, but that would be up to the author

MitchBarnett:
closes #11731 

Added an option to the LoginScreen plugin to set the background to be random.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
@@ -33,6 +33,8 @@ import java.awt.event.KeyEvent;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Random;
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
@@ -276,6 +278,22 @@ public class LoginScreenPlugin extends Plugin implements KeyListener
 					return;
 				}
 			}
+		}
+		if (config.loginScreen() == LoginScreenOverride.RANDOM)
+		{
+
+			ArrayList<String> loginScreenOverrides = new ArrayList<String>();
+
+			for (LoginScreenOverride loginScreenOverride : LoginScreenOverride.values())
+			{
+				if (loginScreenOverride.getFileName() != null)
+				{
+					loginScreenOverrides.add(loginScreenOverride.getFileName());
+				}
+			}
+
+			Random random = new Random(System.nanoTime());
+			pixels = getFileSpritePixels(loginScreenOverrides.get(random.nextInt(loginScreenOverrides.size())));
 		}
 		else
 		{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenOverride.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenOverride.java
@@ -42,7 +42,8 @@ public enum LoginScreenOverride
 	MONKEY_MADNESS_2("mm2.jpg"),
 	PRIFDDINAS("prifddinas.jpg"),
 	THEATRE_OF_BLOOD("tob.jpg"),
-	CUSTOM;
+	CUSTOM,
+	RANDOM;
 
 	@Getter
 	private final String fileName;


runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -277,6 +279,22 @@ private void overrideLoginScreen()
 				}
 			}
 		}
+		if (config.loginScreen() == LoginScreenOverride.RANDOM)
```
Spedwards, line 282: Should be an `else if`

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -277,6 +279,22 @@ private void overrideLoginScreen()
 				}
 			}
 		}
+		else if (config.loginScreen() == LoginScreenOverride.RANDOM)
+		{
+
+			ArrayList<String> loginScreenOverrides = new ArrayList<>();
```
raiyni, line 285: You can just use the values array.

MitchBarnett: OFF, CUSTOM and RANDOM don't have a associated file name so I build a new list leaving them out.

raiyni: The enum is in a specific order.

MitchBarnett: I don't see how the order affects things.

Nightfirecat: The order isn't important, but there is no need to create a new list for this when you can `Arrays.stream()` the values and choose a random from that, via `Stream.skip(randomInt)` for instance.

raiyni:
```java
/**
* Returns an array containing the constants of this enum   
* type, in the order they're declared.  This method may be  
* used to iterate over the constants as follows:  
*  
*    for(E c : E.values())  
*        System.out.println(c);  
*  
* @return an array containing the constants of this enum   
* type, in the order they're declared  
*/  
public static E[] values();
```

Then you can do

```
int idx = ThreadLocalRandom.current().nextInt(1, enum.values().length - 1);
//or
int idx = new Random().nextInt(enum.values().length - 1) + 1;
String fileName = enum.values()[idx];
```

MitchBarnett: > ```
> int idx = ThreadLocalRandom.current().nextInt(1, enum.values().length - 1);
> //or
> int idx = new Random().nextInt(enum.values().length - 2) + 1;
> String fileName = enum.values()[idx];
> ```

Personally I don't like the idea of using magic numbers to filter out the the first and last 2 elements of the enum.

raiyni: Those aren't magic numbers.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -326,4 +337,36 @@ private SpritePixels getFileSpritePixels(String file)
...
+
+			if (image.getHeight() > Constants.GAME_FIXED_HEIGHT)
+			{
+				final double scalar = Constants.GAME_FIXED_HEIGHT / (double) image.getHeight();
+				image = ImageUtil.resizeImage(image, (int) (image.getWidth() * scalar), Constants.GAME_FIXED_HEIGHT);
+			}
+
+			return ImageUtil.getImageSpritePixels(image, client);
+		}
+		catch (IOException | NullPointerException e)
```
Nightfirecat, line 365: Catching `NullPointerException` is new compared to master's version of this code. Where would we expect to see that? I see `ImageIO.read()` would throw `IllegalArgumentException` if `file` was null, but that case isn't handled either?

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -245,59 +251,58 @@ private void overrideLoginScreen()
...
+						.map(Path::toFile)
+						.collect(Collectors.toList());
+
+					File selectedCustomScreen = customScreens.get(random.nextInt(customScreens.size()));
+
+					pixels = getFileSpritePixels(selectedCustomScreen);
 				}
 				catch (IOException e)
 				{
 					log.error("error loading custom login screen", e);
```
Nightfirecat, line 299: Is this accurate when `IOException` would be thrown by `Files.walk()`?

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -245,59 +251,58 @@ private void overrideLoginScreen()
...
 				{
-					BufferedImage image;
-					synchronized (ImageIO.class)
-					{
-						image = ImageIO.read(CUSTOM_LOGIN_SCREEN_FILE);
-					}
+					pixels = getFileSpritePixels(CUSTOM_LOGIN_SCREEN_FILE);
+				}
+				break;
+			case RANDOM:
```
Nightfirecat, line 268: For the RANDOM and RANDOM_CUSTOM options, it might be nice to choose a random image each time the login screen is reached, not just when the config is set. Not a must-have, but I feel that would be what users using this option would want.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: From my testing it threw `NullPointerException ` but I will do some more testing to confirm.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: I will do some testing on this.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: I just attempted to add this using the following code

```
if (event.getGameState() == GameState.LOGIN_SCREEN)
{
		applyUsername();

		if (config.loginScreen() == LoginScreenOverride.RANDOM ||
		config.loginScreen() == LoginScreenOverride.RANDOM_CUSTOM)
		{
				overrideLoginScreen();
		}
}
```
This results in an infinite loop as overrideLoginScreen(); causes a GameStateChanged.LOGIN_SCREEN event. I will have to think of an alternative method, for this functionality.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: I'm not too sure what you mean by accurate but when IOException exception is thrown when "login-screen" folder is not found, therefore pixels will still be null so the following applies:

```
if (pixels != null)
{
	client.setLoginScreen(pixels);
}
```

My latest commit address the case that the "login-screen" folder exists but contains no image files.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: I added this as `NullPointerException `is thrown when an invalid file is read (such as an .mp3 file renamed to .png). In the current code there is no way for `file` to be null but it would be worth adding a check for this in case the code is restructured in the future.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -243,43 +251,64 @@ private void overrideLoginScreen()
...
 
-		if (config.loginScreen() == LoginScreenOverride.OFF)
-		{
-			restoreLoginScreen();
-			return;
-		}
-
 		SpritePixels pixels = null;
-		if (config.loginScreen() == LoginScreenOverride.CUSTOM)
+		Random random = new Random(System.nanoTime());
```
raiyni, line 255: You don't need to seed random.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -308,4 +343,36 @@ private SpritePixels getFileSpritePixels(String file)
...
+				final double scalar = Constants.GAME_FIXED_HEIGHT / (double) image.getHeight();
+				image = ImageUtil.resizeImage(image, (int) (image.getWidth() * scalar), Constants.GAME_FIXED_HEIGHT);
+			}
+
+			return ImageUtil.getImageSpritePixels(image, client);
+		}
+		catch (IOException | NullPointerException e)
+		{
+			log.error("error loading custom login screen: " + file, e);
+			restoreLoginScreen();
```
raiyni, line 374: This method shouldn't be doing login screen operations.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
raiyni: You would have to set a boolean that gets reset when you log in.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: Ok I will remove I just saw that is was done in [WorldSupplier.java](https://github.com/runelite/runelite/blob/5494249636322830e2672da905bd1e1eec81a53f/runelite-client/src/main/java/net/runelite/client/rs/WorldSupplier.java) so I did it to be safe

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: No I'm not sure why I added that there.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
raiyni: You would have to set a boolean that gets reset when you log in.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: This functionality has been added on latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: This functionality has been added on latest commit

runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
MitchBarnett: Fixed on latest commit.

hjdarnel:
This would be cool to get merged in. It looks up to date and no conflicts. Thanks for the work, author and reviewers :)

PR #11739, opened 2020-05-28
worldhopper: Fix other favorited worlds to retain menu state after re…

### LINKED ISSUE #11726: Removing favorited worlds
**Describe the bug**
While removing favorited world from World Switcher plugin it resets menu state of every world, not just the one selected, so when you click on any other favorited world except the one you already removed it says "Add to favorites" instead of "Remove from favorites".
It applies to menu only, when you restart the client the rest of favorited worlds is still there.

**To reproduce**
1. Open World Switcher tab.
2. Add any amount of worlds to favorited list.
3. Remove any world from favorites.
4. If you try to remove any other world it will now say "Add" instead of "Remove" world from favorites.

RuneliteBug

**Expected behavior**
It should be possible to remove multiple worlds from favorites without restarting the client/adding the world again & then removing it.

**Environment**
RuneLite 1.6.16

**Additional info**
From what I've checked it happens because `panel.resetAllFavoriteMenus()` is being called in the `WorldHopperPlugin.clearFavoriteConfig` method.

naleba97:
…moving a favorited world

Replaced `panel.resetAllFavoriteMenus()` in `WorldHopperPlugin.clearFavoriteConfig()` with `updateList()`. The world hopper plugin no longer resets the menu state of every world to non-favorite after removing a favorited world. Right-clicking favorited worlds will now display "Remove from favorites".

Closes #11726

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -298,7 +298,7 @@ public class WorldHopperPlugin extends Plugin
 	private void clearFavoriteConfig(int world)
 	{
 		configManager.unsetConfiguration(WorldHopperConfig.GROUP, "favorite_" + world);
-		panel.resetAllFavoriteMenus();
+		updateList();
 	}
 
 	boolean isFavorite(World world)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
@@ -298,7 +298,6 @@ public class WorldHopperPlugin extends Plugin
 	private void clearFavoriteConfig(int world)
 	{
 		configManager.unsetConfiguration(WorldHopperConfig.GROUP, "favorite_" + world);
-		updateList();
 	}
 
 	boolean isFavorite(World world)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -298,7 +298,7 @@ private boolean isFavoriteConfig(int world)
 	private void clearFavoriteConfig(int world)
 	{
 		configManager.unsetConfiguration(WorldHopperConfig.GROUP, "favorite_" + world);
-		panel.resetAllFavoriteMenus();
```
Nightfirecat, line 301: This method is no longer used after this change, delete it from `WorldSwitcherPanel`.

naleba97: Got it, removed.

runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldHopperPlugin.java
```diff
@@ -298,7 +298,7 @@ private boolean isFavoriteConfig(int world)
...
+		updateList();
```
Adam-, line 301: Is this actually necessary? `removeFromFavorites` looks like it is already correctly updating the favorite menu so I don't see why this is required.

Adam-: I've tested this and it seems unnecesary

PR #11742, opened 2020-05-28
remove stamina timer upon entering the Gauntlet

### LINKED ISSUE #11700: Stamina timer in Gauntlet
**Describe the bug**
If you leave Gauntlet with stamina boost on, you'll keep it in the main game. But if you enter The Gauntlet again you lose the stamina boost but the stamina timer is still there.

**To Reproduce**
1. Get stamina boost (by drinking stamina potion in main game OR drinking Egniol potion in Gauntlet and leaving (works the same if you finish gauntlet with stamina boost still on))
2. Enter Gauntlet again

**Screenshots**
https://streamable.com/krkf2u

branisk:
Closes #11700


Nightfirecat:
This would partially address #11700

branisk:
> 
> 
> This would partially address #11700

Can you elaborate on the other part of the problem?  What am i missing?

Nightfirecat:
Egniol potions from the gauntlet grant a stamina effect while in the minigame which should also be cleared when leaving.

branisk:
> 
> 
> Egniol potions from the gauntlet grant a stamina effect while in the minigame which should also be cleared when leaving.

I just tested it, it seems the effect is still active upon leaving the gauntlet, so it should not be cleared.

Nightfirecat:
Ah, my bad then--I misunderstood the issue. :+1:

Issue #11746, opened 2020-05-29
Kebos Lowlands mining spot.


KennyXOXX:
When highlighting over this spot it says 2 mith 9 iron. However.
[![Screenshot from Gyazo](https://gyazo.com/d97cd39c4a290fdf9a4030e0ba2e28f7/raw)](https://gyazo.com/d97cd39c4a290fdf9a4030e0ba2e28f7)
 
What is actually here is. 
[![Screenshot from Gyazo](https://gyazo.com/500688fa4c9c7ae44b67b2401e7a887d/raw)](https://gyazo.com/500688fa4c9c7ae44b67b2401e7a887d)

PR #11752, opened 2020-05-29
Kebo's Lowlands fix mining rocks

DovydasAL:
Mining rocks were listed incorrectly as per issue #11746


makingcomment:
Before this potentially gets closed because #11746 was already closed by #11748, I'd like to note that #11748 listed the adamantite before the mithril, which is inconsistent with the rest of the mining sites.

dekvall:
Superseded by #11748 (which had the wrong order :( )

PR #11769, opened 2020-05-30
Show GE and HA value in the title while searching

JZomerlei:
Bank value shows as normal on tags but with an additional <br>
image

Bank value shows when using tag searching
image

Farming tag shows along with Faceguards (search term `fa`)
image


raiyni:
I'm pretty sure this will break tag tab selection while using the chatbox search

JZomerlei:
> I'm pretty sure this will break tag tab selection while using the chatbox search

It's odd behavior but works. :)
image

raiyni:
That has nothing to do with what I said.

raiyni:
This needs to be done a different way probably. It breaks the the edit-tags options and probably other menu options that rely on the bank title.

JZomerlei:
2899dd2 Fixes that issue. The open tag string was including the price string so it no longer matched. 

Not sure of a better way of doing it than what I did though.

raiyni:
I think the better solution is offloading most of the update method into a post fired of script 277. There are probably edge cases that need be handled differently at that point though. 

Then you can set the title after all of that logic.

JZomerlei:
How should I deal with the (GE: val) (HA: val) when the activateTab is tried again? When you first load up the tag tabs it uses`client.getVar(VarClientStr.INPUT_TEXT)` but when you right click edit tags that value is blank so it uses the bank title which has the price text in it. 

I've got three methods that I've fiddled around with and they all `work` but I don't know what would be best.

- substring the text to remove `str = str.substring(0, str.indexOf("</col>") + 6)`
- Regex match on `Pattern.compile("<col=ff0000>(.+?)</col>", Pattern.DOTALL)`  and use it to get the value between the color tags which is the search term/the selected tag tab name.
```		Matcher matcher = pattern.matcher(str);
		if (matcher.find())
		{
			str = matcher.group(1); // gets value within the color tags
		}
		else
		{
			str = Text.standardize(str); // normal happy path
		}
```
- The last option I've tried is to `replaceAll(priceText, "")` I would have to get the priceText used from the onScriptPostFired (Also make it static/get for it)

Ideally I would get the text from the ChatboxPanelManager but if the item has multiple tags it would break anyhow. So I stopped going down that route once I saw multiple tags mucking with it.

JZomerlei:
Nuances that still work:
Tag tab stuff
- Renaming a bank tag tab changes all the tags on the items it contained. 
- Delete Tab and and tag from all items
- Delete tab only
- Create Tag tab
- Import Tag tabs
- View Tag tabs(and subsequently clicking the tab you want to see)

Items stuff
- Right clicking item and removing current tab tag (if you're on the active tag)
- Right clicking item and editing tags

I think that's all there was to check

runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
```diff
@@ -542,6 +542,11 @@ else if (!Strings.isNullOrEmpty(config.tab()))
 		if (bankTitle != null && !bankTitle.isHidden() && !str.startsWith(TAG_SEARCH))
 		{
 			str = bankTitle.getText().replaceFirst("Showing items: ", "");
+			// trim off the price
+			if (str.contains("<br>"))
```
Nightfirecat, line 546: This should be `else if` after the startsWith("Tab ") check

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -283,6 +293,48 @@ public void onGrandExchangeSearched(GrandExchangeSearched event)
 		client.setGeSearchResultIds(Shorts.toArray(ids));
 	}
 
+	@Subscribe
+	public void onScriptPreFired(ScriptPreFired event)
+	{
+		// 277 is the ID of BankSearchLayout.rs2asm, just want to clear the container before searching by tag.
```
Nightfirecat, line 299: Define it in ScriptID instead of hardcoding it

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -283,6 +293,48 @@ public void onGrandExchangeSearched(GrandExchangeSearched event)
...
+		{
+			tabContainer.clear();
+		}
+	}
+
+	@Subscribe
+	public void onScriptPostFired(ScriptPostFired event)
+	{
+		// 277 is the ID of BankSearchLayout.rs2asm, set the bank title after it runs.
+		if (event.getScriptId() == 277)
```
Nightfirecat, line 310: Change this into a guard clause instead of a giant if statement block.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -283,6 +293,48 @@ public void onGrandExchangeSearched(GrandExchangeSearched event)
...
+				Item[] foundItems = tabContainerToCalculate.toArray(new Item[tabContainerToCalculate.size()]);
+				final ContainerPrices prices = bankCalculation.calculate(foundItems);
+				if (prices != null)
+				{
+					priceText = createValueText(prices);
+				}
+			}
+
+			Widget bankTitle = client.getWidget(WidgetInfo.BANK_TITLE_BAR);
+			bankTitle.setText(bankTitle.getText() + "<br>" + priceText);
```
Nightfirecat, line 334:
```suggestion
			if (!priceText.empty())
			{
				Widget bankTitle = client.getWidget(WidgetInfo.BANK_TITLE_BAR);
				bankTitle.setText(bankTitle.getText() + "<br>" + priceText);
			}
```

JZomerlei:
When fixing the merge conflicts about how ContainerCalculations was moved to the bank plugin, I moved calculate back into its own utility function. BankPlugin &  BankTagsPlugin both have their own calculate method but both use ContainerCalculations' calculate method. 

I also reintroduced ContainerCalculationTest because the calculation is not being done within the BankPlugin anymore. (It wasn't done when in BankPlugin I first started this project but was moved into BankPlugin  in a recent update and now it's back to being done separately)

runelite-client/src/main/java/net/runelite/client/util/ContainerCalculation.java
```diff
@@ -0,0 +1,91 @@
...
+import net.runelite.api.Item;
+import net.runelite.api.ItemID;
+import net.runelite.client.game.ItemManager;
+
+public class ContainerCalculation
+{
+	@Getter(AccessLevel.PUBLIC)
+	private static HashSet<Integer> itemContainer = new HashSet<>();
+
+	private static ItemManager itemManager;
```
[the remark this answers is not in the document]
JZomerlei: When it isn't static the BankPlugin's value is different than the one found in BankTagsPlugin. I thought they ought to be combined.

(I assigned `chaos` to the feathers
image

runelite-client/src/main/java/net/runelite/client/util/ContainerCalculation.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+import net.runelite.api.Item;
+import net.runelite.api.ItemID;
+import net.runelite.client.game.ItemManager;
+
+public class ContainerCalculation
+{
+	@Getter(AccessLevel.PUBLIC)
+	private static HashSet<Integer> itemContainer = new HashSet<>();
```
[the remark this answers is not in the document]
JZomerlei: HashSet is now an ArrayList to prevent the duplicate issue you mentioned. I didn't think about that because I don't really have a bank with emblems or bonds etc. 

Also the ContainerCalculation that I removed was more of a MOVE than anything, if you compare the code what is on master to what I have in the class it was basically the same besides the additional shared itemContainer which is now removed.

I don't really know how to share the tag search values and regular search.  :/ 
I was hoping to find an example of a utility variable that is shared between plugins as an example

Adam-:
added in ead0b2819316481fadfab1d19b51d85732e4a895

PR #11772, opened 2020-05-30
Fix cryptic clue to show Iorwerth's correct location before and after…

### LINKED ISSUE #11674: Iorwerth cryptic clue uses different NPC after completion of Song of the Elves
Before completing the quest, [you must speak to Lord Iorwerth](https://oldschool.runescape.wiki/w/Lord_Iorwerth) in the Iorwerth camp southwest of Prifddinas. After completion of the quest, [you must speak to Lord Iestin Iorwerth](https://oldschool.runescape.wiki/w/Lord_Iestin_Iorwerth) who is in a building in the southwest quadrant of Prifddinas.

naleba97:
… Song of the Elves completion

Closes #11674

Originally, the cryptic clue only added Iorwerth's location before the
completion of Song of the Elves. The solution text however temporarily
both noted pre-quest and post-quest completion. With some changes in the
base `ClueScroll` class, the `CrypticClue` class has been expanded to add
multiple locations and contain alternate clues based on quest
completion.

These changes attempt to help better represent clues whose NPCs change based on certain quests (e.g. Steve before/after DS2, Iorwerth and Eluned before/after SotE) and clues whose final locations do not take place on the Runescape surface (e.g. Steve in the Asgarnian Ice Dungeon, Genie in the crevice west of Nardah, Eluned in Prifddinas). 

The base `ClueScroll` class has been added with a `Quest` array and a reference to an alternative `ClueScroll`. The `questConditions` `Quest` array contains quests that need to be completed in order for the clue scroll to be valid. For example, the cryptic clue for Lord Iorwerth now contains its post-quest location and quest requirements, and `alternateClue` references his original pre-quest location. This can be ported to anagram clues for the pre-quest and post-quest locations of Steve.

Before SotE completion:
image

After SotE completion:
ezgif com-optimize

In addition, RuneLite currently does not put intermediate `ClueScrollWorldMapPoint`s on the Runescape surface for clue scrolls whose final location is elsewhere. Without this, Lord Iorwerth's post-quest location (i.e. the clue scroll icon on the world map) will only appear in the real version of Prifddinas and not on the Runescape surface when viewing the world map. Cryptic clues now support multiple locations to help guide players if the final world point is not on the Runescape Surface (e.g. in the crevice west of Nardah).

Other changes: 
 - The command `::clue` now takes place on a delayed game tick to support on-demand quest progress lookup in `ClueScrollPlugin::findClueScroll()`.
 - Added test to check for clues with alternate clues and multiple locations.

However, these changes come with a few negative/unclear implications for future clue scroll additions.
 - The final location for Cryptic clue scrolls with intermediate locations is always at the end of the array.
 - The more restrictive post-quest clue should be at the head of the linked list with a reference to the less restrictive pre-quest clue.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -60,6 +60,8 @@ import net.runelite.api.MenuAction;
 import net.runelite.api.NPC;
 import net.runelite.api.ObjectComposition;
 import net.runelite.api.Point;
+import net.runelite.api.Quest;
+import net.runelite.api.QuestState;
 import net.runelite.api.Scene;
 import net.runelite.api.ScriptID;
 import net.runelite.api.Tile;
@@ -90,6 +92,7 @@ import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetID;
 import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.client.callback.ClientThread;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.ConfigChanged;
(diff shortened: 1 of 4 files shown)


Adam-:
I don't particularly like this approach since it is so closely tied specifically with quests, and we would like to sometimes change things conditionally on other things, such as levels (eg. suggest certain agility shortcuts). I would expect something which just overrides getLocation/getSolution etc to probably be better? As long as passed around Client and the plugin probably into them.

PR #11775, opened 2020-05-31
Update farming skill calculator with Sq'irk fruits and POH garden plants

### LINKED ISSUE #11262: Please add all Sq'irk Fruit and Bagged Plants to Farming Skill Calculator.
Although they are not traditional farming methods, they do provide farming XP.
Not sure if the farmers outfit effects them and I don't have the outfit for testing, but I'm assuming they most likely need to be exempt from the outfit bonus (I tried searching the wiki but it doesn't appear to have info confirming or denying this.)

Plondrein:
closes #11262


Plondrein:
Need to check about bonus XP

Plondrein:
Both sq'irks and garden plants XP is affected by farming outfit

PR #11776, opened 2020-05-31
Respect multiple displays when saving/loading client position

### LINKED ISSUE #4778: Client window doesn't remember size/position anymore.
Starting today 8/8/2018, the client for some reason resets settings on its own every time i launch it, it doesn't remember what size and position i set for it,was working just fine yesterday, now i have to resize it all over again and adjust my settings every time i open it, tried to reinstall but nothing changed.

Spedwards:
Closes #4778
Revival of #5763

All credits to @raiyni for this.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -513,11 +513,15 @@ public class ClientUI
 					{
 						frame.setBounds(clientBounds);
 
+						// frame.getGraphicsConfiguration().getBounds() returns the bounds for the primary display.
+						// We have to find the correct graphics configuration by using the intersection of the client boundaries.
 						GraphicsConfiguration gc = getIntersectingDisplay(clientBounds);
 						if (gc != null)
 						{
 							double scale = gc.getDefaultTransform().getScaleX();
 
+							// When Windows screen scaling is on, the position/bounds will be wrong when they are set.
+							// The bounds saved in shutdown are the full, non-scaled co-ordinates.
 							if (scale != 1)
 							{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -510,7 +512,28 @@ public void show()
...
 					{
 						frame.setBounds(clientBounds);
-						frame.revalidateMinimumSize();
+
+						GraphicsConfiguration gc = getIntersectingDisplay(clientBounds);
+						if (gc != null)
+						{
+							double scale = gc.getDefaultTransform().getScaleX();
+
+							if (scale != 1)
```
raiyni, line 521:
```suggestion
							// When windows screen scaling is on, the position/bounds will be wrong when they are set. The bounds saved in shutdown are the full, non-scaled coordinates. 
							if (scale != 1)
```

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -510,7 +512,28 @@ public void show()
 					if (clientBounds != null)
 					{
 						frame.setBounds(clientBounds);
-						frame.revalidateMinimumSize();
+
+						GraphicsConfiguration gc = getIntersectingDisplay(clientBounds);
```
raiyni, line 516:
```suggestion
						// frame.getGraphicsConfiguration().getBounds() returns the bounds for the main display. So we have to find the correct graphics configuration by using the intersection of the client boundaries.
						GraphicsConfiguration gc = getIntersectingDisplay(clientBounds);
```

PR #11778, opened 2020-05-31
Rename weapon type enum & add missing attack styles

Spedwards:
Supersedes/Closes #11626


runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStyle.java
```diff
@@ -32,10 +32,13 @@
...
 	DEFENSIVE("Defensive", Skill.DEFENCE),
 	CONTROLLED("Controlled", Skill.ATTACK, Skill.STRENGTH, Skill.DEFENCE),
+	ACCURATE_RANGING("Accurate Ranging", Skill.RANGED),
 	RANGING("Ranging", Skill.RANGED),
 	LONGRANGE("Longrange", Skill.RANGED, Skill.DEFENCE),
 	CASTING("Casting", Skill.MAGIC),
 	DEFENSIVE_CASTING("Defensive Casting", Skill.MAGIC, Skill.DEFENCE),
+	AIM_AND_FIRE("Aim and Fire"),
+	BLOCK("Blocking"),
 	OTHER("Other");
```
Adam-, line 42: What is this used for?

ThePharros: > What is this used for?

`AIM_AND_FIRE` is used by the Gun category, which is currently only used by Fixed Device

`BLOCK` is used by the Dinh's Bulwark

Adam-: This comment is on the `OTHER` line, which looks unused to me.

ThePharros: Ah mis-observed, my bad. `OTHER` is what was used prior for those two mentioned unique combat styles. I believe it is being retained just in case there is another style not covered by the enum, but perhaps Spedwards can elaborate better.

XrioBtw:
I propose the following additions and changes:
AttackStyle.java
- `ACCURATE_RANGING("Accurate Ranging", Skill.RANGED)` to
`ACCURATE_RANGING("Accurate", Skill.RANGED)`
- `RANGING("Ranging", Skill.RANGED)` to
`RAPID("Rapid", Skill.RANGED)` (or `RAPID("Rapid Ranging", Skill.RANGED)`)
- `ACCURATE_CASTING("Accurate Casting", Skill.MAGIC)`
- `LONGRANGE_CASTING("Longrange Casting", Skill.MAGIC, Skill.DEFENCE)`

WeaponType.java
- `RANGING` to
`RAPID`
- `STAFF_POWERED(CASTING, CASTING, null, DEFENSIVE_CASTING)` to
`STAFF_POWERED(ACCURATE_CASTING, ACCURATE_CASTING, null, LONGRANGE_CASTING)`

If we end up strictly following RuneScape terminology and are worried about confusion for `Accurate` when using a ranged weapon and a melee weapon then I propose adding the appropriate sprites for the skills that you will gain experience in into the infobox.

honeyhoney:
> If we end up strictly following RuneScape terminology and are worried about confusion for `Accurate` when using a ranged weapon and a melee weapon then I propose adding the appropriate sprites for the skills that you will gain experience in into the infobox.

When the plugin was first created the descriptors of the attack styles were primarily to indicate what experience you might gain if you were to attack. A large reason this plugin exists is for pures or other "restricted" accounts to monitor and prevent gaining experience in unwanted styles.

I agree differentiating between Accurate and Rapid for Ranging would be nice, I think the name Defensive Casting serves a purpose, although I guess Longrange Casting could be a useful descriptor depending on why the person is using the plugin.

Accurate Casting serves no purpose. You're either "Accurate" i.e. Casting or "Longrange" i.e. Defensive/Longrange Casting. There is no Rapid Casting as far as I know.

There was even a discussion as to whether the enum should even be called WeaponType back in the day 😂 Naming is hard.

XrioBtw:
> Accurate Casting serves no purpose. You're either "Accurate" i.e. Casting or "Longrange" i.e. Defensive/Longrange Casting. There is no Rapid Casting as far as I know.

I disagree. According to Bitterkoekje ([source](https://secure.runescape.com/m=forum/forums?317,318,461,66138854)) [powered staves](https://oldschool.runescape.wiki/w/Weapons/Types#Powered_staves) have a stance bonus that [normal staves](https://oldschool.runescape.wiki/w/Weapons/Types#Staves) do not have:

> Magic attack:
• Accurate +3 (trident only)
• Long range +1 (trident only)

This means that there is a difference between "Casting" and "Accurate Casting", though it is not visible to the player unless you do the accuracy calculations. As for "Rapid Casting" this could be used for the [Harmonised nightmare staff](https://oldschool.runescape.wiki/w/Harmonised_nightmare_staff) which can attack with 4 instead of 5 tick speed, but this is not straight forward as it only applies to standard spellbook spells.

I my opinion there is no conflict in having this plugin both correctly show weapon types / attack styles and serving as an indicator for what experience you will gain. If the latter is harder to understand with the more precise naming I again refer to my previous suggestion to add skill sprites.

honeyhoney:
> I disagree. According to Bitterkoekje ([source](https://secure.runescape.com/m=forum/forums?317,318,461,66138854)) [powered staves](https://oldschool.runescape.wiki/w/Weapons/Types#Powered_staves) have a stance bonus that [normal staves](https://oldschool.runescape.wiki/w/Weapons/Types#Staves) do not have:

I don't think this relates to my point that Rapid Casting does not exist as an Attack Style. The stance bonus is from selecting either Accurate or Longrange, not from changing from the first "Accurate" style to the second "Accurate" style. i.e. both these "Accurate" options on a powered staff are identical.

image

The plugin already distinguishes between the two stances available to powered staves by referring to the first two "Accurate" options as "Casting" and the "Longrange" option as "Defensive Casting".

> In addition, I believe powered staves always attack with 4 tick speed VS 5 tick speed for standard "Casting".

It is correct that powered staves have a 4 tick attack speed whereas autocasting spells from a staff is typically 5 tick. Are you suggesting we should look to indicate attack speed on weapons? An abyssal whip and dharok's greataxe both have what we refer to as an "Accurate" (i.e. attack experience granting) attack style but have an attack speed difference of 3 ticks that we don't currently indicate.

> This means that there is a difference between "Casting" and "Accurate Casting", though it is not visible to the player unless you do the accuracy calculations.

I'm not sure I follow this at all. You're either Casting (indicated by "Casting" - whether you want to call this something else is up for debate I guess), or Longrange Casting (indicated by "Defensive Casting"/"Longrange Casting"), or not Casting at all (indicated by whatever Attack Style you're using instead). 

> As for "Rapid Casting" this could be used for the [Harmonised nightmare staff](https://oldschool.runescape.wiki/w/Harmonised_nightmare_staff) which does not have the invisible stance bonus of the powered staves and can attack with 4 instead of 5 tick speed, but this is not straight forward as it only applies to standard spellbook spells.

Perhaps there's an argument for calling the regular autocast style of the Harmonised staff "Rapid Casting" but it's not like the player can actually change the attack speed of the staff like they can on ranged weapons. I don't think it's a particularly useful distinction to make outside of exposing players to game knowledge they made not already have - "Hey this weapon autocasts regular spells 1 tick faster!".

XrioBtw:
Let's forget about the "Rapid Casting" as I think we both don't want that, am I right?

> The stance bonus is from selecting either Accurate or Longrange, not from changing from the first "Accurate" style to the second "Accurate" style. i.e. both these "Accurate" options on a powered staff are identical.

I am aware of this. I am only concerned with the following distinction:
powered_staff_VS_staff
The skill(s) that gains experience is the same on both sides, the only difference is the stance bonus. If we don't rename "Casting" to "Accurate Casting" and "Defensive Casting" to "Longrange Casting" for the case of powered staves only we have no way to distinguish what is shown on the left and right in the above image, which is useful for a plugin that both serves to indicate/warn you what skill you will gain experience in AND imply if and what kind of stance bonus is currently active.

I do not want to rename or remove CASTING, I want to add additional versions of CASTING that are called ACCURATE_CASTING and LONGRANGE_CASTING because when you account for stance bonus there are 4 types of spellcasting: CASTING, ACCURATE_CASTING, LONGRANGE_CASTING and DEFENSIVE_CASTING.

PR #11785, opened 2020-06-01
widitemoverlay: clip item text around incinerator box when box isvisible

### LINKED ISSUE #11728: Item identification plugin tags are being drawn over the incinerator window
**Describe the bug**
Names/tags from the Item identification plugin are being drawn over the bank's incinerator window.

**To reproduce**
1. Enable Item Identification plugin.
2. Open your bank.
3. Enable incinerator in the bank settings.
4. Scroll to a place in the bank where you have weeds/seeds stored.
5. Put any item in the incinerator.

RuneliteBug2

**Expected behavior**
(Easier solution) The names shouldn't be drawn when the incinerator window is open.
(Harder solution) The names should be drawn under the incinerator window.

**Environment**
RuneLite 1.6.16

johannfrias:
Closes #11728. When incinerator box is visible, draw a Polygon around the box and clip the text to fit in that space. Apparently `setClip` is expensive, so the efficiency of calling that should be double checked. (Right now I seem to enter the incinerator logic just once per `render()`.)

Also, is there a reason the `shouldClip` calculation isn't simplified by using Rectangle's `contains()` and `intersects()`? (Tried a possible rewrite (not here), found no apparent problems)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/WidgetItemOverlay.java
@@ -26,13 +26,16 @@ package net.runelite.client.ui.overlay;
 
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.awt.Polygon;
 import java.awt.Rectangle;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Setter;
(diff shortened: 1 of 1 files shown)


Adam-:
This seems fairly hacky, maybe just don't draw the bank widget item overlays when the incinerator confirmation dialog is open?

johannfrias:
Well, I gave issue OP's "harder solution" a shot. I don't see a simpler way to redefine the clip area, and I don't see how to render the overlays before the incinerator confirm. So, not drawing the overlays, which is also issue OP's "easier solution," seems much easier.

PR #9572, opened 2019-08-04
Add Config Sections

Hydrox6:
Closes #11602

Certain configs have always been begging for some form of section system. When merging multiple plugins together, being able to separate their options would be very useful.

Sections can be toggled by clicking anywhere on them, and are persisted through the config. It is possible to nest configs inside one another. Each level of nesting adds a small but noticeable margin on the left side.

I have added sections to Chat Color (which desperately needed it) to show users that there are 2 sets of options, and to Ground Items to hide potentially gigantic item lists.

Feedback on the design of the sections is very much appreciated

(Above gif is outdated, as the text is now RL orange. See comments below)

PR #11788, opened 2020-06-01
Add dedicated toggles for NPC highlighting styles

haakonrp:
title

Useful when you want to know both sw tile and whole tile of an npc.

Example:
image


haakonrp:
Changed to individual toggles.

Updated example:
https://i.gyazo.com/1bbbd94997cf6d7bd39d3fd1f08c0337.gif

Nightfirecat:
Now that we've got config sections available (see #9572), these could be put into a section at the top of the list which is expanded by default

haakonrp:
Added a config section for it 👍

PR #11801, opened 2020-06-03
emoji: add flag emojis

mcnabbian:
Added 7 flag emojis (US, UK, Germany, Australia, white flag, black flag, and a rainbow flag).


dekvall:
You have left two `.DS_Store` in there and also i'm not sure what the consensus is on people going around typing in hashtags. Personally, i am fine with it.

makingcomment:
My opinion, for what it's worth (not much, I know), is that hashtags suck. Beyond that, I'm also not a fan of `#Rbw` and `#Wht` (and `#Blk` to a lesser extent) because I don't think it's clear what they mean for those not using the emoji plugin. However, the recently added gorilla emoji is also not clearly a gorilla for non-plugin-users, so based on precedent maybe this is not a valid concern.

mcnabbian:
@makingcomment why are hashtags bad? Is there another trigger that would make more sense? As for the last three flags, they can be expanded to full words (i.e. #rainbow)

Broooklyn:
It has been requested in the past that emoji triggers look somewhat like the object they are referencing, to ensure those without the plugin enabled can still tell what is being written.

PR #11810, opened 2020-06-04
grandexchange: remove dupe buy limit reset timers

### LINKED ISSUE #11808: buy limit reset timers duplicate themselves
**Describe the bug**
The buy limit reset timers duplicate every time you click any button in the grand exchange interface, guide price, enter quantity, and enter price buttons duplicate it once, any other button duplicates it twice.

It's also duplicated once upon opening the sell interface via clicking the item in your inventory instead of clicking the item after opening the sell interface.

This seems to only happen with items missing buy limits, items with known limits such as vial of water or brass key I cannot reproduce this with. Here's a list on the wiki. https://oldschool.runescape.wiki/w/Category:Exchange_items_missing_limit

**To Reproduce**
1. Buy any item that doesn't have a buy limit. Some examples are Elven top, Blighted karambwan, Blighted manta ray, Dragonstone boots, etc.
2. Either open the same item's buy interface or that item's sell interface.
3. Click buttons.
4. Observe dupe bugs.

**Expected behaviour**
Should only set a timer when you actually buy an item.

**Screenshots**
(gh doesn't support mp4 so youtube link it is)
https://youtu.be/FX1N0YjrtgQ

 - windows 10
 - RuneLite version: 1.6.18
 - Launcher version: 2.0.2

mcnabbian:
Closes #11808

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
@@ -107,8 +107,6 @@ import net.runelite.http.api.osbuddy.OSBGrandExchangeResult;
 import net.runelite.http.api.worlds.WorldType;
 import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.apache.commons.text.similarity.FuzzyScore;
-import sun.lwawt.macosx.CSystemTray;
-import sun.tools.jconsole.JConsole;
 
 @PluginDescriptor(
 	name = "Grand Exchange",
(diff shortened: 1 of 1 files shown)


Adam-:
It's not obvious to me; why does this fix the bug? What is the root cause?

mcnabbian:
@Adam- I just revisited the issue and I think I misunderstood what the expected behavior would be. This solution forgoes any reset timer for items without a limit. The issue lies in executing `getLimitResetTime(itemId)` on an item without a limit, which later results in `if (resetTime != null)` evaluating to true when it should not, causing the unexpected behavior

Adam-:
fixed in 3d762d7ca2f5cad30009f1d48dca5093155a8fc8

PR #11815, opened 2020-06-05
Add alternative highlight option for dead tagged NPCs and dead NPC menu entries.

jesse1412:
Add an optional colour for dead NPCs to the NPC indicators plugin.

Additionally add an option to make that highlighting apply to all dead NPCs in menu entries, even those not tagged.

Useful for barraging/chinning stacks of NPCs and for identifying important dead NPCs at a glance.

[Video example](https://streamable.com/l6qcmp)


Adam-:
added in bda66e8da47309b8117175f46feaf8b126c061df

PR #11835, opened 2020-06-06
Fix the 24+ player team size bug for cox pb and allow for pb times from raid end message

melkypie:
image

This will fix the bug where if you get a pb with 24+ players as the team size it did not count is a new pb.
Also adds the ability for cox pb tracking to check if a pb mentioned in the raid end message is better than the one saved in config.

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/chatcommands/ChatCommandsPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/chatcommands/ChatCommandsPluginTest.java
@@ -400,7 +400,7 @@ public class ChatCommandsPluginTest
 	@Test
 	public void testCoXKill()
 	{
-		ChatMessage chatMessage = new ChatMessage(null, FRIENDSCHATNOTIFICATION, "", "<col=ef20ff>Congratulations - your raid is complete!</col><br>Team size: <col=ff0000>4 players</col> Duration:</col> <col=ff0000>37:04</col> (new personal best)</col>>", null, 0);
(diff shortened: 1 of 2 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/chatcommands/ChatCommandsPluginTest.java
```diff
@@ -459,7 +459,7 @@ public void testAdventureLogCountersPage()
 			"<br><br>Hespori<br>Fastest kill: <col=d0c0b0>0:57</col><br><br>Nightmare<br>" +
 			"Fastest kill: <col=d0c0b0>3:30</col><br><br>The Gauntlet<br>Fastest run: <col=d0c0b0>-</col>" +
 			"<br><br>The Corrupted Gauntlet<br>Fastest run: <col=d0c0b0>-</col><br><br>Fragment of Seren<br>Fastest kill: <col=d0c0b0>-</col>" +
-			"<br><br>Chambers of Xeric<br>Fastest run - (Team size: 4 players): <col=d0c0b0>24:17</col>" +
+			"<br><br>Chambers of Xeric<br>Fastest run - (Team size: 24+ players): <col=d0c0b0>24:17</col>" +
```
Nightfirecat, line 477: Has the POH adventurer's log widget text changed in this way?

Adam-:
`matchPb()` doesn't actually check if the new pb is lower - and the game is now tracking per-team-size pbs right?

melkypie:
`matchPb()` itself doesn't check if the pb is lower, it only sets it inside a variable to wait for the kc message and then it is checked if the pb is actually lower.
Yes it is per team size pb's now and the pb in the raid end message would be for that specific team size pb. Since we are currently only tracking the best pb not individual team size pb's, it checks if that is better. 
Also doing per team size pbs the way it was done in sepulchre would not be feasible as you would need a 25 case switch statement for each team size.
As for the adventure log, I can't give a screenshot of it since I don't do 24+ player raids but I assume it will be the same way as Solo is.

PR #11849, opened 2020-06-07
updated LootTracker to track Coffins from Sepulchre

whaevr:
My account isn't able to get to the lowest Floor so I wasn't able to add the specific event Id for the Grand Coffin. Adding the event Id for that and using it in an Or in the onMenuOptionClicked hook check I added would work, unless that Coffin should be tracked separately? 

These chests give out a good amount of items, its nice to track all the supplies I've gotten from them and see that total value


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -206,6 +211,7 @@
 	@VisibleForTesting
 	LootRecordType lootRecordType;
 	private boolean chestLooted;
+	private boolean coffinLooted;
```
Hydrox6, line 214: This var is never reset anywhere

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -143,6 +143,11 @@
 	// Chest loot handling
 	private static final String CHEST_LOOTED_MESSAGE = "You find some treasure in the chest!";
 	private static final Pattern LARRAN_LOOTED_PATTERN = Pattern.compile("You have opened Larran's (big|small) chest .*");
+
+	private static final String COFFIN_LOOTED_MESSAGE = "You push the coffin lid aside.";
+	private static final int HALLOWED_SEPULCHRE_COFFIN_REG = 39535;
```
Hydrox6, line 148: This shouldn't be hardcoded, thought I'm also not sure if it should be referencing the NullObjectID list either. Not sure why the coffin is a null object anyway.

whaevr: Interesting, Im not aware the difference between Object/NullObject. I was wondering why I couldn't find the coffin's # in the ObjectID class though. So that explains it

I tried to use RegionIds like the plugin is currently using for Chest events like Larrans, Crystal, Brimstone, etc. The Sepulchre must be shared instanced or something? though because they would change every run. Not sure how to go about handling this, this is first time I've ever done anything with RL code. 

I will properly at least reference it from NullObjectID in the meanwhile?

Bram91:
In its current state this will also miss certain drops, if you loot the chest with a full inventory the items drop to the floor and wont be part of the inventory snapshot.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
whaevr: fixed

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
whaevr: So I looked into how Kraken was operating and using DevOptions grabbed a Map region # from each floor. Using the Location Plugin in dev plugins. Did 3 runs and took a new screenshot eachtime per floor to make sure the Map region number was reliable. 

I can only reach the 3rd floor though, someone will have to find numbers for the 4th and fifth

Hydrox6:
Also, as was discussed on the Discord, the loot you get is broadcasted in the chat. Using this, we can get a 100% accurate readout of the loot even if it went to the floor. It could be done similar to the Herbiboar tracking.

whaevr:
All set, wrote test coverage as well for all the types of messages discussed

PR #11851, opened 2020-06-08
allow shop menu buying and selling swaps to be configurable per item

thomaschow:
### Description

https://github.com/runelite/runelite/pull/11453 introduced a feature to the [Menu Entry Swapper ](https://github.com/runelite/runelite/wiki/Menu-Entry-Swapper)plugin to allow users to globally configure option swaps when buying and selling to in game shops. 

There was a mention https://github.com/runelite/runelite/pull/11453#issuecomment-624915040 and https://github.com/runelite/runelite/pull/11453#issuecomment-626033179  about individual buy/sell quantities. I thought this would be useful so I took a stab to implement. Open for further discussion on any aspect of this PR. 

I reviewed https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features#menu-entry-swapping and [Jagex's 3rd party client statement ](https://secure.runescape.com/m=news/a=13/another-message-about-unofficial-clients?oldschool=1) and the way I understood it, it looks like there should be no offense with this particular feature. 

#### Implementation

- Adds text boxes for each of the buy/sell quantities where the user can input a csv for items they'd want swaps. 

- If a user indicates the same item for multiple quantities, it will use the selection in order of least to greatest precedence: global `buy` option, whitelist `Off`, whitelist `buy 1`, whitelist `buy 5`, whitelist `buy 10`, whitelist `buy 50` and similar for `sell`.  

- A user can simulate a blacklist by setting a global value other than `Off`, then adding an item to whitelist `Off`. 

- Attached a GIF of it in action:
ezgif com-video-to-gif (2)

### Testing

- [x] local testing
- [x] unit tests


dekvall:
Having lists for just these feels very niche, cluttered and slightly confusing in regards to what's prioritized over what. I'd rather keep the current shift-click option since imo that's enough to prevent you from accidently selling your stack.

Adam-:
Maybe once the config group stuff is merged (#9572) we could do this and just collapse it by default? Currently I don't like that this takes up basically all of the side panel for a feature almost noone will use.

I also don't see any reason why this has to have its own separate config class like this, its very odd and not something we do elsewhere.

thomaschow:
> I'd rather keep the current shift-click option since imo that's enough to prevent you from accidently selling your stack.

This still operates via shift click, the point is that you can vary the behavior per type of item. 


> Maybe once the config group stuff is merged (#9572) we could do this and just collapse it by default? Currently I don't like that this takes up basically all of the side panel for a feature almost noone will use.

Great, I was thinking the same but I didn't realize something was already in flight. Happy to wait till that gets merged, and I'll update the PR.


> I also don't see any reason why this has to have its own separate config class like this, its very odd and not something we do elsewhere.

I just figured these are logically a grouped set of configs and should not interact with any other part of the plugin. I'm fine moving it to the main config.

Adam-:
If you want you could add code folding for the options, though we don't typically do that.

Hydrox6:
tbh, this should be done like the current Shift Click configuration for the normal inventory, though I'm not sure where you would trigger the edit mode from.

XrioBtw:
@Hydrox6 Could this be a possible location for the shift-click configuration (the shop exit button)?
General_Store_shift-click_crop

thomaschow:
> Could this be a possible location for the shift-click configuration (the shop exit button)?

That makes sense to me, if people are in agreement with that I can change the approach.

thomaschow:
@XrioBtw  @Nightfirecat  @Hydrox6 I've updated the PR with the implementation asked for in https://github.com/runelite/runelite/pull/11851#issuecomment-640456073 and https://github.com/runelite/runelite/pull/11851#issuecomment-640679542. 

The only caveat here is that the customizable shift click doesn't take precedence over the in game "shift click drop" option....

thomaschow:
@Nightfirecat @Hydrox6  would you have thoughts on whether we could merge this?

Nightfirecat:
This approach looks fine to me. This PR needs rebasing since the menu entry swapper plugin refactoring.

raiyni:

PR #11857, opened 2020-06-08
timers: fix home teleport timer plugin from Underwater

### LINKED ISSUE #11838: Home and Minigame teleport timers don't appear when teleporting from underwater
Because these timers check for the animation (not the graphic) of your player, if you are floating underwater and trigger a teleport, your animation does not change. We can change this to check your player's graphics change events instead of animations.

mcnabbian:
Closes #11838 

I basically copy and pasted code from `onAnimationChanged` to `onGraphicChanged` to include a check for `config.showHomeMinigameTeleports()`. I also added `GraphicID.BOOK_HOME_TELEPORT_4` attribute to `HOME_TELEPORT` and `MINIGAME_TELEPORT` so I could use `.getGraphicID()`. 

`actor.getGraphic() == HOME_TELEPORT.getGraphicId() || actor.getGraphic() == MINIGAME_TELEPORT.getGraphicId()` returns the same int in either case so this could be shortened to one of those two. I kept both in just to be clear/explicit but lmk if those should be removed. I'm still a nub at this so lmk if there's a more optimal solution that doesn't require copy and pasting code to a different function and I'll try to make that change. 

Also I got 100 kudos and completed Bone Voyage just to test this solution lol


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -755,6 +755,20 @@ public void onGraphicChanged(GraphicChanged event)
 			createGameTimer(IMBUEDHEART);
 		}
 
+		if (config.showHomeMinigameTeleports()
+			&& (actor.getGraphic() == HOME_TELEPORT.getGraphicId()
+			|| actor.getGraphic() == MINIGAME_TELEPORT.getGraphicId()))
```
Nightfirecat, line 760: This would create the teleport timer too early--when the final graphic starts, not when it ends. This should be changed to behave more like the animation check where it checks for final animation -> idle animation.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -67,8 +67,8 @@
 	CHARGE(SpriteID.SPELL_CHARGE, GameTimerImageType.SPRITE, "Charge", 6, ChronoUnit.MINUTES),
 	STAFF_OF_THE_DEAD(ItemID.STAFF_OF_THE_DEAD, GameTimerImageType.ITEM, "Staff of the Dead", 1, ChronoUnit.MINUTES),
 	ABYSSAL_SIRE_STUN(ItemID.ABYSSAL_ORPHAN, GameTimerImageType.ITEM, "Abyssal Sire Stun", 30, ChronoUnit.SECONDS, true),
-	HOME_TELEPORT(SpriteID.SPELL_LUMBRIDGE_HOME_TELEPORT, GameTimerImageType.SPRITE, "Home Teleport", 30, ChronoUnit.MINUTES),
-	MINIGAME_TELEPORT(SpriteID.TAB_QUESTS_RED_MINIGAMES, GameTimerImageType.SPRITE, "Minigame Teleport", 20, ChronoUnit.MINUTES),
+	HOME_TELEPORT(SpriteID.SPELL_LUMBRIDGE_HOME_TELEPORT, GameTimerImageType.SPRITE, "Home Teleport", GraphicID.BOOK_HOME_TELEPORT_4, 30, ChronoUnit.MINUTES, false),
```
Nightfirecat, line 70: I haven't checked with dev tools lately but I'm pretty sure the twisted themed book teleport and cow teleport both show different graphics than the standard home teleport.

mcnabbian: I tested the cow teleport and you're right, it does. I bet the twisted themed does as well, but I'm unable to test for it and figure out its graphic id right now. Is there another way to get the graphic id? I can't find it in the api.

mcnabbian:
How do I undo this mess :\

PR #11859, opened 2020-06-09
LootTracker support for shade chests; support ground items

### LINKED ISSUE #8817: Add loot from chests in Shades of Mort'ton minigame to loot tracker
I think this would be consistent with the tracking of Brimestone, clue scroll, and barrows chests.

leejt:
This commit makes it so LootTracker tracks items that fall on the floor during an inventory-based loot. This works because ItemSpawned/ItemQuantityChanged events happen before ItemContainerChanged events.

It keeps a running track of items that spawned directly below the player during the current tick, and if that is in the same tick as as an inventory change that correlates to (e.g.) a chest or seed pack event, we include the ground items as part of the loot. I've tested this fairly extensively for crystal chests, with good results.

The shades part is relatively straightforward, and uses the object IDs. The only tricky thing is that unlike seed packs, just clicking "Open" is not sufficient – you need to also have a key in your inventory. The game gives you a message if you fail to have a key, so we match against this. Fixes #8817

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -96,6 +96,7 @@ import net.runelite.client.events.SessionClose;
 import net.runelite.client.events.SessionOpen;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.game.ItemStack;
+import net.runelite.client.game.LootManager;
 import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -619,6 +649,13 @@ public void onChatMessage(ChatMessage event)
 					break;
```
Nightfirecat, line 649: All the clue scroll event cases should have their `break`s become `return`s.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -657,6 +698,13 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 			lootRecordType = LootRecordType.EVENT;
 			takeInventorySnapshot();
 		}
+
+		if (event.getMenuOption().equals("Open") && SHADE_CHEST_OBJECTS.containsKey(event.getId()))
+		{
+			eventType = SHADE_CHEST_OBJECTS.get(event.getId());
```
Nightfirecat, line 704: As written, this would make every chest of different key color recorded in a different loot group, which is kind of weird. I'd expect users to care only about the chest color, not about the key color.

PR #11862, opened 2020-06-09
Chest-a-palooza: track grubby, stone, Dorgesh-Kaan and HAM chests in …

leejt:
…loot tracker

The Dorgesh-Kaan tracking could be done better (there are actually 4 distinct drop tables) but it's hard to do this without significantly complicating the code.


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -629,10 +650,16 @@ public void onItemContainerChanged(ItemContainerChanged event)
 			return;
 		}
 
+		if (eventType == null)
```
Nightfirecat, line 653: This should be merged with the guard clause above to become:

```java
if (event.getContainerId() != InventoryID.INVENTORY.getId()
	|| eventType == null)
{
	return;
}
```

Nightfirecat:
I can verify all the region IDs here are correct.

Issue #11863, opened 2020-06-09
Slayer plugin does not update when querying the task through Activating a Slayer Gem


Nyanaro:
The Plugin shows a different task with a different number of monsters left. I have tried resetting the plugin with an enchanted gem, but that has proved futile.

It started after I completed a task on mobile, and once I returned to the runelite client the plugin wasn't working. I've uploaded a video of the occurrence here https://streamable.com/10x2z6

I am on Windows 7, Runelite version 1.6.18.1, Launcher version 2.1.0

Hydrox6:
Check your gem via the right click menu to fix this right now. It should probably check for the dialogue text as well though.

PR #11866, opened 2020-06-10
Slayer plugin: add activate slayer gem functionality and fix existing…

PresNL:
… task for konar and wilderness tasks

I updated the regex for when you talk to a slayer master when already having a task to also work for Konar and Wilderness tasks and I also updated it to work for when you activate the slayer gem and I have added tests for these scenarios.

Issue:
closes https://github.com/runelite/runelite/issues/11863

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -111,7 +111,7 @@ public class SlayerPlugin extends Plugin
 	private static final Pattern NPC_ASSIGN_MESSAGE = Pattern.compile(".*(?:Your new task is to kill|You are to bring balance to)\\s*(?<amount>\\d+) (?<name>.+?)(?: (?:in|on|south of) (?:the )?(?<location>.+))?\\.");
 	private static final Pattern NPC_ASSIGN_BOSS_MESSAGE = Pattern.compile("^(?:Excellent\\. )?You're now assigned to (?:kill|bring balance to) (?:the )?(.*) (\\d+) times.*Your reward point tally is (.*)\\.$");
 	private static final Pattern NPC_ASSIGN_FIRST_MESSAGE = Pattern.compile("^We'll start you off (?:hunting|bringing balance to) (.*), you'll need to kill (\\d*) of them\\.$");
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -111,7 +111,7 @@
 	private static final Pattern NPC_ASSIGN_MESSAGE = Pattern.compile(".*(?:Your new task is to kill|You are to bring balance to)\\s*(?<amount>\\d+) (?<name>.+?)(?: (?:in|on|south of) (?:the )?(?<location>.+))?\\.");
 	private static final Pattern NPC_ASSIGN_BOSS_MESSAGE = Pattern.compile("^(?:Excellent\\. )?You're now assigned to (?:kill|bring balance to) (?:the )?(.*) (\\d+) times.*Your reward point tally is (.*)\\.$");
 	private static final Pattern NPC_ASSIGN_FIRST_MESSAGE = Pattern.compile("^We'll start you off (?:hunting|bringing balance to) (.*), you'll need to kill (\\d*) of them\\.$");
-	private static final Pattern NPC_CURRENT_MESSAGE = Pattern.compile("^You're still (?:hunting|bringing balance to) (?<name>.+)(?: (?:in|on|south of) (?:the )?(?<location>.+), with|; you have) (?<amount>\\d+) to go\\..*");
+	private static final Pattern NPC_CURRENT_MESSAGE = Pattern.compile("^You're (still|currently assigned to) (?:meant to be )?(?:hunting|bringing balance to|kill|bring balance to|slaying) (?<name>.+?)(?: (?:in|on|south of) (?:the )?(?<location>.+))?(, with|; (you have|only)) (?<amount>\\d+)(?: more)? to go\\..*");
```
Nightfirecat, line 114: Use non-capturing groups for the added groups. Also `still meant to be` is always a contiguous string, `currently assigned to meant to be` will never occur, so the optional capturing of `meant to be` should be moved into the case where `still` is matched.
```suggestion
	private static final Pattern NPC_CURRENT_MESSAGE = Pattern.compile("^You're (?:still(?: meant to be)?|currently assigned to) (?:hunting|bringing balance to|kill|bring balance to|slaying) (?<name>.+?)(?: (?:in|on|south of) (?:the )?(?<location>.+))?(?:, with|; (?:you have|only)) (?<amount>\\d+)(?: more)? to go\\..*");
```

PR #11880, opened 2020-06-12
Add Stamina Timer for when wearing Ring of Endurance

dabolink:
adds stamina timer of double length if wearing Ring of Endurance
halves timer if ring is removed


Adam-:
What about https://github.com/Adam-/runelite/commit/afb8d41573e2b1da42b139b6c49ab7f6b28ce7e6? It avoids having to juggle two different timer types. Not sure if there is anyplace else we could reuse `setDuration` at.

Adam-:
added in 51f0b7752dac968d19dddd37081e62c9e5b5d546

PR #11894, opened 2020-06-14
2005 Interface Chatbox Fix

Brochir:
Currently the 2005 Interface chatbox has **neither** the proper "torn page" effect nor is it even the side and top lines of text get cut off frequently.

Please make it **even** like this while having the "unclean" effect. 

I understand we cannot do the torn effect properly but making it like this it still has the 2005 aesthetic while enabling us to be able to read all text in game. 

EabgYskX0AAG0_k


Hydrox6:
This is not how you make a feature request

PR #11897, opened 2020-06-14
chat commands: add hallowed sepulchre

Adam-:


makingcomment:
Does the message for opening the Grand Hallowed Coffin end in "times" even the first time you open it, or should it be "time" in that case?

Adam-:
> 
> 
> Does the message for opening the Grand Hallowed Coffin end in "times" even the first time you open it, or should it be "time" in that case?
> 
> EDIT: It's "time" in that case. From Woox's stream VOD:
> 1time

Good catch.

PR #11900, opened 2020-06-14
ChatFilter: Add config section for Filter Lists

Broooklyn:
The chat filter config can get pretty unruly if you have a lot of word and regex filters. This adds a config section akin to the `Item Lists` section in `Ground Items`.  `Filter type` config is shifted down to be with the rest of the configs.

Screen Shot 2020-06-14 at 6 26 23 PM

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
@@ -125,7 +125,7 @@ public interface ChatFilterConfig extends Config
 		keyName = "collapseGameChat",
 		name = "Collapse Game Chat",
 		description = "Collapse duplicate game chat messages into a single line",
-		position = 8
+		position = 9
 	)
 	default boolean collapseGameChat()
 	{
@@ -136,7 +136,7 @@ public interface ChatFilterConfig extends Config
 		keyName = "collapsePlayerChat",
 		name = "Collapse Player Chat",
 		description = "Collapse duplicate player chat messages into a single line",
-		position = 9
+		position = 10
 	)
 	default boolean collapsePlayerChat()
 	{
@@ -147,7 +147,7 @@ public interface ChatFilterConfig extends Config
 		keyName = "maxRepeatedPublicChats",
 		name = "Max repeated public chats",
 		description = "Block player chat message if repeated this many times. 0 is off",
-		position = 10
+		position = 11
 	)
 	default int maxRepeatedPublicChats()
 	{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -113,7 +125,7 @@ default boolean filterLogin()
 		keyName = "collapseGameChat",
 		name = "Collapse Game Chat",
 		description = "Collapse duplicate game chat messages into a single line",
-		position = 9
+		position = 8
```
Nightfirecat, line 128: Why are these positions changed?

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
```diff
@@ -28,26 +28,25 @@
...
 {
-	@ConfigItem(
-		keyName = "filterType",
-		name = "Filter type",
-		description = "Configures how the messages are filtered",
-		position = 1
+	@ConfigSection(
+		name = "Filter Lists",
+		description = "Custom Word, Regex, and Username filter lists",
+		position = 0,
```
Nightfirecat, line 39: Just make this position 1 like "Filter type" was before

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
Broooklyn: iirc my intent was to have the positions reflect their actual order, making it as clear as possible when viewing `ChatFilterConfig`. Previously it went from 7 to 9, so I corrected it.

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
Broooklyn: I was just following[ the precedent](https://github.com/runelite/runelite/commit/40f17648b4e83d1659d3fde5b7fb1d3ba2218a04#diff-637c5ee9b4bc1dae6fb2a818d2b34443) set by Hydrox, but I'm happy to change it if you feel I should.

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
Nightfirecat: There's no need for positions to be contiguous. Please revert the unneeded position changes to this file.

PR #11903, opened 2020-06-15
agility: Add config for Sepulchre obstacles and skill obstacles

Broooklyn:
This commit moves the Sepulchre obstacles from `COURSE_OBSTACLE_IDS` into its own set so it can be more easily toggled. Also adds config to optionally enable highlighting of the skill obstacles, which also now has its own set. `Sepulchre Highlight` has been renamed `Projectile Color` as to not be confusing with the newly added configs as they get their color from `Overlay Color`.  The other sepulchre configs have been renamed to prevent redundancy and were added to a new `sepulchreSection` at the bottom of `AgilityConfig`.

Screen Shot 2020-06-24 at 6 41 38 PM

Skill challenge highlights:
Screen Shot 2020-06-15 at 12 24 21 PM
Screen Shot 2020-06-15 at 12 20 57 PM
Screen Shot 2020-06-15 at 12 13 32 PM
Screen Shot 2020-06-15 at 12 35 49 PM


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
```diff
@@ -234,12 +234,34 @@ default boolean highlightSepulchreNpcs()
 
 	@ConfigItem(
 		keyName = "sepulchreHighlightColor",
-		name = "Sepulchre Highlight",
+		name = "Projectile Color",
```
Nightfirecat, line 237: Maybe add a "Hallowed Sepulchre" config section since "Projectile Color" on its own doesn't really mean anything in this plugin, and having four sepulchre configs is starting to get a bit busy at the bottom of this config.

Broooklyn: By adding the config section, it makes "Sepulchre" redundant in the config names, so those can probably be removed. Thoughts? 
Screen Shot 2020-06-21 at 12 43 42 PM
Screen Shot 2020-06-21 at 12 50 48 PM

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
```diff
@@ -28,11 +28,19 @@
...
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
 import net.runelite.client.config.Units;
 
 @ConfigGroup("agility")
 public interface AgilityConfig extends Config
 {
+	@ConfigSection(
+		name = "Hallowed Sepulchre",
```
Nightfirecat, line 38: You already have the section name defined in a variable, use that here.
```suggestion
		name = seepulchreSection,
```

Broooklyn: That'd be an illegal forward reference

Broooklyn:
Rebased with since-merged changes from #11799. Should be good to go now.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
Adam-: I think that you have renamed this in the wrong commit.

Broooklyn: @Adam- Should hopefully be fixed now, thanks for pointing this out.

PR #11916, opened 2020-06-18
Change thin progress bar to have darker background for the unfinished portion

### LINKED ISSUE #8262: Different color for timer lines
It's hard to differentiate the inactive part of timers against timers that are done but where the status is unknown.

image

Both the top and bottom are Done. But it seems the third one is not done when referencing against the second one.

### LINKED ISSUE #11915: Time tracking panel progress bars are hard to read/inaccessible color contrasts
**Describe the bug**
The color of the green progress bar in the farming tab, for instance, has too little contrast between the "filled" and "unfilled" parts of the progress bar.

Currently the contrasts are 1.86:1 (measured with [this tool](https://webaim.org/resources/contrastchecker/)) and the [accessibility guidelines](https://www.w3.org/TR/WCAG20/) for the web require a contrast ratio of 3:1 or 4.5:1 depending on the content

I think the background of the unfinished portion of the bar should be a darker color.

hjdarnel:
Closes #8262

The color of the green progress bar in the farming tab, for instance, has too little contrast between the "filled" and "unfilled" parts of the progress bar.

Currently the contrasts are 1.86:1 (measured with [this tool](https://webaim.org/resources/contrastchecker/)) and the [accessibility guidelines](https://www.w3.org/TR/WCAG20/) for the web require a contrast ratio of 3:1 or 4.5:1 depending on the content

I think the background of the unfinished portion of the bar should be a darker color.

Preview here (before and after): 
contrast

Adding another `darker()` would make it look like this, which IMO is too dark. 👇 
contrast 2

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/components/ThinProgressBar.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/components/ThinProgressBar.java
@@ -61,7 +61,7 @@ public class ThinProgressBar extends JPanel
 	public void setForeground(Color color)
 	{
 		super.setForeground(color);
-		setBackground(color.darker());
+		setBackground(color.darker().darker());
 	}
 
 	public void setMaximumValue(int maximumValue)


Nightfirecat:
How does this look with other bar colors?

hjdarnel:
Red: 
image

Yellow (I think? I had to force "dead" to have the same color as diseased since nothing is diseased atm 😛):
image

Issue #7940, opened 2019-02-19
Use XDG basedir instead of ~/.runelite


PureTryOut:
**Describe the bug**
Currently, Runelite stores some files in `~/.runelite`. However, it should use the XDG basebir spec instead as described [here](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), as to not clutter the home directory.

For example config files should go to `$XDG_CONFIG_HOME` with as fallback `~/.config`, and data should go to `$XDG_DATA_HOME` with as fallback `~/.local/share`. Other environment variables to be used can be found [here](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables).

**Expected behavior**
To have the files stored in the proper location rather than it's own folder in `$HOME`.

**Environment (please complete the following information):**
 - OS: Linux, but should happen on any non-Windows and non-macOS operating system
 - RuneLite version: 1.5.12
 - Launcher version: 1.6.2

ThaChillera:
I want to look into this, @Adam- can you tell me where runelite data directories are defined?

ThaChillera:
required functionality:

- Change directory location based on OS
- Add functionality to move existing directory to new locations
- Optional: add boot flag to use old location (potentionally users might want to use the old system)
- Optional: change screenshot location

coreybruce:
Oops I made a duplicate post of it but if you simply add the code I added to the launch script than it would solve the issue.

Adam-:
> Oops I made a duplicate post of it but if you simply add the code I added to the launch script than it would solve the issue.
> 
> [#14510 (comment)](https://github.com/runelite/runelite/issues/14510#event-5798607954)

Something like this would have to go into the launcher or into whatever is packaging the launcher (aur, flathub, etc). By the time this (the client) is running, the launcher has already populated ~/.runelite quite a bit and the vanilla client is writing to user.home.

coreybruce:
> > Oops I made a duplicate post of it but if you simply add the code I added to the launch script than it would solve the issue.
> > [#14510 (comment)](https://github.com/runelite/runelite/issues/14510#event-5798607954)
> 
> Something like this would have to go into the launcher or into whatever is packaging the launcher (aur, flathub, etc). By the time this (the client) is running, the launcher has already populated ~/.runelite quite a bit and the vanilla client is writing to user.home.

[...]

PR #11921, opened 2020-06-18
[WIP]  Use XDG basedir instead of ~/.runelite

ThaChillera:
This PR implements #7940 
Current state:
- [x] Change directory location based on OS
- [x] Add functionality to move existing directory to new locations
- [ ] Optional: add boot flag to use old location (potentionally users might want to use the old system)
- [ ] Optional: change screenshot location

This is my first PR, comments are welcome


Adam-:
Couple of notes:

Both the client *and* the launcher both use `~/.runelite`, so changing just this wouldn't change everything. We have limited ability to change the launcher due to it requiring people to re-download it manually.

The Jagex client writes several files into `$HOME` which this won't move and we can't move easily.

I also like the current consistency of paths between systems so I'm not sure if we want to change this?

ThaChillera:
> Both the client _and_ the launcher both use `~/.runelite`, so changing just this wouldn't change everything. We have limited ability to change the launcher due to it requiring people to re-download it manually.
Does the client need the launcher files? If not, this shouldn't pose a problem should it?

> The Jagex client writes several files into `$HOME` which this won't move and we can't move easily.
I know, that's a separate issue I've tried to contact the Jagex engine team multiple times over, but this is entirely outside runelite's scope.

> I also like the current consistency of paths between systems so I'm not sure if we want to change this?
Well let me know as soon as you decide, I hope you accept it cause its really annoying :(

Hydrox6:
The launcher's entire job is to get the files used by the client, and then launch the client.

So yeah, the client does need the launcher files

ThaChillera:
Some choices in the latest commit:
- When transferring the .runelite folder, it only deletes it once it's complete to prevent losing half your config to an error.
- If transfer fails runelite doesn't start, since having variable folders based on transfer success would be a nightmare
- Since the `OS name contains` code is repeated multiple times, this is centralised

Adam-:
What about cli switches for these:
```java
	public static final File CACHE_DIR = new File(RUNELITE_DIR, "cache");
	public static final File PLUGINS_DIR = new File(RUNELITE_DIR, "plugins");
	public static final File LOGS_DIR = new File(RUNELITE_DIR, "logs");
```

Would that do what you want? It would be a lot less intrusive/potentially destructive.

ThaChillera:
I'd prefer storing the save location on disk somewhere, since starting runelite with 6 arguments just to set save locations seems a bit much. This could default to the current situation, since that's what you prefer (if I understand correctly). Manually editing these locations would change the storage location, and would require the user to manually move the files too.
The save location config file would support environment variables, since that's what the XDG Spec is built around.

One more thing: you don't mention the Session file, should this just be removed? I don't understand its function anyway and up to now have simply thrown it into data storage (~/.local/share/runelite)

Adam-:
I think I would *prefer* for it to default to where it is now, with the option for people to change it. Occasionally we get requests to change the path on non-linux too so it would be nice if whatever the final solution is works with that too. What do you mean by the "save location config file"? I am not that familiar with XDG.

ThaChillera:
> What do you mean by the "save location config file"? I am not that familiar with XDG.

This is not part of the XDG spec, this is just an idea of mine: create a config file that's saved in the install location (C:\program files\runelite, /usr/bin/runelite) that contains the save location for each aspect (cache, plugins, screenshots etc) per OS. The file installed along runelite would contain the locations as used currently, and users could modify this file to change the locations.

An example config file: https://gist.github.com/ThaChillera/b0c6ab1dcbdde47cb994d37185610517
In this example, the $ (dollar sign) denotes environment variables.

This example is of course incomplete, missing Mac OS & other locations. And this syntax is just the first suggestion, other more verbose options could be used like JSON or XML. Alternative ways of denoting ENV variables can be discussed as well.

ThaChillera:
An issue with this proposal: If a default location file is installed along with runelite, every reinstall/update would overwrite this file.

A better solution would be that it defaults to no file, and the current location. The file can contain location(s) that overwrite the default lcoations. This would also simplify the syntax of the file, since it doesn't have to specify every OS location anymore.

I'll try to implement this today and make a new Pull request, since this isn't about XDG base spec anymore

ThaChillera:
I can't find a consistent way to get the install directory - any ideas?

Adam-:
You can't find the launcher install directory from the client. The client really shouldn't even be aware of it.

ThaChillera:
So my latest solutions are based on environment variables:

1. We could have each save location be changed based on an environment variable, (unset = default/current location)
2. Or, we could set the 'location file' location based on environment variable (cleaner solution for the end user and better solution IMO, similar to how Android Studio does it [STUDIO_PROPERTIES env var](https://developer.android.com/studio/command-line/variables)

PR #11924, opened 2020-06-18
Add toggle to remove spellbook tooltips

### LINKED ISSUE #8634: Mouse Tooltips Feature Request - Toggle off Inventory Tooltips independant of Magic
Currently, the Mouse Tooltips when enabled, in the Spellbook has a tooltip saying "Cast [x spell]", which frequently covers the description and names of the spells. This can be disabled, but this also disables Mouse Tooltips in the invntory, which I find very useful, as clicking on items in the inventory do not have a universal effect. I would argue that it is *nearly* useless in the spellbook, as a tooltip+description already appears when hovering over the spell - provided by default by Jagex - and all spells have the same function when left clicked - cast. 

I suggest separating these functions as a toggle within the Mouse Tooltips setting, There may be even more options possible that I'm not considering currently, but I am suggesting at least making disabling Inventory Tooltips a separate toggle from disabling Magic Tooltips. 

Here are a couple examples of my issue with the Magic Tooltips. It can cover the descriptions, which are often quite useful. Disabling Tooltips fixes this issue for me - but introduces the larger problem of not having Inventory Tooltips - which can be essential in many situations. You can view the top left corner for the default tooltips, but that is very out of the way and difficult to use.

zeffirojoe:
Closes #8634

Adding a config option to disable spell book tooltips.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
@@ -55,11 +55,11 @@ public interface MouseHighlightConfig extends Config
 
 	@ConfigItem(
 		position = 2,
-		keyName = "spellbookTooltip",
+		keyName = "disableSpellbooktooltip",
 		name = "Disable Spellbook Tooltips",
 		description = "Disable Spellbook Tooltips so they don't cover descriptions"
 	)
-	default boolean spellbookTooltip()
+	default boolean disableSpellbooktooltip()
 	{
 		return false;
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
@@ -52,4 +52,15 @@ public interface MouseHighlightConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 2,
+			keyName = "spellbookTooltip",
+			name = "Disable Spellbook Tooltips",
+			description = "Disable Spellbook Tooltips so they don't cover descriptions"
+	)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
```diff
@@ -52,4 +52,15 @@ default boolean chatboxTooltip()
...
 		return true;
 	}
+
+	@ConfigItem(
+			position = 2,
+			keyName = "spellbookTooltip",
+			name = "Disable Spellbook Tooltips",
+			description = "Disable Spellbook Tooltips so they don't cover descriptions"
+	)
+	default boolean spellbookTooltip()
```
Nightfirecat, line 62: This is a really confusing method name as it is opposite the actual behavior of the method.

runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
```diff
@@ -52,4 +52,15 @@ default boolean chatboxTooltip()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			position = 2,
+			keyName = "spellbookTooltip",
+			name = "Disable Spellbook Tooltips",
+			description = "Disable Spellbook Tooltips so they don't cover descriptions"
```
Nightfirecat, line 60: Reduce indents by one level.
```suggestion
		position = 2,
		keyName = "spellbookTooltip",
		name = "Disable Spellbook Tooltips",
		description = "Disable Spellbook Tooltips so they don't cover descriptions"
```

runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightConfig.java
zeffirojoe: Refactoring method name and committing suggested changes

PR #11938, opened 2020-06-19
widget-overlay: make encounter health bar moveable

### LINKED ISSUE #12020: Make CoX boss hp bar widget movable/removable
The new CoX boss hp bar widget added by Jagex takes up a large portion of the screen and often gets in the way (especially bad on fixed mode)

Make the widget movable/removable

n/a

**Additional context**
Here is an example of the new boss hp bar getting in the way (can't see the crystal hp without zooming out or rotating camera)
image

Edit: I found that you can actually disable the overlay in settings>display>advanced options>health overlays, but having it be movable would be nice
image

CopyPastaOSRS:
Allows the new toggleable health bar widget used for various boss encounters (nightmare, cox) to be moveable.

Closes #12020

image

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
+++ b/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
@@ -155,6 +155,7 @@ public class WidgetID
 	public static final int COUNTERS_LOG_GROUP_ID = 625;
 	public static final int GAUNTLET_TIMER_GROUP_ID = 637;
 	public static final int BANK_PIN_GROUP_ID = 213;
+	public static final int ENCOUNTER_HEALTH_BAR_GROUP_ID = 303;
 
 	static class WorldMap
 	{
@@ -929,5 +930,10 @@ public class WidgetID
 	static class BankPin
 	{
 		static final int CONTAINER = 0;
+	}
+
+	static class EncounterHealthBar
+	{
+		static final int CONTAINER = 6;
 	}
 }
--- a/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
+++ b/runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
@@ -155,7 +155,7 @@ public class WidgetID
 	public static final int COUNTERS_LOG_GROUP_ID = 625;
 	public static final int GAUNTLET_TIMER_GROUP_ID = 637;
 	public static final int BANK_PIN_GROUP_ID = 213;
-	public static final int ENCOUNTER_HEALTH_BAR_GROUP_ID = 303;
+	public static final int HEALTH_OVERLAY_BAR_GROUP_ID = 303;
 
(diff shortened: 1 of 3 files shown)


CopyPastaOSRS:
> 
> 
> I don't raid regularly so I'm not a good person to test this. Does making this widget into an overlay not flicker? Other similar bars (Kourend favor, underwater breath bar, etc.) tend to snap back to their original position for a frame whenever they're animated/updated.

Here is a clip of the health bar widget during Tekton. I don't see any artifacts where the bar is normally (top-center).
https://streamable.com/blr6kv

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -548,7 +548,9 @@
 
 	SKILLS_CONTAINER(WidgetID.SKILLS_GROUP_ID, WidgetID.Skills.CONTAINER),
 
-	GAUNTLET_TIMER_CONTAINER(WidgetID.GAUNTLET_TIMER_GROUP_ID, WidgetID.GauntletTimer.CONTAINER);
+	GAUNTLET_TIMER_CONTAINER(WidgetID.GAUNTLET_TIMER_GROUP_ID, WidgetID.GauntletTimer.CONTAINER),
+
+	ENCOUNTER_HEALTH_BAR(WidgetID.ENCOUNTER_HEALTH_BAR_GROUP_ID, WidgetID.EncounterHealthBar.CONTAINER);
```
Nightfirecat, line 553: This should be named `COX_BOSS_HEALTH_BAR` since it's specific to raids

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -155,6 +155,7 @@
 	public static final int COUNTERS_LOG_GROUP_ID = 625;
 	public static final int GAUNTLET_TIMER_GROUP_ID = 637;
 	public static final int BANK_PIN_GROUP_ID = 213;
+	public static final int ENCOUNTER_HEALTH_BAR_GROUP_ID = 303;
```
Nightfirecat, line 158: Same here, this should be renamed

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
CopyPastaOSRS: This health bar widget is not unique to raids (Galvek, The Nightmare, Vanstrom Klause). A generic toggle was added in last weeks update so I'd expect Jagex to also use it for other encounters in the future.

image

image

Nightfirecat: Ah I didn't realize it was generic to all of those. Maybe renaming this to be more in line with Jagex's naming then? `HEALTH_OVERLAY_BAR`?

PR #11940, opened 2020-06-20
Add GE links to the offers panel

Shingyx:
This allows looking up GE prices for items already on offer, without the need to manually search for the item on the Search tab. Accessed via a new popup menu.

I made this PR because I've always wanted a way to lookup prices for an item in the Offers tab to see if prices have changed since I made the offer.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeItemPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeItemPanel.java
@@ -39,9 +39,8 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
-import net.runelite.client.util.AsyncBufferedImage;
 import net.runelite.client.ui.ColorScheme;
-import net.runelite.client.util.LinkBrowser;
+import net.runelite.client.util.AsyncBufferedImage;
 import net.runelite.client.util.QuantityFormatter;
 
 /**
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeLinks.java
```diff
@@ -0,0 +1,39 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.grandexchange;
+
+import net.runelite.client.util.LinkBrowser;
+
+class GrandExchangeLinks
+{
+	static void openGeLink(String name, int itemId)
```
Adam-, line 31: this class seems unnecessary, instead you could just move this function into the plugin.

runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeLinks.java
```diff
@@ -0,0 +1,39 @@
...
+	{
+		final String url = "http://services.runescape.com/m=itemdb_oldschool/"
```
Adam-, line 33: I believe they recently changed this to always redirect to https, so it may be worth updating it here.

Shingyx: I also made it use + instead of _ to avoid the redirect.

I.e. compare
https://services.runescape.com/m=itemdb_oldschool/Limpwurt+root/viewitem?obj=225
and
https://services.runescape.com/m=itemdb_oldschool/Limpwurt_root/viewitem?obj=225

The second redirects to https://secure.runescape.com/m=itemdb_oldschool/a=13/Limpwurt+root/viewitem?obj=225

PR #11973, opened 2020-06-24
raid plugin: add raid scouted/reset events

Trevor159:
This is made for external plugins


runelite-client/src/main/java/net/runelite/client/plugins/raids/RoomType.java
```diff
@@ -30,7 +30,7 @@
 
 @AllArgsConstructor
 @Getter
-enum RoomType
+public enum RoomType
```
Nightfirecat, line 33: All these classes being made public should probably move out of the plugin package. Maybe net.runelite.client.game.raids would be a good package for them?

PR #11990, opened 2020-06-26
ClientUI: Allow hotkey plugin panel toggles on login screen

abextm:
I'm not sure why this wasn't included with the rest of them, but this is confusing and should be the same for both the sidebar and plugin panel toggles


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -394,7 +394,7 @@ public void hotkeyPressed()
 					togglePluginPanel();
 				}
 			};
-
+			pluginPanelListener.setEnabledOnLogin(true);
```
Nightfirecat, line 397: newline above I guess

PR #12010, opened 2020-06-28
Consolidate 'logged out/in' messages to a single 'hopped worlds' message

TheStonedTurtle:
Not sure if there's a better way to check when someones login status changes as the chat messages are delayed.

image

This also suppresses log in/out messages if they logged back in to the same world within the 10 tick timeframe.


TheStonedTurtle:
I've updated this PR to live inside the `Chat Filter` plugin instead of the `Friends List` plugin.

PR #12019, opened 2020-06-28
music: add option to mute pet sounds

Broooklyn:
It is currently only possible to mute pet sounds (e.g., Pet general graardor stomp, electric hydra idle sound, etc.) via `Mute NPCs' area sounds`. However, that option turns off _all_ NPC sounds, which can be disadvantageous in places like ToB, CoX, or any other activities where game mechanics are closely tied to area sounds.

This commit introduces the option to disable pet sounds, so you never have to hear the  _thump thump thump_ of an undeserved Bandos pet again, while still retaining the advantages of enabled area sounds.

Pet sound behavior for some is slightly different when a pet is placed in a PoH menagerie compared to having it as a follower. For example, Jagex removed the sounds from the pet Kree'arra wing flap and the Electric Hydra idle animation—but only while they are following you or another player. The sounds still play when they are free-roaming in a PoH. Other pets, like Pet general graardor and Skotos, still have sounds when they walk regardless of whether they are a follower or free roaming. This commit does not take into account the fact that many pets can make noise when interacting with a scratching post, as I deemed this willing behavior (get rid of the scratching post if you don't want pets interacting with it).

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/SoundEffectID.java
+++ b/runelite-api/src/main/java/net/runelite/api/SoundEffectID.java
@@ -96,4 +96,14 @@ public final class SoundEffectID
 	public final static int TOWN_CRIER_BELL_DONG = 3817;
 	public final static int TOWN_CRIER_SHOUT_SQUEAK = 3816;
 
+	// Pet sounds
+	public final static int CAT_HISS = 333;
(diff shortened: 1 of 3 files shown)


Broooklyn:
Annoyance Mute on the plugin hub accomplishes this now

PR #12026, opened 2020-06-28
Fix Var Inspector and make it more efficient

aeramos:
This makes the Var Inspector work for all varbits, and it should be
futureproof for when Jagex adds more or if there are holes in the list
of varbits.

This also simplifies the code for checking which varp changed by using
the VarbitChanged class.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/devtools/VarInspector.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/devtools/VarInspector.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018 Abex
+ * Copyright (c) 2020, Alejandro Ramos (github.com/aeramos)
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/devtools/VarInspector.java
```diff
@@ -226,10 +228,7 @@ public void onVarbitChanged(VarbitChanged ev)
 				int neew = client.getVarbitValue(varps, i);
 				if (old != neew)
 				{
-					// Set the varbit so it doesn't show in the varp changes
```
abextm, line 229: this is wrong if a varb change doesn't consume the whole varp

Adam-:
I've exposed some of the varbit and config index to allow the inspector to more intelligently know how many varbits there are in 62bbf35baf6061fcb1736c470208507d1d09b800. fixed in 53c8593929c89290530f18f52eed8117b3ef8a21.

With regard to the `oldVarps2` removal, it will break if a varp is modified that partially crosses a varbit boundary as abex has said. I'm not sure this ever happens? But I've left the logic in to detect that currently.

aeramos:
@Adam- I think that your commit broke varbit detection. I'm trying it out right now and it's only recording varp changes, and if a varbit changes it just shows the change to its varp.

aeramos:
Everything works fine now, thank you. I'll close my pull request.

PR #12032, opened 2020-06-29
Adding config option to persist tags through restart

### LINKED ISSUE #11961: Shift+Right click to add/remove NPC name to NPC Indicators Highlight list
It is not an issue, but would be an amazing addition towards the NPC indicator, rather than having to type in / delete everytime you'd like to change it up.

Have an option in the NPC indicators plugin where you can shift right click NPCs, to add it to the highlighted list, or remove it. (kind of like the option when you can mark ground tiles)

zeffirojoe:
Closes #11961

Adding config option to have the "Tag" or "Untag" option to instead modify the "Npcs To Highlight" option which allows the tags to persist through restart.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
@@ -88,6 +88,13 @@ public interface NpcIndicatorsConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "npcToHighlight",
+		name = "",
+		description = ""
+	)
+	void setNpcToHighlight(String npcsToHighlight);
+
+	@ConfigItem(
 		position = 4,
 		keyName = "npcColor",
 		name = "Highlight Color",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
@@ -85,6 +85,9 @@ public class NpcIndicatorsPlugin extends Plugin
 	private static final String TAG = "Tag";
 	private static final String UNTAG = "Un-tag";
 
+	private static final String TAG_ALL = "Tag-All";
+	private static final String UNTAG_ALL = "Un-tag-All";
+
 	private static final Set<MenuAction> NPC_MENU_ACTIONS = ImmutableSet.of(MenuAction.NPC_FIRST_OPTION, MenuAction.NPC_SECOND_OPTION,
 		MenuAction.NPC_THIRD_OPTION, MenuAction.NPC_FOURTH_OPTION, MenuAction.NPC_FIFTH_OPTION, MenuAction.SPELL_CAST_ON_NPC,
 		MenuAction.ITEM_USE_ON_NPC);
@@ -270,16 +273,48 @@ public class NpcIndicatorsPlugin extends Plugin
 		}
 		else if (menuAction == MenuAction.EXAMINE_NPC && client.isKeyPressed(KeyCode.KC_SHIFT))
 		{
-			// Add tag option
+			// Add tag and tag-all options
+			final int id = event.getIdentifier();
+			final NPC[] cachedNPCs = client.getCachedNPCs();
+			final NPC npc = cachedNPCs[id];
+
+			if (npc == null || npc.getName() == null)
+			{
+				return;
+			}
+
+			final String npcName = npc.getName();
+			boolean matchesList = highlights.stream()
+				.filter(highlight -> !highlight.equalsIgnoreCase(npcName))
+				.anyMatch(highlight -> WildcardMatcher.matches(highlight, npcName));
+
 			MenuEntry[] menuEntries = client.getMenuEntries();
-			menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 1);
+
+			// Only add Untag-All option to npcs not highlighted by a wildcard entry, because untag-all will not remove wildcards
+			if (!matchesList)
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 2);
+				final MenuEntry tagAllEntry = menuEntries[menuEntries.length - 2] = new MenuEntry();
+				tagAllEntry.setOption(highlights.stream().anyMatch(npcName::equalsIgnoreCase) ? UNTAG_ALL : TAG_ALL);
+				tagAllEntry.setTarget(event.getTarget());
+				tagAllEntry.setParam0(event.getActionParam0());
+				tagAllEntry.setParam1(event.getActionParam1());
+				tagAllEntry.setIdentifier(event.getIdentifier());
+				tagAllEntry.setType(MenuAction.RUNELITE.getId());
+			}
+			else
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 1);
+			}
+
 			final MenuEntry tagEntry = menuEntries[menuEntries.length - 1] = new MenuEntry();
-			tagEntry.setOption(highlightedNpcs.stream().anyMatch(npc -> npc.getIndex() == event.getIdentifier()) ? UNTAG : TAG);
+			tagEntry.setOption(highlightedNpcs.contains(npc) ? UNTAG : TAG);
 			tagEntry.setTarget(event.getTarget());
 			tagEntry.setParam0(event.getActionParam0());
 			tagEntry.setParam1(event.getActionParam1());
 			tagEntry.setIdentifier(event.getIdentifier());
 			tagEntry.setType(MenuAction.RUNELITE.getId());
+
 			client.setMenuEntries(menuEntries);
 		}
 	}
@@ -288,13 +323,13 @@ public class NpcIndicatorsPlugin extends Plugin
 	public void onMenuOptionClicked(MenuOptionClicked click)
 	{
 		if (click.getMenuAction() != MenuAction.RUNELITE ||
-			!(click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG)))
+			!(click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG) ||
+				click.getMenuOption().equals(TAG_ALL) || click.getMenuOption().equals(UNTAG_ALL)))
 		{
 			return;
 		}
 
 		final int id = click.getId();
-		final boolean removed = npcTags.remove(id);
 		final NPC[] cachedNPCs = client.getCachedNPCs();
 		final NPC npc = cachedNPCs[id];
 
@@ -303,19 +338,29 @@ public class NpcIndicatorsPlugin extends Plugin
 			return;
 		}
 
-		if (removed)
+		if (click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG))
 		{
-			highlightedNpcs.remove(npc);
-			memorizedNpcs.remove(npc.getIndex());
+			final boolean removed = npcTags.remove(id);
+
+			if (removed)
+			{
+				highlightedNpcs.remove(npc);
+				memorizedNpcs.remove(npc.getIndex());
+			}
+			else
+			{
+				if (!client.isInInstancedRegion())
+				{
+					memorizeNpc(npc);
+					npcTags.add(id);
+				}
+				highlightedNpcs.add(npc);
+			}
 		}
 		else
 		{
-			if (!client.isInInstancedRegion())
-			{
-				memorizeNpc(npc);
-				npcTags.add(id);
-			}
-			highlightedNpcs.add(npc);
+			final String name = npc.getName();
+			updateNpcsToHighlight(name);
 		}
 
 		click.consume();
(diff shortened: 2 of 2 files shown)


Adam-:
I think I'd rather see a `Tag-All` option instead of a config option to change which menu is shown.

zeffirojoe:
Removing config option for original idea and adding a new context option which will either add or remove the npc name from the "Highlighted Npcs" textbox in the config. This option is denoted as Tag-All or Un-tag-All

Adam-:
NPCs can also be highlighted with wildcards, eg `*demon*`, which will "tag all" of an npc type. This looks like you are making it so the menu option would be `Tag-All` when something is highlighted with a wildcard which when clicked wouldn't visibly do anything since they are already tagged - is this intentional? Oh the other hand it might be weird if "Untag-all" removed a wildcard entry which was also used to highlight other things.

Adam-:
After some more thinking I think it would be best to only add the tag-all/untag-all options if the entry doesn't wildcard match something in the list eg:

```java
boolean matchesList = highlights.stream()
	.filter(highlight -> !highlight.equals(npc.getName().toLowerCase()))
	.anyMatch(highlight -> WildcardMatcher.matches(highlight, npc.getName().toLowerCase()));
if (!matchesList) ...
```

Otherwise it is either always 1) tagging something which is already tagged or 2) not actually untagging something when you click untag all - which I think would be confusing.

zeffirojoe:
Adding a check to only display option based on wildcard matching as suggested.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,50 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			boolean matchesList = false;
+
+			if (npcCached == null || npcCached.getName() == null)
+			{
+				npcFound = false;
+			}
+			else
+			{
+				matchesList = highlights.stream()
+					.filter(highlight -> !highlight.equals(npcCached.getName().toLowerCase()))
```
Adam-, line 284: Declare `final String npcName = npc.getName().toLowerCase();` once and then reference it and below.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,50 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			}
+
 			MenuEntry[] menuEntries = client.getMenuEntries();
-			menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 1);
+
+			if (npcFound && !matchesList)
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 2);
+				final MenuEntry tagAllEntry = menuEntries[menuEntries.length - 2] = new MenuEntry();
+				tagAllEntry.setOption((Arrays.stream(config.getNpcToHighlight().split(",")).anyMatch(npc -> npc.toLowerCase().equals(npcCached.getName().toLowerCase())) ? UNTAG_ALL : TAG_ALL));
```
Adam-, line 294:
```suggestion
				tagAllEntry.setOption(highlights.contains(npcName) ? UNTAG_ALL : TAG_ALL));
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,50 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+				tagAllEntry.setType(MenuAction.RUNELITE.getId());
+			}
+			else
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 1);
+			}
+
 			final MenuEntry tagEntry = menuEntries[menuEntries.length - 1] = new MenuEntry();
-			tagEntry.setOption(highlightedNpcs.stream().anyMatch(npc -> npc.getIndex() == event.getIdentifier()) ? UNTAG : TAG);
+			tagEntry.setOption((highlightedNpcs.stream().anyMatch(npc -> npc.getIndex() == event.getIdentifier()) ? UNTAG : TAG));
```
Adam-, line 307: This just becomes `highlighedNpcs.contains(npc)`

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -288,37 +319,69 @@ else if (menuAction == MenuAction.EXAMINE_NPC && client.isKeyPressed(KeyCode.KC_
...
 			{
-				memorizeNpc(npc);
-				npcTags.add(id);
+				highlightNames = new ArrayList<>(Arrays.asList(npcNames.split(",")));
 			}
-			highlightedNpcs.add(npc);
-		}
 
-		click.consume();
+			final int id = click.getId();
```
Adam-, line 367: These 3 lines + the conditionals below are duplicated in both branches, so move them to be above.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -288,37 +319,69 @@ else if (menuAction == MenuAction.EXAMINE_NPC && client.isKeyPressed(KeyCode.KC_
...
+			}
+
+			final String name = npc.getName().toLowerCase();
+			final boolean removed = highlightNames.remove(name);
+
+			if (!removed)
+			{
+				highlightNames.add(name);
+			}
+			config.setNpcToHighlight(String.join(",", highlightNames));
```
Adam-, line 383: This is force tolowering the user's config option which it should not be doing.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,50 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			// Add tag and tag-all options
+			final int id = event.getIdentifier();
+			final NPC[] cachedNPCs = client.getCachedNPCs();
+			final NPC npcCached = cachedNPCs[id];
+			boolean npcFound = true;
+			boolean matchesList = false;
+
+			if (npcCached == null || npcCached.getName() == null)
+			{
+				npcFound = false;
```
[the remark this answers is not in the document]
zeffirojoe: Even if this fails the normal "Tag" option can still be displayed since it's based on the event information. Only reason I did this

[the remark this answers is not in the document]
Adam-: Right, but I don't think that makes sense? I think it either can't happen, or if it did happen isn't something we would want to try and handle anway.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
zeffirojoe: How would you suggest doing this then? Just directly adding to or removing the "highlights" List?

Adam-: I recommend looking at `updateList` in the ground items plugin

zeffirojoe:
Making attempt at implementing suggested changes.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -303,22 +331,33 @@ public void onMenuOptionClicked(MenuOptionClicked click)
...
+					npcTags.add(id);
+				}
+				highlightedNpcs.add(npc);
+			}
+
+			click.consume();
 		}
+		else
+		{
+			final String name = npc.getName().toLowerCase();
```
Adam-, line 357: This lowercase isn't necessary and just causes it to be added to the user's list in lowercase.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,47 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+				tagAllEntry.setType(MenuAction.RUNELITE.getId());
+			}
+			else
+			{
+				menuEntries = Arrays.copyOf(menuEntries, menuEntries.length + 1);
+			}
+
 			final MenuEntry tagEntry = menuEntries[menuEntries.length - 1] = new MenuEntry();
-			tagEntry.setOption(highlightedNpcs.stream().anyMatch(npc -> npc.getIndex() == event.getIdentifier()) ? UNTAG : TAG);
+			tagEntry.setOption((highlightedNpcs.contains(npcCached) ? UNTAG : TAG));
```
Adam-, line 304:
```suggestion
			tagEntry.setOption(highlightedNpcs.contains(npcCached) ? UNTAG : TAG);
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,47 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+			final NPC[] cachedNPCs = client.getCachedNPCs();
+			final NPC npcCached = cachedNPCs[id];
+
+			if (npcCached == null || npcCached.getName() == null)
+			{
+				return;
+			}
+
+			final String npcName = npcCached.getName().toLowerCase();
+			boolean	matchesList = highlights.stream()
```
Adam-, line 281: This should include a comment explaining what it is trying to do, eg
```java
// Only add Untag options to npcs not highlighted by a wildcard entry, because untag-all will not remove wildcards
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -270,16 +267,47 @@ public void onMenuEntryAdded(MenuEntryAdded event)
 		}
 		else if (menuAction == MenuAction.EXAMINE_NPC && client.isKeyPressed(KeyCode.KC_SHIFT))
 		{
-			// Add tag option
+			// Add tag and tag-all options
+			final int id = event.getIdentifier();
+			final NPC[] cachedNPCs = client.getCachedNPCs();
+			final NPC npcCached = cachedNPCs[id];
```
Adam-, line 273:
```suggestion
			final NPC npc = cachedNPCs[id];
```

The "cached" name the API uses is basically mis-named currently.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -388,6 +427,18 @@ public void onGameTick(GameTick event)
...
+	void updateList(String item)
+	{
+		final List<String> highlightedNpcs = new ArrayList<>(highlights);
+
+		if (!highlightedNpcs.removeIf(item::equalsIgnoreCase))
+		{
+			highlightedNpcs.add(item);
+		}
+
+		config.setNpcToHighlight(Text.toCSV(highlightedNpcs));
```
Adam-, line 439: This works due to this triggering the config change event to rebuild npcs right? Would add
```java
// this triggers the config change event and rebuilds npcs
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -288,13 +316,13 @@ else if (menuAction == MenuAction.EXAMINE_NPC && client.isKeyPressed(KeyCode.KC_
 	public void onMenuOptionClicked(MenuOptionClicked click)
 	{
 		if (click.getMenuAction() != MenuAction.RUNELITE ||
-			!(click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG)))
+			(!(click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG)) &&
```
Adam-, line 319: This reduces to:

```suggestion
			!(click.getMenuOption().equals(TAG) || click.getMenuOption().equals(UNTAG) || click.getMenuOption().equals(TAG_ALL) || click.getMenuOption().equals(UNTAG_ALL))
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -388,6 +427,18 @@ public void onGameTick(GameTick event)
 		lastPlayerLocation = client.getLocalPlayer().getWorldLocation();
 	}
 
+	void updateList(String item)
```
Adam-, line 430:
```suggestion
	private void updateNpcsToHighlight(String npc)
```

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
```diff
@@ -87,6 +87,13 @@ default String getNpcToHighlight()
 		return "";
 	}
 
+	@ConfigItem(
+		keyName = "npcToHighlight",
+		name = "",
+		description = ""
+	)
+	void setNpcToHighlight(String key);
```
Adam-, line 95:
```suggestion
	void setNpcToHighlight(String npcsToHighlight);
```

zeffirojoe:
Adding the most recent set of suggestions

PR #12035, opened 2020-06-29
ChatNotifications: Add ticksSinceLogin check

### LINKED ISSUE #11706: Repeated Chat Notification trigger on world hop
**Describe the bug**
With `Chat Notifications` and `Chat History`: `Retain Chat History` enabled, if a notification word is sent to chat, the chat notification is triggered every time you hop worlds. This only occurs when the trigger for the chat notification is one word.

**To Reproduce**
1. Enable `Retain Chat History` within `Chat History`
2. Trigger a single word `Chat Notification`.
3. Hop worlds or log out.
4. On login the chat notification triggers again.

**Expected behavior**
I would expect to not be spammed by the same chat notification every time I hop worlds, while still being able to see chat history.

**Screenshots**
Original trigger:
Screen Shot 2020-05-26 at 7 54 47 PM

After world hop:
Screen Shot 2020-05-26 at 7 55 04 PM

**Environment:**
 - OS: MacOS
 - RuneLite version: 1.6.16
 - Launcher version: 2.1.2

Broooklyn:
Prevents chat notifications from firing immediately upon login due to `ChatHistoryPlugin` re-printing a highlighted (and already notified) string.

Closes #11706

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
@@ -38,6 +38,7 @@ import net.runelite.api.Client;
 import net.runelite.api.MessageNode;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.client.Notifier;
 import net.runelite.client.RuneLiteProperties;
 import net.runelite.client.chat.ChatColorType;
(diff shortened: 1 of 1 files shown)


Broooklyn:
So I can update this to do the same as https://github.com/runelite/runelite/pull/12052 but it doesn't seem to work consistently, although I now realize after further testing that neither does this PR. It seems sometimes chat history is loaded prior to the first game tick, but sometimes that isn't the case and it loads after. I Could set the delay to `ticksSinceLogin >= 2`, which seems to work properly, but I'm not sure if that's the best approach or not.

runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
```diff
@@ -74,6 +75,8 @@
 	private String usernameReplacer = "";
 	private Pattern highlightMatcher = null;
 
+	private int ticksSinceLogin;
```
Adam-, line 78: This can just be a `boolean`

PR #12048, opened 2020-07-01
Key Mapping Plugin and authenticator

### LINKED ISSUE #12045: Key Remapping Plugin does not allow numeric input when asking for 6 digit authentication code
This bug occurred to me after putting my password in my RuneScape account and being asked for my authentication code. WIth the Key Remapping plugin enabled, I have the F Key Remapping section checked as I have a 60% keyboard that has no F keys without holding the Fn button, and this plugin works great to circumvent remapping my keyboard or holding the Fn key when using my F keys. With this plugin turned on, RuneLite would not accept any numeric input when I was prompted for my authentication code until I turned off the plugin. Once the plugin was turned off, it would accept my numeric input. The numeric keys used are the ones above QWERTY on the keyboard and not on the num pad as my keyboard does not have a num pad. With the plugin turned on as well, I have no issues entering numeric values into my username or password fields when logging into RuneScape; this only occurred to me when I was asked for my 6 digit authentication code.

**To Reproduce**
1. Turn the Key Remapping plugin on and ensure the F keys box is checked
2. Log-in to your RuneScape account in RuneLite
3. When prompted for the 6-digit authentication code, try entering in the numbers. This is where it would not accept my input and I would have to turn off the plugin to enter my 6 digit code.

**Expected behavior**
When prompted for my 6-digit authentication code, I should be able to enter my code with the numbers on the keyboard, but with the Key Remapping plugin turned on, this was not working. This is when I attempt to enter the code with the numbers above the letters on the keyboard, and not from the num pad as my keyboard does not have a num pad.

**Environment:**
 - OS: Win 10 Pro version 1903, OS build 18362.900
 - RuneLite version: 1.6.21.1
 - Launcher version: 2.1.0

zeffirojoe:
Closes #12045

Adding change to allow authenticator use with Key Remap Plugin. Currently the functionality is only prevented on the Login Screen and not the authenticator screen as well

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -33,7 +33,6 @@ import java.util.Map;
 import java.util.Set;
 import javax.inject.Inject;
 import net.runelite.api.Client;
-import net.runelite.api.GameState;
 import net.runelite.api.VarClientStr;
 import net.runelite.client.callback.ClientThread;
 import net.runelite.client.input.KeyListener;
@@ -68,7 +67,7 @@ class KeyRemappingListener implements KeyListener
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (client.getGameState() == GameState.LOGIN_SCREEN || !plugin.chatboxFocused())
+		if (!plugin.chatboxFocused())
 		{
 			return;
 		}
--- a/runelite-client/src/main/java/net/runelite/client/input/KeyManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/input/KeyManager.java
@@ -145,6 +145,6 @@ public class KeyManager
 			return true;
 		}
 
-		return client.getGameState() != GameState.LOGIN_SCREEN;
+		return client.getGameState() != GameState.LOGIN_SCREEN && client.getGameState() != GameState.LOGIN_SCREEN_AUTHENTICATOR;
 	}
 }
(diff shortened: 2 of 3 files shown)


Adam-:
This looks to me like it is doing the opposite of what #12045 wants.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
```diff
@@ -68,7 +68,7 @@ public void keyTyped(KeyEvent e)
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (client.getGameState() == GameState.LOGIN_SCREEN || !plugin.chatboxFocused())
+		if (client.getGameState() == GameState.LOGIN_SCREEN || client.getGameState() == GameState.LOGIN_SCREEN_AUTHENTICATOR || !plugin.chatboxFocused())
```
Nightfirecat, line 71: This change should be made for all plugins, not just the keyremapping plugin. Move this change to KeyManager.

zeffirojoe:
Was the most recent commit the correct way to move it to KeyManager?

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
Adam-: I don't think the key remapper uses the key manager though, since it has its own remap logic?

zeffirojoe:
It looks like this is approved but not merged, is there anything else I'm supposed to do?

PR #12051, opened 2020-07-02
WorldService retries world fetch when very first request fails

### LINKED ISSUE #11819: Worldhopper crashes and requires a runelite restart on bad connections
**Describe the bug**
The worldhopper seems to be crashing a lot here (my connection is not that good), so the plugin tab keeps blank. And I can't use it from plugins if it has crashed.

**To Reproduce**
1. Have a jittery connection
2. Open runelite
3. Wait
4. See stacktrace

**Expected behavior**
The worldhopper should recover from a bad connection, restarting it again.

**Screenshots**
Sadly I don't have a screenshot right now, if it happens again I will update this issue

**Stacktace**
You can find the stacktrace here: https://pastebin.com/MrWJ5KwK

 - OS: Ubuntu
 - RuneLite version: 1.6.19 (i've tracked this bug since 1.6.17)
 - Launcher version: Compiled myself

UgiR:
First contribution.. Let me know if I missed anything.

--------

This PR introduces some basic retry logic for the WorldService, addressing #11819.

Note that the world fetch is only retried if the very first request fails. Any subsequent failures will simply use the 'outdated' world list. This is only meant to address the issue where the world hopper tab is left completely blank after the first fetch for worlds fails.

The retries are limited to three (with increasing delays) as to not send too many requests.

While this closes #11819, it is still possible to get a blank world hopper tab (if all three requests fail), but that's just the nature of network calls..

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/WorldService.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/WorldService.java
@@ -50,12 +50,14 @@ import okhttp3.OkHttpClient;
 public class WorldService
 {
 	private static final int WORLD_FETCH_TIMER = 10; // minutes
+	private static final int MAX_FETCH_RETRY = 3; // maximum retries for *first* request failure
 
 	private final Client client;
 	private final ScheduledExecutorService scheduledExecutorService;
 	private final WorldClient worldClient;
 	private final EventBus eventBus;
 	private final CompletableFuture<WorldResult> firstRunFuture = new CompletableFuture<>();
+	private int firstRunNumRetries = 0;
 
 	private WorldResult worlds;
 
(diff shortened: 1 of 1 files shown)


Adam-:
okhttp has the ability to automatically retry failed requests I believe. I don't know much about it (is it on now? maybe we need to turn it on?). This may be worth investigating first before we try to do something like this.

Adam-:
I'm unsure #11819 is even a bug? It suggests that if the worlds fail to load then they will never ever load in the future, but I don't think that is true based off of what the code does currently?

UgiR:
Honestly don't think it's a bug.. They would reload after 10 minutes (assuming the next request doesn't fail.

As for retries.. f I understand correctly, all API calls from RuneLite and made synchronously. An interceptor can be added to the okhttp client to do simple retries, but that can cause other issues. For example, if your server is already overloaded and dropping requests, you don't want every client retrying request immediately. For retries with backoff, you would want to use some asynchronous calls.

Adam-:
https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#retryOnConnectionFailure-boolean- this is what I was referring to, which looks like it is on by default. I'm unsure if it would apply in this situation though.

PR #12062, opened 2020-07-04
tooltips: make overlay color background option affect tooltips

### LINKED ISSUE #11779: "Overlay Color" option does not affect Mouse Tooltips
The tooltips are designed similarly to other overlays and infoboxes, but they're unaffected due to being an `Overlay`, which can be confusing for users and feels like an oversight.

Thanks to `Florian` on the Discord for bringing this up.

melkypie:
Fixes #11779


deathbeam:
Not sure why is this labeled as bug, but looks good

PR #12071, opened 2020-07-06
Add ability to automatically screenshot Barbarian Assault high gamble rewards

LA:
The clan chat I'm in has a game that requires having screenshots for BA high gambles.

I thought this feature might be helpful for others, as well.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -645,8 +645,9 @@ public class ScreenshotPlugin extends Plugin
 	 * Parses the Barbarian Assault high gamble reward dialog text into a shortened string for filename usage.
 	 *
 	 * @param text The {@link Widget#getText() text} of the {@link WidgetInfo#DIALOG_SPRITE_TEXT} widget.
-	 * @return Shortened string in the format "(100) Watermelon seed (x50)"
+	 * @return Shortened string in the format "High Gamble(100)"
 	 */
+	@VisibleForTesting
 	static String parseBAHighGambleWidget(final String text)
 	{
 		final Matcher highGambleMatch = BA_HIGH_GAMBLE_REWARD_PATTERN.matcher(text);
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -110,6 +110,7 @@
 	private static final ImmutableList<String> PET_MESSAGES = ImmutableList.of("You have a funny feeling like you're being followed",
 		"You feel something weird sneaking into your backpack",
 		"You have a funny feeling like you would have been followed");
+	private static final Pattern BA_HIGH_GAMBLE_REWARD_PATTERN = Pattern.compile("(.+)!<br>High level gamble count: <col=7f0000>(.+)</col>");
```
Nightfirecat, line 113: What is an example of text you'd receive when doing a high gamble? That should be added to this plugin's test cases.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -626,12 +641,33 @@ else if (verb.contains("completely"))
...
+	 */
+	static String parseBAHighGambleWidget(final String text)
+	{
+		final Matcher highGambleMatch = BA_HIGH_GAMBLE_REWARD_PATTERN.matcher(text);
+		if (highGambleMatch.find())
+		{
+			String reward = highGambleMatch.group(1);
+			String gambleCount = highGambleMatch.group(2);
+
+			return String.format("(%s) %s", gambleCount, reward);
```
Nightfirecat, line 658: This should be formatted `High Gamble(count)` like other screenshots. I don't think putting the reward in the filename is valuable.

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
Nightfirecat, line 113: Now that we're not capturing the reward, we can remove the capturing group and use `.group(1)` to get the gamble count. Alternatively, each of the groups should be given a name using the syntax `(?<name>pattern)` and the matcher changed to use [`Matcher.group(String name)`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#group-java.lang.String-) to fetch the captured text instead.
```suggestion
	private static final Pattern BA_HIGH_GAMBLE_REWARD_PATTERN = Pattern.compile("(?<reward>.+)!<br>High level gamble count: <col=7f0000>(?<gambleCount>.+)</col>");
```

PR #12073, opened 2020-07-06
Add per-NPC highlight settings

jesse1412:
Allows you to provide NPC-specific color and highlight styles by appending variables to the NPCs name in the "NPCs to Highlight" config option. Example input: "Banker:r255:g255:hull:tile" would highlight all NPCs named "Banker" yellow using the "Hull" and "Tile" highlighting styles.

Video demonstration [here](https://streamable.com/0zbjlw).


Adam-:
This seems pretty hacky to me; we don't have anything else that behaves like this. The similar things like multicolor ground markers just remember the settings at the time it was marked. Which I think would be a lot more friendly UX-wise.

PR #12083, opened 2020-07-08
Improve readability of texts inside infoboxes with outline and/or background

molteber:
A littlebit outdated, see https://github.com/runelite/runelite/pull/12083#issuecomment-655442844 for newer information

---

When I've been playing in NMZ, I have to struggle a little bit to read how many absorption points I have left.
I saw that you recently added an update with some outlined text and took that chance to add that same functionality to the info boxes, but also explicitly on the NMZ plugin.

Some pictures:  
**Before:**  
Screenshot from 2020-07-08 03-33-31

**After:**  
Screenshot from 2020-07-08 03-46-35

Did not think about taking pictures when there was 100+ points left, but I think it's even harder to read then (without the outline).

Hope this little change is taken into consideration. 

*Maybe the info boxes can get a setting in Runelite in general in the future, opt-in for outline in all info boxes. Who knows*

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/BackgroundComponent.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/BackgroundComponent.java
@@ -47,24 +47,11 @@ public class BackgroundComponent implements RenderableEntity
 	private Color backgroundColor = ComponentConstants.STANDARD_BACKGROUND_COLOR;
 	private Rectangle rectangle = new Rectangle();
 	private boolean fill = true;
+	private boolean border = true;
 
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		Color outsideStrokeColor = new Color(
-			(int) (backgroundColor.getRed() * OUTER_COLOR_OFFSET),
-			(int) (backgroundColor.getGreen() * OUTER_COLOR_OFFSET),
-			(int) (backgroundColor.getBlue() * OUTER_COLOR_OFFSET),
-			Math.min(255, (int) (backgroundColor.getAlpha() * ALPHA_COLOR_OFFSET))
-		);
-
-		Color insideStrokeColor = new Color(
-			Math.min(255, (int) (backgroundColor.getRed() * INNER_COLOR_OFFSET)),
-			Math.min(255, (int) (backgroundColor.getGreen() * INNER_COLOR_OFFSET)),
-			Math.min(255, (int) (backgroundColor.getBlue() * INNER_COLOR_OFFSET)),
-			Math.min(255, (int) (backgroundColor.getAlpha() * ALPHA_COLOR_OFFSET))
-		);
-
 		// Render background
 		if (fill)
 		{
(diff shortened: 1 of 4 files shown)


melkypie:
This should probably be a runelite config option for all infoboxes rather than just adding it for a single plugin.

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBox.java
```diff
@@ -81,4 +83,14 @@ public boolean cull()
 	{
 		return false;
 	}
+
+	public boolean getOutline()
```
melkypie, line 87: Rather than writing get/set methods by hand use `@Getter` and `@Setter` annoations on the `outline` field

molteber:
Fair point about adding it to the RuneLite settings immediately.

I've removed things from the plugin in it's full, and only focused on the client config and the infobox.
Now the infoboxes can only be manipulated through the config.

Attaching some final screenshots
Screenshot from 2020-07-08 12-25-26
Screenshot from 2020-07-08 12-25-38
Screenshot from 2020-07-08 12-25-51
Screenshot from 2020-07-08 12-25-56
Screenshot from 2020-07-08 12-26-30

runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBox.java
molteber: Thanks for the tip :+1: 

But for now I've removed the possibility to change this on the InfoBox class directly, and made it so InfoBoxOverlay decides if it should be outlined and/or have a background.

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/BackgroundComponent.java
```diff
@@ -47,45 +47,49 @@
 	private Color backgroundColor = ComponentConstants.STANDARD_BACKGROUND_COLOR;
 	private Rectangle rectangle = new Rectangle();
 	private boolean fill = true;
+	private boolean border = true;
```
Adam-, line 50: This doesn't look used?

Adam-:
IMO the background on the text component in the infobox looks pretty bad. Can't we just change the text color or border on the text or something instead?

runelite-client/src/main/java/net/runelite/client/ui/overlay/components/BackgroundComponent.java
molteber: It's used here:
https://github.com/runelite/runelite/blob/8129b369ae4fb9d71b40842233d48a5c36c3712c/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/InfoBoxComponent.java#L102

molteber:
> IMO the background on the text component in the infobox looks pretty bad. Can't we just change the text color or border on the text or something instead?

I've tried a bit with other colors, and that can help in some situations, but not all.

The border on the text you are proposing is basically the `outline` functionality here, which helps out a bit.

I was originally just going for the outline on the text, but after conferring with a friend, he pointed that it could make the text look more blurry, whereas i then suggested to him to add a transparent background to make the color gap more convenient for the eyes .

The background on the text looks a bit, unique. Do you think it would be better to make the whole infobox be covered in that transparent background, to avoid that difference on the background, @Adam- ?

Adam-:
Yes. Though I think you could just change the text color/bordering and it would be fine.

molteber:
So, basically what you're saying/suggesting, is to remove the option of background behind the text, and just go for the option of outlined text?

Adam-:
It looks to me like the text color is actually already configurable, so why don't we just make it something less terrible then yellow by default?

molteber:
Yeah, that's the case for NMZ at least. It could probably be done, but could rather be another issue to deal with (to also potentially include other info boxes who might need a new color default).

IMO, it would be great to add something likes this outline option, to allow people to "solve it once and for all", instead of manually changing color on each of the info boxes which is troublesome.

Adam-:
What about https://github.com/Adam-/runelite/commits/nmz? It looks like this 
image

pushaton42:
Is this implemented already? If not, is there an ETA?

Thanks a lot =]

Skretzo:
I think the background on the infobox text should be revisited as this is what the stat boosts bar on the steam client looks like.
steam client example

PR #12105, opened 2020-07-10
Reduce repetitiveness in HTTP clients + Add tests

UgiR:
This PR consolidates much of the repeated logic in each client into one base class.

* Method signatures remain the same except if a client's method is async and returns void, it now returns CompletableFuture<Void>.

* Originally I wanted to move all the 'properties' logic from RuneLiteAPI into RuneLiteApiClient, but it looks like there are references to RuneLiteAPI across a lot of other modules, so I did not touch it.

* HiscoreClient was not changed. I will leave this for a separate PR. There exists an enum for all hiscore URL's, but the client also exposes a lookup method taking an arbitrary URL (it looks like maybe for testing purposes?).. I would like to fix this as well so the lookup URL's are limited to the enum URL values.. slightly more involved.

Let me know how it looks..thanks.

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/RuneLiteApiClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/RuneLiteApiClient.java
@@ -229,7 +229,7 @@ public class RuneLiteApiClient
 	{
 		Request request = requestBuilder
 			.url(urlInterceptor.apply(baseUrl.newBuilder()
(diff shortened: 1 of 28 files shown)


Nightfirecat:
The HTTP API has moved to its own repository: https://github.com/runelite/api.runelite.net. You can reopen this PR there if you feel we should still consider these changes.

PR #12118, opened 2020-07-11
LoginScreen: add timer backgrounds

Graviton1647:
Added timer backgrounds and ability to include custom backgrounds to it


Graviton1647:
Hydrox has made a patch for someone that allows you to edit the background images. This will require that patch or one like it so the login box does not get reset every single time

PR #12130, opened 2020-07-13
Check diseased and dead states on the Farming Contract Indicator

### LINKED ISSUE #12124: Time tracking plugin shows inaccurate info when farming plot contains diseased crop
**Overview of the Issue**
The issue pertains to the farming contract aspect of the Time Tracking official plugin. This plugin normally gives the status of the plot you're currently assigned to using an infobox overlay in the upper left corner of the screen. The infobox includes the estimated time until growth is complete if the correct plant is currently growing already. 

Specifically, if the plant is diseased when user logs in, it will report that the plot is ready to be harvested, even if it's not usually the time it would be finished had it not become diseased.

**Java Version and Operating System**
_(copied from About tab in client)_
Runelite version: 1.6.21.1
Oldschool revision: Rev 190
Launcher version: Unknown

**Reproduce the Error**
1. Plant seed/sapling according to current farming contract in empty plot
    a. Ideally plant a snapdragon seed after receiving a snapdragon as a contract assignment (unsure if necessary step)
2. Use ultracompost from bottomless bucket (unsure if necessary step)
3. Logout within the farming guild
4. Log back in after at least one growth stage has occurred

Expected: Infobox overlay in upper left should show that the contracted plot is diseased and must be cured (perhaps warning yellow text instead of white or green) else the crop will be lost. If not, at least show the actual time remaining as if it were not diseased.

Actual: Infobox overlay shows the name of the plant in green and status of "Ready" also in green (see screenshot below)

image

Cyborger1:
Fixes #12124 

Also fixes an issue with crops requiring health checks showing as in progress instead of occupied if they were already planted and checked before the contract.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractInfoBox.java
@@ -82,9 +82,23 @@ class FarmingContractInfoBox extends InfoBox
 				contractColor = ColorScheme.PROGRESS_ERROR_COLOR;
 				break;
 			case IN_PROGRESS:
-				contractDescription = "Ready " + TabContentPanel.getFormattedEstimate(manager.getCompletionTime() - Instant.now().getEpochSecond(),
-					config.timeFormatMode());
-				contractColor = Color.GRAY;
+				CropState cropState = manager.getContractCropState();
+				switch (cropState)
+				{
+					case DISEASED:
+						contractDescription = "Diseased";
+						contractColor = cropState.getColor();
+						break;
+					case DEAD:
+						contractDescription = "Dead";
+						contractColor = cropState.getColor();
+						break;
+					default:
+						contractDescription = "Ready " + TabContentPanel.getFormattedEstimate(manager.getCompletionTime() - Instant.now().getEpochSecond(),
+							config.timeFormatMode());
+						contractColor = Color.GRAY;
+						break;
+				}
 				break;
 			case EMPTY:
 			case UNKNOWN:
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractManager.java
@@ -257,6 +257,8 @@ public class FarmingContractManager
 					&& !hasEmptyPatch)
 				{
 					summary = SummaryState.OCCUPIED;
+					// Don't let this run into the "Completed" section!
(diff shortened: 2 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractInfoBox.java
```diff
@@ -82,9 +82,23 @@ public String getTooltip()
...
 				break;
 			case IN_PROGRESS:
-				contractDescription = "Ready " + TabContentPanel.getFormattedEstimate(manager.getCompletionTime() - Instant.now().getEpochSecond(),
-					config.estimateRelative());
-				contractColor = Color.GRAY;
+				switch (manager.getContractCropState())
+				{
+					case DISEASED:
+						contractDescription = "Diseased";
+						contractColor = Color.YELLOW;
```
melkypie, line 89: You should use the exact `CropState` color from enum instead of just defining it as `Color.YELLOW`. Same goes for L93.

Cyborger1: Ok, I have pushed the requested change.

melkypie (approved):

Nightfirecat:
It would be lovely if you could squash your work into separate commits for each fix you have included here. (One for the health check status, another for the disease/dead state check fix) As-is, it's not clear what code is needed for each of the fixes.

Nightfirecat:
This also strikes me as something which could use a test (or multiple?) to demonstrate correct behavior. Whether you want to do that yourself or seek help in adding tests, that would be really useful.

Cyborger1:
I'm not the most well versed in git commands, but I think I managed to clean this up. Also, it seems my command line was set up for my other account, so that's a bit of an oops for the force-push there.

I'm going to start looking into getting some tests in for this.

Nightfirecat:
Go ahead and add the tests to whichever commit updates the behavior they are testing.

Cyborger1:
I added the tests, although I am not sure if I did these correctly. I had to Mock/Bind lots of classes before the tests would compile.

There's probably a way to simplify the tests, it would be nice if someone that knows more about how to write test classes for Runelite could take a look and help me out.

Nightfirecat:
You definitely had more stuff `@Mock @Bind`ed than necessary, but the tests were, for the most part, already pretty good at covering the changed code. I added a few to finish out full coverage of the modified FarmingContractManager code and did some other minor cleanup/rewording of the commits. imo this is looking good to merge now.

Nightfirecat (approved):

melkypie (approved):

PR #12151, opened 2020-07-15
Fix minimum overlay size resizing

Nightfirecat:
This PR fixes [a bug which would "push" overlays resized in the right/down-ward directions beyond their minimum sizes](https://cdn.discordapp.com/attachments/419891709883973642/732737803447959672/2020-07-14_19-18-16.gif). A video demonstrating the updated behavior can be downloaded here: https://cdn.discordapp.com/attachments/419891709883973642/732809556131840081/simplescreenrecorder-2020-07-14_21.02.28.webm

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -460,53 +460,83 @@ public class OverlayRenderer extends MouseAdapter implements KeyListener
 
 		if (inOverlayResizingMode)
 		{
-			final Rectangle r = currentManagedBounds;
+			final int left = p.x - currentManagedBounds.x; // Distance (in pixels) from the left edge of the bounds
+			final int top = p.y - currentManagedBounds.y;  // Distance (in pixels) from the top edge of the bounds
+			final int originalX = currentManagedBounds.x;
+			final int originalY = currentManagedBounds.y;
+			int x = originalX;
+			int y = originalY;
+			int width = currentManagedBounds.width;
+			int height = currentManagedBounds.height;
 
-			final int dx = p.x - r.x;
-			final int dy = p.y - r.y;
 			switch (clientUI.getCurrentCursor().getType())
 			{
 				case Cursor.N_RESIZE_CURSOR:
-					int height = r.height - dy;
-					r.setRect(r.x, r.y + dy, r.width, height);
+					y += top;
+					height -= top;
 					break;
 				case Cursor.NW_RESIZE_CURSOR:
-					int width = r.width - dx;
-					height = r.height - dy;
-					r.setRect(r.x + dx, r.y + dy, width, height);
+					x += left;
+					y += top;
+					width -= left;
+					height -= top;
 					break;
 				case Cursor.W_RESIZE_CURSOR:
-					width = r.width - dx;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
```diff
@@ -216,6 +216,11 @@ public void finishCreation(boolean aborted)
 		{
 			final ScreenMarkerOverlay screenMarkerOverlay = new ScreenMarkerOverlay(marker);
 			screenMarkerOverlay.setPreferredLocation(overlay.getBounds().getLocation());
+
+			// Ensure created overlay is of at least minimum width/height
+			final Rectangle markerBounds = overlay.getBounds();
```
deathbeam, line 221: This should just be checked in Overlay class (for example override setPrefferedSize there and check it also there)

runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerOverlay.java
```diff
@@ -49,6 +50,7 @@
 		setLayer(OverlayLayer.ALWAYS_ON_TOP);
 		setPriority(OverlayPriority.HIGH);
 		setResizable(true);
+		setMinimumSize(new Dimension(OverlayRenderer.OVERLAY_RESIZE_TOLERANCE * 2 + 1, OverlayRenderer.OVERLAY_RESIZE_TOLERANCE * 2 + 1));
```
deathbeam, line 53: Why is allowing smaller minimum size useful?

Nightfirecat: This particular change is in response to a user finding that resizing a marker they placed over a spell widget, which is notably smaller than 32x32, caused it to grow much larger and they could not resize it back to its original size.

deathbeam: I see, makes sense

runelite-client/src/main/java/net/runelite/client/plugins/screenmarkers/ScreenMarkerPlugin.java
Nightfirecat: I agree this is a better approach. For the time being, I'm having some trouble with some edge cases around the screen marker plugin specifically, so I'm removing these changes from this PR while I adjust it so the resizing logic fix can be considered on its own.

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -460,53 +460,83 @@ public MouseEvent mouseDragged(MouseEvent mouseEvent)
...
 
-			final int dx = p.x - r.x;
-			final int dy = p.y - r.y;
 			switch (clientUI.getCurrentCursor().getType())
 			{
 				case Cursor.N_RESIZE_CURSOR:
-					int height = r.height - dy;
-					r.setRect(r.x, r.y + dy, r.width, height);
+					y += top;
+					height -=  top;
```
deathbeam, line 476: Double space

PR #12173, opened 2020-07-17
Adding config option to support CTRL + and CTRL - for zooming

### LINKED ISSUE #12104: Super Simple Request which would make me permanently switch from OSBUDDY
I would like to zoom in and out using the CONTROL and + or -  symbols. I don't have a mouse with a mousewheel and I'm unable to zoom in and out during gameplay, a feature which I used frequently when gaming with OSBuddy.

"CTRL" and "+" zooms the camera toward the character.
"CTRL" and "-" zooms the camera away from the character.

I would have to buy a different mouse which has a mouse-wheel - regular mouses cause inflammation in my wrists and detract from my enjoyment of the game. Otherwise, I'll have to settle for switching back to OSBuddy, despite how much I love and appreciate RuneLite.

**Additional context**
I just want to zoom in, man.

zeffirojoe:
Closes #12104

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
@@ -79,7 +79,10 @@ public class CameraPlugin extends Plugin implements KeyListener, MouseListener
 	 * Whether or not the current menu has any non-ignored menu entries
 	 */
 	private boolean menuHasEntries;
-	
+
+	//Integer used to keep track of zoom value
+	private int currentZoom;
+
 	@Inject
 	private Client client;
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/camera/CameraPlugin.java
```diff
@@ -79,7 +79,10 @@
 	 * Whether or not the current menu has any non-ignored menu entries
 	 */
 	private boolean menuHasEntries;
-	
+
+	//Integer used to keep track of zoom value
+	private int currentZoom;
```
Adam-, line 84: Why are you doing this instead of just fetching the zoom value in `keyPressed`? This seems like this would not work if you zoomed with the mouse in combination with the keys.

PR #12184, opened 2020-07-18
cooking: fix sinew not registering as cooked

### LINKED ISSUE #12090: Successfully cooked sinew not registered as cooked
**Describe the bug**
While cooking raw beef (and I would assume raw bear meat as well) into sinew, successful cooks do not count in counter to the top left. Any burns that happen will be tracked. This leaves the cooked value as 0 and the burns to go up, with burn percentage staying at 100.
**To Reproduce**
1. Use raw beef on a range
2. Select "Sinew"
3. Successfully cook at least 1

**Expected behavior**
The amount cooked should represent the amount of sinew you cooked. Then the proper burn percentage could be found.

**Screenshots**
I was cooking sinew before and the counter was at 0, but after cooking one meat normally, the counter went up. It did not go up after cooking more sinew.

Cooking(23) 2020-07-09_01-12-16

 - OS: Windows 10
 - RuneLite version: 1.6.21.1
 - Launcher version: 2.1.3

**Additional context**
I was at the Hosidius kitchen, but I would assume this problem persists at all ranges.

jualkaup:
Successfully cooking raw beef into sinew didn't count in the cooking
counter. Added the chat message for successfully cooking sinew to
CookingPlugin.

Close #12090

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cooking/CookingPlugin.java
@@ -176,7 +176,8 @@ public class CookingPlugin extends Plugin
 			|| message.startsWith("You successfully bake")
 			|| message.startsWith("You manage to cook")
 			|| message.startsWith("You roast a")
-			|| message.startsWith("You cook"))
+			|| message.startsWith("You cook")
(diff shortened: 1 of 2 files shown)


jualkaup:
> 
> 
> This PR should add the full sinew message to the plugin's test.

Added. Amended the test changes to the previous commit.

PR #12194, opened 2020-07-18
flash-notifier: Add custom time and color options

bryangreener:
Added the following features to the flash notification section in Runelite settings:
- Custom flash notification duration (in milliseconds)
- Custom flash notification color using color picker with static alpha channel value of 70.


bryangreener:
I somehow goofed up my branch. Closing until resolved.

Adam-:
A custom time option seems fine, but we already have a custom color option, and it looks like you are removing it and replacing it with one which doesn't have a configurable alpha, which seems worse. Why is that?

bryangreener:
> A custom time option seems fine, but we already have a custom color option, and it looks like you are removing it and replacing it with one which doesn't have a configurable alpha, which seems worse. Why is that?

I actually didn't see where there was a color option for the flash. How long ago was the color selection added?

Adam-:
It was added in 05c4a5c1d5e1cc64672fead0c223f47cfcae1ff1

bryangreener:
> It was added in [05c4a5c](https://github.com/runelite/runelite/commit/05c4a5c1d5e1cc64672fead0c223f47cfcae1ff1)

Ah yeah I wrote mine before that came out. That would explain why I hadn't seen it.
How should I proceed. Do I need to remove the Color Picker code then push back up?

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -217,19 +215,22 @@ public void processFlash(final Graphics2D graphics)
...
+				case FLASH_FOR_DURATION:
+				case SOLID_FOR_DURATION:
 					flashStart = null;
 					return;
 				case SOLID_UNTIL_CANCELLED:
 				case FLASH_UNTIL_CANCELLED:
 					// Any interaction with the client since the notification started will cancel it after the minimum duration
-					if ((client.getMouseIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
-						|| client.getKeyboardIdleTicks() < MINIMUM_FLASH_DURATION_TICKS
+					if ((client.getMouseIdleTicks() < minimum_flash_duration_ticks
```
Adam-, line 232: This currently is using MINIMUM_FLASH_DURATION_MILLIS for two different things, the flash time and the minimum time before activity can cancel a notif (this is so notifs don't get triggered and then cancelled prior to you being able to see them). Since you are replacing both things with 1 new configuration value it is strange since now the min cancel time for flash_until_cancelled is actually the "flash notification duration"

It would probably make more sense to have the config be a simple off/flash/solid and then have a 2nd option for time, and then use time = 0 as until cancelled.

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -244,11 +244,24 @@ default FlashNotification flashNotification()
...
+	{
+		return 2000;
+	}
+
 	@ConfigItem(
 		keyName = "notificationFocused",
 		name = "Send notifications when focused",
 		description = "Toggles all notifications for when the client is focused",
-		position = 25,
+		position = 27,
```
Adam-, line 264: You're reordering the existing focused/flash color options here. Instead this should be 26 and the next one 27.

runelite-client/src/main/java/net/runelite/client/config/FlashNotification.java
```diff
@@ -32,8 +32,8 @@
 public enum FlashNotification
 {
 	DISABLED("Off"),
-	FLASH_TWO_SECONDS("Flash for 2 seconds"),
-	SOLID_TWO_SECONDS("Solid for 2 seconds"),
+	FLASH_FOR_DURATION("Flash for specified duration"),
+	SOLID_FOR_DURATION("Solid for specified duration"),
```
Adam-, line 36: Additionally this needs to be improved (though may be irrelevant with a simple off/flash/solid)

image

bryangreener: Thank you for the feedback. I'll work on making these changes. It might take a while since I need to set my entire environment back up after getting a new computer but I'll let you know when I have an update.

bryangreener:
Alright Adam I've made some changes following your suggestions. Let me know if these changes fit what you're looking for but I've tested it and I couldn't find any issues.

Adam-:
This looks like you've rebased incorrectly since it includes 129 commits.

bryangreener:
> This looks like you've rebased incorrectly since it includes 129 commits.

Lol idk how im screwing this up every single time. Here is my work flow.

```
git checkout my-branch
git fetch upstream
git rebase upstream/master
```
Make changes to files.
```
git add Notifier.java RuneLiteConfig.java FlashNotifierConfig.java
git commit -m 'blah'
```
If there is any conflict when pushing I do
```
git pull
```
fix merge conflicts in files then finish pull and add those files again and attempt to push with
```
git push origin my-branch
```

bryangreener:
Okay I think I've resolved it by using the following to clean up the branch.
```
git fetch upstream
get reset --hard upstream/master
git cherry-pick <hash I want to keep>
git push -f origin my-branch
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -140,7 +142,7 @@ private Notifier(
 		// First check if we are running in launcher
 		this.terminalNotifierAvailable =
 			!Strings.isNullOrEmpty(RuneLiteProperties.getLauncherVersion())
-			&& isTerminalNotifierAvailable();
+				&& isTerminalNotifierAvailable();
```
Nightfirecat, line 145: Revert this indentation change.
```suggestion
			&& isTerminalNotifierAvailable();
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -217,33 +219,52 @@ public void processFlash(final Graphics2D graphics)
 			return;
 		}
 
-		if (Instant.now().minusMillis(MINIMUM_FLASH_DURATION_MILLIS).isAfter(flashStart))
+		final int flash_notification_duration_config_val = runeLiteConfig.flashNotificationDuration();
+		final int flash_duration_millis;
```
Nightfirecat, line 223: Our code convention is to use camelCase for variables. These vars can be much more brief anyway.
```suggestion
		final int configFlashDurationMillis = runeLiteConfig.flashNotificationDuration();
		final int flashDurationMillis;
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -217,33 +219,52 @@ public void processFlash(final Graphics2D graphics)
...
+		}
+		else if (flash_notification_duration_config_val < MINIMUM_FLASH_DURATION_MILLIS)
+		{
+			// If less than minimum duration but not set to 0 then set to minimum.
+			flash_duration_millis = MINIMUM_FLASH_DURATION_MILLIS;
+		}
+		else
+		{
+			flash_duration_millis = flash_notification_duration_config_val;
+		}
```
Nightfirecat, line 238: This is just `else { flash_duration_millis = Math.min(MINIMUM_FLASH_DURATION_MILLIS, flash_notification_duration_config_val); }`
```suggestion
		else
		{
			flash_duration_millis = Math.min(MINIMUM_FLASH_DURATION_MILLIS, flash_notification_duration_config_val);
		}
```

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -217,33 +219,52 @@ public void processFlash(final Graphics2D graphics)
...
+
+		if (flash_duration_millis == 0 || Instant.now().minusMillis(flash_duration_millis).isAfter(flashStart))
 		{
 			switch (flashNotification)
```
Nightfirecat, line 242: This control flow no longer makes sense here because we're guaranteed to run the same code for any value of `flashNotification` which makes the `switch` is completely useless, and the `duration == 0` check is executed multiple times every iteration for no reason.

This can instead be:

```java
if (flash_duration_millis == 0)
{
	// interaction check code here
}
else if (Instant.now().minusMillis(flash_duration_millis).isAfter(flashStart))
{
	flashStart = null;
	return;
}
```

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -244,11 +244,24 @@ default FlashNotification flashNotification()
 		return FlashNotification.DISABLED;
 	}
 
+	@ConfigItem(
+		keyName = "flashNotificationDuration",
+		name = "Notification duration",
+		description = "Duration of flash notification. Value of 0 will force notification to wait until cancelled",
```
Nightfirecat, line 250: This controls the duration of both flash and solid notification types.
```suggestion
		description = "Duration of notification overlay. A value of 0 will force notification to wait until cancelled",
```

Nightfirecat (changes requested):
Needs rebase also

PR #12201, opened 2020-07-22
Notifier: Disable flash notification when off is selected

dereklopes:
I didn't see a bug reported for this so I decided to make a fix it myself.

When the "Flash Notification" setting is set to "Flash until cancelled" or "Solid until cancelled" and the notification goes off, changing the setting to "Off" keeps the notification solid even if you move your mouse or click in the game window panel.

This fix simply disables any flash notification currently active if the setting is changed to off.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -208,13 +208,14 @@ public class Notifier
 
 	public void processFlash(final Graphics2D graphics)
 	{
-		if (flashStart == null || client.getGameState() != GameState.LOGGED_IN)
+		FlashNotification flashNotification = runeLiteConfig.flashNotification();
+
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -246,6 +246,13 @@ public void processFlash(final Graphics2D graphics)
 			return;
 		}
 
+		if (runeLiteConfig.flashNotification() == FlashNotification.DISABLED)
```
Adam-, line 249: You can just move this up to L211 + move up L217

dereklopes: Good call, updated

dereklopes:
@Adam- is there anything else I should do for this PR to be merged?

Adam-:
It would be nice if the history was squashed and the commit message was updated to include a description of why the change is necessary/what it is fixing.

PR #12213, opened 2020-07-23
EmoteClue: Update location of clues from Digsite to Exam Centre

MarbleTurtle:
Changes location of "Clap in the main exam room in the Exam Centre. Equip a white apron, green gnome boots and leather gloves." easy clue and "Headbang at the exam center. Beware of double agents! Equip a mystic fire staff, a diamond bracelet and rune boots." hard clue to Exam Centre from Digsite to more accurately describe their locations.


makingcomment:
The Digsite is more noteworthy than its Exam Center, so maybe "Digsite Exam Centre" would be better?

PR #12222, opened 2020-07-24
Searchable VarInspector

alowaniak:
Adds a search input in the VarInspector.
searchable-var-inspector


abextm:
you should just add the search to the options pane rather than trying to float it over the content pane

alowaniak:
> you should just add the search to the options pane rather than trying to float it over the content pane

Moved the search input to the options pane, now looks like:
image

Issue #12196, opened 2020-07-20
RS3 style examine


AildirMike:
Possibility to add a hovering textbox with the examine text over, around or near the object that is being examined? Sort of like in RS3?
![Unicorns](https://user-images.githubusercontent.com/51796970/87934121-b9d2fb00-ca8e-11ea-968c-7a53423e4a82.png)

Thanks in advance!

mrboots123:
#12233

Hydrox6:
Whoops, this shouldn't have been closed by plugin-hub#500

mrboots123:
#12248

mrboots123:
#12250

PR #12233, opened 2020-07-27
feat(examineHover): examine tooltip hover

mrboots123:
Add ability to see a tooltip on examine text of objects when hovering

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/examine/ExamineClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/examine/ExamineClient.java
@@ -45,6 +45,11 @@ public class ExamineClient
 
 	private final OkHttpClient client;
 
+	public String getObject(int id)
+	{
+		return get("object",id);
+	}
+
 	public void submitObject(int id, String text)
 	{
 		submit("object", id, text);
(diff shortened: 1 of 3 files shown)


mrboots123:
Screen Shot 2020-07-26 at 7 31 49 PM
Screen Shot 2020-07-26 at 7 32 19 PM

Screen Shot 2020-07-26 at 7 30 20 PM

Adam-:
I think this approach would send too much traffic at us that we would either be unable to handle it or unwilling to handle it. If we really wanted this feature I think we should work it into the wiki scraper and then fetch it down like the other things like it. Also the current examine API anyone can submit examines to, and once we start using it to show examines publicly to people, bad guys will send bad stuff to it, and the wiki is generally good at content moderation.


Additionally you are doing blocking http requests on the client thread which is incorrect and will cause noticeable client freezes whenever you mouse over anything not in your cache.

makingcomment:
You referenced this in #12196, but this doesn't do what #12196 asks. The RS3 examine only brings up the tooltip when an object is actually examined. If you mimicked that approach, you could take the examine info given in the game's chat and make a tooltip out of that, which would not only more closely resemble RS3 but also avoid the problem Adam pointed out.

PR #12237, opened 2020-07-27
Add bird nest to loot tracker plugin.

SirGirion:
This will track seed bird nests in the loot tracker plugin and ultimately get the data sent to the wiki to further improve their data sets. Inspired from JihadSquad's sheet where he asks people to send him seed data to improve his rates.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -188,7 +188,7 @@ public class LootTrackerPlugin extends Plugin
 
 	private static final Pattern PICKPOCKET_REGEX = Pattern.compile("You pick (the )?(?<target>.+)'s? pocket.*");
 
-	private static final Pattern BIRDNEST_REGEX = Pattern.compile("You take (a|an) (.+) out of the bird's nest\\.");
+	private static final Pattern BIRDNEST_REGEX = Pattern.compile("You take an? (.+) out of the bird's nest\\.");
 	private static final String BIRDNEST_EVENT = "Bird nest";
 
 	/*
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -188,6 +188,9 @@
 
 	private static final Pattern PICKPOCKET_REGEX = Pattern.compile("You pick (the )?(?<target>.+)'s? pocket.*");
 
+	private static final Pattern BIRDNEST_REGEX = Pattern.compile("You take (a|an) (.+) out of the bird's nest\\.");
```
SirGirion, line 191: Mostly took inspiration from the pickpocket regex above w/ the exception of not having a named capture group since we don't care about the seed from the message itself due to the snapshot function handling that for us.

Nightfirecat, line 191: Don't capture `an?`
```suggestion
	private static final Pattern BIRDNEST_REGEX = Pattern.compile("You take an? (.+) out of the bird's nest\\.");
```

SirGirion: Fixed

Nightfirecat:
This change is not working, it appears the chat message is sent after the inventory change occurs.

SirGirion:
Do you mean with my original regex or your suggested regex? I've been using the plugin for a few days now (90 nests) and have had no issues with it.

SirGirion:
Whoops, forgot to add the event type to the onItemContainerChanged event, this change now works

Nightfirecat:
This works fine for seed nests, but won't work properly for clue or ring nests. I'd recommend taking a similar approach to how we handle seed pack loot tracking--capture inventory snapshots when the nests are clicked instead of when the chat message is received.

SirGirion:
Made this change, works on jewelry nests now too. Haven't had the RNG to get a clue nest to fully test those but operating on the assumption that the IDs are correct from ItemID and it will work for those too.

SirGirion:
> This looks better, though there are a couple things that should be changed:
> 
> 1. Clue nests use an "Open" menu option, not a "Search" menu option
> 2. This doesn't include all bird nest IDs. [Searching "nest" in the item database](https://www.osrsbox.com/tools/item-search/) can help you identify more nests. Notably missing IDs include 5070-5073, 13653, and 22800

Would it be better to have a different set for the nests that use Open rather than search or is it fine to keep them in one?

Re: missing IDs, admittedly I made this with the idea of tracking the nests from birdhouses which I thought were considered different than the regular nests.

Issue #12125, opened 2020-07-13
Agility rooftop course laps per hour inconsistencies


AndreasBoegh:
**Describe the bug**
Laps per hour is inconsistent to the true laps per hour see screenshot for example

**Expected behavior**
Doing my example of 1 hour of laps on agility course for example 1 hour doing 64 laps. Should yield a 64 laps per hour estimate

**Screenshots**
![Screenshot_2](https://user-images.githubusercontent.com/32385164/87293313-6991f100-c502-11ea-8ad6-51c85d085097.png)

Cyborger1:
The agility plugin estimates laps per hour from the 10 previous lap completions. The discrepancy would be from slowing down slightly for the last 10 laps of the hour.

I'm not sure this counts as a bug, but the behaviour could be communicated to the player.

Nightfirecat:
10 laps seems like a very small amount considering it's cheap to increase that number and prevent people from thinking the estimation is as wrong as it is sometimes.

PR #12246, opened 2020-07-29
agility plugin: increase number of laps for laps per hour calculation

arafian:
This commit increases the number of laps used to calculate laps per hour
from 10 to 30.
Resolves: #12125

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilitySession.java
@@ -41,7 +41,7 @@ class AgilitySession
 	private Instant lastLapCompleted;
 	private int totalLaps;
 	private int lapsTillGoal;
-	private final EvictingQueue<Duration> lastLapTimes = EvictingQueue.create(10);
+	private final EvictingQueue<Duration> lastLapTimes = EvictingQueue.create(30);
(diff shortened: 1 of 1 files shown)


dumbitdownjr:
The number of laps used is pretty subjective. I tested 10 (the original), 30, and 50 and found 30 laps to be a good middle ground between taking a good amount of past laps into account so that one or two outlier laps won't skew the laps per hour, and including so many laps that genuine increases or decreases in lap speed will be factored in better.

makingcomment:
I'm sure everyone has different opinions on what the ideal tradeoff between recent and overall performance should be. Making this a config would allow it to better cater to people's preferences.

arafian:
Good point. I think it would be nice to have a config too, although I'm not sure to what extent there should be a config for every little thing. But I would like to figure out how to make one if everyone thinks this a good thing make a config for.

I do think it was a good idea to merge as is because the default really shouldn't be 10 laps. A better default might be based on the average laps per hour for each agility course, but maybe that's overkill, and I'm not sure if it's possible to detect which agility course the player is on anyway.

Nightfirecat:
I don't think a config is really valuable. Setting a sane default like this PR does is already a big improvement, and I don't anticipate users caring to adjust that default--much like how the XP tracker estimates actions needed to level based on the past 10 actions, this shouldn't need to be configurable.

PR #12265, opened 2020-08-01
achivement dairy plugin: add satisfiesRequirement to Requirement

JZomerlei:
Add satisfiesRequirement to Requirement interface that all other requirements implement Instead of using instanceOf to see if each requirement is satisfied.


runelite-client/src/main/java/net/runelite/client/plugins/achievementdiary/CombatLevelRequirement.java
```diff
@@ -38,4 +39,16 @@ public String toString()
 	{
 		return level + " " + "Combat";
 	}
+
+	public boolean satisfiesRequirement(Client client)
+	{
+		try
+		{
+			return client.getLocalPlayer().getCombatLevel() >= level;
```
deathbeam, line 47: This should simply null-check localPlayer instead of catching NPE

PR #12279, opened 2020-08-03
Hiscore Lookup - Cancel HTTP request when clearing input box

TheStonedTurtle:
Currently when you clear the input box the HTTP request will continue to run, making you wait for it to finish/timeout before allowing another request. Since the Normal Hiscores consistently has problems this issue is pretty common, especially when making a typo or forgetting to change to a different endpoint.

This PR fixes that by canceling the HTTP request when the input box is cleared.

52WsVf1qrZ


deathbeam:
Should there really be the failed icon when you cancel the request?

TheStonedTurtle:
> 
> 
> Should there really be the failed icon when you cancel the request?

The `lookup` method will catch the IOException from closing the socket/canceling the call, which is on another thread and overrides any icon change made inside the clear listener. The only way I see to do this would be checking the message on the error, do you have any better ideas?

PR #12284, opened 2020-08-03
Add Ring of Endurance mapping

MeeranB:
Maps the untradeable charged version of the ring of endurance to the uncharged version


Nightfirecat:
Added in a57721b41a0af670eb0a4ccfec072cf00bbbe0de

PR #12306, opened 2020-08-07
Adding option to quick hop between favorites

### LINKED ISSUE #12238: Quick-hop option to hop between multiple customized worlds
Have a way to customize the quick-hop hotkey (or another hotkey all together) to hop between customized world list in sequence instead of the next closest world number. 

When you need to hop between a few selected worlds, for example bossing while waiting for long death timers, you need to hop to worlds that are available at the time.

Also since this would most likely be used multiple client alting as well there should be a way to copy paste text chunk into each alt client into the plugin settings to update the new world list for ease of use.  

something like:

312
366
477
480

zeffirojoe:
Closes #12238

Adding config option to quick-hop between worlds marked as "favorite" instead of incrementing or decrementing by 1 for quick hop


Adam-:
Any reason you didn't put the filter in with the other ones like the region filter? It would be ~4 lines instead.

PR #12311, opened 2020-08-08
Barrows prayer infobox fix

### LINKED ISSUE #10902: Barrows plug in Timer
**Describe the bug**
Barrows plug in displays multiple prayer timers that are all different ticks.

**To Reproduce**
Not sure exactly how to reproduce but,
1. activate barrows plug with timer
2. Do some barrows like normal
3. Either disconnect from inactivity or log out and back in. (normally I am still at barrows site when I log back in.)
4. reload client to cancel all timers.

**Expected behavior**
I should only have 1 prayer timer that is accurate

**Screenshots**
https://imgur.com/gallery/Bk2grru    -- screen shot of how many timers I have at the moment.

 - OS: windows 10
 - RuneLite version: 1.6.8
 - Launcher version: 2.1.0

**Additional context**

Alexsuperfly:
I believe the problem is because the checking for when to add/remove the prayer infobox can only be done on the single gamestate change of going into or out of the crypt.  This can lead to some sort of condition where it somehow did not satisfy the requirement to remove an infobox when it should have and then never can until another one is added, causing the reference to the original infobox to be lost.  

The first commit checks that when adding a timer, if one is already present, to remove it before adding a new one to prevent loss of references.

The second commit changes the logic of determining when to add or remove an infobox to every game state change and only caring about where the player currently is and the current status of the prayer infobox.  

I believe with this second commit the first one becomes obsolete but i left both of them in there so that if you wish to not mess with the current logic i can just drop the second commit and still have the fix for lost references in the PR

this should fix #10902

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
@@ -77,7 +77,6 @@ public class BarrowsPlugin extends Plugin
 	private static final int CRYPT_REGION_ID = 14231;
 
 	private LoopTimer barrowsPrayerDrainTimer;
-	private boolean wasInCrypt = false;
 
 	@Getter
 	private Widget puzzleAnswer;
@@ -128,7 +127,6 @@ public class BarrowsPlugin extends Plugin
 		overlayManager.remove(barrowsOverlay);
 		overlayManager.remove(brotherOverlay);
 		puzzleAnswer = null;
-		wasInCrypt = false;
 		stopPrayerDrainTimer();
 
 		// Restore widgets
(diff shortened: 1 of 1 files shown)


greg-wolff:
plz merge this, i have 12 prayer timers on my screen all day

Nightfirecat:
@greg-wolff Could you join the `#support` channel of our Discord server? We might want your help to test for the root cause of this issue then--the main reason we haven't merged a fix for this bug is that we haven't clearly identified its source.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsPlugin.java
```diff
@@ -233,6 +227,10 @@ private void startPrayerDrainTimer()
 	{
 		if (config.showPrayerDrainTimer())
 		{
+			if (barrowsPrayerDrainTimer != null)
```
Adam-, line 230: Why is this added? The only place which calls this already tests that this is null. You could maybe assert this if you really wanted.

Alexsuperfly: I explained in the description, that in case you didn't want to change the logic that determined when you start or stop the timer, that this would also prevent the boxes from stacking up.  So I made each "solution" a separate commit to easily drop whichever one you'd least prefer.

Adam-:
added in c96c8b979e327ae1d9d79d9c4198141717223204

PR #12329, opened 2020-08-12
Update ImageCapture.java

Macweese:


Adam-:
Why do you want this changed? I know some players are named `#PlayerXXX` but `#` is valid in most file systems so this seems fine?

Macweese:
While I understand the pound symbol is supported, it is not the reason for the change.

This change aims to better organize the way screenshots are managed. In particular to accounts who have do not have a proper name set, which to my understanding can occur in two ways, either deliberately by the user or as default by Jagex when creating an account (upon spawning on tutorial island).

Many players deliberately choose to scrub themselves of a player name, the main reason is usually to avoid being searched on the hiscores, who after succeding are assigned a temporary name each time they log in. The same way as new accounts are assigned once spawned in on tutorial island.

As a result, if a screenshot is taken it is placed in the temporary name directory, and if none exists creates one. Which potentially leads to hundreds or thousands of directories. Navigating them in general becomes near impossible. Additionally, this would remedy the same scenario if a user takes screenshots whilst creating an account on tutorial island, I'm not sure why someone would but with this it'd at least cram it into one place.

This resolves that and places all the images in a single folder. Of course however if you were to have multiple accounts without a set name they would end up mixing together inside the folder and you'd not be able to tell which belongs to which but I believe it's an improvement to the current way the plugin organizes the images.

I really can't think of any downsides to having this implemented, 'normal' players aren't affected. It just helps organize things, the code speaks for itself.

------
### Examples

Here's the result from hopping worlds and taking one screenshot then hopping again
<details>
  <summary>Current</summary>
  
</details>

<details>
  <summary>Proposed change</summary>
  
</details>

makingcomment:
> I really can't think of any downsides to having this implemented, 'normal' players aren't affected. It just helps organize things, the code speaks for itself.

I don't have an opinion either way, but I just want to point out that you gave a downside yourself in the previous paragraph:

> Of course however if you were to have multiple accounts without a set name they would end up mixing together inside the folder and you'd not be able to tell which belongs to which

Macweese:
@makingcomment Fair enough it's a "downside" to have all the images in one folder which sole purpose is to house images of accounts without a username. I feel that it's still an improvement and more appropriate to have e.g. 500 images in one place rather than 500 folders containing one image each. 

It really is strange having to vindicate this PR somehow. I'd imagine that anyone looking at this would surely agree it's a logical decision?

Adam-:
You certainly don't have to "vindicate" this, but you sent a PR with no commit description nor PR description, and I don't think it is out of line for me to inquire what it is you are trying to do & why you are trying to do it. Now you've explained that, thanks.

Macweese:
Man, it's like 30 characters and 2 lines of code and like I said above it speaks for itself.

I was lazy and just cba because the title and description of the PR would take longer to write than the code was but alright in hindsight I should have.

PR #12330, opened 2020-08-12
timetracking: add configuration for displaying time in 12 hour format

### LINKED ISSUE #4896: Change farming timer to AM/PM

Hydrox6:
Closes #4896

Closes #4959
Closes #7449
Closes #7612

I've always assumed that the locale of the OS set the time format before, but I can't find evidence of that. Even if that was the case, relying on the OS screws over those users who want to go against the norm for their locale.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
@@ -39,14 +39,14 @@ public interface TimeTrackingConfig extends Config
 	String STOPWATCHES = "stopwatches";
 
 	@ConfigItem(
-		keyName = "estimateRelative",
-		name = "Show relative time",
-		description = "Show amount of time remaining instead of completion time",
+		keyName = "timeFormatMode",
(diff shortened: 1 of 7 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -39,14 +39,14 @@
 	String STOPWATCHES = "stopwatches";
 
 	@ConfigItem(
-		keyName = "estimateRelative",
-		name = "Show relative time",
-		description = "Show amount of time remaining instead of completion time",
+		keyName = "timeFormatMode",
+		name = "Display time as",
```
Adam-, line 43: A better name would be `Time format`

Hydrox6: That's much better. Honestly had no idea what to call this.

Nightfirecat (approved):

PR #12336, opened 2020-08-14
Clues: Support Trouver items

MarbleTurtle:
Changes Falo and Emote steps to support Trouver versions of items


Nightfirecat:
Do STASH units also support items which have had trouver's parchments used on them? If so, this should also include fire and infernal capes for the Tzhaar area master emote clue.

PR #11052, opened 2020-03-22
Nmz Plugin - Fix area check to exclude KBD lair

TheStonedTurtle:
Fixes #3692

PR #12337, opened 2020-08-14
grounditems plugin: fix despawn timer for private KBD instances

### LINKED ISSUE #12332: Ground Items: Despawn timer is inaccurate in King Black Dragon instance
**Describe the bug**
The Ground Items plugin adjusts item despawn timers when you are in a private boss instance. However, the King Black Dragon private instance works differently from all other private instances -- items despawn at the same rate as the rest of the game. The plugin inaccurately displays a 30-minute despawn timer when it should use the standard timer.

**To Reproduce**
1. Enable the Ground Items plugin and its "Show despawn timers" option
2. Enter the King Black Dragon private instance
3. Defeat the King Black Dragon or drop any item on the ground

**Expected behavior**
The displayed despawn timer should use the standard 2-minute timer.

**Actual results**
The displayed despawn timer uses the 30-minute private instance timer.

**Screenshots**
image

arafian:
Resolves #12332

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -395,7 +395,14 @@ public class GroundItemsOverlay extends Overlay
 			// Items in the KBD instance use the standard despawn timer
 			else if (isInKBD())
 			{
-				despawnTime = spawnTime.plus(DESPAWN_TIME_LOOT);
+				if (groundItem.getLootType() == LootType.DROPPED)
+				{
+					despawnTime = spawnTime.plus(DESPAWN_TIME_DROP);
+				}
+				else
+				{
+					despawnTime = spawnTime.plus(DESPAWN_TIME_LOOT);
+				}
 			}
 			else
 			{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -419,7 +419,6 @@ public class GroundItemsOverlay extends Overlay
 						despawnTime = spawnTime.plus(DESPAWN_TIME_LOOT);
 					}
 				}
-
 			}
 			else
 			{
@@ -515,5 +514,4 @@ public class GroundItemsOverlay extends Overlay
 	{
 		return ArrayUtils.contains(client.getMapRegions(), KBD_NMZ_REGION);
 	}
-
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -78,6 +78,7 @@ public class GroundItemsOverlay extends Overlay
 	private static final Duration DESPAWN_TIME_LOOT = Duration.ofMinutes(2);
 	private static final Duration DESPAWN_TIME_DROP = Duration.ofMinutes(3);
 	private static final int KRAKEN_REGION = 9116;
+	private static final int KBD_REGION = 9033;
 
 	private final Client client;
 	private final GroundItemsPlugin plugin;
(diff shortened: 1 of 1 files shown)


makingcomment:
Do items dropped in NMZ despawn after the same amount of time as items dropped in these KBD private instances? I can't check, but if not, would something like what was done in #11052 (but in reverse) be needed?

arafian:
I think this is a little different because this isn't about items dropped in the KBD instance, but rather how fast the loot from KBD drops. To my knowledge, there are no monster drops in NMZ, so I don't think that's an issue there.

makingcomment:
> I think this is a little different because this isn't about items dropped in the KBD instance, but rather how fast the loot from KBD drops. 

But doesn't this change affect the despawn timers of items dropped in the private KBD instance by KBD _and_ the player?

> To my knowledge, there are no monster drops in NMZ, so I don't think that's an issue there.

From the wiki: "all bosses will drop their 100% drops, Tangleroot drops leaves, Karamel drops Choc-ice, Dad drops Big Bones, etc..". Plus, players in NMZ can drop items themselves.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -391,8 +392,16 @@ private void drawTimerOverlay(Graphics2D graphics, int textX, int textY, GroundI
 			{
 				return;
 			}
+			// Items in the KBD instance use the standard despawn timer
+			else if (isInKBD())
+			{
+				despawnTime = spawnTime.plus(DESPAWN_TIME_LOOT);
```
honeyhoney, line 398: We need to check `LootType.DROPPED` to determine whether the despawn time should be set to `DESPAWN_TIME_LOOT` (2 minutes) or `DESPAWN_TIME_DROP` (3 minutes).

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -478,4 +487,9 @@ private boolean isInKraken()
 		return ArrayUtils.contains(client.getMapRegions(), KRAKEN_REGION);
 	}
 
+	private boolean isInKBD()
+	{
+		return ArrayUtils.contains(client.getMapRegions(), KBD_REGION);
```
honeyhoney, line 492: Put a local player plane check here (plane should be 0).

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
arafian: Thanks for the catch, for some reason I was thinking that the timers would be the same like they would be in an instance just shorter.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
arafian: Why do we do this? Should we also put the check in the IsInKraken method too, or is it a different case?

honeyhoney: NMZ uses the same map region at the KBD lair. The way we differentiate between the two is is the plane.

This is of course assuming NMZ instances use normal instance logic when it comes to dropped and looted items - definitely worth checking. If they use the same logic as a private KBD instance then of course you don't need to add a plane check but would want to rename the method to isInKBDorNMZ or something akin to that. If they use alternative logic to regular instances and private KBD instances then you'll want to add a new isInNMZ method (using the same map region but different plane check) and implement the new despawn logic similar to how you did for KBD.

arafian:
I somehow missed your response, but you're right on both counts. @makingcomment 
I'm going to put what I found the reviewed changes comments.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
arafian: Ah, I see that makes sense.

So I tested out how loot and dropped items work in NMZ and it's pretty confusing...
From what I can tell items dropped in NMZ never despawn. I waited for over an hour but they didn't despawn.
Loot in NMZ uniformly despawn after 3 minutes which is what you would expect, but the weird thing is RuneLite only seems to be detecting some of the dropped object items for timers. It doesn't show timers for any of the bones/ashes dropped by monsters, but it does show timers for some of the other dropped items such as leaves dropped by Tanglefoot and vampyre dust dropped by Dessous. However, other drops have no timers such as choc-ice dropped by Karamel and cake dropped by the Culinromancer. I didn't test every boss, but I have no idea why some items have timers and some don't. Apart from that, it should be straightforward to add another if/else for NMZ's drop/loot mechanics too. I attached a image that shows some of what I'm talking about.
image

arafian: Ok, so the reason the bones had to timers is because I was dumb and had them hidden by accident. I've updated it so dropped items in NMZ have no despawn timer (since they never despawn) and monster drops have the regular loot timer which is accurate based on my testing.

PR #12350, opened 2020-08-16
Player Indicators: Allow hiding indicators for ignored players

### LINKED ISSUE #12245: Player Indicators - Add Bot Busting Option
It would be cool to see an option added to the Player Indicators plugin which would remove the highlighted player's name once you've reported them and added to ignore. The plugin could search your ignore list and remove highlighted usernames from players near you. This would be useful for heavily-botted areas where there are stacks of bots on top of each other. It would help the playerbase report more accurately and assure us that we've gotten all of them!

pilino1234:
When active, this setting will hide all indicators related to ignored players, including name overlays in game and on the minimap, tile indicators, and right-click menu indicators.

Closes #12245


pilino1234:
Just realized that defaulting this to off might be safer for HCIMs and similar.

honeyhoney:
@raiyni raised concerns about this PR and how it is typically a feature we would stay from. This stems from concerns regarding the potential abuse of this feature and how it can act like a filter / adds conditional highlighting for HIghlighted players.

A user could add their friends that they're PKing with to their ignore list, enable Highlighting other players, and enable Hiding indicators for ignored players. This means all potential enemies are highlighted and allies are not, reducing clutter on the screen. This would allow for easier scouting and other activities in the wilderness.

How concerned we are about this issue I am not sure but I figured it was worth raising this issue on the PR directly as it's buried several hundred messages deep in Discord at this point.

Furthermore, https://github.com/runelite/runelite/pull/12354 which adds alpha transparency to player indicators allows for conditional highlighting in a more convenient manner. Friends/clan chat highlighting can be enabled and simply setting the alpha channel to 0 will make those indicators invisible and since the plugin will not also highlight your friends/clan chat with the "Highlight other players" indicator, the same effect as mentioned above is achieved.

pilino1234:
With all this in mind, I'd be okay with closing this PR unmerged, anything that can be used to get any kind of advantage probably will be. To prevent abuse of the alpha transparency in my other PR, I can go through those color settings and remove the transparency for player indicators.

vindux:
> A user could add their friends that they're PKing with to their ignore list, enable Highlighting other players, and enable Hiding indicators for ignored players. This means all potential enemies are highlighted and allies are not, reducing clutter on the screen. This would allow for easier scouting and other activities in the wilderness.


But you can already do that. Have your friends in your friends chat, disable the indicator for friends chat members and enable it for others. Exactly the same result as you just described.
Technically this would also work if you're pking with people but are not in a cc together, but that pretty much doesn't matter.

pilino1234:
> But you can already do that. Have your friends in your friends chat, disable the indicator for friends chat members and enable it for others. Exactly the same result as you just described.

No, they will be highlighted as "Other players" in that case.

vindux:
> > But you can already do that. Have your friends in your friends chat, disable the indicator for friends chat members and enable it for others. Exactly the same result as you just described.
> 
> No, they will be highlighted as "Other players" in that case.

No they won't.

As you can see, the player on my left is in my cc, but not getting highlighted.

image

pilino1234:
Huh, I got that confused with indicators for people on the friends list then. The friends chat indicators seems like a bug that should be fixed then I guess.

PR #12354, opened 2020-08-16
Add alpha channel to relevant color config options

### LINKED ISSUE #12353: Add alpha channel for configs with color options
Allow plugins that allow you to choose colors for marked tiles to include an alpha channel in the options.

As one specific example, the fishing plugin will automatically mark fishing spots, but its tile color selector does not have the `@Alpha` decorator and so no alpha value can be set.

Adding this alpha channel would match the behavior of the Ground Marker and Tile Indicator plugins.

pilino1234:
Add the ability to configure alpha for various different color config options.

Implement a new ColorJButton to preview colors with alpha in the config sidebar by overlaying them on top of the classic checkerboard pattern.
ColorUtil: Add new methods: `toAlphaHexColor` (yields a colors hex string including alpha), `colorOfAlpha` (change alpha of a given RGB color), and `luminance` (calculate luminance of a color). Also add unit tests for these methods.
OverlayUtil: Remove alpha from minimap dots and text overlays.
TextComponent: Remove alpha from text color.

Closes #12353

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
@@ -367,6 +367,7 @@ public interface RuneLiteConfig extends Config
 		return false;
 	}
 
+	@Alpha
 	@ConfigItem(
 		keyName = "overlayBackgroundColor",
 		name = "Overlay Color",
@@ -374,7 +375,6 @@ public interface RuneLiteConfig extends Config
 		position = 44,
 		section = overlaySettings
 	)
-	@Alpha
 	default Color overlayBackgroundColor()
 	{
 		return ComponentConstants.STANDARD_BACKGROUND_COLOR;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
@@ -89,6 +89,7 @@ import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.DynamicGridLayout;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.ui.PluginPanel;
+import net.runelite.client.ui.components.ColorJButton;
 import net.runelite.client.ui.components.ComboBoxListRenderer;
 import net.runelite.client.ui.components.colorpicker.ColorPickerManager;
 import net.runelite.client.ui.components.colorpicker.RuneliteColorPicker;
@@ -407,24 +408,23 @@ class ConfigPanel extends PluginPanel
 
 			if (cid.getType() == Color.class)
 			{
-				String existing = configManager.getConfiguration(cd.getGroup().value(), cid.getItem().keyName());
+				Color existing = configManager.getConfiguration(cd.getGroup().value(), cid.getItem().keyName(), Color.class);
 
-				Color existingColor;
-				JButton colorPickerBtn;
+				ColorJButton colorPickerBtn;
+
+				boolean alphaHidden = cid.getAlpha() == null;
 
 				if (existing == null)
 				{
-					existingColor = Color.BLACK;
-					colorPickerBtn = new JButton("Pick a color");
+					colorPickerBtn = new ColorJButton("Pick a color", Color.BLACK);
 				}
 				else
 				{
-					existingColor = ColorUtil.fromString(existing);
-					colorPickerBtn = new JButton(ColorUtil.toHexColor(existingColor).toUpperCase());
+					String colorHex = "#" + (alphaHidden ? ColorUtil.colorToHexCode(existing) : ColorUtil.colorToAlphaHexCode(existing)).toUpperCase();
+					colorPickerBtn = new ColorJButton(colorHex, existing);
 				}
 
 				colorPickerBtn.setFocusable(false);
-				colorPickerBtn.setBackground(existingColor);
 				colorPickerBtn.addMouseListener(new MouseAdapter()
 				{
 					@Override
@@ -432,14 +432,14 @@ class ConfigPanel extends PluginPanel
 					{
 						RuneliteColorPicker colorPicker = colorPickerManager.create(
 							SwingUtilities.windowForComponent(ConfigPanel.this),
-							colorPickerBtn.getBackground(),
+							colorPickerBtn.getColor(),
 							cid.getItem().name(),
-							cid.getAlpha() == null);
+							alphaHidden);
 						colorPicker.setLocation(getLocationOnScreen());
 						colorPicker.setOnColorChange(c ->
 						{
-							colorPickerBtn.setBackground(c);
-							colorPickerBtn.setText(ColorUtil.toHexColor(c).toUpperCase());
+							colorPickerBtn.setColor(c);
+							colorPickerBtn.setText("#" + (alphaHidden ? ColorUtil.colorToHexCode(c) : ColorUtil.colorToAlphaHexCode(c)).toUpperCase());
 						});
 						colorPicker.setOnClose(c -> changeConfiguration(colorPicker, cd, cid));
 						colorPicker.setVisible(true);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterConfig.java
@@ -25,6 +25,7 @@
 package net.runelite.client.plugins.hunter;
 
 import java.awt.Color;
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
@@ -32,6 +33,7 @@ import net.runelite.client.config.ConfigItem;
 @ConfigGroup("hunterplugin")
 public interface HunterConfig extends Config
 {
+	@Alpha
 	@ConfigItem(
 		position = 1,
 		keyName = "hexColorOpenTrap",
@@ -43,6 +45,7 @@ public interface HunterConfig extends Config
 		return Color.YELLOW;
 	}
 
+	@Alpha
 	@ConfigItem(
 		position = 2,
 		keyName = "hexColorFullTrap",
@@ -54,6 +57,7 @@ public interface HunterConfig extends Config
 		return Color.GREEN;
 	}
 
+	@Alpha
 	@ConfigItem(
 		position = 3,
 		keyName = "hexColorEmptyTrap",
@@ -65,6 +69,7 @@ public interface HunterConfig extends Config
 		return Color.RED;
 	}
 
+	@Alpha
 	@ConfigItem(
 		position = 4,
 		keyName = "hexColorTransTrap",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
@@ -144,7 +144,6 @@ public interface ItemStatConfig extends Config
 		return new Color(0x9CEE33);
 	}
 
-
 	@ConfigItem(
 		keyName = "colorBetterCapped",
 		name = "Better (Capped)",
@@ -155,6 +154,7 @@ public interface ItemStatConfig extends Config
 	{
 		return new Color(0xEEEE33);
 	}
+
 	@ConfigItem(
 		keyName = "colorNoChange",
 		name = "No change",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
@@ -50,6 +50,7 @@ import net.runelite.client.ui.overlay.Overlay;
 import net.runelite.client.ui.overlay.OverlayLayer;
 import net.runelite.client.ui.overlay.OverlayPosition;
 import net.runelite.client.ui.overlay.OverlayUtil;
+import net.runelite.client.util.ColorUtil;
 
 class PyramidPlunderOverlay extends Overlay
 {
@@ -157,8 +158,7 @@ class PyramidPlunderOverlay extends Overlay
 				}
 
 				graphics.draw(objectClickbox);
-				graphics.setColor(new Color(highlightColor.getRed(), highlightColor.getGreen(),
-					highlightColor.getBlue(), 50));
+				graphics.setColor(ColorUtil.colorWithAlpha(highlightColor, highlightColor.getAlpha() / 5));
 				graphics.fill(objectClickbox);
 			}
 		});
--- a/runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
@@ -1,16 +1,15 @@
 /*
- * Copyright (c) 2018, Psikoi <https://github.com/psikoi>
- * Copyright (c) 2018, Ron Young <https://github.com/raiyni>
+ * Copyright (c) 2020, Martin H <pilino@posteo.de>
  * All rights reserved.
  *
- *  Redistribution and use in source and binary forms, with or without
- *  modification, are permitted provided that the following conditions are met:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
  *
  * 1. Redistributions of source code must retain the above copyright notice, this
- *     list of conditions and the following disclaimer.
+ *    list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
- *     this list of conditions and the following disclaimer in the documentation
- *     and/or other materials provided with the distribution.
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
  *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
  * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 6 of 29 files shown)


pilino1234:
Just realized: Should the `@Alpha` come before the `@ConfigItem` or after?

pilino1234:
I've pushed a commit that removes the ability to have transparency for minimap colors and player indicators, since these could be used to reduce clutter in crowded areas and gain unfair advantages.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -158,7 +158,7 @@ public Dimension render(Graphics2D graphics)
 
 				graphics.draw(objectClickbox);
 				graphics.setColor(new Color(highlightColor.getRed(), highlightColor.getGreen(),
-					highlightColor.getBlue(), 50));
+					highlightColor.getBlue(), highlightColor.getAlpha() / 50));
```
Nightfirecat, line 161:
```suggestion
					highlightColor.getBlue(), highlightColor.getAlpha() / 5));
```

runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
```diff
@@ -0,0 +1,89 @@
...
+			textColor = Color.BLACK;
+		}
+
+		this.setForeground(textColor);
+	}
+
+	@Override
+	public void paint(Graphics g)
+	{
+		// Checkerboard drawing from PreviewPanel.java
```
Nightfirecat, line 73: This code should be moved to some common public class and referenced in both spots instead of copied.

pilino1234:
> Please revert all the default color changes you've made unless they're functional.

None of them are directly functional such that they would affect the appearance of the client, however IntelliJ does provide colour previews for the `Color()` constructors but not for some of the numerous other ways to declare colours. But I can revert it if the code churn outweighs the better tooling support and consistency :+1:

pilino1234:
I could squash 7e48ff2 into fb04442 to clean up the history a bit once this is approved, just doing it like this for now to keep track of the work.

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -410,36 +411,51 @@ public void focusLost(FocusEvent e)
...
 				{
 					existingColor = ColorUtil.fromString(existing);
-					colorPickerBtn = new JButton(ColorUtil.toHexColor(existingColor).toUpperCase());
+					String colorHex = ColorUtil.toHexColor(existingColor).toUpperCase();
+					if (!alphaHidden)
+					{
+						colorHex = ColorUtil.toAlphaHexColor(existingColor).toUpperCase();
+					}
+
+					colorPickerBtn = new ColorJButton(colorHex);
```
Adam-, line 432: Why is this not just `colorPickerBtn = new ColorJButton(existing);`?

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
```diff
@@ -410,36 +411,51 @@ public void focusLost(FocusEvent e)
...
-							colorPickerBtn.setBackground(c);
-							colorPickerBtn.setText(ColorUtil.toHexColor(c).toUpperCase());
+							colorPickerBtn.setColor(c);
+							if (alphaHidden)
+							{
+								colorPickerBtn.setText(ColorUtil.toHexColor(c).toUpperCase());
+							}
+							else
+							{
+								colorPickerBtn.setText(ColorUtil.toAlphaHexColor(c).toUpperCase());
```
Adam-, line 457: Internally all of the config Colors always have alpha=0xff, even when the picker doesn't have alpha enabled, so I think this can just be `colorPickerBtn.setText(ColorUtil.toAlphaHexColor(c).toUpperCase());`

runelite-client/src/main/java/net/runelite/client/ui/components/colorpicker/CheckerboardBackground.java
```diff
@@ -0,0 +1,46 @@
...
+package net.runelite.client.ui.components.colorpicker;
+
+import java.awt.Color;
+import java.awt.Graphics;
+
+public class CheckerboardBackground
+{
+	private static final int CHECKER_SIZE = 10;
+
+	public static void paintCheckerboard(Graphics g, int width, int height)
```
Adam-, line 34: This seems like it is in an odd place since it is also used outside of the colorpicker but I'm not sure where a better place would be? Maybe just duplicate it both places?

runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
```diff
@@ -44,6 +45,7 @@ default boolean hideSkillIcons()
 		return false;
 	}
 
+	@Alpha
```
Adam-, line 48: I don't think this works because it looks like the Widgets don't support alpha text colors? Does it work for you?

runelite-client/src/main/java/net/runelite/client/plugins/hunter/HunterConfig.java
```diff
@@ -25,13 +25,15 @@
...
 import java.awt.Color;
+import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
 @ConfigGroup("hunterplugin")
 public interface HunterConfig extends Config
 {
+	@Alpha
```
Adam-, line 36: None of this has any effect due to the overlay hardcoding an alpha value of 100.

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -34,6 +35,7 @@
 {
 	String GROUP = "inventorytags";
 
+	@Alpha
```
Adam-, line 38: Does this work? Even if it does, the menu colors can't be alpha either so it would look weird when configuring ingame.

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
pilino1234: `existing` seems to be an int representing the colour (like `-16711936` for colour `#ff00ff00`), but the button text that the constructor expects is a hex-encoded color like #ff00ff00. Do you mean to move the `ColorUtil.to[Alpha]HexColor(…)` calls and surrounding logic to the constructor and let it take an `int color` parameter instead?

Adam-: It just seems strange due to the string -> int -> string conversion. But I see it is to get the hex-encoded value of the integer. Maybe you can just change this to `Color existing = configManager.getConfiguration(cd.getGroup().value(), cid.getItem().keyName(), Color.class);` and then do the conversion from that to a hex string which would look better.

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
pilino1234: Currently I built the logic to only show the 8-digit hex codes (#aarrggbb) on config options that actually support opacity, the other ones keep the 6-digit version (#rrggbb). Using only `.toAlphaHexColor` here would cause those buttons to also have the long hex codes after they have been changed, which would be quite confusing IMO.

runelite-client/src/main/java/net/runelite/client/ui/components/colorpicker/CheckerboardBackground.java
pilino1234: @Nightfirecat asked me to unify the duplicate code into a single public class: https://github.com/runelite/runelite/pull/12354#discussion_r477614502

runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
pilino1234: That overloaded getConfiguration is nifty, I'll use that :+1:

Adam-:
Can you remove all of the `@Alpha` from config options which do not support it? I mentioned some of them in the previous review.

pilino1234:
> Can you remove all of the `@Alpha` from config options which do not support it? I mentioned some of them in the previous review.

Will do, I intend to finally get around to it this weekend :+1:

pilino1234:
Before and after removing alpha from overlay text colors.

textalpha

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayUtil.java
```diff
@@ -92,7 +92,7 @@ public static void renderTextLocation(Graphics2D graphics, Point txtLoc, String
 		graphics.setColor(Color.BLACK);
 		graphics.drawString(text, x + 1, y + 1);
 
-		graphics.setColor(color);
+		graphics.setColor(new Color(color.getRed(), color.getGreen(), color.getBlue()));
```
Nightfirecat, line 95: We'd honestly be better off making some kind of `ColorUtil` function along the lines of `Color colorOfAlpha(Color color, int alpha)`, as this can be used in many other places as well.

runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
```diff
@@ -0,0 +1,80 @@
...
+		Color textColor = Color.WHITE;
+
+		if (lum > 0.5)
+		{
+			textColor = Color.BLACK;
+		}
+		else if (color.getAlpha() < ALPHA_TEXT_CUTOFF)
+		{
+			textColor = Color.BLACK;
+		}
```
Nightfirecat, line 65:
```java
final Color textColor;

if (lum > 0.5 || color.getAlpha() < ALPHA_TEXT_CUTOFF)
{
	textColor = Color.BLACK;
}
else
{
	textColor = Color.WHITE;
}
```

runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
```diff
@@ -0,0 +1,80 @@
...
+		this.setContentAreaFilled(false);
+	}
+
+	public void setColor(Color color)
+	{
+		this.color = color;
+
+		// Calculate perceptive luminance of new color to choose a readable font color
+		// Based on https://stackoverflow.com/a/1855903
+		double lum = (0.299 * color.getRed() + 0.587 * color.getGreen() + 0.114 * color.getBlue()) / 255;
```
Nightfirecat, line 54: A luminance calculation function would also be a good candidate for a ColorUtil function.

runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
```diff
@@ -0,0 +1,80 @@
...
+import net.runelite.client.ui.components.colorpicker.CheckerboardBackground;
+
+public class ColorJButton extends JButton
+{
+	private static final int ALPHA_TEXT_CUTOFF = 120;
+
+	@Getter
+	private Color color;
+
+	public ColorJButton(String text)
```
Nightfirecat, line 40: It seems to me the color of this class should be provided on creation instead of only allowing it to be set via setter.

runelite-client/src/main/java/net/runelite/client/ui/components/ColorJButton.java
```diff
@@ -0,0 +1,80 @@
...
+			textColor = Color.BLACK;
+		}
+
+		this.setForeground(textColor);
+	}
+
+	@Override
+	public void paint(Graphics g)
+	{
+		CheckerboardBackground.paintCheckerboard(g, getWidth(), getHeight());
```
Nightfirecat, line 73: This can be wrapped in a check for the color's alpha value (`if (color.getAlpha() != 255)`) since it is more expensive to do a double-draw on the background.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
```diff
@@ -122,6 +123,7 @@ default boolean alwaysShowBaseStats()
 		return false;
 	}
 
+	@Alpha
```
Nightfirecat, line 126: I'm not sure we want to draw transparent text on the tooltip since that could look very odd.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -355,6 +355,7 @@ default int infoBoxSize()
 		return 35;
 	}
 
+	@Alpha
```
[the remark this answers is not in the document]
pilino1234: I moved the annotation in front of the `@ConfigItem` to make it consistent with all the other places. It's what was removed at https://github.com/runelite/runelite/pull/12354/files#r503572044

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -374,7 +375,6 @@ default boolean infoBoxTextOutline()
 		position = 44,
 		section = overlaySettings
 	)
-	@Alpha
```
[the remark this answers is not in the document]
pilino1234: This isn't removed, just moved to https://github.com/runelite/runelite/pull/12354/files#r503572409 to make things consistent.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
pilino1234: Agreed. I'm going through all places where I added alpha to check if the colors are also used for text, haven't gotten to this one just yet though :wink:

runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayUtil.java
pilino1234: I was thinking about something similar, just a `Color opaque(Color col)` that would remove alpha from a color. But specifying the actual alpha is more flexible :+1:

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -131,6 +142,31 @@ public static String colorToAlphaHexCode(final Color color)
...
+	}
+
+	/**
+	 * Calculate perceptive luminance of a color.
+	 * Based on https://stackoverflow.com/a/1855903
+	 *
+	 * @param color The to calculate luminance of.
+	 * @return      Perceptive luminance of the color.
+	 */
+	public static double luminance(Color color)
```
Adam-, line 165: This can probably just be in the colorjbutton

runelite-client/src/main/java/net/runelite/client/ui/components/colorpicker/CheckerboardBackground.java
Adam-: I'd probably rather have it duplicated since it is so small than have the backwards dep of ui components -> colorpicker

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -131,6 +142,31 @@ public static String colorToAlphaHexCode(final Color color)
...
 
+	/**
+	 * Gets the same RGB color with the specified alpha value.
+	 *
+	 * @param color The RGB color to use.
+	 * @param alpha The alpha value to use (0-255).
+	 * @return      A new Color with the given RGB and alpha.
+	 */
+	public static Color colorOfAlpha(final Color color, int alpha)
+	{
```
Adam-, line 153: +`if (color.getAlpha() == alpha) return color;`

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -44,6 +44,12 @@
 		put(Color.WHITE, "ffffff");
 	}};
 
+	private static final Map<Color, String> COLOR_ALPHA_HEXSTRING_MAP = new HashMap<Color, String>()
```
Adam-, line 47: use `ImmutableMap.of`

pilino1234: Should I update the `COLOR_HEXSTRING_MAP` above too while I'm at it?

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -147,10 +147,11 @@ public static String colorToAlphaHexCode(final Color color)
 	 *
 	 * @param color The RGB color to use.
 	 * @param alpha The alpha value to use (0-255).
-	 * @return      A new Color with the given RGB and alpha.
+	 * @return      A Color with the given RGB and alpha.
 	 */
 	public static Color colorOfAlpha(final Color color, int alpha)
 	{
+		if (color.getAlpha() == alpha) return color;
```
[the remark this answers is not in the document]
pilino1234: Fixed, I thought it might be fine since Checkstyle didn't complain about it :+1:

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -131,6 +142,23 @@ public static String colorToAlphaHexCode(final Color color)
...
+	 * @param alpha The alpha value to use (0-255).
+	 * @return      A Color with the given RGB and alpha.
+	 */
+	public static Color colorOfAlpha(final Color color, int alpha)
+	{
+		if (color.getAlpha() == alpha)
+		{
+			return color;
+		}
+		alpha = constrainValue(alpha);
```
abextm, line 158: newline above

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -131,6 +142,23 @@ public static String colorToAlphaHexCode(final Color color)
...
 	}
 
+	/**
+	 * Gets the same RGB color with the specified alpha value.
+	 *
+	 * @param color The RGB color to use.
+	 * @param alpha The alpha value to use (0-255).
+	 * @return      A Color with the given RGB and alpha.
+	 */
+	public static Color colorOfAlpha(final Color color, int alpha)
```
abextm, line 152: This should be named `withAlpha`. The color **of** an alpha value will always be gray.

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
```diff
@@ -85,6 +85,17 @@ public static String toHexColor(final Color color)
 		return "#" + colorToHexCode(color);
 	}
 
+	/**
+	 * Converts a given color with alpha to its hexadecimal equivalent.
+	 *
+	 * @param color Color to get hexadecimal string from.
+	 * @return      Hexadecimal string representing the given color, in the form "#aarrggbb".
+	 */
+	public static String toAlphaHexColor(final Color color)
```
abextm, line 94: this method name is easily confused with `colorToAlphaHexCode` and probably should just not exist at all.

abextm (changes requested):

runelite-client/src/main/java/net/runelite/client/util/ColorUtil.java
pilino1234: I suppose I can remove it (the naming was done for consistency with existing methods, so renaming would also add confusion). It's only used in two places in ConfigPanel.java, although those uses are within ternaries that are approaching questionable line lengths. Is there a preferred way to add line breaks to ternaries to make them more readable?

[ConfigPanel.java:423](https://github.com/runelite/runelite/pull/12354/files#diff-078b2bca651a2befbe603356afbeda841673e0706eb05b62f99e57156c826437R423) for example
```diff
- String colorHex = (alphaHidden ? ColorUtil.toHexColor(existing) : ColorUtil.toAlphaHexColor(existing)).toUpperCase();
+ String colorHex = (alphaHidden ? ColorUtil.toHexColor(existing) : "#" + ColorUtil.colorToAlphaHexCode(existing)).toUpperCase();
```

abextm (approved):

PR #12361, opened 2020-08-18
CombatLevel: Properly removes combat skills from tooltip when they are maxed

rralyks:
Removes Attack/Strength/Defence/Hitpoints from showing in the "Next combat level:" tooltip when that individual skill is maxed instead of just when the pairs (Att/Str, Def/HP) are maxed.


Adam-:
I'm not sure this really needs fixing since it is understood that the `Attack/Strength` are a group and you can do either. Additionally this is adding a code path for eg. `sb.append(meleeNeed).append(" ").append("</br>")` which seems wrong.

Issue #12320, opened 2020-08-10
Mining overlay for dense runestones [Mining /  Not Mining]


YunusEmreSener:
**Feature request**

**Request description**
Mining overlay that tells if you are mining or not. Added an example below from a wc overlay. This is helpful when afk mining dense runestones, daeyalt essence and other afk mining resources. Preferably a toggle option. 

**example from another plugin (WC plugin)** 
![image](https://user-images.githubusercontent.com/39036992/89772438-4a3ab380-db02-11ea-99a6-db7dacfb448c.png)
![image](https://user-images.githubusercontent.com/39036992/89772492-5de61a00-db02-11ea-8c90-bda71ef8b482.png)

**edit** 
Just realised that there is a mining overlay for other rocks but not for dense runestones.

**See also**
https://github.com/runelite/runelite/pull/12033

Adam-:
added in 5a524471f8358e75405f787964391c986e529029

PR #12366, opened 2020-08-19
Mining: Add Mining/Not Mining Overlay to Dense Essence

chasertw123:
Mining overlay for dense runestones [Mining / Not Mining] #12320

Needed additional way to start a MiningSession as mining dense essence doesn't have a chat message when you succeed.
Added check for addition animation as mining dense essence using multiple animations.


Adam-:
added in 5a524471f8358e75405f787964391c986e529029

PR #12368, opened 2020-08-19
Agility Plugin: Highlight Hallowed Sepulchre Crossbowman Statues

### LINKED ISSUE #12191: Hallowed sepulchre: highlight statues before they shoot
Would be amazing if shooting statues were highlighted, default animations are sometimes confusing when there are many things to watch out for.

Pratted:
These changes to the agility plugin enable the crossbowman statues to be highlighted during their animation. An additional config option has been added to Hallowed Sepulchre section that gives the player the option to toggle the statue highlighting.

image
image
image

Sample clip 1:
https://streamable.com/ndivsa

Sample clip 2:
https://streamable.com/wuo15h

Closes #12191. 

Related to #11903


runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.google.common.base.Stopwatch;
+import lombok.Getter;
+import net.runelite.api.GameObject;
+import net.runelite.api.Renderable;
+
+import java.time.Duration;
+
+// There is no easy way to determine if a statue in animating so this
+// so this implementation uses the statue's renderable to determine if it
```
Adam-, line 35:
```suggestion
// implementation uses the statue's renderable to determine if it
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,95 @@
...
+// is animating. If the renderable changes, we know that a new frame in
```
Adam-, line 36: I'm pretty sure this isn't correct? I think every frame uses the same renderable. The renderable changes when they send the command to play a new animation. This seems to match up with the code you have below. I also think this is all that really needs to be explained in the comment.

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,95 @@
...
+// At first, one solution might seem to be to listen for the first
+// renderable change and start a timer for the duration of the animation.
+// This would work in most scenarios, but it breaks when the floor is not a
+// new instance because the statue may already be mid-animation so the first
+// renderable change would just be another frame in the animation - not
+// necessarily the start.
+//
+// Until there is some sort of event that indicates when the animation is
+// finished for a game object, there doesn't seem to be an easier workaround.
+public class CrossbowmanStatue
```
Adam-, line 61:
```suggestion
class CrossbowmanStatue
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/Obstacles.java
```diff
@@ -138,4 +138,8 @@
 		// They are multilocs, thus we use the NullObjectID
 		NULL_39524, NULL_39525, NULL_39526, NULL_39527, NULL_39528, NULL_39533
 	);
+
+	static final Set<Integer> SEPULCHRE_CROSSBOWMAN_STATUES = ImmutableSet.of(
+			CROSSBOWMAN_STATUE, CROSSBOWMAN_STATUE_38445, CROSSBOWMAN_STATUE_38446
```
Adam-, line 143:
```suggestion
		CROSSBOWMAN_STATUE, CROSSBOWMAN_STATUE_38445, CROSSBOWMAN_STATUE_38446
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityConfig.java
```diff
@@ -276,4 +276,16 @@ default boolean highlightSepulchreSkilling()
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "highlightSepulchreStatues",
```
Adam-, line 281:
```suggestion
		keyName = "highlightSepulchreStatues",
		name = "Highlight Crossbowman Statues",
		description = "Highlights the crossbowman statues when their animation begins.",
		position = 21,
		section = sepulchreSection
```

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,95 @@
...
+	public boolean isAnimating()
+	{
+		// The renderable changes when a different frame of the animation is displayed.
+		if (renderable != gameObject.getRenderable())
+		{
+			renderable = gameObject.getRenderable();
+
+			// Restart the stopwatch because this is a new renderable in the animation.
+			stopwatch.reset();
+			stopwatch.start();
```
Adam-, line 89: you can add `return true;` here.

Adam- (commented):
Overall this seems fine.

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
Pratted: I agree with you for the most part, although I have a couple points I would like to bring up:

1. What I was trying to say is that when the reference to the `Renderable` changes, the statue visibly changes. I'm assuming that when this happens, the statue is animating. I think we're both on the same page here.

2. Generally speaking, an animation is composed of frames where each frame is a static image. Whenever some graphical changes occur in animation, a new frame must be being displayed because each frame is a static image. When I noticed that the renderable changing caused a graphical change to the statue, naturally it seemed like a new frame of the animation must have been displayed because of the reasoning above. However, if every frame uses the same renderable like you said, then I agree that won't be the case. I just find it counterintuitive that every frame would use the same renderable because typically at least some frames (images) in an animation would have a slight graphical difference between one another. 

As long as you agree with 1, I think we're good to go and I'll revise the comment to be a lot more concise. I brought up point 2 because your reply seemed counterintuitive to me, but I trust your judgement.

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityOverlay.java
```diff
@@ -169,6 +170,24 @@ public Dimension render(Graphics2D graphics)
 			}
 		}
 
+		Collection<CrossbowmanStatue> statues = plugin.getStatues().values();
+		if (!statues.isEmpty() && config.highlightSepulchreStatues())
+		{
+			for (CrossbowmanStatue statue : statues)
+			{
+				if (statue.isAnimating() && statue.getGameObject() != null && statue.getGameObject().getClickbox() != null)
```
Adam-, line 178: `statue.getGameObject() != null` is always true

Adam-, line 178: `.getClickbox()` is a very expensive call, call it once and then store the result.

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,74 @@
...
+// the ANIMATION_TIMEOUT is exceeded.
+class CrossbowmanStatue
+{
+
+	// When this timeout is exceeded, we assume the statue stops animating.
+	private static final Duration ANIMATION_TIMEOUT = Duration.ofMillis(700);
+
+	// This stopwatch gets reset every time the renderable changes and is
+	// used to determine if the statue has timed out.
+	private Stopwatch stopwatch = Stopwatch.createUnstarted();
```
Adam-, line 46: `final`

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Getter
+	private GameObject gameObject;
```
Adam-, line 49: `final`

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,74 @@
...
+	private Renderable renderable;
+
+	public CrossbowmanStatue(final GameObject gameObject)
```
Adam-, line 52: `@RequiredArgsConstructor`

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,74 @@
...
+	{
+		this.gameObject = gameObject;
+	}
+
+	public boolean isAnimating()
+	{
+		// The renderable changes when a different frame of the animation is displayed.
```
Adam-, line 59: This comment is incorrect.

runelite-client/src/main/java/net/runelite/client/plugins/agility/CrossbowmanStatue.java
```diff
@@ -0,0 +1,74 @@
...
+		if (renderable != gameObject.getRenderable())
+		{
+			renderable = gameObject.getRenderable();
+
+			// Restart the stopwatch because this is a new renderable in the animation.
```
Adam-, line 64: and also this one

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityOverlay.java
Adam-: It may be worth using `getConvexHull` instead actually since we generally try to avoid computing clickboxes for things you don't actually have to click, due to how slow it is.

Pratted: Did not know this. I took your advice here and used convex hull since it basically outlines the same thing.

DalderupMaurice:
is this issue in the clear? @Adam-

lemonphresh:
Any updates here? :)

ar1a:
Would love to see this merged

superkitten12:
bump

PR #12369, opened 2020-08-20
Scene uploader optimizations

symaxian:
Optimize SceneUploader.upload(Scene scene)
- Remove the reset(Tile tile) codepath, a single iteration over the tiles is sufficient.
- Create an uploadModel() method optimized for static models.
- Modify the Buffer.ensureCapacity() methods to avoid allocating intermediate buffers.

Some quick testing indicates a 10-20% improvement in initial scene load time.


Adam-:
Can you check out https://github.com/Adam-/runelite/commits/gpu-opt2 and see if its performance is similar? I've rebased the commits.

symaxian:
Yeah looks like most of it came through and the speedup is similar.

There is a minor loop hoist missing for the tile iteration and some de-duplication is possible if uploadModel() takes in a Renderable, but that's all that stands out different that would matter.

Adam-:
this has been merged.

PR #12375, opened 2020-08-21
xpglobes: add option to keep all globes on screen until oldest expires

Enriath:
Currently, XP Globes get removed as soon as they reach the timeout time. This PR adds an option that makes it so XP Globes only get removed when the newest globe reaches the timeout time.


Nightfirecat (commented):
LGTM, but needs rebase

PR #12380, opened 2020-08-23
World Map: adjusted the height of tooltip text to match vanilla tooltips

honeyhoney:
Current behaviour: 
image

New behaviour:
image

Vanilla behaviour for reference:
image


makingcomment:
Your screenshots make it look like vanilla has a horizontal offset that isn't being emulated as well. Is that something that could also be fixed, or is that horizontal difference just due to different cursor positions used to take the screenshots?

honeyhoney:
> 
> 
> Your screenshots make it look like vanilla has a horizontal offset that isn't being emulated as well. Is that something that could also be fixed, or is that horizontal difference just due to different cursor positions used to take the screenshots?

This pull request pertains only to the positioning of the text label within the tooltip, not the position of the tooltip itself. I primarily did it to prevent the overlap of text and tooltip border.

We don't emulate the vanilla positioning of tooltips, we just use a static offset from position of the icon.

Vanilla tooltips are rendered relative to where your mouse enters the icon. I'm sure it could be "fixed" but I don't personally feel like it's a behaviour worth my time in attempting to emulate it.

PR #12395, opened 2020-08-26
Fix barrows reward percentage calculation

### LINKED ISSUE #12009: Barrows Reward Potential not being colored properly
For every monster killed, the player gains Reward Potential equal to the combat level, up to a cap of 1000. Upon item roll, for every Barrows brother killed the player gains an additional +2 points, which increases the cap to 1000 + 2 * # brothers killed, up to 1012. 

The Reward Potential being returned from the client `client.getVar(Varbits.BARROWS_REWARD_POTENTIAL)` doesn't have the +2 points for each brother, just the combat levels. This means that when the client states that the user has 880 Reward Points, which is displayed as green in RuneLite because it's under 88.0%, for the sake of item rolling they will actually have 892 Reward Points and accidentally unlock Bolt Racks, which are unlocked at 881. 

One possible solution would be to re-calculate both the Reward Potential and the total possible points to include the +(2 * # brothers killed) bonus points for item rolls, as this is what the user actually cares about. The color thresholds would need to be calculated, rather than hard coded, in this scenario. Another solution would be to simply adjust the current upper threshold for green to 86.8%, which is the ideal number - 1.2% from the 12 invisible points.

I think my favorite solution would be to move away from using percentages and instead just use the Reward Potential integer, as this is how all of the resources around Barrows track the information. One could either show the Reward Potential and correct it by adding the invisible points, or show Base Reward Potential as the number returned from the API, Bonus Reward Potential (which is the 2 * brothers killed figure), and then lastly Total Reward Potential for maximum clarity. 

Sources: 
[Wiki](https://oldschool.runescape.wiki/w/Barrows/Strategies#Optimising_the_average_value_of_the_rewards)
[Mod Kieren](https://twitter.com/JagexKieren/status/705428283509366785/photo/1)

MarcGuasch:
rewardPercent incorporate Barrows brother's increase in max points. percentage format to 1 decimal place. Now with barrows kills, rewardPercent displays true percentage.

Fixes #12009

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -85,10 +85,10 @@ public class BarrowsBrotherSlainOverlay extends OverlayPanel
 				.build());
 		}
 
-		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / (10.0f + BrotherReward*.02f);
+		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / (10.0f + BrotherReward * .02f);
 		panelComponent.getChildren().add(LineComponent.builder()
 				.left("Potential")
-				.right(rewardPercent != 0 ? String.format("%.1f",rewardPercent) + "%" : "0%")
+				.right(rewardPercent != 0 ? String.format("%.1f" , rewardPercent) + "%" : "0%")
 				.rightColor(rewardPercent >= 73.0f && rewardPercent <= 88.0f ? Color.GREEN : rewardPercent < 65.6f ? Color.WHITE : Color.YELLOW)
 				.build());
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -72,9 +72,11 @@ public class BarrowsBrotherSlainOverlay extends OverlayPanel
 			potential.setHidden(true);
 		}
 
+		float BrotherReward = 0;
 		for (BarrowsBrothers brother : BarrowsBrothers.values())
 		{
 			final boolean brotherSlain = client.getVar(brother.getKilledVarbit()) > 0;
+			BrotherReward += brotherSlain ? 1.0f : 0.0f;
 			String slain = brotherSlain ? "\u2713" : "\u2717";
 			panelComponent.getChildren().add(LineComponent.builder()
 				.left(brother.getName())
@@ -83,10 +85,10 @@ public class BarrowsBrotherSlainOverlay extends OverlayPanel
 				.build());
 		}
 
-		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / 10.0f;
+		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / (10.0f + BrotherReward*.02f);
 		panelComponent.getChildren().add(LineComponent.builder()
 				.left("Potential")
-				.right(rewardPercent != 0 ? rewardPercent + "%" : "0%")
+				.right(rewardPercent != 0 ? String.format("%.1f",rewardPercent) + "%" : "0%")
 				.rightColor(rewardPercent >= 73.0f && rewardPercent <= 88.0f ? Color.GREEN : rewardPercent < 65.6f ? Color.WHITE : Color.YELLOW)
 				.build());
 


runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
```diff
@@ -72,9 +72,11 @@ public Dimension render(Graphics2D graphics)
 			potential.setHidden(true);
 		}
 
+		float BrotherReward = 0;
```
mikearche, line 75: nit: please fix the casing `brotherReward`

Nightfirecat, line 75: We can avoid adding a new variable by moving `rewardPercent`'s definition to this line and simply adding `0.2f` to the reward percent for each brother that's killed.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
```diff
@@ -83,10 +85,10 @@ public Dimension render(Graphics2D graphics)
 				.build());
 		}
 
-		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / 10.0f;
+		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / (10.0f + BrotherReward * .02f);
```
Nightfirecat, line 88: This calculation does not look correct. Why would killing brothers increase the amount we divide by? As written, this would calculate the maximum potential as `1000 / 10.12` instead of `1012 / 10`.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
MarcGuasch: "For every Barrows brother that was slain, the player will receive an extra 2 points on their rewards potential, capping the maximum reward potential at 1,012." ~ https://oldschool.runescape.wiki/w/Barrows/Strategies#Medium%20level 
What we want to display is a percentage so we want (points/totalpoints) * 100%. As max total points in 1012, and we multiple by 100, that gives 10.12

Nightfirecat: Is the varbit accurate to that description then? That is, does killing Ahrim add 98 or 100 to its value? Regardless, if changing the denominator happens in this PR, it becomes ambiguous what % should be stopped at to avoid getting bolt racks since the actual number of points (not the percentage) is what determines that.

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
MarcGuasch: good idea, thanks

MarcGuasch: > We can avoid adding a new variable by moving `rewardPercent`'s definition to this line and simply adding `0.2f` to the reward percent for each brother that's killed.

I don't see how one would do this without doing some bad variable use and have 'rewardPercent' represent different things at different points. The brother variable (alpha) is in the denominator so there is no clean way to separate it. If you see something i don't, could you please elaborate?

runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
MarcGuasch: do you think then, it would be better to just give the raw points instead, sense that's what matters and that's what people know?

MarcGuasch: thinking it over, giving the raw points seems best because it easily displays what is important, plays on what the community knows, and prevents a new variable from being made

Nightfirecat: I agree. Since percentage isn't completely clear, it would likely be better to display points instead especially since the wiki references those same points for calculating reward potential.

rchinerman: > Is the varbit accurate to that description then? That is, does killing Ahrim add 98 or 100 to its value?

The varbit is technically accurate, per Mod Kieren. You can gain 1000 total points, which is the sum of the combat levels of all the monsters you've killed with a cap of 656 from the brothers. Then, at item roll, your total points is added to 2 x # of brothers slain and this new number can reach 1012. 

The player is interested in the point value they will be rolling at, however, which is why the bonus points should likely be included the entire time in one way or another. My original suggestion was: 

> One could either show the Reward Potential and correct it by adding the invisible points, or show Base Reward Potential as the number returned from the API, Bonus Reward Potential (which is the 2 * brothers killed figure), and then lastly Total Reward Potential for maximum clarity.

I do somewhat worry that just including the invisible points and adding 100 points for a level 98 kill could confuse players considering the added points are supposed to match the combat level, which is why I think showing all of the information could potentially be useful. I'm not sure how cluttered that would feel in the UI though.

Nightfirecat:
This is fixed in 807a22172f.

PR #12400, opened 2020-08-26
Varbits: Add Hallowed Sepulchre Varbits

Pratted:
Added some useful Hallowed Sepulchre varbits to `Varbits`.


Nightfirecat:
We don't typically add varbit definitions for things we don't use in the client. If needed in a hub plugin, you can access these values using [`Client.getVarbitValue(int varbitId)`](http://static.runelite.net/api/runelite-api/net/runelite/api/Client.html#getVarbitValue-int-)

Pratted:
Ok, didn't know that. I'm using these varbits for a hub plugin so I'll use your suggestion and just define them there.

PR #12414, opened 2020-08-29
Added hallowed sepulchre floors into the agility calculator

arthur798:
This merge adds each hallowed sepulchre agility course into the agility calculator by floor, not sure how we would go about making "combined" run when a person completes floors from 1-5 in one go, not sure if that needs to be handled? Icon might need to be changed so happy to take suggestions

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_agility.json
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_agility.json
@@ -79,6 +79,12 @@
       "xp": 571
     },
     {
+      "level": 52,
+      "icon": 11849,
+      "name": "Hallowed Sepulchre Floor 1",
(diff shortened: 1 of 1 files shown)


Broooklyn:
Since Hallowed Sepulchre is not an Agility course that produces Marks of grace, you should not use `11849` for the icons. Perhaps `24736` (Ring of Endurance) or `24711` (Hallowed mark), although that's only a single mark and maybe not very identifiable. 

java_aAfRLXtQkY
java_1ZFlVVldhm

arthur798:
I am happy to change if that's what is better? Anyone else have suggestions are agree with @Broooklyn

Hydrox6:
The Ring looks the better out of the two, and it feels like the icon for the area

arthur798:
no problem I will ammend the icons now

arthur798:
Is anything else needed to get this merged?

PR #12423, opened 2020-08-31
MTA: Clean up config, Enchanting Chamber update

Matthew-nop:
Changed config room names to reflect those on the wiki (e.g. Telekinetic Room -> Telekinetic Theatre). Changed config to use collapsible sections to reflect separation of different room options. Added a toggle for marking dragonstones in the enchantment room. Changed the indicator for dragonstone spawns from a hint arrow to a highlighted tile to allow highlighting of all dragonstones instead of nearest.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
@@ -32,6 +32,13 @@ import net.runelite.client.config.ConfigSection;
 @ConfigGroup("mta")
 public interface MTAConfig extends Config
 {
+	enum MarkingMode
+	{
+		ALL,
+		NEAREST,
+		NONE
+	}
+
 	@ConfigSection(
 		name = "Alchemists' Playground",
 		description = "",
@@ -101,10 +108,23 @@ public interface MTAConfig extends Config
 		keyName = "enchantment",
 		name = "Enable",
 		description = "Configures whether or not the Enchanting Chamber overlay is enabled.",
+		position = 0,
 		section = enchantmentGroup
 	)
 	default boolean enchantment()
 	{
 		return true;
+	}
+
+	@ConfigItem(
+		keyName = "dragonstoneMode",
+		name = "Highlighting mode",
+		description = "Chooses which dragonstones should be highlighted.",
+		position = 1,
+		section = enchantmentGroup
+	)
+	default MarkingMode dragonstoneMode()
+	{
+		return MarkingMode.ALL;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
@@ -27,15 +27,48 @@ package net.runelite.client.plugins.mta;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
 
 @ConfigGroup("mta")
 public interface MTAConfig extends Config
 {
+	@ConfigSection(
+		name = "Alchemists' Playground",
+		description = "",
+		position = 0,
+		closedByDefault = true
+	)
+	String alchemyGroup = "alchemy";
+
+	@ConfigSection(
+		name = "Creature Graveyard",
+		description = "",
+		position = 1,
+		closedByDefault = true
+	)
+	String graveyardGroup = "graveyard";
+
+	@ConfigSection(
+		name = "Telekinetic Theatre",
+		description = "",
+		position = 2,
+		closedByDefault = true
+	)
+	String telekineticGroup = "telekinetic";
+
+	@ConfigSection(
+		name = "Enchanting Chamber",
+		description = "",
+		position = 3,
+		closedByDefault = true
+	)
+	String enchantmentGroup = "enchantment";
+
 	@ConfigItem(
 		keyName = "alchemy",
-		name = "Enable alchemy room",
-		description = "Configures whether or not the alchemy room overlay is enabled.",
-		position = 0
+		name = "Enable",
+		description = "Configures whether or not the Alchemists' Playground overlay is enabled.",
+		section = alchemyGroup
 	)
 	default boolean alchemy()
 	{
@@ -44,9 +77,9 @@ public interface MTAConfig extends Config
 
 	@ConfigItem(
 		keyName = "graveyard",
-		name = "Enable graveyard room",
-		description = "Configures whether or not the graveyard room overlay is enabled.",
-		position = 1
+		name = "Enable",
+		description = "Configures whether or not the Creature Graveyard overlay is enabled.",
+		section = graveyardGroup
 	)
 	default boolean graveyard()
 	{
@@ -55,9 +88,9 @@ public interface MTAConfig extends Config
 
 	@ConfigItem(
 		keyName = "telekinetic",
-		name = "Enable telekinetic room",
-		description = "Configures whether or not the telekinetic room overlay is enabled.",
-		position = 2
+		name = "Enable",
+		description = "Configures whether or not the Telekinetic Theatre overlay is enabled.",
+		section = telekineticGroup
 	)
 	default boolean telekinetic()
 	{
@@ -66,9 +99,9 @@ public interface MTAConfig extends Config
 
 	@ConfigItem(
 		keyName = "enchantment",
-		name = "Enable enchantment room",
-		description = "Configures whether or not the enchantment room overlay is enabled.",
-		position = 3
+		name = "Enable",
+		description = "Configures whether or not the Enchanting Chamber overlay is enabled.",
+		section = enchantmentGroup
 	)
 	default boolean enchantment()
 	{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
@@ -24,6 +24,9 @@
  */
 package net.runelite.client.plugins.mta.enchantment;
 
+import java.awt.Color;
+import java.awt.Graphics2D;
+import java.awt.Polygon;
 import java.util.ArrayList;
 import java.util.List;
 import javax.inject.Inject;
@@ -31,9 +34,11 @@ import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.ItemID;
+import net.runelite.api.Perspective;
 import net.runelite.api.Player;
 import net.runelite.api.Tile;
 import net.runelite.api.TileItem;
+import net.runelite.api.coords.LocalPoint;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -81,18 +88,57 @@ default boolean alchemy()
...
+	}
+
+	@ConfigItem(
+			keyName = "defaultColor",
+			name = "Highlight color",
+			description = "The highlight color for dragonstones.",
+			position = 3,
+			section = enchantmentGroup
+	)
+	default Color dragonstoneColor()
```
Nightfirecat, line 132: This doesn't need to be configurable.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -81,18 +88,57 @@ default boolean alchemy()
...
 	}
 
+	@ConfigItem(
+			keyName = "highlightDragonstones",
+			name = "Highlight dragonstones",
+			description = "Enable markers indicating spawned dragonstones.",
+			position = 1,
+			section = enchantmentGroup
+	)
+	default boolean highlightDragonstones()
```
Nightfirecat, line 108: Add a "No marking" marking mode and this config is no longer needed.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -81,18 +88,57 @@ default boolean alchemy()
...
+		return Color.YELLOW;
+	}
+
+
 	@ConfigItem(
 		keyName = "graveyardEnable",
 		name = "Enable",
-		description = "Configures whether or not the graveyard room overlay is enabled.",
+		description = "Configures whether or not the Creature Graveyard overlay is enabled.",
+		position = 0,
```
Nightfirecat, line 142: Items which are alone in their config sections do not need positions assigned to them.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
```diff
@@ -67,67 +70,47 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
...
 	{
-		final TileItem item = itemDespawned.getItem();
-		final Tile tile = itemDespawned.getTile();
-
-		if (item.getId() == ItemID.DRAGONSTONE_6903)
-		{
-			WorldPoint location = tile.getWorldLocation();
-			log.debug("Removed dragonstone at {}", location);
-			dragonstones.remove(location);
+		if (config.enchantment()) {
```
Nightfirecat: Why is this added?

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
```diff
@@ -34,22 +36,23 @@
...
+import net.runelite.client.ui.overlay.OverlayUtil;
 
 @Slf4j
 public class EnchantmentRoom extends MTARoom
 {
 	private static final int MTA_ENCHANT_REGION = 13462;
 
 	private final Client client;
-	private final List<WorldPoint> dragonstones = new ArrayList<>();
+	private final List<Tile> dragonstones = new ArrayList<>();
```
Nightfirecat, line 55: Why are tiles used here instead of WorldPoints like before? We only ever call `getLocalLocation()` of dragonstones.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -27,50 +27,86 @@
...
 	{
 		return true;
 	}
 
 	@ConfigItem(
-		keyName = "graveyard",
-		name = "Enable graveyard room",
-		description = "Configures whether or not the graveyard room overlay is enabled.",
-		position = 1
+			keyName = "enchantmentEnable",
```
Nightfirecat: Double check your formatter--these extra tabs are unwanted. `@ConfigItem(` should cause only one extra level of indentation.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
Matthew-nop: Removed with latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
Matthew-nop: Good point, implemented in MTAConfig.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
Matthew-nop: Removed with latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
Matthew-nop: I added this as a guard since the enchantment room was the only room without them. I modified it to be more consistent with alchemy room (early return).

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
Matthew-nop: I originally did this when testing, but forgot to revert it before pushing the changed. Reverted it back to WorldPoints in the latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -27,50 +27,105 @@
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
 
-@ConfigGroup("mta")
+@ConfigGroup(MTAConfig.GROUP)
 public interface MTAConfig extends Config
 {
+	String GROUP = "mta";
```
Nightfirecat, line 35: Since this is never referenced elsewhere, this can be reverted.

Nightfirecat (changes requested):
Additionally, can these changes be rebased into clean commits? One for code cleanup and another for the behavior changes. The diff is rather messy right now since there's fixup and merge commits littering the history.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
```diff
@@ -27,50 +27,105 @@
...
+		closedByDefault = true
+	)
+	String telekineticGroup = "telekinetic";
+
 	@ConfigItem(
-		keyName = "alchemy",
-		name = "Enable alchemy room",
-		description = "Configures whether or not the alchemy room overlay is enabled.",
-		position = 0
+		keyName = "alchemyEnable",
```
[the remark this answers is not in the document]
Matthew-nop: My bad, I didn't know that was the case. All keyName changes have been reverted.

runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
Matthew-nop: Reverted.

Matthew-nop:
Done and done, hopefully this has organized the changes.

Nightfirecat:
I've rewritten some of this code. (prioritize `inside()` checks over config checks, less code duplication, less config order shuffling, split into one more commit) I think this should be good to go now.

Matthew-nop:
Thanks for going the extra mile. After reading what Adam said on the subject I'll make sure I'm more careful with config checks.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
```diff
@@ -106,12 +105,15 @@ private WorldPoint findNearestStone()
 	@Subscribe
 	public void onItemSpawned(ItemSpawned itemSpawned)
 	{
-		final TileItem item = itemSpawned.getItem();
-		final Tile tile = itemSpawned.getTile();
+		if (!inside() || !config.enchantment())
```
Adam-, line 96: Generally the way we implement on/off options for features is by having the on/off toggle the overlay while having the code which does the actual tracking run regardless. This is so it allows you to toggle the option on/off and have that worked as expected instead of requiring a relog. This is breaking that.

Also I don't really see any need to modify the existing code.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
```diff
@@ -143,4 +131,48 @@ public boolean inside()
 		return player != null && player.getWorldLocation().getRegionID() == MTA_ENCHANT_REGION
 			&& player.getWorldLocation().getPlane() == 0;
 	}
+
+	@Override
+	public void under(Graphics2D graphics)
+	{
+		final MarkingMode mode = config.dragonstoneMode();
```
Adam-, line 138: Prefer config checks to be after `inside()` as it is much more expensive.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
Matthew-nop: I've removed the config checks from here and `onItemDespawned()`. Although if there is a method to get all nearby items it may be better to repopulate the list of dragonstones when there's a config change.

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
Matthew-nop: Changed with the latest commit

Matthew-nop: Changed with the latest commit

runelite-client/src/main/java/net/runelite/client/plugins/mta/enchantment/EnchantmentRoom.java
Adam-: I would recommend just dropping this commit since it doesn't actually do anything and imo doesn't clean up anything either.

Matthew-nop: Sure, the changes are pretty inconsequential.

Adam-:
Additionally I have removed the `inside()` check  in the enchantment room since it is unnecessary due to the MTASceneOverlay already checking that prior to caling `under`.

I've noticed this removes the hint arrow for the dragonstone, which probably a lot of users like, since it shows on the minimap and will flash.  This kind of seems like a step down?

Matthew-nop:
Would it be preferable if there was another value in the enum which would maintain current behavior? I thought it seemed a bit out of place so I didn't include it, but if it's desired I can.

Matthew-nop:
I've added another mode for the hint arrow functionality. Should this be made the default again?

Nightfirecat:
I think having it add tile markers for all dragonstones and a hint arrow for the nearest one would be fine

Matthew-nop:
> I think having it add tile markers for all dragonstones and a hint arrow for the nearest one would be fine


> I've noticed this removes the hint arrow for the dragonstone, which probably a lot of users like, since it shows on the minimap and will flash. This kind of seems like a step down?

I've changed it such that now the `ALL` mode will both highlight all dragonstones and assign a hint arrow to the nearest. Hopefully this can alleviate any concerns.

Issue #11520, opened 2020-05-09
Give higher priority to plugins with title match in plugin searching


gc:
**Is your feature request related to a problem? Please describe.**
Searching for plugins can be weird sometimes and be sorted weirdly, for example see the image below.

**Describe the solution you'd like**
If the title includes the search query, it should be above any that don't include it in the title.

**Describe alternatives you've considered**

**Additional context**
Here I'm searching for the status bars plugin, but  its the second last result despite being the only one with "status" in the title.

![image](https://user-images.githubusercontent.com/30398469/81478541-8b9c9800-9261-11ea-8329-3e3df7062ca1.png)

PR #12432, opened 2020-09-01
Sort plugin search results with better text matches higher in search results

jrhodkinson:
Resolves https://github.com/runelite/runelite/issues/11520

- Introduce `SearchablePlugin` and extract plugin search code to `PluginSearch`, which is now used by both internal and external plugin searches
- Order search results using a few heuristics such that exact or close text matches occur higher up in the results

I tried using JaroWinklerDistance but it didn't have very good results, probably due to the varying name-length of plugins, so I settled on more human heuristics.

Screenshot 2020-09-01 at 20 50 39
Screenshot 2020-09-01 at 20 51 24
Screenshot 2020-09-01 at 20 52 34

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
@@ -87,7 +87,6 @@ import net.runelite.client.ui.components.IconTextField;
 import net.runelite.client.util.ImageUtil;
 import net.runelite.client.util.LinkBrowser;
 import net.runelite.client.util.SwingUtil;
-import net.runelite.client.util.Text;
 import net.runelite.client.util.VerificationException;
 
 @Slf4j
@@ -115,13 +114,15 @@ class PluginHubPanel extends PluginPanel
 		CONFIGURE_ICON_HOVER = new ImageIcon(ImageUtil.alphaOffset(configureIcon, -100));
 	}
 
-	private class PluginItem extends JPanel
+	private class PluginItem extends JPanel implements SearchablePlugin
 	{
 		private static final int HEIGHT = 70;
 		private static final int ICON_WIDTH = 48;
 		private static final int BOTTOM_LINE_HEIGHT = 16;
 
 		private final ExternalPluginManifest manifest;
+
+		@Getter
 		private final List<String> keywords = new ArrayList<>();
 
 		@Getter
@@ -332,6 +333,12 @@ class PluginHubPanel extends PluginPanel
 						.addComponent(configure, BOTTOM_LINE_HEIGHT, BOTTOM_LINE_HEIGHT, BOTTOM_LINE_HEIGHT)
 						.addComponent(addrm, BOTTOM_LINE_HEIGHT, BOTTOM_LINE_HEIGHT, BOTTOM_LINE_HEIGHT))
 					.addGap(5)));
+		}
+
+		@Override
+		public String getSearchableName()
+		{
+			return manifest.getDisplayName();
 		}
 	}
 
@@ -547,22 +554,19 @@ class PluginHubPanel extends PluginPanel
 
 		Stream<PluginItem> stream = plugins.stream();
 
-		String search = searchBar.getText();
-		boolean isSearching = search != null && !search.trim().isEmpty();
+		String query = searchBar.getText();
+		boolean isSearching = query != null && !query.trim().isEmpty();
 		if (isSearching)
 		{
-			String[] searchArray = SPACES.split(search.toLowerCase());
-			stream = stream
-				.filter(p -> Text.matchesSearchTerms(searchArray, p.keywords))
-				.sorted(Comparator.comparing(p -> p.manifest.getDisplayName()));
+			PluginSearch.search(plugins, query).forEach(mainPanel::add);
 		}
 		else
 		{
-			stream = stream
-				.sorted(Comparator.comparing(PluginItem::isInstalled).thenComparing(p -> p.manifest.getDisplayName()));
+			stream
+				.sorted(Comparator.comparing(PluginItem::isInstalled).thenComparing(p -> p.manifest.getDisplayName()))
+				.forEach(mainPanel::add);
 		}
 
-		stream.forEach(mainPanel::add);
 		mainPanel.revalidate();
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginSearch.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginSearch.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2020, Jack Hodkinson <https://github.com/jrhodkinson>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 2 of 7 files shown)


jrhodkinson:
In discord raiyni requested a comparision with the current approach.

Currently all matching results are ordered alphabetically by name, with pinned items first.

So for the three searches in the original post, assuming no pinned items, we would have:

Search: `statu`
|Master|Proposed|
|---|---|
|Discord|Status Bars|
|Fishing|Discord|
|Hunter|Fishing|
|Item Stats|Hunter|
|Smelting|Item Stats|
|Status Bars|Smelting|
|XP Updater|XP Updater|

Search: `gran`
|Master|Proposed|
|---|---|
|Bank|Grand Exchange|
|Chat Commands|Bank|
|Grand Exchange|Chat Commands|
|Ground Items|Ground Items|
|Item Prices|Item Prices|

Search: `bank va`
|Master|Proposed|
|---|---|
|Bank Heatmap|Bank Value Tracker|
|Bank Screenshot|Bank Value|
|Bank Value|Bank Heatmap|
|Bank Value Tracker|Bank Screenshot|
|High Alch Profit Highlighter|High Alch Profit Highlighter|

Adam-:
It seems unlikely to me that we require this much complexity to do a simple search when we are also pulling in `commons-text`, are you sure theres nothing there that is useful?

jrhodkinson:
I think the changes are warranted, the PR does two things:
1. Extracts search functionality across both plugin lists so that the same search is used for both
2. Orders search results more logically

In fact the _set_ of things returned from a search is identical to the set in master.

There are three new classes, `PluginSearch`, `SearchablePlugin`, and `SearchablePluginComparator`. Each piece is actually quite simple, so I'd push back on the argument that this is complex. The plugin search solution in this pr is now neatly composed of small classes.

- `PluginSearch` is where the search is glued together. Previously this was spread across the two list classes, now they both just have a `PluginSearch` and query it when they need to search
- `SearchablePlugin` is a simple "value" (almost) object which just houses the things relevant to searching that are common: a name, keywords
- `SearchablePluginComparator` is the bulk of the change, it implements the sorting of search results

I looked into the options from commons-text. We're using commons-text to decide the set of plugins to return (that's unchanged). I experimented with options for implementing the comparator with commons-text, but nothing worked very well. The distance functions are designed for strings that are all about the same length, rather than a 'name' and a set of keywords. Nevertheless I've tried to make the comparator clear by composing smaller comparators

jrhodkinson:
As an example of how I think this approach is simpler:

Previously we would compute search results in two passes with `showMatchingPlugins(boolean pinned, String text)`: the first pass would add the matching pinned plugins, the second would add the matching non-pinned plugins. With this PR, we find the matching results, then sort them with the comparator. Conceptually I think it's easier to reason about

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
```diff
@@ -87,8 +86,8 @@
 import net.runelite.client.util.ImageUtil;
 import net.runelite.client.util.LinkBrowser;
 import net.runelite.client.util.SwingUtil;
-import net.runelite.client.util.Text;
 import net.runelite.client.util.VerificationException;
+import static java.util.stream.Collectors.toList;
```
abextm, line 90: Don't static import

runelite-client/src/main/java/net/runelite/client/plugins/config/SearchablePlugin.java
```diff
@@ -0,0 +1,70 @@
...
+	@VisibleForTesting
+	SearchablePlugin(T plugin, String name, List<String> keywords, Supplier<Boolean> isPinnedSupplier)
+	{
+		this.plugin = plugin;
+		this.name = name;
+		this.keywords = keywords;
+		this.isPinnedSupplier = isPinnedSupplier;
+	}
+
+	public static SearchablePlugin<PluginListItem> from(PluginListItem pluginListItem)
```
abextm, line 51: These should be `to` methods on the PluginListItem/PHP.PluginItem and those classes made private again.

It would probably better for this to become an interface and make those classes implement it

runelite-client/src/main/java/net/runelite/client/plugins/config/SearchablePluginComparator.java
```diff
@@ -0,0 +1,115 @@
...
+		return comparator().compare(plugin1, plugin2);
+	}
+
+	public Comparator<SearchablePlugin<?>> comparator()
+	{
+		if (StringUtils.isBlank(query))
+		{
+			return nullsLast(isPinned()).thenComparing(name());
+		}
+		return nullsLast(exactMatch())
```
abextm, line 61: the comparator should be cached in a final field. Most of these methods are single use and should be inlined (+ a comment if its not otherwise obvious what they do)

runelite-client/src/main/java/net/runelite/client/plugins/config/SearchablePluginComparator.java
```diff
@@ -0,0 +1,115 @@
...
+public class SearchablePluginComparator implements Comparator<SearchablePlugin<?>>
+{
+
+	private final String query;
+	private final List<String> queryPieces;
+
+	public SearchablePluginComparator(String query)
+	{
+		this.query = query.trim().toLowerCase();
+		this.queryPieces = Arrays.asList(this.query.split(" "));
```
abextm, line 46: maybe use a guava `Splitter` so you can filter empties

runelite-client/src/test/java/net/runelite/client/plugins/config/PluginSearchTest.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2018, Jack Hodkinson <https://github.com/jrhodkinson>
```
abextm, line 2: current year

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginSearch.java
```diff
@@ -0,0 +1,53 @@
...
+{
+
+	private final List<SearchablePlugin<T>> searchablePlugins;
+
+	public PluginSearch(List<SearchablePlugin<T>> searchablePlugins)
+	{
+		this.searchablePlugins = new ArrayList<>(searchablePlugins);
+	}
+
+	public List<T> query(String query)
```
abextm, line 42: this can probably just be a static method on the comparator

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginSearch.java
```diff
@@ -0,0 +1,53 @@
...
+import static java.util.stream.Collectors.toList;
+
+public class PluginSearch<T>
+{
+
+	private final List<SearchablePlugin<T>> searchablePlugins;
+
+	public PluginSearch(List<SearchablePlugin<T>> searchablePlugins)
+	{
+		this.searchablePlugins = new ArrayList<>(searchablePlugins);
```
abextm, line 39: why do you need a copy?

runelite-client/src/main/java/net/runelite/client/plugins/config/SearchablePluginComparator.java
```diff
@@ -0,0 +1,115 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.config;
+
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Function;
+import org.apache.commons.lang3.StringUtils;
+import static java.util.Comparator.comparing;
```
abextm, line 32: don't static import

abextm (changes requested):
You have a newline at the top of all your class blocks; remove those

runelite-client/src/main/java/net/runelite/client/plugins/config/SearchablePlugin.java
jrhodkinson: Went for the interface in the end. One wart is that `getName` already exists on one of the implementers so I had to call the method `getSearchableName`, but it's still better than the original proposal

runelite-client/src/main/java/net/runelite/client/plugins/config/PluginSearch.java
jrhodkinson: Decided to merge the old `PluginSearch` and `PluginSearchComparator`, making this a static method and making the comparator constructor private

jrhodkinson:
Thanks for all the comments @abextm, I found them very helpful. Learnt a thing a two; I think I don't reach for interfaces often enough.

I've updated the PR resolving all of your comments

abextm (approved):

PR #12434, opened 2020-09-02
Feature: Pet list chat command

### LINKED ISSUE #12385: Pet Count
Could there be a command to show our total pet count? Just like "!pets". This would be nice for flexing in CCs and public chats.

IllyaMyshakov:
Close #12385

**Demo**

https://gyazo.com/21156d618a83813a7583377010efca33

- invoke `!pets` without any pets, followed by

- invoking `!pets` after checking the `All Pets` tab in the Collection Log

image


Nightfirecat:
As mentioned on discord you need to make associated HTTP service and API changes to properly send this data to the server to be shared with other players. As written, this will always show your own pet count whenever any player uses the `!pets` command.

Additionally, it would be good to add some kind of error/notice message when using the command without having looked up your pet count so players know how to make it show the correct number.

IllyaMyshakov:
If the current API + Service changes behave correctly, I'll have to think of adding a test for ChatCommandsPluginTest before the PR is merge ready

Adam-:
This seems like a neat idea. It would be more neat if we could figure out a way to actually share the pets instead of just a count, in a way that isn't abusable to spam. Also I think all 40 something? pets wouldn't even fit inside of one message.

IllyaMyshakov:
@Adam- Right. Hence why pet count is currently just a total count, instead of sharing a whole list of specific pets in chat (can be upwards of 46 entries/ new lines).

Adam-:
We could just take a very long list and add it as multiple chat messages instead of one. This combined with maybe a client-side !pets cooldown so that you cannot just spam !pets over and over and send >1 line each time could maybe work.

IllyaMyshakov:
Interesting idea. I can investigate how that would work.

Adam-:
For clarification, I mean each client would have a per-player !pets cooldown that is maybe based on the last time they used !pets + also how many pet lines they have. And then you wouldn't have to add a cooldown for 1 pet line probably since that is just the normal chat rate. I don't mean making the client throttle how often you can send !pets.

Adam-:
Following some discussion in Discord, it looks like the messages actually don't have a max length and will wrap itself as many times as it needs. So this could probably just be one message.

Nightfirecat (changes requested):
This is failing to build

runelite-client/src/main/resources/net/runelite/client/plugins/chatcommands/pets.json
```diff
@@ -0,0 +1,194 @@
+[
+  {
```
Adam-, line 2: This would probably be better represented just by having a `Pet` enum with a `(name, item)` constructor. Then you can just use `Pet.values()` to get a `Pet[]`

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -562,6 +701,9 @@ public void onGameStateChanged(GameStateChanged event)
 			case LOADING:
 			case HOPPING:
 				pohOwner = null;
```
Adam-, line 703: missing `break;`

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -469,6 +585,26 @@ public void onGameTick(GameTick event)
...
 		}
 
+		if (collectionLogLoaded)
+		{
+			Widget collectionLogEntryHeader = client.getWidget(WidgetInfo.COLLECTION_LOG_ENTRY_HEADER);
+			if (collectionLogEntryHeader != null && collectionLogEntryHeader.getChildren() != null)
+			{
+				Widget entryTitle = collectionLogEntryHeader.getChild(COL_LOG_ENTRY_HEADER_TITLE_INDEX);
+				// Make sure that the player is looking in the All Pets tab of the collection log
+				if (entryTitle.getText().equals("All Pets"))
```
Adam-, line 595: It's possible to look at someone else's collection log in their poh, so this should check that similar to how the boss log checks it.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -262,6 +285,99 @@ private double getPb(String boss)
...
+
+	/**
+	 * Looks up the list of owned pets in local storage for the local player
+	 */
+	private Pet[] getPetList()
+	{
+		String petListJson = configManager.getRSProfileConfiguration("pets",
+			client.getUsername().toLowerCase(), String.class);
+
+		Pet[] petList = RuneLiteAPI.GSON.fromJson(petListJson, Pet[].class);
```
Adam-, line 355: Typically we inject `Gson` into the plugin, since is provided by the injector.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -262,6 +285,99 @@ private double getPb(String boss)
...
+			if (child.getOpacity() == 0)
+			{
+				Pet pet = new Pet(
+					child.getName().replace("<col=ff9040>", "").replace("</col>", ""),
+					child.getItemId());
+				petList.add(pet);
+			}
+		}
+
+		configManager.setRSProfileConfiguration("pets", client.getUsername().toLowerCase(),
```
Adam-, line 343: `RSProfileConfiguration` is already per RS-profile, so it is not necessary to set the key to the client username. I would probably just use `chatcommands` as the group and `pets` as the key.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -262,6 +285,99 @@ private double getPb(String boss)
...
+			return;
+		}
+
+		List<Pet> petList = new ArrayList<>();
+		for (Widget child : totalPetList)
+		{
+			if (child.getOpacity() == 0)
+			{
+				Pet pet = new Pet(
+					child.getName().replace("<col=ff9040>", "").replace("</col>", ""),
```
Adam-, line 337: see `Text.removeTags`

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -262,6 +285,99 @@ private double getPb(String boss)
...
+		{
+			final Pet pet = pets[i];
+
+			try
+			{
+				final BufferedImage image = ImageUtil.resizeImage(itemManager.getImage(pet.getIcon()), 18, 16);
+				final IndexedSprite sprite = ImageUtil.getImageIndexedSprite(image, client);
+				newModIcons[modIconIdx + i] = sprite;
+			}
+			catch (Exception ex)
```
Adam-, line 309: What is this for? Does anything here throw?

http-api/src/main/java/net/runelite/http/api/chat/Pet.java
```diff
@@ -0,0 +1,71 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.http.api.chat;
+
+import java.util.Objects;
+import lombok.Getter;
+
+@Getter
+public class Pet
```
Adam-, line 31: Classes are placed in `http-api` if they are used in both `runelite-client` and `http-service` as part of the client <-> service communication, as its classpath is available in both modules. However, you are not using this in http-service, which means this class is in the wrong place.

Adam- (changes requested):

IllyaMyshakov (commented):
Addressed nits

Adam-:
I was thinking about this the other day more, and if you were to get a pet drop and then do !pets it wouldn't show without first seeing the collection log right? We should probably also figure out a way to identify whenever a pet is received, maybe by checking the follower var?

IllyaMyshakov:
You are correct that getting a pet drop will not update the pet list showed when invoking `!pets`.

I can look into a combination of checking the follower var (whenever the new pet follows the player) and the inventory (if the pet appears there when a player already has a follower out) when a pet is received.

Adam-:
> 
> 
> You are correct that getting a pet drop will not update the pet list showed when invoking `!pets`.
> 
> I can look into a combination of checking the follower var (whenever the new pet follows the player) and the player inventory (if the pet appears there when a player already has a follower out) when a pet is received.

We maybe could just use the collection log unlock message, since most people have that on, and it would be  lot easier.

PR #12436, opened 2020-09-02
Visual help plugin Request (AKA Poor eye sight plugin)

Chevalric:
I was wondering if anyone has made or is willing to make a plugin that could aid people with poor eye sight like me? Something i  need it for is Zulrah, due to me not able to see the poison clouds properly due to them being the same ish colour as the enviroment, it makes it hella difficult for me to avoid them. 

I was thinking about a way to change the colour of the clouds, for example make them pink or yellow; a colour that sticks out. 

But maybe there could be other uses for the plugin aswell? maybe certain animations or attacks or small objects, would there be a way to increase the size? is my initial suggestion and what i need it for (the changing of the colour of the clouds at zulrah) even possible to begin with?  

i am new to this forum, so i hope this is a valid request and i have no idea of the possibilities :D


Hydrox6:
1. This is not how you make a request; it should be in Issues
2. We used to have a plugin to help those who can't see the Cerb ghosts properly, and that's now banned under the statement. This would likely come under the same thing. It's probably worth looking into software to change the colours of your entire screen (I think NVidia have something for their GPUs, and I bet there are open-source alternatives)

PR #12442, opened 2020-09-02
Fix grubby chest loot tracking

### LINKED ISSUE #12440: Grubby chest incorrectly tracks loot
I noticed that grubby chests are appearing on the loot tracker now, but the loot does not appear to be getting tracked correctly.

Currently in the source (runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin) there is:
private static final String GRUBBY_CHEST_LOOTED_MESSAGE = "You unlock the chest with your key.";

There are 2 messages that are presented when looting the chest. For large KCs (>2000) this message is shown:
"You have opened the Grubby Chest lots of times."

leejt:
Fixes #12440. 

the issue is that the following things happen in order:
1) "You unlock the chest with your key."
2) Key removed
3) "You have opened the Grubby Chest 3 times."
4) New items added

This makes the old version of the code only pick up the removal of the key. The new version hooks onto the correct message.


Nightfirecat (changes requested):
Please add some tests to demonstrate the fixed behavior.

Adam-:
I've gotten the chat messages
```
2020-09-28 20:08:19 [Client] DEBUG client-patch - Chat message type SPAM: You unlock the chest with your key.
(inv change to remove key)
2020-09-28 20:08:20 [Client] DEBUG client-patch - Chat message type GAMEMESSAGE: You have opened the Grubby Chest 3 times.
(inv change to add loot)
```

Adam-:
fixed in c774117d93eadcea367553c5d0c985d7a56bfc33

PR #12460, opened 2020-09-06
slayer: Alphabetical sorting in Task enum. Add Black Knights and pirates.

Matthew-nop:
This should cover all of Krystilia's tasks. I'm not quite sure if the pirate hat is the best item to use for pirates. Any suggestions are welcome.


Nightfirecat:
I'm not too worried about alphabetical sorting here, but please split that into a separate commit if you want to see that changed.

Matthew-nop:
Split the changes into seperate commits. I originally only sorted them so I could easily cross reference with the wiki, figured might as well include it since it appeared somewhat sorted already.

PR #12461, opened 2020-09-06
Add Stray Dog Menu Entry Swap

vitorebom:


makingcomment:
It would be cool if the config for this didn't default to `true`.

Also, this might be a better fit for the Menu Swapper Extended plugin hub plugin.

Nightfirecat:
I'm closing this as this PR is quite out-of-date. If it is still wanted, recreating this should be much easier now since the refactor of this plugin and addition of config sections.

PR #12464, opened 2020-09-06
timers: add Mage Arena 2 teleblock patterns.

### LINKED ISSUE #12459: Add one minute teleblock timer for Mage Arena 2
The Mage Arena 2 bosses can cast a teleblock which lasts one minute (with pray mage enabled). We should add a timer variant for this.

The message is roughly as follows: (not confirmed exactly, this should be confirmed ingame before being implemented) `<col=\w+>A Tele Block spell has been cast on you. It will expire in 1 minute.</col>`

Matthew-nop:
Closes #12459 

Added timers for both full and half length Mage Arena 2 teleblock patterns.


Nightfirecat:
What are the chat messages you got from testing this?

Matthew-nop:
These are the lines debug output for each variant:
`Chat message type GAMEMESSAGE: <col=c356ef>A Tele Block spell has been cast on you. It will expire in 2 minutes.</col>`
`Chat message type GAMEMESSAGE: <col=c356ef>A Tele Block spell has been cast on you. It will expire in 1 minute.</col>`

PR #12471, opened 2020-09-08
discord: Fix action timeout, add in game time elapsed option

### LINKED ISSUE #12438: Discord "Time elapsed" reset when you start training something new.
This is more of a fix but it's where, on the discord rich presence,  training times out after a set amount of time and the time elapsed at the bottom resets. It does not happen for new locations, just different forms of training reset it.

I would like the time elapsed to stay the same no matter what you are doing to make it feel more synced to the gameplay.

Remove the time elapsed possibly? Not sure about any alternatives.

Matthew-nop:
Fixed the discord plugin's action timeout to match expected behavior and clear the activity from presence. Previously if there was only an event removed (such as training herblore), then the presence was not updated. Updating with the current game state fixes this.

I've also added an option to show total time in game rather than time elapsed performing the current activity. This option is the default.
Closes #12438

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
@@ -42,8 +42,8 @@ import net.runelite.api.Varbits;
 enum DiscordGameEventType
 {
 
-	IN_GAME("In Game", -3),
-	IN_MENU("In Menu", -3),
+	IN_MENU("In Menu", -3, false, true, true),
+	IN_GAME("In Game", -3, false, false, true),
 	PLAYING_DEADMAN("Playing Deadman Mode", -3),
 	PLAYING_PVP("Playing in a PVP world", -3),
 	TRAINING_ATTACK(Skill.ATTACK),
@@ -453,6 +453,7 @@ enum DiscordGameEventType
 	private int priority;
 	private boolean shouldClear;
 	private boolean shouldTimeout;
+	private boolean shouldRestart;
 
 	@Nullable
 	private DiscordAreaType discordAreaType;
@@ -485,11 +486,18 @@ enum DiscordGameEventType
 		this.shouldClear = true;
 	}
 
-	DiscordGameEventType(String state, int priority)
+	DiscordGameEventType(String state, int priority, boolean shouldClear, boolean shouldTimeout, boolean shouldRestart)
 	{
 		this.state = state;
 		this.priority = priority;
-		this.shouldClear = true;
+		this.shouldClear = shouldClear;
+		this.shouldTimeout = shouldTimeout;
+		this.shouldRestart = shouldRestart;
+	}
+
+	DiscordGameEventType(String state, int priority)
+	{
+		this(state, priority, true, false, false);
 	}
 
 	DiscordGameEventType(String areaName, DiscordAreaType areaType, Varbits varbits)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
@@ -42,8 +42,8 @@ import net.runelite.api.Varbits;
 enum DiscordGameEventType
 {
 
-	IN_MENU("In Menu", -3, false, true, true),
-	IN_GAME("In Game", -3, false, false, true),
+	IN_MENU("In Menu", -3, true, true, true),
+	IN_GAME("In Game", -3, false, false, false),
 	PLAYING_DEADMAN("Playing Deadman Mode", -3),
 	PLAYING_PVP("Playing in a PVP world", -3),
 	TRAINING_ATTACK(Skill.ATTACK),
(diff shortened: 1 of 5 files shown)


deathbeam:
@Matthew-nop Can you recheck? I cleaned up this a bit and also decided to go with the correct updating instead of force triggering of events. Based on my testing its working fine (note: the total time is now reset after you login/logout, so only total in-game time and total in-menu time is tracked as imo that makes more sense).

Matthew-nop:
@deathbeam thanks for the work on this PR.
I've changed/fixed a few things I found while testing:
- The elapsed time would reset on teleporting due to `checkForAreaEvent()` triggering an `IN_GAME` event. I changed `IN_GAME` to have `shouldRestart` equal false to remedy this.
- The state would not update when moving from a known region (e.g. Al Kharid) to a generic unknown region (In Game). I made `triggerEvent` clear all clearable events on a `IN_GAME` event to fix this. 
- I changed `shouldClear` on `IN_MENU` to true so an `IN_GAME` event will clear it along with all other areaEvents. 
If there's anything you'd like changed please let me know.

CoopJax:
Oh hey, isn't this what I was asking for?

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordState.java
```diff
@@ -115,32 +117,51 @@ void refresh()
...
 			events.add(event);
 		}
 
 		event.setUpdated(Instant.now());
 
-		if (event.getType().isShouldClear())
+		// An IN_GAME (or IN_MENU) event should remove all other areas from events
+		// but since events is fully cleared every time before IN_MENU is added,
+		// we will never need to run this on an IN_MENU event trigger
+		if (event.getType().isShouldClear() || event.getType() == DiscordGameEventType.IN_GAME)
```
deathbeam, line 138: Imo this behaviour should be controlled by enum. Basically what you want here is flag to mark IN_GAME event to be "sticky" or something, e.g it should not be removed when shouldClear event triggers. Or add shouldBeCleared and shouldClear flags where shouldClear will clear other events that need to be cleared and shouldBeCleared will determine if the event will be cleared.

So in this case you would end up with IN_GAME event with shouldBeCleared: false, shouldClear: true

deathbeam:
I tried to incorporate what I mentioned in my comment above in 8ccea05. What i did is:

1. I added `root` flag to game events to avoid checking specifically for IN_GAME or IN_MENU when determining "root" event what should be used for total time tracking
2. I added `shouldBeCleared` flag that determines if event should be cleared when processed
3. I changed original `shouldClear` flag to just determine if event should clear *other* events when triggered

I also added documentation for each flag so hopefully this will be less confusing.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordState.java
```diff
@@ -209,10 +206,10 @@ private void updatePresenceWithLatestEvent()
...
-				// IN_GAME or IN_MENU event as this indicates start of tracking and find last updated one
-				// to determine if we are in game or menu
+				// root event as this indicates start of tracking and find last updated one
+				// to determine correct state we are in
 				startTime = events.stream()
-					.filter(e -> e.type == DiscordGameEventType.IN_GAME || e.type == DiscordGameEventType.IN_MENU)
+					.filter(e -> e.getType().isRoot())
 					.sorted((a, b) -> b.getUpdated().compareTo(a.getUpdated()))
 					.map(EventWithTime::getStart)
 					.findFirst()
```
Matthew-nop: Assuming there's only 1 root event in events list at any given time, this could be simplified.

Matthew-nop (commented):
With the current implementation, it's possible for `events` to contain both an `IN_GAME` and `IN_MENU` event which seems odd given that the player can't concurrently be both in menu and in game.

Would it be a good idea to enforce there being strictly 1 root event in the events list? I'm not sure if it makes sense that there is more than 1 root in the list, and this way we could simplify the code in the switch statement for total time tracking. When processing a new and unique root event we could clear all other root events.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordState.java
deathbeam: Well in-menu theoretically can be active at same time as in-game timer

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
```diff
@@ -451,10 +451,32 @@
...
+
+	/**
+	 * Determines if event start time should be reset when processed
+	 */
 	private boolean shouldRestart;
 
+	/**
+	 * Determines if event should be cleared when processed
+	 */
+	private boolean shouldBeCleared = true;
```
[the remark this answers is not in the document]
deathbeam: processed indicates that its cleared when that happens (as events are processed either when triggered or in the periodic timeout check). and i think shouldBeCleared is more clear, idk

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordState.java
Matthew-nop: Yeah, this is just to clarify which section of code I was referring to in the above post. Although both can be in the events list, is it possible for the player to actually be both in-menu and in-game at the same time?

PR #12488, opened 2020-09-13
runecraft: cleanup

Matthew-nop:
Refactored `updateRifts()` and general cleanup. No functionality changes.


Broooklyn:
In general, you don't want to change keyNames as it would unset everyone's runecraft configs to default

Matthew-nop:
The keyNames themselves should be unchanged. I moved them into variables to reduce coupling between the config and plugin.

Matthew-nop:
I've removed any tangential changes that made the diff large and modified 'updateRifts()'. Hopefully this is a more acceptable PR now.

Nightfirecat (approved):

PR #12495, opened 2020-09-14
inventory tags: improve outline and add configurable thickness

### LINKED ISSUE #6844: Avernic defender is not properly outlined
Title really sums this one up, the inventory tag does not properly outline the avernic defender

https://gyazo.com/f7817dabd582d901bb775c01dc675d71

### LINKED ISSUE #9768: Inventory tag color borders obscure edges of base item sprites [OSB Feature Mirroring]
**Describe the bug**
Inventory tag colors should add a clear outline to item sprites without covering or obscuring any part of the original item sprite. Instead, inventory tag colors cover parts of the edges of items, making the tags look messy. In OSBuddy, these tags outline items only outside of the existing sprites, avoiding the messy look.

**To Reproduce**
1. Add an inventory tag group to any group of items in RuneLite.
2. Add an inventory tag group to the same group of items in OSBuddy.
3. Compare the look and feel of the tags between OSBuddy and RuneLite

**Expected behavior**
The appearance of tags should be identical between OSBuddy and Runelite. Tags look cleaner when they do not obscure the original sprite.

**Actual Result**
The tags in Runelite obscure the original sprite with elements of the original sprite sometimes bleeding out behind the tag outline.

**Screenshots**
Item Tags
Item Tags Close Up

winterdaze:
closes #6844 
closes #9768

Improves inventory tag outlines and adds a config item to choose outline thickness
Before: 
before
After: 
after
After with thickness: 
after2


runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -478,12 +480,13 @@ public AsyncBufferedImage getImage(int itemId, int quantity, boolean stackable)
 	 * @param itemId item id
 	 * @param itemQuantity item quantity
 	 * @param outlineColor outline color
+	 * @param thickness thickness of outline
 	 * @return image
 	 */
-	private BufferedImage loadItemOutline(final int itemId, final int itemQuantity, final Color outlineColor)
+	private BufferedImage loadItemOutline(final int itemId, final int itemQuantity, final Color outlineColor, final int thickness)
```
Nightfirecat, line 486: Maybe instead of adding a breaking change to this API, add an overload with the default being 1 thickness.

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -478,12 +480,13 @@ public AsyncBufferedImage getImage(int itemId, int quantity, boolean stackable)
...
 	{
-		final SpritePixels itemSprite = client.createItemSprite(itemId, itemQuantity, 1, 0, 0, true, 710);
-		return itemSprite.toBufferedOutline(outlineColor);
+		final BufferedImage outline = ImageUtil.outlineImage(getImage(itemId, itemQuantity, false), outlineColor, false, thickness);
+		return outline;
```
Nightfirecat, line 489:
```suggestion
		return ImageUtil.outlineImage(getImage(itemId, itemQuantity), outlineColor, false, thickness);
```

The default for `outlineCorners` is false already. This also does not need to be saved to a variable since it is simply returned on the next line.

runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -326,6 +326,22 @@ public static BufferedImage outlineImage(final BufferedImage image, final Color
...
+	 * corners in addition to edges.
+	 *
+	 * @param image          The image to be outlined.
+	 * @param color          The color to use for the outline.
+	 * @param outlineCorners Whether to draw an outline around corners, or only around edges.
+	 * @param thickness      How thick the outline is.
+	 * @return               The BufferedImage with its edges--and optionally, corners--outlined
+	 *                       with the given color.
+	 */
+	public static BufferedImage outlineImage(final BufferedImage image, final Color color, final Boolean outlineCorners, final int thickness)
```
Nightfirecat, line 340: Another overload of signature `BufferedImage image, Color color, int thickness` can be added to reduce verbisoty in usage.

Broooklyn:
I've noticed the normal white outline from the `Use` option doesn't work with this enabled, i.e., there is no way to tell if you're using an item on another without checking the text in the top left corner or the tooltip.

javaw_JpcxKV1jfL

winterdaze:
> I've noticed the normal white outline from the `Use` option doesn't work with this enabled, i.e., there is no way to tell if you're using an item on another without checking the text in the top left corner or the tooltip.
> 
> javaw_JpcxKV1jfL

it looks like a mess in the old version too. where is the outline behavior of 'use' defined?
img

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
winterdaze: the `false` you removed is for `stackable`, not `outlineCorners`
stackable should be false so the quantity text doesn't get outlined as well

Broooklyn:
It seems like this is taking priority over any change to the inventory icons, not just the `Use` white outline. I've noticed that it also doesn't darken/fade the sprite when you click to equip, for example.

winterdaze:
fixed the issue with the overlay/outline taking priority over interactions. also, the white outline from 'use' shows more now, even with a thicker border. is it good enough?
Updated:
updatedUse
Updated with thickness:
updatedUseThicker
Old:
oldUse
as you can see in the old version, you could only see like two pixels of the white 'use' outline

winterdaze:
updated the PR, it now fixes rendering text over the outline

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsOverlay.java
```diff
@@ -63,8 +64,29 @@ public void renderItemOverlay(Graphics2D graphics, int itemId, WidgetItem itemWi
...
 				if (displayMode == DisplayMode.OUTLINE)
 				{
-					final BufferedImage outline = itemManager.getItemOutline(itemId, itemWidget.getQuantity(), color);
+					final int quantity = itemWidget.getQuantity();
+					final BufferedImage outline = itemManager.getItemOutline(itemId, quantity, color);
 					graphics.drawImage(outline, (int) bounds.getX(), (int) bounds.getY(), null);
+					if (itemManager.getItemComposition(itemId).isStackable())
+					{
+						graphics.setFont(FontManager.getRunescapeSmallFont());
+						String quantityText = Integer.toString(quantity);
```
Nightfirecat, line 73: This strikes me as something which should be added to QuantityFormatter's API instead of only being used locally here.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/util/QuantityFormatter.java
```diff
@@ -145,6 +145,29 @@ public static synchronized String quantityToRSDecimalStack(int quantity, boolean
...
+		if (quantity >= 10000000)
+		{
+			quantityText = quantityText.substring(0, quantityText.length() - 6) + "M";
+
+		}
+		else if (quantity >= 100000)
+		{
+			quantityText = quantityText.substring(0, quantityText.length() - 3) + "K";
+		}
+		return quantityText;
```
Nightfirecat, line 168: Avoiding multiple string operations would be better here.
```java
if (quantity >= 10_000_000)
{
	return Integer.toString(quantity / 1_000_000) + 'M';
}
else if (quantity >= 100_000)
{
	return Integer.toString(quantity / 1_000) + 'K';
}

return Integer.toString(quantityText);
```

Nightfirecat (approved):

Nightfirecat:
I've rebased this and reordered the commits to make more sense. This looks fine to me, and does seem to be identical to ingame item text. :+1:

PR #12517, opened 2020-09-20
Fix Nightmare Zone Overload timer and notification

### LINKED ISSUE #12493: Nightmare Zone overload timer and notification are not cleared when leaving
The overload infobox timer is not cleared when leaving the instance. Additionally, the overload expiry notification can trigger immediately upon re-entering NMZ which also should not happen. Both of these features should have their times reset or cleared when leaving the NMZ instance

spatiag:
Close #12493 

The timers plugin was not checking to see if the player was still inside NMZ, causing the timer to stay when the player left or hopped worlds. Now checks to see if the player is inside the NMZ region.

For the nightmarezone plugin, `overloadNotificationSend = true` when the player drinks an Overload potion. When the player leaves NMZ before the Overload expires and re-enters, `overloadNotificationSend` is still `true`. This results in the notification still being sent based on when the potion was drunk in the previous NMZ instance. Now sets `overloadNotificationSend` to `false` upon leaving.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
@@ -136,6 +136,8 @@ public class NightmareZonePlugin extends Plugin
 				resetPointsPerHour();
 			}
 
+			overloadNotificationSend = false;
+
 			return;
 		}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
@@ -89,6 +89,7 @@ public class NightmareZonePlugin extends Plugin
 		overlayManager.add(overlay);
 		overlay.removeAbsorptionCounter();
 
+		absorptionNotificationSend = true;
 		overloadNotificationSend = false;
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -121,6 +121,7 @@ public class TimersPlugin extends Plugin
 
 	static final int FIGHT_CAVES_REGION_ID = 9551;
 	static final int INFERNO_REGION_ID = 9043;
+	private static final int NMZ_MAP_REGION_ID = 9033;
 	private static final Pattern TZHAAR_WAVE_MESSAGE = Pattern.compile("Wave: (\\d+)");
 	private static final String TZHAAR_DEFEATED_MESSAGE = "You have been defeated!";
 	private static final Pattern TZHAAR_COMPLETE_MESSAGE = Pattern.compile("Your (TzTok-Jad|TzKal-Zuk) kill count is:");
@@ -718,6 +719,11 @@ public class TimersPlugin extends Plugin
 		return client.getMapRegions() != null && ArrayUtils.contains(client.getMapRegions(), INFERNO_REGION_ID);
 	}
 
+	private boolean isInNightmareZone()
+	{
+		return client.getLocalPlayer() != null && client.getLocalPlayer().getWorldLocation().getPlane() > 0 && ArrayUtils.contains(client.getMapRegions(), NMZ_MAP_REGION_ID);
+	}
+
 	private void createTzhaarTimer()
 	{
 		removeTzhaarTimer();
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -708,6 +712,11 @@ private boolean isInInferno()
 		return client.getMapRegions() != null && ArrayUtils.contains(client.getMapRegions(), INFERNO_REGION_ID);
 	}
 
+	private boolean isInNightmareZone()
+	{
+		return client.getMapRegions() != null && ArrayUtils.contains(client.getMapRegions(), NMZ_MAP_REGION_ID);
+	}
```
Nightfirecat, line 718: This should use the same logic as is used in the nightmare zone plugin. I don't think it's an important distinction in this particular case, but this would give false positive results when in KBD's lair

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -143,6 +144,7 @@
 	private int lastAnimation;
 	private boolean loggedInRace;
 	private boolean widgetHiddenChangedOnPvpWorld;
+	private boolean overloadNMZ;
```
Nightfirecat, line 147: What purpose does this serve? We can safely call `removeGameTimer(OVERLOAD)` even if the timer hasn't been added.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
```diff
@@ -135,6 +135,11 @@ public void onGameTick(GameTick event)
 				resetPointsPerHour();
 			}
 
+			if (overloadNotificationSend)
+			{
+				overloadNotificationSend = false;
+			}
```
Nightfirecat, line 141:
```suggestion
			overloadNotificationSend = false;
```

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
spatiag: Sorry, I forgot about NMZ and KBD using the same region. Fixed in a new commit.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
spatiag: I added `overloadNMZ` to mimic how the `tzhaarTimer` is used and to also limit the number of calls to `removeGameTimer(OVERLOAD)`. I originally had this 
```
if (!isInNightmareZone())
{
	removeGameTimer(OVERLOAD);
}
```
but it gets triggered frequently outside NMZ by just walking around and loading in new regions.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
spatiag: I was thinking of doing it this way, but decided to imitate how `absorptionNotificationSend` is handled. 
```
if (!absorptionNotificationSend)
{
	absorptionNotificationSend = true;
}
```

Should we change that block of code to just `absorptionNotificationSend = true` along with `overloadNotificationSend = false`?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: Is it bad for it to be called when walking around? I don't think that's an issue.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
Nightfirecat: I'm not really sure why it was written as it is, but leave it be. This, however, should not be written the same way since it makes no sense.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -800,6 +823,7 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
 
 				removeTzhaarTimer(); // will be readded by the wave message
 				removeGameTimer(TELEBLOCK);
+				removeNMZOverloadTimer();
```
Adam-, line 826: The timer removal only has to be in LOADING I think.

runelite-client/src/main/java/net/runelite/client/plugins/nightmarezone/NightmareZonePlugin.java
Adam-: I've also noticed `absorptionNotificationSend = true` is missing from startUp()

PR #12533, opened 2020-09-22
Replace toScreen function and geometry shader with a projection matrix

Runemoro:
Currently vertices are being projected manually by the `toScreen` shader function, rather than a projection matrix, which is then followed by a second projection using a different matrix. Near-plane clipping is also handled manually using a geometry shader to remove triangles closer than the near plane. 

This PR replaces all of that with a single multiplication by a projection matrix. This simplifies the rendering code, and improves performance by reusing the same matrix for the entire frame, and removing the geometry shader.

I've compared screenshots before and after, and everything looks pixel-perfectly the same.


runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -50,14 +50,15 @@
 import java.nio.IntBuffer;
 import javax.inject.Inject;
 import javax.swing.SwingUtilities;
+
```
abextm, line 53: remove this newline

abextm:
This is somewhat different in that it is near culling per frag, instead of per face. I'm not aware of anywhere that would make this infeasible though. Otherwise looks fine

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1151,6 +1123,16 @@ private void drawFrame()
 			gl.glUniform1f(uniBrightness, (float) textureProvider.getBrightness());
 			gl.glUniform1f(uniSmoothBanding, config.smoothBanding() ? 0f : 1f);
 
+			// Calculate projection matrix
+			Matrix4 matrix = new Matrix4();
+			matrix.scale(0.04f / viewportWidth, 0.04f / viewportHeight, 1);
+			matrix.makeFrustum(-1, 1, -1, 1, 50f, 10000);
```
abextm, line 1129: this can cause far clipping in some cases, 20k should be safe (scene is 128*104 units in each dimension)

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -1151,6 +1123,16 @@ private void drawFrame()
...
+			matrix.scale(client.getScale(), client.getScale(), 1);
+			matrix.rotate((float) (Math.PI - pitch * Math.PI / 1024), -1, 0, 0);
```
Adam-, line 1131: We have the jau-to-rad unit defined as `Perspecta.UNIT`

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/comp_common.glsl
```diff
@@ -38,6 +38,8 @@
    ivec2 sinCosTable[2048];
  };
 
+uniform mat4 projectionMatrix;
```
Adam-, line 41: The projection matrix in the compute shader is uninitialized so this completely breaks all of the face sorting.

image

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
Adam-, line 1131: We have the jau-to-rad constant pi/1024 defined as `Perspective.UNIT`

runelite-client/src/main/resources/net/runelite/client/plugins/gpu/comp_common.glsl
Runemoro: I might be blind... but what exactly is wrong in that pic? Anyway, I fixed initializing the projection matrix.

Adam-:
added in 8c00f6da8847f45b7dac4ba0537cf54a218762d2

PR #12534, opened 2020-09-22
xpupdater: remove onGameTick

Matthew-nop:
Changes the updating of `lastXp` from checking each game tick if an update is needed to just update when you've logged in/out.


Matthew-nop:
In retrospect, this code needlessly complicates the plugin and isn't a clear improvement. Closing.

PR #12536, opened 2020-09-22
chatcommands: add shorthand names for all agility courses

### LINKED ISSUE #12529: !kc rooftop
could you add !kc to the same rooftops as ape atoll and prifddinas? thank you

Broooklyn:
Adds shorthand names for all agility courses, including common misspellings.

I have confirmed each `longBossName` in game, as they don't all match what is shown when talking to Grace.

Closes #12529


makingcomment:
Maybe Pollnivneach should also include poly and/or polly, and maybe Gnome Stronghold should also include gnome.

dekvall:
Imo we should not normalize the usage of misspelled versions of city names such as `canafis`, `relekka` or `polnivneach`.

Adam-:
I've removed the misspellings.

PR #12544, opened 2020-09-23
XP Tracker: Goal time formatting options

d-as:
Add option to display XP goal times in hours rather than days, since some people find it easier to visualize goals that way. This means hours can go above 23 in XP goals. In addition, add an option for showing the current percentage of an XP goal in the progress bar tooltip.

Preview


makingcomment:
I think the "101 hours 26:24" in your example screenshot would look better as "101:26:24".

PR #12559, opened 2020-09-26
Add birdhouses to loot tracker plugin

SirGirion:
Adds actual birdhouses to the loot tracking plugin. I wasn't sure if Jagex puts a comma in the XP drop, they don't for redwoods but I didn't test for yew and magics. No reason to think they would but made the regex account for it anyways.


Adam-:
Tests for these messages would be good to add, and then combine the 3 regexes just into one.

PR #12567, opened 2020-09-29
Bugfix: Update sprite ID MINIMAP_ORB_RUN_ICON from 1070 to 1069

andreykaipov:


andreykaipov:
Corresponding widget is 160.27 (id 10485787) if one would like to confirm :eye::eye:

Adam-:
It looks like someone else has since added `MINIMAP_ORB_WALK_ICON = 1069;` for statusbars, so this should probably just be `MINIMAP_ORB_WALK_ICON_ACTIVATED`?

Adam-:
I removed the duplicated symbol in 20dec72416f984dcc2bc8c8bfcf2cf5c961b99aa

PR #12578, opened 2020-10-01
wiki: Always include item names in widget optarget lookups

abextm:


runelite-api/src/main/java/net/runelite/api/Constants.java
```diff
@@ -114,4 +114,9 @@
 	 * @see ItemComposition#getPrice
 	 */
 	public static final float HIGH_ALCHEMY_MULTIPLIER = .6f;
+
+	/**
+	 * The standard color of items in the right click context menu
+	 */
+	public static final String ITEM_COLOR = "<col=ff9040>";
```
Nightfirecat, line 121: This is just `ColorUtil.colorTag(JagexColors.MENU_TARGET)`

Issue #12283, opened 2020-08-03
Setting to show Chambers of Xeric health overlay only in certain rooms


sadegr:
**Is your feature request related to a problem? Please describe.**
The CoX health overlay that was added to the vanilla client can only be toggled on or off for all rooms. I personally only want it for vanguards, and I'm sure others would also like the freedom to only see it in certain rooms of their choosing.

**Describe the solution you'd like**
A new setting in the Chambers of Xeric plugin called something like "Health overlay exceptions" which would allow you to configure in which rooms the overlay would *not* be shown. This could work similarly to the Blacklisted rooms setting - a comma separated list of rooms is entered by the user, and the health overlay will not be visible in any rooms that are in the list. This is just one way it could work, but I'm open to any settings scheme/naming that would allow you to toggle the health overlay on or off for particular rooms.

A related, but different issue: #12230

Issue #12230, opened 2020-07-26
Chambers of Xeric Opponent info changes


Kurama2121:
With the addition of native health bars in chambers of xeric it would be nice to be able to not have to turn the Opponent info plugin off while in certain rooms of chambers of xeric of free up screen space. In rooms like Tekton or Olm, there is a built in HP tracker so the opponent info health bar isnt needed (see image) but in guardians where there isnt a native health bar, the opponent info bar would still be needed.

![unknown](https://user-images.githubusercontent.com/39833426/88473510-f2863080-cee3-11ea-8e1a-f69591ea958c.png)

makingcomment:
Related (but not quite the same): #12137

Hydrox6:
Duplicate of #5570

Issue #12137, opened 2020-07-14
Changing Runelites enemy health bar for CoX


TheTargets:
Since Runescape itself added the ability to see the exact value of health on each boss in Chambers of Xerics. Would we be able to have it changed instead of seeing a % of the health.

![image](https://user-images.githubusercontent.com/9783476/87418958-8bef4180-c5ca-11ea-81f8-e1552adc9d7e.png)

PR #12585, opened 2020-10-02
Add the ability to merge the raid hp bar overlay with the opponent info overlay

CasvM:
### before:
image

### after: 
image

This would essentially close these isssues: https://github.com/runelite/runelite/issues/12283 https://github.com/runelite/runelite/issues/12230 https://github.com/runelite/runelite/issues/12137

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -304,6 +304,8 @@ public enum Varbits
 	PERSONAL_POINTS(5422),
 	RAID_PARTY_SIZE(5424),
 
+	RAID_HEALTH_BAR_OVERLAY_VALUE(6099),
+
 	// 0 = raid not started, >0 = raid started
 	RAID_STATE(5425),
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
@@ -63,4 +63,15 @@ public interface OpponentInfoConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "mergeRaidOverlay",
+			name = "Merge raid health bar overlay",
+			description = "Hides the raid health bar overlay, but adds the info to the existing opponent info",
+			position = 4
+	)
+	default boolean mergeRaidOverlay()
+	{
+		return true;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
@@ -31,12 +31,18 @@ import java.awt.FontMetrics;
 import java.awt.Graphics2D;
 import java.awt.Point;
 import java.awt.Rectangle;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import net.runelite.api.Actor;
 import net.runelite.api.Client;
+
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
 import net.runelite.api.NPC;
 import net.runelite.api.Player;
+import net.runelite.api.widgets.Widget;
+import net.runelite.client.callback.ClientThread;
 import net.runelite.client.game.HiscoreManager;
 import net.runelite.client.game.NPCManager;
 import static net.runelite.client.ui.overlay.OverlayManager.OPTION_CONFIGURE;
@@ -60,11 +66,16 @@ class OpponentInfoOverlay extends OverlayPanel
 	private final OpponentInfoConfig opponentInfoConfig;
 	private final HiscoreManager hiscoreManager;
 	private final NPCManager npcManager;
+	private final ClientThread clientThread;
 
 	private Integer lastMaxHealth;
 	private int lastRatio = 0;
 	private int lastHealthScale = 0;
 	private String opponentName;
+	private int raidOpponentHealth = 0;
+	private int raidOpponentMaxHealth = 0;
+	private String raidOpponentName;
+	private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
 
 	@Inject
 	private OpponentInfoOverlay(
@@ -72,7 +83,8 @@ class OpponentInfoOverlay extends OverlayPanel
 		OpponentInfoPlugin opponentInfoPlugin,
 		OpponentInfoConfig opponentInfoConfig,
 		HiscoreManager hiscoreManager,
-		NPCManager npcManager)
+		NPCManager npcManager,
+		ClientThread clientThread)
 	{
 		super(opponentInfoPlugin);
 		this.client = client;
@@ -80,6 +92,7 @@ class OpponentInfoOverlay extends OverlayPanel
 		this.opponentInfoConfig = opponentInfoConfig;
 		this.hiscoreManager = hiscoreManager;
 		this.npcManager = npcManager;
+		this.clientThread = clientThread;
 
 		setPosition(OverlayPosition.TOP_LEFT);
 		setPriority(OverlayPriority.HIGH);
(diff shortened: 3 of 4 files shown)


Hydrox6:
How does this work with Vanguards, considering they have a really overengineered health bar that tries to show all 3 at once iirc (#12283 would want vanilla for *just* vanguards, likely because of this)

Hydrox6 (changes requested):
You've got a lot of checkstyle violations. You can use this page to get our style guide and import our formatter

CasvM:
I've added the ability to force the vanilla HP bar by typing it in the config:
image

I've also added the percentage to the vanilla HP bar 
image

I'm not sure why the checks are failing, I formatted the code before submitting.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -164,6 +165,43 @@ public void onGameTick(GameTick gameTick)
...
+			lastHealth = client.getVar(Varbits.RAID_HEALTH_BAR_OVERLAY_VALUE);
+			opponentInfoOverlay.updateRaidVars(lastHealth);
+			update = false;
+		}
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (event.getGroup().equals("opponentinfo") || event.getGroup().equals("runelite"))
```
Nightfirecat, line 187: Why is `runelite` in here? I don't really understand this at all as written.

runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -304,6 +304,8 @@
 	PERSONAL_POINTS(5422),
 	RAID_PARTY_SIZE(5424),
 
+	RAID_HEALTH_BAR_OVERLAY_VALUE(6099),
```
Nightfirecat, line 307: What value does this hold? Seems like it's an int of the current health of the monster? Add a comment to describe this.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
```diff
@@ -63,4 +63,39 @@ default boolean showOpponentsInMenu()
...
 	}
+
+	@ConfigItem(
+		keyName = "showVanillaPercentages",
+		name = "Show HP % on the Vanilla HP Bar",
+		description = "Adds a percentage value next to the health value of the vanilla HP Bar",
+		position = 4
+	)
+
+	default boolean showVanillaPercentages()
```
Nightfirecat, line 74: Isn't this already covered by `HitpointsDisplayStyle.BOTH`?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -31,15 +31,25 @@
...
+
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
+
 import net.runelite.api.NPC;
 import net.runelite.api.Player;
+import net.runelite.api.widgets.Widget;
+import net.runelite.client.callback.ClientThread;
 import net.runelite.client.game.HiscoreManager;
 import net.runelite.client.game.NPCManager;
+
```
Nightfirecat, line 50: Don't add empty lines in the import block.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -60,26 +71,35 @@
...
 	private final HiscoreManager hiscoreManager;
 	private final NPCManager npcManager;
+	private final ClientThread clientThread;
 
 	private Integer lastMaxHealth;
 	private int lastRatio = 0;
 	private int lastHealthScale = 0;
 	private String opponentName;
+	private int raidOpponentHealth = 0;
+	private int raidOpponentMaxHealth = 0;
```
Nightfirecat, line 81: int fields default to `0`, you don't need to explicitly set that starting value.
```suggestion
	private int raidOpponentHealth;
	private int raidOpponentMaxHealth;
```

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -60,26 +71,35 @@
...
+	private String raidOpponentName;
+	private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
+	boolean useVanilla = false;
+	boolean shouldUnHide = false;
```
Nightfirecat, line 85: boolean fields default to `false`, you don't need to explicitly set that starting value.
```suggestion
	boolean useVanilla;
	boolean shouldUnHide;
```

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -60,26 +71,35 @@
...
+	private final ClientThread clientThread;
 
 	private Integer lastMaxHealth;
 	private int lastRatio = 0;
 	private int lastHealthScale = 0;
 	private String opponentName;
+	private int raidOpponentHealth = 0;
+	private int raidOpponentMaxHealth = 0;
+	private String raidOpponentName;
+	private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
Nightfirecat, line 83: SES should be injected, not instantiated here.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -89,44 +109,174 @@ private OpponentInfoOverlay(
...
 
+	public void updateRaidVars(int lastHealth)
+	{
+		if (opponentInfoConfig.mergeRaidOverlay() || opponentInfoConfig.showVanillaPercentages())
+		{
+			clientThread.invokeLater(() ->
+			{
+				Widget raidOpponentWrapper = client.getWidget(303, 6);
+				Widget raidOpponentNameWidget = client.getWidget(303, 11);
+				Widget raidOpponentHealthWidget = client.getWidget(303, 20);
```
Nightfirecat, line 120: These should all be defined in WidgetID and WidgetInfo.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -89,44 +109,174 @@ private OpponentInfoOverlay(
...
+				{
+					//if lastHealth is zero, either the Raid Health Bar is supposed to disappear,
+					//or the opponent is dead, so remove the raid opponent name and max health vars
+					raidOpponentName = null;
+					raidOpponentMaxHealth = 0;
+
+					if (raidOpponentWrapper != null)
+					{
+						//unhide so that initial values can be read again, but wait 7 seconds so that the hp bar isn't shown
+						//while the boss is dying. It's possible to cancel the unhide by changing shouldUnHide back to false
```
Nightfirecat, line 172: What does this mean? Is the varbit for the opponent's health not updated when the widget is hidden?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -89,44 +109,174 @@ private OpponentInfoOverlay(
...
+		if (useVanilla && opponentInfoConfig.mergeRaidOverlay())
 		{
-			opponentName = null;
 			return null;
 		}
 
-		if (opponent.getName() != null && opponent.getHealthScale() > 0)
+		final Actor opponent = opponentInfoPlugin.getLastOpponent();
+
+		if (opponent != null)
```
Nightfirecat, line 216: Why is this guard clause reversed?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
```diff
@@ -154,8 +304,12 @@ else if (opponent instanceof Player)
 				// This is the reverse of the calculation of healthRatio done by the server
 				// which is: healthRatio = 1 + (healthScale - 1) * health / maxHealth (if health > 0, 0 otherwise)
 				// It's able to recover the exact health if maxHealth <= healthScale.
-				int health = 0;
-				if (lastRatio > 0)
+
+				//raidOpponentHealth is non-zero if the raid health overlay is active;
+				//The max and min health don't have to be calculated then
+				int health = raidOpponentHealth;
```
Nightfirecat, line 310: What about when the health overlay isn't active?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -26,25 +26,20 @@
...
-import net.runelite.api.MenuAction;
-import net.runelite.api.MenuEntry;
-import net.runelite.api.NPC;
-import net.runelite.api.WorldType;
-import net.runelite.api.events.GameStateChanged;
-import net.runelite.api.events.GameTick;
-import net.runelite.api.events.InteractingChanged;
-import net.runelite.api.events.MenuEntryAdded;
+import net.runelite.api.*;
+import net.runelite.api.events.*;
```
Nightfirecat, line 38: Reverse these wildcard imports.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -164,6 +165,43 @@ public void onGameTick(GameTick gameTick)
 		}
 	}
 
+	@Subscribe
+	public void onVarbitChanged(VarbitChanged event)
+	{
+		if ((client.getVar(Varbits.RAID_HEALTH_BAR_OVERLAY_VALUE) != lastHealth) || update ||
+			(config.showVanillaPercentages() && (event.getIndex() == 1682 || event.getIndex() == 1683)))
```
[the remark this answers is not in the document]
abextm: VarbitChanged::getIndex is not a varbit id; this is probably wrong

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -164,6 +165,43 @@ public void onGameTick(GameTick gameTick)
 		}
 	}
 
+	@Subscribe
+	public void onVarbitChanged(VarbitChanged event)
+	{
+		if ((client.getVar(Varbits.RAID_HEALTH_BAR_OVERLAY_VALUE) != lastHealth) || update ||
```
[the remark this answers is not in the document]
abextm: VarbitChanged::getIndex is not a varbit id; changing this would be wrong

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
CasvM: Done in the latest commit. Apologies, I think intellij snuck that in

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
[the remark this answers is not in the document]
CasvM: Currently it isn't checking values at all anymore, I could make it like before where it checks if one of the values is changed, but I don't know if that's wanted

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
CasvM: When the plugin itself is enabled or disabled through the config it should also remove/unhide the health bar, but I forgot I could do that through the startup and shutdown. I've removed the runelite part in the latest commit

runelite-api/src/main/java/net/runelite/api/Varbits.java
CasvM: I've added another varbit for the max health of a monster, and a comment to describe these varbits

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
CasvM: While I could make it so that the value on the vanilla health bar is equal to that of the opponent information, I thought someone might want just the percentage on the vanilla health bar, and something else on the opponent information panel, that's why this toggle exists

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
CasvM: With the new code, using varbits instead of the widget, I don't need the scheduler anymore, so I've removed it. Also all the formatting and initialization guidelines you pointed out are added in the latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
CasvM: This was the case. When the widget is hidden, the values on the widget don't update. The varbits still do update though, and that is used for the variables in the latest commit.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
CasvM: When the vanilla bar and the opponentinfo are merged, the opponentinfo shouldn't disappear after 5 seconds, or when the opponent is null, since the variables are still known from the vanilla bar. When opponent is null, it shouldn't return yet, but rather go on and check if the raid variables are known. There is an additional gaurd clause which checks if the raid variables are null after this guard clause:
`		if ((opponent == null && (vanillaHealthBarMaxHealth == 0 || !opponentInfoConfig.mergeVanillaHealthBar())) ||
			opponentName == null)
		{
			opponentName = null;
			return null;
		}`

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoOverlay.java
CasvM: By now I've changed this to `int health = opponentMaxHealth`, so that it only runs when the overlay actually isn't active. Previously it could also run when the opponent simply had zero health. 

But to answer your question: Both the opponenthealth and raidopponenthealth are zero when the overlay isn't active.

CasvM:
#### Notable changes in the latest commit:

- The 'merge vanilla health bar' feature also works outside of COX.

- The opponent information overlay now takes on the same colours as the boss health bar overlay:
image
image
image

Nightfirecat:
Aside from code review, I'm not sure what the general opinion is around changing the healthbar colors. I'd think it would be better to not change them since, with the vanilla bar hidden, it would seemingly change for no reason.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -164,6 +190,62 @@ public void onGameTick(GameTick gameTick)
...
+				}
+			}
+			//reset/check values after config change.
+			clientThread.invokeLater(() ->
+			{
+				opponentInfoOverlay.setVanillaHealthText();
+				opponentInfoOverlay.setHealthBarColors();
+			});
+
+			updateVanillaHealthBarVars();
```
Nightfirecat, line 245: Should these all be running even if mergeVanillaHealthBar wasn't the changed config within opponentinfo?

Nightfirecat (changes requested):

ulises-c:
This is just what I was looking for

destruct7780:
Can this be implemented already? it's crazy we still can't have exact hp values on the runelite hp bar i can't stand the official one

PR #12586, opened 2020-10-02
Render Ground Items tile overlay at correct height

Unmoon:
Added `zOffset` argument for `Perspective.getCanvasTilePoly` and `Perspective.getCanvasTileAreaPoly` to allow Ground Items to render the tile overlay at the correct height (for example tables).
image

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Perspective.java
+++ b/runelite-api/src/main/java/net/runelite/api/Perspective.java
@@ -365,6 +365,20 @@ public class Perspective
 	}
 
 	/**
+	 * Calculates a tile polygon from offset worldToScreen() points.
+	 *
+	 * @param client the game client
+	 * @param localLocation local location of the tile
+	 * @param zOffset offset from ground plane
+	 * @return a {@link Polygon} on screen corresponding to the given
+	 * localLocation.
+	 */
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityOverlay.java
```diff
@@ -161,7 +161,7 @@ public Dimension render(Graphics2D graphics)
 				int size = npcComposition.getSize();
 				LocalPoint lp = npc.getLocalLocation();
 
-				Polygon tilePoly = Perspective.getCanvasTileAreaPoly(client, lp, size);
+				Polygon tilePoly = Perspective.getCanvasTileAreaPoly(client, lp, size, 0);
```
Raminios, line 164: Surely an extra overload method in `Perspective.java`

`public static Polygon getCanvasTileAreaPoly(
		@Nonnull Client client,
		@Nonnull LocalPoint localLocation,
		int size,
		int zOffset)
	{
    getCanvasTileAreaPoly(client, localLocation, size, 0);
}
`

Is better than adding a zero on every usage?

deathbeam (approved):

PR #12588, opened 2020-10-03
discord: add random events regions

Broooklyn:
Adds discord regions for all instanced random events.  The Pinball, Beekeeper, and Gravedigger randoms are located in the same region square, so I have opted to include those within ScapeRune as I don't think there is a better way to describe those if they're together.


abextm:
Are you sure the pinball random is ScapeRune?

deathbeam:
Can you also add that as comment above the pinball/gravedigger/beekeper region definition @Broooklyn ?

Issue #11403, opened 2020-04-29
screenshot doesn't capture full screen with gpu on and dpi scaling at 125%


kobyhinnant:
**Describe the bug**
Taking a screenshot with windows dpi scaling set to 125% doesn't capture the full game area unless GPU plugin is turned off

OS: Windows 10
RuneLite version: 1.6.12
Launcher version: 2.1.0

**To Reproduce**
Set windows 10 dpi scaling to be 125%
Turn on GPU plugin
Launch runelite
Take a screenshot with the screenshot plugin

**Expected behavior**
The screenshot plugin should capture the full game regardless of GPU being turned on or off

**Screenshots**
With GPU turned on
![image](https://user-images.githubusercontent.com/42243069/80556625-0ecc0d00-8989-11ea-9050-74d291291492.png)

With GPU turned off
![image](https://user-images.githubusercontent.com/42243069/80556633-1d1a2900-8989-11ea-84e0-26fa4a1c8271.png)

Alexsuperfly:
idk why you are flip flopping launcher versions but this should be fixed in 2.1 so just stick with that and stop using the 2.0 one randomly

kobyhinnant:
meant to write 2.1, i'm not using 2.0

kobyhinnant:
> idk why you are flip flopping launcher versions but this should be fixed in 2.1 so just stick with that and stop using the 2.0 one randomly

kobyhinnant:
> idk why you are flip flopping launcher versions but this should be fixed in 2.1 so just stick with that and stop using the 2.0 one randomly

meant to write 2.1, i'm not using 2.0, edited now

raiyni:
This is fixed for me in 2.1

Biice:
@kobyhinnant Slightly off-topic, but when using windows scaling at e.g. 125%, does your Runescape also turn blurry? The new UI scaling options under the _GPU_ plugin make it somewhat more bearable, but I still find it virtually unplayable. Feels like I need glasses playing with windows scaling—due to the blurriness, not the size of the launcher of course, which is where windows scaling helps.

kobyhinnant:
[...]

PR #12594, opened 2020-10-03
Fix screenshot with hidpi scaling with GPU plugin

### LINKED ISSUE #12463: Screenshot size issue with GPU and Stretched Mode plugins on Linux
Note: [11403](https://github.com/runelite/runelite/issues/11403) is the same issue for Windows, not Linux.

**Describe the bug**
When using RuneLite on Linux with GPU and Stretched Mode plugins, screenshot functionality only captures ~25% of the game window.

**To Reproduce**
1. Use RuneLite on Linux
2. Enable GPU plugin
3. Enable Stretched Mode plugin
4. Take screenshot

**Expected behavior**
Full-size client screenshot of entire window.

**Screenshots**
[https://imgur.com/a/j29YI4B](https://imgur.com/a/j29YI4B)

 - OS: Pop!_OS 20.04
 - RuneLite version: 1.6.26.1
 - Launcher version: 2.1.3

**Additional context**
I am using the .jar file on a 4K monitor with 200% desktop scaling.

jacobmischka:
The ClientUI changes are somewhat gross and non-DRY, but it seemed better than overengineering more abstractions.

Fixes #12463

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -1303,6 +1303,13 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 			height = dim.height;
 		}
 
+		if (OSType.getOSType() != OSType.MacOS)
+		{
+			final AffineTransform t = ((Graphics2D) canvas.getGraphics()).getTransform();
+			width = getScaledValue(t.getScaleX(), width);
+			height = getScaledValue(t.getScaleY(), height);
+		}
+
 		ByteBuffer buffer = ByteBuffer.allocateDirect(width * height * 4)
 			.order(ByteOrder.nativeOrder());
 


jacobmischka:
Ah whoops, sorry about those checkstyle fails.

Adam-:
This should probably look something similar to what `glDpiAwareViewport` looks like.

jacobmischka:
Ah nice, I missed that bit. You're right, that is cleaner and better.

Adam-:
Are you able to reproduce this problem currently, and are you using Linux?

jacobmischka:
Yes, and I am, yes. Both of the following screenshots were taken with GPU and Stretched plugins enabled, on Linux, with GDK_SCALE=2.

Before this change:



After this change:

jacobmischka:
I don't have a Windows machine to test right now, but I imagine this should also fix #11403.

Adam-:
I've tested this some and this breaks Windows with UI scaling on. I will look into it more.

jacobmischka:
Oh darn, breaks how?

Adam-:
So master is broken on Windows with scaling, but this PR is also broken, but they are broken differently. Which is odd.

This pr does:



and master does:

jacobmischka:
Interesting. I'll try to get a Windows machine set up and look into it soon.

Adam-:
For additional levels of confusion, *release* on *release launcher* works correctly, as well as master on release launcher and this PR on release launcher. I believe this is due to https://github.com/runelite/launcher/commit/123bb6590643e6e645dbd1cd070f00b1a8c16cae which is what allows RuneLite to UI scale on Windows as an application (the commit removes the dpi awareness flag from the RuneLite.exe manifest, resulting in Windows thinking RuneLite isn't DPI aware). This is due to the Java 8/11 DPI awareness stuff being some combination of completely missing or not working and we can't really change it regardless. 

So we can maybe merge this since it doesn't seem to break anything more. It would be nice to make this work outside of the launcher I suppose but I think it might be some other issues.

PR #12595, opened 2020-10-04
Imageutil log path

### LINKED ISSUE #10708: ImageUtil.getResourceStreamFromClass does not log attempted path if load fails
As it currently stands, the method will (after a lengthy set of stacktraces) yield the following exceptions:

```
Caused by: java.lang.IllegalArgumentException: file.png
 
Caused by: java.lang.IllegalArgumentException: input == null!
```

It would be helpful to developers to print the searched path so they can more easily figure out why their images are not being loaded.

Trentr9:
Add debug logging into ImageUtil#getResourceStreamFromClass to allow easier troubleshooting

Closes #10708

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
@@ -373,6 +373,7 @@ public class ImageUtil
 	 */
 	public static BufferedImage getResourceStreamFromClass(final Class c, final String path)
 	{
+		log.debug("Attempting to load image from class: {}, path: {}", c.getName(), path);
 		try
 		{
 			synchronized (ImageIO.class)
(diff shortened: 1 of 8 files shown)


runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -373,6 +373,7 @@ public static BufferedImage outlineImage(final BufferedImage image, final Color
 	 */
 	public static BufferedImage getResourceStreamFromClass(final Class c, final String path)
 	{
+		log.debug("Attempting to load image from class: {}, path: {}", c.getName(), path);
```
Nightfirecat, line 376: I don't think we need to log all image access attempts, only failed ones.

runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -382,10 +383,12 @@ public static BufferedImage getResourceStreamFromClass(final Class c, final Stri
 		}
 		catch (IllegalArgumentException e)
 		{
+			log.error("Failed to load image from class: {}, path: {}", c.getName(), path);
```
Nightfirecat, line 386: Wouldn't this just log something like `Failed to load image from class: ExamplePlugin, path example.png`? That would not be helpful since it would not describe the physical file path where the attempted load would have occurred.

Nightfirecat (changes requested):

deathbeam:
Implemented in https://github.com/runelite/runelite/pull/12631

PR #12602, opened 2020-10-05
Add Fill Option for Inventory Tags

### LINKED ISSUE #12518: cool new kind of inventory tags
i have really bad eyesight and i hate how all the inventory tags options look and i saw a screenshot of someone having this kind of invy tags which look really cool

unknown

they also had like an opacity meter so you could have the opacity of the invy tag color go from like 0 to 255
this is a really cool idea imo please consider this

1jz:
Closes #12518

Includes a spinner for alpha values from 0-255

Looks like this at 100:
filled tags

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
@@ -28,20 +28,74 @@ import java.awt.Color;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
+import net.runelite.client.config.Range;
 
-@ConfigGroup("inventorytags")
+@ConfigGroup(InventoryTagsConfig.GROUP)
 public interface InventoryTagsConfig extends Config
 {
-	enum DisplayMode
-	{
-		OUTLINE,
-		UNDERLINE
-	}
-
 	String GROUP = "inventorytags";
+
+	@ConfigSection(
+		name = "Tag display mode",
+		description = "How tags are displayed in the inventory",
+		position = 0
+	)
+	String tagStyleSection = "tagStyleSection";
 
 	@ConfigItem(
 		position = 0,
+		keyName = "showTagOutline",
+		name = "Outline",
+		description = "Configures whether or not item tags show be outlined",
+		section = tagStyleSection
+	)
+	default boolean showTagOutline()
+	{
+		return true;
+	}
+
+	@ConfigItem(
+		position = 1,
+		keyName = "tagUnderline",
+		name = "Underline",
+		description = "Configures whether or not item tags should be underlined",
+		section = tagStyleSection
+	)
+	default boolean showTagUnderline()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "tagFill",
+		name = "Fill",
+		description = "Configures whether or not item tags should be filled",
+		section = tagStyleSection
+	)
+	default boolean showTagFill()
+	{
+		return false;
+	}
+
+	@Range(
+		max = 255
+	)
+	@ConfigItem(
+		position = 3,
+		keyName = "fillOpacity",
+		name = "Fill opacity",
+		description = "Configures the opacity of the tag \"Fill\"",
+		section = tagStyleSection
+	)
+	default int fillOpacity()
+	{
+		return 50;
+	}
+
+	@ConfigItem(
+		position = 1,
 		keyName = "groupColor1",
 		name = "Group 1 Color",
 		description = "Color of the Tag"
@@ -52,7 +106,7 @@ public interface InventoryTagsConfig extends Config
 	}
 
 	@ConfigItem(
-		position = 1,
+		position = 2,
 		keyName = "groupColor2",
 		name = "Group 2 Color",
 		description = "Color of the Tag"
@@ -63,7 +117,7 @@ public interface InventoryTagsConfig extends Config
 	}
 
 	@ConfigItem(
-		position = 2,
+		position = 3,
 		keyName = "groupColor3",
 		name = "Group 3 Color",
 		description = "Color of the Tag"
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -28,18 +28,23 @@
...
 
 @ConfigGroup("inventorytags")
 public interface InventoryTagsConfig extends Config
 {
 	enum DisplayMode
 	{
 		OUTLINE,
-		UNDERLINE
+		UNDERLINE,
+		OUTLINE_FILL
```
Nightfirecat, line 40: Instead of adding another option here, I'd advise going the route of multiple toggles, [similar to the approach used in the NPC Indicators Plugin](https://github.com/runelite/runelite/blob/3c7a6380798f9821ec0477ee3746819f69e2897f/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java#L37-L78) so users can optionally enable outline, underline, and/or fill. (Maintaining the opacity option for the fill color is a good idea)

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
1jz: I removed the DisplayMode enums and followed the approach shown in the NPC indicators for multiple toggles. I also lowered the default opacity to 50 because 100 felt little too opaque. Thanks.

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -28,20 +28,78 @@
...
+import net.runelite.client.config.ConfigSection;
 
 @ConfigGroup("inventorytags")
 public interface InventoryTagsConfig extends Config
 {
-	enum DisplayMode
+	String GROUP = "inventorytags";
+
+	int OPACITY_LIMIT_MIN = 0;
+	int OPACITY_LIMIT_MAX = 255;
```
Nightfirecat, line 40: I suppose these can be replaced with `ColorUtil.(MIN|MAX)_RGB_VALUE` given opacity is also represented by the same range

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -28,20 +28,78 @@
...
+
+	@ConfigSection(
+			name = "Tag style",
+			description = "The tag style of highlighted items",
```
Nightfirecat, line 44: This should more closely match the existing display style config.
```suggestion
			name = "Tag display mode",
			description = "How tags are displayed in the inventory",
```

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsOverlay.java
```diff
@@ -57,16 +57,23 @@ public void renderItemOverlay(Graphics2D graphics, int itemId, WidgetItem widget
...
+				if (config.showTagOutline())
 				{
 					final BufferedImage outline = itemManager.getItemOutline(itemId, widgetItem.getQuantity(), color);
 					graphics.drawImage(outline, (int) bounds.getX(), (int) bounds.getY(), null);
 				}
-				else
+
+				if (config.showTagFill())
+				{
+					final Color fillColor = new Color(color.getRed(), color.getGreen(), color.getBlue(), config.getAlphaValue());
```
Nightfirecat, line 71:
```suggestion
					final Color fillColor = ColorUtil.colorWithAlpha(color, config.getAlphaValue());
```

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -28,20 +28,78 @@
...
+
+	@Range(
+			min = OPACITY_LIMIT_MIN,
+			max = OPACITY_LIMIT_MAX
+	)
+	@ConfigItem(
+			position = 2,
+			keyName = "alphaValue",
+			name = "Opacity Level",
+			description = "Configures the opacity of each color",
```
Nightfirecat, line 81: This should indicate it only affects the "Fill" display mode

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -28,20 +28,78 @@
...
+			section = tagStyleSection
+	)
+	default int getAlphaValue()
+	{
+		return 50;
+	}
```
Nightfirecat, line 87: Given the fill + alpha configs must be paired together, I think this order would be better for the modes:

1. Outline
2. Underline
3. Fill
4. Fill opacity

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsConfig.java
```diff
@@ -25,23 +25,79 @@
...
 
+	@Range(
+			min = ColorUtil.MIN_RGB_VALUE,
+			max = ColorUtil.MAX_RGB_VALUE
+	)
 	@ConfigItem(
-		position = 0,
+			position = 3,
+			keyName = "alphaValue",
+			name = "Opacity Level",
```
Nightfirecat, line 90:
```suggestion
			name = "Fill opacity",
```

Nightfirecat (commented):
LGTM otherwise

Nightfirecat (approved):

Adam-:
What's the reasoning behind this being 3 separate options instead of just adding a new one to the enum? Is there need to have multiple enabled at once? Also I think the fillImage `ImageUtil.fillImage(itemManager.getImage(itemId), fillColor)` should probably be cached between frames.

Adam-:
I've figured out that the original picture OP posted is of both fill + outline and not just fill, so it does then make sense to have configurations for both I suppose. I've tested this and it doesn't work with stackable items due to not using the 3-arg getImage() though:

image

PR #12609, opened 2020-10-06
Add woodcutting clickboxes

CarlOlson:
This adds click boxes to trees, similar to the agility plugin. I searched, but couldn't find any references to this being a rejected feature.

I based this off the agility plugin code, so I used the existing copyright and added my name to the new file. (I didn't see how to handle copyright notices in the docs/wiki...)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) 2018, Adam <Adam@sigterm.info>
+ * Copyright (c) 2018, Cas <https://github.com/casvandongen>
+ * Copyright (c) 2020, Carl <https://github.com/CarlOlson>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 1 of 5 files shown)


Nightfirecat:
This will cause heavy performance impacts to most players. The agility plugin fairly aggressively limits the distance to which it'll draw clickboxes out to.
This may be better-suited as a hub plugin?

CarlOlson:
> This will cause heavy performance impacts to most players.

Is there some rule-of-thumb or telemetry on performance for future reference?

> The agility plugin fairly aggressively limits the distance to which it'll draw clickboxes out to.

I copied the same distance limiting to this plugin, but wasn't sure why it was used to begin with. :+1: 

> This may be better-suited as a hub plugin?

Doesn't matter to me either way, but is performance still a concern if this is just off by default?

discomethod:
Seems like a common use case would be to not draw ALL woodcutting clickboxes, but only clickboxes for the specific tier(s) of tree the player wants to target (i.e., at 45 WC, there's no need to show Magic tree clickboxes, and at 95 WC, there's little reason to highlight willow trees). A checklist of which clickboxes to enable would help reduce clutter, which would in turn improve performance and usability.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+	private static final int MAX_DISTANCE = 2350;
+	private static final Color TREE_HIGH_LEVEL_COLOR = Color.ORANGE;
+	private static final int OVERLAY_ALPHA = 50;
+
+	private final Client client;
+	private final WoodcuttingPlugin plugin;
+	private final WoodcuttingConfig config;
+
+	@Inject
+	private WoodcuttingClickboxOverlay(Client client, WoodcuttingPlugin plugin, WoodcuttingConfig config)
```
Nightfirecat, line 53: This should render before the axe overlay so it is not obscured.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+		LocalPoint playerLocation = client.getLocalPlayer().getLocalLocation();
+		Point mousePosition = client.getMouseCanvasPosition();
+		int woodcuttingLevel = client.getRealSkillLevel(Skill.WOODCUTTING);
+
+		plugin.getTreeObjects().forEach((object) ->
+		{
+			Tree tree = Tree.findTree(object.getId());
+
+			if (tree != null
+				&& tree != Tree.REGULAR_TREE
```
Nightfirecat, line 80: Why are regular trees excluded?

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+
+		LocalPoint playerLocation = client.getLocalPlayer().getLocalLocation();
+		Point mousePosition = client.getMouseCanvasPosition();
+		int woodcuttingLevel = client.getRealSkillLevel(Skill.WOODCUTTING);
+
+		plugin.getTreeObjects().forEach((object) ->
+		{
+			Tree tree = Tree.findTree(object.getId());
+
+			if (tree != null
```
Nightfirecat, line 79: Will this ever be null?

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingClickboxOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+		{
+			return null;
+		}
+
+		LocalPoint playerLocation = client.getLocalPlayer().getLocalLocation();
+		Point mousePosition = client.getMouseCanvasPosition();
+		int woodcuttingLevel = client.getRealSkillLevel(Skill.WOODCUTTING);
+
+		plugin.getTreeObjects().forEach((object) ->
+		{
```
Nightfirecat, line 76: The nested if statements in here should become guard clauses, e.g.

```java
plugin.getTreeObjects().forEach((object) ->
{
	Tree tree = Tree.findTree(object.getId());

	if (tree == null
		|| tree == Tree.REGULAR_TREE
		|| object.getPlane() != client.getPlane()
		|| object.getLocation().distanceTo(playerLocation) < MAX_DISTANCE)
	{
		return;
	}

	Shape objectClickbox = object.getClickbox();
	if (objectClickbox == null)
	{
		return;
	}

	Color configColor = tree.getLevel() <= woodcuttingLevel ? config.getOverlayColor() : TREE_HIGH_LEVEL_COLOR;
	// continue code here...
});
```

Nightfirecat (changes requested):

PR #12610, opened 2020-10-06
ground markers: add tile labels

Adam-:


runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerOverlay.java
```diff
@@ -101,11 +104,18 @@ private void drawTile(Graphics2D graphics, WorldPoint point, Color color)
...
 		Polygon poly = Perspective.getCanvasTilePoly(client, lp);
-		if (poly == null)
+		if (poly != null)
 		{
-			return;
+			OverlayUtil.renderPolygon(graphics, poly, color);
 		}
 
-		OverlayUtil.renderPolygon(graphics, poly, color);
+		if (!Strings.isNullOrEmpty(label))
```
Nightfirecat, line 112: I don't think we should try to render the label if the tile poly is null.

Adam-: I think this won't hurt, maybe it is possible to get it at weird camera angles or something.

PR #12612, opened 2020-10-06
Time Tracking: Add right click menu to reset farming patches

Raitab:
Add a right click menu to reset farming patches in time tracking. 

Useful for when you collect a crop on mobile and you don't want to go back to check it so it doesn't show "Ready" when it's not.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTabPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTabPanel.java
@@ -33,6 +33,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import javax.swing.JLabel;
+import javax.swing.JMenuItem;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTabPanel.java
```diff
@@ -103,6 +105,31 @@
...
+			{
+				farmingTracker.resetData(p.getTimeable().getRegion(), p.getTimeable().getVarbit());
+
+				itemManager.getImage(Produce.WEEDS.getItemID()).addTo(p.getIcon());
+				p.getIcon().setToolTipText("Unknown state");
+				p.getProgress().setMaximumValue(0);
+				p.getProgress().setValue(0);
+				p.getProgress().setVisible(false);
+				p.getEstimate().setText("Unknown");
+				p.getProgress().setBackground(null);
```
Nightfirecat, line 129: Can't this all be replaced by a call to `update()`?

PR #12615, opened 2020-10-07
Update CoordinateClue.java

HughRaynor:
Corrected a Wilderness Clue Step with the CoordinateClueInfo of 'Chaos Temple' to 'Chaos Altar', to better distinguish the location from the Low-level Wilderness Chaos Temple location.


Hydrox6 (changes requested):
There are two clues by the Chaos Altar, which have been mixed up and merged together here.

makingcomment:
> to better distinguish the location from the Low-level Wilderness Chaos Temple location.

That's what the "(level 38)" that you removed was for.

Now it needs to be distinguished from the runecrafting chaos altar.

Nightfirecat:
This clue is already unambiguous and clear as-is, definitely moreso than the proposed change. I don't see a reason to change it.

PR #12616, opened 2020-10-07
menuentryswapper: enable shift withdraw/deposit for seed vault

SirGirion:
Makes your bank withdraw/deposit options in MES work in the seed vault. Doesn't support withdraw-all-but-1 because that isn't an option for the vault so does nothing in that case. Tested the options swapped for all relevant values.


SirGirion:
@Nightfirecat Please review

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -539,10 +541,11 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 		// Swap to shift-click withdraw behavior
 		// Deposit- op 1 is the current withdraw amount 1/5/10/x
 		if (shiftModifier() && config.bankWithdrawShiftClick() != ShiftWithdrawMode.OFF
-			&& menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 1
-			&& menuEntryAdded.getOption().startsWith("Withdraw"))
+			&& menuEntryAdded.getType() == MenuAction.CC_OP.getId() && (menuEntryAdded.getIdentifier() == 2 || menuEntryAdded.getIdentifier() == 1)
+			&& menuEntryAdded.getOption().startsWith("Withdraw-"))
```
Adam-, line 545: This will break the storage unit withdraw swap I believe. Maybe there should be a test for that. I think it does not include the `-`?

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -553,7 +556,7 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 			else
 			{
 				actionId = shiftWithdrawMode.getMenuAction().getId();
-				opId = shiftWithdrawMode.getIdentifier();
+				opId = WidgetInfo.TO_GROUP(menuEntryAdded.getActionParam1()) == WidgetID.SEED_VAULT_GROUP_ID ? shiftWithdrawMode.getIdentifierSeedVault() : shiftWithdrawMode.getIdentifier();
```
Adam-, line 559: This is already defined to `widgetGroupId`

Nightfirecat (approved):
LGTM, works fine ingame

PR #12620, opened 2020-10-08
add button to hide all screen markers

### LINKED ISSUE #12210: Screen Markers |  Add a hide all option to the screen markers plugin
Title.  It can be a bit tedious to scroll throughout my screen markers list to find which ones to turn off.  If there was a hide all button at the top it would be greatly appreciated.

MaxGlore:
Closes #12210 

Change to screen markers plugin that allows users to set all screen markers to hidden

[https://imgur.com/mtFdjhy](url)
Gif ^


Adam-:
I'm not sure I see this as being very useful, since it is almost equivalent to just turning off the plugin? Maybe a "delete-all" could make some sense, but with this you would still have to go through and delete them anyway. And also you cannot mass-toggle them back on either?  There was some talk of screen marker groups before but I don't think I've seen any good ui/ux proposals for them, but that could solve this instead.

Issue #12280, opened 2020-08-03
Allow XP Globes to be stacked vertically


Smellypniz:
**Is your feature request related to a problem? Please describe.**
In fixed mode, the XP globes, no matter the size can be a little intrusive as they're stacked horizontally.

**Describe the solution you'd like**
If they were able to be stacked vertically they could be placed in nicer areas such as the columns in the inventory.

**Describe alternatives you've considered**
Potentially re-arranging independent globes, much simpler to have them wrap vertically though.

**Additional context**

PR #12625, opened 2020-10-10
XPGlobes: allow stacking vertically.

Bram91:
Allows for the xp globes to stack vertically.

image

Closes: #12280


deathbeam:
I think adding Flip option similar to infoboxes for this would be better, not sure

Bram91:
Changed the PR to use a `Flip` option similar to infoboxes.

deathbeam (approved):

Issue #8515, opened 2019-04-13
Customisable Anchors


ThimbleFire:
I'd like to propose customisable anchors.

![default](https://i.imgur.com/2zIOn2W.png)

**Why I'm making this suggestion**

I thought it'd be nice if the anchor in the middle-top of the screen were a bit lower, and then I thought it'd be even better if there were a way to customise anchors.

I looked to see if anyone had suggested this but I didn't find any posts about it, so I thought _what the hell_, what's the worst that could happen.

Anyway, I made a plugin interface design below.

![default](https://i.imgur.com/IWtjotj.png)

This step could probably skipped, but it'd allow the user to quickly select an area for their anchor.
![default](https://i.imgur.com/S5jzlfV.png)

![default](https://i.imgur.com/wNf2R0b.png)

 ![default](https://i.imgur.com/TAgwyT4.png) ![default](https://i.imgur.com/A1T1FZB.png) You can control the position of the anchor by changing the vertical and horizontal values.

![default](https://i.imgur.com/9Wz6Uw7.png) You can also change the direction info boxes stack by selelecting an side, corner or center

![default](https://i.imgur.com/zNzj4K4.png) Lastly, you can specify (toggle) between horizontally and vertically displaying info boxes by clicking on the horizontal and vertical icons. To elaborate what I mean here are some examples

![default](https://i.imgur.com/xQIDGB9.png) Vertical

![default](https://i.imgur.com/6lqhUnY.png) Horizontal

**Closing thoughts**
I make this suggestion not entirely invested in the idea. I can't think of many practical uses for it. It'd be nice to have more diverse control over how info boxes are displayed and their locations. I'm mostly posting it because I'm sure there are other players who would feel more strongly toward wanting this feature.

Reasel:
I think this would be a great addition to the client.

Alexsuperfly:
duplicate of #4461

Nightfirecat:
I'm closing 4461 in favor of this, as this issue is much more detailed with a good mockup of this feature's possible implementation.

qnps:
When playing with status bars in resizeable, they hide widgets locked to the anchor next to the tabs.
![uKbP1io](https://user-images.githubusercontent.com/44282659/72757639-1ca61100-3b96-11ea-9ac7-eee895934196.png)
![dycad2W](https://user-images.githubusercontent.com/44282659/72757641-1e6fd480-3b96-11ea-83d5-20e073b2e93a.png)

[...]

PR #12627, opened 2020-10-11
Reposition top center snap corner in Resizable Mode

aloudasian:
## Issue
The top center snap corner is misaligned in Resizable Mode. It is positioned to the left of the character instead of above the character in Fixed Mode.

## Change
Reposition the top center snap corner so it is aligned with the character when in Resizable Mode

## Testing
Locally tested

## Screenshots
Before: 
image
After:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -810,7 +810,7 @@ public class OverlayRenderer extends MouseAdapter implements KeyListener
 			viewportOffset + BORDER_TOP);
 
 		final Point topCenterPoint = new Point(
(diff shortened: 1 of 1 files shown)


raiyni:
This snap point is based on the center of the widget viewport.

aloudasian:
Yes but in resizable mode I think you could argue that the viewport expands to fill the entire client, even if the widget remains bound by the map and chat widget. It was a personal change but I've received feedback from a few people that they would like this in their client.

raiyni:
This is probably better fixed by #8515

Unmoon:
IMO this issue describes the kind of default behavior these snap corners have, even though customizable would fix it.

deathbeam:
This should definitely not reposition anything. If you want add new snap corner and we can decided if we want it or implement #8515

PR #12631, opened 2020-10-12
ImageUtil: add error log with path

### LINKED ISSUE #10708: ImageUtil.getResourceStreamFromClass does not log attempted path if load fails

kkozera:
Close #10708

I came with this idea to show exact place where file should be.

Is there any better way to handle it?

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
@@ -351,15 +351,18 @@ public class ImageUtil
 		}
 		catch (IllegalArgumentException e)
 		{
+			String filePath;
+
 			if (path.startsWith("/"))
 			{
-				log.warn("Failed to load image from class: {}, path: {}", c.getName(), path);
+				filePath = path;
 			}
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -351,6 +351,16 @@ public static BufferedImage getResourceStreamFromClass(final Class c, final Stri
 		}
 		catch (IllegalArgumentException e)
 		{
+			if (path.startsWith("/"))
```
deathbeam, line 354: Just store the path and then call log.warn only once

deathbeam (changes requested):

deathbeam (approved):

PR #12632, opened 2020-10-12
ChatCommands: Command Aliases

DevRuto:
Added a way to have aliases for certain commands

Such as using `!level` for the `!lvl` command, or `!pc` for the `!price` command.

The code checks for registered aliases and replaces the command in the message with the normal command


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -189,6 +189,11 @@ public void startUp()
 		chatCommandManager.registerCommandAsync(PB_COMMAND, this::personalBestLookup, this::personalBestSubmit);
 		chatCommandManager.registerCommandAsync(GC_COMMAND_STRING, this::gambleCountLookup, this::gambleCountSubmit);
 		chatCommandManager.registerCommandAsync(DUEL_ARENA_COMMAND, this::duelArenaLookup, this::duelArenaSubmit);
+
+		chatCommandManager.registerCommandAlias(TOTAL_LEVEL_COMMAND_STRING, "!overall");
+		chatCommandManager.registerCommandAlias(PRICE_COMMAND_STRING, "!pc");
+		chatCommandManager.registerCommandAlias(LEVEL_COMMAND_STRING, "!level");
+		chatCommandManager.registerCommandAlias(CMB_COMMAND_STRING, "!cb", "!cblvl");
```
Nightfirecat, line 196: In my opinion, this would be better and simpler as either (a) invoking `registerCommand(Async)?` with the alternative command strings, or (b) changing the command registration API to accept some array or collection of strings.

DevRuto: I agree with that, I thought about option b before but wasn't sure how to go about where the alias would be. Maybe changing the command string constant to an array. When I can, I'll make the changes and see how it goes

Adam-: I'd vote on just registering the command twice probably.

Unmoon:
https://github.com/runelite/runelite/issues/5294 was closed because other clients would not support this - if this is no longer a concern, maybe we could add a few other aliases like `combat`, `overall` and `ge` as well.

DevRuto:
Thoughts on this implementation?
I had to make some command handlers less dependent on the command string, as I don't think it's necessary.
If the check against the command string is necessary, I can add it back and have it check the aliases as well.

Adam-:
> 
> 
> Thoughts on this implementation?
> I had to make some command handlers less dependent on the command string, as I don't think it's necessary.
> If the check against the command string is necessary, I can add it back and have it check the aliases as well.

I'd vote on just registering the command twice probably. If you want to simplify the parameter parsing we could maybe change the callback signature to `(ChatMessage chatMessage, String command, String message)`

DevRuto:
> 
> 
> > Thoughts on this implementation?
> > I had to make some command handlers less dependent on the command string, as I don't think it's necessary.
> > If the check against the command string is necessary, I can add it back and have it check the aliases as well.
> 
> I'd vote on just registering the command twice probably. If you want to simplify the parameter parsing we could maybe change the callback signature to `(ChatMessage chatMessage, String command, String message)`

I've updated the code to register the command alias with their original command handler, or do you mean something else?
I might've been misleading saying `this implementation` as I changed my approach from the original implementation

Adam-:
I mean this:

```java
chatCommandManager.registerCommandAsync("!total", this::playerSkillLookup);
chatCommandManager.registerCommandAsync("!overall", this::playerSkillLookup);
```

DevRuto:
I see. I pushed a change that registers the command in the same place as others

Adam-:
I don't think any of these changes to `ChatCommandManager` are necessary; you can just add the additional commands next to the existing ones. And then, change the command handler to only compute `indexOf(' ')` once and it should be good.

Nightfirecat (changes requested):

DevRuto:
@Nightfirecat I think the rebase went horribly wrong on this so I'll probably close this and reopen the pull request on a new branch with the changes you requested

MESLewis:
Force push to this existing github branch and this pr will update without you needing to create a new one.

DevRuto:
> 
> 
> Force push to this existing github branch and this pr will update without you needing to create a new one.

Ok cool, I'll look into that once I get the branch sorted out, thanks.

DevRuto:
This looks good now, thanks for the suggestion

Nightfirecat (approved):
LGTM

PR #12636, opened 2020-10-13
itemmanager: calculate price of scrolls for scroll book price

SirGirion:
Makes the scroll book show the value of the scrolls a player has. Technically if someone had 2 scroll books the value would be double the reality since the scrolls are shared between books, but IMO the number of people that keep 2+ charged scroll books is probably pretty low. If there is enough concern about this I can look into changing the behavior. I also mapped every scroll, even the untradeable ones, just to have them.


SirGirion:
@Nightfirecat Please review

Nightfirecat (approved):

SirGirion:
Rebased to new changes, removed the change for int -> long on prices since I'm assuming that would probably break phub stuff so not worth at this point

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
```diff
@@ -279,6 +298,22 @@ public int getItemPriceWithSource(int itemID, boolean useWikiPrice)
 		{
 			return 1000;
 		}
+		if (itemID == MASTER_SCROLL_BOOK)
+		{
+			// Can only hold 13k tradeable teleport scrolls, unlikely this overflows an int
+			// as each scroll would need to be over 165k ea
```
Nightfirecat, line 304: I guess this must have changed since the time of writing. There are now 15 types of scrolls, making the amount ~143k each.

Nightfirecat (commented):
As it's currently written, this only works for a stack of exactly 1 scroll book. If a user ever had 2 in their bank, this would incorrectly double the value of the contained scrolls. This was brought up in #8595 where a similar implementation was written.

runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
SirGirion: Watson and Keymaster teleports can't be traded. Arguably I suppose they shouldn't even be in the map then since their price will always be 0?

Nightfirecat: As was mentioned on Discord, only Keymaster teleports are untradeable, so I suppose it would be fine to leave them out. (maybe include a comment explaining that)

Nightfirecat:
I have a feeling that we can assume that nobody is stacking up so many master scroll books that their stack would exceed int's max value, so a change from `int` to `long` return value shouldn't be needed. That said, the existing API should not change--ensure whatever overloads are added do not impact existing plugins in any way.

SirGirion:
I added new APIs with a Long suffix denoting their return type and updated the code base to use them where appropriate. I'm not sure if it is the best way to do it, did my best within the constraint of not touching the public API.

PR #12638, opened 2020-10-13
Refactor QuantityFormatter so method names and javadocs are more clear.

winterdaze:
Summary of changes:
- changed quantityToRSDecimalStack to quantityToSIStack, which takes in a parameter for level of precision and doesn't match in-game stack behavior. replaced quantityToStackSize with quantityToPreciseSIStack, which overloads quantityToSIStack and specifies thousandths precision (so in most places prices are now more precise to an additional decimal point; behavior for XP panels [already 3 digits precision] and infoboxes [1 digit precision, to avoid overflowing the box] was not changed)
- added quantityToRSStack, which follows in-game stack behavior (e.g. 2147M instead of 2.147B and no decimal precision).
- removed support for negative stacks, as I don't think they're used anywhere (not calculating any price differences AFAIK)
- improved some of the docs
changes

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoCounter.java
@@ -46,7 +46,7 @@ class AmmoCounter extends Counter
 	@Override
 	public String getText()
 	{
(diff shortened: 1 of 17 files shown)


makingcomment:
Having so many decimal places (e.g. 723.548M) somewhat defeats the purpose of the M abbreviation.

PR #12641, opened 2020-10-14
client util: prevent regex injection in WildcardMatcher

Trevor159:
Currently if the user is dumb enough to add some special regex characters to the ground items plugin/other places we use the wildcard matcher it can error. The test added will fail on master.


Adam-:
fixed in 811842ef2f46ade48bff5b4148256dbd8f791c79

PR #12646, opened 2020-10-15
Chat Notifications: Fixed highlighting messages that end with a period.

### LINKED ISSUE #12524: Chat notifications/Chat colors red dot
**Describe the bug**
When using the "chat notifications" plugin to change the colour of text, such as 'your divine has worn off',
you can't colour the dot at the end without it turning back to default.

The top message uses: Your divine potion effect is about to expire
The bottom message uses: The effects of the divine potion have worn off.

Putting the dot at the end of the sentence makes it not work, and not putting it there obviously doesn't make it red.

**To Reproduce**
► Go to Runelite settings
► Go to 'Chat Color' plugin
► Pick/change colour of 'Server message highlight' (16th from the top)
► If you use resizable, pick/change colour of 'Server message highlight (transparent)' (12th from bottom) 
► Then go to 'Chat notifications' plugin and add text to 'Highlight words' :
The effects of the divine potion have worn off, Your divine potion effect is about to expire.

**Expected behavior**
What I expect to happen is for the red dot to turn purple (in this case)

 - OS: Windows 10
 - RuneLite version: 1.6.26.1
 - Launcher version: 1.6.0

jbman223:
Closes #12524 

Fixes highlight message matcher to correctly match messages that end in a `.`

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsPlugin.java
@@ -120,8 +120,8 @@ public class ChatNotificationsPlugin extends Plugin
 				.map(this::quoteAndIgnoreColor) // regex escape and ignore nested colors in the target message
 				.collect(Collectors.joining("|"));
 			// To match <word> \b doesn't work due to <> not being in \w,
-			// so match \b or \s
-			highlightMatcher = Pattern.compile("(?:\\b|(?<=\\s))(" + joined + ")(?:\\b|(?=\\s))", Pattern.CASE_INSENSITIVE);
+			// so match \b or \s, as well as \A and \z for beginning and end of input respectively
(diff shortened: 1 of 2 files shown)


Adam-:
The opposite of this is also an issue, though it would require something which is maybe a non-sentence. Such as trying to match `.Your divine potion effect is about to expire`. It may be worth fixing it in addition to this to keep it consistent. You can also use `\\A` and `\\z` for beginning/end of input (as opposed to line as is `^` and `$`). And so you could use pattern `"(?:\\b|(?<=\\s)|\\A)(" + joined + ")(?:\\b|(?=\\s)|\\z)"` and then also change the test message to `.Your divine potion effect is about to expire.`

jbman223:
Updated PR to your suggestions. Great input, thanks a lot.

Adam-:
I'm pretty sure its `\z` and not `\Z` (there is some difference but I'm unsure what it is). Also I've noticed the inner group doesn't require capturing (since the code which does the replacements just replaces the entire match) so it can be `"(?:\\b|(?<=\\s)|\\A)(?:" + joined + ")(?:\\b|(?=\\s)|\\z)"`

PR #12648, opened 2020-10-15
Updated cache tools

JasonM97:
Created a DefaultExporter to replace a few classes in: net.runelite.cache.definitions.exporters
Added a command to dump obj and mtl files


Adam-:
Are all of the things you've replaced with `DefaultExporter`  unused now? I think so. If so they should be removed. Also this should have one commit to add the `DefaultExporter` and a second to add the new `models` arg. Additionally please review the codestyle on the wiki.

PR #12651, opened 2020-10-16
Next Combat Level: handle level 125 empty tooltip edge case

### LINKED ISSUE #12467: Next Combat Level tooltip empty
With 97 defence an 94 prayer the tooltip does not display the levels require for 126cmb.
image

Per Nightfirecat's investigation it seems that the issue is inside the CombatLevelOverlay and since neither 2 defence levels nor 5 prayer levels gets me 126 combat the overlay doesn't add anything.

```java
if ((hitpointsLevel + defenceLevel + hpDefNeed) <= Experience.MAX_REAL_LEVEL * 2)
{
    sb.append(hpDefNeed).append(" Defence/Hitpoints</br>");
}
// ...
if ((prayerLevel + prayerNeed) <= Experience.MAX_REAL_LEVEL)
{
    sb.append(prayerNeed).append(" Prayer");
}
```

1jz:
fixes #12467

The next combat tooltip only shows how many levels of a certain skill you need in order to level up.

With 97 defence and 94 prayer the tooltip does not display the levels required for 126 because the values returned will make you surpass the max level of 99 like so:
```
Attack/Strength needed: 3
Defence needed: 4
Ranged needed: 35
Mage needed: 35
Prayer needed: 8
```

This only happens when you reach level 125, in order to get level 126 you have to max all combat skills.

This fix handles this edge case for players at level 125 by basically telling them to max, prayer is the only limiting factor requiring a level of 98.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelOverlay.java
@@ -104,47 +104,38 @@ class CombatLevelOverlay extends Overlay
 		int prayerNeed = Experience.getNextCombatLevelPrayer(attackLevel, strengthLevel, defenceLevel, hitpointsLevel,
 			magicLevel, rangeLevel, prayerLevel);
 
-		if (client.getLocalPlayer().getCombatLevel() >= 125)
-		{
-			if (defenceLevel != Experience.MAX_REAL_LEVEL)
-			{
-				hpDefNeed = Experience.MAX_REAL_LEVEL - defenceLevel;
-			}
-			if (strengthLevel != Experience.MAX_REAL_LEVEL || attackLevel != Experience.MAX_REAL_LEVEL)
-			{
-				meleeNeed = (Experience.MAX_REAL_LEVEL - attackLevel) + (Experience.MAX_REAL_LEVEL - strengthLevel);
-			}
-			if (prayerLevel != Experience.MAX_REAL_LEVEL)
-			{
-				prayerNeed = Experience.getNextCombatLevelPrayer(Experience.MAX_REAL_LEVEL,
-						Experience.MAX_REAL_LEVEL, Experience.MAX_REAL_LEVEL, Experience.MAX_REAL_LEVEL,
-						Experience.MAX_REAL_LEVEL, Experience.MAX_REAL_LEVEL, prayerLevel);
-			}
-		}
-
 		// create tooltip string
 		StringBuilder sb = new StringBuilder();
-		sb.append(ColorUtil.wrapWithColorTag("Next combat level:</br>", COMBAT_LEVEL_COLOUR));
+		boolean isMeleeCombatDominating = ((attackLevel+strengthLevel) > (Experience.MAX_REAL_LEVEL / 2 + Experience.MAX_REAL_LEVEL));
 
-		if ((attackLevel + strengthLevel + meleeNeed) <= Experience.MAX_REAL_LEVEL * 2)
+		if (client.getLocalPlayer().getCombatLevel() >= Experience.MAX_COMBAT_LEVEL)
 		{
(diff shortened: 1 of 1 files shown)


Adam-:
I think you are changing what the overlay is showing with this. Normally the overlay shows how many of each group is required, eg

```
1 Attack/Strength
2 Defence/Hitpoints
```

would mean `(1 attack or strength levels) OR (2 defence or hitpoints levels)`. You are changing this from `OR` to `AND` which seems confusing?

I don't see anything too wrong with just uncapping it so it can show past 99. Eg for your example

```
Attack/Strength needed: 3
Defence needed: 4
Ranged needed: 35
Mage needed: 35
Prayer needed: 8
```
Is *actually correct * despite not being able to get all 4 defence or 8 prayer, so we could just show that.

1jz:
Yeah, I agree it was confusing as I suddenly switch it from OR to AND for players that are level 125.

I changed it so that it will show uncapped leveling requirements so assuming 97 defense and 94 prayer it'll show
```
Defence needed: 4
Prayer needed: 8
```
I also made it so that range/mage levels wont show at all if the melee combat level is dominating the range and mage levels as per https://oldschool.runescape.wiki/w/Combat_level#Mathematics

This fix looks much cleaner too.

runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelOverlay.java
```diff
@@ -106,27 +106,36 @@ private String getLevelsUntilTooltip()
...
 		// create tooltip string
 		StringBuilder sb = new StringBuilder();
-		sb.append(ColorUtil.wrapWithColorTag("Next combat level:</br>", COMBAT_LEVEL_COLOUR));
+		boolean isMeleeCombatDominating = ((attackLevel + strengthLevel) > (Experience.MAX_REAL_LEVEL / 2 + Experience.MAX_REAL_LEVEL));
 
-		if ((attackLevel + strengthLevel + meleeNeed) <= Experience.MAX_REAL_LEVEL * 2)
+		if (client.getLocalPlayer().getCombatLevel() >= Experience.MAX_COMBAT_LEVEL)
 		{
-			sb.append(meleeNeed).append(" Attack/Strength</br>");
+			sb.append(ColorUtil.wrapWithColorTag("Max combat level achieved", COMBAT_LEVEL_COLOUR));
```
Adam-, line 113: This is already validated in the `render` method, which is why on release it does not show any tooltip, so this is never hit. I'd recommend just removing this.

runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelOverlay.java
```diff
@@ -106,27 +106,36 @@ private String getLevelsUntilTooltip()
...
+
+			if ((attackLevel + strengthLevel) < Experience.MAX_REAL_LEVEL * 2)
+			{
+				sb.append(meleeNeed).append(" Attack/Strength</br>");
+			}
+			if ((hitpointsLevel + defenceLevel) < Experience.MAX_REAL_LEVEL * 2)
+			{
+				sb.append(hpDefNeed).append(" Defence/Hitpoints</br>");
+			}
+			if (rangeLevel < Experience.MAX_REAL_LEVEL && !isMeleeCombatDominating)
```
Adam-, line 127: I don't understand why the check for `!isMeleeCombatDominating` is added? Even with the combat level being computed by the "melee combat level", that doesn't mean you cannot gain combat levels from ranged levels. It just means that you have to gain ranged levels until `range * 3/2 > attack+str`

Adam-:
Also the `getLevelsUntilTooltip` could be made testable pretty easy which may be useful to have at least the case in #12467

Adam-:
I've added this in 6f60750c313d586b5130717c7b9b8dc2383aa319

PR #12666, opened 2020-10-18
friendschat: recolor ignored players in ToB interfaces

### LINKED ISSUE #10515: Display users differently in the ToB interface if they are on the player's ignore list
Inside of any community, there are users that a player doesn't want to interact with. For most of the game, there is the ignore list, but trying to use that at ToB is quite a handful, especially if the player has a large list.

To fix this, the users on the player's ignore list should show up differently inside the Recruitment interface and in the party display in Vir Sinhaza. The most easy, and probably most effective way to do this would be to colour the name red.

This same idea could also in future extend to the Clan Chat Interface, for use in the Chambers of Xeric.

winterdaze:
Closes #10515

img


abextm (approved):

Nightfirecat (changes requested):
This strikes me as a change which should be in a plugin specific to TOB, not shoehorned into the friendschat plugin.

winterdaze:
I put it here so it's with all the other ignored players settings and uses that config field

Adam-:
I think it would make more sense in the friends list plugin, since tob really has nothing to do with fc/cc. Otherwise this seems fine mostly, you can replace the string building with a `StringBuilder` instead of a whole bunch of `+=` though.

runelite-client/src/main/java/net/runelite/client/plugins/friendlist/FriendListPlugin.java
```diff
@@ -109,6 +124,15 @@ else if (event.getScriptId() == ScriptID.IGNORE_UPDATE)
...
 	}
 
+	@Subscribe
+	public void onGameTick(GameTick gameTick)
+	{
+		if (config.showTobIgnores())
+		{
+			colorIgnoredPlayersTob(config.showIgnoresColor());
+		}
+	}
```
Nightfirecat, line 134: This can be replaced with a `PostScriptFired` subscriber for the scripts which build these interfaces instead of running every tick.

PR #12668, opened 2020-10-19
OverlayManager: Unset stored tooltip overlay location

### LINKED ISSUE #11643: Mouse Tooltips Misaligned While Logged In

**Describe the bug**
While logged into my runelite account the mouse tool tips becomes misaligned .

**To Reproduce**
1. Open Runelite while logged into account
2. Mouse over an item
3. Tool tip is misaligned 
4. Logging out of Runelite profile fixes the error

**Expected behavior**
The mouse tool tip to be aligned with the item the mouse is over

**Screenshots**
https://imgur.com/a/XusM1iz

 - OS: Windows 64 Bit
 - RuneLite version: [e.g 1.6.15]
 - Launcher version: [e.g 2.1.0]

**Additional context**
Happened a day ago while playing I was changing a few setting in different plugins, mainly XP Globes and Ground Items when it happened. I have tried uninstalling and reinstalling as well as deleting everything related to runelite off of my computer to no avail.

Nightfirecat:
In the past, prior to commit 6df3016bb, the tooltip overlay was errantly
able to be repositioned by the user. While that commit fixes that issue,
users who had already given their tooltip overlay some saved location
will have that issue persist. This commit adds a method, to be removed
at some later time after users will have run it for a while, which
unsets that stored location from their config.

Closes #11643

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayManager.java
@@ -49,6 +49,9 @@ import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.ConfigChanged;
 import net.runelite.client.events.OverlayMenuClicked;
 import net.runelite.client.events.PluginChanged;
+import net.runelite.client.events.SessionClose;
+import net.runelite.client.events.SessionOpen;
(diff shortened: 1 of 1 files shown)


deathbeam:
I dont think this is correct as it will work only for local config

Nightfirecat:
I think this is fixed in other code now, so this shouldn't be needed anymore.

PR #12674, opened 2020-10-20
timers plugin: Improve imbued heart detection

### LINKED ISSUE #3516: Change imbued heart infobox condition
I assume as of now it's set to the animation ID for using the heart. Sometimes in combat the animation gets ignored/overwritten depending on what you're doing, causing the infobox not to show up.

Is it possible to change the infobox condition for appearing to be when your magic level gets set to what the heart boosts it to, which is 10% of level (rounded down) +1? No other magic boost in the game does the same thing (maybe there could be some overlap if you do something like use a magic potion when level 30-39). Not sure if OSRS allows some type of condition where you can set code to execute if item in inven = true, or something.

Nightfirecat:
As is the case with other graphics-based timers, the imbued heart timer
will not fire if other graphics animations are triggered (such as those
created when fighting the Dagannoth Kings). To add this timer more
reliably, this commit will add the timer when a Magic stat boost occurs
of the appropriate amount after recently clicking on an imbued heart.
Because the magic level boost, combined with an imbued heart click
check, is a reliable way to detect usage of an imbued heart, the
graphics check is removed.

Closes #3516

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/GraphicID.java
+++ b/runelite-api/src/main/java/net/runelite/api/GraphicID.java
@@ -44,7 +44,6 @@ public class GraphicID
 	public static final int BOOK_HOME_TELEPORT_3 = 803;
 	public static final int BOOK_HOME_TELEPORT_4 = 804;
 	public static final int STAFF_OF_THE_DEAD = 1228;
-	public static final int IMBUED_HEART = 1316;
 	public static final int FLYING_FISH = 1387;
 	public static final int NPC_CONTACT = 728;
 	public static final int POT_SHARE = 733;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
@@ -54,7 +54,7 @@ enum GameTimer
 	ICEBURST(SpriteID.SPELL_ICE_BURST, GameTimerImageType.SPRITE, "Ice burst", GraphicID.ICE_BURST, 16, GAME_TICKS, true),
 	ICEBLITZ(SpriteID.SPELL_ICE_BLITZ, GameTimerImageType.SPRITE, "Ice blitz", GraphicID.ICE_BLITZ, 24, GAME_TICKS, true),
 	ICEBARRAGE(SpriteID.SPELL_ICE_BARRAGE, GameTimerImageType.SPRITE, "Ice barrage", GraphicID.ICE_BARRAGE, 32, GAME_TICKS, true),
-	IMBUEDHEART(ItemID.IMBUED_HEART, GameTimerImageType.ITEM, "Imbued heart", GraphicID.IMBUED_HEART, 420, ChronoUnit.SECONDS, true),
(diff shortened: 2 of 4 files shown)


makingcomment:
Could this give a false positive between levels 40-49 if a spicy stew gives a +5 boost?

Nightfirecat:
As-is, it can only do that if you have an imbued heart which you click at the same time. Even when I wrote this the first time around (without the click logic--which had false positives at CoX) that wasn't a case I bothered handling since ~0 users will ever boost their combat stats with spicy stews.

Nightfirecat:
After some Discord discussion, I've removed the lower bound for level checking since all ambiguous cases should be properly handled by the imbued heart click check.

discomethod:
I've noticed that even when the animation is overridden, the sound effect of the imbued heart plays reliably. Could this be used to improve the detection of imbued heart usage?

Adam-:
This was added once before in 45c5df3379767146f91e1e4a1f48d9692599a1d4 and then reverted without a reason in 967fadf59d8808dd84f4a5f4c441031fe8bba4bb by you - do you know why this was done?

Nightfirecat:
Yes. The original PR which added this feature made the false assumption that you could only ever achieve a boost equivalent to that of the imbued heart with the heart itself. Chambers of Xeric overload potions combined with some combination of Xeric's aid and revitalization and/or overload re-boosts can cause you to end up at the same boost amount which caused a lot of erroneous timers to appear for players doing raids. This PR addresses that by ensuring an imbued heart is actually clicked recently before the boost.

Adam-:
Does this create the timer twice then? Is there any reason to keep the existing graphic check?

Nightfirecat:
This change would cause the timer to be added by both the stat change and graphic change events. It's probably safe to remove the graphic change event trigger, but I don't own an imbued heart so I can't say with confidence that would work.

Adam-:
It does seem to add two

```
2020-11-21 19:14:08 [Client] DEBUG n.r.c.plugins.timers.TimersPlugin - Creating timer IMBUEDHEART
2020-11-21 19:14:08 [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Adding InfoBox Timer(startTime=2020-11-22T00:14:08.278379200Z, endTime=2020-11-22T00:21:08.278379200Z, duration=PT7M)
2020-11-21 19:14:08 [Client] DEBUG n.r.c.plugins.timers.TimersPlugin - Creating timer IMBUEDHEART
2020-11-21 19:14:08 [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Removed InfoBoxes for filter net.runelite.client.plugins.timers.TimersPlugin$$Lambda$1519/0x000000080086c040@2cb522e from net.runelite.client.ui.overlay.infobox.InfoBoxOverlay@29f03df0
2020-11-21 19:14:08 [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Adding InfoBox Timer(startTime=2020-11-22T00:14:08.281383700Z, endTime=2020-11-22T00:21:08.281383700Z, duration=PT7M)
```

Nightfirecat:
I've revisited this and updated it to remove the graphics trigger for the timer, and to allow partial boosts and boosts from debuffed stats to trigger the timer.

PR #12689, opened 2020-10-25
Add "Placeholder" withdraw option to the Menu Entry Swapper's "Bank W…

Chicagolight:
Add "Placeholder" withdraw option to the Menu Entry Swapper's "Bank Withdraw Shift-Click" menu.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftWithdrawMode.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftWithdrawMode.java
@@ -37,9 +37,10 @@ public enum ShiftWithdrawMode
 	WITHDRAW_10("Withdraw-10", MenuAction.CC_OP, 4, 3),
 	WITHDRAW_X("Withdraw-X", MenuAction.CC_OP, 5, 5),
 	WITHDRAW_ALL("Withdraw-All", MenuAction.CC_OP_LOW_PRIORITY, 7, 4),
-	// chambers of xeric storage units do not have an "all-but-1" option, so this option will choose "Withdraw-all"
-	// instead when using the storage unit.
+	// chambers of xeric storage units do not have an "all-but-1" option or a "placeholder" option, so these options will choose "Withdraw-all"
+	// choose "Withdraw-all" instead when using the storage unit.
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -540,7 +540,7 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
 		// Deposit- op 1 is the current withdraw amount 1/5/10/x
 		if (shiftModifier() && config.bankWithdrawShiftClick() != ShiftWithdrawMode.OFF
 			&& menuEntryAdded.getType() == MenuAction.CC_OP.getId() && menuEntryAdded.getIdentifier() == 1
-			&& menuEntryAdded.getOption().startsWith("Withdraw"))
+			&& (menuEntryAdded.getOption().startsWith("Withdraw") || menuEntryAdded.getOption().equals("Placeholder")))
```
Adam-, line 543: This is here to detect top-most op, which is added last, in order to do the swap once. `Placeholder` is never that (it is always the current withdraw mode). So I think you should be able to remove this?

Chicagolight: The file has been reverted.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/ShiftWithdrawMode.java
```diff
@@ -40,6 +40,7 @@
 	// chambers of xeric storage units do not have an "all-but-1" option, so this option will choose "Withdraw-all"
 	// instead when using the storage unit.
```
Nightfirecat, line 41:
```suggestion
	// chambers of xeric storage units do not have an "all-but-1" or a "placeholder" option, so these options will
	// choose "Withdraw-all" instead when using the storage unit.
```

Chicagolight: Updated comments

makingcomment:
Why does the comment say `choose "Withdraw-all"` twice?

PR #12693, opened 2020-10-26
bank: show value of costume room containers

SirGirion:


SirGirion:
@Nightfirecat Please review

Adam-:
I don't understand how there can be a `COSTUME_ROOM` inv, but also you have to compute the stored items from the widgets? Especially so when you are identifying changes to it via the inv change event? Why can't this all just use itemcontainers like everything else we have?

Issue #11264, opened 2020-04-14
Menu Entry Swapper for Captain Barnaby


OliverBroadbent:
**Is your feature request related to a problem? Please describe.**
Simple request. Support Captain Barnaby in the Menu Entry Swapper plugin. Useful for fruit tree runs.

https://imgur.com/a/UhiIr9c

Plondrein:
Barnaby has different options in each location - how would you want that to be configured?
Per location? Or route priority?

Nightfirecat:
I would imagine this would prioritize one menu entry and therefore not do anything for the Barnaby menu options when at the selected location.

Plondrein:
If someone is often uses boat to go Ardy <-> Brim, then he would set 'Brimhaven' option - but would get 'Talk-to' option as first while talking to Barnaby in Brimhaven - not sure if it's intended behaviour, but maybe it's enough as most fruit tree runners just tele out after getting to Brimhaven.

As alternative, we could have options with routes ( A <-> B") and in each location first option would be A given we are at B and vice versa.

ShaunDreclin:
I think having a selection of all possible routes is the best way to do it. `Ardougne <-> Brimhaven`, `Ardougne <-> Rimmington`, `Brimhaven <-> Rimmington`

Similar options for veos, mistag, and any other multi-destination npcs would be good too.

Adam-:
This can be swapped now via the custom left click swap option in dadf3661e13be1e201ade1addadd554fda346cc3

PR #12697, opened 2020-10-26
Brimhaven and Ardougne travel options added to menu entry swapper

honeyhoney:
This updates the menu entry swapper plugin to include swapping `Brimhaven` and `Ardougne` with `Talk-to` as part of the `Travel` option/config.

Rimmington is unfortunately a bit of a second-class citizen. I figured both Brimhaven and Ardougne options are more commonly used and therefore worthwhile prioritising over Rimmington. Brimhaven is prioritised over Ardougne when travelling from Rimmington.

Closes https://github.com/runelite/runelite/issues/11264


Nightfirecat:
I'm inclined to say if we're going to add swaps for this, we should add full configuration options for it

honeyhoney:
> 
> 
> I'm inclined to say if we're going to add swaps for this, we should add full configuration options for it

I agree. I think this PR fills 99% of the population's needs in terms of wanting to swap common travel options but I have wanted to revisit the Travel options for the menu entry swapper to be more configurable as this isn't the only instance of an NPC with multiple conflicting options (conflicts with other destinations and/or "Collect"/"Pay" options etc.).

Adam-:
Superseded by dadf3661e13be1e201ade1addadd554fda346cc3

PR #12698, opened 2020-10-26
plugin-panel: stick plugin hub button to bottom of plugin list

dekvall:
Since the plugin hub "isn't a total disaster" I thought it would be neat if you didn't have to scroll so much when accessing it.


deathbeam (approved):

raiyni:
@dekvall see https://github.com/runelite/runelite/commit/b8456fc1a580cf5d7d5387f3297516b5886ae1a6 for the search layout fix

The `DynamicGridLayout::preferredLayoutSize(parent)` height is not filled to be the same size as `parent.getHeight()`. So the math when it does `final double sh = (1.0 * parent.getHeight() - hborder) / (pd.height - hborder);` results in a scaled height value.

Without it, it looks like 
image

dekvall:
I see, thanks for that!

PR #12700, opened 2020-10-26
agility: Remove Al-Kharid window shortcut highlight from unusable window

### LINKED ISSUE #11759: Agility shortcut highlight of unusable obstacle
Suggestion to remove Agility shortcut highlight of west 'Big window' obstacle on the south wall of Al Kharid palace. 
This one can't be used to leave the palace.

Screenshot: https://i.imgur.com/2PIf3gY.png

andmcadams:
The unusable western window shortcut on the southern wall of Al-Kharid palace has the same object id as the usable one slightly to the east, which caused it to get highlighted along with the usable window. This splits the Al-Kharid window AgilityShortcut into two shortcuts, one for the usable window and one for the unusable window. The usable window has had its location updated to better match it since the old location was in between the two windows. This adds a check to make sure the unusable window is not given an overlay.

Fixes #11759


Adam-:
fixed in 1f3634d429d561165132c0616b051c65592a152f

PR #12701, opened 2020-10-26
Update to the Region Locker Plugin

dkonc:
Added a dropdown menu, where you can select regions that automatically select chunks, that are in that region.


dkonc:
Did it wrong, didnt do it on the correct repo.

PR #12723, opened 2020-10-30
hiscore: correctly look up ironmen players from chat message or self lookup

Hydrox6:
Currently, when using the shortcut to lookup yourself on the hiscores (double clicking the magnifying glass), the normal account tab is used no matter what was set before.

Earlier, a conversation was had on Discord about getting this to also work off of chat message, since the information is available. The PR now covers this.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
@@ -38,7 +38,6 @@ import java.awt.event.KeyListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.image.BufferedImage;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -53,7 +52,6 @@ import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.Experience;
 import net.runelite.api.Player;
-import net.runelite.api.WorldType;
 import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.ui.PluginPanel;
@@ -118,7 +116,7 @@ public class HiscorePanel extends PluginPanel
 		HiscoreEndpoint.NORMAL, HiscoreEndpoint.IRONMAN, HiscoreEndpoint.HARDCORE_IRONMAN, HiscoreEndpoint.ULTIMATE_IRONMAN, HiscoreEndpoint.DEADMAN, HiscoreEndpoint.LEAGUE
 	};
 
-	private final Client client;
+	private final HiscorePlugin plugin;
 	private final HiscoreConfig config;
 	private final NameAutocompleter nameAutocompleter;
 	private final HiscoreClient hiscoreClient;
@@ -138,10 +136,10 @@ public class HiscorePanel extends PluginPanel
 	private boolean loading = false;
 
 	@Inject
-	public HiscorePanel(@Nullable Client client,
-		HiscoreConfig config, NameAutocompleter nameAutocompleter, OkHttpClient okHttpClient)
+	public HiscorePanel(@Nullable Client client, HiscorePlugin plugin, HiscoreConfig config,
+		NameAutocompleter nameAutocompleter, OkHttpClient okHttpClient)
 	{
-		this.client = client;
+		this.plugin = plugin;
 		this.config = config;
 		this.nameAutocompleter = nameAutocompleter;
 		this.hiscoreClient = new HiscoreClient(okHttpClient);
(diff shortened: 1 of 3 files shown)


Adam-:
I think this has already been fixed for looking up yourself, since `resetEndpoints()` calls `selectWorldEndpoint()` which identifies the world type. We could still do the per-account type lookup though but this is conflicting atm.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -138,8 +138,8 @@
 	private boolean loading = false;
 
 	@Inject
-	public HiscorePanel(@Nullable Client client,
-		HiscoreConfig config, NameAutocompleter nameAutocompleter, OkHttpClient okHttpClient)
+	public HiscorePanel(@Nullable Client client, HiscorePlugin plugin, HiscoreConfig config,
+						NameAutocompleter nameAutocompleter, OkHttpClient okHttpClient)
```
Nightfirecat, line 142:
```suggestion
		NameAutocompleter nameAutocompleter, OkHttpClient okHttpClient)
```

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -217,4 +231,25 @@ private void lookupPlayer(String playerName)
...
 		});
 	}
+
+	private HiscoreEndpoint getLocalHiscoreEndpointType()
+	{
+		EnumSet<WorldType> worldType = client.getWorldType();
+		if (worldType.contains(WorldType.LEAGUE))
+		{
+			return HiscoreEndpoint.LEAGUE;
+		}
```
Nightfirecat, line 241: This part of the logic could be replaced by `HiscorePanel#selectWorldEndpoint()`, e.g.:

```java
private HiscoreEndpoint getLocalHiscoreEndpoint()
{
	final HiscoreEndpoint profile = HiscorePanel.selectWorldEndpoint();
	if (profile != HiscoreEndpoint.NORMAL)
	{
		return profile;
	}

	switch (client.getAccountType())
	// ...
}
```

Nightfirecat (changes requested):

Hydrox6:
I've moved some stuff around. 
* `selectWorldEndpoint` is now in the Plugin, renamed to `getWorldEndpoint`
* `getLocalHiscoreEndpointType` now has a null check, and uses `getWorldEndpoint`

Nightfirecat (approved):

Hydrox6:
Earlier, a conversation was had on Discord about getting this to also work off of chat message, since the information is available. The PR now covers this.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -203,13 +207,22 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 			}
 			else
 			{
+				// Prefer the world-based endpoint if the world is special
+				endpoint = getWorldEndpoint();
+				if (endpoint == HiscoreEndpoint.NORMAL)
```
Adam-, line 212: Players can dm you from other worlds, so it probably would be best to just not check this. Or maybe check the flags of the world they are in, but those aren't always available.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -241,14 +254,19 @@ private void insertMenuEntry(MenuEntry newEntry, MenuEntry[] entries)
 	}
 
 	private void lookupPlayer(String playerName)
```
Adam-, line 256: This is only use one other place, might aswell remove it and use the one which takes the endpoint.

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -296,4 +314,32 @@ private HiscoreEndpoint getLocalHiscoreEndpointType()
...
+		}
+		else if (rawName.startsWith(IconID.ULTIMATE_IRONMAN.toString()))
+		{
+			return HiscoreEndpoint.ULTIMATE_IRONMAN;
+		}
+		else if (rawName.startsWith(IconID.HARDCORE_IRONMAN.toString()))
+		{
+			return HiscoreEndpoint.HARDCORE_IRONMAN;
+		}
+		return null;
```
Adam-, line 343: I see no reason why this should ever be able to default to anything other than `NORMAL`

runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
```diff
@@ -296,4 +314,32 @@ private HiscoreEndpoint getLocalHiscoreEndpointType()
 		}
 		return HiscoreEndpoint.NORMAL;
 	}
+
+	private HiscoreEndpoint getHiscoreEndpointFromName(String rawName)
+	{
+		// Remove <col> tags, as they come before the <img> tags we need.
+		Matcher macther = COL_STRIPPER.matcher(rawName);
```
Adam-, line 321: Instead of this, you could just contains() the name with IconID.toString()

Issue #12658, opened 2020-10-17
Allow key remapping of Control key


willediger:
The Control key is used for a somewhat little-known feature in Old School Runescape, [[Control]+click to run](https://oldschool.runescape.wiki/w/Game_controls#Running). This is useful in PvM situations where you need to be able to either switch from walking to running quickly, or vice versa.

If you don't need to run in a situation, but need to be able to run at a certain point quickly in reaction to something, you keep the run orb off, and when you need to run, you [Control]+click to run and do your thing. When you no longer need to run, you just click normally, and you're back to running.

If you don't need to walk in a situation, but need to be able to run at a certain point quickly in reaction to something, you keep the run orb off, [Control]+clicking to run where you need to go, and when you need to walk, you simply click to walk where you need to go.

It's also useful if you are running all the time and have the run orb on, and then forget to drink a stamina potion. If you run until your energy runs out, and you switch to walking, but don't notice it immediately, or aren't bothered by it incredibly at the time, if you realize that you need to run to react to something, [Control]+click is a great way to get running without having to click the run orb first.

This all avoids having to move your mouse first to the run orb to change your speed. Very helpful in intense PvM situations such as Inferno, ToB, and CoX. The problem is that the Control key is far away from where my fingers are normally, and the Control key is small as well. I would like to remap the Control key to [space] so that I can easily hit it while using my F-keys.

I know there's a workaround I could set up by using AutoHotkey, but it would be great if it was built into RuneLite.

I appreciate all of your work on this great client!

mzegar:
@willediger 

Seemed like a reasonable request to me, PR implements rebinding of "Control".

mzegar:
Looks like this issue can be closed : D

Adam-:
added in #12724

PR #12724, opened 2020-10-31
Added ability to rebind control key

mzegar:
Implements #12658

runelite_after

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.keyremapping;
 
+import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
@@ -285,5 +286,16 @@ public interface KeyRemappingConfig extends Config
 	default ModifierlessKeybind space()
 	{
 		return new ModifierlessKeybind(KeyEvent.VK_SPACE, 0);
+	}
+
+	@ConfigItem(
+		position = 21,
+		keyName = "control",
+		name = "Control",
+		description = "The key which will replace {Control}."
+	)
+	default ModifierlessKeybind control()
+	{
+		return new ModifierlessKeybind(KeyEvent.VK_UNDEFINED, InputEvent.CTRL_DOWN_MASK);
 	}
 }
--- a/runelite-client/src/test/java/net/runelite/client/plugins/keyremapping/KeyRemappingListenerTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/keyremapping/KeyRemappingListenerTest.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.keyremapping;
 import com.google.inject.Guice;
 import com.google.inject.testing.fieldbinder.Bind;
 import com.google.inject.testing.fieldbinder.BoundFieldModule;
+import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import javax.inject.Inject;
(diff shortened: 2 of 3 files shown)


Adam-:
Did you look at why it is throwing an exception? That might mean something else is wrong.

mzegar:
> Did you look at why it is throwing an exception? That might mean something else is wrong.

Yeah, I am just completely lost as to why the exception is being thrown. I attempted to make break points in the `.matches(e)` method, but the exception is thrown right after that method is called.

I'm assuming it has something to do with the fact that `mappedKeyCode` cannot already be assigned.

Let me know if there is any other helpful information I could provide.

Adam-:
The stack trace should tell you why... can you push a branch that the test fails on? I could look.

mzegar:
> The stack trace should tell you why... can you push a branch that the test fails on? I could look.

I just pushed it directly to the branch. Check out `KeyRemappingListenerTest.java`. I appreciate the help, I genuinely gave it 30 minutes or so but was just extremely dumbfounded. I use a debugger all the time and I've never seen such behavior (maybe because it's in a test?)

Adam-:
This is just because `config.control()` is `null` in the test due to it not being mocked.

mzegar:
> This is just because `config.control()` is `null` in the test due to it not being mocked.

Thanks, I going to look into how I missed that. Will fix, thanks for taking a look at it.

mzegar:
@Adam- fixed, thanks again for the help.

mzegar:
@Adam- Fixed, thanks for the help.

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
```diff
@@ -276,14 +277,34 @@ default ModifierlessKeybind esc()
 		return new ModifierlessKeybind(KeyEvent.VK_ESCAPE, 0);
 	}
 
+	@ConfigSection(
+		name = "Miscellaneous Remapping",
```
Adam-, line 281: I don't think a new section for this is required. Having no section usually implies miscellaneous.

runelite-client/src/test/java/net/runelite/client/plugins/keyremapping/KeyRemappingListenerTest.java
```diff
@@ -73,6 +74,7 @@ public void testTypingStateChange()
 		when(keyRemappingConfig.down()).thenReturn(new ModifierlessKeybind(KeyEvent.VK_S, 0));
 		when(keyRemappingConfig.left()).thenReturn(new ModifierlessKeybind(KeyEvent.VK_A, 0));
 		when(keyRemappingConfig.right()).thenReturn(new ModifierlessKeybind(KeyEvent.VK_D, 0));
+		when(keyRemappingConfig.control()).thenReturn(new ModifierlessKeybind(KeyEvent.VK_CONTROL, InputEvent.CTRL_DOWN_MASK));
```
Adam-, line 77: You can also just put this in `setUp` I think

runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
```diff
@@ -286,4 +287,15 @@ default ModifierlessKeybind space()
...
+
+	@ConfigItem(
+		position = 22,
+		keyName = "control",
+		name = "Control",
+		description = "The key which will replace {Control}."
+	)
+	default ModifierlessKeybind control()
+	{
+		return new ModifierlessKeybind(KeyEvent.VK_CONTROL, InputEvent.CTRL_DOWN_MASK);
```
Adam-, line 299: From testing on my system and talking to abex, it seems like this should be `new ModifierlessKeybind(KeyEvent.VK_UNDEFINED, InputEvent.CTRL_DOWN_MASK);`. Optionally we could just make this not have a default value and check for null, although we don't do that for the other keybinds.

PR #12737, opened 2020-11-03
xptracker: account for relic modifiers when tracking kills left

dekvall:
The varbits are gathered from [\[proc, league_relics_draw_selections\]](https://github.com/RuneStar/cs2-scripts/blob/1c404f9e5b31ba09661f7f1fbc6af90dbf30f867/scripts/%5Bproc%2Cleague_relics_draw_selections%5D.cs2#L72-L87) and verified with my first three selected.

This also replaces the Endless Endurance HP regeneration modifier from twisted league with fluid strikes from trailblazer.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
@@ -377,7 +377,7 @@ public class XpTrackerPlugin extends Plugin
 		final Actor interacting = client.getLocalPlayer().getInteracting();
 		if (interacting instanceof NPC && COMBAT.contains(skill))
 		{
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -390,6 +392,35 @@ public void onStatChanged(StatChanged statChanged)
...
+
+		if (worldType == LEAGUE)
+		{
+			return (int) (worldType.getXpModifier() * relicModifier());
+		}
+
+		return worldType.getXpModifier();
+	}
+
+	private float relicModifier()
```
Adam-, line 407: I'd probably prefer to just see a `modifier()` overridden in the enum or something

PR #12740, opened 2020-11-03
dpscounter: add autoreset and specific NPC counter

### LINKED ISSUE #12692: DPS Counter Plugin
Currently DPS Counter plugin has 2 issues which makes it useless for many situations.
1) DPS Counter doesn't reset after the target NPC is killed
2) DPS Counter counts damage dealt to any NPC, its not separate for different NPCs, but they are combined

1st issue is when a player uses the plugin to make sure they do most damage to an NPC for example when bossing with an alt account to make sure alt doesn't snioe the kill. If damage doesn't reset, it would be very hard to calculate if the main or alt has dealt more damage for that specific kill.

2nd issue is if you are bossing in a situation where there are multiple NPCs for example like at GWD. DPS counter will count the damage dealt to minions and the boss at the same time, which is not most users want. 

These options could be added as toggles to current plugin. Thanks

akhtar-u:
Close #12692 

Just added features to address the 2 points mentioned in the issue.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsConfig.java
@@ -52,4 +52,26 @@ public interface DpsConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "autoreset",
+		name = "Auto reset",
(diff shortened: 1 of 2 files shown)


Adam-:
added in d8e97dd1f7ac40b2f99e3cba31b78732a9537fcb and e9acabf54915efcdf6e7f8c29aef6d96f245b688

PR #12741, opened 2020-11-03
grounditems: add numerical despawn timers

Hydrox6:
By sudden popular request, here's some numerical despawn timers. Available in both seconds (accurate to 0.1s) and ticks!

record_2020-11-02__23-29-00

The code that calculated the despawn time and pie colour has been refactored out into 2 functions; it's execution should remain unchanged.

The pie will still appear when holding Alt, as it did before, since the numbers would interfere with the controls.

record_2020-11-02__23-36-40


deathbeam:
cant you simply put the timer at beginning where pie is? so it wont interfere with controls (and also so it will be consistent)

Hydrox6:
I tried that before, and the timer jumps around as the width of it changes; it doesn't look good. 

Another reason for keeping the pie when holding Alt is there isn't a way to select something different when it's off; on release, the pie will still appear even when the feature is disabled, and I felt adding an extra checkbox just to be able to select the Alt behaviour was clunky

PR #12742, opened 2020-11-03
zalcano: add damage counter

dekvall:
Adds a panel which tracks damage dealt to Zalcano's health and shield
to make it easier prioritising what to focus on at a low level.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
@@ -28,26 +28,32 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.inject.Inject;
 import lombok.Getter;
+import net.runelite.api.Actor;
 import net.runelite.api.Client;
 import net.runelite.api.GameObject;
 import net.runelite.api.GameState;
 import static net.runelite.api.GraphicID.GRAPHICS_OBJECT_ROCKFALL;
 import net.runelite.api.GraphicsObject;
+import net.runelite.api.Hitsplat;
 import net.runelite.api.NPC;
 import net.runelite.api.NpcID;
 import static net.runelite.api.NpcID.ZALCANO;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
```diff
@@ -183,4 +223,47 @@ public void onProjectileMoved(ProjectileMoved event)
...
+
+	private boolean checkArea()
+	{
+		final Player localPlayer = client.getLocalPlayer();
+		if (localPlayer == null)
+		{
+			return false;
+		}
+
+		final WorldPoint point = WorldPoint.fromLocalInstance(client, localPlayer.getLocalLocation());
```
Adam-, line 266: We think Zolcano is never an instance. Also this means you can move this off of the client thread if you need.

runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
```diff
@@ -132,6 +161,11 @@ public void onNpcDespawned(NpcDespawned event)
 		if (npc.getId() == ZALCANO_WEAKENED)
 		{
 			client.clearHintArrow();
+
+			if (npc.isDead())
```
Adam-, line 165: If you were to leave in the middle of a round this would not reset, is there a better way to do that?

runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
```diff
@@ -111,6 +135,11 @@ public void onGameStateChanged(GameStateChanged event)
 		{
 			rocks.clear();
 		}
+		else
+		{
+			inCavern = checkArea();
```
Adam-, line 140: This and the rocks.clear() should just be in `if (event.getGameState() == GameState.LOADING)`

PR #12745, opened 2020-11-04
camera: increase maximum value for outer zoom limt

Runemoro:
Zooming out only breaks at when the limit is expanded by 1260 or more (resets to fully zoomed in if you try to zoom out more), so I don't see a reason to limit it at 400.


abextm:
Anything over ~700 breaks the scroll wheel. Anything over 400 has some very extreme distortion due to how the projection works, which is why the limit is where it is now.

PR #12748, opened 2020-11-04
Fix antidrag in bank interface

kmarius:
When accessing the bank, the bank widget is loaded and the drag delay is set before the inventory widget is even loaded. This fix works assuming the inventory is always loaded last. If this is not reasonable one should set the delay for each widget individually as they are loaded.


ShuaMiszyn:
Yeah, the inventory tends to take a while to load anti-drag. After a while it works, but as soon as the bank is open anti-drag does not work for the inventory. 
Here's a gyazo showing the issue with anti-drag when it is just opened. https://gyazo.com/412c219a0a59a6eabcbd7b7a867f926b

PR #12752, opened 2020-11-06
AgilityShortcut.java: Update Neitiznot Bridge agility requirements

robuhde:
Update Four Neitiznot bridges to 1 Agility requirement, only northeast bridge to the center island mine requires 40 agility


Adam-:
1 agility is the same as no requirement - should these just be removed?

robuhde:
yes

Adam-:
Removing them also removes the clickboxes, so probably this is ok just changing it to 1.

robuhde:
yeah clickbox is still valuable for these

robuhde:
It is actually a 0 agility requirement technically. I just checked it.

PR #12761, opened 2020-11-08
Fix MES shift click for Ardougne cloak/Karamja gloves

### LINKED ISSUE #10087: If you have menu entry swapped to the ardy patch teleport, you can no longer MES an ardy cloak 1 on other accounts
When a player has menu entry swapped the ardy cloak 2/3/4 to be shift-click teleport to the ardougne farm patch, they can no longer menu entry swap the ardy cloak 1.

1. Have an account with an ardy cloak 2/3/4
2. Right click on the inventory icon, select "configure shift click"
3. Right click your cloak, select "farm teleport"
4. Log out, close your client, whatever needs to be done for it to save your config
5. Restart client, now log in on an account with an ardy cloak 1
6. Right click on the inventory icon, select "configure shift click"
7. Right click your cloak, but no menu entry swapping

It should let the player menu entry swap

https://gyazo.com/c7c2868e93d218528fc48335afef28eb

kmarius:
Fixes #10087. If shift click is configured to an option which doesn't exist, `option` is `null` and causes an exception at `option.equals()`. This allows the shift click option to be reset or switched to one that does exist.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -494,7 +494,7 @@ public class MenuEntrySwapperPlugin extends Plugin
 			{
 				entry.setType(MenuAction.RUNELITE.getId());
 
-				if (option.equals(entry.getOption()))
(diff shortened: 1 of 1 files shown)


Adam-:
fixed in b1ebf7814b114105e4997678922817109709c1d1

PR #12766, opened 2020-11-10
Add 'flash once' option to flash notifications

### LINKED ISSUE #12765: Allow for 'Single Flash' as flash notification option
When using Runelite on a second monitor, a single flash as a notification help get attention through peripheral vision and multiple flashes are overwhelming.

Allow for a single flash as an option in the 'Flash notifications' options

It seems like this issue can be handled in the Notifier.java class. I can try and make this change given some guidance.

Sorgrum:
This allows for a "Flash once" flash notification option. I'm not sure if it flash a bit too quickly but I'm also not sure how I would go about slowing it down so that the duration of the flash is a bit longer.

Fixes #12765

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -247,6 +247,10 @@ public class Notifier
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


Sorgrum:
Closing in favor of PR #12767

PR #12767, opened 2020-11-10
Adds a flash notification setting for 1 singular flash (one tick)

### LINKED ISSUE #12765: Allow for 'Single Flash' as flash notification option

Noahh16:
As per the title, this change adds a new menu option for the flash notification setting to only be a singular tick long.

Fixes #12765

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -229,7 +229,8 @@ public class Notifier
 					return;
 				case FLASH_TWO_SECONDS:
 				case SOLID_TWO_SECONDS:
-					if (Instant.now().minusMillis(FLASH_DURATION_TWO_SECONDS).isAfter(flashStart)) {
+					if (Instant.now().minusMillis(FLASH_DURATION_TWO_SECONDS).isAfter(flashStart))
+					{
 						flashStart = null;
 						return;
 					}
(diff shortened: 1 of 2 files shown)


Sorgrum (approved):
These changes look more complete than the PR I introduced. Will close out mine in favor of this one.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -217,14 +218,22 @@ public void processFlash(final Graphics2D graphics)
...
+				case FLASH_ONE_TICK:
 					flashStart = null;
 					return;
+				case FLASH_TWO_SECONDS:
+				case SOLID_TWO_SECONDS:
+					if (Instant.now().minusMillis(FLASH_DURATION_TWO_SECONDS).isAfter(flashStart))
+					{
+						flashStart = null;
+						return;
+					}
```
Adam-, line 236: this is missing a `break;`

Adam- (commented):
I don't think this actually does one singular flash. There are 30 client ticks in 600ms (600/20) and this flashes the screen repeatedly in 20 ticks on and then 20 ticks off. So, you would get two flashes ~33% of the time.

Issue #12711, opened 2020-10-28
Mac client is impossible to resize if you're moving from a larger monitor to a smaller one


JamesM123:
If the client on Mac is stretched beyond the size of the monitor due to it assuming the same size as the previous session, there's nothing to grab to make the client smaller. This usually results in being unable to click the menu bar, or any of the RuneScape tabs (inventory, prayers, magic etc) as the icons for them are off the bottom of the screen. On the Mac client the resizing can only happen from the bottom of the client or the top, but neither are accessible. I have a feeling the client is centering on the available screen resolution, which results in the top and bottom being inaccessible.

**To Reproduce**
1. Open RuneLite client on monitor with greater vertical size than Mac Retina display.
2. Close RuneLite and disconnect secondary monitor.
3. Reopen RuneLite on the Retina display.
4. Futilely try to resize the client

**Expected behavior**
The Runelite client should probably open aligned to the top of the display so there is an option to resize it down.

**Screenshots**
[
<img width="1450" alt="Screen Shot 2020-10-28 at 21 21 47" src="https://user-images.githubusercontent.com/3850209/97441390-a5db4700-1963-11eb-8dcd-9ea1747d135f.png">
](url)

**Environment (please complete the following information):**
 - OS: macOS
 - RuneLite version: 2.1.1

**Additional context**
[client.log](https://github.com/runelite/runelite/files/5452098/client.log)
[launcher.log](https://github.com/runelite/runelite/files/5452099/launcher.log)

PR #12785, opened 2020-11-12
UI: Fixed window maximizing over screen boundaries on startup

### LINKED ISSUE #12784: Window maximizes over screen boundaries on startup
**Describe the bug**
If the window is closed while maximized on secondary display with lower resolution than main display the window will expand over screen boundaries on next startup.

**To Reproduce**
1. Maximize the window on secondary display that is lower resolution than the primary display
2. Close RuneLite
3. Start RuneLite
4. Window expands over the boundaries of the display and can't be moved since it is in maximized mode 

**Expected behavior**
Window should take the same size on startup that it had when RuneLite was closed

**Environment**
 - OS: Windows 10
 - RuneLite version: 1.6.31
 - Launcher version: 2.1.0

tpssim:
Close #12784 

Use the method overridden in ContainableFrame.java to make sure the boundaries are set correctly.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -554,7 +554,7 @@ public class ClientUI
 					{
 						// Make sure that the bounds are correct for the current display.
 						// The method only uses the parameter on OSX so passing null is ok.
-						if (OSType.getOSType() == OSType.MacOS)
+						if (OSType.getOSType() != OSType.MacOS)
 						{
 							frame.setMaximizedBounds(null);
 						}
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -552,6 +552,12 @@ public class ClientUI
 
 					if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_MAXIMIZED) != null)
 					{
+						// Make sure that the bounds are correct for the current display.
+						// The method only uses the parameter on OSX so passing null is ok.
+						if (OSType.getOSType() == OSType.MacOS)
+						{
+							frame.setMaximizedBounds(null);
+						}
 						frame.setExtendedState(JFrame.MAXIMIZED_BOTH);
 					}
 				}


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -552,6 +552,12 @@ public void show()
 
 					if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_MAXIMIZED) != null)
 					{
+						// Make sure that the bounds are correct for the current display.
+						// The method only uses the parameter on OSX so passing null is ok.
+						if (OSType.getOSType() == OSType.MacOS)
```
Adam-, line 557: do you mean `!=` here...? Your issue/comment/code all disagree with each other

tpssim: Yes actually. Let me fix that real quick.

Adam-:
More generally, do you understand what is going on here and why this fixes it?

Noahh16:
> More generally, do you understand what is going on here and why this fixes it?

If I am seeing this right, he's counting on the fact that the underlying method in  ContainableFrame.java will always recalculate the maximum size (does this because some JDKs incorrectly maximize over the taskbar.)

If that's the case, could this not just pass in null regardless of MacOS or not, since the [overridden method](https://docs.oracle.com/javase/7/docs/api/java/awt/Frame.html#setMaximizedBounds(java.awt.Rectangle)) on Frame.java is called and can accept a null value to indicate using the bounds supplied by the system? That would also fix this other issue a user posted about on Macs doing a similar thing (Issue #12711 )

tpssim:
> More generally, do you understand what is going on here and why this fixes it?



> More generally, do you understand what is going on here and why this fixes it?

The problem seems to be that maximized bounds are set incorrectly to be the ones of the main display even if the frame is on the secondary display. Why this happens and why only at startup I don't know. 
The fix only works because the setMaximizedBounds() is overridden in ContainableFrame to ignore the parameter and calculate the maximized bounds correctly.

tpssim:
> > More generally, do you understand what is going on here and why this fixes it?
> 
> If I am seeing this right, he's counting on the fact that the underlying method in ContainableFrame.java will always recalculate the maximum size for non Mac-os clients (does this because some JDKs incorrectly maximize over the taskbar.)
> 
> If that's the case, could this not just pass in null regardless of MacOS or not, since the [overridden method](https://docs.oracle.com/javase/7/docs/api/java/awt/Frame.html#setMaximizedBounds(java.awt.Rectangle)) on Frame.java is called and can accept a null value to indicate using the bounds supplied by the system? That would also fix this other issue a user posted about on Macs doing a similar thing (Issue #12711 )

I was afraid that passing null in OSX would break the fix in ContainableFrame.java so I chose to prevent it with the if statement since I don't have any OSX system to do testing with

Noahh16:
> I was afraid that passing null in OSX would break the fix in ContainableFrame.java so I chose to prevent it with the if statement since I don't have any OSX system to do testing with

I personally couldn't see why it would, but safer to do that since I don't have one to test on either.

Adam-:
Can  you also test this on the latest launcher (2.1.5) and see if the problem still exists there?

tpssim:
Did some further testing and it seems that ignoring getWindowAreaBounds() in

tpssim:
> Can you also test this on the latest launcher (2.1.5) and see if the problem still exists there?

With the 2.1.5 launcher the problem seems to be different.

1. Close the window while maximized on the secondary display
2. Next time RuneLite is opened it opens maximized on the main display
3.  When minimize(restore) is clicked the window changes it position to the secondary display

JamesM123:
Sorry I haven't been following this, life got in the way and I forgot I'd added this. What happens if you have the window on the secondary display, bigger than the primary display but not maximised, then you close it and reopen? Do you at least have access to resize the window on the topbar or is the client topbar still hidden behind the OSX topbar? I seem to have discovered that there might be a one pixel wide line on the current client where you can actually grab the client, but that's also not really great. If you need me to test a build on an OSX platform, I'm happy to oblige.

Issue #10811, opened 2020-02-16
Music plugin: Make the volume control sliders toggleable


ShaunDreclin:
I prefer the volume control snapping to 25% increments, but if I turn the plugin off then I can't use the sound effect muting or music search features.

abextm:
Just wondering, why do you like not like having granular control of the volume?

ShaunDreclin:
I dunno really haha I just like the feel of the vanilla volume controls

abextm:
That's really weird tbh

PR #12799, opened 2020-11-15
Option to snap volume changes to a specified percentage

Shingyx:
Adds support for snapping volume changes to 5, 10, or 25 percent increments. Defaults to OFF to match current behavior. This makes it easier to change the three sliders to have the exact same volume value.

Closes https://github.com/runelite/runelite/issues/10811.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicConfig.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2019, Adam <Adam@sigterm.info>
+ * Copyright (c) 2020, Shingyx <https://github.com/Shingyx>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


Adam-:
I can't really tell what this is doing, but it isn't doing what #10811 wants, since that just wants us to not replace the volume sliders.

Shingyx:
Setting the option to "Nearest 25%" will solve the linked issue. The extra changes in this PR were more for me personally.

I like having _some_ granularity but "25% snap or nothing" isn't enough. I like to have the volume at 20% while it's the main game I'm playing, but more like 5-10% when it's in the background doing some AFK task. Without any snapping, it can be annoying to try get them all to match at 5% or 20%.

Adam-:
I suppose, but some people really like the vanilla feature and I do find it odd our plugin doesn't have the ability to turn off  this feature, since we do usually have config opts for stuff like this.

Adam-:
The game update today rewrote all of the volume interfaces, so we've removed our volume sliders in order to rewrite them. So we'll have to revisit this later.

PR #12806, opened 2020-11-17
Fishing: Show fishing spot overlay when crystal harpoon is equipped/in inventory

jagmeetb:


makingcomment:
The trailblazer harpoon probably has the same issue

Paveldin:
Solution proposed here: #12829

PR #12814, opened 2020-11-18
Calculate price of rune pouch contents

### LINKED ISSUE #9072: Include the value of runes inside rune pouches to be counted towards Bank's Value
It'd be nice to have the bank value include the values of runes inside the rune pouch if the rune pouch is in the bank

MKuijpers:
- Calculate rune pouch price in `getItemPrice` of `ItemManager`
- Moved `Runes` enum from the rune pouch plugin to the RuneLite API
- Extracted the rune pouch varbits from the rune pouch plugin to `RunePouchVarbits` in RuneLite API

Fixes #9072


MKuijpers:
Also, I know there was already another PR (#8645) created for this issue. However, there was still some discussion in the PR, and not much has happened for 1.5 years. I thought I would pick it back up with those discussion points in mind. If this PR is accepted, the other one could be closed I think.

Adam-:
I don't really like the idea of special casing specific items within the item manager getPrice(). It is also possible to have multiple runepouches, in which case this would total your bank value incorrectly due to double counting all of the runes. What plugins really need this? Maybe just item prices and bank value? Can we just work it into this individually?

MKuijpers:
It was mainly made for the bank value and examine plugins, but there could be more plugins that would like to have the rune pouch value, I am not sure. Therefore I implemented the getPrice() method, so all plugins would get the rune pouch value in the future.
Is there any particular reason to not add it in the item manager getPrice() method, besides minimizing clutter of different items in the getPrice() method?

PR #12818, opened 2020-11-19
cluescrolls: Add X Marks The Spot quest solutions

Nightfirecat:
This mostly speaks for itself. `itemComposition != null` is removed from the plugin as it is vestigial--it will never be false.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
@@ -323,11 +323,22 @@ public class ClueScrollPlugin extends Plugin
 		{
 			return;
 		}
-		if (event.getMenuOption().equals("Read"))
+
+		final boolean itemClicked = event.getMenuAction() == MenuAction.ITEM_FIRST_OPTION
+			|| event.getMenuAction() == MenuAction.ITEM_SECOND_OPTION
+			|| event.getMenuAction() == MenuAction.ITEM_THIRD_OPTION
+			|| event.getMenuAction() == MenuAction.ITEM_FOURTH_OPTION
+			|| event.getMenuAction() == MenuAction.ITEM_FIFTH_OPTION;
+		final boolean isXMarksTheSpotOrb = event.getId() == ItemID.MYSTERIOUS_ORB_23069;
+
+		if (itemClicked && (isXMarksTheSpotOrb || event.getMenuOption().equals("Read")))
 		{
 			final ItemComposition itemComposition = itemManager.getItemComposition(event.getId());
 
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -289,11 +289,16 @@ public void onOverlayMenuClicked(OverlayMenuClicked overlayMenuClicked)
 	@Subscribe
 	public void onMenuOptionClicked(final MenuOptionClicked event)
 	{
-		if (event.getMenuOption() != null && event.getMenuOption().equals("Read"))
+		final boolean isXMarksTheSpotOrb = event.getId() == ItemID.MYSTERIOUS_ORB_23069;
```
Adam-, line 292: You should verify the menu type before assuming the id is an item id. I think the old code works because you can getItemComposition() with any id and it always returns a non null item with the name `"null"` if the id isn't valid. However this is bad practice and we should prefer to remove that whenever we see it.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CipherClue.java
```diff
@@ -88,12 +96,15 @@ private CipherClue(String text, String npc, WorldPoint location, String area, @N
 	@Override
 	public void makeOverlayHint(PanelComponent panelComponent, ClueScrollPlugin plugin)
 	{
-		panelComponent.getChildren().add(TitleComponent.builder().text("Cipher Clue").build());
-		panelComponent.getChildren().add(LineComponent.builder().left("NPC:").build());
-		panelComponent.getChildren().add(LineComponent.builder()
-			.left(getNpc())
-			.leftColor(TITLED_CONTENT_COLOR)
-			.build());
+		if (getNpc() != null)
```
Adam-, line 99: Use a variable for `getNpc()`

Nightfirecat:
I've addressed both of your comments

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CipherClue.java
```diff
@@ -88,12 +100,16 @@ private CipherClue(String text, String npc, WorldPoint location, String area, @N
...
-		panelComponent.getChildren().add(TitleComponent.builder().text("Cipher Clue").build());
-		panelComponent.getChildren().add(LineComponent.builder().left("NPC:").build());
-		panelComponent.getChildren().add(LineComponent.builder()
-			.left(getNpc())
-			.leftColor(TITLED_CONTENT_COLOR)
-			.build());
+		final String clueNpc = getNpc();
+		if (clueNpc != null)
+		{
+			panelComponent.getChildren().add(TitleComponent.builder().text("Cipher Clue").build());
```
Adam-, line 106: Shouldn't this be outside of the if? It's part of the overlay title

PR #12824, opened 2020-11-21
timetracking: add leagues botanist multiplier

MadCatzAno:
Adds the Botanist ×5 multiplier to the timers.


abextm (changes requested):
You can't assume the user is logged in from `predictPatch`

abextm:
Added in 374cf05c6d7481eea4f9446696bbef2e78b18905

PR #12835, opened 2020-11-23
Charge lasts for seven minutes, not six.

emerald000:


Adam-:
I've just tested this ingame with 1 charge and it lasted 7 min 22 sec, so I don't think it lasts a fixed time.

emerald000:
After further testing, it looks like the timer gets extended while an interface stall is happening (such as being in a bank).

PR #12843, opened 2020-11-26
Blurry screen marker fix

### LINKED ISSUE #12821: Screen Markers: Markers with even thickness are blurry
**Screenshots**
Screenshot

**Environment:**
 - OS: Windows 10
 - RuneLite version: 1.6.33-SNAPSHOT

waspider:
Fix #12821 blurry screen markers on every even number of thickness.


Adam-:
Do you know why this requires an `KEY_INTERPOLATION` value when antialiasing is also disabled? The docs say `KEY_INTERPOLATION controls how image pixels are filtered or resampled during an image rendering operation` - however I don't see why they would be resampled with AA off?

waspider:
Yeah that was a mistake on my end when I was looking through the docs. I'll edit it when I get the chance

PR #12844, opened 2020-11-26
add item wiki to popup in grandexchange and loottracker plugins

### LINKED ISSUE #12840: Wiki button to work on Grand Exchange plugin items
I always find myself wanting to quickly look up items on the wiki while searching with the GE plugin.

Currently, clicking an item in the GE plugin opens the listing on the official GE guide website. Clicking the Lookup Wiki button and then clicking a GE plugin item does not open the relevant wiki page.

Clicking the Lookup Wiki button and then a GE plugin item should open the relevant wiki page.

I'm not familiar with how these two elements work behind the scenes, so if this isn't possible, there should be an option directly in the GE plugin to do this. Maybe a `right click -> Wiki [x]` option similar to right clicking a skill on the skills tab?

No alternatives besides manually opening and searching the wiki.

**Additional context**
n/a

acdvs:
Adds wiki links to the right click menus on items in the Grand Exchange search tab and the Loot Tracker.

Closes #12840


Adam-:
Both of these should be looking up items by id, not by name. It avoids name collisions from multiple items having the same name. Look at how the wiki plugin does it to see how to do it.

Nightfirecat:
Needs rebase

PR #12851, opened 2020-11-26
Add individual value search to bank plugin.

### LINKED ISSUE #12771: Bank search values should have an individual search
When searching your bank you can type ha 100k and it would show everything over 100k.

What I suggest is another option to search ha iv 100k and it show only items that have an individual price of 100k.

For example, If I searching ha 100k, 100k bronze arrows would show up as well as dragon platebody. With the new method only the dragon platebody would show.

You would never alch the bronze arrows but you would the dragon platebody.

waspider:
Added additional functionality to bank value search where you can specify with  "i", "iv", or "individual" to search for the individual value of an item. This is particularly useful for when you want to find items with large individual high alch values and want to filter out high quantity items like feathers, arrows, etc. because nobody alchs those.

example usage: 
"ha i > 20k"
"ha iv > 20k"
"ha individual > 20k"

close #12771

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
@@ -83,7 +83,7 @@ public class BankPlugin extends Plugin
 	private static final String SEED_VAULT_TITLE = "Seed Vault";
 
 	private static final String NUMBER_REGEX = "[0-9]+(\\.[0-9]+)?[kmb]?";
-	private static final Pattern VALUE_SEARCH_PATTERN = Pattern.compile("^(?<mode>ge|ha|alch)?" +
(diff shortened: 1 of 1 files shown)


raiyni:
That form of search doesn't really make sense to ever discover

waspider:
> That form of search doesn't really make sense to ever discover

Can you elaborate? It is useful for finding alchables lying around your bank that have decent HA value to them that you want to clean out. The issue with the current system is that it always aggregates the total value of an item stack so you end up getting items that are pointless to alch like runes, feathers, arrows, etc.

Also, the original functionality of aggregate value is still there

raiyni:
I mean the search activator phrasing

waspider:
> I mean the search activator phrasing

Any suggestions of what i should change it to? It allows a few phrases already that point to the same functionality. Can you give me an example usage that is acceptable?

waspider:
maybe it should be a checkbox in the bank plugin section instead?

Adam-:
This seems okay. Can you extend the existing test for bank value search for this?

waspider:
Yeah when I get the chance. I'll let you know

Fritz231 (approved):

PR #12868, opened 2020-12-01
Add visual metronome 

### LINKED ISSUE #12850: Visual metronome for hearing impaired players
Currently, if you're doing actions that require a metronome, like tick manipulation, the metronome i a good guide. However, if you are watching a movie, or listening to music, the additional beat might throw you off the rythm

Similar to the prayer orb, or another vosual counter that hops inbetween selected colors if the beat is not linear. it flashes, or slides from side to side to mimic the rythm of one or multiple tick.

similar to what once was on vorkath tick timer in the past, but make it movable around the client so that it can be places next to the inventory, or next to player switches where it'd be most fit. scaling and changing colors could also be nice. aswell as selecting inbetween color switching inbeween eacother to represent a tick, or hat it just flashes for each tick

**Additional context**
no.

acdvs:
Adds a visual metronome (a flashing circle) with several color and size options to the metronome plugin. Ties directly into the existing metronome functionality and makes use of overlay panels to allow dragging to reposition on the screen.

Resolves #12850 

demo

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePluginConfiguration.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePluginConfiguration.java
@@ -37,11 +37,11 @@ public interface MetronomePluginConfiguration extends Config
 	int VOLUME_MAX = SoundEffectVolume.HIGH;
 
 	@ConfigItem(
-		keyName = "tickCount",
-		name = "Tick count",
-		description = "Configures the tick on which a sound will be played."
+		keyName = "tickInterval",
+		name = "Tick interval",
+		description = "The game tick interval after which a sound will be played."
 	)
-	default int tickCount()
+	default int tickInterval()
 	{
 		return 1;
 	}
@@ -52,7 +52,7 @@ public interface MetronomePluginConfiguration extends Config
 	@ConfigItem(
 		keyName = "tickVolume",
 		name = "Tick volume",
-		description = "Configures the volume of the tick sound. A value of 0 will disable tick sounds."
+		description = "The volume of the <em>tick</em> sound. A value of 0 will disable <em>tick</em> sounds."
 	)
 	default int tickVolume()
 	{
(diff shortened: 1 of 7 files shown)


Bram91:
Could you add some screenshots/gifs of this?

Adam-:
Is this still something we need? There is a visual metronome plugin on the pluginhub now. I'm also not sure why you are doing this `MetronomeComponent` <-> `MetronomePanel` <-> `MetronomeOverlay` thing since you have a single-component overlay it seems unnecessary to build a dedicated component for it?

acdvs:
@Adam- I needed custom panel features (implemented in `MetronomeComponent`) and I figured creating `MetronomePanel` to replace `OverlayPanel` was the only way to do that. I'm no expert in Java so if you know a better way to do this I'll change it.

To answer your first question, I do think this would be beneficial as a native feature rather than a plugin since a lot of people play with no sound. There's also the accessibility argument as well.

runelite-client/src/main/java/net/runelite/client/plugins/metronome/MetronomePluginConfiguration.java
```diff
@@ -25,23 +26,37 @@
...
+		position = 4
+	)
+	String visualSettings = "visualSettings";
 
 	@ConfigItem(
-		keyName = "tickCount",
-		name = "Tick count",
-		description = "Configures the tick on which a sound will be played."
+		position = 1,
+		keyName = "tickInterval",
```
Nightfirecat, line 55: Changing the key name will reset user's configs--don't do that.

Nightfirecat (commented):
I'm also confused by `MetronomePanel`. Why can't the component be a field of `MetronomeOverlay` which can extend `OverlayPanel` and be set to be resizable?

DaRealMajister:
hello, someone whose trying his best to learn how to operate GitHub to see if they can start contributing like you legends. 

I was hoping to expand upon the metronome system to create a plugin called "OSRS rhythm finder". It would operate very similarly to a visual metronome like you've setup here but more gamified. Like playing guitar hero it would have a scrolling tick system that would display upcoming player or enemy attacks. This way players could have more accurate manual prayer flicking. Ideally it would be placed in the upper right screen right below exp drops so players would have both visual queues for damage displayed beside each other. 

I'm not at all an experienced coder but I would love to see something like this in game so I don't get carpal tunnel trying to 1tick prayer flick when I want to save on prayer pots.

PR #12880, opened 2020-12-03
Added Offenders

vitorebom:
Created a list of Offenders which can be updated for more offenders and updated if statements to include them. Tested and it works as intended:
- Using a special attack will add the counter
- Using another special attack on the same boss will update the counter
- Attacking one of the offenders will not reset the counter
- Using a special attack on one of the offenders will not update the counter
- Attacking another npc will reset the counter

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -28,17 +28,10 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import javax.inject.Inject;
+
+import com.google.common.collect.ImmutableSet;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.Actor;
-import net.runelite.api.Client;
-import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.GameState;
-import net.runelite.api.Hitsplat;
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemContainer;
-import net.runelite.api.NPC;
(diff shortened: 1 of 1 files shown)


Adam-:
Do you know what this `dark core` npc is?

vitorebom:
My bad, thought the Dark Energy Core had more Ids, just tested and both of there are not the actual NPC Corp spawns:
- DARK_CORE
- DARK_CORE_388
Only the DARK_ENERGY_CORE is the one that's supposed to be in the Offenders list

Adam-:
I've taken this pr and made https://github.com/Adam-/runelite/commit/589a362d140b098a08316ddbed81512a2ee4b998 instead, can you test that?

vitorebom:
On your version, specing one of the ignored npcs will still add to the counter but not remove it

Adam-:
> 
> 
> On your version, specing one of the ignored npcs will still add to the counter but not remove it

hmm do you see why? It should be just resetting the counter and counting the new spec instead. Did you test it and get this behavior?

vitorebom:
LN177  if (!wasSpec && IGNORED_NPCS.contains(interactingId))
With this logic the ignored npcs will only matter for normal hits that aren't specs, let me test removing the spec part completely

Adam-:
..? My intent is if you *spec* one of the ignored npcs it still counts. Since that doesn't happen normally. Right? Except maybe for the dummy...

vitorebom:
You are correct, and honestly, the dummy only mattered for Trailblazer as a make-do ornate pool.
I can't think of any other scenarios when we would Spec one of the Ignored NPCs

Adam-:
If the dummy is to be included then probably the `!wasSpec` should be removed. Maybe there is some niche use for it now or in future leagues.

Adam-:
added in aa1cd2c7ff79e368db6769d1c09f6b8a11c917d9

PR #12881, opened 2020-12-03
Added overlay tooltip to motherlode plugin

0anth:
Added a tooltip to the ores overlay, which displays estimated GE values.

EY73Ypk

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeConfig.java
@@ -143,4 +143,14 @@ public interface MotherlodeConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "oreValueMouseover",
+		name = "Show ore values overlay tooltip",
+		description = "Displays GE values for ores found in ore overlay tooltip"
+	)
+	default boolean oreValueMouseover()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


0anth:
Removed mouseover overlay. 

Ore values is now takes the place of "Ores found" overlay upon mouseover.

vCdrI6K

Still image:

IUXT3dP

0anth:
Changed to dropdown in config and added shift right click options to the overlay.

HWsOoZN

cvniBus

runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodePlugin.java
```diff
@@ -208,6 +211,19 @@ public void onOverlayMenuClicked(OverlayMenuClicked overlayMenuClicked)
 		{
 			session.resetRecent();
 		}
+
+		if (overlayMenuClicked.getEntry() == MotherlodeOreOverlay.DISABLED_ENTRY)
```
Nightfirecat, line 215: These should be checking the clicked overlay, similar to the reset branch above. In fact, that should be added as a guard clause so none of these code paths can be hit should some other overlay be clicked.

Nightfirecat (changes requested):
I'm not sure if we want to add this, given we have no other overlays which exhibit different behavior when hovered. The only existing instance of an overlay using `onMouseOver()` is the DPS plugin, which adds a tooltip. This would be a more notable change in behavior by contrast.

0anth (commented):
Updated code as per suggestions.

0anth:
> I'm not sure if we want to add this, given we have no other overlays which exhibit different behavior when hovered. The only existing instance of an overlay using `onMouseOver()` is the DPS plugin, which adds a tooltip. This would be a more notable change in behavior by contrast.

I'm happy to remove the hover function, though it is not enabled by default and must be switched on in config first. This gives the user the decision on whether to enable it or not. 

I trialled using the hover option at motherlode mine for a while and it actually "felt" nice in that context. As majority of the time at mlm, you are stood idly waiting, not moving your mouse around the screen. But it allows you to quickly check the current values, then go back to quantities without having to keep switching configs.

Nightfirecat (commented):
Coming back to this with a fresher perspective, I think I'd like this a lot more if it were a simple config to toggle display of a single large, multi-line tooltip with ore values (similar in design to the item stats tooltip) when hovering the overlay, similar to the screenshot in the PR, and removing the feature of changing the overlay's text on hover completely.

PR #12889, opened 2020-12-05
music: re-add slider granularity 

### LINKED ISSUE #10811: Music plugin: Make the volume control sliders toggleable

abextm:
Requires !123
Closes #10811


Seppobb:
Commented on previous threads, would be amazing to get this back! Helps out miles with inferno.

PR #12895, opened 2020-12-07
camera: add slider tooltip to main settings window 

abextm:


Adam-:
This require some sort of validity check on the widget 
image

PR #12901, opened 2020-12-08
[bug] barrows coloring not aligning with drop table

TheRealNoob:
This will make the coloring work as follows.  I believe this was the intended solution?
WHITE pre death rune
YELLOW include death runes
GREEN include blood runes
YELLOW include bolt racks

see drop table rates
https://oldschool.runescape.wiki/w/Chest_(Barrows)#Main_table

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -87,7 +87,7 @@ public class BarrowsBrotherSlainOverlay extends OverlayPanel
 		panelComponent.getChildren().add(LineComponent.builder()
 				.left("Potential")
 				.right(rewardPercent != 0 ? rewardPercent + "%" : "0%")
-				.rightColor(client.getVar(Varbits.BARROWS_REWARD_POTENTIAL >= 756f && client.getVar(Varbits.BARROWS_REWARD_POTENTIAL <= 880f ? Color.GREEN : client.getVar(Varbits.BARROWS_REWARD_POTENTIAL < 631f ? Color.WHITE : Color.YELLOW)
+				.rightColor(client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) >= 756f && client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) <= 880f ? Color.GREEN : client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) < 631f ? Color.WHITE : Color.YELLOW)
 				.build());
 
 		return super.render(graphics);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsBrotherSlainOverlay.java
@@ -83,8 +83,8 @@ public class BarrowsBrotherSlainOverlay extends OverlayPanel
 				.build());
 		}
 
-		float rewardPotential = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL);
-		float rewardPercent = client.getVar(Varbits.BARROWS_REWARD_POTENTIAL) / 10.12f;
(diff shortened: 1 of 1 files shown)


Adam-:
It may be good to have the contents of https://pbs.twimg.com/media/CcovSPiXIAEPKjm?format=png&name=large in a comment here to explain this magic. Also, you want `< 63.1` instead of `< 63.0` for the death rune check I think.  It may be also be good to verify that this percentage is actually out of 1000 like you are assuming and not 1012, which is what I would guess it is based on this picture.

TheRealNoob:
Good catches, I missed both of those.
So I changed the percentage to calculate off the 1012 instead of 1000.  And then correspondingly updated the reward percentages to be calculated by the same.  So your `63.1` has become `62.35 `.  The problem with the approach is that reward percentage is based off comparing percentages which are rounded off.  I'm sure the game is comparing the 4-digit reward potential.  It would be preferable if we could do the same, and then divide by `10.12` just for the display in the panel, but I don't have the Java skill to do it (never written Java).  I also don't have a way to test the code before commiting it, etc etc.  However, I am sure this solution is fine.  We're talking about a fraction of a percent and I'm sure people will survive.

TheRealNoob:
I gave it another look and I think this makes sense.  Comparing the in-game loot potential vs computed percentages.  I trust you'll tell me if this won't work, but pretty sure it will

Adam-:
Did you verify that the percentage is out of 1012 in game?

TheRealNoob:
Another good point.  No I didn't, but I have now.  Turns out the vanilla client displays percentages as well which makes it a little harder.  I killed zero brothers and just skeletons adding up their combat levels.  On the left calculator you can see the total of their combat levels and on the right, combat levels / 1012.  It looks like the vanilla client just drops the decimals rather than rounding.

https://imgur.com/a/EZ4J85V
https://imgur.com/a/NaEhOfH

TheRealNoob:
I just wonder if there isn't a way to communicate to the player what the colors of the potential (in the panel) means.  I remember for the longest time I didn't know.  Currently it's not explained anywhere.  It's not configurable so it doesn't fit nicely into the plugin config.

Adam-:
This is the code the client uses

```
def_int $int2 = calc(%varbit457 + %varbit458 + %varbit459 + %varbit460 + %varbit461 + %varbit462);
def_int $int3 = calc(%varbit463 + $int2 * 2);
if_settext("Rewards potential: <tostring(interpolate(0, 100, 0, 1012, $int3))>%", $component0);
```

varbit457-462 each being 0-1 for whether each brother was killed, 463 being the reward potential. So, the UI is percentage out of 1012, but the var is out of 1000. 

Probably the best thing to do then is to compute the $int3 value and then check it against the known thresholds (631, 765, 881).

TheRealNoob:
That looks like exactly how the wiki explained it `reward potential is the sum of all the combat levels of the monsters killed, capped at 1000 - including the brothers (all six have a combat sum of 656). This is then added to the two times the number of brothers killed, giving a total maximum reward potential of 1012.`

it sounds like varbits.BARROWS_REWARD_POTENTIAL is $int3.  But I can't rule out that it's not varbit463 either.  I have to make some assumptions here, but I'm guessing that what you shared is obfuscated variables.  If `varbits.BARROWS_REWARD_POTENTIAL  == varbit463` then that would explain why varbits.BARROWS_REWARD_POTENTIAL was being divided by 10 (and not 10.12) before I opened this PR.  Do you have a way to confirm what varbits.BARROWS_REWARD_POTENTIAL is?

Adam-:
We know what the varbits are
```java
	BARROWS_KILLED_AHRIM(457),
	BARROWS_KILLED_DHAROK(458),
	BARROWS_KILLED_GUTHAN(459),
	BARROWS_KILLED_KARIL(460),
	BARROWS_KILLED_TORAG(461),
	BARROWS_KILLED_VERAC(462),
	BARROWS_REWARD_POTENTIAL(463),
	BARROWS_NPCS_SLAIN(464),
```

TheRealNoob:
okay perfect i'll redo the math now

TheRealNoob:
I would have preferred to simply do `varbits(463) + varbits(464)` but since you said the client doesn't use 464 i figure that makes it an unknown.  At least in that I can't confirm what 464 is.  So I added them all up same as the client.

TheRealNoob:
This java syntax is throwing me off, but if i'm reading this right the rewardPercent is not going to be rounded off at all is it?

Adam-:
fixed in 807a22172f0fd09a89f1e72e35e5eb7fb6816881

PR #12918, opened 2020-12-11
12916 fix barrows overlay rounding

### LINKED ISSUE #12916: Barrows overlay percentage shows too many digits
Following the merging of 807a22172f0fd09a89f1e72e35e5eb7fb6816881, the barrows overlay shows too many digits for the reward potential. It should be capped to some smaller number, perhaps 2 digits? eg. `12.34%`

solo-fish:
Closes #12916 

Currently, the reward percentage displayed on the BarrowsBrotherSlainOverlay isn't rounded - this simple change rounds the percentage to the second decimal place.


TheRealNoob:
tested.  works 👍

Adam-:
In-game it's just shown as an int, which is probably also fine. Then it's just `int rewardPercent = (int) (rewardPotential / 10.12f);`

PR #12919, opened 2020-12-11
music: make volume percent always visible

### LINKED ISSUE #12909: Volume slider funk
**Describe the bug**
Volume % tooltip only appears over volume bar, not when using the slider.

**To Reproduce**
1. Hover over volume slider

**Expected behavior**
Show tooltip on both slider and volume bar

Thank you friends

TaylorAbraham:
Fixes #12909

JPpF8VBPUZ

Behaviour before:
- Tooltip only shows up when hovering over the part of the slider WITHOUT the handle

Behaviour after this PR:
- Tooltip now shows up when hovering over the slider, when hovering over the handle, and when dragging the handle

Current Issues:
- I was *not* able to get it to display the tooltip when you press down on the mouse but haven't started dragging yet. I tried using `handle.setOnClickListener` and `handle.setOnHoldListener` but it still did not cause the tooltip to show up in those cases.


runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
```diff
@@ -466,17 +466,20 @@ private void updateList(Widget musicList, String filter)
...
+			JavaScriptCallback setSliderTooltip = ev ->
 			{
 				int value = channel.getValue();
 				int percent = (int) Math.round((value * 100.0 / channel.getMax()));
 
 				sliderTooltip = new Tooltip(channel.getName() + ": " + percent + "%");
-			});
+			};
+
+			handle.setOnMouseRepeatListener(setSliderTooltip);
```
Adam-, line 477: This additionally needs to be unset in shutdown.

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
TaylorAbraham: Yeah I was curious about this. This line actually exists currently (I just refactored it) and was never unset in shutdown. Should we still add that in?

Adam-: This line actually doesn't exist currently

TaylorAbraham: Ah right my bad, it only existed on the slider and not the handler.

abextm, line 477: you can avoid this listener by setting noClickThrough to false

TaylorAbraham: Good call; fixed

TaylorAbraham:
I verified that the tooltip still always shows up with the latest changes I've made.

Nightfirecat:
I've squashed and condensed your changes in this commit: https://github.com/Nightfirecat/runelite/commit/fb68bacfccf8fcdd8041a77fa37e1c62a6812178
I'd have pushed them to this PR, but force-pushes are rejected by your master branch

Nightfirecat:
Added in 27fb56f5b8cdfc8b376b1d42b48953e6c348277f

PR #12932, opened 2020-12-14
Raw GP Banked Option for Bank Plugin

Ricklet:
Added an option to show the raw gp a player has banked.  This includes both gp/plats, items deemed as alchables by the isAlchable function, and processable items such as hides, bars, and ores. Processable items are only counted if the player has all of the required materials and the required level to process them. Useful tool for ironmen to see how much gp they have stored in items.


Nightfirecat:
In my opinion this greatly exceeds the bounds of what we want this plugin to do. Not every user will have the same idea of what they want to process, or what kind of margin would make it worth their while to alch something rather than selling it. On top of that, this PR is riddled with hardcoded raw values that don't seem to correspond to anything in particular?

I don't think we are likely to add this to the main client, though you are welcome to [submit this to the plugin hub](/runelite/plugin-hub) if you'd like to enjoy it for your own use.

PR #12936, opened 2020-12-15
loottracker: added bird nests to track type

weiyih:
~~This modifies the current bird nest loot tracking to improve the wiki data sets. Currently, the system combines all types of bird nests and it is tracked under "Bird nest".~~

~~**Changes made:**~~
~~- Wyson nests -> Bird nest (Wyson)~~
~~- Regular nests (miscellania, birdhouses, woodcutting) - Bird nest (Seed)~~
~~- Ring nests -> Bird nest (Ring)~~
~~- Egg nests-> Bird nest (Egg)~~
~~- Empty nests -> Not tracked~~

~~_Note: Historical nests with the ID 5073 and 13563 are tracked with under the -> Bird nest (Old) and (Old v2) respectively for data purposes._~~

Update: This PR modifies the current bird nest loot tracking to include the object id as metadata.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -477,8 +477,8 @@ public class LootTrackerPlugin extends Plugin
 		if (config.npcKillChatMessage())
 		{
 			final String prefix = VOWELS.contains(Character.toLowerCase(name.charAt(0)))
-					? "an"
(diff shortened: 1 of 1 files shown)


Nightfirecat:
The metadata field is not used by the client, it is thus far only useful to the wiki for their drop log project. @leejt Has this approach been signed off on with y'all?

leejt:
> The metadata field is not used by the client, it is thus far only useful to the wiki for their drop log project. @leejt Has this approach been signed off on with y'all?

I can confirm this is going to be used on the wiki side.

PR #12944, opened 2020-12-17
add skull sceptre 'invoke' to teleport item menu entry swapper

### LINKED ISSUE #10295: Menu Entry Swapper - Teleport Item Skull Sceptre(i) request
I just enabled my Menu Entry Swapper and I still can't believe this is not bannable! 
**Problem description**
The only thing that is not satisfying is that the Skull Sceptre(i) does not have his options switched. Maybe it is because the teleport option is called "Invoke". Probably the unimbued version has the same problem. 

**Solution**: Add the Menu Entry Swapper support for the Skull Sceptre(i) (and unimbued version).

Untitled

jamesahhh:
Fixes #10295
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -319,6 +319,7 @@ public class MenuEntrySwapperPlugin extends Plugin
 		swap("wear", "rub", config::swapTeleportItem);
 		swap("wear", "teleport", config::swapTeleportItem);
 		swap("wield", "teleport", config::swapTeleportItem);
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -319,6 +319,7 @@ void setupSwaps()
 		swap("wear", "rub", config::swapTeleportItem);
 		swap("wear", "teleport", config::swapTeleportItem);
 		swap("wield", "teleport", config::swapTeleportItem);
+		swap( "wield", "invoke", config::swapTeleportItem);
```
Nightfirecat, line 322: I'll commit this via GitHub's UI
```suggestion
		swap("wield", "invoke", config::swapTeleportItem);
```

PR #12947, opened 2020-12-18
Add farming tick offsets to time tracking plugin

### LINKED ISSUE #8381: Farming timer optimization

MESLewis:
Automatically tracks and applies your accounts offset to farming ticks. For a general user, your farming times will become more accurate over time as you inadvertently observe crops growing.

For anyone looking to quickly get accurate times the simplest solution is to plant a normal tree and keep it within your render distance for 40 minutes. After observing the tree growing one stage your timers for all crops will be 100% accurate.

Closes #8381
Requires !127

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
@@ -33,6 +33,8 @@ import net.runelite.client.config.Units;
 public interface TimeTrackingConfig extends Config
 {
 	String CONFIG_GROUP = "timetracking";
+	String FARM_TICK_OFFSET = "farmTickOffset";
+	String FARM_TICK_OFFSET_PRECISION = "farmTickOffsetPrecision";
 	String AUTOWEED = "autoweed";
 	String BIRD_HOUSE = "birdhouse";
 	String BOTANIST = "botanist";
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPanel.java
@@ -25,6 +25,8 @@
  */
 package net.runelite.client.plugins.timetracking;
 
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
 import java.awt.GridLayout;
@@ -38,16 +40,18 @@ import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.SwingUtilities;
 import javax.swing.border.EmptyBorder;
-import net.runelite.client.util.AsyncBufferedImage;
+import net.runelite.client.config.ConfigManager;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.plugins.timetracking.clocks.ClockManager;
 import net.runelite.client.plugins.timetracking.farming.FarmingContractManager;
+import net.runelite.client.plugins.timetracking.farming.FarmingNextTickPanel;
 import net.runelite.client.plugins.timetracking.farming.FarmingTracker;
 import net.runelite.client.plugins.timetracking.hunter.BirdHouseTracker;
 import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.PluginPanel;
 import net.runelite.client.ui.components.materialtabs.MaterialTab;
 import net.runelite.client.ui.components.materialtabs.MaterialTabGroup;
+import net.runelite.client.util.AsyncBufferedImage;
 
 class TimeTrackingPanel extends PluginPanel
 {
@@ -64,9 +68,11 @@ class TimeTrackingPanel extends PluginPanel
 	@Nullable
 	private TabContentPanel activeTabPanel = null;
 
-	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config,
-		FarmingTracker farmingTracker, BirdHouseTracker birdHouseTracker, ClockManager clockManager,
-		FarmingContractManager farmingContractManager)
+	@Inject
+	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config, FarmingTracker farmingTracker,
+		BirdHouseTracker birdHouseTracker, ClockManager clockManager,
+		FarmingContractManager farmingContractManager, ConfigManager configManager,
+		@Named("developerMode") boolean developerMode)
 	{
 		super(false);
 
@@ -92,6 +98,11 @@ class TimeTrackingPanel extends PluginPanel
 		for (Tab tab : Tab.FARMING_TABS)
 		{
 			addTab(tab, farmingTracker.createTabPanel(tab, farmingContractManager));
+		}
+
+		if (developerMode)
+		{
+			addTab(Tab.TIME_OFFSET, new FarmingNextTickPanel(farmingTracker, config, configManager));
 		}
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
@@ -25,6 +25,7 @@
  */
 package net.runelite.client.plugins.timetracking;
 
+import com.google.inject.Inject;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
 import java.time.Instant;
@@ -32,20 +33,21 @@ import java.time.temporal.ChronoUnit;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
-import javax.inject.Inject;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.coords.WorldPoint;
-import net.runelite.client.events.ConfigChanged;
 import net.runelite.api.events.ChatMessage;
+import net.runelite.api.events.CommandExecuted;
 import net.runelite.api.events.GameTick;
+import net.runelite.api.events.WidgetClosed;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetInfo;
+import net.runelite.api.widgets.WidgetModalMode;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.events.ConfigChanged;
 import net.runelite.client.events.RuneScapeProfileChanged;
-import net.runelite.client.game.ItemManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import static net.runelite.client.plugins.timetracking.TimeTrackingConfig.CONFIG_GROUP;
@@ -89,16 +91,13 @@ public class TimeTrackingPlugin extends Plugin
 	private ClockManager clockManager;
 
 	@Inject
-	private ItemManager itemManager;
-
-	@Inject
-	private TimeTrackingConfig config;
-
-	@Inject
 	private InfoBoxManager infoBoxManager;
 
 	@Inject
 	private ScheduledExecutorService executorService;
+
+	@Inject
+	private ConfigManager configManager;
 
 	private ScheduledFuture panelUpdateFuture;
 
@@ -108,6 +107,8 @@ public class TimeTrackingPlugin extends Plugin
 
 	private WorldPoint lastTickLocation;
 	private boolean lastTickPostLogin;
+
+	private int lastModalCloseTick = 0;
 
 	@Provides
 	TimeTrackingConfig provideConfig(ConfigManager configManager)
@@ -125,7 +126,7 @@ public class TimeTrackingPlugin extends Plugin
 
 		final BufferedImage icon = ImageUtil.getResourceStreamFromClass(getClass(), "watch.png");
 
-		panel = new TimeTrackingPanel(itemManager, config, farmingTracker, birdHouseTracker, clockManager, farmingContractManager);
+		panel = injector.getInstance(TimeTrackingPanel.class);
 
 		navButton = NavigationButton.builder()
 			.tooltip("Time Tracking")
@@ -175,6 +176,16 @@ public class TimeTrackingPlugin extends Plugin
 	}
 
 	@Subscribe
+	public void onCommandExecuted(CommandExecuted commandExecuted)
+	{
+		if (commandExecuted.getCommand().equals("resetfarmtick"))
+		{
+			configManager.unsetRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION);
+			configManager.unsetRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET);
+		}
+	}
+
+	@Subscribe
 	public void onGameTick(GameTick t)
 	{
 		if (client.getGameState() != GameState.LOGGED_IN)
@@ -206,7 +217,7 @@ public class TimeTrackingPlugin extends Plugin
 		}
 
 		boolean birdHouseDataChanged = birdHouseTracker.updateData(loc);
-		boolean farmingDataChanged = farmingTracker.updateData(loc);
+		boolean farmingDataChanged = farmingTracker.updateData(loc, client.getTickCount() - lastModalCloseTick);
 		boolean farmingContractDataChanged = farmingContractManager.updateData(loc);
 
 		if (birdHouseDataChanged || farmingDataChanged || farmingContractDataChanged)
@@ -233,6 +244,15 @@ public class TimeTrackingPlugin extends Plugin
 		}
 
 		farmingContractManager.setContract(null);
+	}
+
+	@Subscribe
+	private void onWidgetClosed(WidgetClosed ev)
+	{
+		if (ev.getModalMode() != WidgetModalMode.NON_MODAL)
+		{
+			lastModalCloseTick = client.getTickCount();
+		}
 	}
 
 	@Schedule(period = 10, unit = ChronoUnit.SECONDS)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingNextTickPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingNextTickPanel.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2020 Morgan Lewis
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *     list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright notice,
+ *     this list of conditions and the following disclaimer in the documentation
+ *     and/or other materials provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ *  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ *  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ *  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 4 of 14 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -136,4 +140,23 @@ default Tab activeTab()
...
+	)
+	String advancedSection = "advanced";
+
+	@ConfigItem(
+		keyName = FARM_TICK_OFFSET_RESET,
+		name = "Reset Automatic Farming Tick Tracking",
+		description = "Reset automatic farming tick offset tracking.",
+		section = advancedSection
+	)
+	default boolean resetFarmTickOffset()
```
abextm, line 158: this would probably be better as a :: command. maybe `::resetfarmtick`?

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPanel.java
```diff
@@ -64,9 +68,8 @@
 	@Nullable
 	private TabContentPanel activeTabPanel = null;
 
-	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config,
-		FarmingTracker farmingTracker, BirdHouseTracker birdHouseTracker, ClockManager clockManager,
-		FarmingContractManager farmingContractManager)
+	@Inject
+	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config, FarmingTracker farmingTracker, BirdHouseTracker birdHouseTracker, ClockManager clockManager, FarmingContractManager farmingContractManager, ConfigManager configManager, @Named("developerMode") boolean developerMode)
```
abextm, line 72: newlines

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
```diff
@@ -172,6 +169,13 @@ else if (clockManager.getStopwatches().isEmpty() && e.getKey().equals(STOPWATCHE
 		{
 			clockManager.loadStopwatches();
 		}
+
+		if (e.getKey().equals(FARM_TICK_OFFSET_RESET) && e.getNewValue().equals("true"))
```
abextm, line 173: `newValue` can be null. In general prefer `"value".equals(thing)` to `thing.equals(value)`

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -100,13 +104,16 @@ public boolean updateData(WorldPoint location)
...
+		Collection<FarmingRegion> newRegions = farmingWorld.getRegionsForLocation(location);
+
+		if (!newRegions.equals(lastRegions))
 		{
-			if (!region.isInBounds(location))
-			{
-				continue;
-			}
+			newRegionLoaded = true;
+			log.debug("Farming: New region loaded. " + newRegions.toString() + " at " + client.getTickCount() + " ticks");
```
abextm, line 112: Use `{}` replacement rather than string concatenation.
```suggestion
			log.debug("Farming: New region loaded. {} at {} ticks", newRegions, client.getTickCount());
```

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,30 +128,81 @@ public boolean updateData(WorldPoint location)
...
+						else if (!newRegionLoaded)
 						{
-							continue;
+							PatchState previousPatchState = patch.getImplementation().forVarbitValue(Integer.parseInt(parts[0]));
+							PatchState currentPatchState = patch.getImplementation().forVarbitValue(client.getVar(varbit));
+							int patchTickRate = previousPatchState.getTickRate();
+
+							if (isObservedGrowthTick(previousPatchState, currentPatchState))
+							{
+								String storedOffsetPrecisionString = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION);
```
abextm, line 157:
```suggestion
								Integer storedOffsetPrecision = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION, Integer.class);
```

The config manager is capable of doing the conversion, you should use that since it handles corrupted configs.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,30 +128,81 @@ public boolean updateData(WorldPoint location)
...
+								}
+
+								if (storedOffsetString != null)
+								{
+									storedOffsetMins = Integer.parseInt(storedOffsetString);
+								}
+
+								int offsetMins = (int) Math.abs(((Instant.now().getEpochSecond() / 60) % patchTickRate) - patchTickRate);
+								String growthTickMessage = "Farming: Observed an exact growth tick. Offset is: " + offsetMins + " from a " + patchTickRate + " minute tick";
+								log.debug(growthTickMessage);
```
abextm, line 175: again use formatting and not concatenation

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,30 +128,81 @@ public boolean updateData(WorldPoint location)
...
+
+								if (storedOffsetMins != 0 && offsetMins != storedOffsetMins % patchTickRate)
+								{
+									WorldPoint playerLocation = client.getLocalPlayer().getWorldLocation();
+									String info = "Farming: Offset error! Observed new offset of " + offsetMins + ", previous observed offset was " + storedOffsetMins + " (" + storedOffsetMins % patchTickRate + ")" + " Player Loc:" + ((playerLocation != null) ? playerLocation.toString() : "");
+									log.error(info);
```
abextm, line 181: ditto

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,30 +128,81 @@ public boolean updateData(WorldPoint location)
...
+								}
+							}
 						}
 					}
 				}
-
 				String value = strVarbit + ":" + unixNow;
 				configManager.setRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, key, value);
 				changed = true;
+
```
abextm, line 198: remove this newline; readd the one above

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -153,6 +211,44 @@ public boolean updateData(WorldPoint location)
...
+	private boolean isObservedGrowthTick(PatchState previous, PatchState current)
+	{
+		//Check the previous state so it will still calculate during the final growth tick
+		int patchTickRate = previous.getTickRate();
+		CropState previousCropState = previous.getCropState();
+		CropState currentCropState = current.getCropState();
+		Produce previousProduce = previous.getProduce();
+
+		//Ignore weeds growing or being cleared.
+		if (previousProduce == Produce.WEEDS || current.getProduce() == Produce.WEEDS || patchTickRate <= 0)
```
abextm, line 223: probably also add a safety for the patch changing to a different produce

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -153,6 +211,44 @@ public boolean updateData(WorldPoint location)
...
+		{
+			return false;
+		}
+
+		if (previousCropState == CropState.GROWING)
+		{
+			if (currentCropState == CropState.GROWING || currentCropState == CropState.DISEASED)
+			{
+				log.debug("Farming: Found GROWING -> GROWING or GROWING -> DISEASED");
```
abextm, line 232: also maybe a safety that the state is +1 of the old one

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -244,6 +338,34 @@ public PatchPrediction predictPatch(FarmingPatch patch)
...
+	}
+
+	public long getTickTime(int tickRate, int ticks, long requestedtime)
+	{
+		String offsetPrecisionMins = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION);
+		String offsetTimeMins = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET);
+
+		//All offsets are negative but are stored as positive
+		long calculatedOffsetTime = 0L;
+		if (offsetPrecisionMins != null && (Integer.parseInt(offsetPrecisionMins) >= tickRate || Integer.parseInt(offsetPrecisionMins) >= 40))
```
abextm, line 353: unsafe parseInt. Use the ConfigManager to do the conversion.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -244,6 +338,34 @@ public PatchPrediction predictPatch(FarmingPatch patch)
 		);
 	}
 
+	public long getTickTime(int tickRate, int ticks)
+	{
+		return getTickTime(tickRate, ticks, Instant.now().getEpochSecond());
+	}
+
+	public long getTickTime(int tickRate, int ticks, long requestedtime)
```
abextm, line 346:
```suggestion
	public long getTickTime(int tickRate, int ticks, long requestedTime)
```

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingWorld.java
```diff
@@ -306,4 +342,9 @@ private void add(FarmingRegion r, int... extraRegions)
 				.add(p);
 		}
 	}
+
+	Collection<FarmingRegion> getRegionsForLocation(WorldPoint location)
+	{
+		return this.regions.get(location.getRegionID()).stream().filter(region -> region.isInBounds(location)).collect(Collectors.toSet());
```
abextm, line 348: newlines

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingWorld.java
```diff
@@ -292,7 +328,7 @@ public boolean isInBounds(WorldPoint loc)
 		this.tabs = Collections.unmodifiableMap(umtabs);
 	}
 
-	private void add(FarmingRegion r, int... extraRegions)
+	private void add(Multimap<Integer, FarmingRegion> regions, FarmingRegion r, int... extraRegions)
```
abextm, line 331: I dislike this change, especially since this is private.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/PatchImplementation.java
```diff
@@ -2787,4 +2787,6 @@ PatchState forVarbitValue(int value)
 	private final Tab tab;
 
 	private final String name;
+
+	private final boolean requiresHealthCheck;
```
abextm, line 2791:
```suggestion
	private final boolean healthCheckRequired;
```

or write your own getter idc

abextm (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingWorld.java
MESLewis: pmd was upset with FarmingWorld.regions not being final once the getter was removed. What is the suggested solution?

abextm: you can opt it out of pmd since its incorrect.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPlugin.java
```diff
@@ -174,6 +171,16 @@ else if (clockManager.getStopwatches().isEmpty() && e.getKey().equals(STOPWATCHE
 		}
 	}
 
+	@Subscribe
+	public void onCommandExecuted(CommandExecuted commandExecuted)
+	{
+		if (commandExecuted.getCommand().equals("resetfarmtick"))
```
Adam-, line 177: Should this be behind developer mode?

MESLewis: I think it is important for users to be able to reset their timing if anything goes wrong.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingConfig.java
```diff
@@ -33,6 +33,9 @@
 public interface TimeTrackingConfig extends Config
 {
 	String CONFIG_GROUP = "timetracking";
+	String FARM_TICK_OFFSET = "farm_tick_offset";
```
abextm, line 36: normally config keys are camel case

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,20 +128,52 @@ public boolean updateData(WorldPoint location)
...
+								Integer storedOffsetPrecision = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION, int.class);
+								Integer storedOffsetMins = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET, int.class);
+
+								int offsetMins = (int) Math.abs(((Instant.now().getEpochSecond() / 60) % patchTickRate) - patchTickRate);
+								log.debug("Farming: Observed an exact growth tick. Offset is: {} from a {} minute tick", offsetMins, patchTickRate);
+
+								if (storedOffsetMins != null && storedOffsetMins != 0 && offsetMins != storedOffsetMins % patchTickRate)
+								{
+									WorldPoint playerLocation = client.getLocalPlayer().getWorldLocation();
+									log.error("Farming: Offset error! Observed new offset of {}, previous observed offset was {} ({}) Player Loc:{}", offsetMins, storedOffsetMins, storedOffsetMins % patchTickRate, ((playerLocation != null) ? playerLocation.toString() : ""));
```
abextm, line 166:
```suggestion
									log.error("Farming: Offset error! Observed new offset of {}, previous observed offset was {} ({}) Player Loc:{}",
										offsetMins, storedOffsetMins, storedOffsetMins % patchTickRate, playerLocation));
```

playerLocation will never be null, and even if it could be this would be better

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -244,6 +326,34 @@ public PatchPrediction predictPatch(FarmingPatch patch)
...
+	}
+
+	public long getTickTime(int tickRate, int ticks, long requestedTime)
+	{
+		Integer offsetPrecisionMins = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET_PRECISION, int.class);
+		Integer offsetTimeMins = configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, TimeTrackingConfig.FARM_TICK_OFFSET, int.class);
+
+		//All offsets are negative but are stored as positive
+		long calculatedOffsetTime = 0L;
+		if (offsetPrecisionMins != null && (offsetPrecisionMins >= tickRate || offsetPrecisionMins >= 40))
```
abextm, line 341: should have a null check for `offsetTimeMins` too

abextm (changes requested):

abextm:
If it doesn't have a known offset it predicts early, then when it finds that is incorrect it will shift the prediction a tick later.

abextm (approved):

MESLewis:
Farming varbits don't get sent while the bank is open. Probably when other interfaces are open as well. This causes the plugin to incorrectly track the offset when closing the bank.

melkypie:
If you login after a growth tick has already happened, it will set the offset to the time when you login rather than when the actual growth tick had happened, resulting in the wrong offset. The offset is set after going past the click to play screen.

MESLewis:
> If you login after a growth tick has already happened, it will set the offset to the time when you login rather than when the actual growth tick had happened, resulting in the wrong offset. The offset is set after going past the click to play screen.

Fixed by clearing `lastRegions ` in `loadCompletionTime()` which is called every time accounts are changed or when logging out

MESLewis:
> Farming varbits don't get sent while the bank is open. Probably when other interfaces are open as well. This causes the plugin to incorrectly track the offset when closing the bank.

Fixed by using WidgetNode::modalMode

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/TimeTrackingPanel.java
```diff
@@ -64,9 +68,11 @@
 	@Nullable
 	private TabContentPanel activeTabPanel = null;
 
-	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config,
-		FarmingTracker farmingTracker, BirdHouseTracker birdHouseTracker, ClockManager clockManager,
-		FarmingContractManager farmingContractManager)
+	@Inject
+	TimeTrackingPanel(ItemManager itemManager, TimeTrackingConfig config, FarmingTracker farmingTracker,
+					BirdHouseTracker birdHouseTracker, ClockManager clockManager,
```
abextm, line 73: too many indents

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingNextTickPanel.java
```diff
@@ -0,0 +1,107 @@
...
+import net.runelite.client.plugins.timetracking.TimeTrackingConfig;
+import net.runelite.client.plugins.timetracking.TimeablePanel;
+import net.runelite.client.ui.ColorScheme;
+
+public class FarmingNextTickPanel extends TabContentPanel
+{
+	private final FarmingTracker farmingTracker;
+	private final TimeTrackingConfig config;
+	private final ConfigManager configManager;
+	private final List<TimeablePanel> patchPanels;
```
abextm, line 45: Add type parameters to all these raw `TimeablePanel`s
```suggestion
	private final List<TimeablePanel<Void>> patchPanels;
```

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -58,9 +63,11 @@
 	 */
 	private final Map<Tab, Long> completionTimes = new EnumMap<>(Tab.class);
 
+	private boolean newRegionLoaded;
+	private Collection<FarmingRegion> lastRegions = Collections.EMPTY_LIST;
```
abextm, line 67:
```suggestion
	private Collection<FarmingRegion> lastRegions = Collections.emptyList();
```
or I think this can be null

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
```diff
@@ -121,20 +138,56 @@ public boolean updateData(WorldPoint location)
...
+								continue;
+							}
 						}
-						catch (NumberFormatException e)
+						else if (!newRegionLoaded && timeSinceModalClose > 1)
 						{
-							// ignored
+							PatchState previousPatchState = patch.getImplementation().forVarbitValue(Integer.parseInt(parts[0]));
+							PatchState currentPatchState = patch.getImplementation().forVarbitValue(client.getVar(varbit));
+							int patchTickRate = previousPatchState.getTickRate();
```
abextm, line 163: (previous|current)PatchState should have null checks

PR #12950, opened 2020-12-19
When shift clicking on skills in the skills tab it pauses them?

CnPx98:
Can this be added in the next update? It would be a useful feature saving right clicking and then selecting pause.


Hydrox6:
You want to create an issue instead; Pull Requests are for code contributions.

PR #12951, opened 2020-12-19
Extend Clue Scroll tag tab functionality to be generic + support for custom tag tab sub-sections

Zoinkwiz:
Previously, clue scrolls had functionality added to the Bank Tags plugin to allow for the insertion of current clue specific items, such as spades or equipment for emote clues.

The core of this change is to extend that functionality to be easily usable by other plugins should they want to. Notably, this would be used by the Quest Helper plugin to provide information on what items to bring for a quest.

In addition, this includes three other features:

- When a player doesn't have an item nor a placeholder for an item a plugin has stated should exist in a bank, it now has the option to insert a fake placeholder item, which a player can check 'Details' to get more  information. This currently is only the quantity + name of the item (or the collective name for it, such as 'Any boots'), but this could be easily extended to provide more useful information
- There is now the option to organise Bank Tabs into sections, much like the main bank tab has lines dividing up the different official tabs. These can be labeled, and a plugin can choose what items should fall under each tab.
- Now if you search a tab's name in the GE which has plugin-inserted items, they will appear in the search.

In tabs with plugin-inserted sections, any items tagged to match the tab's tag will be added to the 'General' section at the bottom.

clue2

There's a couple things worth noting, and thoughts for things beyond this:

- If an item falls into multiple sections of a tab, it will only appear as the actual item in the first occurrence, and afterwards appear as fake placeholders. I imagine it'd be a better experience for the item to have a clickable widget in each section for taking it out, but equally I wasn't certain 1. How to, 2. If it'd be problematic from a rules perspective. This can work either way, but insight would be great 😄.
- The GE search could be improved quite a bit. I've yet to look particularly into it, but I could imagine a button existing to search for a plugin-inserted tab in the GE would be handy as I think most people don't even know that the functionality exists. This would be assuming most plugin-inserted tabs are circumstantial (clue only appears when a clue is open, quest helper only when a quest helper is explicitly open, etc, to avoid noise).
- Extending on the above, having tabs appear when relevant, then disappear when not relevant, could be useful for visibility. I would imagine this working as, assuming someone doesn't have a 'clue' tab already, a tab being added to the top of the tabs list with some indication of being separate from the rest. Once you finish the clue it'd disappear, assuming the player hasn't made it a tab themselves. If they had it'd convert to a normal tab with the existing insertion.


Zoinkwiz:
Made an implementation of this also with the Quest Helper, seems to work well for it, certainly enough detail for now.

questhelpertags

Adam-:
I can't review all of this at once, can you reduce it to only the letting other plugins define bank tags part?

Adam-:
What about this? https://github.com/runelite/runelite/compare/master...Adam-:clue-tag2

Zoinkwiz:
That looks good for that part of the change, would you want me to close this PR?

Adam-:
I've pushed this now. You should rebase this on it since it looks like this contains a number of other things?

PR #12953, opened 2020-12-19
WIP: Add Inventory Value Tracker 

wrightmalone:
Original plugin by @wikiworm: https://github.com/wikiworm/InventoryValue

Inventory Value 

Displays the current inventory value in an overlay.

Configuration options:
ignoreCoins -- coins will no longer be shown in total value
useHaValue -- use the high alchemy value of an item instead of the GE price
ignoreItems -- list of item names, case-insensitive, delimited by either comma or semicolon

image

Allows a user to see value fluctuation in a farm run, supplies used vs drops gained, and many other scenarios.

image

With the addition of the ignoreItems configuration, items in your inventory used for a kill which are not consumable can be ignored, allowing an easier tally of the overall gp gain from a trip.
image

image


wrightmalone:
Plugin was determined too niche to be in the main client. Made a PR against wikiworm's pluginhub plugin instead.

PR #12956, opened 2020-12-21
Timers Plugin: Update Sire Timer to account for stacking stuns

Mrnice98:
Update Sire Timer to account for stacking stuns, change method of detection to chat message. should make it more robust so it should not break if jagex makes changes in the future.


Mrnice98:
rebased to upstream head

Issue #12787, opened 2020-11-12
"Scorched" Lletya Fruit Tree Patch Not Recognized in Time Tracking plugin


llamositopia:
**Describe the bug**
I am far enough along in Song of the Elves that Lletya has been attacked, and is "scorched". Without completing the quest, I have planted a palm tree in the fruit tree patch there, but it is not timed in the Time Tracking plugin. I am not sure yet whether Lletya actually returns back to normal after the quest, but maybe someone can fill me in on that.

**To Reproduce**
1. Advance in Song of the Elves until Lletya is attacked. Do not continue.
2. Plant a fruit tree in the fruit tree patch just west of the Lletya Bank.
3. Observe that Time Tracking does not recognize anything planted in that patch.

**Expected behavior**
The tree should be recognized and timed within the plugin interface.

**Screenshots**
![image](https://user-images.githubusercontent.com/1868974/98991303-aa723480-24f9-11eb-842b-1f40d7b306da.png)
![image](https://user-images.githubusercontent.com/1868974/98991330-b2ca6f80-24f9-11eb-827d-85ed8ac883e8.png)

**Environment (please complete the following information):**
 - OS: Windows 10, Version 2004
 - RuneLite version: 1.6.31
 - Launcher version: 2.1.3

**Additional context**
I'm not sure whether the patches are ground objects or game objects, but the ground object of the patch is 7517, and the game object looks to be 26579.
![image](https://user-images.githubusercontent.com/1868974/98992082-b9a5b200-24fa-11eb-8954-9424540a3e3e.png)
![image](https://user-images.githubusercontent.com/1868974/98992100-be6a6600-24fa-11eb-83a2-9abbbd838d95.png)

Tile location is (2794, 6105, 0) to (2795, 6106, 0).

llamositopia:
I've just learned that when passing through the tree entrance to Lletya, 
![image](https://user-images.githubusercontent.com/1868974/98995760-d42e5a00-24ff-11eb-8d50-932fe5b5a05b.png)
your character is warped from region 9265 on the left side, to region 11103 on the right side. Presumably this creates the distinction between scorched and unscorched Lletya. 

The existing code (snippet below) references Lletya by region 9265.
https://github.com/runelite/runelite/blob/b520e58034f034da81bc6b40d70a9847752403dd/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingWorld.java#L194-L196

[...]

PR #12957, opened 2020-12-21
Add "scorched" Lletya regionID recognition to TimeTracking and Discord plugins

llamositopia:
Fixes https://github.com/runelite/runelite/issues/12787. 

From the issue:
> I am far enough along in Song of the Elves that Lletya has been attacked, and is "scorched". Without completing the quest, I have planted a palm tree in the fruit tree patch there, but it is not timed in the Time Tracking plugin.
> [...] when passing through the tree entrance to Lletya, your character is warped from region 9265 on the left side, to region 11103 on the right side.

This patch adds recognition for region 11103, so that players who are partially through Song of the Elves still have their fruit tree timers accurately tracked.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
@@ -121,7 +121,7 @@ enum DiscordGameEventType
 	CITY_JATIZSO("Jatizso" , DiscordAreaType.CITIES, 9531),
 	CITY_KELDAGRIM("Keldagrim" , DiscordAreaType.CITIES, 11423, 11422, 11679, 11678),
 	CITY_LANDS_END("Land's End", DiscordAreaType.CITIES, 5941),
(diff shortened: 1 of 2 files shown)


makingcomment:
I bet this region id is also missing from the discord plugin.

llamositopia:
> I bet this region id is also missing from the discord plugin.

Good catch. Just added a new commit to add it there as well.

MESLewis (approved):

abextm:
Is this region an instance?

llamositopia:
I'm not sure, actually. I'm no longer at this point of SotE, so I wouldn't be able to check on my account.

I don't think it would be, it replaces the entirety of Lletya when you are partially through SotE, for all visits, not just a fight. It remains that way until you complete the quest.

Is the handling different for instances in either the farming tracking or the discord region tracking?

abextm:
The farming tracker doesn't account for instances, so this wouldn't work if it is one.

llamositopia:
Before submitting this, I did test that adding the region worked to recognize the patch when I was there, and it did work. So presumably this isn't an instanced area then.

PR #12958, opened 2020-12-21
--debug Logback Classic Logger instance check

llamositopia:
Spoke with @abextm about this on Discord...

When using the example-plugin base project, the build.gradle imports slf4j-simple, and excludes logback-classic. However, this (slightly misleadingly) causes a crash if you're using the --debug flag, since the runtime log level setting is a logback-classic specific feature (it's not an action you can do with just slf4j-api).

This PR adds a check to make sure that the root logger being modified is actually a logback-classic logger, by checking which SLF4J implementation has been bound. If --debug was set, but we don't have a logback-classic logger instance, it prints out a warning to the developer to set the root logger level using another method.

I used fully-qualified names for the Logger classes to avoid any confusion about which type we have at any given time.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/RuneLite.java
+++ b/runelite-client/src/main/java/net/runelite/client/RuneLite.java
@@ -25,7 +25,6 @@
 package net.runelite.client;
 
 import ch.qos.logback.classic.Level;
-import ch.qos.logback.classic.Logger;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Guice;
 import com.google.inject.Inject;
@@ -86,6 +85,7 @@ import net.runelite.http.api.RuneLiteAPI;
 import okhttp3.Cache;
 import okhttp3.OkHttpClient;
 import org.slf4j.LoggerFactory;
+import org.slf4j.impl.StaticLoggerBinder;
 
 @Singleton
 @Slf4j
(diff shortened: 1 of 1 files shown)


llamositopia:
Amended commit to satisfy checkstyle errors

runelite-client/src/main/java/net/runelite/client/RuneLite.java
```diff
@@ -223,8 +223,22 @@ public ClientUpdateCheckMode convert(String v)
 
 		if (options.has("debug"))
 		{
-			final Logger logger = (Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
-			logger.setLevel(Level.DEBUG);
+			// during plugin-hub dev, not all devs use logback-classic
+			// make sure we have logback-classic before attempting to set level
+			// setting level at runtime is not supported by the slf4j-api alone
+			final org.slf4j.Logger rootLogger = LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
+			if (StaticLoggerBinder.getSingleton().getLoggerFactoryClassStr().equals("ch.qos.logback.classic.util.ContextSelectorStaticBinder"))
```
Adam-, line 230: Can't this just be an `instanceof Logger` check?

llamositopia: > Can't this just be an `instanceof Logger` check?

Unfortunately not. I anticipated that an instanceof check should work, since the instanceof operator does not actually cause the ClassLoader to load in the right-hand operand class, but apparently it does still require that the ClassLoader can *find* the class. Therefore, when I tested it with an instanceof check, it still threw the ClassNotFoundEx.

For context on how I determined that the information was in StaticLoggerBinder, that is the same class SLF4J uses to determine which binding is in the classpath. This is also the same class that will throw warnings if there are multiple bindings on the classpath. It provides `getLoggerFactoryClass()` to get the `Class<?>` instance, or `getLoggerFactoryClassStr()` which is exclusively the class name.

Adam-:
I think I would rather just have the pluginhub plugins use logback. abex says that should work.

llamositopia:
Resolved in runelite/plugin-hub#881

PR #12971, opened 2020-12-23
discord: add main menu status option

### LINKED ISSUE #12940: Discord: Option to not show game status when on login screen
Keep having misunderstandings where people think I'm playing when I actually just left the client open in the background not logged in, would love if it removed runelite from my status on the login screen.

dillydill123:
Closes #12940

When the client starts up, there is still around ~15 seconds where it will show the status as "In Menu". Afterwards it will go away. I'm not sure how to solve this as it doesn't seem to be related to the startup of the discord plugin. Turning the plugin on and off does not reproduce this behavior. It is only on client startup.


dillydill123:
I was using the wrong game state. It is correctly not showing up on client startup now.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordPlugin.java
```diff
@@ -372,6 +372,12 @@ private void resetState()
 
 	private void checkForGameStateUpdate()
 	{
+		if (!config.showMainMenu() && client.getGameState() != GameState.LOGGED_IN)
```
Adam-, line 375: Just call `client.getGameState()` once and then use it here + also down below.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordConfig.java
```diff
@@ -72,11 +72,22 @@ default int actionTimeout()
 		return 5;
 	}
 
+	@ConfigItem(
+			keyName = "showMainMenu",
```
Adam-, line 76: Fix the indenting here.

dillydill123:
I changed the logic because there was a duplicate call to `resetState()`. Everywhere  `checkForGameStateUpdate()` is called, `resetState()` is called right before it.

PR #12973, opened 2020-12-24
Java 14-16 compatability

### LINKED ISSUE #10113: Using Gson to marshal many java.* classes is platform specific
Gson reflects into private fields of types it does not have [an explicit understanding](https://github.com/google/gson/blob/f649e051411e092f0123878e16c5132500a2d01e/gson/src/main/java/com/google/gson/Gson.java#L231-L276) of, which means it depends on the vendor/version of the JVM how/if the type is serialized. This will also fail in java >= 16 due to illegal access being denied due to JEP 396.

The solution to this would be to create TypeAdapters for any types Gson does not understand already, or not use the type.

Some (__not all__) cases of this:
 - `java.util.UUID` used all over in the session stuff
 - `java.awt.Color` used in screen markers (this is painful because its in the config and has an *interesting* format)

abextm:
Includes workarounds for:
 - [JDK-8173978](https://bugs.openjdk.java.net/browse/JDK-8173978) (Java 14)
 - [JEP396/JDK-8255363](https://bugs.openjdk.java.net/browse/JDK-8255363) (Java 16)

Closes #10113

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/gson/ColorTypeAdapter.java
+++ b/http-api/src/main/java/net/runelite/http/api/gson/ColorTypeAdapter.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2020 Abex
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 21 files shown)


http-api/src/main/java/net/runelite/http/api/gson/ColorTypeAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+	{
+		switch (in.peek())
+		{
+			case NULL:
+				in.nextNull();
+				return null;
+			case BEGIN_OBJECT:
+				in.beginObject();
+				double value = 0;
+				for (; in.peek() != JsonToken.END_OBJECT; )
```
Nightfirecat, line 57:
```suggestion
				while (in.peek() != JsonToken.END_OBJECT)
```
for all the type adapter loops

http-api/src/main/java/net/runelite/http/api/gson/InstantTypeAdapter.java
```diff
@@ -0,0 +1,82 @@
...
+package net.runelite.http.api.gson;
+
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.time.Instant;
+
+// Just add water!
```
Nightfirecat, line 34: :joy:

http-api/src/main/java/net/runelite/http/api/gson/ColorTypeAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.gson.stream.JsonWriter;
+import java.awt.Color;
+import java.io.IOException;
+
+public class ColorTypeAdapter extends TypeAdapter<Color>
+{
+	@Override
+	public void write(JsonWriter out, Color value) throws IOException
+	{
+		int rgba = value.getRGB();
```
Nightfirecat, line 39: Do we not need to handle null color values for this?

Nightfirecat (commented):
I can't speak to the ReflectUtil change, but e9b3c16 looks good to me

Adam- (commented):
The only gson instance left being created I can see is in the skill calc, did you intentionally keep that? I don't think it requires using any of the new serializers though so maybe it's ok.

PR #12974, opened 2020-12-24
worldmap: change from skill level to boosted level for map icons

### LINKED ISSUE #12790: world map: Take boosted levels into account when showing if you can chop a tree/use a shortcut

akhtar-u:
Close #12790

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
@@ -195,7 +195,7 @@ public class WorldMapPlugin extends Plugin
 		{
 			case AGILITY:
 			{
-				int newAgilityLevel = statChanged.getLevel();
+				int newAgilityLevel = statChanged.getBoostedLevel();
 				if (newAgilityLevel != agilityLevel)
 				{
 					agilityLevel = newAgilityLevel;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
```diff
@@ -195,7 +195,7 @@ public void onStatChanged(StatChanged statChanged)
 		{
 			case AGILITY:
 			{
-				int newAgilityLevel = statChanged.getLevel();
+				int newAgilityLevel = statChanged.getBoostedLevel();
```
Nightfirecat, line 198: While most shortcuts reference your boosted level, not all do. The Crabclaw Caves shortcuts (used in [The Depths of Despair quest](https://oldschool.runescape.wiki/w/The_Depths_of_Despair)) require 18 agility to cross, but you cannot boost your agility level to start the quest therefore must have the base level to cross them.

That said this might be too much of an edge case to be worth considering, unless other similar shortcuts exist.

akhtar-u: Ya I am not sure which shortcuts are unboostable or not. Will probably need to come back to this in the future.

Adam-: Based on my experience there's more shortcuts that use boosted level than not, so it's probably better to have it just use the boosted level.

PR #12977, opened 2020-12-26
Corp: Add dark core notification sound when spawned

EliasLahham:


Broooklyn:
I believe this goes against Jagex's statement on third party clients, specifically ```Adds additional visual or audio indicators of a boss mechanic.```

https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1

EliasLahham:
> 
> 
> I believe this goes against Jagex's statement on third party clients, specifically `Adds additional visual or audio indicators of a boss mechanic.`
> 
> https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1

I see where you're coming from, but if that's the case, why is the 
image allowed then? It adds a visual red square over the dark core. Not to mention the many other audio notifiers in existing plugins

Adam-:
Doesn't the core despawn if you teleport out?

EliasLahham:
> 
> 
> Doesn't the core despawn if you teleport out?

Not from what I've experienced. I've been doing 170~ kills on my ironman and when I spawn the core during my specing period it will still be there when I teleport out and back into the cave.

Adam-:
Okay, this seems fine then. Overall I think this is okay since we know the NPC highlight is okay for the core, I don't see why a spawn notification wouldn't either.

Nightfirecat:
Does the core die if you kill the corporeal beast without killing it?

EliasLahham:
I'm pretty sure it disappears as corp dies without dropping ashes

Nightfirecat:
Then wouldn't that skip the `isdead()` check in that despawn branch, and the plugin wouldn't send a notification the next time the core spawned?

Adam-:
> 
> 
> Then wouldn't that skip the `isdead()` check in that despawn branch, and the plugin wouldn't send a notification the next time the core spawned?

You could just reset it when the corp dies too I think.

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpPlugin.java
```diff
@@ -140,6 +150,22 @@ public void onNpcSpawned(NpcSpawned npcSpawned)
...
+				if (config.notifyDarkCore() && !coreSpawned)
+				{
+					notifier.notify("Dark core has spawned!");
+
+					String message = new ChatMessageBuilder()
+							.append(ChatColorType.HIGHLIGHT)
+							.append("Dark core has spawned!")
+							.build();
+
+					chatMessageManager.queue(QueuedMessage.builder()
```
Adam-, line 162: The notifier already has support for sending notifs to the chatbox, so this isn't necessary.

PR #12980, opened 2020-12-26
fix mining amounts when using varrock armour

### LINKED ISSUE #12910: Mining Counter doesn't account for Varrock Platebody
Mining plugin uses xptracker plugin to keep a count of actions, which doesn't equal ores when the varrock platebody special effect doubles the ore mined. 

I suspect the same is true for the smelting effect of the platebody, but haven't tested that. 

Could keep a second counter based on chat messages: "The Varrock platebody enabled you to mine an additional ore.". There would then need to be some maths to fix the Mined/hr amount, which could be done by getting the session time and adding the platebody additions per hour to the xptracker amount.

akhtar-u:
Close #12910 

I'll try to fix the smelting plugin counter and ring of forging charges as well sometime.


Nightfirecat:
What are the actual chat messages received from the Varrock armour effect? We should add tests and possibly refactor `MINING_PATERN` to handle them

PR #12983, opened 2020-12-27
chatfilter: add filter chat option to right click menu

### LINKED ISSUE #12845: Right click chat filter
Would it be possible to have the option to add a chat filter option to right clicking or shift+right clicking on a player. My ignore list is full and all these recent bots or whatever spamming that they're buying items or quitting have random names.

akhtar-u:
Close #12845 

Could also add right click option from chat box in the future.


runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
```diff
@@ -383,4 +395,63 @@ boolean shouldFilterByName(final String playerName)
...
+		String filteredNames = configManager.getConfiguration("chatfilter", "filteredNames");
+
+
+		if (click.getMenuOption().equals(FILTER_CHAT))
+		{
+			config.setNamesToFilter(filteredNames.isEmpty() ? playerName : filteredNames + "\n" + playerName);
+		}
+		else
+		{
+			filteredNames = filteredNames.replaceAll(playerName, "");
```
Nightfirecat, line 451: A newline splitter is defined at the top of this plugin class, use that to fetch filtered names  and filter entries exactly matching the clicked player's name instead of using `replaceAll`. This would, for instance, break other name filters if you tried to unfilter the name of a player with a one-letter name.

Nightfirecat (changes requested):

Adam-:
The entity hider lets you hide ignored players now, so it seems like that might be a better solution than this?

akhtar-u:
> 
> 
> The entity hider lets you hide ignored players now, so it seems like that might be a better solution than this?

But you still have to type their name no? This is more convenient imo.

Nightfirecat:
If a player is chatting, you can right-click their chat message and ignore them from that menu.

akhtar-u:
> 
> 
> If a player is chatting, you can right-click their chat message and ignore them from that menu.

Only works if your ignore list isn't full. I guess we could check if ignore list is full and add to the chat filter list using the same right click option.

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
```diff
@@ -383,4 +396,65 @@ boolean shouldFilterByName(final String playerName)
...
+
+		int type = event.getType();
+
+		if (type >= MENU_ACTION_DEPRIORITIZE_OFFSET)
+		{
+			type -= MENU_ACTION_DEPRIORITIZE_OFFSET;
+		}
+		final MenuAction menuAction = MenuAction.of(type);
+
+		if (menuAction.equals(RUNELITE_PLAYER))
```
Nightfirecat, line 416: This will not work when no other plugin uses MenuManager to add player menu actions. (Disable the hiscore plugin to see how this doesn't work)

This plugin should also make use of [`MenuManager#addPlayerMenuItem(String menuText)`](http://static.runelite.net/api/runelite-client/net/runelite/client/menus/MenuManager.html#addPlayerMenuItem(java.lang.String)) instead of adding this entry manually.

Nightfirecat (changes requested):

PR #12990, opened 2020-12-29
opponentinfo: make health bar more active

chaticon:
Made the health bar continue to display while the opponent attacks the player.  Useful for keeping track of enemy hp while recoiling.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
@@ -25,6 +25,7 @@
  */
 package net.runelite.client.plugins.opponentinfo;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Provides;
 import java.time.Duration;
 import java.time.Instant;
@@ -80,6 +81,8 @@ public class OpponentInfoPlugin extends Plugin
 	@Getter(AccessLevel.PACKAGE)
 	private Actor lastOpponent;
 
+	@Getter(AccessLevel.PACKAGE)
+	@VisibleForTesting
 	private Instant lastTime;
 
 	@Provides
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2016-2018, Adam <Adam@sigterm.info>
  * Copyright (c) 2018, Jordan Atwood <jordan.atwood423@gmail.com>
+ * Copyright (c) 2021, Andre Araya <araya.andre7@gmail.com>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -137,33 +138,50 @@ public class OpponentInfoPlugin extends Plugin
 	@Subscribe
 	public void onInteractingChanged(InteractingChanged event)
 	{
-		if (event.getSource() != client.getLocalPlayer())
+		final Actor player = client.getLocalPlayer();
+		if (player == null)
 		{
 			return;
 		}
+		final Actor opponent = player.getInteracting();
+		final Actor source = event.getSource();
+		final Actor target = event.getTarget();
 
-		Actor opponent = event.getTarget();
-
-		if (opponent == null)
+		if (source == player)
+		{
+			// You have attacked an enemy
+			if (target != null)
+			{
+				lastOpponent = target;
+				lastTime = null;
+			}
+			// You have stopped attacking an enemy which is not attacking you
+			else if (lastOpponent != null && lastOpponent.getInteracting() != player)
+			{
+				lastTime = Instant.now();
+			}
+		}
+		// You are attacked while not attacking anything
+		else if (target == player && opponent == null)
+		{
+			lastOpponent = source;
+			lastTime = null;
+		}
+		// An enemy which was previously attacking you (which you were not attacking back) has changed its target
+		else if (source == lastOpponent && opponent != lastOpponent)
 		{
 			lastTime = Instant.now();
-			return;
 		}
-
-		lastOpponent = opponent;
 	}
 
 	@Subscribe
 	public void onGameTick(GameTick gameTick)
 	{
-		if (lastOpponent != null
-			&& lastTime != null
-			&& client.getLocalPlayer().getInteracting() == null)
+		if (lastTime != null
+			&& Duration.between(lastTime, Instant.now()).compareTo(WAIT) > 0)
 		{
-			if (Duration.between(lastTime, Instant.now()).compareTo(WAIT) > 0)
-			{
-				lastOpponent = null;
-			}
+			lastOpponent = null;
+			lastTime = null;
 		}
 	}
 
--- a/runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2021, Jordan Atwood <jordan.atwood423@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.opponentinfo;
+
+import com.google.inject.Guice;
+import com.google.inject.Inject;
+import com.google.inject.testing.fieldbinder.Bind;
+import com.google.inject.testing.fieldbinder.BoundFieldModule;
+import net.runelite.api.Actor;
+import net.runelite.api.Client;
+import net.runelite.api.events.InteractingChanged;
+import net.runelite.api.NPC;
+import net.runelite.api.Player;
+import net.runelite.client.ui.overlay.OverlayManager;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import org.junit.Before;
+import org.junit.runner.RunWith;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.Mock;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OpponentInfoPluginTest
+{
+	@Mock
+	@Bind
+	Client client;
+
+	@Mock
+	@Bind
+	OverlayManager overlayManager;
+
+	@Mock
+	@Bind
+	OpponentInfoConfig config;
+
+	@Mock
+	@Bind
+	OpponentInfoOverlay opponentInfoOverlay;
+
+	@Mock
+	@Bind
+	PlayerComparisonOverlay playerComparisonOverlay;
+
+	@Inject
+	OpponentInfoPlugin plugin;
+
+	private final Player localPlayer = mock(Player.class);
+
+	@Before
+	public void before()
+	{
+		Guice.createInjector(BoundFieldModule.of(this)).injectMembers(this);
+
+		when(client.getLocalPlayer()).thenReturn(localPlayer);
+	}
+
+	@Test
+	public void testNpcInteractions()
+	{
+		final NPC npc = mock(NPC.class);
+
+		interactingChanged(npc, localPlayer);
+
+		assertNull(plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(npc, null);
+
+		assertNull(plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(localPlayer, npc);
+
+		assertSame(npc, plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(npc, localPlayer);
+
+		assertSame(npc, plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(localPlayer, null);
+
+		// last opponent is remembered for 5 seconds
+		assertSame(npc, plugin.getLastOpponent());
+		assertNotNull(plugin.getLastTime());
+	}
+
+	@Test
+	public void testOtherPlayerInteractions()
+	{
+		final Player otherPlayer = mock(Player.class);
+
+		interactingChanged(otherPlayer, localPlayer);
+
+		assertNull(plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(otherPlayer, null);
+
+		assertNull(plugin.getLastOpponent());
+		assertNull(plugin.getLastTime());
+
+		interactingChanged(localPlayer, otherPlayer);
+
+		assertSame(otherPlayer, plugin.getLastOpponent());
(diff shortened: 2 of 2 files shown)


Adam-:
This looks like it would break whenever you are fighting something in a multi combat area and then had something else attack you.

chaticon:
> This looks like it would break whenever you are fighting something in a multi combat area and then had something else attack you.

My bad, I had tested it in multi but clearly not well enough.  Should correctly prioritise the NPC you're attacking now.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -134,42 +134,55 @@ else if (worldType.contains(WorldType.LEAGUE))
...
-			&& client.getLocalPlayer().getInteracting() == null)
+		if (lastTime != null 
+			&& Duration.between(lastTime, Instant.now()).compareTo(WAIT) > 0)
 		{
-			if (Duration.between(lastTime, Instant.now()).compareTo(WAIT) > 0)
-			{
-				lastOpponent = null;
-			}
+			lastOpponent = null;
+			lastTime = null;
```
Nightfirecat, line 176: Why is this nulled here now?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -134,42 +134,55 @@ else if (worldType.contains(WorldType.LEAGUE))
...
+		else if (target == player
+				|| source == lastOpponent)
+		{
+			Actor opponent = player.getInteracting();
 
-		lastOpponent = opponent;
+			if (target != null && opponent == null)
+			{
+				lastOpponent = source;
+				lastTime = null;
```
Nightfirecat, line 160: Why is this nulled here now?

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
chaticon: Hello, so, in the current version of the plugin, it checks whether the player is interacting with an NPC on line 158 as part of determining whether to hide the health bar after WAIT time has passed or not.  Since my goal is to have the health bar visible while the player isn't technically interacting with the NPC but is still damaging it with recoil/cannon, I instead made lastTime null when it hasn't been set to a value in onInteractingChanged.  Otherwise, it would hide the health bar every tick after the initial WAIT time has passed.

Nightfirecat:
The behavior this has now seems very odd. If you are in a multi-combat area and are being attacked by multiple NPCs, it will prioritize whichever NPC targeted you last, as opposed to whichever attacked you first? Also, if a monster attacking you changed targets to another player--for instance, a hellhound attacking you became aggressive to a player running through your area--it would start tracking that no-longer-attacking-you hellhound as the active opponent?

This change very much would benefit from some tests being added, given its complexity.

chaticon:
> The behavior this has now seems very odd. If you are in a multi-combat area and are being attacked by multiple NPCs, it will prioritize whichever NPC targeted you last, as opposed to whichever attacked you first? Also, if a monster attacking you changed targets to another player--for instance, a hellhound attacking you became aggressive to a player running through your area--it would start tracking that no-longer-attacking-you hellhound as the active opponent?
> 
> This change very much would benefit from some tests being added, given its complexity.

I added some tests to clarify behaviour in some ambiguous scenarios.

To address the specific scenarios you brought up, you were partially right that I goofed the second scenario.  It wouldn't start tracking the hellhound if it wasn't already active opponent, but if it was, it would stay the active opponent until that NPC died or some other NPC became active opponent.  I've fixed that now.

As for the first scenario you mentioned, I don't think it's as weird as you say.  If the player is attacking an NPC, then that NPC will always be the active opponent regardless of what happens in multi.  The only time the most recent attacker will become the active opponent is if the player is not attacking any NPC at all.  That might sound odd, but if you're, say, cannoning while idle, that NPC is probably the most recent enemy your cannon hit, in which case it's behaving similarly to the current version of the plugin.

Regardless, I hope the tests are to your liking and can help illustrate my explanation.  Cheers.

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,199 @@
...
+    {
+        Guice.createInjector(BoundFieldModule.of(this)).injectMembers(this);
+    }
+
+    /*
+     * Verify that the current opponent in multi is always the npc that the player is attacking
+     */
+    @Test
+    public void testAttackingEnemyInMulti() {
+        Player thePlayer = mock(Player.class), otherPlayer = mock(Player.class);
```
Nightfirecat, line 56: We generally use the name `localPlayer` for the local player
```suggestion
        Player localPlayer = mock(Player.class), otherPlayer = mock(Player.class);
```

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,199 @@
...
+        opponentInfoPlugin.onInteractingChanged(aggroEnemy);
+
+        // verify that the current opponent is the aggressive npc
+        assertSame(opponentInfoPlugin.getLastOpponent(), aggro);
+
+        // the player attacks a different npc
+        when(thePlayer.getInteracting()).thenReturn(attacked);
+        opponentInfoPlugin.onInteractingChanged(new InteractingChanged(thePlayer, attacked));
+        // the npc attacks them back
+        opponentInfoPlugin.onInteractingChanged(new InteractingChanged(attacked, thePlayer));
```
Nightfirecat, line 76: Shouldn't there be another assert between these verifying `attacked` is now the last opponent?

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -134,33 +137,46 @@ else if (worldType.contains(WorldType.LEAGUE))
...
-			return;
+			if (target != null)
+			{
+				lastOpponent = target;
+				lastTime = null;
+			}
+			else if (lastOpponent != null
+				&& lastOpponent.getInteracting() != player)
+			{
+				lastTime = Instant.now();
```
Nightfirecat, line 152: It looks like this line is not covered in the added tests. What case does this cover?

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,199 @@
...
+        // verify that there is no longer an opponent
+        // (if lastTime is not null, then the lastOpponent will become null after OpponentInfoPlugin.WAIT time)
+        assertTrue(opponentInfoPlugin.getLastTime() != null || opponentInfoPlugin.getLastOpponent() == null);
+
+        // the player attacks the npc
+        when(thePlayer.getInteracting()).thenReturn(playerNPC);
+        opponentInfoPlugin.onInteractingChanged(new InteractingChanged(thePlayer, playerNPC));
+        // the npc attacks them back
+        when(playerNPC.getInteracting()).thenReturn(thePlayer);
+        opponentInfoPlugin.onInteractingChanged(new InteractingChanged(playerNPC, thePlayer));
```
Nightfirecat, line 172: Same here, shouldn't there be another assert between these verifying `playerNPC` is now the last opponent?

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,199 @@
...
+        assertTrue(opponentInfoPlugin.getLastTime() != null || opponentInfoPlugin.getLastOpponent() == null);
+    }
+
+    /*
+     * Verify that the current opponent is the expected npc in singles
+     */
+    @Test
+    public void testSingles() {
+        Player thePlayer = mock(Player.class), otherPlayer = mock(Player.class);
+        NPC playerNPC = mock(NPC.class);
```
Nightfirecat, line 146: This is a very confusing var name. Couldn't we just call this `npc`?

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        // verify that there is currently no opponent
+        assertNull(opponentInfoPlugin.getLastOpponent());
+
+        // some npc attacks the player
+        InteractingChanged aggroEnemy = new InteractingChanged(aggro, thePlayer);
+        opponentInfoPlugin.onInteractingChanged(aggroEnemy);
+
+        // verify that the current opponent is the aggressive npc
+        assertSame(opponentInfoPlugin.getLastOpponent(), aggro);
```
Nightfirecat, line 70: FYI, the argument order for assertions is `assertSame(expected, actual)`, so every assertion has the parameters in reverse order.
```suggestion
        assertSame(opponentInfoPlugin.getLastOpponent(), aggro);
```

Nightfirecat (changes requested):
While reviewing this, I've thought of another case which causes this code to fail, and I've written a brief test case which demonstrates it:

```java
@Test
public void singleCombatIssue()
{
    Player localPlayer = mock(Player.class), two = mock(Player.class), three = mock(Player.class);
    when(client.getLocalPlayer()).thenReturn(localPlayer);
    // TODO: find some unique var/condition representing a player being in multi-combat and mock it to be false here

    opponentInfoPlugin.onInteractingChanged(new InteractingChanged(localPlayer, two));
    assertSame(two, opponentInfoPlugin.getLastOpponent());

    // Another player tries to attack or cast magic on the player which causes interaction target changed from null -> player -> null in succession
    opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, null));
    assertSame(two, opponentInfoPlugin.getLastOpponent());

    opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, localPlayer));
    assertSame(two, opponentInfoPlugin.getLastOpponent()); // this line fails

    opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, null));
    assertSame(two, opponentInfoPlugin.getLastOpponent());
}
```

Keep in mind this is, as far as I'm aware, the only case that would be affected by the player being in a multi-combat zone (or not, in this case)

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
chaticon: This prevents the hp bar disappearing in the case where the player attacks an npc, then stops attacking anything, but the npc hasn't stopped attacking the player.  There actually is a test for it on line 184.

chaticon:
> While reviewing this, I've thought of another case which causes this code to fail, and I've written a brief test case which demonstrates it:
> 
> ```java
> @Test
> public void singleCombatIssue()
> {
>     Player localPlayer = mock(Player.class), two = mock(Player.class), three = mock(Player.class);
>     when(client.getLocalPlayer()).thenReturn(localPlayer);
>     // TODO: find some unique var/condition representing a player being in multi-combat and mock it to be false here
> 
>     opponentInfoPlugin.onInteractingChanged(new InteractingChanged(localPlayer, two));
>     assertSame(two, opponentInfoPlugin.getLastOpponent());
> 
>     // Another player tries to attack or cast magic on the player which causes interaction target changed from null -> player -> null in succession
>     opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, null));
>     assertSame(two, opponentInfoPlugin.getLastOpponent());
> 
>     opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, localPlayer));
>     assertSame(two, opponentInfoPlugin.getLastOpponent()); // this line fails
> 
>     opponentInfoPlugin.onInteractingChanged(new InteractingChanged(three, null));
>     assertSame(two, opponentInfoPlugin.getLastOpponent());
> }
> ```
> 
> Keep in mind this is, as far as I'm aware, the only case that would be affected by the player being in a multi-combat zone (or not, in this case)

If I'm understanding correctly, the reason this appears to change the value of lastOpponent is because you didn't also mock the localPlayer.getInteracting() method to return 'two'.  It uses that to check if the player is still attacking a different actor when an actor attacks them, to prevent lastOpponent from taking on the value of the actor they aren't attacking.  I did a test for this from lines 73-85.

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
chaticon: You're right, at one point I was using two npcs and I just never changed the name when I removed one of them.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
Nightfirecat: That isn't what this condition is written to be though, nor is it what that test is testing.

This condition, combined with all the parent conditions, works out to be:

```java
if (source == player && target == null && lastOpponent != null && lastOpponent.getInteracting() != player)
```

That is, when the player stops interacting with an NPC which _is not_ attacking the player. That is why, despite the test being written to test the scenario you described, this line is not being covered by tests. I highly recommend running the tests in IntelliJ using the "Run Test with Coverage" button and checking to see which lines of the plugin (most importantly the changed lines) have test coverage. That is how I identified that this `lastTime = Instant.now();` call is never executed by the tests.

Nightfirecat (changes requested):
I've squashed and rebased these changes. Please see [my comment here](https://github.com/runelite/runelite/pull/12990#discussion_r569111571) regarding the line which isn't covered by tests

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
chaticon: Sorry, I explained poorly.  What I meant was exactly what you described, I was just explaining what happens when the condition is false.  It is designed so that in the case I mentioned, where the player stops attacking an npc, but the npc is still attacking the player, the condition is false, so the hp bar stays visible.  That's why when—as you've described—the player stops interacting with an npc that isn't attacking the player, the condition is true.  The result of that is, as you would expect, the hp bar is set to disappear after WAIT time.  In the test, I was testing that case where the condition is false.  I've added a test where the condition is true now.

Nightfirecat:
The tests do add full coverage of the changed code, though I'm still not 100% sure I understand this code inside and out. I'll review this more in depth at a later time.

Nightfirecat:
I've added some further tests and cleaned up the test class a bit. @chaticon, does this still look good to you? From some quick ingame testing (in addition to the full test coverage provided by this PR's tests) it seems to still work just fine.

runelite-client/src/test/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPluginTest.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	@Test
+	public void testNpcInteractions()
+	{
+		final NPC npc = mock(NPC.class);
+
+		interactingChanged(npc, localPlayer);
+
+		assertNull(plugin.getLastOpponent());
```
chaticon: These tests aren't quite right; they test the behaviour of the current OpponentInfo, but not this one.  This one continues to track the lastOpponent for as long as the player is attacking or being attacked.  So, in testNpcInteractions for example, line 91 should be assertSame(npc, plugin.getLastOpponent()), because the player is being attacked by npc.  Then, on lines 96-97, the last opponent is being remembered for 5 seconds.  Finally, on line 113, even though the player has stopped attacking npc, npc is still attacking the player, so plugin.getLastTime() should be null.

chaticon:
> I've added some further tests and cleaned up the test class a bit. @chaticon, does this still look good to you? From some quick ingame testing (in addition to the full test coverage provided by this PR's tests) it seems to still work just fine.

Sorry for the late response.  Looks good to me.  And I think I may have commented on something outdated that you'd already fixed, sorry about that.

Nightfirecat (approved):
I think this is working as intended now, and has sufficiently thorough tests to demonstrate that.

Adam-:
This throws an exception on login
```
	at net.runelite.client.plugins.opponentinfo.OpponentInfoPlugin.onInteractingChanged(OpponentInfoPlugin.java:161)
	at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:67)
	at net.runelite.client.eventbus.EventBus.post(EventBus.java:238)
	at net.runelite.client.callback.Hooks.post(Hooks.java:169)
```

Additionally we should probably add comments to all of this logic since it looks complicated and I don't follow all of it.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -134,33 +138,48 @@ else if (worldType.contains(WorldType.LEAGUE))
...
+			}
 		}
+		else if (target == player
```
Adam-, line 158: Can the case of `target == player && opponent != null && opponent != lastOpponent` ever happen? You have 4 different branches here:

1) `target == player && opponent == null`
2) `target == player && opponent != null && opponent != lastOpponent` <-- does this happen?
3) `target != player && source == lastOpponent && target == player ...` obviously this cannot happen
4) `target != player && source == lastOpponent && opponent != lastOpponent`

I think this can't happen because it would require something to attack you while you were attacking something that also wasn't your opponent. And I don't think you can be attacking something which isn't your opponent.

If so, this can just be split I think into two separate ifs

```java
if (target == player && opponent == null) {
// Case of something attacking you when you aren't in combat
}
else if (source == lastOpponent && opponent != lastOpponent) {
// Case of something attacking you switching off while you are not attacking it
}
```

Adam-:
We may want to keep an eye on this since there is no despawn event hooks for the opponent actor, and it is still invoking getInteracting() on it. But I think it is probably okay.

Nightfirecat:
Since this was merged, I've found a couple unusual behaviors that this causes:

1. If a player follows you, the player comparison overlay appears. This is odd because they don't need to be fighting with you to cause this to appear--this happens even on non-PvP worlds/outside of the wilderness. If that player were to have a visible health bar when they followed you, such as if something was attacking them, I believe the opponent info overlay would appear as well.
2. I experienced a bug today where while in the nylocas room of Theatre of Blood, I died then saw the opponent info overlay persist with a nylocas (I'm not sure whether it was one of the smaller wave nylocas or if it was the boss) throughout the boss fight and after the fight was over, even when I was not interacting with it, and nothing was interacting with me. This persisted until I followed another player then stopped following them for 5 seconds.

Nightfirecat:
I've reverted this feature in 36ec314ab7d592c5b0f59183d7a5ee72bbf1fe35 for the time being. If we want to revisit adding this down the line, it will need more thorough ingame testing to ensure it doesn't cause these types of issues.

PR #3069, opened 2018-05-21
Add Low Energy Notification

Reasel:
As requested in issue: [#3057](https://github.com/runelite/runelite/issues/3057)

This will allow the user to set a threshold for run energy to be notified of.

PR #12995, opened 2020-12-30
Add Low/High Energy notifications to idle notifier

### LINKED ISSUE #3057: Notification on low run energy
Similiar to low hp/pray have a notification on low run energy where you can set the limit

Would be useful for some of the more draining bosses like inferno/olm/sara

Cyborger1:
This is a redo of PR #3069.

It seems the code was good, the only thing that was outdated would have been the config positions conflicting with other fields added over time.

I took the liberty to add an energy regen notification to accompany the low energy one.

Closes #3057

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
@@ -27,6 +27,7 @@ package net.runelite.client.plugins.idlenotifier;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
 import net.runelite.client.config.Units;
 
 @ConfigGroup("idlenotifier")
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
```diff
@@ -110,10 +111,36 @@ default int getPrayerThreshold()
 		return 0;
 	}
 
+	@ConfigItem(
+		keyName = "lowEnergy",
+		name = "Low Energy Notification Threshold",
+		description = "The amount of energy points remaining to send a notification at. A value of 0 will disable notification.",
```
Nightfirecat, line 117: Wouldn't users want to be able to be notified when they run out of energy completely?

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -93,6 +93,8 @@
 	private boolean notifyPosition = false;
 	private boolean notifyHitpoints = true;
 	private boolean notifyPrayer = true;
+	private boolean notifyLowEnergy = true;
+	private boolean notifyHighEnergy = true;
```
Nightfirecat, line 97: These names seem very strange, since they're backwards from what they actually do. They're basically only ever checked as `if (!notify) notifier.notify();`

Instead, flip the values and checks for these booleans so they start as `false` and trigger notifications when they're `true`.

Nightfirecat (changes requested):
This should credit @Reasel for the code they wrote. [`Co-authored-by` is a reasonable way to do that.](https://github.blog/2018-01-29-commit-together-with-co-authors/#how-it-works)

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierConfig.java
Cyborger1: I'll set this one to not trigger when the value is set to 100 then. I had it like this because I wanted it to be consistent with the other notifications, but I agree this is better.

runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
Cyborger1: I was following the way prayers and hitpoints was implemented, but I switched them around and made the var names a bit more descriptive.

PR #12996, opened 2020-12-31
Add Maven Assembly plugin to pom

nmcmahon-personal:
Add Maven assembly plugin to Runelite.

This adds an extra step when creating the Runelite Jar file. The extra step creates a "fat" jar i.e. a jar which contains runelite and all of its dependencies.

This will allow developers to easily run Runelite locally from the generated jar following a build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,6 +263,29 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<artifactId>maven-assembly-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


Adam-:
The build already outputs a fat jar (the artifact classifier is `shaded`), using the maven shade plugin. So I don't think this is necessary.

nmcmahon-personal:
@Adam- my bad, you're correct! Closing PR.

PR #12999, opened 2020-12-31
QuestListPlugin: Fix names for Rag and Bone Man and Dragon Slayer

JZomerlei:
Quests are not 'Dragon Slayer' or 'Rag and Bone Man' they now have 'I' at the end similar to monkey madness already did.


runelite-api/src/main/java/net/runelite/api/Quest.java
```diff
@@ -36,7 +36,7 @@
 	THE_CORSAIR_CURSE(301, "The Corsair Curse"),
 	DEMON_SLAYER(302, "Demon Slayer"),
 	DORICS_QUEST(303, "Doric's Quest"),
-	DRAGON_SLAYER(304, "Dragon Slayer"),
+	DRAGON_SLAYER_I(304, "Dragon Slayer I"),
```
Nightfirecat, line 39: Changing the enum names here will break uses in hub plugins.

PR #13004, opened 2021-01-03
Item Stats: Fix Anglerfish boost heal in wilderness

EliasLahham:
Fixes tooltip bug with anglerfish showing that a boosted heal is possible in the wilderness when it's not.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/food/Anglerfish.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/food/Anglerfish.java
@@ -25,21 +25,36 @@
 package net.runelite.client.plugins.itemstats.food;
 
 import net.runelite.api.Client;
+import net.runelite.api.Varbits;
 import net.runelite.client.plugins.itemstats.FoodBase;
+
+import javax.inject.Inject;
 
 public class Anglerfish extends FoodBase
 {
+	@Inject
+	private Client client;
+
 	public Anglerfish()
 	{
-		setBoost(true);
+
 	}
 
 	@Override
 	public int heals(Client client)
 	{
 		int maxHP = getStat().getMaximum(client);
-
 		int C;
+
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This is not correct, and is also something we can't easily fix. From the blog post where this change was added to the game:

> Anglerfish no longer overheal if the player is in the Wilderness or on a PvP or BH world AND has been in combat or recently attacked a player (within the last 10 seconds).

I can confirm eating an anglerfish in the wilderness works as expected, so this statement should be interpreted as `((inWilderness || inPvpWorld) && recentlyInCombat)`. We can easily check the former statements, but have no reliable way to know if we were recently in combat.

EliasLahham:
> 
> 
> This is not correct, and is also something we can't easily fix. From the blog post where this change was added to the game:
> 
> > Anglerfish no longer overheal if the player is in the Wilderness or on a PvP or BH world AND has been in combat or recently attacked a player (within the last 10 seconds).
> 
> I can confirm eating an anglerfish in the wilderness works as expected, so this statement should be interpreted as `((inWilderness || inPvpWorld) && recentlyInCombat)`. We can easily check the former statements, but have no reliable way to know if we were recently in combat.

I think from that blog post it would be `(inWilderness || (inPvpWorld && recentlyInCombat))` right? And if we can't check if someone is in combat at the present time, could we at least get the `inWilderness` change in?

Nightfirecat:
Like I said, you absolutely can overheal with an anglerfish in the wilderness if you are outside of combat--I confirmed this ingame yesterday before making my comment. Hence, it would not be correct to assume anglerfish overheals don't work in the wilderness generally.

EliasLahham:
Oh ya you're right. I incorrectly assumed this was the case after I implemented my logic and tested it in my dev environment. So since we can't check if combat happened recently, part of this bug will have to stay in the client for the time being.

Nightfirecat:
Given that this presently can't be fixed, I'm closing this PR. You can open a an issue for this so we can track it and potentially find a fix down the road for it.

PR #13005, opened 2021-01-04
Add shareable ground markers

### LINKED ISSUE #7853: Allow for (an easy way) to import ground markers
Would be useful to get started in raids/inferno rather than having to go in and marking everything yourself, sort of simular to bank tags

LlemonDuck:
Closes #7853.

Adds two menu options to the World Map globe: `Export Ground Markers` and `Import Ground Markers`.

Selecting `Export Ground Markers` will serialize all ground markers in active scenes into JSON, then copy the result onto the clipboard.

Selecting `Import Ground Markers` will deserialize clipboard JSON, then copy the results into the config, merging the points with already existing points for that region, and reloading. Importing points does not overwrite existing points.

Here's a video of the import in action:

https://user-images.githubusercontent.com/1868974/103497492-aa165b80-4e0f-11eb-8611-0296581c6b0d.mp4

Since this feature got a bit large to implement, I placed it into its own GroundMarkerSharingManager class, which takes an injected GroundMarkerPlugin to communicate with the plugin instance. The plugin passes over the `onWidgetMenuClicked` event, and as part of this change, provides package-level access to `getPoints`, `savePoints` and `loadPoints`.

To-do:
* other ClipboardUtil migrations
* Tighter serialization format

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
@@ -51,6 +51,7 @@ import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.MenuEntryAdded;
 import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.client.config.ConfigManager;
+import net.runelite.client.eventbus.EventBus;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.game.chatbox.ChatboxPanelManager;
 import net.runelite.client.plugins.Plugin;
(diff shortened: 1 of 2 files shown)


makingcomment:
Why does this import from and export to the clipboard instead of an actual file?

I feel like giving users a file to share would be much easier.

loldudester:
Presumably to match the behaviour of bank tags

LlemonDuck:
> Why does this import from and export to the clipboard instead of an actual file?
> 
> I feel like giving users a file to share would be much easier.



> Presumably to match the behaviour of bank tags

This is correct. It also removes the step of naming the file within the flow. A user can always paste into notepad or  such

Hydrox6:
Text is a lot easier to share, given that if it has a text box, it can take the text. Files require specific vectors.
Text is obvious at a glance; what you're importing is all visible, and you have to interact with it to make it work. A file requires you open it, and so can be missed. This is also a positive for moderation of the data.
Text is used elsewhere in the client. It's simple to grasp (given good UX). A file requires linking into the operating system for a file browser, and is more difficult to handle entirely correctly.

makingcomment:
I don't use the bank tags import/export, so I never noticed how that works. But consistency and ease of use/implementation are both good reasons, fair enough.

Maybe telling the user somewhere that the import/export process uses the clipboard would be helpful? (especially for export, where overwriting someone's clipboard when they weren't expecting it might not be appreciated)

LlemonDuck:
I've added two things:
1. ClipboardUtil is now used across the whole project, where applicable.
2. There is now a confirmation prompt before importing ground markers. New video below:

https://user-images.githubusercontent.com/1868974/103585591-4f354080-4eb1-11eb-83b7-71b7428dd4b1.mp4

PR #13010, opened 2021-01-05
XP Tracker: Add sort by most xp to XP Tracker

EliasLahham:
Added a new option to the XP tracker plugin. You can now select "Sort by XP" which will rearrange your current session skills from most XP gained to least.

sortbyXpgif

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
@@ -234,7 +234,6 @@ class XpInfoBox extends JPanel
 	void setInfoBoxPosition(int position)
 	{
 		panel.setComponentZOrder(this, position);
-		panel.revalidate();
 	}
 
 	void reset()
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
@@ -177,6 +177,7 @@ class XpPanel extends PluginPanel
 			infoBoxes.get(skill).setInfoBoxPosition(i);
 			i++;
 		}
+		revalidate();
 	}
 
 	void resetAllInfoBoxes()


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -282,6 +281,33 @@ boolean hasOverlay(final Skill skill)
 		return overlayManager.anyMatch(o -> o instanceof XpInfoBoxOverlay && ((XpInfoBoxOverlay) o).getSkill() == skill);
 	}
 
+	void sortByXp()
+	{
+		for (Skill skill : Skill.values())
+		{
+			if (xpState.getSkillSnapshot(skill).getXpGainedInSession() != 0 && !skill.getName().equals("Overall"))
```
Nightfirecat, line 288:
```suggestion
			if (xpState.getSkillSnapshot(skill).getXpGainedInSession() != 0 && skill != Skill.OVERALL)
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -28,14 +28,12 @@
...
+import com.google.common.collect.Lists;
 import com.google.inject.Binder;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
 import java.time.temporal.ChronoUnit;
-import java.util.Arrays;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
Nightfirecat, line 36: Don't star import here. https://github.com/runelite/runelite/wiki/Code-Conventions#imports

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanel.java
```diff
@@ -27,8 +27,7 @@
 
 import java.awt.BorderLayout;
 import java.awt.GridLayout;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
Nightfirecat, line 30: Same here, no star import

Nightfirecat (changes requested):
Also, is there some delay in your gif when clicking the added option? Seems like it does not immediately sort the boxes but waits until the next game tick... This should sort them immediately upon selection

EliasLahham:
@Nightfirecat Ya it must be waiting for the next game tick, not sure how to remedy that off the top of my head

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
```diff
@@ -282,6 +288,33 @@ boolean hasOverlay(final Skill skill)
 		return overlayManager.anyMatch(o -> o instanceof XpInfoBoxOverlay && ((XpInfoBoxOverlay) o).getSkill() == skill);
 	}
 
+	void sortByXp()
+	{
+		for (Skill skill : Skill.values())
+		{
+			if (xpState.getSkillSnapshot(skill).getXpGainedInSession() != 0 && skill != Skill.OVERALL)
```
Nightfirecat, line 295: Since the xp gained in session is used twice here, store it in a variable instead of calling `getSkillSnapshot()` twice.

Nightfirecat:
You may need a `repaint()` or `revalidate()` call on the xp panel then. As-is, I would not want to add this unless this can be updated to reflect instantly in the UI.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -231,6 +231,11 @@ public void popupMenuCanceled(PopupMenuEvent popupMenuEvent)
 		add(container, BorderLayout.NORTH);
 	}
 
+	void sortByXp(int position)
```
Nightfirecat, line 234: This isn't a very good method name since it doesn't sort the box, it simply sets its z order. Change the method name accordingly

EliasLahham:
@Nightfirecat There was a revalidate() method that fixed the delayed sort, so I implemented that along with the other requested changes.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -231,6 +231,12 @@ public void popupMenuCanceled(PopupMenuEvent popupMenuEvent)
 		add(container, BorderLayout.NORTH);
 	}
 
+	void setInfoBoxPosition(int position)
+	{
+		panel.setComponentZOrder(this, position);
+		panel.revalidate();
```
Nightfirecat, line 237: Move this to the end of `XpPanel#sortByXp()` so it is called only once per sort instead of once per box. I've tested this and can confirm the UI is visually updated the same way with that change in place.

Nightfirecat (changes requested):

Nightfirecat (approved):

Adam-:
This is currently massively over complicated for what it's doing. I'm not sure this feature is even that useful though? All this would need to do is sort the infoboxes in the xppanel, which would only be a few lines (get the keyset of `infoBoxes`, sort that, and then apply z order).

EliasLahham:
@Adam- I'll try and find some time and simplify the code based on your comments. I guess the only reason I added this feature is because on long play sessions I like having my xp tracker sorted from XP gained especially when there are 10+ skills listed

PR #13045, opened 2021-01-13
chatcommands: Add Soul Wars Zeal command (!sw)

Broooklyn:
Currently many people have been using `!lvl soul_wars_zeal`, which isn't exactly the easiest to type, and due to it being recognized as a skill, the output incorrectly displays it as such. 

This commit adds the `!sw` command to more easily and appopriately display one's Soul Wars Zeal and rank.

image


Nightfirecat (approved):
LGTM

PR #13053, opened 2021-01-13
Add item requirement filtering for skill challenge clues

LlemonDuck:
Skill challenge clues, although they have item requirements, are not currently filtered by the `clue` bank tag.

A more proper fix might be to create an interface for clue scrolls to implement, like `ItemRequiringClueScroll` or something. Let me know if you'd prefer that.


Nightfirecat (approved):

Adam-:
This seems ok. Maybe the item requirement is something we could put into the base ClueScroll instead of having to duplicate this? Regardless this needs rebase since this logic moved into the clue scroll plugin.

PR #13062, opened 2021-01-15
emoji: add weary face emoji

### LINKED ISSUE #13046: Can we please get the IOS moaning emoji
download

Hi, I am in desperate need of this emoji from IOS and I think many people would use it in the chat plugin. Runelite already has many different emojis similar to ios emojis  and I would appreciate it if this one was added.

WolffTech:
Closes #13046 

adds the weary face emoji using `<_>` as the way to trigger it.


dekvall:
Your image file-size is too large, please run it through `optipng` or use this 1f629

Issue #12717, opened 2020-10-29
XP Bar


Conzec:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Under "XP Tracker" i would love to have the option of hiding for example "XP Per Hour" and ONLY see the procentage between levels. I know OSBuddy used to have this.

**Describe alternatives you've considered**
No.

**Additional context**
No

PR #13065, opened 2021-01-15
XpTracker: Toggle stats

mattBrzezinski:
## Overview
Resolves: #12717 

Toggles compact view of XP Info boxes on click.

Single skill toggled:
toggleOne

Multiple skills toggled:
toggleTwo

One addition that might be worth while considering is placing the skill icon when toggling stats somewhere on the progress bar? Feel free to let me know if there are some changes you'd like to see.

## Alternatives
The feature request only asked for a stats toggle. Personally I don't think this looks as visually appealing, containing wasted whitespace.

toggledStats

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
@@ -29,6 +29,8 @@ package net.runelite.client.plugins.xptracker;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.math.RoundingMode;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
@@ -81,20 +83,27 @@ class XpInfoBox extends JPanel
 	private static final String REMOVE_STATE = "Remove from canvas";
 	private static final String ADD_STATE = "Add to canvas";
 
+	private static final EmptyBorder DEFAULT_PROGRESS_WRAPPER_BORDER = new EmptyBorder(0, 7, 7, 7);
+	private static final EmptyBorder COMPACT_PROGRESS_WRAPPER_BORDER = new EmptyBorder(5, 1, 5, 5);
+	private static final int PROGRESS_SKILL_ICON_POSITION = 0;
+
 	// Instance members
 	private final JComponent panel;
 
 	@Getter(AccessLevel.PACKAGE)
 	private final Skill skill;
 
-	/* The tracker's wrapping container */
+	// The tracker's wrapping container
 	private final JPanel container = new JPanel();
 
-	/* Contains the skill icon and the stats panel */
+	// Contains the skill icon and the stats panel
 	private final JPanel headerPanel = new JPanel();
 
-	/* Contains all the skill information (exp gained, per hour, etc) */
+	// Contains all the skill information (exp gained, per hour, etc)
 	private final JPanel statsPanel = new JPanel();
+
+	// Contains progress bar and compact-view icon
+	private final JPanel progressWrapper = new JPanel();
 
 	private final ProgressBar progressBar = new ProgressBar();
 
@@ -182,18 +191,12 @@ class XpInfoBox extends JPanel
 			}
 		});
 
-		JLabel skillIcon = new JLabel(new ImageIcon(iconManager.getSkillImage(skill)));
-		skillIcon.setHorizontalAlignment(SwingConstants.CENTER);
-		skillIcon.setVerticalAlignment(SwingConstants.CENTER);
-		skillIcon.setPreferredSize(new Dimension(35, 35));
-
 		headerPanel.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		headerPanel.setLayout(new BorderLayout());
 
 		statsPanel.setLayout(new DynamicGridLayout(2, 2));
 		statsPanel.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		statsPanel.setBorder(new EmptyBorder(9, 2, 9, 2));
-
 
 		topLeftStat.setFont(FontManager.getRunescapeSmallFont());
 		bottomLeftStat.setFont(FontManager.getRunescapeSmallFont());
@@ -205,20 +208,24 @@ class XpInfoBox extends JPanel
 		statsPanel.add(bottomLeftStat);  // bottom left
 		statsPanel.add(bottomRightStat); // bottom right
 
-		headerPanel.add(skillIcon, BorderLayout.WEST);
+		JLabel headerSkillIcon = getSkillIcon(iconManager, skill, 35, 35, false);
+		headerPanel.add(headerSkillIcon, BorderLayout.WEST);
 		headerPanel.add(statsPanel, BorderLayout.CENTER);
 
-		JPanel progressWrapper = new JPanel();
 		progressWrapper.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		progressWrapper.setLayout(new BorderLayout());
-		progressWrapper.setBorder(new EmptyBorder(0, 7, 7, 7));
+		progressWrapper.setBorder(DEFAULT_PROGRESS_WRAPPER_BORDER);
 
 		progressBar.setMaximumValue(100);
 		progressBar.setBackground(new Color(61, 56, 49));
 		progressBar.setForeground(SkillColor.find(skill).getColor());
 		progressBar.setDimmedText("Paused");
 
-		progressWrapper.add(progressBar, BorderLayout.NORTH);
+		JLabel progressSkillIcon = getSkillIcon(iconManager, skill, 25, 16, true);
+		progressSkillIcon.setVisible(false);
+
+		progressWrapper.add(progressSkillIcon, BorderLayout.WEST, PROGRESS_SKILL_ICON_POSITION);
+		progressWrapper.add(progressBar, BorderLayout.CENTER);
 
 		container.add(headerPanel, BorderLayout.NORTH);
 		container.add(progressWrapper, BorderLayout.SOUTH);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I think any "collapsed" state we add to xp info boxes should include the skill's icon at minimum so it can be identified.

mattBrzezinski:
> Edit: I don't mean the alternative design you showed, though I couldn't say what design would look best when displaying only the bar itself.

Agreed, the alternative looks quite poor. I just forced-pushed some changes below are a couple of images of what it looks like.
Other alternatives could be to overlay the skill icon in either the `foreground / background` of the progress bar? Although from quickly googling around that seems to be a bit more complicated than it's worth.

toggledSkillIcon1
toggledSkillIcon2

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -26,9 +26,8 @@
  */
 package net.runelite.client.plugins.xptracker;
 
-import java.awt.BorderLayout;
-import java.awt.Color;
-import java.awt.Dimension;
+import java.awt.*;
```
mattBrzezinski, line 29: Looks like IntelliJ decided to import all here. I think I need to change my IDE settings, will follow the `Wiki` guide on it shortly.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -200,20 +205,25 @@ public void popupMenuCanceled(PopupMenuEvent popupMenuEvent)
...
-		progressWrapper.setBorder(new EmptyBorder(0, 7, 7, 7));
+		progressWrapper.setBorder(DEFAULT_PROGRESS_WRAPPER_BORDER);
 
 		progressBar.setMaximumValue(100);
 		progressBar.setBackground(new Color(61, 56, 49));
 		progressBar.setForeground(SkillColor.find(skill).getColor());
 		progressBar.setDimmedText("Paused");
 
-		progressWrapper.add(progressBar, BorderLayout.NORTH);
+		JLabel progressSkillIcon = getSkillIcon(iconManager, skill, 10, 10, true);
```
Nightfirecat, line 221: Instead of using the full size skill images here, [use `SkillIconManager`'s `small` argument to get a smaller image to work with](http://static.runelite.net/api/runelite-client/net/runelite/client/game/SkillIconManager.html#getSkillImage(net.runelite.api.Skill,boolean)). Even if the icon needs to be further resized, this will be a better starting point for smaller icons.

mattBrzezinski:
Using the smaller icons fixes the resolution issue as well.

single
multiple

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -243,6 +252,37 @@ void update(boolean updated, boolean paused, XpSnapshotSingle xpSnapshotSingle)
...
+		headerPanel.setVisible(!headerPanel.isVisible());
+		progressWrapper.getComponent(progressSkillIconPosition).setVisible(!progressSkillIconState);
+	}
+
+	private JLabel getSkillIcon(SkillIconManager iconManager, Skill skill, int width, int height, boolean small)
+	{
+		JLabel skillIcon = new JLabel();
+		Image image = small ? iconManager.getSkillImage(skill, true) : iconManager.getSkillImage(skill);
+
+		skillIcon.setIcon(new ImageIcon(image));
```
Nightfirecat, line 278:
```suggestion

		skillIcon.setIcon(new ImageIcon(iconManager.getSkillImage(skill, small)));
```

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
mattBrzezinski: I did modify [Line 208](https://github.com/runelite/runelite/pull/13065/files/a2866ce87fa911950d78083988cb0f6ba5ec102a#diff-3f7bdad15fc0f983fae5f9172f851786bc503296baf07e81bb7e10a392fbecb5R208) to use this function, but get the full sized image instead of the `small`.

Nightfirecat: I see that, my point is that `SkillIconManager#getSkillImage(Skill)` is an overload which resolves to `SkillIconManager#getSkillImage(Skill, false)`, so `small` can be passed as the second argument to the latter signature instead of using a ternary to indirectly pass `false` to it.

mattBrzezinski: Yeah you're definitely right... duh.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -29,6 +29,7 @@
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
+import java.awt.Image;
```
Nightfirecat, line 32: This is unused now. (I'm committing its removal via GitHub's UI)
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -243,6 +251,36 @@ void update(boolean updated, boolean paused, XpSnapshotSingle xpSnapshotSingle)
...
+		else
+		{
+			progressWrapper.setBorder(DEFAULT_PROGRESS_WRAPPER_BORDER);
+		}
+
+		headerPanel.setVisible(!headerPanel.isVisible());
+		progressWrapper.getComponent(progressSkillIconPosition).setVisible(!progressSkillIconState);
+	}
+
+	private JLabel getSkillIcon(SkillIconManager iconManager, Skill skill, int width, int height, boolean small)
```
Nightfirecat, line 272: This method can either be made static or have the SkillIconManager argument removed from it, one or the other.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -243,6 +251,36 @@ void update(boolean updated, boolean paused, XpSnapshotSingle xpSnapshotSingle)
 		SwingUtilities.invokeLater(() -> rebuildAsync(updated, paused, xpSnapshotSingle));
 	}
 
+	void toggleHeaderPanel(JPanel progressWrapper)
```
Nightfirecat, line 254:
```suggestion
	private void toggleHeaderPanel(JPanel progressWrapper)
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -243,6 +251,36 @@ void update(boolean updated, boolean paused, XpSnapshotSingle xpSnapshotSingle)
...
+	{
+		int progressSkillIconPosition = 0;
```
Nightfirecat, line 256: This can be made into a class constant instead of a method field as it is never changed.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -135,6 +137,13 @@
 		final JMenuItem resetOthers = new JMenuItem("Reset others");
 		resetOthers.addActionListener(e -> xpTrackerPlugin.resetOtherSkillState(skill));
 
+		// ProgressWrapper containing SkillIcon and ProgressBar
+		final JPanel progressWrapper = new JPanel();
```
Nightfirecat, line 141: This can remain on L211 since it's not used until then

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
mattBrzezinski: I brought this up here as we add an action listener for in a couple lines lower. I felt that was more appropriate than moving the listener lower down.

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -82,6 +82,9 @@
 	private static final String REMOVE_STATE = "Remove from canvas";
 	private static final String ADD_STATE = "Add to canvas";
 
+	private final EmptyBorder DEFAULT_PROGRESS_WRAPPER_BORDER = new EmptyBorder(0, 7, 7, 7);
+	private final int PROGRESS_SKILL_ICON_POSITION = 0;
```
Nightfirecat, line 86: (committing via GitHub's UI)
```suggestion
	private static final int PROGRESS_SKILL_ICON_POSITION = 0;
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpInfoBox.java
```diff
@@ -82,6 +82,9 @@
 	private static final String REMOVE_STATE = "Remove from canvas";
 	private static final String ADD_STATE = "Add to canvas";
 
+	private final EmptyBorder DEFAULT_PROGRESS_WRAPPER_BORDER = new EmptyBorder(0, 7, 7, 7);
```
Nightfirecat, line 85: (committing this via GitHub's UI)
```suggestion
	private static final EmptyBorder DEFAULT_PROGRESS_WRAPPER_BORDER = new EmptyBorder(0, 7, 7, 7);
```

Nightfirecat:
This code LGTM, however the behavior when you drag+drop them is very strange.

https://user-images.githubusercontent.com/2199511/105423702-f0f0b780-5c3d-11eb-9aa2-9ff9f504cba7.mp4

mattBrzezinski:
> This code LGTM, however the behavior when you drag+drop them is very strange.
> 
>  xp-infobox-drag-drop-reorder-bug.mp4

I would assume that this is because the `headerPanel` is hidden and not actually removed. Could try removing and rebuilding the tracker object on toggle, however I'd assume that's quite an expensive operation.

If you have any suggestions on where to look into I can take the time to try and find a solution. I'm not well versed in Java or AWT anymore to know.

Nightfirecat:
No, this is caused by some Swing weirdness in our DragAndDropReorderPane. I'm looking at fixing that now

Nightfirecat (approved):
I've pushed a commit which fixes the DragAndDropReorderPane behavior with children of varying sizes and squashed your work into a single commit; I believe this is fine to be merged now.

You can see the fixed behavior in the following video:

https://user-images.githubusercontent.com/2199511/105438597-37084400-5c5b-11eb-9ee1-6edff08d8447.mp4

Nightfirecat:
Rebased; I've changed the behavior to toggle the collapsed view on click rather than via right-click, as this matches loot tracker behavior, and is generally much more ergonomic to use.

Nightfirecat:
The DragAndDropReorderPane change to accommodate children of varying sizes has been added in 25c4beb8bc2fff7bdcff77d9f109642766114f82, so my commit (c4d6c2a) can be dropped from this PR now.

PR #13072, opened 2021-01-16
discord: Use regions for raiding activity and remove varbit handling

Broooklyn:
Currently, after being in a ToB party, Jagex does not properly unset `Varbits.THEATRE_OF_BLOOD` when leaving the Theatre area, only when returning. Since we use that varbit for discord activity, if you were to do a ToB and then leave afterward, your discord activity would continue to show `Theatre of Blood` regardless of your location, until you either log out or return to Ver Sinhaza. 

With that in mind, it is much easier to use the regions for raiding activity rather than the varbits. With those changed, it is no longer necessary to handle varbits within the discord plugin, thus all code related to varbits can be removed.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
@@ -307,8 +307,8 @@ enum DiscordGameEventType
 	MG_VOLCANIC_MINE("Volcanic Mine", DiscordAreaType.MINIGAMES, 15263, 15262),
 
 	// Raids
-	RAIDS_CHAMBERS_OF_XERIC("Chambers of Xeric", DiscordAreaType.RAIDS, Varbits.IN_RAID),
-	RAIDS_THEATRE_OF_BLOOD("Theatre of Blood", DiscordAreaType.RAIDS, Varbits.THEATRE_OF_BLOOD),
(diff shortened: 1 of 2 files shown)


Broooklyn:
Updated PR to use regions instead of varbits for raiding activity and to remove varbit-related code from Discord plugin.

runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordGameEventType.java
```diff
@@ -307,8 +303,8 @@
 	MG_VOLCANIC_MINE("Volcanic Mine", DiscordAreaType.MINIGAMES, 15263, 15262),
 
 	// Raids
-	RAIDS_CHAMBERS_OF_XERIC("Chambers of Xeric", DiscordAreaType.RAIDS, Varbits.IN_RAID),
-	RAIDS_THEATRE_OF_BLOOD("Theatre of Blood", DiscordAreaType.RAIDS, Varbits.THEATRE_OF_BLOOD),
+	RAIDS_CHAMBERS_OF_XERIC("Chambers of Xeric", DiscordAreaType.RAIDS, 12889, 13136, 13137, 13138, 13139, 13140, 13141, 13145, 13393, 13394, 13395, 13396, 13397, 13401),
+	RAIDS_THEATRE_OF_BLOOD("Theatre of Blood", DiscordAreaType.RAIDS, 12611, 12612, 12613, 12867, 12869, 13122, 13123, 13125),
```
Nightfirecat, line 307: I think 12868 or 13124 are used for the person selected to run Sotetseg's maze?

PR #13094, opened 2021-01-20
Fix widget ids for barbarian assault

BegOsrs:


LlemonDuck:
Pretty sure this was just fixed in 1bec513137495ad7f7ec8ca976fe51009cf263ca

PR #12862, opened 2020-11-29
Add Green Tears of Guthix Timer (fixes #12858)

cgati:
Screen Shot 2020-11-29 at 3 21 58 PM
Screen Shot 2020-11-29 at 3 22 03 PM

This PR adds an optional Green Tears timer to the Tears of Guthix Plugin (fixes #12858).

PR #13102, opened 2021-01-24
Add Green Tears of Guthix Timer Option (fixes #13023)

### LINKED ISSUE #13023: Add toggle for Green Tears of Guthix Timers
Original issue mentioned here #12858 

Looks like there was a configurable option originally implemented but was striped #12862 

People with varying degrees of color blindness may have difficulties distinguishing between the colors but would still like to take advantage of the plugin. Or anyone that might use blue light filtering on their pc (flux, redshift, etc)

cgati:
@abextm I think the linked issue and comments like https://github.com/runelite/runelite/pull/12862#issuecomment-766186979 are a great reason to add this toggle back. Would you kindly review?

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixConfig.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2020, cgati <https://github.com/cgati>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.tearsofguthix;
+
+import java.awt.Color;
+import net.runelite.client.config.Alpha;
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+import net.runelite.client.util.ColorUtil;
+
+@ConfigGroup("tearsofguthix")
+public interface TearsOfGuthixConfig extends Config
+{
+	@ConfigItem(
+		keyName = "showGreenTearsTimer",
+		name = "Enable Green Tears Timer",
+		description = "Configures whether to display a timer for green tears or not",
+		position = 1
+	)
+	default boolean showGreenTearsTimer()
+	{
+		return true;
+	}
+
+	@Alpha
+	@ConfigItem(
+		keyName = "blueTearsColor",
+		name = "Blue Tears Color",
+		description = "Color of Blue Tears timer",
+		position = 2
+	)
+	default Color getBlueTearsColor()
+	{
+		return ColorUtil.colorWithAlpha(Color.CYAN, 100);
+	}
+
+	@Alpha
+	@ConfigItem(
+		keyName = "greenTearsColor",
+		name = "Green Tears Color",
+		description = "Color of Green Tears timer",
+		position = 3
+	)
+	default Color getGreenTearsColor()
+	{
+		return ColorUtil.colorWithAlpha(Color.GREEN, 100);
+	}
+
+}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
@@ -36,17 +36,18 @@ import net.runelite.client.ui.overlay.Overlay;
 import net.runelite.client.ui.overlay.OverlayLayer;
 import net.runelite.client.ui.overlay.OverlayPosition;
 import net.runelite.client.ui.overlay.components.ProgressPieComponent;
+import net.runelite.client.util.ColorUtil;
 
 class TearsOfGuthixOverlay extends Overlay
 {
-	private static final Color CYAN_ALPHA = new Color(Color.CYAN.getRed(), Color.CYAN.getGreen(), Color.CYAN.getBlue(), 100);
-	private static final Color GREEN_ALPHA = new Color(Color.GREEN.getRed(), Color.GREEN.getGreen(), Color.GREEN.getBlue(), 100);
 	private static final Duration MAX_TIME = Duration.ofSeconds(9);
+	private final TearsOfGuthixConfig config;
 	private final TearsOfGuthixPlugin plugin;
 
 	@Inject
-	private TearsOfGuthixOverlay(TearsOfGuthixPlugin plugin)
+	private TearsOfGuthixOverlay(TearsOfGuthixConfig config, TearsOfGuthixPlugin plugin)
 	{
+		this.config = config;
 		this.plugin = plugin;
 		setPosition(OverlayPosition.DYNAMIC);
 		setLayer(OverlayLayer.ABOVE_SCENE);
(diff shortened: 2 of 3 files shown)


Alexsuperfly:
might as well do a toggle for both, cuz some guy is gonna complain he cant have only green

abextm:
I would rather change the color so it has a different luminance so they can tell them apart

cgati:
@abextm I'm not sure that I agree. I think the color matching the stream makes sense from a default perspective, and adapting the color to cater for a single type of color differentiation issue might result in others having the same sort of issue. 

From a strictly color blindness perspective, I'd argue the Colorblindness Correction option on the GPU plugin could suffice when it can be used. For users with color temperature changes, such as the blue light filtering mentioned in https://github.com/runelite/runelite/issues/13023, I think the ability to turn the green stream timers off is a more than reasonable option.

Adam-:
To me, it seems fine having a green tears toggle.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
```diff
@@ -87,6 +89,15 @@ else if (object.getId() == ObjectID.GREEN_TEARS ||
...
 				: 1));
 
+			if (!config.showGreenTearsTimer())
+			{
+				if (object.getId() == ObjectID.GREEN_TEARS ||
+					object.getId() == ObjectID.GREEN_TEARS_6666)
+				{
+					return;
+				}
+			}
```
Cyborger1, line 99: Personally, I'd put this check either at the beginning of the lambda or at L68 just after the `if (position == null)`. Would avoid calculating and prepping the pie component when it's not going to be used anyway.

cgati: This was a choice I made while during development, noticing that if I toggled the feature, the green timer would be incorrect until the stream moved.

I'm not sure that the calculations are at such a cost that it would be worth introducing this behaviour.

Cyborger1: You must be thinking of having done that on the plugin side such that it would affect what `plugin.getStreams()` returns.

Moving the check on the overlay side up a few lines will not introduce that behaviour.
I ran that change on my end to confirm it doesn't cause any issues.

cgati: Oops. Not used to GitHub here, intended to comment while resolving. Intended message:

Indeed it is! Sorry about that, I had some difficulties getting the build running with tests so I drew from memory at the time of commenting.

Looks good on my end, I've pushed up an amended commit addressing this change.

Adam-: You can move it above even the `getCanvasLocation` call. 

Also do:

```java
			if ((object.getId() == ObjectID.GREEN_TEARS || object.getId() == ObjectID.GREEN_TEARS_6666)
				&& !config.showGreenTearsTimer())
			{
				return;
			}
```

Since the id check is faster than the config check.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixConfig.java
```diff
@@ -0,0 +1,45 @@
...
+{
+	@ConfigItem(
+		keyName = "showGreenTearsTimer",
+		name = "Enable Green Tears Timer",
+		description = "Configures whether to display a timer for green tears or not",
+		position = 1
+	)
+	default boolean showGreenTearsTimer()
+	{
+		return false;
```
Adam-, line 42: This should probably default to `true`

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
cgati: Good call, I'll make this change now. Thanks, @Adam- !

Adam-:
On second thought, if the original issue is colorblindness, then we should probably just make a color config for each of the colors instead?

cgati:
Not opposed to the addition, though it's an increase in scope to the main ask of adding the toggle itself. Would it be reasonable to merge this as-is and circle back on color options as a future enhancement?

Adam-:
I don't see a problem with just adding the color configs. You'd still need to keep the boolean switch for the green color since there is no way to unset colors which default to not-unset I think.

cgati:
@Adam- I wasn't sure of the time required to implement the necessary changes, which is where the push back was coming from. I found it simple enough to add. Looking forward to your feedback!

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
```diff
@@ -39,14 +39,14 @@
 
 class TearsOfGuthixOverlay extends Overlay
 {
-	private static final Color CYAN_ALPHA = new Color(Color.CYAN.getRed(), Color.CYAN.getGreen(), Color.CYAN.getBlue(), 100);
```
Adam-, line 42: These two removed colors should be what the defaults are.

cgati: Oops, thanks for that. I've pushed up a change to resolve.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixConfig.java
```diff
@@ -0,0 +1,71 @@
...
+	{
+		return new Color(Color.CYAN.getRed(), Color.CYAN.getGreen(), Color.CYAN.getBlue(), 100);
+	}
+
+	@Alpha
+	@ConfigItem(
+		keyName = "greenTearsColor",
+		name = "Green Tears Color",
+		description = "Color of Green Tears timer",
+		position = 2
```
Cyborger1, line 64: You've put `position = 2` on both green and blue tears color.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixConfig.java
```diff
@@ -0,0 +1,71 @@
...
+	@Alpha
+	@ConfigItem(
+		keyName = "blueTearsColor",
+		name = "Blue Tears Color",
+		description = "Color of Blue Tears timer",
+		position = 2
+	)
+	default Color getBlueTearsColor()
+	{
+		return new Color(Color.CYAN.getRed(), Color.CYAN.getGreen(), Color.CYAN.getBlue(), 100);
```
Cyborger1, line 56: I think you could use `ColorUtil.colorWithAlpha()` for these.

Cyborger1: The function did not exist at the time `CYAN_ALPHA` and `GREEN_ALPHA` were defined, which is why they weren't using it.

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
```diff
@@ -55,8 +55,19 @@ private TearsOfGuthixOverlay(TearsOfGuthixPlugin plugin)
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
+		Color BLUE_TEARS_ALPHA = config.getBlueTearsColor();
+		Color GREEN_TEARS_ALPHA = config.getGreenTearsColor();
+		Color BLUE_TEARS_COLOR = new Color(BLUE_TEARS_ALPHA.getRed(), BLUE_TEARS_ALPHA.CYAN.getGreen(), BLUE_TEARS_ALPHA.CYAN.getBlue());
+		Color GREEN_TEARS_COLOR = new Color(GREEN_TEARS_ALPHA.getRed(), GREEN_TEARS_ALPHA.getGreen(), GREEN_TEARS_ALPHA.getBlue());
```
Cyborger1, line 61: You should use camelCase notation here as those aren't static variables anymore. I'd propose calling the first two `[green/blue]TearsFill` and the other two `[green/blue]TearsBorder` as they are used for the fill and border of the pie timers.

For the border colors, you could also use `ColorUtil.colorWithAlpha(color, 0xFF)`.

Also I just noticed L60 looks broken, it has some Cyan poking through, `BLUE_TEARS_ALPHA.CYAN.getXX()` .

cgati: Yeah, noticed the extra CYAN. I was about to replace the line with `ColorUtil.colorWithAlpha(color, 255)`. Is the preference on using hexadecimal here?

Cyborger1: I saw it used with `0xFF` in TextComponent.java [here](https://github.com/runelite/runelite/blob/c9cd33a8dd4eff05159fac664420006a74baec99/runelite-client/src/main/java/net/runelite/client/ui/overlay/components/TextComponent.java#L104), but I think either is fine.

Cyborger1: If you're going to use `, 100)` in the config fields you should stay consistent and use `, 255)` here I suppose.

cgati: Oops, didn't see this reply until after I pushed up `0xFF`. I can go either way, replacing `100` with `0x64` or `0xFF` to `255`.

Cyborger1: I'd say go for the `255`.

At the same time, I noticed you named the green border color `greenTearsColor`, while the blue one is correctly named `blueTearsBorder`.

I think after that it's going to look pretty good to merge, but Adam will likely be the final judge.

cgati: Ah, yes. Apologies for the sloppiness here, my attention is a bit split right now. I appreciate the thorough review and guidance, thank you!

runelite-client/src/main/java/net/runelite/client/plugins/tearsofguthix/TearsOfGuthixOverlay.java
```diff
@@ -55,8 +56,19 @@ private TearsOfGuthixOverlay(TearsOfGuthixPlugin plugin)
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
```
Cyborger1, line 58: Actually, one more optimisation I can think of would be to have `if (plugin.getStreams().size() == 0) { return null; }` here. It would avoid having to hit up the config and set up colors on every render pass when there aren't even any streams active.

cgati: Not at all! I appreciate a finely tuned machine. I regret earlier comments that might indicate otherwise where I prioritized expediency over thoroughness. Split attention indeed..

Issue #13105, opened 2021-01-24
Cannon plugin not tracking cannonballs correctly after being destroyed and reclaimed


ryleymcrae:
**Description**
I brought my cannon to KBD with the 'cannon' plugin enabled. The settings for my plugin showed how many cannonballs were remaining and the double-hit spots. After some time, KBD destroyed my cannon when there was 7 cannonballs left in it. Although my cannon was destroyed, the cannon overlay remained where it was. May it also be noted that after I reclaimed my cannon and tried to use it again... 7 cannonballs were what the plugin considered to be the max (actually 30 in game). After filling my cannon with cannonballs again, the overlay displayed that there were 7 cannonballs loaded and starting counting down to -23.

**To Reproduce**
1. Go to KBD with the cannon plugin enabled and the overlay on.
2. Fill your cannon and let it begin to fire cannonballs
3. Wait for KBD to destroy your cannon (overlay remains Error)
4. Reclaim your cannon and try placing it down and letting it fire (see Error).

**Expected behavior**
It would be expected that the cannon overlay would disappear after it was destroyed and perhaps an overlay message in the game window saying "Your cannon has been destroyed. See Nulodion in-game to reclaim it."

**Screenshots**
No Screenshots. Sorry :{

**Environment:**
 - OS: Windows 10 Home x64 Version 10.0.18363
 - RuneLite version: 1.6.36.2

PR #13106, opened 2021-01-24
Timers: track cannon destroyed message

Alexsuperfly:


Nightfirecat:
This helps address part of #13105

PR #13107, opened 2021-01-24
Cannon: fix reclaimed cannon initial cball counter

### LINKED ISSUE #13105: Cannon plugin not tracking cannonballs correctly after being destroyed and reclaimed

Alexsuperfly:
A cannon that has been destroyed with cannon balls loaded
will still have them loaded when reclaimed.  This commit
does a diff of the inventory on the first message that says 
you are loading the cannon.  If the diff and inventory 
after that message both have cannon balls in them, then
we know the maximum amount possible were stored so set 
the counter to max.

Fixes #13105

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -39,6 +39,7 @@ import net.runelite.api.GameObject;
 import net.runelite.api.GameState;
 import net.runelite.api.InventoryID;
 import net.runelite.api.Item;
+import net.runelite.api.ItemContainer;
 import net.runelite.api.ItemID;
 import static net.runelite.api.ObjectID.CANNON_BASE;
 import net.runelite.api.Player;
@@ -301,6 +302,23 @@ public class CannonPlugin extends Plugin
 			cannonPlaced = true;
 			addCounter();
 			cballsLeft = 0;
+
+			final ItemContainer inventory = client.getItemContainer(InventoryID.INVENTORY);
+			if (inventory != null)
+			{
+				int invCballs = inventory.count(ItemID.GRANITE_CANNONBALL) > 0
+						? inventory.count(ItemID.GRANITE_CANNONBALL)
+						: inventory.count(ItemID.CANNONBALL);
+				// Cannonballs are always forcibly loaded after the furnace is added. If the player has more than
+				// the max number of cannon balls in their inventory, the cannon will always be fully filled.
+				// This is preferable to using the proceeding "You load the cannon with x cannon balls" message
+				// since it will show a lower number of cannon balls if the cannon is already partially-filled
+				// prior to being placed.
+				if (invCballs >= MAX_CBALLS)
+				{
+					cballsLeft = MAX_CBALLS;
+				}
+			}
 		}
 
 		if (event.getMessage().contains("You pick up the cannon")
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -384,6 +384,15 @@ public class CannonPlugin extends Plugin
 			}
 		}
 
+		if (event.getMessage().startsWith("Your cannon contains"))
+		{
+			Matcher m = NUMBER_PATTERN.matcher(event.getMessage());
+			if (m.find())
+			{
+				cballsLeft = Integer.parseInt(m.group());
+			}
+		}
+
 		if (event.getMessage().startsWith("You unload your cannon and receive Cannonball")
 			|| event.getMessage().startsWith("You unload your cannon and receive Granite cannonball"))
 		{
--- a/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
@@ -105,6 +105,18 @@ public class CannonPluginTest
 	}
 
 	@Test
+	public void addWrongTypeOfCannonballs()
+	{
+		final ChatMessage message = new ChatMessage();
+		message.setType(ChatMessageType.GAMEMESSAGE);
+		message.setMessage("Your cannon contains 20 x Cannonball.<br>You can only add cannonballs of the same kind.");
+
+		plugin.onChatMessage(message);
+
+		assertEquals(20, plugin.getCballsLeft());
+	}
+
+	@Test
 	public void addMaxCannonballs()
 	{
 		final ItemContainer inventory = mock(ItemContainer.class);
--- a/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) 2021, Alexsuperfly <alexsuperfly@users.noreply.github.com>
+ * Copyright (c) 2021, Jordan Atwood <nightfirecat@protonmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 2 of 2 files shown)


Nightfirecat (commented):
Has been mentioned that this should do an inventory diff to find out how many balls have been loaded.

Nightfirecat (changes requested):
Also the first and second commits can be squashed? The third commit should remain separate.

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -198,6 +206,26 @@ public void onItemContainerChanged(ItemContainerChanged event)
...
+		if (inventorySnapshot != null)
+		{
+			Multiset<Integer> diff = generateInventoryDiff();
+			int snapshotCannonballs = inventorySnapshot.count(ItemID.CANNONBALL);
+			int diffCannonballs = diff.count(ItemID.CANNONBALL);
+			int snapshotGraniteCannonballs = inventorySnapshot.count(ItemID.GRANITE_CANNONBALL);
+			int diffGraniteCannonballs = diff.count(ItemID.GRANITE_CANNONBALL);
+
+			if ((diffCannonballs > 0 && snapshotCannonballs > diffCannonballs)
+					|| (diffGraniteCannonballs > 0 && snapshotGraniteCannonballs > diffGraniteCannonballs))
```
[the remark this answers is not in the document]
Alexsuperfly: good shout, i think this can actually just become `diff.contains(ItemID.CANNONBALL) && inventory.contains(ItemID.CANNONBALL)` then, since all that matters is the diff and current inventory both contain cannonballs in them

Nightfirecat:
Do you know offhand what test cases you were testing ingame to make sure these changes worked? I'd like to add tests to this.

Nightfirecat:
I've added a test for your second commit. I'm happy to write a test for your first commit as well (and maybe the rest of the plugin while I'm at it) if you can share the steps you took to test this ingame

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -24,9 +24,14 @@
  */
 package net.runelite.client.plugins.cannon;
 
+import com.google.common.annotations.VisibleForTesting;
```
Cyborger1, line 27: It appears this line is causing a CheckStyle violation. Did you intend to add this annotation somewhere?

Alexsuperfly:
> I've added a test for your second commit. I'm happy to write a test for your first commit as well (and maybe the rest of the plugin while I'm at it) if you can share the steps you took to test this ingame

The process to test each case would be:

- place the cannon in KQs lair
- load with cannonballs and start firing
- KQ destroys the cannon
- reclaim the cannon from Nulodion
- setup the cannon in barb village (dwarfs dont let you place it next to them)

each test case is:

1. load regular, set back up with excess regular in invent, observe cannon reading 30
2. 1 but with granite cannonballs
3. load regular, set back up with <= amount of space in cannon of balls in invent, observe only the loaded amount displayed (this is a case i dont believe can be solved since the client can be shut down between destruction and claim, so theres no way to preserve the known number of cannonballs in the cannon)
4. 3 but with granite cannonballs
5. load regular, set back up with any amount of granite cannonballs in invent, observe the "you cant mix cannonballs message" and the count is set correctly to the message that displays how many balls are loaded
6. load granite, set back up with only regular in invent, same as 5

Nightfirecat:
Ok, I've added tests which cover the added functionality in your commits. I've also added a commit to remember the previous cannonball count when it is destroyed and later set up. I think this is probably all set now?

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -204,6 +204,7 @@ public void cannonDestroyedSetupNormalCannonballs()
 		assertEquals(30, plugin.getCballsLeft());
```
Alexsuperfly, line 204: after this commit these arnt really testing the inventory diff system since the message will add 5 and 25 to equal 30 regardless of what the inventory diff decides the contents are

Nightfirecat:
Your point on the tests is taken. I've incorporated your recommended changes (along with some further changes on my part to fix up these commits) and I think it's all set now.

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -171,6 +179,145 @@ public void unloadCannon()
...
+		final ItemContainer inventory = mock(ItemContainer.class);
+		final Item[] inventoryContentsBefore = { new Item(ItemID.CANNONBALL, 100), new Item(ItemID.GRANITE_CANNONBALL, 100) };
+		when(inventory.getItems()).thenReturn(inventoryContentsBefore);
+		when(inventory.contains(ItemID.CANNONBALL)).thenReturn(true);
+		lenient().when(inventory.contains(ItemID.GRANITE_CANNONBALL)).thenReturn(true);
+		when(client.getItemContainer(InventoryID.INVENTORY)).thenReturn(inventory);
+
+		plugin.onChatMessage(loadCannonballs(25));
+
+		final Item[] inventoryContentsAfter = { new Item(ItemID.CANNONBALL, 75), new Item(ItemID.GRANITE_CANNONBALL, 100) };
```
Alexsuperfly, line 220: This doesn't really follow in game behavior since it will always load granite cannonballs if they exist in your inventory, no matter what, but for the purposes of testing code paths this could be fine.

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -300,15 +327,20 @@ public void onChatMessage(ChatMessage event)
 		{
 			cannonPlaced = true;
 			addCounter();
+			justPlaced = true;
```
Adam-, line 330: From testing in game, it seems like it's not possible to add the furnace but also not load it if you have cballs in your inventory. So I think you could just check if there are >= 30 cballs here and if so set the cballs left to 30 then. Then you could avoid the whole inv snapshot thing. That would fix most of the scenarios except if you had a partially-filled broken cannon and also <30 cannon balls to re-load it when replacing.

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -0,0 +1,172 @@
...
+	@Bind
+	private Client client;
+
+	@Mock
+	@Bind
+	private OverlayManager overlayManager;
+
+	private static final ChatMessage ADD_FURNACE = new ChatMessage();
+
+	static
```
Adam-, line 93: Prefer `@BeforeClass` - since if this fails it will hard fail the test runner.

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -301,6 +302,23 @@ public void onChatMessage(ChatMessage event)
 			cannonPlaced = true;
 			addCounter();
 			cballsLeft = 0;
+
+			final ItemContainer inventory = client.getItemContainer(InventoryID.INVENTORY);
+			if (inventory != null)
+			{
+				int invCballs = inventory.count(ItemID.GRANITE_CANNONBALL) > 0
+						? inventory.count(ItemID.GRANITE_CANNONBALL)
+						: inventory.count(ItemID.CANNONBALL);
```
maxbla, line 311: I think this logic causes a bug if the player's cannon is destroyed when containing normal cannonballs, then they bring 30+ granite cannonballs and no normal cannonballs the next time they place it. I can't test it because I don't have any granite cannonballs right now.

Alexsuperfly: I dont see how 5bfbff492d4bff1f40aeb3cd411f6766b6a6ba67 doesn't address that.

maxbla: 5bfbff4 does address that. Sorry for the mistake.

PR #13109, opened 2021-01-25
Add world to friend login message

### LINKED ISSUE #13079: Add world friend has logged into in the login message
Show the world a friend has logged into in the notification message. "Friend has logged into World XX."

mlewicki12:
Closes #13079

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/friendlist/FriendListPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/friendlist/FriendListPlugin.java
@@ -25,19 +25,26 @@
  */
 package net.runelite.client.plugins.friendlist;
 
+import com.google.inject.Provides;
 import javax.inject.Inject;
+import net.runelite.api.ChatMessageType;
+import net.runelite.api.ChatPlayer;
 import net.runelite.api.Client;
 import net.runelite.api.Friend;
 import net.runelite.api.Ignore;
+import net.runelite.api.MessageNode;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/friendlist/FriendListPlugin.java
```diff
@@ -109,6 +128,25 @@ else if (event.getScriptId() == ScriptID.IGNORE_UPDATE)
 		}
 	}
 
+	@Subscribe
+	public void onChatMessage(ChatMessage message)
+	{
+		if (config.showWorld() && message.getType() == ChatMessageType.LOGINLOGOUTNOTIFICATION)
+		{
+			// get the player name out of the notification
+			String name = message.getMessage().split(" ")[0];
```
Adam-, line 137: Did you test this on a player with a space in their name? It might actually work if the message uses non-breaking spaces for the player name, and also the `findByName` handles it, but I do not remember.

mlewicki12: I just tested it, works with spaces, but I changed the way of getting name to substring instead of split just to be safe.

Adam-: This isn't much better since if the old approach would break this new approach would also break if the user has ` has` in their name. So you could just use the old approach.

Nightfirecat (changes requested):

PR #13111, opened 2021-01-25
Avoid Javadoc cutting off descriptions

loldudester:
Using a period before cases of `(ie.` to avoid including them in summaries


Nightfirecat:
Added in d68d66b01d6c29f16f8fd50b1a1f022a7066f496

PR #13114, opened 2021-01-26
Boosts Plugin: Update notification to match idle notifier

brad-richardson:
See https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java#L436


Adam-:
We've ended up just adding the player name to the notification title globally since it is used almost everywhere currently.

PR #13128, opened 2021-01-30
Add !cats command, showing traded cats, obtained runes and lost kittens

Cyborger1:
Adding this as a draft until I can figure out how to test the command properly. I mostly followed the way `!duels` got implemented for the http-service part.

Adds "kc" tracking for Cats traded at West Ardougne, Death Runes obtained from said trading and kittens lost from not feeding/stroking. The `!cats` command would display this information to other players.

This also may or may not need some tests implemented.

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/chat/Cats.java
+++ b/http-api/src/main/java/net/runelite/http/api/chat/Cats.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2021, Cyborger1
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


Nightfirecat:
If we had some kind of API for allowing chat commands for hub plugins, this seems like it would be fine there. I don't think we care about adding this to the main client though?

Cyborger1:
I would have definitely done this as a hub plugin if I could, I was mostly doing it for fun, really.

I do agree there's not much of a point for having this in Core, feel free to close without merging I suppose.

Nightfirecat:
Seems like that'd be a good feature request to raise. For now I'll close this PR though

PR #13134, opened 2021-02-01
Add ConfigEnum interface for Tooltips on Config ComboBoxes

Cyborger1:
Adds an interface that allows you to explicitly specify a DisplayName and a Tooltip for an Enum type.

tooltip


Nightfirecat:
I'd prefer https://github.com/runelite/runelite/commit/41eff1b4ca9d9ed0deefd36689992bf8e5e39f7e given we use `Tooltip` over `ToolTip` elsewhere in this project

Cyborger1:
Ok, I'll change that back. I had switched it to "ToolTip" to make it consistent with how the `java.awt` classes call it.

Adam-:
Is there something you'd intended to use this for..? Adding tooltips to existing combo boxes?

Cyborger1:
This is to add tooltips to the individual items in the combo box, like so, for example.

```
@Getter
@AllArgsConstructor
public enum ControlFunction implements TooltipProvider
{
	NONE("None", "No special behavior from pressing Control"),
	CONTROL_TO_ZOOM("Hold to zoom", "Zoom only enabled while Control is held"),
	CONTROL_TO_RESET("Reset zoom", "Press Control to reset zoom to default");

	private final String name;
	private final String tooltip;
	
	@Override
	public String getTooltipText()
	{
		return name + ": " + tooltip;
	}

	@Override
	public String toString()
	{
		return getName();
	}
}

```

Adam-:
I get that, I meant do you intend to actually use it? Since this is unused.

Cyborger1:
I do intend to use it in my plugins, but there's also the possibility to go back to the Core plugins and add tooltip text for enum fields that might need more clarification.

Cyborger1:
I figured adding tooltip text to plugins following this would have been better suited to be in a new PR.

PR #13143, opened 2021-02-03
slayer: Modify to use RSProfiles Config

### LINKED ISSUE #249: Slayer plugin - Task carrying over to other accounts

Cyborger1:
Modifies the Slayer Plugin to use the newly added RSProfiles feature to ConfigManager. This closes #249.

The website actually pulls the info from the old config fields though.
I'm not sure what needs to be done on that end. Someone who knows his way around the rl.net code would have to help with this.
With that said, this shouldn't be merged until the website can support the new config path.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -140,6 +140,9 @@ public class SlayerPlugin extends Plugin
 	private SlayerConfig config;
 
 	@Inject
+	private ConfigManager configManager;
+
+	@Inject
 	private OverlayManager overlayManager;
 
 	@Inject
@@ -227,12 +230,17 @@ public class SlayerPlugin extends Plugin
 		{
 			cachedXp = client.getSkillExperience(SLAYER);
 
-			if (config.amount() != -1
-				&& !config.taskName().isEmpty())
+			migrateConfig();
+
+			if (getIntProfileConfig(SlayerConfig.AMOUNT_KEY) != -1
+				&& !getStringProfileConfig(SlayerConfig.TASK_NAME_KEY).isEmpty())
 			{
-				setExpeditiousChargeCount(config.expeditious());
-				setSlaughterChargeCount(config.slaughter());
-				clientThread.invoke(() -> setTask(config.taskName(), config.amount(), config.initialAmount(), config.taskLocation(), false));
+				setExpeditiousChargeCount(getIntProfileConfig(SlayerConfig.EXPEDITIOUS_CHARGES_KEY));
+				setSlaughterChargeCount(getIntProfileConfig(SlayerConfig.SLAUGHTER_CHARGES_KEY));
+				clientThread.invoke(() -> setTask(getStringProfileConfig(SlayerConfig.TASK_NAME_KEY),
+					getIntProfileConfig(SlayerConfig.AMOUNT_KEY),
+					getIntProfileConfig(SlayerConfig.INIT_AMOUNT_KEY),
+					getStringProfileConfig(SlayerConfig.TASK_LOC_KEY), false));
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -919,7 +932,8 @@ public void testNewAccountSlayerKill()
 	@Test
 	public void infoboxNotAddedOnLogin()
 	{
-		when(slayerConfig.taskName()).thenReturn(Task.BLOODVELD.getName());
+		when(slayerPlugin.getStoredTaskName()).thenReturn(Task.BLOODVELD.getName());
+		when(slayerPlugin.getStoredAmount()).thenReturn(50);
```
Nightfirecat, line 936: Why is the amount set in this test?

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -283,7 +291,8 @@ public void testBossTask()
 
 		assertEquals("Vet'ion", slayerPlugin.getTaskName());
 		assertEquals(3, slayerPlugin.getAmount());
-		verify(slayerConfig).points(914);
+
```
Nightfirecat, line 294: These added newlines in this file can be removed

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -275,27 +292,132 @@ public void onGameStateChanged(GameStateChanged event)
...
+		setIntConfig(SLAUGHTER_CHARGES_KEY, slaughterCharges);
+	}
+
+	private int getIntConfig(String key)
+	{
+		Integer value = configManager.getRSProfileConfiguration(SlayerConfig.GROUP_NAME, key, int.class);
+		return value == null ? -1 : value;
+	}
+
+	private void setIntConfig(String key, int value)
```
Nightfirecat, line 397: This is identical to `setStringConfig` except for the signature. Change the signature to `(String key, Object value)` and direct all setters to use it.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -275,27 +292,132 @@ public void onGameStateChanged(GameStateChanged event)
...
+						getStoredAmount(),
+						getStoredInitialAmount(),
+						getStoredTaskLocation(), false);
 					loginFlag = false;
 				}
 				break;
 		}
 	}
 
+	public int getStoredAmount()
```
Nightfirecat, line 311: All getters should be package-private, and setters should be private.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -133,12 +133,25 @@
...
 
+	// Key names for stored task values
+	public static final String TASK_NAME_KEY = "taskName";
+	public static final String AMOUNT_KEY = "amount";
+	public static final String INIT_AMOUNT_KEY = "initialAmount";
+	public static final String TASK_LOC_KEY = "taskLocation";
+	public static final String STREAK_KEY = "streak";
+	public static final String POINTS_KEY = "points";
+	public static final String EXPEDITIOUS_CHARGES_KEY = "expeditious";
+	public static final String SLAUGHTER_CHARGES_KEY = "slaughter";
```
Nightfirecat, line 144: These should be package-private and `@VisibleForTesting`

Nightfirecat (changes requested):

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
Cyborger1: I figured since this tests [this code branch](https://github.com/runelite/runelite/blob/a58e441f9a236eff8b2012761189f1fe4b6f9db7/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java#L295) I would set an arbitrary amount value along with the name. If this defaults to 0 otherwise I can remove it.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
Cyborger1: Thank you, I forgot about that annotation.

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
Cyborger1: I fiddled around with this some more. The amount check, and `when(slayerConfig.showInfobox()).thenReturn(true);`, which I'll be adding, are both required to properly test this behavior.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -133,12 +133,33 @@
 	private static final Pattern TASK_STRING_VALIDATION = Pattern.compile("[^a-zA-Z0-9' -]");
 	private static final int TASK_STRING_MAX_LENGTH = 50;
 
+	// Key names for stored task values
+	@VisibleForTesting
+	static final String TASK_NAME_KEY = "taskName";
```
abextm, line 138: these should go in the Config

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -275,27 +300,127 @@ public void onGameStateChanged(GameStateChanged event)
...
+						getStoredAmount(),
+						getStoredInitialAmount(),
+						getStoredTaskLocation(), false);
 					loginFlag = false;
 				}
 				break;
 		}
 	}
 
+	int getStoredAmount()
```
abextm, line 319: naming these like getters/setters is somewhat confusing. Its probably clearest for them to be replaced with calls directly to configmanager

Cyborger1: I agree we could change the name, but I'm not sure replacing these with direct calls to the manager would be clearer. The call can get quite long to type out and the getters abstract away a null check to keep the behavior consistent with the old config setup.

abextm: Just `getIntConfig(SlayerConfig.WHATEVER)` seems fine as it resolves the naming problem.

Cyborger1: Thank god for ctrl-alt-n

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -275,27 +281,47 @@ public void onGameStateChanged(GameStateChanged event)
...
+						getIntProfileConfig(SlayerConfig.AMOUNT_KEY),
+						getIntProfileConfig(SlayerConfig.INIT_AMOUNT_KEY),
+						getStringProfileConfig(SlayerConfig.TASK_LOC_KEY), false);
 					loginFlag = false;
 				}
 				break;
 		}
 	}
 
+	int getIntProfileConfig(String key)
```
Nightfirecat, line 300: These should be `@VisibleForTesting` as they won't be used elsewhere.

Nightfirecat (commented):
LGTM otherwise at a glance

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
Cyborger1: Done 👍

abextm (approved):

Nightfirecat (approved):

runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
```diff
@@ -994,4 +1006,13 @@ public void testMultikill()
 		assertEquals("Suqahs", slayerPlugin.getTaskName());
 		assertEquals(229, slayerPlugin.getAmount()); // 2 kills
 	}
+
+	// Helper method for ConfigManager mock
+	// Last call to the manager needs to have the expected value
+	private void verifyIntConfig(String key, Integer expectedValue)
+	{
+		ArgumentCaptor<Integer> captor = ArgumentCaptor.forClass(int.class);
+		verify(configManager, atLeastOnce()).setRSProfileConfiguration(eq(SlayerConfig.GROUP_NAME), eq(key), captor.capture());
```
Adam-, line 1015: Why was this changed to `atLeastOnce()`?

Cyborger1: In case the config gets set more than once, the only thing that matters is that the last call is of the expected value. Without `atLeastOnce()`, the captor serves no purpose and could just be a full equality check as the default behavior is that the call needs to happen exactly once.

PR #13156, opened 2021-02-04
Added two new cow emojis

Woodse07:
Emojis taken from Twemoji.


makingcomment:
... Why?

Also: Those triggers aren't very cow-like. I could definitely see someone typing them and not expecting a cow replacement.

dekvall (changes requested):
The client can't handle semi-transparent pixels. I have a twemoji repo you can get them from or you can modify these and threshold alpha at 50%

Woodse07:
> The client can't handle semi-transparent pixels. I have a twemoji repo you can get them from or you can modify these and threshold alpha at 50%

Cool thanks, I've changed them to use the emojis from your fork.

Woodse07:
> ... Why?
> 
> Also: Those triggers aren't very cow-like. I could definitely see someone typing them and not expecting a cow replacement.

There's gorillas, acorns and aliens, why not a cow ¯\_(ツ)_/¯

Do you have any better ideas in mind for the triggers?

dekvall (approved):

dekvall:
Needs rebase.

Nightfirecat:
I don't think we really need both cow emojis--just one would be fine, though I'm not sure which would be more immediately recognizable from the image.

Issue #13160, opened 2021-02-06
Chat commands plugin NPE


Nightfirecat:
```
2021-02-05 22:13:17 [Client] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
java.lang.NullPointerException: null
	at net.runelite.client.plugins.chatcommands.ChatCommandsPlugin.onGameTick(ChatCommandsPlugin.java:484)
	at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:73)
	at net.runelite.client.eventbus.EventBus.post(EventBus.java:222)
	at net.runelite.client.callback.Hooks.clientMainLoop(Hooks.java:179)
	at client.qb(client.java:47834)
	at client.av(client.java)
	at bd.zq(bd.java:377)
	at bd.run(bd.java:331)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Adam-:
fixed in #13227

PR #13166, opened 2021-02-07
fix NPE in onGameTick function

RyanAFinney:
Fix NPE mentioned in Issue 13160.

https://github.com/runelite/runelite/issues/13160


Nightfirecat (changes requested):
This is not quite correct--`getChild()` should be passed the text index:

```java
Widget exploitsText adventureLog.getChild(ADV_LOG_EXPLOITS_TEXT_INDEX);
if (exploitsText != null)
{
	Matcher advLogExploitsText = ADVENTURE_LOG_TITLE_PATTERN.matcher(adventureLog.getChild(ADV_LOG_EXPLOITS_TEXT_INDEX).getText());
	if (advLogExploitsText.find())
	{
		pohOwner = advLogExploitsText.group(1);
	}
}
```

Moreover, this could be solved in a better way: it's possible this interface is being created via some unique client script. If that is the case, we should run this code in an `onScriptPostFired` subscriber checking for that script being executed.

RyanAFinney:
Thanks. This was my first stab at something on runelite.. been using the client for a while and thought I'd try giving back a little. Will look more into the issue and regroup.

PR #13167, opened 2021-02-07
Adds streak and points to !task command. Adds !slayer as an alias.

### LINKED ISSUE #13116: !streak - to display your current slayer streak (even if you don't currently have a task)
Hi,

Title explains it pretty much. It can be annoying having not done slayer for a while, having no task, and not knowing whether to use Konar or regular master.

The 'Slayer' plugin does keep track of your task streak (but only appears when you're doing a task). So I imagine the data is in there somewhere, would just be nice to be able to pull it any time.

williameschmidt:
Resolves #13116 

Example
Example

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
+++ b/http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
@@ -128,7 +128,7 @@ public class ChatClient
 		}
 	}
 
-	public boolean submitTask(String username, String task, int amount, int initialAmount, String location) throws IOException
+	public boolean submitTask(String username, String task, int amount, int initialAmount, String location, int streak, int points) throws IOException
 	{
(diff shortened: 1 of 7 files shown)


williameschmidt:
What's a man gotta do to get this merged

Adam-:
I guess I don't feel too strongly either way about this (although I dislike the alias, and that the alias doesn't work when !slayer is used with no task). And also the output of the existing command can be quite long if it includes a Konar task, and this seems just like extra info that just clutters it up more.

williameschmidt:
Would you be more inclined to merge it if I changed the behavior of the alias? I'm fine with just closing the PR too if you think the extra fields aren't necessary.

Adam-:
I'm fine with closing it if you want since these new fields aren't even really task related. They exist independent of the task (you can have no task but also have a streak/points). But usually if I'm unsure on something I just leave it open.

williameschmidt:
Rip my beautiful PR

PR #13176, opened 2021-02-07
pestcontrol: add widget definitions

yani:
Adds some more widget definitions for the Pest Control minigame.

I'd like to make #8096 into a plugin for the plugin hub, but I think these definitions should be pushed to the core client.


abextm:
We don't add things to WidgetInfo for hub plugins since they bitrot easily. You can use the 2 arg variant of getWidget in your plugin.

Issue #13144, opened 2021-02-03
Idle notifier not working in ectofunctus


FelixAbrahamsson:
Idle notifier does not trigger after grinding bones or picking up slime

1. Grind bones in ectofuntus or fill buckets of slime
2. Tab out and wait until idle

**Expected behavior**
Idle notifier should go off but it doesnt.

**Environment**
 - OS: Ubuntu 18.04
 - RuneLite version: 1.6.38
 - Launcher version: 2.1.0

Nightfirecat:
Added in #13177

PR #13177, opened 2021-02-08
Idle notifier: ectofuntus

Auenc:
Idle notifier will now trigger after the following animations:
- Filling buckets of slime
- Emptying bone crusher bin
- Grinding the bones (last animation played if the player has no container for the crushed bones)

Fix for: #13144


Broooklyn:
It's `Ectofuntus` not `Ectofungus`.

Economax:
Is there any way to turn off the animation notifications while keeping on the idle notifier? Because it is very annoying to get notifications for every animation when I'm trying to afk!

Nightfirecat:
No, you should set your delay higher if you're being notified so soon.

PR #13194, opened 2021-02-10
Fix emoji plugin losing chat color

Cyborger1:
Chat messages edited through the emoji plugin would lose their color whenever the chat color cache was rebuilt from changing any chat color because they were `setRuneLiteFormatMessage()` on the message nodes without the `<colNORMAL>` indicator.

oky1unszdT

In this gif, the regular message does not change because it has no runelite format, so the widget's text never gets updated by the reloading of the color cache.

Edit: This is somewhat outdated. We are no longer using runelite formatted messages.


Adam-:
A better way to fix this might be to have the emoji plugin set the message directly instead of the RL-formatted message, since it isn't a RL-formatted message. It would make this use the existing behavior of everything else where it does not retroactively recolor chat messages that aren't RL-formatted messages, which I think is fine.

Cyborger1:
I've changed it to only set message node value instead of using message formats. This does make the change a lot more concise.

PR #13196, opened 2021-02-11
Update CannonSpots.java

### LINKED ISSUE #13195: Cannon spot marker under Mourner HQ is now invalid after song of the elves quest release
After the release of song of the elves, the cannon placement marker that was located underground in the mourner HQ in West Ardougene is now invalid due to being unable to cannon in this spot anymore. If you were to place a cannon anywhere underground within the confines of this building the cannon will not fire therefore the cannon marker spot placed by runelite should be removed. [](url)
mourner_Hq

buttonpoker:
Removed coordinates for the elf/mourner cannon spot located under the mourner hq. After the release of the quest Song of The Elves you are still able to place a cannon down here but the cannon will not fire regardless if you have completed the quest or not therefore the cannon marker is not valid.

Closes #13195

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonSpots.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonSpots.java
@@ -49,7 +49,7 @@ enum CannonSpots
 	DUST_DEVIL(new WorldPoint(3218, 9366, 0)),
 	EARTH_WARRIOR(new WorldPoint(3120, 9987, 0)),
 	ELDER_CHAOS_DRUID(new WorldPoint(3237, 3622, 0)),
-	ELVES(new WorldPoint(2044, 4635, 0), new WorldPoint(3278, 6098, 0)),
+	ELVES(new WorldPoint(3278, 6098, 0)),
 	FIRE_GIANTS(new WorldPoint(2393, 9782, 0), new WorldPoint(2412, 9776, 0), new WorldPoint(2401, 9780, 0), new WorldPoint(3047, 10340, 0)),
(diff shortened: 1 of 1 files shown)


buttonpoker:
Reference to issue opened in regards to this matter #13195

Cyborger1:
You can keyword the issue in your initial PR message like so to automatically link the issue to this PR: `Closes #13195`

Mrnice98:
if you reference the issue in your commit message it will link the issue and pr Ex. "Remove Morner Cannon Spot Closes #13195"

buttonpoker:
> You can keyword the issue in your initial PR message like so to automatically link the issue to this PR: `Closes #13195`

Got it, sorry about that

Nightfirecat:
I mentioned this in Discord, but I'll repeat it here--I don't think this is the right way to solve this. Given that these NPCs are attackable, this sounds like a game bug which Jagex should address rather than a special case we should work around. We should get in touch with a jmod (via twitter or otherwise) and inquire about this behavior.

Cyborger1:
I agree, although I have a hunch this might be intended due to engine limitations with invisible NPCs.

Mrnice98:
> I mentioned this in Discord, but I'll repeat it here--I don't think this is the right way to solve this. Given that these NPCs are attackable, this sounds like a game bug which Jagex should address rather than a special case we should work around. We should get in touch with a jmod (via twitter or otherwise) and inquire about this behavior.

I agree with this.

Cyborger1:
I linked this in the discord, but I'll put this here too.

I reached out to Mod Ash who [confirms this is known](https://twitter.com/JagexAsh/status/1359968976760094721).
MultiNPCs don't play well with the cannon.

PR #13197, opened 2021-02-11
Widgets: Update Widgets to better handle Resizable-Modern and no longer reference Account Management as Ignore

MarbleTurtle:
Updates WidgetID and WidgetInfo's missing or incorrectly labeled items to be more inline with existing items


Nightfirecat:
I wouldn't mind the sorting and renaming to keep names consistent across the three views, however I don't think we want to add more widget definitions which aren't used in the client since those tend to bit rot with interface updates.

PR #13209, opened 2021-02-13
emoji: Add shadow to spoon, fork-and-knife emojis

Nightfirecat:
Without a shadow, these emojis were nearly unreadable on the fixed chat background.
I wouldn't say it looks great now, but it's somewhat readable now, and looks fine in overhead text and on transparent chat backgrounds.

P.S. The image is not resized to allow for a row of shadow pixels below the emoji as that offset from the image resizing looked wrong in chat.


Cyborger1:
I made some alternate utensil emojis if you want to consider these

fork_and_knife
spoon

makingcomment:
Maybe a full border would be better? In the image you posted, the right side looks good but the left side is very hard to see.

Nightfirecat:
I did compare with a full border, and it sadly looks very bad that way

Nightfirecat:
I've replaced the images with those provided by @Cyborger1 as they look significantly better than my proposal

Nightfirecat:
One last update to use a less wonky fork graphic and add shadowing to the utensils similar to the knife:

spoon
fork and knife
cutlery on fixed chat background
cutlery on transparent chat background

PR #13211, opened 2021-02-13
Add new utility methods for checking if world points are inside world areas

pwatts6060:
These utility methods are useful in plugins both for the hub and runelite. I swapped out one of the old ways it was done in the NpcAggroAreaPlugin as an example. I ran some unit tests and the methods seem to functionally be the same as the distance == 0 checks.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/coords/WorldArea.java
+++ b/runelite-api/src/main/java/net/runelite/api/coords/WorldArea.java
@@ -152,6 +152,26 @@ public class WorldArea
 	}
 
 	/**
+	 * Checks whether a tile is contained within the area and in the same plane.
+	 *
+	 * @return {@code true} if the tile is contained within the bounds of this area, {@code false} otherwise.
+	 */
+	public boolean contains(WorldPoint worldPoint)
+	{
+		return distanceTo(worldPoint) == 0;
+	}
+
+	/**
+	 * Checks whether a tile is contained within the area while ignoring the plane.
+	 *
+	 * @return {@code true} if the tile is contained within the bounds of this area regardless of plane, {@code false} otherwise.
+	 */
+	public boolean contains2D(WorldPoint worldPoint)
+	{
+		return distanceTo2D(worldPoint) == 0;
+	}
+
+	/**
 	 * Checks whether this area is within melee distance of another.
 	 * <p>
 	 * Melee distance is exactly 1 tile, so this method computes and returns
--- a/runelite-api/src/main/java/net/runelite/api/coords/WorldPoint.java
+++ b/runelite-api/src/main/java/net/runelite/api/coords/WorldPoint.java
@@ -406,4 +406,40 @@ public class WorldPoint
 		}
 		return worldPoint;
 	}
+
+	/**
+	 * Checks whether this tile is located within any of the given areas.
+	 *
+	 * @param worldAreas areas to check within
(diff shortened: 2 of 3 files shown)


Nightfirecat (commented):
Isn't this the same as `WorldPoint.distanceTo(WorldArea other) == 0`?

pwatts6060:
> Isn't this the same as `WorldPoint.distanceTo(WorldArea other) == 0`?

Yes it serves the same purpose but less verbose and can do checks for multiple areas rather than chaining them with ors.

Nightfirecat:
> ... and can do checks for multiple areas rather than chaining them with ors.

I don't understand this at all. Can you give an example where this proposed API is better than what I posted?

pwatts6060:
> > ... and can do checks for multiple areas rather than chaining them with ors.
> 
> I don't understand this at all. Can you give an example where this proposed API is better than what I posted?
People search the api to find a method for if a point is contained within an area. With the current api you won't find anything you want for that, most people won't immediately realize they can do the distance == 0 check. You shouldn't need to get the distance and compare it with 0 though. 

It's a lot more readable and often you need to check if a point is in any areas such as the one I swapped out in the NpcAggroAreaPlugin. I'm sure it will also get lots of use for plugins on the hub. e.g.
`inLobby = client.getLocalPlayer().getWorldLocation().distanceTo(lmsCompetitiveLobby) == 0
			|| client.getLocalPlayer().getWorldLocation().distanceTo(lmsCasualLobby) == 0
			|| client.getLocalPlayer().getWorldLocation().distanceTo(lmsHighStakesLobby) == 0;`
vs 
`inLobby = client.getLocalPlayer().getWorldLocation().inArea(lmsCompetitiveLobby, lmsCasualLobby, lmsHighStakesLobby);`

runelite-api/src/main/java/net/runelite/api/coords/WorldArea.java
```diff
@@ -88,6 +88,24 @@ public WorldArea(WorldPoint location, int width, int height)
 		this.height = height;
 	}
 
+	/**
+	 * @return true if the WorldPoint is contained within the bounds of this area its plane, false otherwise.
+	 */
+	public boolean contains(WorldPoint worldPoint)
+	{
+		return plane == worldPoint.getPlane()
+			&& contains2D(worldPoint);
```
Nightfirecat, line 97:
```suggestion
		return distanceTo(worldPoint) == 0;
```

runelite-api/src/main/java/net/runelite/api/coords/WorldArea.java
```diff
@@ -88,6 +88,24 @@ public WorldArea(WorldPoint location, int width, int height)
...
+	}
+
+	/**
+	 * @return true if the WorldPoint is contained within the bounds of this area regardless of plane, false otherwise.
+	 */
+	public boolean contains2D(WorldPoint worldPoint)
+	{
+		return worldPoint.getX() >= x && worldPoint.getX() <= x + width
+			&& worldPoint.getY() >= y && worldPoint.getY() <= y + height;
```
Nightfirecat, line 106:
```suggestion
		return distanceTo2D(worldPoint) == 0;
```

Nightfirecat (changes requested):
Also please write full javadoc, eg. description, `@param`(s), and `@return`

pwatts6060:
@Nightfirecat Added javadocs and commited your changes

PR #13212, opened 2021-02-13
Move wiki varbit crowdsourcing into default client

andmcadams:
This is a slight modification of varbit tracking found here https://github.com/leejt/osrs-wiki-crowdsourcing. The CrowdsourcingVarbits class is itself a modified version of [Abex's VarInspector devtool](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/devtools/VarInspector.java).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/varbits/CrowdsourcingVarbits.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/varbits/CrowdsourcingVarbits.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018 Abex
+ * Copyright (c) 2021 andmcadams
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -24,10 +25,10 @@
  */
 package net.runelite.client.plugins.crowdsourcing.varbits;
 
+import com.google.common.collect.ImmutableSet;
 import net.runelite.client.plugins.crowdsourcing.CrowdsourcingManager;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
-import java.util.HashSet;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/varbits/CrowdsourcingVarbits.java
```diff
@@ -0,0 +1,193 @@
...
+	{
+		/* Wait a tick before grabbing location.
+		 *
+		 * This seems to cause fewer issues than not waiting a tick.
+		 * Only noticeable issues seem to be in places like Dorgesh-Kaan, where you can go up and down
+		 * stairs quickly to trick it into thinking varbs were updated in a different location.
+		 * Without waiting a tick, certain loads/instances are messed up, including the Dorgesh-Kaan light
+		 * varbs when loading a part of the map you haven't been to.
+		 */
+		clientThread.invokeLater(() ->
```
Nightfirecat, line 142: This doesn't delay the invoked runnable by a game tick, it delays it by a client tick. (See `Constants.CLIENT_TICK_LENGTH` and `Constants.GAME_TICK_LENGTH`) Hence I don't believe this comment is accurate.

andmcadams: The issue seems to happen only on certain plane changes. If I go to Dorgesh Kaan ground floor, then go up a floor, the varb changes recorded get the worldpoint of the ground floor instead of the previous plane. This only seems to happen the first time I load the floor though. Waiting for the next client tick seems to solve this problem. But yeah, I definitely was under the impression this was waiting a game tick rather than a client tick, so I can clean up that comment.

runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/varbits/CrowdsourcingVarbits.java
```diff
@@ -93,6 +93,10 @@ public void startUp()
 			{
 				return false;
 			}
+			if (client.getGameState() == GameState.STARTING || client.getGameState() == GameState.UNKNOWN)
```
andmcadams, line 96: This feels like a hack, but I'm not sure how to tell if `client.getIndexConfig()` will throw an NPE otherwise.

andmcadams: Oh, it's actually happening in `client.getVarbit`, not `client.getIndexConfig`. I can't see this part of the code afaik, so not sure what exactly is happening, but it's not the `client` that's null. Here's the exception if I remove the gamestate check and run it:
```
2021-06-06 12:27:56 [Client] WARN  n.r.client.callback.ClientThread - Exception in invoke
oz: null
	at fh.n(fh.java:103)
	at mv.n(mv.java:31)
	at ge.v(ge.java:24)
	at client.bp(client.java)
	at client.xd(client.java:43892)
	at client.getVarbit(client.java:59445)
	at net.runelite.client.plugins.crowdsourcing.varbits.CrowdsourcingVarbits.lambda$startUp$0(CrowdsourcingVarbits.java:100)
	at net.runelite.client.callback.ClientThread.invoke(ClientThread.java:99)
	at net.runelite.client.callback.Hooks.clientMainLoop(Hooks.java:195)
	at gg.zc(gg.java:9184)
	at client.ad(client.java)
	at af.ix(af.java:395)
	at af.run(af.java:349)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NullPointerException: null
	at mv.n(mv.java:32)
	... 12 common frames omitted
```

PR #13217, opened 2021-02-14
Add minimap arrow pointing to the active daeyalt rock

### LINKED ISSUE #12314: Daeyalt rock minimap indicator
To my knowledge, this feature/plugin would either be very straightforward to do or near impossible. I would like to see a feature implemented that would show on the minimap which of the three rocks in the Daeyalt mines is currently active. There are circumstances (most notably 1.5t daeyalt mining) in which it is not feasible to have all the rocks visible on the screen. When the rock the character is mining expires, it's not reasonable to path straight to the next active rock without knowing which it may be, and it's a lot of scrolling very regularly just to still have suboptimal pathing overall. This could potentially be something expandable to other features, and to my knowledge would be withing the third party client guidelines handed down by Jagex.

NJohnsonJ:
Closes #12314
Example

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
@@ -39,6 +39,7 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameObject;
 import net.runelite.api.GameState;
+import static net.runelite.api.HintArrowType.WORLD_POSITION;
 import net.runelite.api.MenuAction;
 import static net.runelite.api.ObjectID.DEPLETED_VEIN_26665;
 import static net.runelite.api.ObjectID.DEPLETED_VEIN_26666;
@@ -133,6 +134,7 @@ public class MiningPlugin extends Plugin
 		pickaxe = null;
 		overlayManager.remove(overlay);
 		overlayManager.remove(rocksOverlay);
+		respawns.forEach(respawn -> clearHintArrowAt(respawn.getWorldPoint()));
 		respawns.clear();
 	}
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -247,6 +247,7 @@ public void onGameObjectDespawned(GameObjectDespawned event)
 			{
 				final WorldPoint point = object.getWorldLocation();
 				respawns.removeIf(rockRespawn -> rockRespawn.getWorldPoint().equals(point));
+				client.clearHintArrow();
```
Nightfirecat, line 250: This should check that the hint arrow being cleared is at the same point as the despawned essence via `Client#getHintArrowType()` and `Client#getHintArrowPoint()`. It's possible that the hint arrow was set by the client or by another plugin, so we should not presume that this plugin is the only thing setting it.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -256,6 +258,12 @@ public void onGameObjectDespawned(GameObjectDespawned event)
 		}
 	}
 
+	private void clearDaeyaltHintArrow(GameObject object)
+	{
+		if (client.getHintArrowType() == WORLD_POSITION && client.getHintArrowPoint().equals(object.getWorldLocation()))
+			client.clearHintArrow();
```
Nightfirecat, line 264: This project avoids the use of one-line `if` statements. Add braces to the `clearHintArrow()` call.
```suggestion
		if (client.getHintArrowType() == WORLD_POSITION && client.getHintArrowPoint().equals(object.getWorldLocation()))
		{
			client.clearHintArrow();
		}
```

Nightfirecat (changes requested):
The only other issue I noticed is that shutting down the plugin does not clear the hint arrow if it is set. You should be able to call `clearDaeyaltHintArrow()` on the contents of `respawns` before it is cleared to resolve that.

Nightfirecat:
I've pushed a commit which cleans up this PR to apply with fewer deletions than 6ea82e7 and clean up the flow a bit.

NJohnsonJ:
Is anything else required of me to get this merged?

PR #13225, opened 2021-02-16
move slayer bracelet charge tracking to item charges

### LINKED ISSUE #2407: Add Slaughter/Exped Brace to Indicators

Hydrox6:
Closes #2407
Closes #7902
Closes #8090
Closes #12558

Simply moves the tracking for the slayer bracelets over to Item Charges, allowing for infoboxes. All that remains in the slayer plugin is what's needed to correctly modify the task value.


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -469,4 +469,88 @@ default int chronicle()
...
+	@ConfigItem(
+		keyName = "braceletOfSlaughter",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default int braceletOfSlaughter()
+	{
+		return -1;
+	}
```
Nightfirecat, line 494: These should be set using `setRSProfileConfiguration`

Hydrox6: To be honest, the entire plugin should probably be moved over to use RSProfiles for shared tracking now they exist. I think I'll PR that first instead, then change this PR later after that one has been accepted in some way.

Hydrox6:
Updated to work alongside the rest of the Item Charges plugin, which is now using RSProfile

deathbeam (approved):

PR #13228, opened 2021-02-17
Add new LMS regions

Broooklyn:
Adds the new Wild Varrock LMS regions to Discord as well as the loot tracker region blacklist to prevent player loots received in the new map from appearing in the loot tracker.

Also adds a missing Misthalin discord region because I just happened to noticed it.


Broooklyn:
As an alternative, in [today's game update blog](https://secure.runescape.com/m=news/8th-birthday-event-and-lms-changes?oldschool=1), Jagex said "The default Last Man Standing map has been renamed ‘Deserted Island,’" which gives us the opportunity to split the regions into `LMS - Deserted Island` and `LMS - Wild Varrock` or something to that regard. Thoughts?

PR #13230, opened 2021-02-18
ProgressBar: Remove unused maximumValue field, fix center label centering

### LINKED ISSUE #13115: ProgressBar component horizontal alignment is dependent on left and right label size.
The ProgressBar component horizontal alignment is dependent on left and right label size. This means that if one sides label is significantly larger then the other, the alignment of the 'centerLabel' will be off and not truly in the center of the progress bar.

Examples:
Center with same size label. 

Off-center with different size labels.

Nightfirecat:
BorderLayout does not guarantee equal sizing of its children, so having
a component centered between two components of unequal sizes will leave
it centered relative to their sizes, such as the following:

```
|-----BorderLayout-------------------------------------------|
||----WEST----||------CENTER------||----------EAST----------||
||------------||------------------||------------------------||
|------------------------------------------------------------|
```

Conversely, GridLayout uses fixed, equal-width columns for its children,
ensuring the center component will always occupy the center third of the
parent's dimensions.

P.S. This change causes labels to truncate earlier as they do not expand
beyond a third of the parent container's dimensions. Because the use
case for this component is with small labels, I do not anticipate this
will cause breakage.

Fixes #13115

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/components/ProgressBar.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/components/ProgressBar.java
@@ -24,10 +24,10 @@
  */
 package net.runelite.client.ui.components;
 
-import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics;
+import java.awt.GridLayout;
 import java.util.Collections;
 import java.util.List;
 import javax.swing.JLabel;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Some screenshots demonstrating the change
| Before | After |
|:------:|:-----:|
| |  |
| - |  |

Adam-:
@Nightfirecat  I think this is being used now with the new party panel stuff

Nightfirecat:
I've dropped the commit which removed that field.

PR #13234, opened 2021-02-19
key remapping: fix F-Keys remapping while using Keyboard Bankpin

BossHuso:
Closing the bank pin window with ESC shouldn't stop the F-keys from being remapped anymore.


Adam-:
We think this is a Jagex bug which causes it to incorrectly reload the pin interface after it is closed, and have patched around it internally, so it should no longer happen.

PR #13235, opened 2021-02-19
item stats: Add locator orb and dwarven rock cake

Nightfirecat:


makingcomment:
I thought the rock cake's left click only does 1 damage. Maybe that should be displayed too?
Or did Jagex swap Eat/Guzzle at some point? 😮

Nightfirecat:
I didn't think the plugin was set up to support multiple stat changes of the same stat, but it apparently works just fine, so I've updated it accordingly.

PR #13239, opened 2021-02-20
Make Integer.class work when deserializing config

BrandtHill:


Nightfirecat:
Why is this useful when you can assign the value given from this method to an `Integer` anyway? eg. `Integer i = stringToObject("1", int.class);`

BrandtHill:
Because `stringToObject("1", Integer.class)` returns a string.

BrandtHill:
More specifically, `boolean.class` and `Boolean.class` are interchangeable as expected, whereas `int.class` and `Integer.class` are not. This works: `int beans = stringToObject("69", int.class)`, and this throws an exception: `int beans = stringToObject("69", Integer.class)`.

PR #13247, opened 2021-02-20
Add item stats in equipment tab

### LINKED ISSUE #12628: Item Stats Plugin: Improve Equipment changes and show in worn equipment tab
I've had a couple of times that my inventory would be full and I was wearing something that I wanted to check the equipment stats of. But because it is not shown in the worn equipment tab you would have to make inventory space and unequal the item to see the item stats overlay. Now I did find out there is a "Always show base stats" Option in the Item stats plugin which is a nice addition but it does not work as expected because it also changes the way how the overlay is shown in the inventory and also, it only shows the base stats in the worn equipment tab now, what I would like to see is how the stats change when you unequip it. So if the "Always show base stats" is toggled off it would show the stats change in the worn equipment tab in the same way it is presented in the inventory tab, so it would be "turned on" on default.

Please note that I am already working on the solution but it is not finished yet as I would like to hear if it is a desired feature/solution and if the "extra improvements/features" described below are okay and fit in this feature request. See in the images below how it would look like:

This is how it currently looks like in the inventory: 
Screenshot 2020-10-11 at 12 50 30
What I also don't like about this is that the weight is described as +2.721 in red, suggesting that when equipping it, it would add weight. My suggestion would be to describe the weight just as 2.721 in red, maybe weight reducing armour shown as e.g. -4 in green as it reduces the weight when equipping it. But to be consistent it should be shown in red +4 (or more including the weight of the item itself) when unequiping it as it actually increases the weight when unequiping. , could these improvements also be fixed in this feature, or should a separate bug be made?

This is how it currently looks in the worn equipment tab:
Screenshot 2020-10-11 at 12 54 42
Note that it only shows the weight currently which is weird

This is how it currently looks in the worn equipment tab with "always show base stats" turned on:
Screenshot 2020-10-11 at 12 56 33
This is how I like the weight should be shown, but than maybe in red

This is how it currently looks in the inventory  tab with "always show base stats" turned on:
Screenshot 2020-10-11 at 12 58 26

My suggestion is to make the worn equipment tab consistent with the inventory tab, when the "always show base stats" is turned off it would look like this:
Screenshot 2020-10-11 at 13 00 27
Note that the weight is not right in this image yet, see suggestion above.
Note that the stats are reversed because you are "removing" the equipment
Note that speed is only -2 because the base attack speed Is 4 and the attack speed of the battleaxe is 6

When the "always show base stats" is turned on it would look like this:
Screenshot 2020-10-11 at 13 01 01
Note that the weight is a little bit better here, but should be red, see suggestion above.

NJohnsonJ:
Resolves #12628 
* Add unit tests for new methods and to test the 2H logic
* Fix and simplify the 2H logic
* Add config option to turn off equipment stats in the equipment tab
* Add javadoc for buildItemTooltip

### DIFF
--- a/http-api/src/main/java/net/runelite/http/api/item/ItemStats.java
+++ b/http-api/src/main/java/net/runelite/http/api/item/ItemStats.java
@@ -30,6 +30,16 @@ import lombok.Value;
 @Value
 public class ItemStats
 {
+	// Unarmed attack speed is 4
+	public static final ItemStats UNARMED = new ItemStats(false, true, 0, 0,
+			ItemEquipmentStats.builder()
+					.aspeed(4)
+					.build());
+
+	public static final ItemStats NOTHING = new ItemStats(false, true, 0, 0,
+			ItemEquipmentStats.builder()
+					.build());
+
 	private boolean quest;
 	private boolean equipable;
 	private double weight;
@@ -53,7 +63,7 @@ public class ItemStats
 		{
 			final ItemEquipmentStats equipment = this.equipment != null
 				? this.equipment
-				: new ItemEquipmentStats.ItemEquipmentStatsBuilder().build();
+				: NOTHING.getEquipment();
 
 			newEquipment = new ItemEquipmentStats.ItemEquipmentStatsBuilder()
 				.slot(equipment.getSlot())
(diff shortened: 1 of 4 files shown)


http-api/src/main/java/net/runelite/http/api/item/ItemStats.java
```diff
@@ -30,6 +30,16 @@
 @Value
 public class ItemStats
 {
+	// Unarmed attack speed is 4
+	public static final ItemStats UNARMED = new ItemStats(false, true, 0, 0,
```
Nightfirecat, line 34: This shouldn't be stored here since it's not used in this class. It should probably continue to remain in the overlay class.

http-api/src/main/java/net/runelite/http/api/item/ItemStats.java
```diff
@@ -81,5 +91,49 @@ public ItemStats subtract(ItemStats other)
...
+
+		final double newWeight = weight + other.weight;
+		final ItemEquipmentStats newEquipment;
+
+
+		if (other.equipment != null)
+		{
+			final ItemEquipmentStats equipment = this.equipment != null
+					? this.equipment
+					: NOTHING.getEquipment();
```
Nightfirecat, line 110: These lines along with the ItemEquipmentStatsBuilder continuation lines below are indented one level too far.
```suggestion
				? this.equipment
				: NOTHING.getEquipment();
```

Nightfirecat (changes requested):
This PR should be comprised of a couple commits:

1. Simplify and fix 2h logic (and add tests for 2h logic; these tests should fail on master and pass after this commit is applied)
2. Add equipment tab stats config, add tooltip method javadoc (?)

As-is, this is hard to review because multiple features are all being added in one giant change without any distinction regarding which code changes are attributed to which logical changes.

NJohnsonJ:
I'm going to close this PR, update the commits and reopen once it is fixed.

PR #13251, opened 2021-02-21
Feature/equipped item stat changes

### LINKED ISSUE #12628: Item Stats Plugin: Improve Equipment changes and show in worn equipment tab

NJohnsonJ:
Resolves #12628


NJohnsonJ:
Reopening this PR with more commits and fewer changes/refractors to the original code. 

This PR also addresses the requested enhancement to the display of weight.

Nightfirecat:
For what it's worth, it's much more helpful to reopen a closed PR or mark it WIP while it's still undergoing changes so we don't have to refer back to the closed PR to see what comments were left and resolved. Please keep that in mind for future PRs.

PR #13257, opened 2021-02-23
Camera Zoom Tooltip

superiorser9:
Adds the ability to show tooltip while pressing the camera zoom slider and also while dragging the slider, instead of only showing the tooltip when hovering it.


Adam-:
added in 8fe75748f9842e2a497da0e0bcb40a01271766df

PR #13261, opened 2021-02-24
Change default font size in FontManager to 12 (Fix panel component line spacing)

Cyborger1:
This fixes the line spacings all interfaces that use the `checkmark` and `x` icons.

The `defaultFont` is only ever used for Falo clues, Skills clues, emote clues, the Barrows plugin and `HotkeyButton`s in the key remapping plugin. All current cases would use a size of 12.

The `defaultBoldFont` was changed as well for consistency and is now derived to a size of `16` in the one place where it was ever used so far, which is the suggestion button on the plugin search bar.

image

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/HotkeyButton.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/HotkeyButton.java
@@ -41,7 +41,7 @@ class HotkeyButton extends JButton
 
 	public HotkeyButton(Keybind value, boolean modifierless)
 	{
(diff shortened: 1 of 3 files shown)


Al3xPro:
I only found this due to a friend having a fit about the spacing at barrows and now I can't unsee it everywhere in the game. Is there a way this could be merged or fixed in some way to correct this issue. Thanks

Adam-:
reopen if this is still an issue post b9ea1f8ff5fbeac7594e8dc6bdbf31ec1023fd43

PR #13265, opened 2021-02-24
Add activity filter to world switcher

### LINKED ISSUE #13252: Add search functionality to world swither
I would love a search bar that lets you quickly find the world you are looking for. For example, typing in "House Party" would show the worlds that are labeled as such.

joelewis43:
Closes #13252

Adds a textfield to the top of the WorldSwitcher panel used to filter worlds based on the worlds activity field.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
@@ -35,13 +35,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
-import javax.swing.JPanel;
-import javax.swing.SwingUtilities;
+import javax.swing.*;
+
 import lombok.AccessLevel;
 import lombok.Setter;
 import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.DynamicGridLayout;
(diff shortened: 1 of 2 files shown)


joelewis43:
Can anyone shed some lite on the issue I'm seeing in the checks? I performed a clean install on my machine and am not seeing this error so I do not know where to go from here.

Broooklyn:
```PMD Failure: net.runelite.client.plugins.worldhopper.WorldSwitcherPanel:214 Rule:UseEqualsToCompareStrings Priority:3 Use equals() to compare strings instead of '==' or '!='.```

Adam-:
I'm not keen with the idea of a search box for the activity taking up space, when you can already just sort the worlds by activity. It seems not that useful?

joelewis43:
@Broooklyn Thank you! In the future I will be sure to look just above the runtime summary for the error message.

@Adam- Well we could add a checkbox to the plugin that displays the search box if the user wants to see it. But as for its usefulness, sorting the whole world list by activity can still result in the use scrolling down the list looking for their desired world. With the filter option, they can just type it in, say "pvp" and have all pvp worlds readily available to them. So I see value in it for that reason.

joelewis43:
Pushed the changes I mentioned above making the activity filter configurable. The user can check the Display Activity Filter box in the settings to have the filter render at the top of the World Hopper. Otherwise the panel remains unchanged.

PR #13266, opened 2021-02-24
Adds right-click options to Configuration side bar

### LINKED ISSUE #13003: Add a Master Switch to turn off and on plugins.
Hi, I was wondering if it would be possible to add a master switch for runelites plugins.

One larger button that turns on and off your active plugins. I would love for it to be possible to remember what plugins where on when you turn them off, so you dont have to remember each time what plugins you last used. 

I made this Gif of how I think it could look like.

Master switch

It makes it easier to turn on and off all your plugins. for times that you would like to play vanilla or just check something out real quick instead of turning off 10-20 plugins manually to check something fast out.

superiorser9:
Closes #13003 

Based off the issue above, I came up with the concept of right-clicking the side bar wrench (Configuration) so it will appear these options:
- "Restore default plugins" (enable enabledByDefault plugins & disable everything else including external plugins);
- "Disable all plugins" (except Configuration & Dev tools if enabled);
- "Enable all plugins" (except GPU plugins);
- "Revert changes" (this will basically revert the plugin status after using one of the above options to last stored).

I think this is more than just a master switch/using rs=vanilla , specially the "Restore default plugins" option might come in handy on discord #support.

configoptions

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPlugin.java
@@ -24,29 +24,39 @@
  */
 package net.runelite.client.plugins.config;
 
-import java.awt.image.BufferedImage;
-import javax.inject.Inject;
-import javax.inject.Provider;
-import javax.swing.SwingUtilities;
+import com.google.common.collect.ImmutableMap;
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.MenuAction;
 import net.runelite.client.config.ChatColorConfig;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.config.RuneLiteConfig;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.OverlayMenuClicked;
+import net.runelite.client.externalplugins.ExternalPluginManager;
+import net.runelite.client.externalplugins.ExternalPluginManifest;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.plugins.PluginInstantiationException;
+import net.runelite.client.plugins.PluginManager;
 import net.runelite.client.ui.ClientToolbar;
 import net.runelite.client.ui.NavigationButton;
 import net.runelite.client.ui.overlay.Overlay;
 import net.runelite.client.ui.overlay.OverlayMenuEntry;
 import net.runelite.client.util.ImageUtil;
 
+import javax.inject.Inject;
+import javax.inject.Provider;
+import javax.swing.SwingUtilities;
+import java.awt.image.BufferedImage;
+import java.util.HashSet;
+import java.util.Set;
+
 @PluginDescriptor(
 	name = "Configuration",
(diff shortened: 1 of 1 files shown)


abextm:
Disabling every plugin is a bad idea because it will make people think they can do that as a substitute to reproducing a bug on vanilla. Enabling everything is also a bad idea, as the things that are disabled by default generally have a good reason to be and may actively conflict with other plugins.
Having a restore default plugins button seems ok, but I don't think this is a good place for it. Maybe it could live in the info panel.

superiorser9:
As suggested, removed everything but restore default plugins which I added to the info panel.

image

And also a confirmation dialog:

image

runelite-client/src/main/java/net/runelite/client/plugins/info/InfoPanel.java
```diff
@@ -170,12 +186,79 @@ void init()
...
+						if (stopPlugin)
+						{
+							try
+							{
+								pluginManager.setPluginEnabled(plugin, false);
+								pluginManager.stopPlugin(plugin);
+							}
+							catch (PluginInstantiationException ex)
+							{
+								log.warn("Unable to stop {}", plugin.getClass().getSimpleName());
```
Nightfirecat, line 230: This should log the exception as well. See [line 324 of PluginListPanel](https://github.com/runelite/runelite/blob/78eb69c8df8f823a590fe267bf9663591c3c09c0/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginListPanel.java#L324):

```java
log.warn("Error when stopping plugin {}", plugin.getClass().getSimpleName(), ex);
```

runelite-client/src/main/java/net/runelite/client/plugins/info/InfoPanel.java
```diff
@@ -170,12 +186,79 @@ void init()
...
+							if (externalPluginManager.getInstalledExternalPlugins().contains(manifest.getInternalName()))
+							{
+								stopPlugin = true;
+							}
+						}
+						if (stopPlugin)
+						{
+							try
+							{
+								pluginManager.setPluginEnabled(plugin, false);
```
Nightfirecat, line 225: This does not throw any exceptions, move it outside the `try` block.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/info/InfoPanel.java
```diff
@@ -126,7 +142,7 @@ void init()
 	{
 		setLayout(new BorderLayout());
 		setBackground(ColorScheme.DARK_GRAY_COLOR);
-		setBorder(new EmptyBorder(10, 10, 10, 10));
+		setBorder(new EmptyBorder(10, 10, 10, 5));
```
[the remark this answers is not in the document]
superiorser9: setBorder(new EmptyBorder(10, 10, 10, 10))

image

setBorder(new EmptyBorder(10, 10, 10, 5))

image

`emailLabel.setText("<a href=\"" + RUNELITE_LOGIN + "\">Login</a> to sync settings to the cloud.")` this text is too large and Box.createGlue() messes with spaces.

PR #13268, opened 2021-02-25
MusicPlugin: Fix after today's update

superiorser9:
Fixes music granular sliders after today's update (Script ID 3885 arguments changed).


abextm:
Thanks

PR #13270, opened 2021-02-25
Time tracking: Hide nonfunctional notification icons on the birdhouse panel

### LINKED ISSUE #13262: Birdhouse notification setting not preserved between sessions
**Describe the bug**
I'm not sure if this is intentional or not, but when I click the new bell icon to enable notifications in the time tracker, they are not saved and everything is reverted to the default (muted notifications) when I restart the client. If this is intentional then I think it would make more sense to save the settings.

**To Reproduce**
Turn on notifications for a farming patch / birdhouse and restart the client.

**Expected behavior**
When restarting the client, patches which I set to notify should be automatically enabled.

**Screenshots**
image

 - OS: Winblows 10
 - RuneLite version: 1.7.0
 - Launcher version: 2.1.3

Hannah-GBS:
Fixes #13262 
These were erroneously included in my notification PR so for now I'm just fixing this in lieu of figuring out an appropriate UI for the single birdhouse notification config.


Nightfirecat:
For what it's worth, I think this would be better solved by removing the "birdhouse notifications" config and having all 4 birdhouses controlled by a shared toggle. (eg. clicking the bell on any one would toggle all of them on or off)
This solution is fine in the short term, but leaves the odd UX difference between other patch notifications and birdhouses.

PR #13273, opened 2021-02-25
item identification: add option to not show names on unqiue models

### LINKED ISSUE #13272: Item Identification Change
Can we get a revert on this or have a separate checkbox that marks those seeds instead? The whole point of the plugin from the beginning was to mark the seeds that shared the same sprite. Nothing shares the same sprites as the seeds that were added a few days ago. 

3d93c857c7f1089e44765cf05ab00f46
5a6a952badab17b27fa886c71b2a27d6
5871122d427faa7e7a1055c2019e10db

Hydrox6:
This plugin is a mess. What started as a tool to allow users to actually discern items with the same inventory model has devolved into something that spams basically everything because people are too lazy to learn the colours of items. It's honestly saddening to see it go this way, so here's the fix: you want to spam all your items with letters? Go ahead, but if you just want to use the plugin for what it was originally meant for, now you can. 

All items have been sorted into one of two categories: whether the icon is "Unique" or not. The non-unique category also includes items with icons so close that they usually cause issues (eg. Str and Sara pots, or Opal and Jade).

Closes #13272


Nightfirecat:
I like this approach, but not this implementation. We could [reuse ChatbotItemSearch's shared item icon filtering](https://github.com/runelite/runelite/blob/11f32eb5ce26fe02b03fd5bdbc8a793162be6e36/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java#L307-L328) to determine which items would have unique icons rather than adding an extra argument to each entry.

Hydrox6:
The issue with that shared icon filtering is it's checking for exact dupes. As I mentioned in the original post, "The non-unique category also includes items with icons so close that they usually cause issues", and these items would not get caught without some more advanced way of looking at items.

Nightfirecat:
That's fair, though I'm still not excited about the 300-line change in ItemIdentification. Adding an `isUniqueIcon` field to `Type` could help remove a lot of these, since many item types either have all unique icons or no unique icons. (eg. herb seeds have no unique icons, impling jars and teleport tabs all have unique icons, etc.) This would require an extra `HERB_SEED` Type and another ctor, but would yield a smaller change and simpler data entry for most new entries.

PR #13274, opened 2021-02-25
world map overlay: make icons display fully on the map when edge snapped

Hydrox6:
Currently, icons snapped to the edge of the world map hang off the edge in a way that looks bad. This PR addresses that by making those icons use a slightly smaller area to position themselves, which keeps them on the map when snapped. This doesn't affect icons that don't snap, as they use a different rect.

This also fixes an issue that arose with icons overlapping the close button, which close the map when clicked. This is due to the overlay being drawn on top of the close button and completely covering it. I fixed this by consuming the mouse click event if an icon is clicked.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
@@ -131,7 +131,24 @@ public class WorldMapOverlay extends Overlay
 
 				if (worldPoint.isSnapToEdge())
 				{
-					if (worldMapRectangle.contains(drawPoint.getX(), drawPoint.getY()))
+					// Get a smaller rect for edge-snapped icons so they display correctly at the edge
+					final Rectangle snappedRect = widget.getBounds();
+					snappedRect.grow(-image.getWidth() / 2, -image.getHeight() / 2);
+
+					final Rectangle unsnappedRect = new Rectangle(snappedRect);
+					if (worldPoint.getImagePoint() != null)
+					{
+						int dx = worldPoint.getImagePoint().getX() - (image.getWidth() / 2);
+						int dy = worldPoint.getImagePoint().getY() - (image.getHeight() / 2);
+						unsnappedRect.translate(dx, dy);
+					}
+					// Make the unsnap rect slightly smaller so a smaller snapped image doesn't cause a freak out
+					if (worldPoint.isCurrentlyEdgeSnapped())
+					{
+						unsnappedRect.grow(-image.getWidth(), -image.getHeight());
+					}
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
```diff
@@ -57,6 +57,7 @@
 	private static final int TOOLTIP_PADDING_HEIGHT = 1;
 	private static final int TOOLTIP_PADDING_WIDTH = 2;
 	private static final int TOOLTIP_TEXT_OFFSET_HEIGHT = -2;
+	private static final int ICON_EDGE_SNAPPED_INSET = -15;
```
Nightfirecat, line 60: Where does this value come from? This won't look right for all icons; we should calculate this for each icon's size

Nightfirecat (commented):
This change looks very nice visually, and the click consume is a good fix. :+1:

runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
Hydrox6: I arbitrarily chose it from the current plugins that use it (Clues and Party) as a decent value to use. I thought about calculating it for each icon separately, and in hindsight it would probably be better to do it like that.

Nightfirecat:
The only issue I see left worth fixing is that the edge snap detection doesn't account for imagePoint being set, so you end up with these being possible: (the latter being the closest the image can be to the bottom edge before snapping)

map-north-overlap
map-south-limit

A minor issue besides that is that because this image actually changes size when snapped, it can flicker if it is close enough to the edge that the pre-snap image would be considered snapped, but the post-snap image is not close enough to be snapped. This may not be easy enough to be worth solving in the short term.

Hydrox6:
Updated to make snapping take into account imagePoints, and also removed the flickering where the snapped and unsnapped arrows were fighting due to differing sizes.
(The red line is the outer limit of where the icon would be when snapped, and the blue line is where the point transitions from snapped to unsnapped and vice versa. They were made visible for debugging and are not visible in the commits)
record_2021-02-27__23-09-30

Hydrox6:
I have a suspicion that the click consuming is causing dead clicks; going to investigate, fix, then reopen

Hydrox6:
So I was correct: the map overlay listener is always active, and so it's checking for clicks even when the overlay is closed. Before that wasn't a problem, and just caused a little extra work for the program, but with the consume changes I made this caused dead clicks. `mouseMoved` already implemented a guard check to make sure the map was open, so I've added this check to `mousePressed`

PR #13275, opened 2021-02-25
XPGlobes: fix

### LINKED ISSUE #13060: XP globes require two xp drops after being turned on to start showing
**Describe the bug**
When turning the plugin on after starting the client, XP globes will not appear right away after gaining experience, likely because the saved experience amounts had not been initialized.

**To Reproduce**
1. Turn off the XP globes plugin prior to starting the client
2. Log in
3. Turn on the plugin
4. Gain experience in any skill
5. Observe that no globes appear (gain experience in that skill again to see them appear as expected)

superiorser9:
Closes #13060 

The problem resides on turning on XPGlobes Plugin when logged in.
`globalCache` array is updated on `onStatChanged` event, therefore on first xp drop the index content will be null and resulting on xp globe not being drawn.
To overcome this issue, once the plugin is turned on it will write to `globalCache` array all the skills data, so by the time `onStatChanged` event is fired, the index won't be null and it will draw the xp globes.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesPlugin.java
@@ -33,7 +33,9 @@ import java.util.Comparator;
 import java.util.List;
 import javax.inject.Inject;
 import lombok.Getter;
+import net.runelite.api.Client;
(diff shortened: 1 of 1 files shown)


Adam-:
This seems like such a minor issue, may not be worth fixing. But if we do want to fix it, you should just refactor the common globe setup code into a method and call it from here and also in `onStatChanged`.

PR #13284, opened 2021-02-27
Fix POH altar burners false positive burn message

### LINKED ISSUE #13281: False positive burn message from lighting POH altar burners
This message from lighting POH altars errantly triggered the cooking and wiki crowdsource plugins:

You burn some marrentill in the incense burner.

This should be fixed in both plugins and have a test case added in the cooking plugin.

superiorser9:
Closes #13281

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/cooking/CookingPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/cooking/CookingPluginTest.java
@@ -60,7 +60,8 @@ public class CookingPluginTest
 		"You roast a lobster.",
 		"You cook a bass.",
 		"You successfully bake a tasty garden pie.",
-		"You dry a piece of meat and extract the sinew."
+		"You dry a piece of meat and extract the sinew.",
+		"You burn some marrentill in the incense burner."
 	};
 
 	@Inject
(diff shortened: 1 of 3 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/cooking/CookingPluginTest.java
```diff
@@ -60,7 +60,8 @@
 		"You roast a lobster.",
 		"You cook a bass.",
 		"You successfully bake a tasty garden pie.",
-		"You dry a piece of meat and extract the sinew."
+		"You dry a piece of meat and extract the sinew.",
+		"You burn some marrentill in the incense burner."
```
Nightfirecat, line 64: Don't add this to the successful cooking messages. Define and test it separately from this array.

Nightfirecat (changes requested):
Based on the linked spreadsheet in #13222, it seems that the only burn message of this format is "You burn the mushroom in the fire." Changing the affected lines to be `message.equals()` would be a better solution here.

runelite-client/src/test/java/net/runelite/client/plugins/cooking/CookingPluginTest.java
superiorser9: Changed as requested.

PR #13316, opened 2021-03-06
hiscore: Add lookup fallback for uncached players

### LINKED ISSUE #13310: Question: Right-click lookup after user disappears
I was curious if anyone else noticed this:

When you right-click lookup a player in RuneLite you used to be able to look them up even if they had left the area or you teleported away. Now (as of maybe a month ago?) clicking lookup after they are no longer in your chunk results in nothing happening. Not sure if this was by design but wanted to check, seems kind of silly since you could theoretically screen-cap their name and look it up later.

Thanks for your time and my apologies if this has been posted before, I did look.

Cyborger1:
Adds parsing logic on the menu entry target as a fallback, allowing the lookup to work on players that are no longer cached after they moved out of view of the client.

Fixes #13310.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePlugin.java
@@ -36,6 +36,7 @@ import javax.inject.Provider;
 import javax.swing.SwingUtilities;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
+import net.runelite.api.IconID;
 import net.runelite.api.MenuAction;
 import net.runelite.api.MenuEntry;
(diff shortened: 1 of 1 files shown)


Adam-:
If you're going to do this, you should at the very least look at the original code, which covers more edge cases than what you have here - https://github.com/runelite/runelite/commit/f6c68eefc8a3c5415451bdbd6190b1f745ed5489#diff-70e4e692a15055e3d73faf2fcb173f3ce80b9aa4562b53e9d05d91f115c9c0adL261-L267. However I am not really a fan of this.

Cyborger1:
Sorry, I didn't think to check if the original parsing was done in a different class and missed it. I have changed the code to use the old parsing logic.

Cyborger1:
Closing this as #13330 is a much better solution.

PR #13324, opened 2021-03-09
Menu Entry Swapper: Achievement Diary Teleports

GG46KG:
Add configuration for Achievement Diary items with nonstandard teleports


Nightfirecat:
These swaps [already exist in the Menu Swapper Extended plugin](https://github.com/raiyni/menu-swapper-extended/blob/4a2c8d1edd7f490e0c92c5767ee36eda74b5804c/src/main/java/io/ryoung/menuswapperextended/MenuSwapperConfig.java#L251-L282). I'm not sure we want to add these to the main client.

PR #13329, opened 2021-03-12
Add more options for Inventory Grid

richardcane:


runelite-client/src/main/java/net/runelite/client/plugins/inventorygrid/InventoryGridOverlay.java
```diff
@@ -72,6 +73,30 @@ private InventoryGridOverlay(InventoryGridConfig config, Client client, ItemMana
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
+		final Color GRID = config.gridColor();
+
+		if (config.showGrid() == GridMode.ALWAYS) {
+			Widget backpack = client.getWidget(WidgetInfo.INVENTORY);
+			Widget bank = client.getWidget(WidgetInfo.BANK_INVENTORY_ITEMS_CONTAINER);
+			Widget activeInventoryWidget = backpack.isHidden() ? bank : backpack;
```
richardcane, line 81: This works, but looks like it'd probably want to check whether both are hidden and early return, right?

PR #13331, opened 2021-03-12
Idle: Add Shades of Mort'ton Minigame

### LINKED ISSUE #12905: [Idle Notifier] Trigger Idle notifications at Shades of Mort'on
Repairing the temple wall during the Shades of Mort'on minigame is an automated idle activity that occasionally times out. However, even with Animation/Interaction checked in the Idle Notifier plugin settings, no notification triggers when the repair process times out, even though the player character stops performing the repair animation.

Adjust the plugin so that an idle notification triggers when the temple repairing animation stops, if "Idle Animation" notifications are enabled.

"Idle Interaction" would work too if RuneLite can distinguish temple repair as an interaction activity.

**Additional context**

WolffTech:
Add's reinforce temple wall to the idle animations alert.

Closes #12905


Nightfirecat:
`LOOKING_INTO` is already defined as the animation with ID 832, which is currently breaking the `checkAnimationReset()` test. To fix this, I would remove the new AnimationID definition you created, replace its usage in the plugin with `LOOKING_INTO`, and change the test to use some known never-used animation ID, like `-2`.

WolffTech:
Would you recommend that I create an Animation ID definition for -2 and use that for the test? Or should I just use an already existing definition that is unused?

runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -33,6 +33,7 @@
 public final class AnimationID
 {
 	public static final int IDLE = -1;
+	public static final int UNUSED = -2;
```
Nightfirecat, line 36: Don't declare this here; add this to the test file instead.

runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
```diff
@@ -153,7 +153,7 @@ public void checkAnimationReset()
 		animationChanged.setActor(player);
 		plugin.onAnimationChanged(animationChanged);
 		plugin.onGameTick(new GameTick());
-		when(player.getAnimation()).thenReturn(AnimationID.LOOKING_INTO);
+		when(player.getAnimation()).thenReturn(AnimationID.UNUSED);
```
Adam-, line 156: this should just be `0x7fffffff` or some other unused value.

Nightfirecat:
Added in d58cecdb8d5a2d6958a5bcf1381805c6d64be7b8

PR #13332, opened 2021-03-12
Improve the Bank Tags plugin's "tag tab tab" view.

geheur:
Improves the "tag tab tab" view:

1. Allowing the left-click option on the button at the top left of the bank to be swapped to any of the options in the right-click menu, using a config dropdown menu.

2. Changes tab selection to always scroll the tab into view in the sidebar - this is so that when a tab is selected from the "tag tab tab" view it will be visible in the sidebar so that you know which tab you're looking at, and can drag new items into the tab.

Also just in case this feature is obscure (I didn't know it existed until a few days ago), this is the thing I'm talking about: 
image


geheur:
Changed the icon for "View tag tabs" to this: 
open-tab-menu

geheur:
closing because it is somewhat superfluous now that the option can be swapped in other ways, also making the tag tab list scroll to the selected tag is sometimes annoying and would have to be reworked.

Zoboomaku:
> closing because it is somewhat superfluous now that the option can be swapped in other ways, also making the tag tab list scroll to the selected tag is sometimes annoying and would have to be reworked.

Do you mind listing some of the other ways this can be swapped?  I stumbled upon this coming to suggest this change, but am failing to find a way to do it now, and I'm sure others might as well

geheur:
@Zoboomaku "Custom Menu Swaps" plugin hub plugin.

PR #13341, opened 2021-03-16
Clarify how to enter multiple words

FWDekker:
A simple non-functional change that should clarify usage.

Clarify that multiple words can be entered in the input box by separating them with commas.


runelite-client/src/main/java/net/runelite/client/plugins/chatnotifications/ChatNotificationsConfig.java
```diff
@@ -44,7 +44,7 @@
 		position = 1,
 		keyName = "highlightWordsString",
 		name = "Highlight words",
-		description = "Highlights the following words in chat",
+		description = "Highlights the following comma-separated words in chat",
```
Nightfirecat, line 47: This is more in line with existing config description wording, such as Chat Filter
```suggestion
		description = "Highlights the following words in chat, separated by commas",
```

Nightfirecat (changes requested):

PR #13352, opened 2021-03-19
Add barrows door highlight

iBelg:
This adds a green/red highlight on doors depending on unlocked/locked. This will help being able to see what path is useful to take at a glance.

Currently I am using strange old lockpicks to reach the center, but obviously I want to minimize the use of them so I try and take as many unlocked doors as possible, but it's not always obvious which ones are unlocked as sometimes it's too dark to distinguish the slightly paler unlocked doors. This change allows me to see it way more easily. I believe this isn't against the 3rd party client guidelines set out by Jagex.


runelite-client/src/main/java/net/runelite/client/plugins/barrows/BarrowsOverlay.java
```diff
@@ -24,18 +24,16 @@
  */
 package net.runelite.client.plugins.barrows;
 
-import java.awt.Color;
-import java.awt.Dimension;
-import java.awt.Graphics2D;
-import java.awt.Rectangle;
+import java.awt.*;
```
mdennis1318, line 27: Remove * imports across review

raiyni:
Exists as a phub plugin

PR #13370, opened 2021-03-22
Menu Entry Swapper: Add Barbarian Assault scroll use-swap

rsfost:
The default left-click option is "Read" which is pretty useless. This makes the left-click option "Use".


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -375,6 +375,8 @@ void setupSwaps()
 		swap("eat", "guzzle", config::swapRockCake);
 
 		swap("travel", "dive", config::swapRowboatDive);
+
+		swap("read", "scroll", "use", config::swapBarbarianAssaultScroll);
```
domien96, line 379: How will this make a difference with other scrolls? Not related to BA.

Nightfirecat:

PR #13376, opened 2021-03-23
change hs kc patterns to match values greater than 999

jpritcha3-14:
This PR changes the capture groups for the the Hallowed Sepulchre kc patterns to match values that contain commas, and then strips those commas before invoking setKc.


Adam-:
This looks good. You could also change the `testHsFloorKc` and `testHsGhcKc` tests to test using a number with a comma so this doesn't regress later accidentally.

PR #13379, opened 2021-03-24
xp globes: scale icon while respecting aspect ratio

Hydrox6:
Currently the scaling does not look good, as it just forces the image into the size of the orb, not respecting the aspect ratio
image
image

This PR streamlines the code a lot (not sure why the old code even defined width and height given the values were always the exact same), while also correcting the scaling so it actually looks good. This is done through an interesting feature of `Image.getScaledInstance` whereby giving a negative value for either width or height will result in it using a value that keeps the aspect ratio for the image. 
record_2021-03-24__15-34-19

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
@@ -209,7 +209,38 @@ public class ImageUtil
 	 */
 	public static BufferedImage resizeImage(final BufferedImage image, final int newWidth, final int newHeight)
 	{
-		final Image resized = image.getScaledInstance(newWidth, newHeight, Image.SCALE_SMOOTH);
+		return resizeImage(image, newWidth, newHeight, false);
+	}
+
+	/**
+	 * Re-size a BufferedImage to the given dimensions.
+	 *
+	 * @param image the BufferedImage.
+	 * @param newWidth The width to set the BufferedImage to.
+	 * @param newHeight The height to set the BufferedImage to.
+	 * @param preserveAspectRatio Whether to preserve the original image's aspect ratio. When {@code true}, the image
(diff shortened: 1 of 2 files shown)


Nightfirecat (approved):

runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -199,17 +199,39 @@ public static BufferedImage grayscaleImage(final BufferedImage image)
 		return ImageUtil.bufferedImageFromImage(grayImage);
 	}
 
+
+	public static BufferedImage resizeImage(final BufferedImage image, final int newWidth, final int newHeight)
```
Nightfirecat, line 203: This should have the javadoc of the original method

runelite-client/src/main/java/net/runelite/client/util/ImageUtil.java
```diff
@@ -199,17 +199,39 @@ public static BufferedImage grayscaleImage(final BufferedImage image)
...
+		return resizeImage(image, newWidth, newHeight, false);
+	}
+
 	/**
 	 * Re-size a BufferedImage to the given dimensions.
 	 *
 	 * @param image the BufferedImage.
 	 * @param newWidth The width to set the BufferedImage to.
 	 * @param newHeight The height to set the BufferedImage to.
+	 * @param preserveAspectRatio Whether to preserve the original image's aspect ratio.
```
Nightfirecat, line 214:
```suggestion
	 * @param preserveAspectRatio Whether to preserve the original image's aspect ratio. When {@code true}, the image
	                              will be scaled to have a maximum of {@code newWidth} width and {@code newHeight}
	                              height.
```

Nightfirecat (changes requested):
Code LGTM

Nightfirecat (approved):

PR #13384, opened 2021-03-25
loottracker: Add support for Tempoross Reward pool and new Casket

Broooklyn:
Adds loot tracking for the Reward pool (Tempoross) and the new Casket. Per the request of the wiki, fishing level metadata is included.


Broooklyn:
Added a commit that separates events that do not produce ground items

PR #13391, opened 2021-03-27
ge plugin : panel item right click label according configuration (wiki/runescape ge)

### LINKED ISSUE #13382: Grand exchange panel item right-click menu should say "Open RuneScape Wiki prices website" when actively traded price is enabled
Currently, this right-click option has static text, despite having a dynamic target based on the "Use actively traded price" option under RuneLite's settings.

Haashi:
Closes #13382


Nightfirecat (commented):
Rather than passing `RuneLiteConfig` down to the offer slot, a package-private `openGeLinkText` method could be created in the plugin to return the appropriate string for the config. That said, another issue which needs to be considered is that the text must be rebuilt if that config is changed--currently that does not happen, so that would need to be added also.

PR #13396, opened 2021-03-28
loottracker: Use real skill level for Tempoross loot metadata

Broooklyn:
Per request by the Wiki, it turns out the Tempoross reward pool does not take into account boosted level when deciding loot, so getRealSkillLevel should be used for metadata instead.


Adam-:
wiki no longer requires this.

PR #13397, opened 2021-03-28
MenuEntrySwapper: add take to light swap for dropped logs in MenuEntrySwapper

### LINKED ISSUE #13369: Menu Entry Swapper Suggestion: [Dropped Item] - Left Click - Light Logs
Menu Entry Swapper Suggestion: [Dropped Item] - Left Click - Light Logs

mdennis1318:
Tested by checking menu entries on dropped logs with config true and false

Closes #13369

droppedLogsOff

droppedLogsOn


Adam-:
Swaps on ground items like this is currently unprecedented, so it gives me some pause. I'd probably prefer to not do it just to avoid any potential firestorms.

PR #13403, opened 2021-03-29
Combine BGS and BGS(OR) as one special weapon to count as same counter

### LINKED ISSUE #13402: Bandos godsword and Bandos godsword (or) dont combine damage for corp

When using the special attack plugin along with party for duoing corp, if one person has a normal BGS and another has the ornament BGS, you will get two different damage counters

**To Reproduce**
One person with BGS, another with BGS (or)

**Expected behavior**
Two different damage counters, instead of one

**Screenshots**
BGS damage

mdennis1318:
Closes #13402

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/specialcounter/SpecialCounterPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/specialcounter/SpecialCounterPluginTest.java
@@ -258,6 +258,11 @@ public class SpecialCounterPluginTest
 		specialCounterPlugin.onVarbitChanged(new VarbitChanged());
 		specialCounterPlugin.onHitsplatApplied(hitsplat(target, Hitsplat.HitsplatType.DAMAGE_ME));
 
+		// Set up spec weapon as BGS(OR)
+		ItemContainer equipment = mock(ItemContainer.class);
+		when(equipment.getItem(EquipmentInventorySlot.WEAPON.getSlotIdx())).thenReturn(new Item(ItemID.BANDOS_GODSWORD_OR, 1));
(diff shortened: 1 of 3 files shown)


Adam-:
This same issue should also be fixed for the bone dagger I think.

mdennis1318:
I agree, but they are more "different" than bgs vs bgs(or) so it is not as clear cut. I will make the update.

Adam-:
I was under the impression their spec was the same, is it not?

mdennis1318:
It is the same, just the weapon itself is slightly different, where the BGS(or) is purely cosmetic

Adam-:
As long as the specs are the same, they should probably be counted as the same within the plugin.

runelite-client/src/test/java/net/runelite/client/plugins/specialcounter/SpecialCounterPluginTest.java
```diff
@@ -258,6 +258,11 @@ public void testNotification()
 		specialCounterPlugin.onVarbitChanged(new VarbitChanged());
 		specialCounterPlugin.onHitsplatApplied(hitsplat(target, Hitsplat.HitsplatType.DAMAGE_ME));
 
+		// Set up spec weapon as BGS(OR)
+		ItemContainer equipment = mock(ItemContainer.class);
+		when(equipment.getItem(EquipmentInventorySlot.WEAPON.getSlotIdx())).thenReturn(new Item(ItemID.BANDOS_GODSWORD_OR, 1));
+		when(client.getItemContainer(InventoryID.EQUIPMENT)).thenReturn(equipment);
+
```
Nightfirecat, line 265: Instead of adding this to an existing test, another test should be added to verify specs from varied weapon types are (a) identified as special attacks, and (b) stacked instead of tracked separately.

PR #13419, opened 2021-03-31
menu entry swapper: add Tempoross leave swap

molo-pl:
After subduing Tempoross, to save some time, most players choose to "Leave" instantly by interacting with one of four NPCs in the cove. I believe swapping the default "Talk-to" with "Leave" would be a minor QOL improvement for the boss.

Some self-asked QA questions:

**Is there a risk that players will leave by accident before the fight is over?**

Nope, because the option to leave early is called "Forfeit", which is not being swapped. So throughout the fight, the left-click option for each NPC is going to be "Talk-to".

**Should this swap be enabled by default?**

In my opinion, yes. I believe in 99% or even more cases players just want to leave the cove once the fight is finished. These NPCs are available outside of the cove should anyone want to learn how to play. There's also plenty of time to talk to the NPCs during the fight.

**Should the swap be limited just to the 4 NPCs, or should it be a general thing?**

I think limiting it makes it safer to enable by default. Although I'm not aware of any other NPCs with a "Talk-to" and "Leave" interactions combo. As of now, the name of the config item reflects the fact that it is limited to Tempoross ("Tempoross Leave"). If it's made more generic, it should probably be called just "Leave".

**How was it tested?**

I've manually tested it during a few Tempoross encounters, with both captains and first mates NPCs. ~Also added a unit test, not sure if it adds value though, as it is quite basic.~ The unit test is gone, after discussing on Discord.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -661,4 +661,15 @@ public interface MenuEntrySwapperConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "swapTemporossLeave",
+		name = "Tempoross Leave",
+		description = "Swap Talk-to with Leave after subduing Tempoross",
+		section = npcSection
+	)
+	default boolean swapTemporossLeave()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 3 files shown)


Adam-:
There is an exiting `quick-leave` option, shouldn't this just use that?

molo-pl:
I can think of a few reasons why perhaps not:

- Tempoross leave is not an **Object** swap, nor it is a **Quick-X** action.
- Similar case: NPC swap `Start-minigame` for Pyramid Plunder is not grouped with Object swap `Quick Pass/Open/Start/Travel`.
- `Quick-Leave` setting is **OFF** by default (see my prev comment), and perhaps stays **OFF** for a reason (it can cause you to exit Pyramid Plunder early on a missclick).

SKDudeMan (approved):

SKDudeMan:
> I can think of a few reasons why perhaps not:
> 
> * Tempoross leave is not an **Object** swap, nor it is a **Quick-X** action.
> * Similar case: NPC swap `Start-minigame` for Pyramid Plunder is not grouped with Object swap `Quick Pass/Open/Start/Travel`.
> * `Quick-Leave` setting is **OFF** by default (see my prev comment), and perhaps stays **OFF** for a reason (it can cause you to exit Pyramid Plunder early on a missclick).

Agreed it's better to make this it's own swap entry. Hope we can get this in soon :)

PR #13435, opened 2021-04-03
Update DPS counter

JoRouss:
[FIX] When in a party, "Only boss damage" would work only if everyone in the party had it setup the same.
Members who had it enabled would not send their hit to those who had it disabled. Members who had it enabled would receive and count hits from those who had it disabled.

[FEATURE] Added Tempoross boss IDs.

[FEATURE] Ability to recolor your damage line in the overlay when in a party.
I found it hard to keep track of my own damage when in a party. Having it recolored for myself makes it really easy.

[FEATURE] Ability to order damage lines by DPS in the overlay.
I found it hard to quickly know who has the most damage in the party. Combined with the previous recoloring feature, it makes it easy to know who has the most damage, and if it's actually me.

[FEATURE] Added the ability to reset the DPS counter when a boss is hit, after a previous one died.
With my party members, we wanted to be able to see the DPS done to a boss in between kills, and have it reset when we engage the next one. Currently, it was only possible to have the counters reset as soon as the boss died, leaving no time to actually review the DPS.

[FEATURE] Added Party and Boss config sections


runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -145,6 +145,8 @@ public class DpsCounterPlugin extends Plugin
 	@Getter(AccessLevel.PACKAGE)
 	private final DpsMember total = new DpsMember("Total");
 
+	private boolean bossDied;
```
Nightfirecat, line 148: Reset this on `shutDown()`

JoRouss: Fixed in 5b0b2dccde0a52144f3a5d67eb5797c8bb938011

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsMember.java
```diff
@@ -96,4 +97,10 @@ Duration elapsed()
 	{
 		return Duration.between(start, end == null ? Instant.now() : end);
 	}
+
+	@Override
+	public int compareTo(DpsMember o)
+	{
+		return Integer.compare(damage, o.getDamage());
```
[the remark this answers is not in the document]
JoRouss: Very true. And it even makes the code more simple.
Fixed in c45a1ad0f60a55dc11234841756f70a84bdf1e9d

Adam-:
It might be better to make the existing autoreset reset on next boss hit instead of adding a new config op for that, since I think the current behavior isn't very useful?

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -168,6 +174,12 @@ protected void shutDown()
 		members.clear();
 	}
 
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		sort();
```
Adam-, line 183: this should be checking at least the config group being changed, maybe also the key.

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -326,4 +340,32 @@ private void unpause()
...
+		boolean inParty = !partyService.getMembers().isEmpty();
+		boolean showDamage = dpsConfig.showDamage();
+
+		if (inParty && dpsConfig.sortByDps())
+		{
+			if (showDamage)
+			{
+				sortedMembers = members.values()
+					.stream()
+					.sorted((e1, e2) -> Integer.compare(e2.getDamage(), e1.getDamage()))
```
Adam-, line 381: `.sorted(Comparator.comparing(DpsMember::getDamage).reversed())`

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsCounterPlugin.java
```diff
@@ -326,4 +340,32 @@ private void unpause()
...
+					.collect(Collectors.toList());
+			}
+			else
+			{
+				sortedMembers = members.values()
+					.stream()
+					.sorted((e1, e2) -> Float.compare(e2.getDps(), e1.getDps()))
```
Adam-, line 388: same here

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsConfig.java
```diff
@@ -68,10 +87,35 @@ default boolean autoreset()
...
 	}
+
+	@ConfigItem(
+		position = 5,
+		keyName = "selfColor",
+		name = "Self color",
+		description = "Change your own damage line color",
+		section = partySection
+	)
+	default Color selfColor()
```
Adam-, line 117: put this selfColor in a separate commit

runelite-client/src/main/java/net/runelite/client/plugins/dpscounter/DpsOverlay.java
```diff
@@ -114,16 +116,26 @@ public Dimension render(Graphics2D graphics)
 		int maxWidth = ComponentConstants.STANDARD_WIDTH;
 		FontMetrics fontMetrics = graphics.getFontMetrics();
 
-		for (DpsMember dpsMember : dpsMembers.values())
+		for (DpsMember dpsMember : dpsCounterPlugin.getSortedMembers())
```
Adam-, line 119: I may reverse my original opinion on moving the sorting out to a plugin, there are too many data races with what you've implemented here, since members is accessed from multiple threads (hence the CHM) and you aren't handling this correctly anywhere currently. Eg here this would NPE if run between member update and sorting. Since there are so few players maybe just sorting here is fine then.

Adam- (changes requested):

Adam-:
The only boss damage fix was commited in bed28462a3f831fbb82c71db50cd947b2740f6ad

PR #13452, opened 2021-04-07
Add default color config for XP drops when no offensive prayer is active

ogoldthwaite:
Add config to XP Drops plugin to set the text color when there is color changing prayer active. 

Prayer colors still take precedence over default color.


Adam-:
added in #13729

PR #13461, opened 2021-04-09
Menu Entry Swapper: Swap Construction Cape Wear with Tele to POH

dfree951:
Allows for a left-click "Tele to POH" for the construction cape without interfering with regular "Teleport Item" functionality.


Adam-:
this was added a bit ago in e3f2b28fac0049a4f24107fb89736f9eb4fa43df

PR #13482, opened 2021-04-14
Region data for Camdozaal and enums for Below Ice Mountain quest

LlemonDuck:


runelite-api/src/main/java/net/runelite/api/SpriteID.java
```diff
@@ -732,6 +732,7 @@
 	public static final int EMOTE_ZOMBIE_DANCE = 734;
 	public static final int EMOTE_RABBIT_HOP = 735;
 	public static final int EMOTE_SKILLCAPE = 736;
+	public static final int EMOTE_FLEX = 2426;
```
Nightfirecat, line 735: ?

LlemonDuck: The quest added a new Flex emote, and this is the ID for the sprite in the emote panel. I also added the locked version a few lines down.

PR #13488, opened 2021-04-15
Fix Fremmenik diary file so skill and quest requirement show properly

### LINKED ISSUE #13480: Fremmy Elite Diary Missing Runecraft Level Requirement
Fremennik Elite diary does not display 82 Rc requirement.

nganzalo:
Changed the string in fremmenik diary for the skill requirment for crafting 56 astral runes, to the updated and correct string so the skill requirment will show again.

Closes #13480.


Nightfirecat:
This has been added in #13485.

PR #13489, opened 2021-04-15
idle notifier: add support for the imcando hammer

Hydrox6:
This is the new fancy wieldable hammer from the Ruins of Camdozaal


Nightfirecat:
It seems like there are other animations up to 8916 which are mirrors of other normal-hammer animations (such as `AnimationID.CONSTRUCTION) which have Imcando hammer variants. These should also be added and used where apporpriate

Hydrox6:
Added the other used hammer animation for Wintertodt. I assume the other 4 are used in quests or one-time events, as we don't seem to have the normal hammer ones mapped.

PR #13517, opened 2021-04-23
fix crystal pickaxe inactive not counting #13515

### LINKED ISSUE #13515: Crystal pickaxe (inactive) Charlie the Tramp clue
Currently the mine a piece of iron ore for Charlie the Tramp clue step, doesn't count the Crystal pickaxe (inactive) as pickaxe. Like it shows a red x instead of a green check symbol on the items required. It is really small and up to you guys if you can be bothered to fix it.

pristit:
The inactive version of the crystal pickaxe doesn't count as a pickaxe during Charlie the Tramp task to mine an iron ore.

This should fix that by adding the inactive version as an item in the requirement collection for pickaxes.

Fixes #13515


makingcomment:
It looks like the crystal axe and crystal harpoon have the same issue.

pristit:
> 
> 
> It looks like the crystal axe and crystal harpoon have the same issue.

Added the inactive versions of the axe and harpoon to the skill challenge requirement categories.

Issue #13290, opened 2021-02-28
Filter in-game world hopper based on region


laska89:
**Is your feature request related to a problem? Please describe.**
The in-game world list shows worlds from all regions, as well as f2p worlds. This is very cluttered and messy,

**Describe the solution you'd like**
If I live in USA east region and am p2p, I would like the in-game world hopper to show only USA p2p worlds, not f2p worlds or European worlds.

**Describe alternatives you've considered**
The runelite world switcher also lists all the worlds. The closest option is quick-hop can switch within your region.

**Additional context**

ragsmaroon:
The Runelite world hopper can remove all f2p worlds. Not a full solution to your problem but it's helpful.

PR #13523, opened 2021-04-24
Add world hopper config option to filter worlds by region

isaacph:
Add the new option to the config, respond to config update events, and account for the new filter in the populate function.
Issue #13290 asks for this in the in-game world hopper. This might be a nice place to start.

Note: the WorldSwitcherPanel class had different line endings than the default, so the auto-formatter seems to have made Git think the whole thing was rewritten.


abextm:
You need to revert the changes to the line endings. You will probably need to set `core.autocrlf=false` in your git config

PR #13524, opened 2021-04-25
Adds vanity options to woodcutting and mining plugins' overlay

Megamagepal:
This change will let you choose color, opacity, and diameter of pie circles in the woodcutting and mining plugins to keep them more in line with other plugins that allow similar functionality.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningConfig.java
@@ -25,18 +25,18 @@
  */
 package net.runelite.client.plugins.mining;
 
-import net.runelite.client.config.Config;
-import net.runelite.client.config.ConfigGroup;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningConfig.java
```diff
@@ -53,4 +59,76 @@ default boolean showMiningStats()
...
+	default Color getRockOverlayOutlineColor()
+	{
+		return Color.ORANGE;
+	}
+
+	@Alpha
+	@ConfigItem(
+		position = 13,
+		keyName = "hexColorRockOverlayMLM",
+		name = "MLM Overlay color",
```
Adam-, line 100: This is used for more than just MLM, but instead anything which can respawn. It is used for showing that enough time has elapsed that it is possible it can respawn. Additionally the default values for these should match the existing values we use so that the default behavior doesn't change.

PR #13532, opened 2021-04-27
Add Ring of Endurance effect to RunEnergyPlugin

JZomDev:
Similar to stamina potions, but applies a 15% reduction to run energy drain rate.


Adam-:
According to the wiki, this only happens with 500 charges or more https://oldschool.runescape.wiki/w/Ring_of_endurance

Issue #12720, opened 2020-10-30
Thieving: Pickpocketing stun timer over player (consider eating while stunned)


CptBrian:
(A timer like those you see on trees and rocks)
• Place over player or the player's head.
• Timer shows stun duration until you can pickpocket again.
• This is not the same as the timer until you can move again(explained below)
• Eating while stunned will extend/pause the stun timer(beyond its visible effect), but only for your ability to pickpocket, not move.
• The duration it extends/pauses should be however long you spent eating while stunned.
• Eating while *not* stunned does not affect the pickpocket cooldown.
I'm sure with a little trial-and-error, you'll figure out the right timings.

Mrnice98:
I like this idea, I am going to take a look into implementing something along these lines in the coming days.

Adam-:
added in #13552 as an infobox

PR #13552, opened 2021-04-30
Added pickpocket stun timer to timers plugin.

Zytion:
A simpler implementation of the feature request brought up in #12720.

pickpocket
Timer displays how long the player is stunned for (5 seconds for most pickpocket targets).


makingcomment:
5 seconds isn't a multiple of 0.6. Do stuns work outside the tick system? 😮

PR #13557, opened 2021-05-01
Increase camera plugin maximum outer zoom limit

mblaurens:


LlemonDuck:
For reference, this is the current limit on the max render distance:

image

And this is your proposed change:

image

Extending the zoom limit on the camera does not innately extend the draw distance.

mblaurens:
There are some niche use cases where a larger outer zoom limit would be useful when used with GPU plugin to increase draw distance. For example, at blood runecrafting altar, it would enable moving directly from the dark altar to the blood altar in one click, which is not possible with the current 400 zoom limit.

raiyni:
This was the last response to this suggestion.

https://github.com/runelite/runelite/pull/12745#issuecomment-749490454

Hydrox6:
> There are some niche use cases where a larger outer zoom limit would be useful when used with GPU plugin to increase draw distance. For example, at blood runecrafting altar, it would enable moving directly from the dark altar to the blood altar in one click, which is not possible with the current 400 zoom limit.

The zoom limit and the click distance are entirely separate; if were at max zoom before and you zoomed out further, you can't click any further. The scene is also limited to 104 tiles in each direction, whereas you need to move over 101 tiles to get between the mine and the altar. Even if you managed to set up the perfect scene, you would trigger load lines and screw that scene up, at which point if you'd click to go up to the altar it would just be dropped as there's no valid path in the current scene.

This change has very little practical purpose to everyone but a snowflake wanting to try out Satellite-man mode. As abex has also mentioned before, it would also have other issues, which is why the limit is set at the reasonable amount it's currently at.

mblaurens:
Thanks for explaining. Those are fair reasons, but they weren't obvious to me as an outsider. Would it be good to document the reasoning for current maximum zoom limit of 400 in a code comment similarly to how the 1004 inner zoom limit is?

PR #13563, opened 2021-05-03
Disable portal nexus swap when holding shift modifier

JZomDev:
Add shift modifier to the portal nexus to disable the swap


Nightfirecat:
I'm not a fan of this, as I'd expect that if this new behavior was added to handle Shift being pressed, I would also expect the reverse behavior to work. (config disabled with shift pressed would swap "Teleport menu" to left-click) At the very least, I would expect this to be explained in the config description, but even that would not make this behavior obvious.

PR #13564, opened 2021-05-05
[loot] - added tracking for lockboxes from the Ruins of Camdozaal Vault

weiyih:
Tracking the lockboxes from the Camdozaal Vault

Screen Shot 2021-05-05 at 5 12 25 PM

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -792,7 +792,8 @@ public class LootTrackerPlugin extends Plugin
 			|| WINTERTODT_SUPPLY_CRATE_EVENT.equals(eventType)
 			|| eventType.endsWith("Bird House")
 			|| eventType.startsWith("H.A.M. chest")
-			|| lootRecordType == LootRecordType.PICKPOCKET)
+			|| lootRecordType == LootRecordType.PICKPOCKET
+			|| eventType.endsWith("lockbox"))
 		{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -246,6 +246,13 @@
 	private static final String TEMPOROSS_LOOT_STRING = "You found some loot: ";
 	private static final int TEMPOROSS_REGION = 12588;
 
+	// Camdozaal Vault lockbox
+	private static final Map<Integer, String> LOCKBOX_IDS = new ImmutableMap.Builder<Integer, String>().
+		put(ItemID.SIMPLE_LOCKBOX_25647, "Simple lockbox").
+		put(ItemID.ELABORATE_LOCKBOX_25649, "Elaborate lockbox").
+		put(ItemID.ORNATE_LOCKBOX_25651, "Ornate lockbox").
+		build();
```
Nightfirecat, line 254: This can be stored as an immutable list of item IDs, and the event name can be set using `MenuOptionClicked#getMenuTarget()` instead of doing this lookup. As-is, this doesn't serve any purpose.

weiyih:
Updated

Adam-:
@weiyih can you retest this now i've rebased it?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
Adam-: while this normally is true, for items it seems that getMenuTarget() is actually just the option too.

Adam-:
nvm, I was able to get it tested

PR #13566, opened 2021-05-06
loot tracking for the Supply crates from Mahogany Homes

weiyih:
Tracking supply crates from Mahogany Homes opened with the Open menu option.

Note: This will not track crates opened with the Open-All menu options. That option opens all the crates on the same tick and rewards the player in one go.

unknown-1


Adam-:
added in 8760471818964fed5b8a34010b799d9816b408f4

PR #13570, opened 2021-05-06
item charges: add Kharedst's Memoirs

molo-pl:
Adding tracking of item charges for Kharedst's Memoirs.
Destroying and reobtaining the Memoir does not remove charges.
Below a few screenshots of tested scenarios, in order: using the memoir, adding charges and checking charges.
(Note 1: sometimes the word is plural even for 1 charge.)
(Note 2: checking when empty produces the same chat message as when using when empty.)

screen-1
screen-2
screen-3


Nightfirecat:
Needs rebase

molo-pl:
Rebased. Also added support for increasing charges with torn pages:

add-torn-page

molo-pl:
No longer works after introduction of Book of the Dead, some messages have changed, some things which used to be widgets turned into chat messages, etc. Also does not support Book of the Dead. Requires general rework.

PR #13589, opened 2021-05-09
Split ground item minimum threshold into stackable item and non-stackable item 

ruvane3:
This lets players set a different, lower, threshold for stackable items. Since they don't take up the same inventory space as non-stackable items it can be useful to have a different threshold for hiding those items.

This change defaults the new threshold for stackable items to the currently set threshold so it will have no affect on what is hidden and shown until a user explicitly changes the two thresholds. The only problem with one default depending on another is resetting the config needs to be done in an order where we don't reset the dependent value first. To solve this, this change introduces 'resetOrder' which is used to order the methods when resetting them.


ruvane3:
Is there something I need to do? First time contributing sorry if im missing something

Adam-:
No. I just haven't looked at this yet. Sometimes it takes me a few months to get to PRs since there are so many.

Looking at this now, I'm really not liking this reset order thing, usually to do something like this we would just use value == 0 as disabled and then anything >0 as enabled.

ruvane3:
Totally fair enough - let me know if you prefer it as value == 0 for disabled and I can update it to that way sometime this week. Also good to know on schedule.

Adam-:
I do prefer value==0

roflmuffin:
@ruvane3 Let me know if you would like me to pick this up and continue work on it.

armantark:
Would love for this feature to be looked at again.

PR #13618, opened 2021-05-19
Add menu entry swap option to swap Pick/Remove with Inspect for crop protection flowers and scarecrow

### LINKED ISSUE #13617: Menu Entry Swapper Suggestion - Swap the 'Pick' option with 'Inspect' on flowers
Issue:
Currently it is easy to misclick a flower used for protecting crops from disease and pick it out, resulting in losing the protection and having to replant said flower.

Solution:
Add the ability to swap the 'Pick' option with 'Inspect' to stop accidentally picking flowers.

Additional context:
Option should exclude Limpwurt Roots and Woad Seeds as they only provide crops, without protecting other crops from disease (so there is no reason not to pick them).

ThePharros:
closes #13617

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -672,4 +672,15 @@ public interface MenuEntrySwapperConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "swapCropProtection",
+		name = "Crop Protection",
+		description = "Swap Pick/Remove with Inspect on crop protectors in flower patches",
+		section = objectSection
+	)
+	default boolean swapCropProtection()
+	{
+		return false;
+	}
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -672,4 +672,15 @@ default boolean swapTemporossLeave()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "swapCropProtection",
+		name = "Crop Protection",
```
Nightfirecat, line 678: This isn't very clear without mentioning that it actually applies to flowers. I'd make this change for the config name, but everything else should be fine.
```suggestion
		name = "Crop Protecting Flowers",

```

ThePharros: I went with flowers at first but was unsure if that's technically fitting since it applies to scarecrows too. Should it still say flowers?

Hydrox6:
This has been superseded by 92aac53f4273e7944c3da767637e083847c03808

PR #13622, opened 2021-05-21
update menuentryswapper to include flower patch swap

### LINKED ISSUE #13617: Menu Entry Swapper Suggestion - Swap the 'Pick' option with 'Inspect' on flowers

bjosullivan:
closes #13617


bjosullivan:
closes #13617

Hydrox6:
This has been superseded by 92aac53f4273e7944c3da767637e083847c03808

PR #13625, opened 2021-05-23
Add option to Screenshot plugin for new collection log entries

### LINKED ISSUE #13673: Add an option to automatically screenshot new collection log items
Since there's an option to display new collection log items as a chat message in osrs, I thought that it would be neat if runelite was able to automatically take a screenshot when that happens.

I hope whoever is reading this has a nice day!

ThePharros:
Closes #13673

Takes a screenshot when the new collection log entry `ChatMessage` appears 

(e.g. `Chat message type GAMEMESSAGE: New item added to your collection log: <col=ff0000>Partial note</col>`)

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -256,7 +256,7 @@ public interface ScreenshotConfig extends Config
 	@ConfigItem(
 		keyName = "collectionLogEntries",
 		name = "Screenshot collection log entries",
-		description = "Configures whether or not screenshots are automatically taken of new collection log entries",
+		description = "Take a screenshot when completing an entry in the collection log",
 		position = 18,
 		section = whatSection
 	)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -254,10 +254,22 @@ public interface ScreenshotConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "collectionLogEntries",
+		name = "Screenshot collection log entries",
+		description = "Configures whether or not screenshots are automatically taken of new collection log entries",
+		position = 18,
+		section = whatSection
+	)
+	default boolean screenshotCollectionLogEntries()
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
```diff
@@ -253,11 +253,23 @@ default boolean screenshotHighGamble()
 		return false;
 	}
 
+	@ConfigItem(
+		keyName = "collectionLogEntries",
+		name = "Screenshot collection log entries",
+		description = "Configures whether or not screenshots are automatically taken of new collection log entries",
```
Nightfirecat, line 259:
```suggestion
		description = "Take a screenshot when completing an entry in the collection log",
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -456,6 +456,13 @@ public void onChatMessage(ChatMessage event)
 				takeScreenshot(fileName, "Duels");
 			}
 		}
+
+		if (config.screenshotCollectionLogEntries() && chatMessage.startsWith("New item added to your collection log: "))
+		{
+			String entry = chatMessage.substring(51, chatMessage.length() - 6);
```
Nightfirecat, line 462: I'd prefer `"New item added to your collection log: "` be made a constant, then this line can be changed to:

```
String entry = Text.removeTags(chatMessage).substring(COLLECTION_LOG_TEXT.length);
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
ThePharros: I was on the fence of how to format the description since most use the "configures whether" format, while more recent ones use the "take a screenshot" format. I have no issue going with either and can easily commit the suggested change, but I think for consistency sake there should be a consistency pass on the descriptions in the plugin in a separate PR.

Nightfirecat: Generally I do not like "Configure whether" wording, as it's clear the user is already viewing plugin configuration at the point this would be displayed. I understand this is not consistent across options, but new entries at least should be better-worded.

PR #13626, opened 2021-05-23
Restrict wintertodt start notifs to wt regions

LlemonDuck:
Varbit `WINTERTODT_TIMER(7980)` is checked to send the player notifications on wintertodt round starts, but this varbit is set in lots of locations around the world that are not wintertodt, causing extraneous start notifications. This limits the notifications to only fire when the player is at wintertodt.

After playing and leaving Wintertodt, I received round start notifications multiple times, including while bank standing at the GE, fishing at Otto's Grotto, and training agility. At no point between these notifications did I revisit the wintertodt area.

From the [varbit map here](https://mejrs.github.io/osrs.html?varbit=7980&m=-1&z=0&p=0&x=2084&y=3647), we can see that the varbit is set *everywhere*, even when not in use. I was unable to find, in my rudimentary look, any other references to/uses of the varbit. 

I am not entirely happy with this approach to solve the problem, since players restocking at their POH/some other location could miss out on notifications. However, I was unable to come up with a better alternative.

I could also use some pointers on how to properly update the unit test, if my current attempt isn't great (I suspect it isn't).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
@@ -89,6 +89,7 @@ import net.runelite.client.util.ColorUtil;
 public class WintertodtPlugin extends Plugin
 {
 	private static final int WINTERTODT_REGION = 6462;
+	private static final int WINTERTODT_CAMP_REGION = 6461;
 
 	@Inject
 	private Notifier notifier;
@@ -129,6 +130,7 @@ public class WintertodtPlugin extends Plugin
 	private Instant lastActionTime;
 
 	private int previousTimerValue;
+	private boolean notifiedOutsideWt;
 
 	@Provides
 	WintertodtConfig getConfig(ConfigManager configManager)
@@ -158,6 +160,7 @@ public class WintertodtPlugin extends Plugin
 		numKindling = 0;
 		currentActivity = WintertodtActivity.IDLE;
 		lastActionTime = null;
+		notifiedOutsideWt = false;
 	}
 
 	private boolean isInWintertodtRegion()
@@ -167,6 +170,16 @@ public class WintertodtPlugin extends Plugin
 			return client.getLocalPlayer().getWorldLocation().getRegionID() == WINTERTODT_REGION;
 		}
 
+		return false;
+	}
+	
+	private boolean isInCampRegion()
+	{
+		if (client.getLocalPlayer() != null)
+		{
+			return client.getLocalPlayer().getWorldLocation().getRegionID() == WINTERTODT_CAMP_REGION;
+		}
+		
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


mfloob:
Nobody restocks for Wintertodt at their POH/some other location.

LlemonDuck:
I restock at my POH pool each round 😄

mfloob:
Regardless this seems silly to restrict all users just for your (silly) use-case. Notifying everywhere seems intended and your idea fits better as an additional feature like a checkbox.

Hannah-GBS:
>  Notifying everywhere seems intended

It's a bug.

LlemonDuck:
This is a PR to fix a bug that caused WT start notifications to play repeatedly for hours after leaving the region. It was only fixed by restarting the client, which is a bug, not an intended feature.

> I am not entirely happy with this approach to solve the problem, since players restocking at their POH/some other location could miss out on notifications. However, I was unable to come up with a better alternative.

I am happy to hear alternative ideas on how to implement this without removing the usefulness of the notifications. I imagine for most users this change won't impact them, but I made this PR aware of the potential negatives.

mfloob:
> > Notifying everywhere seems intended
> 
> It's a bug.

Thanks for your constructive opinion.

Cyborger1:
> I am happy to hear alternative ideas on how to implement this without removing the usefulness of the notifications. I imagine for most users this change won't impact them, but I made this PR aware of the potential negatives.

I can think of a way, you could keep track of if the notification has been received since last visiting the WT regions. If the player doesn't go back after getting a notification while outside of WT, you can stop showing it for every subsequent times.

LlemonDuck:
> I can think of a way, you could keep track of if the notification has been received since last visiting the WT regions. If the player doesn't go back after getting a notification while outside of WT, you can stop showing it for every subsequent times.

I like this. Marking as draft until I get a chance to implement it.

Cyborger1:
Here's a rough draft, but I think something like this would do the trick

```java
// Instance variable
private boolean lastNotificationOutsideWT ;
...
private void reset()
{
	inventoryScore = 0;
	totalPotentialinventoryScore = 0;
	numLogs = 0;
	numKindling = 0;
	currentActivity = WintertodtActivity.IDLE;
	lastActionTime = null;
	// Add this in reset
	lastNotificationOutsideWT = false;
}
...
// In onVarbitChanged()
if (timerValue != previousTimerValue)
{
	if (isInWintertodtRegion() || isInCampRegion())
	{
		// Reset notification guard
		lastNotificationOutsideWT = false;
	}
	else
	{
		if (lastNotificationOutsideWT )
		{
			// Second notification trigger in a row outside of WT, don't notify
			return;
		}
		lastNotificationOutsideWT = true;
	}

	// Notification code
...
```

LlemonDuck:
Done, thanks @Cyborger1 for speeding that up. Ready for review once again.

LlemonDuck:
Rebased and cleaned up unit test ^

Adam-:
fixed in 7e9cb6c88e5efe72963bb58ed936dc45d34088a3

PR #13627, opened 2021-05-23
Rebuild split PM box on history clear and login message removal

emerald000:
When the "Clear Messages" option was used or whenever a login/logout notification message was removed, BUILD_CHATBOX was run which only rebuilds the chatbox.
By running SPLITPM_CHANGED instead, both the chatbox and the split PM box are rebuilt.


Adam-:
fixed in b5375ae48756555bec35ee043701d93687c366ae

PR #13636, opened 2021-05-26
Poh burner notification

JPedrys:
- Added notification when the burner is 10 seconds from entering the random phase (when the timer could burn out at any time).
- Slight change to the random timer calculation subtract 1 from fmLevel. Wiki formula here: 200 + firemaking level + A ticks where A is a random number between 0 and firemaking level - 1
- Added a config option to toggle the burner expiration notification


Adam-:
I've fixed most of this in 44134dc668158da1244b10f8ab1f04e9950e350b. If you want to rebase this for the notif that is fine.

PR #13647, opened 2021-05-26
Reset spec counter when leaving instanced fights

LlemonDuck:
When leaving Vorkath or Alchemical Hydra fights, the spec counter does not reset, despite the boss resetting their stats. This could cause the spec counter to add the player's next trip spec to that, giving an inaccurate total spec count. This clears that counter when leaving the vork + hydra region instances.

Needs to be tested at hydra, since I don't have access to that fight to test myself. Drafted until then.


LlemonDuck:
Confirmed to work at Hydra by @TheStonedTurtle, so removing Draft status.

Issue #10145, opened 2019-10-27
Remember Password


Aholicknight:
**Is your feature request related to a problem? Please describe.**
Not a problem.

**Describe the solution you'd like**
Regular OSRS allows you to save the email/username for your account. Another great addiction to this would be a plugin to save your password if you want.
**Describe alternatives you've considered**
Save the password in notes so you can easily access it.

**Additional context**
None.

Hydrox6:
Storing passwords in the client is a really really bad idea. If you want quick logging in, the `Login Screen` plugin already allows you to paste in passwords, allowing easy use of a password manager such as KeePass.

coreybruce:
Would be fine if you did two step than either way it doesn't matter

binury:
> Storing passwords in the client is a really really bad idea.

This would be the same behavior as the official mobile client

binury:
> Storing passwords in the client is a really really bad idea

This feature would mirror the current behavior of the official mobile client. Could salt the password locally.... or trust the player to secure their own PC.

Adam-:
> > Storing passwords in the client is a really really bad idea
> 
> This feature would mirror the current behavior of the official mobile client. Could salt the password locally.... or trust the player to secure their own PC.

The official mobile client doesn't store passwords

Jin-Jiyunsun:
theres a plugin on the hub, "KeePassXC", it auto fills for you and if setup correctly means you only need to put the master password into keepassxc once per startup on your pc

coreybruce:
In Linux for example it could be stored in your keyring and remembered that way user wide

Aholicknight:
> Storing passwords in the client is a really really bad idea

You can encrypt the password so they are not stored in cleartext

Jin-Jiyunsun:
99.99% chance this isn't going to happen. Use the keepass plugin if you want something close to remembering passwords, great thing about the plugin in you can have multiple accounts saved into keepass and clicking on them in the plugins sidebar will fill username and password for you

coreybruce:
@Hydrox6 not sure why you thumbs down my comment, it's not like anyone else is going to know someone else's password and even if they did they wouldn't have access to it without 2step and they could change their password ontop of that using a Password manager like Bitwarden to generate a decent password for example making it a non issue. Not to mention if they client did save the password or with Mac/Linux or even WIndows if you used a password manager like Linux and Mac has built in it's not like the password wouldn't be encrypted and not stored as plain text, that would be stupid.

[...]

PR #13652, opened 2021-05-27
Remember Login Details

ShaneeexD:
Have runelite able to remember both login details, and keep them unchanged even if other accounts/clients have been used


Hydrox6:
This is not something we want at all (#10145), and have rejected before (runelite/plugin-hub#249)

Adam-:
And, also, this PR would send me everyone's password who had the plugin on and was also logged in, which is very bad.

PR #13661, opened 2021-05-29
rl-client: never cache 4xx requests

abextm:


runelite-client/src/test/java/net/runelite/client/OkHttpCacheSanityTest.java
```diff
@@ -0,0 +1,125 @@
...
+	 * which does hit the upstream, returning 304 Not Modified. Since there
+	 * is no body OkHttp serves the cached 404 as up-to-date. Better yet, since
+	 * there is now an ETag in the 304 response it requests subsequent updates
+	 * with the ETag, which will always 304 Not Modified, causing clients in
+	 * this state to get incorrect 404s until the ETag changes.
+	 */
+	@Test
+	public void testCacheSanity() throws IOException, InterruptedException
+	{
+		MockWebServer server = new MockWebServer();
```
Adam-, line 71: This must be shutdown. The few places we use mockwebserver in the codebase do this with a junit before/after, however, MockWebserver is a junit `ExternalResource` and so I believe you can instead do:
```java
@Rule
public MockWebServer server = new MockWebServer();
```

Adam-:
Your test appears to pass even with the new code commented.

PR #13667, opened 2021-05-30
Add additional Motherlode Mine GroupId and WidgetInfo values

ThePharros:
These will be used in a future PR, however, it's most likely best to add such values separately as they modify core classes of the main RL branch, hence this PR.


ThePharros:
As per Discord discussion, it has been clarified to best include such changes alongside the relevant plugin:

image

Closing this PR.

PR #13674, opened 2021-06-01
Update the "Highlight" capabilities to allow Hull or Tile on slayer

dan-drews:
Similar to how NPC Indicators works, allow users to choose to highlight either the Hull or the tiles of the opponent, as opposed to only allowing the Hull to be highlighted.


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
```diff
@@ -94,11 +94,11 @@ default int statTimeout()
...
 		keyName = "highlightTargets",
 		name = "Highlight Targets",
-		description = "Highlight monsters you can kill for your current slayer assignment"
+		description = "Highlight the hull or tiles of monsters you can kill for your current slayer assignment"
 	)
-	default boolean highlightTargets()
+	default SlayerTaskHighlightMode highlightTargets()
 	{
-		return false;
+		return SlayerTaskHighlightMode.NONE;
```
dan-drews, line 101: Can anything be done here to allow me to pull in their previous default, since this would be changing the type stored on the config? If it was previously `true` we would want it to be `SlayerTaskHighlightMode.HULL`

Adam-: If by previous default you mean previous setting, then, you could write code to migrate from the old value to the new value (assuming they have two different keys), which we will do occasionally. However this feature is little-used and so I probably wouldn't bother with it. You have set the default value to be the same as the existing default value (off) so this PR will not cause a behavior change.

If I recall correctly the config manager has code to deal with this (currently what you have is you are trying to unserialize a true/false into a boolean value, which is not valid) and I think it will force-reset the config value to its default.

dan-drews: So would I be fine to just leave it as is and have it force-reset?

Adam-: I'd have to check but I think so. If you run the client you should see it hit this `log.debug("Setting default configuration value for {}.{} to {}", group.value(), item.keyName(), defaultValue);` and override the default value on the config after being unable to deserialize it.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/TargetClickboxOverlay.java
```diff
@@ -75,11 +88,28 @@ private void renderTargetOverlay(Graphics2D graphics, NPC actor, Color color)
...
 	}
+
+	private void renderTargetTiles(Graphics2D graphics, NPC actor, Color color)
+	{
+		NPCComposition npcComposition = actor.getTransformedComposition();
+		if (npcComposition != null)
+		{
+			int size = npcComposition.getSize();
+			LocalPoint lp = actor.getLocalLocation();
+			Polygon tilePoly = Perspective.getCanvasTileAreaPoly(client, lp, size);
```
Adam-, line 102: this can be `null`

runelite-client/src/main/java/net/runelite/client/plugins/slayer/TargetClickboxOverlay.java
```diff
@@ -75,11 +88,28 @@ private void renderTargetOverlay(Graphics2D graphics, NPC actor, Color color)
...
+			renderPoly(graphics, tilePoly, color);
+		}
+	}
+
+	private void renderPoly(Graphics2D graphics, Shape shape, Color color)
+	{
+		graphics.setColor(color);
+		graphics.setStroke(new BasicStroke(2));
+		graphics.draw(shape);
+		graphics.setColor(new Color(color.getRed(), color.getGreen(), color.getBlue(), 20));
```
Adam-, line 112: ... what is this?

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerTaskHighlightMode.java
```diff
@@ -0,0 +1,8 @@
+package net.runelite.client.plugins.slayer;
```
Adam-, line 1: missing license header. you could also just put this enum as an inner class on the config and that is also acceptable.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/TargetClickboxOverlay.java
dan-drews: This was taken from `renderPoly` in `NpcSceneOverlay.java`

I'll adjust this to match what was here before instead

runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
dan-drews: Yup, I did in fact see that, should be good

Nightfirecat:
I would prefer this adopt the same set of configs as is available in the NPC Indicators plugin, allowing any combination of hull/tile/southwest tile to be highlighted.

* Render style:
  - [ ] Highlight hull
  - [ ] Highlight tile
  - [ ] Highlight south west tile

dan-drews:
@Nightfirecat Just spit-balling here, could we combine this config rather than duplicating? Either having it directly use the NPC Indicators options, or, moving the `Highlight Slayer Monsters` config to NPC Indiciators?

Nightfirecat:
I think if we were to have a co-dependency for these configs, we'd be better served handling slayer task highlighting to the NPC Indicators plugin instead.

Adam-:
added in c7b444d8926e994a9f2c32b80e69a0d3b227136e

PR #13691, opened 2021-06-04
Reset Wintertodt countdown notification on logout or world hop.

### LINKED ISSUE #13468: Logging in outside of Wintertodt's prison always causes a "Game is starting soon" notification
Even when a round of Wintertodt has already started and is underway, logging in while standing at the Wintertodt camp will always trigger a "Game is starting soon" notification. This should only be triggered when the timer is nearly finished counting down, never upon login.

tanlines:
Closes #13468 

Logging out while timer is currently ticking down, the varbit value is saved e.g. 30.
When logging in while wintertodt is running, the varbit is reset to 0, triggering a notification since the **previousTimerValue** was 30.
Issue also occurs when world hopping from a world waiting for wintertodt to a world currently running.

The varbit has to be explicitly set to 0 along with **previousTimerValue** since varbitChanged events are firing continuously, resetting **previousTimerValue** to the varbit value.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
@@ -65,6 +65,7 @@ import net.runelite.api.Player;
 import net.runelite.api.Varbits;
 import net.runelite.api.events.AnimationChanged;
 import net.runelite.api.events.ChatMessage;
+import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.VarbitChanged;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/wintertodt/WintertodtPlugin.java
```diff
@@ -219,6 +220,19 @@ public void onVarbitChanged(VarbitChanged varbitChanged)
...
 
+	@Subscribe
+	public void onGameStateChanged(GameStateChanged event)
+	{
+		switch (event.getGameState())
+		{
+			case HOPPING:
+			case LOGGING_IN:
+				previousTimerValue = 0;
+				client.setVarbit(Varbits.WINTERTODT_TIMER, 0);
```
Nightfirecat, line 231: We really shouldn't be setting the var ourselves. A better way to handle this would be to add a check in the varbit change subscriber to prevent writing to the field when the game state is `LOADING`.

Adam-:
fixed in 7e9cb6c88e5efe72963bb58ed936dc45d34088a3

PR #13692, opened 2021-06-04
Fix stamina restore with Ring of Endurance

### LINKED ISSUE #13560: Hovering stamina pot while wearing Ring of Endurance shows wrong regen amount
**Describe the bug**
While wearing a Ring of Endurance run energy is increased by +40 when sipping a stamina pot and +20 without the ring. When the ring is equipped Runelite still shows +20 when hovering a pot.

**To Reproduce**
1. Wear charged Ring of Endurance
2. Hover any stamina pot in inventory to see run regen

**Expected behavior**
Display +40 instead of +20 when hovering stamina pot while wearing charged Ring of Endurance

**Screenshots**
image

 - OS: Windows 10
 - RuneLite version: 1.7.6

tanlines:
Closes #13560 

Basically duplicating the logic for prayer potions which checks for Ring of Gods and swapping the ring.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
@@ -55,5 +55,4 @@ public class StaminaPotion extends StatBoost
 		}
 		return 20;
 	}
-
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2016-2018, Adam <Adam@sigterm.info>
+ * Copyright (c) 2021, Tanlines <tanlines@outlook.com.au>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2016-2018, Adam <Adam@sigterm.info>
```
Nightfirecat, line 2: Unless you're adam in disguise, you should change or append this with your own copyright.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
```diff
@@ -0,0 +1,59 @@
...
+
+	@Override
+	public int heals(Client client)
+	{
+		boolean hasEquippedChargedRingOfEndurance = false;
+
+		ItemContainer equipContainer = client.getItemContainer(InventoryID.EQUIPMENT);
+		if (equipContainer != null)
+		{
+			Item ring = equipContainer.getItem(RING_SLOT);
```
Nightfirecat, line 53: `RING_SLOT` is only used once; this can be changed to the following, and the constant can be removed.
```suggestion
			Item ring = equipContainer.getItem(EquipmentInventorySlot.RING.getSlotIdx());
```

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/StaminaPotion.java
```diff
@@ -0,0 +1,59 @@
...
+			hasEquippedChargedRingOfEndurance = ring != null && ring.getId() == ItemID.RING_OF_ENDURANCE;
```
Nightfirecat, line 54: `hasEquippedChargedRingOfEndurance` is not needed. Instead, you can simply `return 40;` here if the condition is matched, and `return 20;` as the base case for this function, eg.

```java
@Override
{
	final ItemContainer equipment = client.getItemContainer(InventoryID.EQUIPMENT);
	if (equipment != null)
	{
		final Item ring = equipment.getItem(EquipmentInventorySlot.RING.getSlotIdx());
		if (ring != null && ring.getId() == ItemID.RING_OF_ENDURANCE)
		{
			return 40;
		}
	}
	return 20;
}
```

Nightfirecat (changes requested):

PR #13693, opened 2021-06-04
Disable bank search hotkey while entering search query

### LINKED ISSUE #13608: Bank search shortcut is always enabled while bank is open.
**Describe the bug**
Bank hotkey for search item is not disabled while typing in chat. For example, WASD camera movement does not work when you've yet to press enter to type, are replying to a pm, or searching in the bank. 

**To Reproduce**
1. Set bank search shortcut to a key without a modifier (in this case it was S)
2. Hit the shortcut key in any context while the bank is open.

**Expected behavior**
The shortcut key should be disabled while typing (in bank search or elsewhere) in client. This is not the case.

**Screenshots**
Gfycat example: https://gfycat.com/soreillinformedferret

 - OS: Windows 10
 - RuneLite version: 1.7.7
 - Launcher version: 1.5.1

**Additional context**
None, ty :)

tanlines:
Closes #13608 

Players that reenter the hotkey to reset the search query may be inconvenienced as it will now require pressing Enter -> hotkey to reset the search query, one extra keypress.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
@@ -126,6 +126,12 @@ public class BankPlugin extends Plugin
 					return;
 				}
 
+				Widget chatboxFullInput = client.getWidget(WidgetInfo.CHATBOX_FULL_INPUT);
+				if (chatboxFullInput != null && !chatboxFullInput.isHidden())
+				{
+					return;
+				}
+
 				log.debug("Search hotkey pressed");
 
 				bankSearch.initSearch();


runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -126,6 +126,12 @@ public void keyPressed(KeyEvent e)
 					return;
 				}
 
+				Widget chatboxFullInput = client.getWidget(WidgetInfo.CHATBOX_FULL_INPUT);
+				if (chatboxFullInput != null && !chatboxFullInput.isHidden())
```
Adam-, line 130: Not sure if this change is correct in general, but, this fails if run with `-ea` due to being called off the client thread.

Adam- (changes requested):

Issue #13670, opened 2021-05-31
Special attack counter disappears when hydra changes phases despite stats staying drained (unlike nightmare)


ColemanDunn:
Please check if your issue is not a duplicate by [searching existing issues](https://github.com/runelite/runelite/search?type=Issues)

**Describe the bug**
The counter disappears when it changes forms despite the drained stats being carried over (but the drained stats do go up by 1 each cycle/new form, but that could be due to the regular timer for NPCs regaining their stats over time, if that is a thing) from the previous form. Confirmed multiple times with monster examine. 

**To Reproduce**
1. Go to hydra
2. Spec hydra and see the counter plugging pop up for your BGS hit
3. Get to second/any other form
4. Observe the counter go away

**Expected behavior**
The counter should stay to remind you that you have spec'd, and what you spec'd, in the event that you may want to spec again later in the fight. Also when checking and it is gone gives the impression that you did not spec yet that kill. 

**Environment (please complete the following information):**
 - OS: macOS
 - RuneLite version: 1.7.10
 - Launcher version: 1.6.2

PR #13696, opened 2021-06-05
specialCounter: Adds all phases of Alchemical Hydra 

SkylerPIlot:
This should keep the counter up for all phases of alchemical hydra


SkylerPIlot:
This should fix Issue 
#13670

PR #13702, opened 2021-06-07
skill calc: added urium remains to the list along with 50% bonus

### LINKED ISSUE #13699: [Skill Calculator] Missing Urium Shades
Burning urium shades is not included in the skill calculator for prayer.

SkylerPIlot:
Closes #13699


Nightfirecat:
In the short term, this will work fine. Particularly as the number of various XP multipliers increases, and to possibly accommodate temporary or variable multipliers such as in leagues and DMM, I'd like to move skill calculator definitions back into Java from JSON. This will allow us to create a system where we can have a more complex and correct multiplier system (instead of only `ignoreBonuses`), and comes with the side effect of letting us reference ItemID instead of hard-coding all the icon references.

PR #13704, opened 2021-06-07
!pb Barb Assault

SkylerPIlot:
This is currently a draft. I'd just like a few more eyes on it to make it easier to merge in its final form.
Currently adds:
Saving of current best time into an overall pb and a role-specific pb into the config file at the end of w10
new config setting in Barbarian assault to check if they'd like to save into a role-specific pb (default false)
Fixed display of timer from changing to hh:mm when a run lasted more than 1hr instead staying as mm:ss
and the usage of these commands.
!pb barb assault (with a few more cases in chat commands plugin)
!pb ba healer
!pb ba collector
!pb ba defender

Currently, they all save correctly and logic is being double-checked to make sure there isn't some hidden little bug.
Feel free to build yourself and play around with it.

Gonna mark it as a draft, for now, maybe get some new eyes on it and some review of code then when finalized I'll squash commits and do a final pr


SkylerPIlot:
Reasons for adding is for the general person playing Barbarian assault they get a torso and leave and get lvl5 all for diaries and leave and therefore would never need/want to use the current !gc command for gamble count. These new commands would allow them to be able to show others how well(or poorly) they performed in any role as well as an overall personal best.

Then for the people who do leeches/ba community in general the !gc command is nice to show how much time you've put into the minigame, but at a certain point, you can't really determine the skill level of those players just from the !gc command. An example being someone showing that they have 700gc from just playing in 306 vs someone who does leeches and is at 300gc this would be a better way of showcasing the skill level of other players especially if they only have 700gc from a single role.

Everyone asked from the ba community has been quite receptive and would like this change, and for the average player it gives them some functionality as well

I do know there might be hesitation on saving a personal best from something that can not be verified as it is not tracked serverside by Jagex and however is totally client-side, but my views on this are the people who know how to change them can change how they show up anyways.  And tbh you can also mess with how other people's !pb commands show up regardless with no editing of their config file. And for records in BA, there are already strict guidelines set with video evidence and timing so this would not affect that in any way.

Nightfirecat:
Looks like this was added in https://github.com/runelite/plugin-hub/pull/1562, so I'm closing this PR.

Issue #8702, opened 2019-05-01
NPC indicators "true position highlight"


buldosis:
At the moment, the NPC indicator "tile" highlight only highlights around the monsters character and doesnt show its true location of the monster (i.e. if it has long attack animations, its actual location is very wrong from the one thats shown on your screen if you are moving around).
The npc indicator "south west tile" highlights its true location but only the south west tile, would work on 1x1 npcs but not bigger, so this would be the same thing but would work for other npcs bigger than 1x1's

jachymb:
This would be very useful to have for PvM activities where you specifically want or specifically do not want to stand under the monster.

Nightfirecat:
We've had to remove this feature in the past ([see Adam's comment on this](https://github.com/runelite/runelite/pull/7748#issuecomment-636353341)) so we're not likely to add this feature again.

PR #13712, opened 2021-06-09
npc indicators: add outline width

Bird-Hands:
Added a server tile option and a stroke width config

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
@@ -109,6 +109,17 @@ public interface NpcIndicatorsConfig extends Config
 
 	@ConfigItem(
 		position = 5,
+		keyName = "borderWidth",
+		name = "Border Width",
+		description = "Width of the highlighted NPC border"
+	)
+	default double borderWidth()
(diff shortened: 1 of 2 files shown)


Nightfirecat:
We are not interested in adding "true position" highlighting (see #8702), however a customizable outline size would be acceptable. If you want to remove the former feature, we'd be interested in adding that.

Bird-Hands:
My bad I didn't see that one. I can't update it right now but I'll try to send the update in tomorrow.

runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
```diff
@@ -107,8 +109,22 @@ default Color getHighlightColor()
...
+	@Range(
+		max = MAX_STROKE_SIZE
+	)
 	@ConfigItem(
 		position = 5,
+		keyName = "strokeSize",
+		name = "Outline Width",
+		description = "Width of the NPC highlight"
+	)
+	default int getStrokeSize()
```
Nightfirecat, line 121: This can follow the same example as d4078a95553bee262f0d5ca032cea49dfca5ca3e and be a `double` config.

Bird-Hands:
My branch got kind of broken but I think that works. not too certain though

PR #13715, opened 2021-06-10
Fix toJagexName removing double spaces in some cases

Cyborger1:
**Note:** A blank character in this context represents any of a space (either normal or non-breaking), `_` or `-`. I am also replacing spaces with `+` in this message because GitHub markdown appears to delete extra spaces in their one-liner code blocks.

While the game does not allow players to enter names containing trailing/leading blank characters or multiple blank characters in a row, there are seemingly ways around that (such as the infamous `+Spacehack`, but also `0++Crazy++0` (notice the double spaces). Trailing and leading blanks are currently handled correctly in all cases, but there are issues with blanks inside of the name.

Taking `Crazy` as an example here, in Jagex's database it would be distinct from, say `0+Crazy+0`, and any of the blanks should be able to be substituted to any other blank and it would count as an equivalent name. For example, `0__Crazy--0` is the same as `0++Crazy++0`.

The current implementation of `toJagexName()` will correctly keep `0++Crazy++0` the way it is, it will also, for example, correctly convert `0+_Crazy-+0` to the same. However, unbroken strings of `_`s and `-`s are incorrectly converted to single spaces. `0--Crazy--0` incorrectly becomes `0+Crazy+0`, which is not an equivalent username.

This PR corrects the functionality of the method and adds tests to ensure it performs as expected.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/Text.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/Text.java
@@ -130,7 +130,7 @@ public class Text
 	 */
 	public static String toJagexName(String str)
 	{
-		return CharMatcher.ascii().retainFrom(str.replace('\u00A0', ' ')).replaceAll("[_-]+", " ").trim();
+		return CharMatcher.ascii().retainFrom(str.replaceAll("[\u00A0_-]", " ")).trim();
 	}
 
 	/**
--- a/runelite-client/src/test/java/net/runelite/client/util/TextTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/util/TextTest.java
@@ -56,4 +56,21 @@ public class TextTest
 		assertEquals("a <lt> b", Text.removeFormattingTags("a <lt> b"));
 		assertEquals("Remove no tags", Text.removeFormattingTags("Remove no tags"));
 	}
+
+	@Test
+	public void toJagexName()
+	{
+		assertEquals("lab rat", Text.toJagexName("lab rat"));
+		assertEquals("lab rat", Text.toJagexName("-lab_rat"));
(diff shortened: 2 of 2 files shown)


Nightfirecat (approved):

Adam-:
Are there any bugs this is fixing?

Cyborger1:
Hey Adam, we had talked about this in the Discord if I recall correctly. It's mostly because I use this to normalize names in the Bot Detector hub plugin, but this would actually fix a few potential issues in Core. Below are issues with this method that I could spot at a glance searching https://github.com/runelite/runelite/search?q=tojagexname.

* Friends Note may have some collision issues if a user has two players in their list that `toJagexName` maps to the same name.

* The Team Plugin may miscount nearby clan members as it does a `Text.toJagexName(member.getName())` into a player hashmap check `players.get(memberName)`. Although, I do question if this is correct as the map is populated with `players.put(Text.removeTags(player.getName()), player);`, so names with underscores and hyphens at all will always fail?

* Chat notifications may fail to stop highlighting on certain message type for these players. However, it seems there's more going on in that class that needs investigating, as it assumes `client.getLocalPlayer().getName()` only ever contains spaces, which is apparently not the case.

* Party Plugin may set the overlay name of one of these players' with the incorrect amount of spaces, however minor this may be.

* Chat Channels may add the wrong name to recent friends chat targets if the player name contains these double underscores or hyphens.

Adam-:
Do you have any actual examples of actual bugs that are reproducible that this fixes?

Cyborger1:
I had to dig around a lot to find an example of a name to show this off, but yes, this would fix this bug with Friends Note collisions:
2o9Xoz6lhQ

PR #13725, opened 2021-06-12
timers: add cannon repair timer

Hydrox6:
Adds a simple timer when your cannon breaks, showing the 10 minutes you have to repair it. Uses the icon for the Repair Kit used in the Dwarf Cannon quest.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -97,6 +97,7 @@ public class TimersPlugin extends Plugin
 	private static final String CANNON_PICKUP_MESSAGE = "You pick up the cannon. It's really heavy.";
 	private static final String CANNON_REPAIR_MESSAGE = "You repair your cannon, restoring it to working order.";
 	private static final String CANNON_DESTROYED_MESSAGE = "Your cannon has been destroyed!";
+	private static final String CANNON_BROKEN_MESSAGE = "<col=ef1020>Your cannon has broken!";
 	private static final String CHARGE_EXPIRED_MESSAGE = "<col=ef1020>Your magical charge fades away.</col>";
 	private static final String CHARGE_MESSAGE = "<col=ef1020>You feel charged with magic power.</col>";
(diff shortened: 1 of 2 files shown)


MarbleTurtle:
Confirming that the transparent color code the for the cannon is broken text is the same for both transparent and opaque

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -540,9 +541,25 @@ public void onChatMessage(ChatMessage event)
 			cannonTimer.setTooltip(cannonTimer.getTooltip() + " - World " + client.getWorld());
```
Nightfirecat, line 541: You've forgotten to remove this block since you moved it into L546-L551.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -540,9 +541,25 @@ public void onChatMessage(ChatMessage event)
...
-			removeGameTimer(CANNON);
+			if (message.equals(CANNON_FURNACE_MESSAGE) || message.contains(CANNON_REPAIR_MESSAGE))
+			{
+				removeGameTimer(CANNON_REPAIR);
+				TimerTimer cannonTimer = createGameTimer(CANNON);
+				cannonTimer.setTooltip(cannonTimer.getTooltip() + " - World " + client.getWorld());
+			}
+			else if (message.equals(CANNON_BROKEN_MESSAGE))
+			{
+				removeGameTimer(CANNON);
```
Nightfirecat, line 554: Is there any case where this is needed? This message should only be received after the original 25 minutes has already elapsed, at which point `CANNON` would have run its full time and be culled anyway.

Hydrox6: I think I just wanted to make sure that the timer actually gets removed, given the states shouldn't overlap, and a lot of the timers drift slightly over time. It does look nicer imo to have only one of the two at a time.

PR #13632, opened 2021-05-25
Added cost calculation for the construction calculator

pristit:
Added tooltip text detailing how many of each material is required.

I've edited the entries to only some of the actions in the skill resource json file, needs to be completed.

Possible to use the same actions and calculations to add a cost calculator for herblore.

image

PR #5096, opened 2018-08-24
add gp/xp support for Prayer, Firemaking, Construction, Smithing (WIP), Cooking (WIP)

astaninger:
#3869

PR #13726, opened 2021-06-12
skillcalculator: show material cost for actions that have them

### LINKED ISSUE #3869: add gp/xp to skill calculator
add gp/xp with ge-prices?

Hydrox6:
Closes #3869
Closes #5096
Closes #13632

Draft for the following reasons:
* I would like feedback on the naming in ff82f5c
* It currently is missing all the materials (it's just a PoC atm)
* The multi-selection box doesn't show cost at all if it's 0 (ie. there was no costs associated with all selected methods). I'm not sure if it should just show as 0 gp in that case instead.
* The materials are displayed in the tooltip, and currently only show the requirements for a single action. It should probably also show the total required for the overall actions, and the combined action display should do the same, but I'm not sure what that should look like. Feedback much appreciated.
* ~~I'm not sure how to deal with materials with variable input (eg. Shade Cremation, Lunar Magic processing spells)~~ I've got a method, it's just not pushed yet.
* There is currently no config for turning this off, and I would like to add one since it's not really useful for all players

How it currently looks:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/ItemManager.java
@@ -307,6 +307,36 @@ public class ItemManager
 	 */
 	public int getItemPriceWithSource(int itemID, boolean useWikiPrice)
 	{
+		return getItemPrice(itemID, false, useWikiPrice);
+	}
+
+	/**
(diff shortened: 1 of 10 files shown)


Hydrox6:
Closing until a PR for #14082 is implemented, as that will massively change the work needed

PR #13729, opened 2021-06-12
Xp Drops: Add option to override standard client xp drop colors

Zeusay:
Adds toggle option for player to choose to override the default client xp drop colors, allowing for more options/variety.

Does not affect the ability to change colors when using prayers.

# Before
image

# After
image

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
@@ -45,10 +45,18 @@ public interface XpDropConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "standardColor",
+		name = "Standard Color",
+		description = "XP drop color when no prayer is active",
+		position = 1
+	)
+	Color standardColor();
+
+	@ConfigItem(
 		keyName = "meleePrayerColor",
 		name = "Melee Prayer Color",
 		description = "XP drop color when a melee prayer is active",
-		position = 1
+		position = 2
 	)
 	default Color getMeleePrayerColor()
 	{
@@ -59,7 +67,7 @@ public interface XpDropConfig extends Config
 		keyName = "rangePrayerColor",
 		name = "Range Prayer Color",
 		description = "XP drop color when a range prayer is active",
-		position = 2
+		position = 3
 	)
 	default Color getRangePrayerColor()
 	{
@@ -70,7 +78,7 @@ public interface XpDropConfig extends Config
 		keyName = "magePrayerColor",
 		name = "Mage Prayer Color",
 		description = "XP drop color when a mage prayer is active",
-		position = 3
+		position = 4
 	)
 	default Color getMagePrayerColor()
 	{
@@ -81,12 +89,11 @@ public interface XpDropConfig extends Config
 		keyName = "fakeXpDropDelay",
 		name = "Fake Xp Drop delay",
 		description = "Configures how many ticks should pass between fake XP drops, 0 to disable",
-		position = 4
+		position = 5
 	)
 	@Units(Units.TICKS)
 	default int fakeXpDropDelay()
 	{
 		return 0;
 	}
-
 }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
```diff
@@ -83,10 +83,33 @@ default Color getMagePrayerColor()
 		description = "Configures how many ticks should pass between fake XP drops, 0 to disable",
 		position = 4
 	)
-	@Units(Units.TICKS)
```
Adam-, line 86: this is wrong.

runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
```diff
@@ -83,10 +83,33 @@ default Color getMagePrayerColor()
...
 	default int fakeXpDropDelay()
 	{
 		return 0;
 	}
 
+	@Units(Units.TICKS)
```
Adam-, line 91: this is also wrong.

runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
```diff
@@ -83,10 +83,33 @@ default Color getMagePrayerColor()
...
+		return false;
+	}
+
+	@ConfigItem(
+		keyName = "standardColor",
+		name = "Standard Color",
+		description = "XP drop color when no prayer is active",
+		position = 6
+	)
+	default Color standardColor()
```
Adam-, line 110: You can just do `Color standardColor()` and if it is null not apply it, which removes the need for a separate on/off toggle.

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
Zeusay: This was in the original file, was it wrong to move it or should it just not have been there in the first instance?

Adam-: It was correct originally, you are requesting its removal, which is not correct.

runelite-client/src/main/java/net/runelite/client/plugins/experiencedrop/XpDropConfig.java
Zeusay: I tried the following code snippets...
### Config:
```java 	
        @ConfigItem(
		keyName = "standardColor",
		name = "Standard Color",
		description = "XP drop color when no prayer is active",
		position = 6
	)
	default Color standardColor()
	{
		return null;
	}
```
### Plugin:
```java Color standardColor = config.standardColor();

		if (standardColor != null)
		{
			widget.setTextColor(standardColor.getRGB());
		}
		else
		{
			EnumComposition colorEnum = client.getEnum(EnumID.XPDROP_COLORS);
			int defaultColorId = config.standardColor().getRGB();
			int color = colorEnum.getIntValue(defaultColorId);
			widget.setTextColor(color);
		}
```

But I couldn't figure out how the user would completely disable or be able to reset the Standard Color. I have set it to null by default, but even when resetting the plugin to it's default config, the Standard Color does not reset.

I have also removed the Override Color bool toggle as you suggested.

Any idea how I might be able to achieve this?

Cyborger1: It looks like you're not doing the `else` part correctly. In fact, in your snippet, it would cause a NullPointerException on `int defaultColorId = config.standardColor().getRGB();`. The `else` part should be the unchanged old logic.

Try something like this.
```java
private void resetTextColor(Widget widget)
{
	Color standardColor = config.standardColor();
	if (standardColor != null)
	{
		widget.setTextColor(standardColor.getRGB());
	}
	else
	{
		EnumComposition colorEnum = client.getEnum(EnumID.XPDROP_COLORS);
		int defaultColorId = client.getVar(Varbits.EXPERIENCE_DROP_COLOR);
		int color = colorEnum.getIntValue(defaultColorId);
		widget.setTextColor(color);
	}
}
```

Adam-: I've pushed a fixed commit with this fixed.

PR #13734, opened 2021-06-14
mining: Account for barronite veins in Ruins of Camdozaal

acdvs:
Adds pie progress timers to depleted barronite veins in the Ruins of Camdozaal.


Adam-:
added in #14084

PR #13744, opened 2021-06-16
clues: Add the new cryptic clue at Giant's Den

petter:
Ref: #13745


Adam- (changes requested):
please don't unnecessarily reformat the entire file.

Issue #11969, opened 2020-06-24
Player Indicators: Highlight ToB Party


KakronPrime:
**Is your feature request related to a problem? Please describe.**
Player Indicators are very useful for verzik part 1. But if you have "Highlight other" disabled you wouldn't see their names.

**Describe the solution you'd like**
Add option to highlight ToB party members

Nightfirecat:
Turn on "Highlight other" then? I don't see this being a situation where it is worth us adding a special case for highlighting.

KakronPrime:
Turning on highlight other during the raid means we have to turn it off after. If you're doing 416 tob (ffa world, quite a lot of people in lobby) it becomes a mess and its hard to identify when your team are pre-potting.

Issue #13104, opened 2021-01-24
raid party player indicator


thashaman:
Right now, while doing TOB  in popular worlds, it is very hard to tell where are your team players. it would be nice to be able to highlight their name and tile below them.

thanks!

PR #13765, opened 2021-06-20
Tob tem member highlight

geheur:
* Modifies Player Indicators to be able to indicate tob team members.
   * This is controlled by the "Highlight team members" option. Currently this option is for highlighting players with the same team cape as you, but since team capes aren't worn at tob I don't see an issue lumping these two options together into the same option. I'm open to changing this though - a lot of players may have realized that the "team members" option currently is about team capes (despite the description confusingly never mentioning team capes once), and may not assume that more functionality might be added to it.
   * So far I've only tested this in duo sizes but I see no reason it wouldn't work in larger team sizes.

* Clarifies what "Team members" are in the config description for the "Highlight team members" checkbox.
* Fixes a bug where having only "highlight team members" enabled with no other highlight options enabled results in no player names being drawn at all, even for team members.


geheur:
Would close #11969 and #13104

dominicfrost:
I was just thinking about implementing this extension myself, but was happy to find someone else had already gotten to it. Bummer this is so stale though, its a solid upgrade.

geheur:
This could be made as a hub plugin, you'd just have to make sure that player indicators isn't already drawing that person's name.

If you do, you might want to include toa as well.

dominicfrost:
Where you added feels like a good fit to me, since this is the plugin that controlls player highlights. Though imo it'd make more sense as a seperate toggle than team -- something like, raid-party. Are you thinking it'd be easier to contribute this as a plugin?

geheur:
It is almost always easier to do things as plugin hub plugins where possible (although in this case I see if the existing toa and tob all-in-one plugins will take it before releasing it as standalone). If you want to try getting it into core though, feel free. I agree that it would fit into player indicators well.

PR #13776, opened 2021-06-23
Widget based updates

Hydrox6:
All the previously movable vanilla widgets that became unmovable due to interface changes today are now movable again, along with the other Pest Control info widget since it was confusing me.

The combat level attack range indicator has also been fixed in the second commit. Turns out Jagex made it much nicer to edit, probably for their own Steam client feature, and now the widgets don't need to be moved around.


runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
```diff
@@ -167,6 +167,7 @@
 	public static final int TEMPOROSS_GROUP_ID = 437;
 	public static final int CLAN_GROUP_ID = 701;
 	public static final int CLAN_GUEST_GROUP_ID = 702;
+	public static final int TOB_HEALTH_BAR_GROUP_ID = 28;
```
Nightfirecat, line 170: There's already a `TOB_PARTY_GROUP_ID` defined as `28`. Ideally we shouldn't define the group ID twice, so maybe rename that `TOB_GROUP_ID`.

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -415,7 +422,9 @@
 
 	RAIDS_POINTS_INFOBOX(WidgetID.RAIDS_GROUP_ID, WidgetID.Raids.POINTS_INFOBOX),
 
+	TOB_PARTY_INTERFACE(WidgetID.TOB_PARTY_GROUP_ID, WidgetID.Tob.PARTY_INTERFACE),
 	TOB_PARTY_STATS(WidgetID.TOB_PARTY_GROUP_ID, WidgetID.Tob.PARTY_STATS),
+	TOB_HEALTH_BAR(WidgetID.TOB_HEALTH_BAR_GROUP_ID, 9),
```
Nightfirecat, line 427: A `Tob` child class already exists, we may as well define this child ID there.

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -74,6 +74,9 @@
 	DIARY_QUEST_WIDGET_TITLE(WidgetID.DIARY_QUEST_GROUP_ID, WidgetID.Diary.DIARY_TITLE),
 	DIARY_QUEST_WIDGET_TEXT(WidgetID.DIARY_QUEST_GROUP_ID, WidgetID.Diary.DIARY_TEXT),
 
+	PEST_CONTROL_BOAT_INFO(WidgetID.PEST_CONTROL_BOAT_GROUP_ID, 2),
```
Nightfirecat, line 77: A `PestControl` child class already exists, we may as well define this child ID there.

Nightfirecat (changes requested):

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
QuestingPet: Agreed. That widget actually has more than just party info. It has health info (as mentioned above in this addition), and also it's the same widget that displays the big red text on-screen that scrolls up (like the boss name, or "Seeking Party..." etc.)

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
Hydrox6: The only reason I didn't want to put it in there is that it is a separate widget group entirely, and it would collide with the `KNIGHT_INFO_CONTAINER` that already exists in it.

runelite-api/src/main/java/net/runelite/api/widgets/WidgetID.java
Hydrox6: Can tell how much my brain was frazzled yesterday :d

Nightfirecat (approved):

abextm (approved):

abextm (approved):

Issue #13347, opened 2021-03-18
Tile Marker with Fill Color


chngedd:
**Is your feature request related to a problem? Please describe.**

Some times the tile marks are hard to see on crowded backgrounds, The solution is to use a contrasting color but this can be hard when you are using certain colors to indicate specific things (dont click the red one, click the yellow one when x happens). 

**Describe the solution you'd like**

I think giving control over the fill color of marked tiles would help with this. The Agility extension has an overlay color that effects the solid border and half transparent fill color I think that would be the easiest solution.

**Describe alternatives you've considered**
Being able to make the border Thicker might also be a solution.

PR #13785, opened 2021-06-24
groundmarkers: Added option to fill tiles

WLoumakis:
Might want to have a setting to remember if a tile is filled or not, but I haven't thought of a good way to do that yet


WLoumakis:
Relates to #13347

Adam-:
this was added in 4e0211b4905c0327938e96f0ac78cab228aa8b70

PR #13786, opened 2021-06-25
timers: make thrall timer use boosted magic level

Hydrox6:
turns out that thrall spells are seemingly the only arceuus spell with a
magic-level based duration that uses the boosted level, at least out of
the ones we have timers for

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
@@ -503,7 +503,7 @@ public class TimersPluginTest
 	public void testThrall()
 	{
 		when(timersConfig.showArceuus()).thenReturn(true);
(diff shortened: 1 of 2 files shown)


Nightfirecat (approved):
Maybe the tests should verify the duration of the created timers? LGTM otherwise

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -503,7 +503,7 @@ public void testShadowVailCooldown()
 	public void testThrall()
 	{
 		when(timersConfig.showArceuus()).thenReturn(true);
-		when(client.getRealSkillLevel(Skill.MAGIC)).thenReturn(50);
+		when(client.getBoostedSkillLevel(Skill.MAGIC)).thenReturn(60);
```
Adam-, line 506: this test should probably also test the timer duration to be sure then.

PR #13794, opened 2021-06-27
zalcano: show total damage done and percentage that was yours

Hydrox6:
Before this, the overlay only showed the amount of damage you'd done.

Due to overlay sizes, this was a pain to fit; using the existing LineComponents caused the overlay to become 2 lines at default width (which doesn't look good when unintended), and be super wide if made to be on one line. As a result, I made it use TitleComponents instead, and split up the numbers from the text.

Here it is with the values reasonably maxed out, and it just barely fits into the default
image

and when empty, it still looks good because centre aligned
image

There's a little bit of jitter when values change, but that's unavoidable with more than one variable changing.


runelite-client/src/main/java/net/runelite/client/plugins/zalcano/ZalcanoPlugin.java
```diff
@@ -154,6 +163,28 @@ public void onVarbitChanged(VarbitChanged event)
 		{
 			resetDamageCounter();
 		}
+
```
Adam-, line 166: This method simplifies to:
```java
		boolean wasInCavern = inCavern;
		inCavern = isHealthbarActive();

		if (!inCavern)
		{
			if (wasInCavern)
			{
				resetDamageCounter();
			}
			return;
		}

		// Set up HP correctly for next shield phase
		int barNPCID = client.getVar(VarPlayer.HP_HUD_NPC_ID);
		if (barNPCID == ZALCANO_WEAKENED)
		{
			lastHP = SHIELD_HP;
		}
		else if (barNPCID == ZALCANO && client.getVar(Varbits.LARGE_HP_BAR_MAX_HP) == SHIELD_HP)
		{
			int currentHP = client.getVar(Varbits.LARGE_HP_BAR_CURRENT_HP);
			if (lastHP > currentHP)
			{
				totalShieldDamage += lastHP - currentHP;
			}
			lastHP = currentHP;
		}
```

Nightfirecat (commented):
LGTM, but needs rebase

PR #13800, opened 2021-06-28
xp tracker: show TTG instead of TTL when a goal has been set

Hydrox6:
A small change to show `TTG` for "Time to Goal" when a skill has a goal set.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanelLabel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanelLabel.java
@@ -33,7 +33,7 @@ import net.runelite.client.util.QuantityFormatter;
 @AllArgsConstructor
 public enum XpPanelLabel
 {
-	TIME_TO_LEVEL("TTL", XpSnapshotSingle::getTimeTillGoalShort),
+	TIME_TO_LEVEL("", XpSnapshotSingle::getTimeTillGoalShort),
 
 	XP_GAINED("XP Gained", snap -> format(snap.getXpGainedInSession())),
 	XP_HOUR("XP/hr", snap -> format(snap.getXpPerHour())),
@@ -55,6 +55,11 @@ public enum XpPanelLabel
 	 */
 	public String getActionKey(XpSnapshotSingle snapshot)
 	{
+		if (this == TIME_TO_LEVEL)
+		{
+			return snapshot.isHasGoal() ? "TTG" : "TTL";
+		}
+
 		String actionKey = key;
 		if (snapshot.getActionType() == XpActionType.ACTOR_HEALTH)
 		{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanelLabel.java
```diff
@@ -55,6 +55,11 @@
 	 */
 	public String getActionKey(XpSnapshotSingle snapshot)
 	{
+		if (this == TIME_TO_LEVEL)
```
Nightfirecat, line 58: I feel like changing `key`'s type to `Function<XpSnapshotSingle, String>` and creating an additional ctor for plain string arguments is a better approach to special-casing this. Then the change would be as simple as changing "TTL" to `(snap) -> snap.isHasGoal() ? "TTG" : "TTL"`.

Nightfirecat (approved):

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -61,6 +61,8 @@
 	private int startLevelExp = 0;
 	private int endLevelExp = 0;
 
+	private boolean hasGoal = false;
```
Nightfirecat, line 64: nit-pick: `false` is the default value, doesn't need to be explicitly specified.
```suggestion
	private boolean hasGoal;
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpStateSingle.java
```diff
@@ -294,6 +296,8 @@ boolean update(long currentXp, int goalStartXp, int goalEndXp)
 			{
 				endLevelExp = goalEndXp;
 			}
+
+			hasGoal = goalEndXp != -1;
```
Nightfirecat, line 300: When testing this on my maxed main, I found the value was `0` for skills without a goal set. I'm aware the call originating from `XpTrackerPlugin#onStatChanged()` can supposedly have a `-1` value, so this should be `goalEndXp <= 0` instead.
```suggestion
			hasGoal = goalEndXp <= 0;
```

runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpPanelLabel.java
```diff
@@ -45,8 +42,25 @@
 	;
 
 	private final String key;
+	private final Function<XpSnapshotSingle, String> keyFunc;
```
Nightfirecat, line 45: Added `@Nullable` for clarity
```suggestion
	@Nullable
	private final Function<XpSnapshotSingle, String> keyFunc;
```

Nightfirecat (changes requested):

PR #13801, opened 2021-06-28
Corporeal beast plugin syncs with discord party

igarshep:


Adam-:
I see multiple places here where you are keeping dangling references to players and the corp (holding references to them after they despawn) - this should not be done. The references are invalidated at the point when they are despawned.

PR #13807, opened 2021-06-29
widgets: add WidgetEventManager

LlemonDuck:
The WidgetEventManager enables multiple event listeners to be listening to the same event, without overwriting each other. The underlying widgets only allow a single listener to be registered, so this manager registers itself, then delegates the events to sub-listeners.

I did not update existing call-sites to use this manager; that can be done on a when-needed basis.

The reason for this addition is to resolve an issue I ran into with my hub plugin ToG Indicator, which conflicts with Skills Progress Bars when both are installed and enabled, by attempting to listen to the same widget events. If this PR is merged, I will update ToG Indicators to use this delegator, and will reach out to the maintainer of Skills Progress Bars to do the same.


LlemonDuck:
Closing: Stale, and not a good solution to a mostly non-existent problem

PR #13816, opened 2021-06-30
screenshots plugin: Add feature to automatically take screenshots for high value hitsplats and escapes at low hp.

wontonstrips:
Add feature to automatically take a screenshot (if enabled) when dealing a high "hitsplat" of damage on opponent and above threshold.  Ex. Take a screenshot if you hit a NPC with a 50 or higher damage.

Add feature to automatically take a screenshot (if enabled) when clicking a teleport tablet and under a specified HP.  Ex.  Take a screenshot if you click a teleport tablet and you have 5 or less HP left.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -275,4 +275,40 @@ public interface ScreenshotConfig extends Config
 	{
 		return Keybind.NOT_SET;
 	}
+
+	@ConfigItem(
+		keyName = "takeHitsplatShots",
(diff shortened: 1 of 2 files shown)


Nightfirecat:
In my opinion this should be a hub plugin rather than an addition to the screenshot plugin. Currently though, there's not a straightforward way to capture screenshots in a hub plugin besides duplicating most of the drawing/overlay/capture code from this plugin. (a gap I hope to fix in #11022)

wontonstrips:
Cool yeah, I was kind of wondering that.  Was thinking like "I wonder if i could just imported the screenshot plugin into a separate plugin if I would be able to use it to take screenshots".  Anyway had this idea and figured I would just make it and put it out there.  Perhaps in the future if your Image Capture API goes through I will revisit this.  Thanks for reply :)

PR #13834, opened 2021-07-04
loottracker: Added the ability to split the group loot tracker by current session and aggregate data

Mitchell-Kovacs:
This PR adds the config option "Split group loot section" in the loot tracker plugin (default false).
When enabled, it allows users to quickly see the total amount of loot obtained vs the previous aggregate data. Example: 

image

New menu option:
image

While the "Show each kill separately view" does allow a user see what has been obtained this session it's hard to get the total value gained as the number of kills go up.

All comments/feedback/suggestion welcome.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerBox.java
@@ -411,7 +411,7 @@ class LootTrackerBox extends JPanel
 	{
 		title.setLayout(new BoxLayout(title, BoxLayout.X_AXIS));
 		title.setBorder(new EmptyBorder(5, 7, 5, 7));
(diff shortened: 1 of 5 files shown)


acdvs:
I think having group names, aggregates, and sessions all combined in one single-tier list is good UX. Ideally the aggregate and session tabs should be very clearly displayed as children of their respective group (i.e., outline format).

Mitchell-Kovacs:
> I think having group names, aggregates, and sessions all combined in one single-tier list is bad UX. Ideally the aggregate and session tabs should be very clearly displayed as children of their respective group (i.e., outline format).

I agree. I'm not the happiest overall with my implementation but figured I'd put it out to see if it would have any interest.
Changing the background of the subsection title does help to solve this a bit but could definitely be improved upon.


image

Mitchell-Kovacs:
Closing in favour of #13854 
Also, if someone comes across this the Loot Logger plugin has the option for showing current session data.

PR #3574, opened 2018-06-02
sidebar: fix minimum size when restoring window

jplsek:
Fixes #2612

This fix seems like a hack, but I'm not sure of a better way around it since it seems like a Swing problem. 

While the client is as small as possible and the sidebar gets contracted, the minimum size gets updated. Now maximize the client, then expand the sidebar. Based on the existing `expandby` code, the minimum size does not get updated. However, adding a `revalidateminimumsize` while fullscreen "does" update the minimum size, restoring the window does not actually resize the frame. It will only resize the frame when the window is moved or manually resized. This is why I think it's a Swing problem. While technically correct to call `revalidateminimumsize` in the `expandby` code, I decided to only call it during the restoring window event, instead of having both calls, where it wouldn't make much of a difference.

PR #13835, opened 2021-07-04
Fix restored window size when client is fullscreen

### LINKED ISSUE #2612: Sidebar is hidden in certain condition.
Reproduce :
Hide sidebar in restored mode > maximize > show side bar > restore

ezgif-1-73afb0325e

On both versions 1.3.8 and 1.3.9

MarkKoester:
Fixes #2612 and related to #3574

Currently, when the plugin panel or sidebar is shown while the client is fullscreen, and the client is restored, the updated size of the client is not respected.

This pull request makes two changes to fix this issue. First, `revalidateMinimumSize()` is moved in the `expandBy()` and `contractBy()` functions to be called before the return path when the client is maximized. This should guarantee that the minimum size is still captured in all situations.

Second, the client size is manually set when the window is restored. When the window is restored, it's set back to it's size before it was maximized regardless of the minimum size, so the size has to be manually set. Since the `setSize()` function respects the frames minimum size, I just call `setSize()` with the frames current size and it works as desired.

Alternatively, the onResize listener could check for whether the frame state was previously maximized and is now not maximized, but the current approach seems a bit simpler. I'm not sure which one would necessarily be better.

Also, I'm not exactly sure if adding the component listener should be done in ClientUI or the ContainableFrame class, so I'd be interested in getting feedback there.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -41,6 +41,8 @@ import java.awt.LayoutManager;
 import java.awt.Rectangle;
 import java.awt.Toolkit;
 import java.awt.TrayIcon;
+import java.awt.event.ComponentAdapter;
+import java.awt.event.ComponentEvent;
 import java.awt.event.MouseEvent;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
@@ -357,6 +359,16 @@ public class ClientUI
 					{
 						shutdownClient();
 					}
+				}
+			});
+
+			frame.addComponentListener(new ComponentAdapter()
+			{
+				@Override
+				public void componentResized(ComponentEvent e)
+				{
+					// Manually call setSize to force the frame to respect it's minimum size after restore
+					frame.setSize(frame.getSize());
 				}
 			});
 
--- a/runelite-client/src/main/java/net/runelite/client/ui/ContainableFrame.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ContainableFrame.java
@@ -157,6 +157,7 @@ public class ContainableFrame extends JFrame
 	{
 		if (isFullScreen())
 		{
+			revalidateMinimumSize();
 			return;
 		}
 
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -360,6 +362,16 @@ public void windowClosing(WindowEvent event)
 				}
 			});
 
+			frame.addComponentListener(new ComponentAdapter()
+			{
+				@Override
+				public void componentResized(ComponentEvent e)
+				{
+					super.componentResized(e);
```
abextm, line 370: you don't need a super call here

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -360,6 +362,16 @@ public void windowClosing(WindowEvent event)
...
+					frame.setSize(frame.getSize());
```
abextm, line 371: this should have a comment

abextm (changes requested):

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
MarkKoester: Done

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
MarkKoester: Done

abextm (approved):

WooxSolo:
The minimum size is 8 pixels off in both width and height when unmaximizing the window after expanding or contracting the sidebar in maximized mode with custom chrome enabled from my testing on Windows 10. Would guess this is due to custom chrome adding 4 pixels to the border in unmaximized mode but not in maximized mode, so revalidating the minimum size in maximized mode gives a slightly wrong minimum size.

Adam-:
1) I've looked at these `revalidateMinimumSize` calls but I do not see any behavior change by changing them alone. When the client is maximized everything still behaves correctly since you cannot make the window smaller without unmaximizing it. So I don't think it is even necessary to change this to fix this problem?

At best maybe it ensures the minimum size is always valid even when maximized but I'm not sure if this is useful? Do you know why you are trying to change this?

2) I think that this is just needing to recompute the minimum size whenever the frame state is changed, similar to what #3574 is doing. 

3) I too see the minimum size being slightly too small with this like Woox said - even also with #3574. I figured out just putting the resize call in an invokeLater() fixes it. Not sure.

So I think that this commit fixes these issues instead: https://github.com/Adam-/runelite/commit/ab2658b6a6fa0a644ac754423cbf34d1188b1b02. Can you check and see if it works for you?

cc @jplsek

Adam-:
I tested this some more and think it's probably fine so I've pushed it. Reopen or comment on #2612 if its still a problem.

PR #13849, opened 2021-07-08
loot tracker: Identify corrupted and failed gauntlet loot

### LINKED ISSUE #11279: Seperate loot tracker entries for fail/completed corrupted gauntlet
I think it would be better if loot tracker entered gauntlet completions and fails into seperate entries, it would be nice to be able to get a better idea of average loot from a completion rather than it being like this https://i.imgur.com/M8437qg.png , which is probably like 15 deaths 10 completions

Nightfirecat:
Not my favorite code I've ever written, but I think this should be fine? I have tested both failed and corrupted gauntlet completions and this works fine for them; I have not tested the regular gauntlet case, if someone wouldn't mind testing that still works fine, it would be appreciated.

Fixes #11279

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -77,6 +77,7 @@ import net.runelite.api.SpriteID;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.GameTick;
 import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.MenuOptionClicked;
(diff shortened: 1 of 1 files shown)


Hydrox6:
Just went and tested this, looks like it detects it all fine 
image
(There's no distinction on the failure chests loot wise, I just wanted to show that I tested all the cases and it didn't screw up somehow)

Nightfirecat:
afaik failed gauntlet loot is the same regardless of which type was attempted

makingcomment:
> afaik failed gauntlet loot is the same regardless of which type was attempted

Even if the loot comes from the same table, it still might be useful to have them distinct in the loot tracker.

JuanAntelo:
Hello, looks like this code is ready to be merged. I could really use this, would be great to have it merged

Nightfirecat:
Rebased with recent plugin changes, but dropped the failed gauntlet loot differentiation as it requires some reworking, and I haven't had a chance to test this ingame at all yet.

PR #13854, opened 2021-07-08
loot tracker: add option to only show current session's loot

### LINKED ISSUE #10346: Loot tracker - Filter by session
Would be nice to have an option to filter the loot tracker to show only the current session, or perhaps even options for previous sessions as well as showing everything together.

Sorry if something like this was already suggested, tried searching but didn't see it.

Hydrox6:
Closes #10346
Closes #13834

Adds a nice simple button to toggle between showing all loot and the current session's loot.
I've been testing this for.... quite a while. Was never satisfied with the icons, but at this point they've grown on me. If anyone has some better icons to represent the concepts of "all loot" and "session loot" that are free for this project to use, do point me to them.

<table>
<tr>
	<td>
With all loot showing
	<td>
With only session loot
<tr>
	<td>

	<td>

</table>


QuestingPet:
Just curious, is it not considered out of place that these are inside of the "Overall" info panel, as opposed to near the top with the other "action" buttons?

Hydrox6:
It was partly because the top section is getting *very* full, especially when it's showing the name of the monster for the detailed view. I also just think it looks nicer there tbh. Up for changing it, though I might also need to clean up that top bar if so.

TranceHD:
I’d love this feature so much.

stantheblan:
Can we revisit this PR? Would love to have it implemented after 4 years :)

PR #13856, opened 2021-07-09
timers: add cannon parts

Loze-Put:
Display the cannon timer upon setting up the first cannon part. The decay timer is reset after adding an additional part to the cannon. This change is useful when only setting up the cannon base, instead of the entire cannon. Example use case [here](https://youtu.be/Ce5qU85QPUc?t=14).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -55,7 +55,6 @@ import net.runelite.api.Varbits;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ActorDeath;
 import net.runelite.api.events.AnimationChanged;
-import net.runelite.api.events.StatChanged;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -559,7 +562,11 @@ public void onChatMessage(ChatMessage event)
 
 		}
 
-		if (config.showCannon() && (message.equals(CANNON_FURNACE_MESSAGE) || message.contains(CANNON_REPAIR_MESSAGE)))
+		if (config.showCannon() && (message.equals(CANNON_BASE_MESSAGE) ||
+			message.equals(CANNON_STAND_MESSAGE) ||
+			message.equals(CANNON_BARRELS_MESSAGE) ||
+			message.equals(CANNON_FURNACE_MESSAGE) ||
+			message.contains(CANNON_REPAIR_MESSAGE)))
```
Nightfirecat, line 569: Our usual code style for continued lines is to push the operator to the next line instead of having it at the end of the line.
```suggestion
		if (config.showCannon() && (message.equals(CANNON_BASE_MESSAGE)
			|| message.equals(CANNON_STAND_MESSAGE)
			|| message.equals(CANNON_BARRELS_MESSAGE)
			|| message.equals(CANNON_FURNACE_MESSAGE)
			|| message.contains(CANNON_REPAIR_MESSAGE)))
```

PR #13858, opened 2021-07-11
ground items: Highlight stackable ground items already in inventory

Krysaczek:
### Highlight stackable ground items already in inventory
Less time is spent looking throught the loot and all these items are free to pick up while plenty of space is still kept for big ticket items or alchables. Goes well with afk combat sessions.

### Notes:

- I'm assuming that majority of ground item users don't have min-maxed hidden/highligted item lists.
- Color picker config could be renamed to be more neutral and used as a second tier of highlighted items. I'm already thinking about implementing "Highlight storable ground items that can be stored in inventory storage containers" which could use this Color picker in the future.


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -403,4 +403,27 @@ default boolean textOutline()
...
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "highlightStackable",
+			name = "Highlight stackable in inventory",
+			description = "Highlight stackable ground items already in inventory",
+			position = 30
+	)
+	default boolean highlightStackable()
```
Adam-, line 413: Typically for things like this we'd just make the color default to null, eg:

```java

	@Alpha
	@ConfigItem(
		keyName = "highlightStackableColor",
		name = "Highlight stackable",
		description = "Configures the color for highlighted stackable items",
		position = 31
	)
	Color highlightStackableColor();
```

and then null-test it to see if it is enabled or not, instead of having a separate config option.

Adam- (commented):
Overall this seems fine. Needs rebase.

PR #13862, opened 2021-07-11
plugins: add missing roofs of the lighthouse and falador castle to roofremoval

Bram91:
The lighthouse and falador castle were showing parts of the roof. This commit should add the missing tiles.


runelite-client/src/main/resources/net/runelite/client/plugins/roofremoval/overrides.json
```diff
@@ -1878,6 +1886,14 @@
...
       "z1": 1,
       "z2": 2
+    },
+    { // Lighthouse clockwork
+      "rx1": 12,
+      "ry1": 51,
+      "rx2": 13,
+      "ry2": 51,
+      "z1": 0,
+      "z2": 2
```
Hydrox6, line 1896: This override isn't connected to the other override, which causes some very weird stuff to happen when on plane 1. It also doesn't need to plane 2 since the clockwork itself is on that plane.
```suggestion
      "ry2": 52,
      "z1": 0,
      "z2": 1
```

Hydrox6 (changes requested):

PR #13883, opened 2021-07-15
Slayer highlight min max levels

nateprat:
simple level filter for slayer highlighter before slayer highlighter is moved to NpcIndicators


nateprat:
found bug, will updat

PR #13887, opened 2021-07-16
menu entry swapper: Add ash scatter swap

Nightfirecat:
Similar to the bones entry, this swap will prevent accidental scattering
of demonic ashes for players who wish not to gain prayer experience, or
wish to save them to sell or offer at a later time.


Nightfirecat:

PR #13889, opened 2021-07-16
menu entry swapper: Add Drakan's Medallion swaps

Cyborger1:
Adds the 3 teleport options as entry swaps (Ver Sinhaza/TOB, Darkmeyer and Slepe).


Nightfirecat:

PR #13890, opened 2021-07-16
skillcalculator: Add New Amethyst Items

Vetricci:
Add amethyst darts & dart tips to fletching and crafting calculators respectively.


runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_crafting.json
```diff
@@ -757,6 +757,12 @@
       "name": "Light Orb",
       "xp": 70
     },
+    {
+      "level": 89,
+      "icon": 25853,
+      "name": "Amethyst Dart Tips",
```
Nightfirecat, line 763:
```suggestion
      "name": "Amethyst Dart Tip",
```

runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_crafting.json
```diff
@@ -757,6 +757,12 @@
...
+      "xp": 60
```
Nightfirecat, line 764: I'm not sure that this is the right value to use. Presently, we use the xp given from a single of the item created, even if it's not possible to only create one of the item--case in point, arrow shafts. That would make this value 7.5, not 60.

Nightfirecat (changes requested):

runelite-client/src/main/resources/net/runelite/client/plugins/skillcalculator/skill_crafting.json
Vetricci: The value is consistent with the existing amethyst items (arrowtips, javelin heads, bolt tips). For crafting you're looking for the number of amethyst to process, each providing 60 xp.

Nightfirecat: Each of the items you mentioned can be processed individually, unlike the arrow shaft case I mentioned. You cannot craft fewer than 8 amethyst dart tips.

PR #13893, opened 2021-07-17
Change Grotesque Guardians instance closing time to 5 minutes

emerald000:
The instance closes 5 minutes after killing Dusk if you don't ring the bell again.


Nightfirecat:
I've gotten confirmation from another trusted source on this, it is indeed 5 minutes.

PR #13905, opened 2021-07-19
config panel: show icon on installed hub plugins

Hydrox6:
This should make it excessively clear to users that hub plugin's aren't the same as core plugins.
It could also potentially hint to developers that maybe they should use a shorter name for their hub plugin.

image


Nightfirecat (approved):
LGTM

Wynadorn:
I like the change though maybe the icon could be placed somewhere else as it makes the list's oder look a bit off.
Maybe the list could be sorted afterwards with the plugin-hub one's at the bottom of the list, alternatively the icon could be placed after the title.

Hydrox6:
It's intended to be before to make it more obvious that hub plugins are different. The other positions all look worse

abextm (changes requested):
This should not shift the label, it should either be on the right of the label, near the label, or on the left of the button group. Also I'm not sure if the plug icon is particularly clear since everything in the list is a plugin.

Hydrox6:
As discussed on Discord, feel free to submit an alternative icon, or an icon for all core plugins to have as well. The icon used is a scaled down version of the default hub plugin icon, hence it is already associated with the hub as far as the user is concerned.

Putting the icon to the right of the text does not stand out. Putting it alongside the right-side buttons makes it look as if it should be a button itself, which it isn't, and giving it some token link to the hub just to make it a button isn't a good solution.

Putting it on the left makes it *very* clear to the user that the plugin is different, which is the entire problem that this is trying to solve.

PR #13907, opened 2021-07-20
clues: remove non-existent clues

Hydrox6:
The wiki recently started crowdsourcing clue data with their hub plugin, and came upon some very interesting findings: a handful of clues weren't real! A lot of them can be traced to a post-OSRS clue update (April 2008), and those clues exist to this day on RS3.

The pre-SotE clues definitely seem gone; they were seemingly replaced with newer clues, and the wiki has noted no clue dialogue from the relevant clue givers.


Nightfirecat:
Needs rebase. Also, is there a source for the clues the wiki says no longer exist? eg. some wiki revision or other data source

Nightfirecat (approved):
Cook from the wiki team has been tracking these down via his crowdsourcing, and I've confirmed these changes from his wiki edits. :+1:

PR #13960, opened 2021-07-28
world map: correct locations of quest markers

Hydrox6:
With the update today (28/07/21), Jagex moved many quest markers, split up all previously grouped ones, and added their own tooltips. I've checked and fixed every single quest start location, added in the new ones that were split up, and removed code that is now unnecessary 

Closes #10170


makingcomment:
Now that Jagex added their own tooltips, the "Quest names and status" world map config may be worth renaming/redescribing too.

Hydrox6:
oh I completely forgot about the config. Good shout

makingcomment:
Also I think this (or the game update itself?) closes #10170

Issue #8507, opened 2019-04-13
Automatically Track Slayer Target Aggression Timer


Jademalo:
**Is your feature request related to a problem? Please describe.**
Generally, I find that I primarily want to know aggression timers when I'm on a slayer task safespotting. I have to manually add whichever NPC is my target, and then manually remove it when I;m done.

**Describe the solution you'd like**
Have an option in the NPC Aggression Timer to automatically enable the timer for your current slayer target.

**Describe alternatives you've considered**
Manually adding works, but it's fairly tedious.

**Additional context**
I feel like this would be a fairly simple and extremely useful change. Not having to manually add NPCs every task would be super useful, and make the aggression timer plugin better.
I would also argue adding similar functionality to the NPC indicator plugin would also be extremely useful, but that's one for another ticket.

LlemonDuck:
Added in c669f6462bf9a9bca82befdeb6bc58915b9ed6c4

PR #13963, opened 2021-07-30
Shows NPC Aggression timer for slayer tasks #8507

fatfingers23:
This adds a new config option inside of the NPC Aggression Timer named "Show timer for current slayer task" when selected it will display the timer for the current slayer task NPCs. The aggression timer works finding the target using the same logic as the Slayer Plugin. I move the logic to a shared util. I hope this was an okay location to place it. If not please direct me to the best location for the shared logic. This does not add the slayer NPC to the config NPC list but holds its own check inside of the NPC Aggression Timer plugin. 

This also solves #8507

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
@@ -117,4 +117,15 @@ public interface NpcAggroAreaConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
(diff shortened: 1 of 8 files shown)


fatfingers23:
Added slayer events as suggested in discord. The plugin no longer uses same logic as slayer plugin to decide if the NPC needs to show the agro timer. It now formats the slayer task NPC names inside of the Agro Timer plugin to set the timer properly for slayer task npc. I also added unit tests.

Please let me know if I need to change anything. Thank you!

fatfingers23:
* Fixed merge conflicts 
* Fixed a bug with timer not turning on at task during certain circumstances 
* Fixed bug where turning on and off npc agro plugin would not show timer for slayer task
* Tried to clean up code a litte

PR #13966, opened 2021-08-01
Enable antidrag in pvp

jsuarez5341:
This feature has been added to the official Steam client -- no reason to omit it from RuneLite

First PR to this project -- If I've missed any protocols, please let me know

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
@@ -29,9 +29,7 @@ import java.awt.event.KeyEvent;
 import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
-import net.runelite.api.Varbits;
 import net.runelite.api.events.FocusChanged;
-import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.events.WidgetLoaded;
 import net.runelite.api.widgets.Widget;
 import net.runelite.api.widgets.WidgetID;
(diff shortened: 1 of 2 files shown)


Adam-:
@Hydrox6 weren't you doing something wrt this?

runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
```diff
@@ -113,12 +111,12 @@ public void keyTyped(KeyEvent e)
 	@Override
 	public void keyPressed(KeyEvent e)
 	{
-		if (e.getKeyCode() == KeyEvent.VK_CONTROL && config.disableOnCtrl() && !(inPvp || config.onShiftOnly()))
+		if (e.getKeyCode() == KeyEvent.VK_CONTROL && config.disableOnCtrl())
```
Adam-, line 114: I think this is wrong and should be:
```suggestion
		if (e.getKeyCode() == KeyEvent.VK_CONTROL && config.disableOnCtrl() && !config.onShiftOnly())
```
Since ctrl to disable only makes sense when shift-only is off, otherwise you would have to hit both shift + ctrl at once to enable and then disable which doesn't make sense.

Adam- (changes requested):

jsuarez5341:
Thank you!

runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
jsuarez5341: Now it only works in pvp when you enable onShiftOnly

PR #13968, opened 2021-08-02
Highlight entity when hovering menu

superiorser9:
Highlights the NPC/Object when hovering its menu options.

ezgif com-gif-maker


Adam-:
added in f283871f507d8bd40793294f9d248aba7c258290

PR #13986, opened 2021-08-05
Interact highlight: Add config for click flash

LeikvollE:
Added config to change click flash color and option to disable the click flash.
Additionally added click flash for when show on interact is disabled.


runelite-client/src/main/java/net/runelite/client/plugins/interacthighlight/InteractHighlightOverlay.java
```diff
@@ -97,7 +95,8 @@ private void renderMouseover()
 				TileObject tileObject = plugin.findTileObject(x, y, id);
 				if (tileObject != null && config.objectShowHover() && (tileObject != plugin.getInteractedObject() || !config.objectShowInteract()))
 				{
-					modelOutlineRenderer.drawOutline(tileObject, config.borderWidth(), config.objectHoverHighlightColor(), config.outlineFeather());
+					Color hoverColor = getClickColor(config.objectHoverHighlightColor(), config.objectHoverHighlightColor(), client.getGameCycle() - plugin.getGameCycle());
```
Adam-, line 98: this is very confusing since you are using the same start/end color in order to abuse the clickFlash check below. The clickFlash check should instead be moved either here or to another method.

PR #13988, opened 2021-08-05
NPC Indicator Fill Opacity

Fayyne:
With the new highlights for interactions added, I started looking into configuring my NPC indicators tile thickness and etc, and realized I wanted a way to make the fill have less opacity. Small config and appropriate access added.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
@@ -31,6 +31,7 @@ import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 import net.runelite.client.config.ConfigSection;
 import net.runelite.client.config.Range;
+import net.runelite.client.util.ColorUtil;
 
 @ConfigGroup("npcindicators")
 public interface NpcIndicatorsConfig extends Config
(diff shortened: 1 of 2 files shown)


Fayyne:
A Discussion point, is before all Opacity was defaulted to 20, so this still links all those together (sw tile, hull, outline). In theory, if someone did a 255 settings for a solid color block, it would also completly fill the hull, which may not be exactly the user wants.

Should we (or I in this case) possibly break the render out into more specified  renders to avoid this? In that case I would end up changing this to "Tile Fill Opacity" and only apply it to SW and Tile renders and not hull and outline, which would stay at the default 20 (or I can make that a setting as well).

Or just ignore this "feature" and let users live their life with big blocks of color in their life. (Nothing this in the wiki docs)
image

Adam-:
The color picker has opacity support and we typically use that for this type of thing. Any reason you are not doing that?

Fayyne:
If you are referring to the opacity for the lines in the current color option, that didn't seem to apply to the tile fills. Just the outline itself.

And making the option a color picker over just a numerical value seemed overkill for just an alpha change. 

Default Fill
image

0 Fill 
image

255 Fill

image

Fayyne:
If you think there is value in making the fill color different from the Outline color, I can look into doing that I don't think it would be a large change from what I already have, as you said would be more consistent with other feature sets, just didn't feel right to me in this particular implementation, but I was also blinded by my specific goal during implementation and testing.

Adam-:
I think that we can change this back to `alpha/12` like how it was done in ab4cf1afe913879f9d0c1b031d77798b0b6c865e - this was reverted in fa43ca94dbe97235feb54d4fdce499a6be3f555e due to it being no longer impossible to set the color fully transparent in order to get a border less highlight, which some people want, however now the border width is configurable so that should no longer be an issue.

Fayyne:
Hmm, my only problem with with the `alpha/12` approach is that what if one wants 0 fill, but a bright tile highlight, in which case would not be achievable. I think the best solution would just be a color picker for fill, with a default to the current  Cyan at 20 alpha.

Adam-:
A 2nd color picker for the fill is fine too I guess.

Fayyne:
Cool, I'll get that patched in and tested. Thanks for the feedback and discussion :D

Adam-:
added in 33ff3a2749372a1d98329d2c4aeddbeadf19434e

PR #13989, opened 2021-08-05
menu entry swapper: allow teleport item excludes

molo-pl:
Allows some items to be excluded from the "Teleport item" feature of Menu Entry Swapper.

teleport-item-excludes

This swap covers a wide range of equipment, and for some the "teleport" action might be an undesired default, often leading to unintentional teleports.


Nightfirecat:
8944eb9be6d62ec8384e29484dab14c4c2858d9d was added to master, which sufficiently covers this use case. (the left-click configuration takes priority over other swaps)

PR #13995, opened 2021-08-08
Update to skill calculator "action" plurals

neatclient12345:
Update skill calculators to correctly show plurals of action as "0 actions", "1 action", "2 actions", "3 actions", etc.
Old behavior showed "0 action", "1 action", "2 actions", etc


neatclient12345:
Hi, just wanted to add a clarification to my commit message - the old behavior is "0 action", "1 action", "2 actions" - it gets it correct unless actionCount is 0 (which happens frequently with calculators that aren't setup yet). This PR fixes the pluralization for the actionCount==0 case

PR #13996, opened 2021-08-08
Change slayer plugin to use NPC Indicators plugin for hightlighting.

fatfingers23:
Created new events `onHighlightNpcEvent` and `onUnHighlightNpcEvent` for other plugins to use the Npc Highlight plugin for highlighting. The event takes the NPC and the color it wishes to highlight the NPC. I left the target color option in the slayer plugin and the rest of the overlay options follow what is set in the  NPC Indicators plugin.


Adam-:
added in c7b444d8926e994a9f2c32b80e69a0d3b227136e

PR #13997, opened 2021-08-08
Add the Agility Pyramid to the Agility skill calculator.

neatclient12345:
The Agility Pyramid rewards bonus XP depending on player level. This PR calculates the bonus XP based on the starting level entered into the calculator.


Nightfirecat:
Rather than having code specific to the agility pyramid in the skill calculator, I'd rather see support for level-based variable-experience actions added generally, eg. #10076. If done correctly, this should mean that all of the implementation for such skills could be done in `SkillAction` or some child interface of that one so no one-off code needs to exist in the skill calculator class itself.

You can also refer to some of the discussion on #9940 to understand some of the work that would be involved. Mainly, such a solution should consider that intermediate levels have different XP requirements to level up, and also grant different XP for each actions completed.

PR #14030, opened 2021-08-16
Chat commands: Improve findPet performance and fix error message formatting

IllyaMyshakov:
Improve `findPet` performance in the Pet class
Fix error message formatting in compliance with #13917

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/Pet.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/Pet.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.chatcommands;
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/Pet.java
```diff
@@ -84,15 +85,18 @@
 	private final String name;
 	private final Integer iconID;
 
-	static Pet findPet(String petName)
+	private static final HashMap<String, Pet> PET_HASH_MAP = new HashMap<>();
```
Adam-, line 88: There's only ~40 something pets, this really isn't necessary. And also we typically use `ImmutableMap` for stuff like this.

PR #14035, opened 2021-08-18
chat commands: Add duplicate pets support

### LINKED ISSUE #13923: !pets  chat command to show dupe pets such as <rocky> x 3
Pretty much just as the title states. it would be cool if the pets command listed off hm dupe pets you have

IllyaMyshakov:
Closes #13923

Add duplicate pets support to the !pets chat command

Change Redis storage type from SET to HSET

**Images**
Normal:

image

Double all pets:

image


http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
```diff
@@ -30,8 +30,7 @@
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
-import java.util.Set;
+import java.util.HashMap;
```
Nightfirecat, line 33: Why is this used over `java.util.Map`? I don't see anything which specifies a HashMap specifically is needed.

http-service/src/main/java/net/runelite/http/service/chat/ChatController.java
```diff
@@ -250,9 +252,12 @@ public void submitLayout(@RequestParam String name, @RequestBody LayoutRoom[] ro
 	}
 
 	@PostMapping("/pets")
-	public void submitPetList(@RequestParam String name, @RequestBody int[] petList)
+	public void submitPetList(@RequestParam String name, @RequestBody HashMap<Integer, Integer> petList)
 	{
-		if (petList.length == 0 || petList.length > MAX_PETS)
+		if (petList.size() == 0
+			|| petList.size() > MAX_PETS
+			|| Collections.max(petList.values()) > MAX_PETS
```
Nightfirecat, line 259: Huh? Why would it matter if a user has received a duplicate pet more than 256 times?

http-service/src/main/java/net/runelite/http/service/chat/ChatController.java
```diff
@@ -250,9 +252,12 @@ public void submitLayout(@RequestParam String name, @RequestBody LayoutRoom[] ro
...
+			|| petList.values().stream().reduce(0, Integer::sum) > MAX_PETS)
```
Nightfirecat, line 260: Here also, why would it matter if a user has received enough duplicate pets that across all pets they had gotten more than 256 pet rolls?

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -1208,9 +1207,13 @@ private void petListLookup(ChatMessage chatMessage, String message)
...
 		{
-			if (playerPetList.contains(pet.getIconID()))
+			if (playerPetList.containsKey(pet.getIconID()))
 			{
-				responseBuilder.append(" ").img(modIconIdx + pet.ordinal());
+				Integer petQuantity = playerPetList.get(pet.getIconID());
+				for (int i = 0; i < petQuantity; i++)
+				{
+					responseBuilder.append(" ").img(modIconIdx + pet.ordinal());
+				}
```
Nightfirecat, line 1216: Duplicating the icon seems very weird to me. I think it would be much better implemented like so:
```java
responseBuilder.append(" ").img(modIconIdx + pet.ordinal());
int petQuantity = playerPetList.get(pet.getIconID());
if (petQuantity > 1)
{
	responseBuilder.append("x" + petQuantity);
}
```

Nightfirecat (commented):
This PR generally seems to overqualify all map uses as `HashMap`; `Map` looks fine for most of these as the actual map implementation is not relevant in most cases.

http-api/src/main/java/net/runelite/http/api/chat/ChatClient.java
IllyaMyshakov: This was only for consistency. Can change it

http-service/src/main/java/net/runelite/http/service/chat/ChatController.java
IllyaMyshakov: For data validation. The check may be redundant as filling the chatbox with 250-ish icons might be stretching it, so I can remove this

http-service/src/main/java/net/runelite/http/service/chat/ChatController.java
IllyaMyshakov: This was for data validation. As the Collection Log only counts a given entry up to 250 times, any amount above 250 would definitely be an error state. Can change `MAX_PETS` to 250

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
IllyaMyshakov: Originally I thought duplicating the pet icon would be a more pleasing user aesthetic. However, if this is not recommended, then I can change it

http-service/src/main/java/net/runelite/http/service/chat/ChatController.java
Nightfirecat: Is this confirmed somewhere? There are numerous log entries which definitely count beyond 256 of a drop, though I believe each entry can have a different maximum counted value.

PR #14062, opened 2021-08-25
ClientUI: Run PluginPanel#onDeactivate when switching panels

Cyborger1:
`onDeactivate()` currently only runs when the plugin panel is explicitly closed, but not when it's implicitly closed from switching to a different panel. This fixes that.

There currently doesn't seem to be any core plugin that actively uses `onDeactivate()`, but there are some plugin hub plugins that are affected by this bug.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -959,6 +959,11 @@ public class ClientUI
 
 		int width = panel.getWrappedPanel().getPreferredSize().width;
 		int expandBy = pluginPanel != null ? pluginPanel.getWrappedPanel().getPreferredSize().width - width : width;
+
+		if (pluginPanel != null)
+		{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -959,6 +959,11 @@ private void expand(@Nullable NavigationButton button)
 
 		int width = panel.getWrappedPanel().getPreferredSize().width;
 		int expandBy = pluginPanel != null ? pluginPanel.getWrappedPanel().getPreferredSize().width - width : width;
+
+		if (pluginPanel != null)
+		{
+			pluginPanel.onDeactivate();
+		}
```
Nightfirecat, line 966: A comment helps clarify this bit of code.
```suggestion
		// Deactivate previously active panel
		if (pluginPanel != null)
		{
			pluginPanel.onDeactivate();
		}

```

Nightfirecat (approved):

PR #14073, opened 2021-08-27
xp globes: account for progress arc width when deciding overlay bounds

Hydrox6:
Currently, the xp's bounds do not account for the progress arc, as it is drawn either side of the perimeter of the background circle. This PR addresses that.

On `master`:
record_2021-08-27__16-10-01

With this PR:
record_2021-08-27__16-09-48

This also fixes the progress arc being drawn off-screen or interfering with other overlays when the progress arc size is excessively high (shown here with 20px):
image
record_2021-08-27__16-11-34

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
@@ -63,6 +63,7 @@ public class XpGlobesOverlay extends Overlay
 	private static final int MINIMUM_STEP = 10;
 	private static final int PROGRESS_RADIUS_START = 90;
 	private static final int PROGRESS_RADIUS_REMAINDER = 0;
+	private static final int PROGRESS_BACKGROUND_SIZE = 5;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/xpglobes/XpGlobesOverlay.java
```diff
@@ -108,31 +109,34 @@ public Dimension render(Graphics2D graphics)
 			return null;
 		}
 
-		int curDrawPosition = 0;
+		// The progress arc is drawn either side of the perimeter of the background. This value accounts for that
+		// when calculating draw positions and overall size of the overlay
+		int progressArcOffset = (int) Math.ceil(Math.max(PROGRESS_BACKGROUND_SIZE, config.progressArcStrokeWidth()) / 2.0);
```
Nightfirecat, line 114: `final` for clarity would be nice here.

Nightfirecat (commented):
LGTM. I don't especially like this approach since it works around the fact that `renderProgressCircle()` can draw past the starting point rather than changing it to avoid that overflow drawing, but this approach does have the least user-facing impact while still being correct.

PR #14092, opened 2021-09-02
Chat Channels: Add option for displaying online clan member count

d-as:
Since there is no currently an easy way to see how many clan members are online, this patch allows displaying the number of online clan members at the end of the clan's name.

Demo


runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java
```diff
@@ -341,6 +346,11 @@ public void onGameTick(GameTick gameTick)
 			}
 		}
 
+		if (config.clanChatShowOnlineMemberCount())
```
Adam-, line 349: Typically we prefer to put these into script events if possible, like how the friends chat channel counter works. Have you tried that?

d-as: I'll try to do that, thanks for the tip!

runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java
```diff
@@ -845,4 +855,17 @@ private void colorIgnoredPlayers(Color ignoreColor)
 			listWidget.setTextColor(ignoreColor.getRGB());
 		}
 	}
+
+	private void showClanOnlineMemberCount (boolean show)
+	{
+		Widget clanTitle = client.getWidget(WidgetInfo.CLAN_TITLE);
+		ClanChannel clanChannel = client.getClanChannel();
```
Adam-, line 862: It would be good to show the guest channel count too probably.

d-as: Added a separate option for guest clan chat.

runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java
```diff
@@ -740,4 +755,27 @@ private void colorIgnoredPlayers(Color ignoreColor)
...
+	private void updateClanTitle(WidgetInfo widget, ClanChannel channel, boolean show)
+	{
+		Widget header = client.getWidget(widget);
+
+		if (header != null && channel != null)
+		{
+			Widget title = header.getChild(0);
+			int memberCount = channel.getMembers().size();
+			String clanName = channel.getName();
+			title.setText(show ? String.format("%s (%d)", clanName, memberCount) : channel.getName());
```
Adam-, line 778: It looks to me like CLAN_CHAT_CHANNEL_REBUILD already assigns this text to the clan name - so all this needs to do is append the count to the existing text.

runelite-api/src/main/java/net/runelite/api/ScriptID.java
```diff
@@ -199,6 +199,12 @@
 	@ScriptArguments(integer = 15)
 	public static final int FRIENDS_CHAT_CHANNEL_REBUILD = 1658;
 
+	/**
+	 * Builds the widget that holds all of the players inside a clan chat
+	 */
+	@ScriptArguments(integer = 15)
+	public static final int CLAN_CHAT_CHANNEL_REBUILD = 4396;
```
Adam-, line 206: this is actually named `CLAN_SIDEPANEL_DRAW`

runelite-api/src/main/java/net/runelite/api/ScriptID.java
```diff
@@ -199,6 +199,12 @@
 	@ScriptArguments(integer = 15)
 	public static final int FRIENDS_CHAT_CHANNEL_REBUILD = 1658;
 
+	/**
+	 * Builds the widget that holds all of the players inside a clan chat
+	 */
+	@ScriptArguments(integer = 15)
```
Adam-, line 205: this is actually `7`

runelite-client/src/main/java/net/runelite/client/plugins/chatchannel/ChatChannelPlugin.java
d-as: Does that mean that I don't need to worry about resetting the title text if a player toggles off the setting/plugin, since the script does it anyway (and can be updated immediately by clicking the refresh button next to the clan title)? I removed the "show" flag from this, which makes it more simple.

Adam-: We could just client.runScript() one of the listeners maybe to cause it to rebuild. But just ignoring it is probably fine to if there is a button to reload it / it rebuilds so frequently anyway

Issue #13504, opened 2021-04-20
Idle Notifier Support for Creating Jugs of Water


NixyOSRS:
When turning a Jug into a Jug of Water with a sink or fountain, the Idle Notifier does not trigger upon completion of an inventory. I believe this should happen under the Idle Animation Notifications, as there is a 'reach emote' to the water spot you're utilising.

While I totally appreciate this is fairly niche to ironmen and not at all a priority, it would be handy to have as part of the plugin.

FlashBasedGod:
You have my sympathy here because I had to fill up 2,000 jugs on my ironman. Looking into it, however, it seems the animation for filling up buckets/jugs is a fairly general animation, and including it could cause some annoying issues with the idle notification triggering at unintended times. Regardless we can at least try and add it.

Nightfirecat:
In what situations would it be annoying to add notification triggers to that animation? For people who may be incidentally encountering that animation during active play, it wouldn't trigger anyway due to their activity.

arnaudperalta:
This enhancement would require to edit the test `checkAnimationReset()` in IdleNotifierPluginTest.java. I'm a beginner in this project and I wonder why this test exists in a first time ? How can an animation can be resets and why the player should use the LOOKING_INTO animation to reset it the first one ?

Nightfirecat:
That test exists to ensure that changing from some animation which would trigger a notification if idling afterward would not trigger if some other (non-triggering) animation occurred between them. I don't know why `LOOKING_INTO` was chosen instead of some dummy value, so I agree the test should be changed to accommodate this enhancement.

Nightfirecat:
Added in d58cecdb8d5a2d6958a5bcf1381805c6d64be7b8

PR #14109, opened 2021-09-08
idle notifier: LOOKING_INTO animation is now handled + test modified

arnaudperalta:
Closes https://github.com/runelite/runelite/issues/13504

Animation LOOKING_INTO is added to the list of animation who can trigger the notifier. Hence, the jug filling action can trigger the notifier.

The test has been changed with a Death animation instead of the `LOOKING_INTO` one.

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
@@ -62,6 +62,8 @@ import org.mockito.junit.MockitoJUnitRunner;
 @RunWith(MockitoJUnitRunner.class)
 public class IdleNotifierPluginTest
 {
+	private static final int DUMMY_ANIM = -2;
+
 	@Mock
 	@Bind
 	private Client client;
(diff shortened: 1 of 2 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPluginTest.java
```diff
@@ -153,7 +153,7 @@ public void checkAnimationReset()
 		animationChanged.setActor(player);
 		plugin.onAnimationChanged(animationChanged);
 		plugin.onGameTick(new GameTick());
-		when(player.getAnimation()).thenReturn(AnimationID.LOOKING_INTO);
+		when(player.getAnimation()).thenReturn(AnimationID.DEATH);
```
Nightfirecat, line 156: Rather than changing this footgun into another footgun, I'd like to see some constant defined in the test and used here instead of some other real known animation. eg. `private static final int UNKNOWN_ANIM = -2;`

arnaudperalta: Makes sense :) I created a DUMMY_ANIM attribute to the class

makingcomment:
I think this would also close #13331 and #12905

Nightfirecat:
Added in d58cecdb8d5a2d6958a5bcf1381805c6d64be7b8

Issue #14086, opened 2021-09-01
Inventory tag functionality with gravestone


wzrddd:
Would be amazing if when in a gravestone interface to reclaim your items it displayed the inventory tag colours you have already set on your items, thanks in advance

PR #14113, opened 2021-09-09
feat: inventory-tags implemented in gravestone interface

arnaudperalta:
Closes https://github.com/runelite/runelite/issues/14086

Gravestone widget added in the overlay.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsOverlay.java
@@ -32,13 +32,12 @@ import java.awt.Image;
 import java.awt.Rectangle;
 import java.awt.image.BufferedImage;
 import javax.inject.Inject;
+import static net.runelite.api.widgets.WidgetID.GRAVESTONE_GROUP_ID;
 import net.runelite.api.widgets.WidgetItem;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.ui.overlay.WidgetItemOverlay;
 import net.runelite.client.util.ColorUtil;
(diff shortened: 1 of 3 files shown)


Adam-:
The purpose of `showOnInventory` is for interfaces which either are the inventory, or replace the inventory when a modal is open in order to alter the inventory options (which is how banks, shops, etc work). To the player these are generally indistinguishable. 

The grave stone reclaim interface however is a full modal in the center of the viewport? Which isn't an inventory interface.

So instead it would be better to just `showOnInterfaces()` the gravestone interface.

arnaudperalta:
I understand the problem with my MR. Can I solve it by adding this line : `showOnInterfaces(GRAVESTONE_GROUP_ID);` in the InventoryTagsOverlay constructor instead of my modification in the WidgetItemOverlay class ? It is working as well.

Nightfirecat:
Yes, that is what he suggested.

runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsOverlay.java
```diff
@@ -38,6 +38,8 @@
 import net.runelite.client.util.ColorUtil;
 import net.runelite.client.util.ImageUtil;
 
+import static net.runelite.api.widgets.WidgetID.GRAVESTONE_GROUP_ID;
```
Nightfirecat, line 41: Move this import between the `javax.inject.Inject` and `net.runelite.api.widgets.WidgetItem` imports; we don't separate static and non-static imports.

arnaudperalta: Done !

PR #14115, opened 2021-09-10
cannon: update cannon location when interacting with the player's cannon

### LINKED ISSUE #8106: Cannon Overlay not working at Cave Horrors

**Describe the bug**
The cannon overlay does not work in the mos le harmless caves, in particular when cannoning cave horrors.

**To Reproduce**
1. Go to Mos Le Harmless caves.
(not sure if it makes a difference but I am not wearing a witchwood icon)
2. Place cannon in the recommended cannon spot.

**Expected behavior**
The cannon will be setup but there will be no overlay like there is in other places. The cannonball indicator tile does not update either.

**Screenshots**
image

 - OS: Windows 10
 - RuneLite version: 1.5.14
 - Launcher version: 1.6.0

**Additional context**

Hydrox6:
When interacting with another player's cannon, a completely different
message is returned. Using this, if we ever get a normal cannon message
after clicking on a cannon, we can be 100% certain that it is ours.

Closes #8106 by allowing the player to fix it without a re-place

Also cleaned up the mess that is `onChatMessage`, which was doing an unnecessary amount of string operations on cannon-related messages.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -41,17 +41,21 @@ import net.runelite.api.InventoryID;
 import net.runelite.api.Item;
 import net.runelite.api.ItemContainer;
 import net.runelite.api.ItemID;
+import net.runelite.api.MenuAction;
+import net.runelite.api.ObjectID;
 import static net.runelite.api.ObjectID.CANNON_BASE;
 import net.runelite.api.Player;
 import net.runelite.api.Projectile;
 import static net.runelite.api.ProjectileID.CANNONBALL;
 import static net.runelite.api.ProjectileID.GRANITE_CANNONBALL;
+import net.runelite.api.coords.LocalPoint;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameObjectSpawned;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.ItemContainerChanged;
+import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.ProjectileMoved;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -257,6 +262,19 @@ public void onGameObjectSpawned(GameObjectSpawned event)
 			}
 		}
 	}
+	
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked event)
+	{
+		// Check for the Fire option being selected on the cannon.
+		if (event.getMenuAction() != MenuAction.GAME_OBJECT_FIRST_OPTION || event.getId() != ObjectID.DWARF_MULTICANNON)
```
Nightfirecat, line 270: This should also support using cannonballs on a cannon as that would also reload a cannon.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
Hydrox6: Good shout. Fixed, and also found (and fixed) a weird tick-perfect issue where the indicator could appear on another cannon.

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -77,6 +81,8 @@
 	private CannonCounter counter;
 	private boolean skipProjectileCheckThisTick;
 	private boolean cannonBallNotificationSent;
+	private WorldPoint clickedCannonLocation;
+	private boolean firstCannonLoad = false;
```
Nightfirecat, line 85: `false` is the default initialized value for fields.
```suggestion
	private boolean firstCannonLoad;
```

Nightfirecat (commented):
LGTM

PR #14117, opened 2021-09-10
skill calculator: Define actions and bonuses in Java; add ItemSkillAction interface to remove explicit item name string entries

### LINKED ISSUE #14082: Move Skill Calculator definitions to Java
In order to implement features such as those described/POC'd in #14116 and #13726, skill action definitions and bonuses must be declared in code rather than deserialized from JSON. This will come with other advantages such as referencing ItemID and SpriteID definitions and possible run-time validation of these definitions. (ex. require entries be sorted by level requirement)

Nightfirecat:
Closes #14082

to-do:

- [x] Run this by folks to see if this is an agreeable solution in the first place
   - [ ] Accept or drop the `ItemSkillAction` interface commit
- [x] Actually write the javadoc for these interfaces
- [x] Add copyright headers (d'oh)


runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/UIActionSlot.java
```diff
@@ -121,7 +123,8 @@ public void mouseExited(MouseEvent mouseEvent)
 		uiInfo.setBackground(ColorScheme.DARKER_GRAY_COLOR);
 		uiInfo.setBorder(new EmptyBorder(0, 5, 0, 0));
 
-		JShadowedLabel uiLabelName = new JShadowedLabel(action.getName());
+		JShadowedLabel uiLabelName = new JShadowedLabel();
+		clientThread.invokeLater(() -> uiLabelName.setText(action.getName(itemManager)));
```
Adam-, line 127: You shouldn't call setText off-edt

PR #14125, opened 2021-09-11
cannon: add recommended spots for dagganoths and increase overlay distance

### LINKED ISSUE #13503: cannon plugin: Add recommended spot for dagganoth in Jormungand prison
It is a pretty good task, comparable to Lighthouse. I prefer cannoning them in the western room. Not sure which tile is optimal.

TalSk:
Adds two Dagganoth cannon spots in Jormungand's prison which closes #13503 
One spot it to the West and enjoys more low-level Dagganoths which are better due to their nonexistent Ranged defence, and one to the East which is more open and therefore has more targets to shoot at. The positions themselves are relatively optimal in terms of cannon range and respawn positions but may possibly be altered based on more concrete data.

Also took this opportunity to combine the max rendering distance for both cannon overlays (cannonballs left, double-hit spots when the cannon is placed and the optimal positions when the cannon is not placed) which were for some reason separate but almost identical. Also increased the distance itself to somewhere around the maximal distance in which you can also interact with the cannon itself based on some testing (couldn't find it there's a known figure for that).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonOverlay.java
@@ -39,11 +39,10 @@ import net.runelite.client.ui.overlay.OverlayPosition;
 import net.runelite.client.ui.overlay.OverlayPriority;
 import net.runelite.client.ui.overlay.OverlayUtil;
 import net.runelite.client.ui.overlay.components.TextComponent;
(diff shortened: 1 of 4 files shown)


Nightfirecat:
These changes (dag spot and extended overlay distance) should be added in separate commits.

TalSk:
Right, would you prefer both to be committed over current master as HEAD or on top of each other?

Nightfirecat:
They can be in the same PR

TalSk:
Great, done

Issue #14120, opened 2021-09-11
Change default skybox color in PoH with christmas theme to white


Manderville:
Currently it looks like this by default:
![2021-09-10 at 20 41 40](https://user-images.githubusercontent.com/45472577/132930667-2f024018-2377-4391-b6b0-a4befa8b99c8.png)

And I'd like it changed to this:
![2021-09-10 at 20 41 24](https://user-images.githubusercontent.com/45472577/132930679-487541ab-67c2-4696-8862-04712efe5c6b.png)

The only way to get it to white currently is by manually changing the overworld color to white but that also makes the entire overworld white too though. Thank you

JBerben:
@Manderville Is this what you were wishing for? 🎄

### Preview of skybox color
![SkyboxChristmasPoHToggle](https://user-images.githubusercontent.com/12043246/133029971-7acfe48d-de92-479b-abd5-e84f2517927e.png)

### Preview of the plugin panel
![SkyboxChristmasPoHTogglePanel](https://user-images.githubusercontent.com/12043246/133030036-1463c9a7-b65d-4b60-a7fa-e83d417d97ca.png)

Let me know if there is anything you would like changed or improved! (Even if it is a bit early for Christmas don't you think? 😉)

Manderville:
@JBerben Yeah sure that'd be great.

PR #14128, opened 2021-09-11
skybox: add christmas mode toggle for when in PoH. #14120

JBerben:
Adds a toggle in the skybox plugin to change the color whilst in a PoH to white.
Requested in #14120

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skybox/SkyboxPlugin.java
@@ -29,11 +29,13 @@ import com.google.inject.Provides;
 import java.awt.Color;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.api.Player;
 import net.runelite.api.coords.LocalPoint;
+import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.events.BeforeRender;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.client.config.ConfigManager;
@@ -56,6 +58,8 @@ public class SkyboxPlugin extends Plugin
 	private SkyboxPluginConfig config;
 
 	private Skybox skybox;
+
+	private final int[] REGION_POH = {7513, 7514, 7769, 7770};
 
 	@Override
 	public void startUp() throws IOException
@@ -110,6 +114,11 @@ public class SkyboxPlugin extends Plugin
 
 		Color overrideColor = player.getWorldLocation().getY() < Constants.OVERWORLD_MAX_Y
 			? config.customOverworldColor() : config.customOtherColor();
+		if (config.christmasMode() && Arrays.stream(REGION_POH).anyMatch(i -> i == WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID()))
+		{
+			client.setSkyboxColor(Color.WHITE.getRGB());
+			return;
+		}
 		if (overrideColor != null)
 		{
 			client.setSkyboxColor(overrideColor.getRGB());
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Is there no way to detect whether you're in a Christmas-themed POH?

JBerben:
> Is there no way to detect whether you're in a Christmas-themed POH?

Oh wow, I didn't even realize that Christmas-themed PoHs were a thing. I'll do that now!

Nightfirecat (changes requested):
This change should be made in [skybox.txt](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt), not in code.

JBerben:
> This change should be made in [skybox.txt](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt), not in code.

Do you think the toggle button should be removed too? I just made another commit before seeing your reply. I'll change it now.

Nightfirecat:
Yes. The only change this PR should include is the skybox definition update.

JBerben:
> Yes. The only change this PR should include is the skybox definition update.

All done now!

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
```diff
@@ -957,7 +957,14 @@ c 7 4
 
 m31 89
 
-// 31 89 p0 Hosidius
+// 31 89 p1 Christmas-themed POH
+p 1
+#FFFFFF
```
Nightfirecat, line 962: Pure white is a bit too bright [based on your screenshot](https://github.com/runelite/runelite/issues/14120#issuecomment-917861144). Maybe tone this down to something like `#BBB`? Play with the color and get it closer to the color of the snow.

Nightfirecat (changes requested):

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
abextm: the color should be `#b6cbd8`

JBerben: @abextm I just did so now in a PR.

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
```diff
@@ -957,7 +957,14 @@ c 7 4
...
+C 0 0 7 15
+
+m32 89
```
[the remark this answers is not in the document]
JBerben: My bad. I changed this. Everything should be in order now

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
```diff
@@ -955,6 +955,13 @@ c 6 3
 c 7 4
 
 
+m31 89
+
+// 31 89 p1 Christmas-themed POH
+p 1
+#B6CBD8
+C 0 0 7 15
+
```
abextm, line 964: you are missing dungeon colors still

JBerben: Oops. Do you think white would suit the dungeon, considering dungeons are underground? I was thinking either `#484840` or `#2c2c29`.

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
```diff
@@ -955,6 +955,22 @@ c 6 3
...
+p 1
+#251B09
+c 3 0
+c 0 3
+c 2 3
+c 4 3
+c 6 3
+c 7 4
+
 m31 89
```
Nightfirecat, line 974: The region does not need to be moved twice for the same region. Get rid of this line. See how 30,89 is handled above--move the region once, then change planes between each type.

runelite-client/src/main/resources/net/runelite/client/plugins/skybox/skybox.txt
```diff
@@ -955,6 +955,22 @@ c 6 3
...
+C 0 0 7 15
+// dungeon
+p 1
+#251B09
+c 3 0
+c 0 3
+c 2 3
+c 4 3
+c 6 3
+c 7 4
```
Nightfirecat, line 972: These should occur after the hosidius theme given that is on p0 and these are on p1.

Nightfirecat (changes requested):

PR #14129, opened 2021-09-11
timer plugin: overhaul handling Imbued Heart cooldown

### LINKED ISSUE #14091: Invigorating an Imbued heart while in the bank does not create a timer
The timer plugin code should be updated to allow the menu action type of the heart via the bank inventory.

TalSk:
Found a nice little Varbit (5361) that gets sent to the client in intervals of 6 containing remaining Imbued Heart cooldown time.
So instead of utilizing the menu entry coupled with the actual Magic boost to figure out when the player started the boost (which introduced an issue of "forgetting" remaining time when logging off or hopping worlds), it's switched to using information from the Varbit (which has a nice side-effect of "remembering" cooldown between logins).
Also fixes the bug of using the Heart while in the bank

Technically it would be a little cleaner to simply re-create the timer every 6 seconds, but for the sake of performance I simply saved a reference.

Closes #14091

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -39,7 +39,6 @@ import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.GameState;
 import net.runelite.api.InventoryID;
 import net.runelite.api.Item;
 import net.runelite.api.ItemContainer;
@@ -62,7 +61,6 @@ import net.runelite.api.events.GraphicChanged;
 import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.NpcDespawned;
-import net.runelite.api.events.StatChanged;
 import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.widgets.Widget;
 import static net.runelite.api.widgets.WidgetInfo.PVP_WORLD_SAFE_ZONE;
@@ -105,7 +103,6 @@ public class TimersPlugin extends Plugin
 	private static final String FROZEN_MESSAGE = "<col=ef1020>You have been frozen!</col>";
 	private static final String GAUNTLET_ENTER_MESSAGE = "You enter the Gauntlet.";
 	private static final String GOD_WARS_ALTAR_MESSAGE = "you recharge your prayer.";
-	private static final String IMBUED_HEART_READY_MESSAGE = "<col=ef1020>Your imbued heart has regained its magical power.</col>";
 	private static final String MAGIC_IMBUE_EXPIRED_MESSAGE = "Your Magic Imbue charge has ended.";
 	private static final String MAGIC_IMBUE_MESSAGE = "You are charged to combine runes!";
 	private static final String STAFF_OF_THE_DEAD_SPEC_EXPIRED_MESSAGE = "Your protection fades away";
@@ -153,14 +150,14 @@ public class TimersPlugin extends Plugin
 	private int lastPoisonVarp;
 	private int lastPvpVarb;
 	private int lastCorruptionVarb;
+	private int lastImbuedHeartVarb;
+	private boolean imbuedHeartTimerActive;
 	private int nextPoisonTick;
 	private WorldPoint lastPoint;
 	private TeleportWidget lastTeleportClicked;
 	private int lastAnimation;
 	private boolean widgetHiddenChangedOnPvpWorld;
 	private ElapsedTimer tzhaarTimer;
-	private int imbuedHeartClickTick = -1;
-	private int lastBoostedMagicLevel = -1;
 
 	@Inject
 	private ItemManager itemManager;
@@ -186,10 +183,6 @@ public class TimersPlugin extends Plugin
 	@Override
 	public void startUp()
 	{
-		if (client.getGameState() == GameState.LOGGED_IN)
-		{
-			lastBoostedMagicLevel = client.getBoostedSkillLevel(Skill.MAGIC);
-		}
 	}
 
 	@Override
@@ -205,8 +198,8 @@ public class TimersPlugin extends Plugin
 		nextPoisonTick = 0;
 		removeTzhaarTimer();
 		staminaTimer = null;
-		imbuedHeartClickTick = -1;
-		lastBoostedMagicLevel = -1;
+		imbuedHeartTimerActive = false;
+		lastImbuedHeartVarb = 0;
 	}
 
 	@Subscribe
@@ -218,6 +211,7 @@ public class TimersPlugin extends Plugin
 		int poisonVarp = client.getVar(VarPlayer.POISON);
 		int pvpVarb = client.getVar(Varbits.PVP_SPEC_ORB);
 		int corruptionCooldownVarb = client.getVar(Varbits.CORRUPTION_COOLDOWN);
+		int imbuedHeartCooldownVarb = client.getVar(Varbits.IMBUED_HEART_COOLDOWN);
 
 		if (lastRaidVarb != raidVarb)
 		{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -159,8 +159,8 @@
 	private int lastAnimation;
 	private boolean widgetHiddenChangedOnPvpWorld;
 	private ElapsedTimer tzhaarTimer;
-	private int imbuedHeartClickTick = -1;
-	private int lastBoostedMagicLevel = -1;
+	private TimerTimer imbuedHeartTimer;
```
Nightfirecat, line 162: This doesn't need to be a field, it's never referenced anywhere except to null it.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -308,6 +305,23 @@ else if (poisonVarp >= VENOM_VALUE_CUTOFF)
 
 			lastPvpVarb = pvpVarb;
 		}
+
+		if (lastImbuedHeartVarb != imbuedHeartCooldownVarb && config.showImbuedHeart())
+		{
+			if (imbuedHeartCooldownVarb == 0)
+			{
+				log.debug("The Imbued Heart has regained its magical power, clearing timer");
```
Nightfirecat, line 313: These log statements don't really help clarify anything and can be removed. (there's already debug logging for removal of infoboxes via `InfoboxManager`)

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -308,6 +305,23 @@ else if (poisonVarp >= VENOM_VALUE_CUTOFF)
...
+				removeGameTimer(IMBUEDHEART);
+				imbuedHeartTimer = null;
+			}
+			else if (imbuedHeartTimer == null)
+			{
+				log.debug("Setting Imbued Heart cooldown timer to " + 6L * imbuedHeartCooldownVarb + " seconds.");
+				imbuedHeartTimer = createGameTimer(IMBUEDHEART, Duration.of(6L * imbuedHeartCooldownVarb, ChronoUnit.SECONDS));
```
Nightfirecat, line 320:
```suggestion
				createGameTimer(IMBUEDHEART, Duration.of(10 * imbuedHeartCooldownVarb, RSTimeUnit.GAME_TICKS));
```

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -543,123 +544,43 @@ public void testThrallEnd()
...
-		imbuedHeartClick.setId(ItemID.IMBUED_HEART);
-		timersPlugin.onMenuOptionClicked(imbuedHeartClick);
-
-		when(client.getTickCount()).thenReturn(101);
-
-		for (int level = 1, i = 0; level <= Experience.MAX_REAL_LEVEL; level++, i++)
-		{
-			event = new StatChanged(Skill.MAGIC, 0, level, heartBoostedLevel(level));
-			timersPlugin.onStatChanged(event);
+		when(client.getVar(any(Varbits.class))).thenReturn(0);
```
Nightfirecat, line 550: This is a no-op. Mock return values use the default value of the method's type--for `int`, that is already 0. Remove this entry from all the heart tests.

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -543,123 +544,43 @@ public void testThrallEnd()
...
+		when(client.getVar(Varbits.IMBUED_HEART_COOLDOWN)).thenReturn(0);
+		timersPlugin.onVarbitChanged(new VarbitChanged());
 
-			ArgumentCaptor<InfoBox> captor = ArgumentCaptor.forClass(InfoBox.class);
-			verify(infoBoxManager, times(i + 1)).addInfoBox(captor.capture());
-			TimerTimer infoBox = (TimerTimer) captor.getValue();
-			assertEquals(GameTimer.IMBUEDHEART, infoBox.getTimer());
-		}
+		verify(infoBoxManager).removeIf(any());
```
Nightfirecat, line 569: This isn't specific enough to ensure the proper behavior is being followed. Any other varbit-managed infobox added in the future could trip this and cause a false pass. What coverage does this test add which the others don't? I think it can be removed.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -159,8 +159,8 @@
 	private int lastAnimation;
 	private boolean widgetHiddenChangedOnPvpWorld;
 	private ElapsedTimer tzhaarTimer;
-	private int imbuedHeartClickTick = -1;
-	private int lastBoostedMagicLevel = -1;
+	private TimerTimer imbuedHeartTimer;
+	private int lastImbuedHeartVarb = -1;
```
Nightfirecat, line 163: It doesn't look like this needs to default to `-1` because `0` is a no-op. (doesn't add a timer)

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -308,6 +305,23 @@ else if (poisonVarp >= VENOM_VALUE_CUTOFF)
...
+
+		if (lastImbuedHeartVarb != imbuedHeartCooldownVarb && config.showImbuedHeart())
+		{
+			if (imbuedHeartCooldownVarb == 0)
+			{
+				log.debug("The Imbued Heart has regained its magical power, clearing timer");
+				removeGameTimer(IMBUEDHEART);
+				imbuedHeartTimer = null;
+			}
+			else if (imbuedHeartTimer == null)
```
Nightfirecat, line 317:
```suggestion
			else
```

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -543,123 +544,43 @@ public void testThrallEnd()
...
 	}
 
 	@Test
-	public void testImbuedHeartBoostFromPartialBoost()
+	public void testImbuedHeartCooldown()
```
Nightfirecat, line 573: This is just the same test as `testImbuedHeartStart()`?

Nightfirecat (changes requested):
Functionality looks good, nice find on the varbit! Some changes needed to clean up code/tests.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
TalSk: There's a relevant check on line 317. Otherwise we would re-create the timer every 60 game ticks when the server sends an update on the varbit.
However I agree this should be changed to a flag, I saved a Timer reference since my first idea would be to update the time on varbit changed, which I later forgo.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
TalSk: See my comment above, this would recreate the timer every 60 game ticks. I opted to change this for a flag, let me know if that looks okay to you.

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
TalSk: It was created to replicate behaviour of logging in/hopping worlds while the cooldown is still active, hence the timer is inactive and the client gets some intermediate value from the server (in this test, 40). On another look I agree `testImbuedHeartStart` actually performs the same test internally so I removed it.

runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
TalSk: I altered it to check that the specific infobox for Imbued Heart was removed.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
TalSk: Actually, scratch my above comment. I simply add the timer once we get the initial cooldown value (70) and remove it on 0 and this can be removed. I don't know why I didn't think of it before :|

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -922,9 +922,9 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
 				}
 				break;
 			case LOGIN_SCREEN:
-				lastBoostedMagicLevel = -1;
 				// fall through
 			case HOPPING:
+				imbuedHeartTimerActive = false;
```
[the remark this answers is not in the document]
TalSk: You're right, good catch, I missed that. Removed in the latest commit.

PR #14132, opened 2021-09-12
woodcutting-plugin: control which tiers of clue scroll bird nest drops will be notified

### LINKED ISSUE #14025: Woodcutting plugin - ignore beginner's clue nests notification
Currently when woodcutting it is possible to get a notification whenever a bird's nest is dropped. This notification triggers for every single type of bird's nest. It would be nice to have some level of control over which nests trigger a notification, for example the ability to ignore beginner's clue scrolls.

The main reason I propose this for beginner's clues is because to most people they are not worth doing, but perhaps more importantly they don't even yield an empty nest when the clue is taken out.

TalSk:
Closes #14025 by adding new configuration which allows control on the tier of clue nests to start notifying from. 
I debated several implementations to discover which nest had spawned, ended up using item spawn hook which conveniently happens right before the chat message on nest drop is initiated. 
One could argue that controlling the starting tier from which to notify is too limiting and I'd argue that this is crazy.


runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/config/ClueNestTier.java
```diff
@@ -0,0 +1,39 @@
...
+public enum ClueNestTier
+{
+	BEGINNER(0),
+	EASY(1),
+	MEDIUM(2),
+	HARD(3),
+	ELITE(4),
+	DISABLED(999);
+
+	private final int tier;
```
Nightfirecat, line 36: This can be replaced by using `ClueNestTier.ordinal()` comparisons. I recommend using the nest item IDs instead so the tier can be fetched from the item spawn subscriber.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
```diff
@@ -107,6 +121,7 @@
 	private final List<TreeRespawn> respawns = new ArrayList<>();
 	private boolean recentlyLoggedIn;
 	private int currentPlane;
+	ClueNestTier clueTierSpawned = null;
```
Nightfirecat, line 124: `null` is the default value for object fields.
```suggestion
	ClueNestTier clueTierSpawned;
```

Nightfirecat: Also, this should be reset in shutDown or startUp

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
```diff
@@ -190,11 +205,31 @@ public void onChatMessage(ChatMessage event)
...
 
+	@Subscribe
+	public void onItemSpawned(ItemSpawned itemSpawned)
+	{
+		if (CLUE_NEST_IDS.contains(itemSpawned.getItem().getId()))
+		{
+			// It seems that falling nests spawn before the chat message is displayed, thus we can save state here and notify on game message.
+			clueTierSpawned = clueNestIdToTier.get(itemSpawned.getItem().getId());
+		}
+		else if (NON_CLUE_NEST_IDS.contains(itemSpawned.getItem().getId()))
```
Nightfirecat, line 226: This yields the same behavior?
```suggestion
		else
```

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
```diff
@@ -190,11 +205,31 @@ public void onChatMessage(ChatMessage event)
...
+		{
+			// Clear the tier spawned to prevent player nest drop from affecting notification.
```
Nightfirecat, line 228: How would a player dropped nest affect the notification? A nest has to spawn on the same tick the "A bird's nest falls out of the tree" chat message is received in order to affect the tier.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
TalSk: So, imagine we don't set `clueTierSpawned` to `null` here, and consider the following scenario:
1. Player sets clue tier notification to `Hard`
2. Player gets a beginner clue nest, doesn't get notified (which is fine), picks it up and then decides to drop it.
3. Upon this drop, `onItemSpawned` is invoked with the beginner clue nest itemId, causing `clueTierSpawned` to be equal `ClueNestTier.BEGINNER`
4. Now, imagine some regular seed bird nest is dropped. `onItemSpawned` will be invoked but without setting `clueTierSpawned` to `null`, so it'll still be `ClueNestTier.BEGINNER`.
5. "A bird's nest falls out of the tree" chat message will be then received, and the check ```if (clueTierSpawned == null || clueTierSpawned.ordinal() >= config.clueNestNotifyTier().ordinal())``` will not pass, since `clueTierSpawned` is not `null` and the ordinal of it is for `BEGINNER` which is less than the ordinal for `HARD` as set in config.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
TalSk: Yeah it does, but consider this (very unlikely) case where some random item spawns the same tick as a clue bird nest. We might get `onItemSpawned` invoked for this random item before the chat message will be processed by `onChatMessage`, changing `clueTierSpawned` to `null` by this change which might cause us to notify on a too-low tier clue nest. I think we rather only set it to `null` upon seeing an actual non-clue nest.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/config/ClueNestTier.java
TalSk: I moved to use ordinal, kinda forgot about it, thanks! As per the previous comment I moved the map to the enum so it absolves the need to use the nest item IDs, or any specific value whatsoever.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
```diff
@@ -190,11 +198,31 @@ public void onChatMessage(ChatMessage event)
...
+				// Clear the clue tier that has previously spawned
+				clueTierSpawned = null;
 			}
 		}
 	}
 
+	@Subscribe
+	public void onItemSpawned(ItemSpawned itemSpawned)
+	{
+		if (CLUE_NEST_IDS.contains(itemSpawned.getItem().getId()))
```
Nightfirecat, line 214: I'm quite wary of maintaining lists of bird nest IDs since these have changed over time--specifically more nests are added every now and then. Instead, because the one consistent feature of bird nests is that they have very consistent item names, I think the entirety of this subscriber would be better written as:

```java
final int spawnedItemId = itemSpawned.getItem().getId();
if ("Bird nest".equals(ItemManager.getItemComposition(spawnedItemId).getName()))
{
	clueTierSpawned = null;
}
else
{
	clueTierSpawned = ClueNestTier.getTierFromItem(spawnedItemId);
}
```

This would remove the need for the `CLUE_NEST_IDS` and `NON_CLUE_NEST_IDS` lists entirely.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
Nightfirecat: Your proposed scenario is not accurate as step 4 would reach this code branch and set the spawned tier to `null`.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
```diff
@@ -107,6 +113,7 @@
 	private final List<TreeRespawn> respawns = new ArrayList<>();
 	private boolean recentlyLoggedIn;
 	private int currentPlane;
+	ClueNestTier clueTierSpawned;
```
Nightfirecat, line 116: Reduce this visibility
```suggestion
	private ClueNestTier clueTierSpawned;
```

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
TalSk: Forgive me, I thought that you wanted me to clarify why my comment exists; it does because it explains why the line below it is necessary. Which, it is - if you imagine this code branch (setting the spawned tier to `null`) does not exist, then a player dropping a clue nest could trigger my scenario and introduce a bug

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingPlugin.java
TalSk: I understand, and you're right, I haven't given new item release enough thought. Your suggestion might not work without changing `getTierFromItem` to return `null` on getting itemId not in `CLUE_NEST_IDS` however (and this creates a somewhat unnecessary code since checking the item name for the string `"Bird nest"` is redundant).

If I think about what we expect from the code once an update is released adding some new kind of nests or stuff that falls from trees,
1. Not notify the player _only_ if we're _certain_ that a clue nest of low tier has dropped.
2. Still work if an unknown type of nest fell.
With this in mind, we have two options for `onItemSpawned`, tell me what you think is best:
- Live with the rare bug I mentioned in another comment and always reset `clueTierSpawned` like you already suggested
```
if (CLUE_NEST_IDS.contains(itemSpawned.getItem().getId()))
{
	clueTierSpawned = ClueNestTier.getTierFromItem(itemSpawned.getItem().getId());
}
else
{
	clueTierSpawned = null;
}
```
- Try to guess that any new nests will include the specific string "nest" in them (or anything else you think is better)
```
if (CLUE_NEST_IDS.contains(itemSpawned.getItem().getId()))
{
	clueTierSpawned = ClueNestTier.getTierFromItem(itemSpawned.getItem().getId());
}
else if (itemManager.getItemComposition(itemSpawned.getItem().getId()).getName().contains("nest"))
{
	clueTierSpawned = null;
}
```

Nightfirecat: [`Map#get(Object key)` returns `null` if the given key isn't present in the map](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#get-java.lang.Object-), so `getTierFromItem` already behaves the way you described. I'm not really sure what you mean by the "Bird nest" string check being redundant--[clue nests are named "Clue nest (tier)", and all other bird nests are named "Bird nest".](https://chisel.weirdgloop.org/moid/item_name.html#/nest/)

With those two points in mind, my code handles all the following cases:
* Non-clue bird nest is dropped (not by woodcutting): chat message subscriber isn't fired, no notification
* Non-clue bird nest is dropped by woodcutting: `clueTierSpawned` is set to null, chat message subscriber is fired
* Clue nest is dropped by the player: chat message subscriber isn't fired, no notification
* Clue nest is dropped by woodcutting: `clueTierSpawned` is set to the appropriate tier, chat message subscriber is fired
* Non-nest item is dropped: `clueTierSpawned` is set to null (because non-clue nest items aren't in the map), chat message subscriber isn't fired

The only case which could use coverage would be another (or multiple other) items spawning on the same tick as a bird nest spawn. To handle this, I'd change the following:
1. Set `clueTierSpawned` to null on GameTick. This event fires at the end of a tick, after other subscribers execute, so it will not interfere with the other changes in this PR, and simply resets the value after evaluation
2. Skip all item spawn checks if `clueTierSpawned` is not null--other item spawns don't matter if a clue nest has spawned, as they wouldn't impact the value

TalSk: Thanks! Didn't know this information about `get`, good to know. Fixed this as per your suggestion.
Just to clarify on my comment:
```java
final int spawnedItemId = itemSpawned.getItem().getId();
// if ("Bird nest".equals(ItemManager.getItemComposition(spawnedItemId).getName()))
// {
// 	clueTierSpawned = null;
// }
else
{
	clueTierSpawned = ClueNestTier.getTierFromItem(spawnedItemId);
}
``` 
Would be identical to your code since `clueTierSpawned` will be set to null even if the item is a non-clue bird nest since `getTierFromItem` will return `null`

runelite-client/src/test/java/net/runelite/client/plugins/woodcutting/WoodcuttingPluginTest.java
```diff
@@ -127,4 +133,96 @@ public void testBirdsNest()
...
+		TileItem nestTileItem = mock(TileItem.class);
+		when(nestTileItem.getId()).thenReturn(ItemID.BIRD_NEST_22798);
+		ItemSpawned regularNestSpawned = new ItemSpawned(tile, nestTileItem);
+
+		when(woodcuttingConfig.showNestNotification()).thenReturn(true);
+
+		// Player drops nest
+		woodcuttingPlugin.onItemSpawned(beginnerClueSpawned);
+		// Assume that at least one tick has passed
+		woodcuttingPlugin.onGameTick(null);
```
Nightfirecat, line 222: This should verify no interactions have happened with `notifier` at this point.
```suggestion
		// Player drops nest
		woodcuttingPlugin.onItemSpawned(beginnerClueSpawned);
		woodcuttingPlugin.onGameTick(null);
		verifyNoInteractions(notifier);

```

runelite-client/src/test/java/net/runelite/client/plugins/woodcutting/WoodcuttingPluginTest.java
```diff
@@ -127,4 +133,96 @@ public void testBirdsNest()
...
+		TileItem beginnerTileItem = mock(TileItem.class);
+		when(beginnerTileItem.getId()).thenReturn(ItemID.CLUE_NEST_BEGINNER);
+		ItemSpawned beginnerClueSpawned = new ItemSpawned(tile, beginnerTileItem);
+
+		when(woodcuttingConfig.showNestNotification()).thenReturn(true);
+		when(woodcuttingConfig.clueNestNotifyTier()).thenReturn(ClueNestTier.HARD);
+		woodcuttingPlugin.onItemSpawned(beginnerClueSpawned);
+		woodcuttingPlugin.onChatMessage(nestChatMessage);
+		woodcuttingPlugin.onGameTick(null);
+		verifyNoMoreInteractions(notifier);
```
Nightfirecat, line 202: This use and the one in `testClueNestDisabledConfig()` should be `verifyNoInteractions(notifier)` if no interactions were verified to begin with.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/config/ClueNestTier.java
```diff
@@ -0,0 +1,52 @@
...
+{
+	BEGINNER,
+	EASY,
+	MEDIUM,
+	HARD,
+	ELITE,
+	DISABLED;
+
+
+	private static final ImmutableMap<Integer, ClueNestTier> clueNestIdToTier = new ImmutableMap.Builder<Integer, ClueNestTier>()
```
Nightfirecat, line 40: Constant naming convention
```suggestion
	private static final ImmutableMap<Integer, ClueNestTier> CLUE_NEST_ID_TO_TIER = new ImmutableMap.Builder<Integer, ClueNestTier>()
```

Nightfirecat (changes requested):
One additional test should be added to have coverage of the `clueTierSpawned == null` check in onItemSpawned; that test should have a clue nest and some other non-clue-nest item spawn, in that order, on the same tick.

Changes LGTM otherwise.

TalSk:
All done regarding last changes requested.

PR #14138, opened 2021-09-13
Fix StatusBars overheal in Last Man Standing

### LINKED ISSUE #11378: Status bars for LMS
Runelite's status bars max out where your current player's max hp is.  When you start an LMS game your combat stats including HP go to 99 and def goes to 75. 

So if I make a new account with 10 hp, my hp level is boosted to 99 but my max hp stays at 10 and in turn, my health status bar also stays at 10. This gets very confusing when fighting as you don't know what proportion of health you are missing when glancing at the status bar. 

The number on the status bar is correct but it would be nice if the graphic corresponding to your current hp in proportion form 0-99. 

So in other words, if you are in LMS your status bar total health in the health bar code should be set to 99 in order to properly display your correct proportion of hp left. This would not apply to prayer as prayer does not drain in LMS.

pwatts6060:
In Last Man Standing your health and prayer is kind of faked to 99. Your current health is real but the real skill level is kept as your one outside of LMS which breaks the status bars during the minigame. This fixes that.
closes #11378

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
@@ -246,20 +246,18 @@ class StatusBarsOverlay extends Overlay
 		final Point offsetLeft = curViewport.getOffsetLeft();
 		final Point offsetRight = curViewport.getOffsetRight();
 		final Point location = curWidget.getCanvasLocation();
-		final int width, height, offsetLeftBarX, offsetLeftBarY, offsetRightBarX, offsetRightBarY;
+		final int height, offsetLeftBarX, offsetLeftBarY, offsetRightBarX, offsetRightBarY;
 
 		if (curViewport == Viewport.RESIZED_BOTTOM)
 		{
-			width = config.barWidth();
 			height = RESIZED_BOTTOM_HEIGHT;
-			offsetLeftBarX = (location.getX() + RESIZED_BOTTOM_OFFSET_X - offsetLeft.getX() - 2 * (width - BarRenderer.DEFAULT_WIDTH));
+			offsetLeftBarX = (location.getX() + RESIZED_BOTTOM_OFFSET_X - offsetLeft.getX());
 			offsetLeftBarY = (location.getY() - RESIZED_BOTTOM_OFFSET_Y - offsetRight.getY());
-			offsetRightBarX = (location.getX() + RESIZED_BOTTOM_OFFSET_X - offsetRight.getX() - (width - BarRenderer.DEFAULT_WIDTH));
+			offsetRightBarX = (location.getX() + RESIZED_BOTTOM_OFFSET_X - offsetRight.getX());
 			offsetRightBarY = (location.getY() - RESIZED_BOTTOM_OFFSET_Y - offsetRight.getY());
 		}
 		else
 		{
-			width = BarRenderer.DEFAULT_WIDTH;
 			height = HEIGHT;
 			offsetLeftBarX = (location.getX() - offsetLeft.getX());
 			offsetLeftBarY = (location.getY() - offsetLeft.getY());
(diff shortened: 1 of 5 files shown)


pwatts6060:
image description
Works in Last Man Standing now, before this would show me as full hp on my 10 hp account

runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsPlugin.java
```diff
@@ -72,6 +75,7 @@
 	private boolean barsDisplayed;
 
 	private int lastCombatActionTickCount;
+	boolean inLMS;
```
Nightfirecat, line 78:
```suggestion
	@Getter(AccessLevel.PACKAGE)
	private boolean inLMS;
```

and change reads to `plugin.isInLms()`

runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsPlugin.java
```diff
@@ -133,4 +137,22 @@ else if (client.getTickCount() - lastCombatActionTickCount >= config.hideAfterCo
 			barsDisplayed = false;
 		}
 	}
+
+	@Subscribe
+	public void onWidgetClosed(WidgetClosed ev)
```
Nightfirecat, line 142: This works fine while the plugin is running, but does not work if the plugin is toggled on while in an LMS match.

runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsOverlay.java
```diff
@@ -113,7 +113,7 @@ private void initRenderers()
 	{
 		barRenderers.put(BarMode.DISABLED, null);
 		barRenderers.put(BarMode.HITPOINTS, new BarRenderer(
-			() -> client.getRealSkillLevel(Skill.HITPOINTS),
+			() -> plugin.inLMS ? 99 : client.getRealSkillLevel(Skill.HITPOINTS),
```
Nightfirecat, line 116:
```suggestion
			() -> plugin.inLMS ? Experience.MAX_REAL_LEVEL : client.getRealSkillLevel(Skill.HITPOINTS),
```

Nightfirecat (changes requested):
Even if this PR doesn't handle other plugins which would be affected by LMS stat changes (opponent info, boosts info), it should also change the prayer bar to use a max of 99 as well.

pwatts6060:
> Even if this PR doesn't handle other plugins which would be affected by LMS stat changes (opponent info, boosts info), it should also change the prayer bar to use a max of 99 as well.

Hmm yeh sure I'll do that. Prayer doesn't actually drain in lms but they might do like a 99/99 overlay in the plugin at some point (if not already an option) and it would show 99/1

runelite-client/src/main/java/net/runelite/client/plugins/statusbars/StatusBarsPlugin.java
pwatts6060: Fixed now, when the plugin turns on it checks if the widget is open

Nightfirecat: On second thought, `client.getWidget()` is actually very fast. This could be moved to the overlay instead, making the change into:

```java
private boolean inLms()
{
	return client.getWidget(WidgetInfo.LMS_KDA) != null;
}
```

as a method in the overlay class, with no plugin class changes needed.

pwatts6060:
@Nightfirecat Any update on this?

Nightfirecat:
[This comment](https://github.com/runelite/runelite/pull/14138#discussion_r719070700) is still outstanding, so this PR is needing changes.

Nightfirecat:
Added in c7ea7a7e145467b0c50a619e10fbcbbcd105f0db

PR #14145, opened 2021-09-14
Status Bars Plugin - Customize width of status bars for resizable mode

### LINKED ISSUE #12846: Customizable status bars width
Hello, I would like to suggest the option of making the width of the bars configurable, with the main idea being to make them thinner.

How is it now:
image 

As I suggest having an option:
image

When using some UI configurations the bars are on top of the columns and looks very nice, but with this Resizable Mode configuration the double bars to the left of the inventory take up a lot of space and perhaps giving a width option would make the settup cleaner.

I got this idea while using the skills progress bar plugin, as this plugin have exactly this option (*but for height).  

Thanks and sorry for bothering

pwatts6060:
closes #12846

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/BarRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/statusbars/BarRenderer.java
@@ -41,8 +41,11 @@ class BarRenderer
 	private static final Color OVERHEAL_COLOR = new Color(216, 255, 139, 150);
 	private static final int SKILL_ICON_HEIGHT = 35;
 	private static final int COUNTER_ICON_HEIGHT = 18;
-	private static final int WIDTH = 20;
 	private static final int BORDER_SIZE = 1;
+	private static final int MIN_ICON_AND_COUNTER_WIDTH = 16;
+	static final int DEFAULT_WIDTH = 20;
+	static final int MIN_WIDTH = 3;
+	static final int MAX_WIDTH = 40;
 	private final Supplier<Integer> maxValueSupplier;
 	private final Supplier<Integer> currentValueSupplier;
 	private final Supplier<Integer> healSupplier;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/statusbars/Viewport.java
```diff
@@ -34,16 +34,17 @@
...
+			new Point(20, -4), new Point(0, -4), false),
 	FIXED_BANK(WidgetInfo.BANK_CONTAINER, WidgetInfo.BANK_INVENTORY_ITEMS_CONTAINER,
-			new Point(20, -4), new Point(0, -4));
+			new Point(20, -4), new Point(0, -4), false);
 
 	private WidgetInfo container;
 	private WidgetInfo viewport;
 	private Point offsetLeft;
 	private Point offsetRight;
+	private boolean variableWidth;
```
Nightfirecat, line 49: It looks like this isn't used? Seems like this should be used to prevent bars drawn on the inventory UI from being drawn at variable sizes. (they'd look bad if they were larger or smaller than they currently are)

pwatts6060: I'll check but I tested the plugin and wrote code so that it only affects the width in resizeable. The bar width uses the DEFAULT_WIDTH when it's not resizeable.

Nightfirecat: It looks like it does only affect resizable (see the changes in StatusBarsOverlay), but this added field doesn't cause that behavior. One or the other should be removed/changed.

pwatts6060: oh I see it now sorry yeh I tried that way realized it wouldn't work and forgot about the variable I created

pwatts6060: should be good now @Nightfirecat

Nightfirecat:
During the process of reviewing this, I've added 3c009c8f54791a4dafeee1dbefc665962835d396 and 847c44c44c3fd012073e548b8ad468c1242ab047 to master to make the rendering code more readable, which in turn makes this PR smaller also.

Anyways I rebased this and I've added an enforced minimum width for drawing counters and icons as they look very bad when drawn on thin bars.

pwatts6060:
Hey still not merged, is something wrong with it?

Nightfirecat:
No I just haven't tested it much since reworking this.

pwatts6060:
Could you work on this again? Recently saw someone complaining about in on discord and it's also annoying for me in lms

PR #14172, opened 2021-09-20
party: toggleable party stats overlay

LlemonDuck:
The party plugin provides features that extend beyond the stats overlay, but does not provide the option to turn off the stats overlay. In large groups (depending on UI scaling), the stats overlay can take up an unnecessarily large portion of the screen.

Adds the ability to disable the stats overlay while keeping the other party features (tile marking, non-party plugin features like raid layout sharing, world map, etc.)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
@@ -33,6 +33,17 @@ import net.runelite.client.config.ConfigItem;
 public interface PartyConfig extends Config
 {
 	String GROUP = "party";
+	
+	@ConfigItem(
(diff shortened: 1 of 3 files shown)


Nightfirecat:
As was mentioned in Discord, each party members' hp/prayer overlay is toggleable via the party panel. Is this necessary?

LlemonDuck:
I was literally on my way to comment as well rn

When the party is started it would recreate the panel, I didn't check for whether this is also the case when a single member leaves/joins but it may be as well.

This would still be useful to default to never having these overlays, rather than manually hiding them for each party.

LlemonDuck:
PartyStatsOverlay was removed, and the new overlay supersedes it. Closing because stale + no longer applicable

PR #14197, opened 2021-09-27
Update Lizardman suggested cannon spot

### LINKED ISSUE #14184: Suggested cannon spot at Shayzien Lizardman Canyon outdated
Since the rework to the area the suggested cannon safespot isn't correct anymore.
image
image

geeckon:
Fixes #14184 
New suggested cannon spot shown in the picture below, as well as two spawn points on cannon double hit spots marked near cannon and the old suggested cannon spot marked for reference.
image


StephanBijzitter:
So uhm, this is still needed. The current suggestion is quite literally on a tile where a cannon cannot be placed.

PR #14203, opened 2021-09-28
High alch value and profit to GE buy interface

bebeli555:
Adds high alch value and profit to the GE buy interface below the actively traded price on the grandexchange plugin. The profit thats in the brackets is calculated including the nature rune price so it shows how much you would make if you buy the item with the actively traded price and high alch it.
image


austinmc32:
any updates on these? Id love to see this implemented

Lord-Caim:
this would be great

bebeli555:
I fixed and tested it now, its ready to merge, also set it's default option to be disabled

PR #14204, opened 2021-09-28
Change made to add F2p.wiki to the XP Updater plugin

ChrisNatePollock:
This adds the f2p.wiki for tracking ehp gains for players that are free to play.


Nightfirecat (commented):
1. Why have you modified the indents of all other parts of the files you made changes to?
2. Who owns f2p.wiki, and would they be interested in this change being made to RuneLite, which could cause many new requests to their servers?
3. Is this a change which would even be useful to most OSRS players? This strikes me as much better suited to be a hub plugin than a core change.

ChrisNatePollock:
1. My apologies sheer idiocy
2.  f2p.wiki is owned by personal friends of mine whom I informed about this plugin
3. you are correct that this would be niche to only f2p players looking to be a part of the xp tracking on the site thank you for this direction and I will investigate the creating of a pluginhub creation. My apologies as first contribution issues.

Nightfirecat:
This has been added in runelite/plugin-hub#1919

PR #14215, opened 2021-10-01
skill calculator: assorted tweaks

Hydrox6:
This is a collection of commits to do a few things:

* Use Guice injection for the panel and other elements. Back when I was working on #13726, abex mentioned that it would be worth using guice, so here it is. The change ends up being relatively small, and makes the constructors for a lot of things smaller.

* Adjusts the bonus system to actually use a multiplier. For whatever reason, back when the skill calc was made up of JSON files, it stored the multipliers as the value to be added to 1 to obtain the multiplier. This really does not make sense at all, so this makes it actually a direct multiplier.

* With the recent update to move away from JSON storage, the skill calc can no longer show the names of members items if the client loaded on an F2P world. This is less than ideal, and reveals a secondary issue of clutter for players on F2P worlds, who can't access quite a lot of the actions on some lists. To fix this, the skill calc no longer shows P2P actions on F2P worlds, and reloads if the player switches from P2P to F2P or vice versa.


Hydrox6:
Fixed them all. Some fun observations made during fixing this:

* For pure essence, it seems to only be F2P because of an accident back in 2006. Pure Essence exists because people were botting Rune Essence back in 2006 as the requirements were extremely low, and mining was basically the only way to get the resource needed to train Runecraft. Pure Essence requiring 30 mining and membership was quite a step up for those trying to abuse it. To compensate players for this sudden change, all existing Rune Essence was converted into Pure Essence with that update, but this left F2P players with an item they couldn't use or trade, as it was only used for P2P Runecrafting. The next week, Jagex made the Pure Essence available to be used in F2P, both for trading and crafting, though F2P players could not obtain the essence themselves outside of said trading.
* Bracelets have 2 Item IDs associated with them: one of them being the item you wear in game, and the other being specifically for the jewellery crafting interface in furnaces. The items for the interface are therefore F2P, untradeable, have a value of 1, and lack the wear option, so they'd be useless if they could somehow be obtained in the game itself (though they still have the correct models for players to wear, interestingly enough). What's odd about this is that crafting bracelets is entirely a members activity, and the mould itself is also a members object, without which the interface will show "You need a bracelet mould to craft bracelets" instead. This therefore exists specifically for the situation that a P2P player takes a bracelet mould to F2P and tries to craft something.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/CookingAction.java
```diff
@@ -160,4 +161,16 @@
...
+
+	CookingAction(int itemId, int level, float xp)
+	{
+		this(itemId, level, xp, false);
+	}
+
+	@Override
+	public boolean isMembers(final ItemManager itemManager)
+	{
+		return isMembersOverride() || itemManager.getItemComposition(getItemId()).isMembers();
```
Nightfirecat, line 174: I've realized now that the right side of these overrides should be calling `ItemSkillAction.super.isMembers(itemManager)` instead of duplicating the code it contains.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/CraftingAction.java
```diff
@@ -74,26 +75,26 @@
...
-	SAPPHIRE_BRACELET(ItemID.SAPPHIRE_BRACELET, 23, 60),
+	SAPPHIRE_BRACELET(ItemID.SAPPHIRE_BRACELET_11072, 23, 60),
 	TIARA(ItemID.TIARA, 23, 52.5f),
 	SAPPHIRE_AMULET_U(ItemID.SAPPHIRE_AMULET_U, 24, 65),
 	UNFIRED_POT_LID(ItemID.UNFIRED_POT_LID, 25, 20),
 	POT_LID(ItemID.POT_LID, 25, 20),
 	JADE_NECKLACE(ItemID.JADE_NECKLACE, 25, 54),
 	WILLOW_BIRD_HOUSE(ItemID.WILLOW_BIRD_HOUSE, 25, 25),
-	DRIFT_NET(ItemID.SMALL_FISHING_NET_6209, 26, 55),
+	DRIFT_NET(ItemID.DRIFT_NET, 26, 55, true),
```
Nightfirecat, line 85: This does not need an override--it was only showing up incorrectly as a f2p item because it was entered as a small fishing net rather than a drift net.
```suggestion
	DRIFT_NET(ItemID.DRIFT_NET, 26, 55),
```

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/MiningAction.java
```diff
@@ -42,7 +43,14 @@
 	BARRONITE_DEPOSIT("Barronite deposit", 14, 32, ItemID.BARRONITE_DEPOSIT),
 	IRON_ORE("Iron ore", 15, 35, ItemID.IRON_ORE),
 	SILVER_ORE("Silver ore", 20, 40, ItemID.SILVER_ORE),
-	PURE_ESSENCE("Pure essence", 30, 5, ItemID.PURE_ESSENCE),
+	PURE_ESSENCE("Pure essence", 30, 5, ItemID.PURE_ESSENCE)
+	{
+		@Override
```
Nightfirecat, line 48: This override is well and good, but soft clay also needs an override.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/MiningAction.java
```diff
@@ -65,4 +73,10 @@
 	private final int level;
 	private final float xp;
 	private final int icon;
+
+	@Override
+	public boolean isMembers(final ItemManager itemManager)
```
Nightfirecat, line 78: During the course of reviewing this PR, I've realized this class can be reimplemented to `implements ItemSkillAction`, and that'll remove the need for this override. The only items I see needing overrides for their names are `BARRONITE_SHARDS` and `GEM_ROCKS`.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/SmithingAction.java
```diff
@@ -116,7 +117,14 @@
 	STEEL_SQ_SHIELD(ItemID.STEEL_SQ_SHIELD, 38, 75),
 	STEEL_WARHAMMER(ItemID.STEEL_WARHAMMER, 39, 112.5f),
 	STEEL_BATTLEAXE(ItemID.STEEL_BATTLEAXE, 40, 112.5f),
-	GOLD_BAR_GOLDSMITH_GAUNTLETS(ItemID.GOLD_BAR, 40, 56.2f),
+	GOLD_BAR_GOLDSMITH_GAUNTLETS(ItemID.GOLD_BAR, 40, 56.2f)
```
Nightfirecat, line 120: Speaking of related fixes, this could use a name override to differentiate it fro `GOLD_BAR`.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/CookingAction.java
Hydrox6: I tried to do that originally, but couldn't figure it out. Syntax feels kinda weird, tbh

Nightfirecat (approved):

Issue #14008, opened 2021-08-10
Ring of Forging count doesn't handle Varrock Platebody double smelting effect


Hydrox6:
When smelting in Edgeville furnace with the Varrock Armour on, you have a chance of smelting two bars at once. This is indicated by a filtered chat message (`The Varrock platebody enabled you to smelt your next ore simultaneously.`). When smelting Iron with the Ring of Forging, this causes the counter to desync by 1, as the extra bar used a charge from the ring.

Something very important to note is that the effect can trigger on the final ore in your inventory for whatever reason, meaning that you had the ability to smelt two bars at once, you didn't have the ore available.

Adam-:
fixed in #14220

PR #14220, opened 2021-10-03
Makes Ring of forging correctly reflect charges when using a Varrock platebody

JBarr2710:
Close issue #14008
Plugin: ItemChargePlugin
Makes ring of forging charges correctly reflect upon triggering the Varrock platebody effect.
Decrements charges by two if player has at least two iron ore in inventory.
Decrements charges by only one if player only has one piece of iron ore and the platebody effect is triggered.


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -342,6 +343,24 @@ else if (message.equals(RING_OF_FORGING_BREAK_TEXT))
 
 				updateRingOfForgingCharges(MAX_RING_OF_FORGING_CHARGES);
 			}
+
+			else if (message.equals(RING_OF_FORGING_VARROCK_PLATEBODY))
```
Adam-, line 347: Since this is largely duplicated from L321 I think you can just add it there using eg 

```java
int charges = Ints.constrainToRange(
  getItemCharges(ItemChargeConfig.KEY_RING_OF_FORGING) - message.equals(RING_OF_FORGING_VARROCK_PLATEBODY) ? 2 : 1,
  0, MAX_RING_OF_FORGING_CHARGES);
```

Issue #14219, opened 2021-10-03
Duck Highlight for Bird Slayer Task


6enny6:
Ducks (SW of champions guild) dont highlight under the Slayer plugin, they count as a 'birds' task from turael

Nightfirecat:
Added in #14224

PR #14224, opened 2021-10-04
Update slayer plugin to include "Duck" as a slayer monster for a birds task

Brett-MK:
Duck is missing from the list of monsters to kill for a bird slayer task.

Updated the birds targetNames to include "Duck" within the slayer plugin.


Brett-MK:
Closes #14219

PR #14247, opened 2021-10-08
Change Phosani's Nightmare icon in Highscores plugin

### LINKED ISSUE #14214: Phosani's Nightmare hiscore icon is incorrect.
**Describe the bug**
Phosani's Nightmare has a different icon than the regular version on the hiscores, RuneLite does not distinguish between the two visually.

**To Reproduce**
Look at the Phosani's Nightmare icon on the hiscore function.

**Expected behavior**
Weird egg thing displayed rather than Nightmare's head.

nocommentsgood:
Close #14214

Used provided image from the issue page on GitHub.
This way, there is two different icons for The Nightmare and for Phosani's Nightmare.
The image provided had to be scaled down, so there is quality loss.
Unsure if it meets RuneLite standards.


Adam-:
I've no real opinion on the icon itself, however, all boss icons must be 24x22, and also stripped with `mogrify -strip` or similar. You can see yours is likely not stripped as you are increasing the file size by 2400% despite your image being smaller.

Felanbird:
This change was completed in 0304007f4e3c61eab0d7c1b1e6947c8957426bda instead

PR #14248, opened 2021-10-08
Add collection log inventory ID

Equinox-:
Referenced based on https://github.com/RuneStar/cs2-scripts/blob/189cb2f94be24af3fb4596a3d41afa895cdf4345/scripts/%5Bproc%2Ccollection_draw_log%5D.cs2#L61


Nightfirecat:
We don't add API definitions to core unless we use them there. Do you need this addition for a hub plugin?

Equinox-:
I'm adding collection log tracking to my statistics plugin here: https://github.com/Equinox-/runelite-influxdb/blob/master/src/main/java/net/machpi/runelite/influxdb/InfluxDbPlugin.java#L172

I can work around this by manually checking for this inventory ID, but this felt like a nicer solution.  Either way is fine though

PR #14270, opened 2021-10-13
Add party persistence config and ability to join by party id.

alanbaumgartner:
I've seen requests in the discord to be able to persist parties between sessions so this PR adds that and since you need a party uuid to join a party, I also added the option to join a party by id and a button to copy your current parties id.

[gif of it in action.
](https://cdn.discordapp.com/attachments/836283317036122192/897990216948785272/Animation11.gif)

It's hard to see that I am copy pasting the uuid into the plugin search but its meant to show that after I had left and rejoined, the party id was the same.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
@@ -99,4 +99,23 @@ public interface PartyConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "previousPartyId",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	default String previousPartyId()
+	{
+		return "";
+	}
+
+	@ConfigItem(
+		keyName = "previousPartyId",
+		name = "",
+		description = "",
+		hidden = true
+	)
+	void setPreviousPartyId(String id);
 }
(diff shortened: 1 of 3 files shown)


robmonte:
Hopefully this can get reviewed/merged soon. It is very tedious to use parties as they are currently.

robmonte:
@alanbaumgartner It might also be worthwhile to change the name of this pull request to better reflect the feature changes you implemented. This may assist with its review since Adam has stated previously that persistence isn't a feature he will allow. Since it no longer adds party persistence and adds a rejoin button instead, could the name of the PR be updated to match that?

Adam-:
The party stuff is relatively expensive for us (like the % of our total userbase which we can have actually in a party at one time is actually quite small) - so having people unknowingly in a party would probably be very bad, and isn't something I would be willing to do without building out a much larger/more expensive party service.

Adam-:
Okay after re-reading this, it looks like this is just re-using party ids but it is not actually joining parties until you click on it - so this is probably actually fine. I will look at it later.

alanbaumgartner:
@robmonte @Adam- Yup, sorry I didn't update the PR name, I remember back when I had posted this PR I had posted in discord (Folks#0001 if you'd like to find the chat logs for it) and we had spoken about instead of auto joining a party on startup, to make it save the id so you can quickly hop into your previous party (with friends or whatever).

Thanks for merging 😄

PR #14277, opened 2021-10-17
hiscore: Add player lookup in group ironman tab

### LINKED ISSUE #14257: Add rightclick option to "Lookup" player from the Group Ironman team interface

BrastaSauce:
Close #14257

Adds the lookup option for players in the group ironman tab. Included a check to prevent the option from being added twice.


Adam-:
What is causing it to be added twice?

BrastaSauce:
I'm not sure why exactly but what happens is onMenuEntryAdded() runs fully twice. Once for the 'Message' option and once for the 'Remove Friend' option which ends up adding Lookup twice.

When I debugged it everywhere else Lookup is added it only ever ran once fully for one option in the menu, even if there was multiple.

Adam-:
`Message` and `Remove Friend` are two different menu options and so the event will run for both of them. You should just instead check `groupId == WidgetInfo.GROUP_IRON_LIST.getGroupId() && "Remove friend".equals(option)`

Adam-:
I rewrote some of the existing code around this since it was too hard to follow, can you rebase this?

PR #14280, opened 2021-10-17
Custom Fill Colors for Ground Item, Ground Marker, and Object Marker Plugins

### LINKED ISSUE #14233: Custom Ground Marker Fill Colors - Parity with Screen Markers and NPC Indicators
Screen Markers and NPC Indicators plug-ins currently offer the ability to change the fill color of a highlighted NPC or screen region.  

The Ground Markers, Ground Items, and Object Markers plug-ins should be brought to parity with this functionality for consistency, allowing the user to manipulate the fill color of highlighted tiles, items, and objects.

Currently:
- Ground Markers allows you to fill with a black overlay and adjust the opacity, and you can adjust the outline color.
- Ground Items allows you to change the color of the outline based on item values / highlighting rules, but no fill color option is available. Appears to apply a black overlay with low opacity at the moment in addition to the custom outline.
- Object Markers applies an outline, no option for fill color.

a-rokay:
I added functionality to allow the user to modify the fill colour of highlighted tiles and objects in the Ground Item, Ground Marker, and Object Marker Plugins, rather than being stuck with the stock black. This feature exists in the NPC Indicator plugin so it makes sense to add it here too.

image
image
image

This mainly consisted of altering the `renderPolygon` functions in each of the overlay managers to include the fill colour. 

For the Ground Items plugin, I added a `fillColor` attribute to the `PriceHighlight` class and altered the `getHighlighted` function to return `PriceHighlight` instead of `Color`, to easily access both colors from a single object which are then passed to `renderPolygon`. I had to do something similar for `ColorTileObjects` of the Object Marker plugin so the "Remember color" feature would work.

The only thing I'm unsure about is the Ground Items config with these new options. Let me know if you think this looks too crowded and I'd be happy to make changes.

image

Fixes #14233


LanaDelSlay:
Good stuff! One thing to make this even better is the plugin has an issue where it lets you set the fills' opacity beyond 255 which causes the fill to not show. You could do this in the config

```
@Range(
     max = 255
	)
@ConfigItem(
     keyName = "fillOpacity", 
     etc..
```

OneCent01:
This would be such a useful feature. Any progress on getting this merged in?

jevonsp:
any idea what issues need to be resolved before this can be put in the game?

OllyRN:
Any update on this being merged into the main client?

wheetcracker:
Just commenting to see if anyone has looked at this in the past year. I would very much like this improvement to be made.

PR #14281, opened 2021-10-17
Add ability to exclude specific items from bank value calculation

henrikstorck:
Most players probably have some items, which they would never (or rarely) sell, that cause the bank value to be higher than what the player is willing to liquidate.

Some examples are:
- untradeable currency (marks of grace, gold nuggets, crystal shards)
- skilling/cosmetic items (graceful outfit, twisted slayer helm)
- Imbued items (b-ring, archers ring)
- charged items (crystal tools, ring of suffering)
- items locked by trouver parchment (infernal cape (l))

This change will allow players to exclude certain items from the bank value calculation.
You can either add the items via a text field in the config panel:
image

Or via shift-clicking them in your bank:
image
image

Wildcard entries work aswell and you can disable the filtering by unchecking the "Ignore specific items" checkbox.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankConfig.java
@@ -69,10 +69,39 @@ public interface BankConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "ignoreSpecificitems",
+		name = "Ignore certain items",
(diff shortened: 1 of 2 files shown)


Hydrox6:
An item's value doesn't care about whether the item will be sold; it's a measure of wealth. This can already be done with a bank tag of all the items you don't care about.

As such, this is not a feature we're considering at all.

Issue #14279, opened 2021-10-17
Inventory tags not supported in raids 1 (cox) ?


Ogieeeeee:
When depositing items in the cox bank the inventory tags disappear.  Would be great to see the tags when the chest is opened.

Pic outside raid but it's the same when inside the raid.

![afbeelding](https://user-images.githubusercontent.com/22215187/137633555-7bacea5f-f1e7-4484-b7f0-9c80c5fd1ccb.png)

![afbeelding](https://user-images.githubusercontent.com/22215187/137633547-1827bc58-51c0-4124-89a0-7af358aea9a5.png)

Adam-:
added in 15dfc49d22974f8bacc08b44a852d35407de8afe

PR #14304, opened 2021-10-23
Issue#14279: Add Chambers_Of_Xeric_Storage units to InventoryTags 

BrianSchimert:
Add Chambers_Of_Xeric_Storage units to WidgetItemOverlay showOnInventory function to make InventoryTagsOverlay plugin correctly display inventory tags when COX storage units are open.

Closes issue #14279


Adam-:
added in 15dfc49d22974f8bacc08b44a852d35407de8afe

PR #14334, opened 2021-11-02
camera: Add left click drag to rotate

### LINKED ISSUE #6204: Enhanced Touch Screen Support
Playing OSRS on any touchscreen device, such as a Surface Pro/Go, is near impossible due to the inability to rotate the camera with touch input. 

Having a "touchscreen enabled" mode in the settings. Basically it would allow the user to rotate the camera by clicking and dragging instead of having to hold the middle mouse button. Also having right click be activated by clicking and holding would be nice. 

I believe this would be a pretty straight-forward feature to add and wouldn't be that difficult, but I have no real idea. I was thinking you could just have have dragged clicks be registered as being dragged middle mouse button clicks? Just an idea, I have no idea the best way to go about doing it. My current "homebrew" solution is having the button on the Surface Pen act as a middle mouse button.

aHooder:
Fixes #6204 by adding the following options to the camera plugin:

image

This should qualify as 1:1 actions in my opinion. I've avoided changing modifier bits for mouse events for simplicity, since as far as I can tell, OSRS only cares about the button parameter.

I've included some checks for time-sensitive widgets, like the quick-prayer orb, and non-clickthrough widgets. These should receive normal left clicks like before.


ajhakra:
Lil bump to see if this is still valid

MatthewRorke:
This _is_ still valid

aHooder:
This could be a hub plugin instead I think.

Yocairo:
This would be nice on my Surface Pro 11 😔 

@aHooder any plans to integrate this into the Camera Plugin (or a separate plugin?) otherwise is it okay if I try to continue this work?

aHooder:
> This would be nice on my Surface Pro 11 😔
> 
> @aHooder any plans to integrate this into the Camera Plugin (or a separate plugin?) otherwise is it okay if I try to continue this work?

Feel free to use any of it :slightly_smiling_face:

xaviergmail:
Why can't this be first-party?

Yocairo:
> Why can't this be first-party?

I think that the ARM64 userbase is too small to be considered currently, especially considering that we can just plug in a mouse and keyboard on devices like the Surface Pro. I just bought an iPad for OSRS mobile and am using my Surface Pro for other purposes.

PR #14335, opened 2021-11-03
add salmon cursor and rotate trout so head is mouse point

Jin-Jiyunsun:
Adds a salmon cursor and rotates the trout one so the "head" of the fish is the cursor point, before it was the tail. Not sure why as the head makes more sense 🤷.


Jin-Jiyunsun:
any issues with this?

Jin-Jiyunsun:
could someone review this and either close it or merge it?

Issue #13645, opened 2021-05-26
Incorrect phrasing in Time Tracking notifications


Hydrox6:
With some specific farming patches, crops, or both, the Time Tracking plugin outputs some very oddly phrased notifications. They still convey the information needed, they just look odd.

Examples:
![image](https://user-images.githubusercontent.com/2979691/119710744-a0d96a00-be56-11eb-9bc4-1aec552309ca.png)
`Seaweed` isn't a place

![image](https://user-images.githubusercontent.com/2979691/119710753-a3d45a80-be56-11eb-8a02-2266da4b4923.png)
Anima crops are technically dead when they're "ready", should probably mention that they're ready to be cleared

I've also seen `Your Magic is ready to harvest at Falador`, referring to a Magic tree, but I don't have a screenshot of it happening.

PR #14340, opened 2021-11-05
timetracking: change produce and patch names, add anima specific notification text

pirak3:
…learer and added different notification text for anima crops

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTracker.java
@@ -599,6 +599,10 @@ public class FarmingTracker
 				{
 					stringBuilder.append(" is ready to collect in ");
 				}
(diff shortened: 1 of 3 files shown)


pirak3:
Fixes #13645

pirak3:
I have changed the names of the trees but i don't know if this is the way we would like to do it. It might be better to add a seperate field to tree crops for their display name so the shortened form can be shown in different places and the longer one in the notifications.

Hydrox6 (approved):

Hydrox6:
The name is used in the contract infobox tooltip, the tooltips of the produce in the panel itself, and the notification messages. In all of those places, referring to eg. `Magic tree` ends up being better imo. Those places are not exactly short of space to fit the word `tree`.

Hydrox6 (changes requested):
The hardwood, redwood, spirit, calquat, and celastrus trees should all get the same treatment as the other trees

pirak3:
I also changed the trees you mentioned, thank you for the feedback

PR #14341, opened 2021-11-05
Add random option to login screen plugin

### LINKED ISSUE #11731: Add login screen background option "Random"

hjdarnel:
Closes #11731

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
@@ -33,6 +33,7 @@ import java.awt.event.KeyEvent;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Random;
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loginscreen/LoginScreenPlugin.java
```diff
@@ -287,6 +288,11 @@ private void overrideLoginScreen()
 				}
 			}
 		}
+		else if (config.loginScreen() == LoginScreenOverride.RANDOM)
+		{
+			LoginScreenOverride rand = LoginScreenOverride.values()[new Random().nextInt(LoginScreenOverride.values().length - 1)];
```
Nightfirecat, line 293: `length - 1` is not a clear indicator of the logic happening on this line. Furthermore, it would give greater weight to one login screen given that `OFF` is often/always synonymous with some other login screen, usually `NORMAL`.

Instead, you can filter the available login screens to a list of only the ones with non-null file names, then select a random entry from there.

hjdarnel: How does that look?

PR #14343, opened 2021-11-05
object-markers: fix highlights not showing after restart of plugin

### LINKED ISSUE #14090: Object Markers plugin loses highlights when it is shut down and started back up.
**Describe the bug**
Restarting the Object Markers plugin removes my object markers. Restarting the client works as expected, with the object markers still present on the objects I applied them to.

**To Reproduce**
Use Object marker plugin to mark an object.
Shut down Object Markers
Start up Object Markers
The object will no longer be marked.
Restart the client.
The object will be marked again.

**Expected behavior**
Marked objects stay marked after restarting the plugin

**Screenshots**

windows 10, runelite 1.7.21.1

**Additional context**
alexsuperfly said some stuff about it:

pirak3:
Adds flag that triggers points reload on startup of the plugin. This flag is only set when the shutdown function is called which in turn is only called when the user disables the plugin. This way the points are only reloaded when the user restarts the app.

fixes #14090

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
@@ -116,10 +116,16 @@ public class ObjectIndicatorsPlugin extends Plugin
 		return configManager.getConfig(ObjectIndicatorsConfig.class);
 	}
 
+	private static boolean reloadMapPointsOnStartup = false;
+
 	@Override
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		if (reloadMapPointsOnStartup){
+			reloadMapPoints();
+			reloadMapPointsOnStartup = false;
+		}
 	}
 
 	@Override
@@ -128,6 +134,7 @@ public class ObjectIndicatorsPlugin extends Plugin
 		overlayManager.remove(overlay);
 		points.clear();
 		objects.clear();
+		reloadMapPointsOnStartup = true;
 	}
 
 	@Subscribe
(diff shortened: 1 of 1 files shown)


hex-agon:
Please run checkstyle locally and fix the violations

pirak3:
ran checkstyle, my apologies i had forgotten to run it.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -116,10 +116,17 @@ ObjectIndicatorsConfig provideConfig(ConfigManager configManager)
...
 
+	private static boolean reloadMapPointsOnStartup = false;
+
 	@Override
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		if (reloadMapPointsOnStartup)
+		{
+			reloadMapPoints();
```
Nightfirecat, line 127: You can't be calling client methods from EDT. This must be wrapped in `clientThread.invokeLater()`.

runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsPlugin.java
```diff
@@ -116,10 +116,17 @@ ObjectIndicatorsConfig provideConfig(ConfigManager configManager)
 		return configManager.getConfig(ObjectIndicatorsConfig.class);
 	}
 
+	private static boolean reloadMapPointsOnStartup = false;
```
Nightfirecat, line 119: From what I can tell, this can be removed in favor of checking `if (points.isEmpty())` in `reloadMapPoints()`. In fact, renaming that method to `loadMapPoints()` and not clearing the points map inside of it would be best.

Nightfirecat (changes requested):

PR #14346, opened 2021-11-06
Draft: [Pyramid Plunder] Ability to highlight urns on penultimate floor until configurable seconds left

midasvo:


midasvo:
Cat walked over keyboard, wasn't ready to create MR

PR #14349, opened 2021-11-07
[Pyramid Plunder] Add ability to highlight urns from penultimate floor (until configured timer) and up

midasvo:
Currently the plugin allows you to set a floor from when to highlight urns, this change makes it a bit more plug-and-play because it checks your thieving level and matches it up with the penultimate floor. The wiki recommends to check urns on the penultimate floor until about half the timer has passed. In default state it will highlight the urns on the penultimate floor until ~2:30 remains indicating to the player they should move on to the last floor.

I added it to this plugin instead of creating a pluginhub plugin because to me it seems more intuitive than having to configure it yourself, however they are compatible.


midasvo:
Will run checkstyle asap

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -165,4 +190,75 @@ public Dimension render(Graphics2D graphics)
...
+			return currentFloor >= config.highlightChestFloor();
+		}
+		else if (SARCOPHAGUS_ID == objectId)
+		{
+			return currentFloor >= config.highlightSarcophagusFloor();
+		}
+		return false;
+	}
+
+	private Integer getPenultimateFloor()
```
dekvall, line 240: This looks like it can be `floor = (level - 1) / 10 -1` which feels less complex.

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderConfig.java
```diff
@@ -155,4 +155,20 @@ default int highlightSarcophagusFloor()
...
+
+	@ConfigItem(
+		position = 11,
+		keyName = "highlightPenultimateUrnsAndUp",
+		name = "Highlight urns from penultimate floor",
+		description = "Highlights the urns from the penultimate floor and up"
+	)
+	default boolean highlightPenultimateUrns() { return false;}
+
+	@ConfigItem(
```
dekvall, line 167: There's `@Units` you can use, it might be worth using game ticks here instead to make it eadier to compare with the timer

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -165,4 +190,75 @@ public Dimension render(Graphics2D graphics)
 
 		return null;
 	}
+
+	private void highlightObject(Graphics2D graphics, GameObject object)
+	{
+		ObjectComposition imposter = client.getObjectDefinition(object.getId()).getImpostor();
```
dekvall, line 196: `impostor`

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
```diff
@@ -118,7 +143,7 @@ public Dimension render(Graphics2D graphics)
 			if (!config.highlightDoors() && TOMB_DOOR_WALL_IDS.contains(object.getId())
 				|| !config.highlightSpeartraps() && SPEARTRAP_ID == object.getId()
 				|| tile.getPlane() != client.getPlane()
-				|| object.getLocalLocation().distanceTo(playerLocation) >= MAX_DISTANCE)
+				|| objectTooFarAway(object))
```
dekvall, line 146: I don't think this abstraction is needed

runelite-client/src/main/java/net/runelite/client/plugins/pyramidplunder/PyramidPlunderOverlay.java
midasvo: Ah that's a nice clean way of getting the max floor, I'll just use `floor = (level - 1) / 10 -2` and round down

dekvall: I think you should keep the method, so you can handle cases where players are below lvl 21 or have boosted above 100 here.

midasvo: Hm yeah I'll see what I can do with it

PR #14359, opened 2021-11-10
Fixes the flickering scrollbar when withdrawing or depositing items i…

geheur:
…n a bank tag while "Remove tab separators" is enabled.

I've added a new `runelite_callback` in bankmain_finishbuilding which allows plugins to change the bank scrollbar height argument. Fixing and re-using the old bank scroll height callback was not possible because it was coupled to the tag tab tab's skipping item drawing, so I removed it and updated the tag tab tab to use the new callback to set the scroll height.

Potential issues:
- I'm not sure what https://github.com/runelite/runelite/pull/12466 does or if it still happens after my change. I deleted the code for it but I didn't notice any change in the behavior it was supposed to change.


Adam-:
It looks to me like this regresses #13761 since it is removing the `isTagTabActive` check to rebuild the scroll yet not replacing it?

geheur:
It does not regress #13761. #13761 is re-implemented here in the change to TabInterface where a new callback for `setScrollbarHeight` is added. Scrollbar flickering occured with #13761, so it had to be changed to use the new `setScrollbarHeight` rather than the `invokeLater`.

the history of tag tab tab scrolling is:
1) initially it worked due to `setBankScroll`.
2) it broke when the script was incorrectly updated.
3) I fixed it using `invokeLater` in #13761.
4) Now I've updated it to not scrollbar flicker in this br.

Adam-:
fixed in c8ec004b7af874ab670b6494b9ba61a37db77502

PR #14363, opened 2021-11-11
chat filter: Ignore character accents for matching

### LINKED ISSUE #13097: Ignore special characters in chat filter Regex
Gambling and other scamming bots at G.E. have started using special characters to avoid chat filters.

Could we somehow make Chat Filter more resilient to this? Regexing all special characters seems to not be viable due to the amount of characters available.

Could we call Normalizer.normalize on all chat messages passed to Chat filter or something?

Nightfirecat:
This lets plain latin-character filters to match messages with accents
and diacritics which are not easily typed on all keyboard layouts.

Fixes #13097

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.regex.MatchResult;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
```diff
@@ -332,27 +333,38 @@ String censorMessage(final String username, final String message)
...
+							return null;
+					}
 				}
 			}
-			m.appendTail(sb);
 
-			strippedMessage = sb.toString();
+			for (MatchResult match : matches)
+			{
+				strippedMessage = strippedMessage.substring(0, match.start())
```
Adam-, line 364: This looks like it is applying indexes from the match result of `Text.stripDiacritics(strippedMessage)` into `strippedMessage` which wouldn't always be correct

PR #14364, opened 2021-11-12
xptracker: add "reset paused" menu entry

LlemonDuck:
Allows players to reset their paused skills only, removing skills that they are no longer working on without resetting the skills still being trained. Useful for reducing clutter when you forget to reset the panel between activities.

ezgif com-gif-maker


LlemonDuck:
Stale, undesirable

PR #14366, opened 2021-11-12
chat filter: Ignore character accents for matching

### LINKED ISSUE #13097: Ignore special characters in chat filter Regex

Nightfirecat:
This lets plain latin-character filters to match messages with accents
and diacritics which are not easily typed on all keyboard layouts.

This PR fixes a bug which was present in #14363 where `Text.stripDiacritics()` also stripped raw diacritic characters from input strings, meaning that the resulting string could be of a different length. This is now guaranteed as part of the method's execution and has a new test which covers this case.

Fixes #13097


Adam-:
added in 1362af414a1a7c4e6ec3fc678f00a9aecb233b6f

Issue #14353, opened 2021-11-08
Right Click Teleport for Max Cape


marlenekoh:
**Edit:** 
Feature Request to swap Max Cape 'Teleports' with 'Wear' in Menu Entry Swapper

**Is your feature request related to a problem? Please describe.**
Difficult to teleport when max cape in inventory. If I wear the max cape, I am able to teleport to POH but this requires me to change tab from inventory to worn equipment.

**Describe the solution you'd like**
Add feature request to configure teleport for menu entry swapper
- When max cape is inventory, swap wear with click to teleport like Ardougne cloak but choose from all available max cape teleports 
- Will be great if can configure teleport to POH, or to Crafting bank for example

**Describe alternatives you've considered**
Alternatively, add one more option to max cape context menu when right click from inventory

**Additional context**

> Menu entry swapper (Ardougne cloak)

![image](https://user-images.githubusercontent.com/28558150/140743286-462f626d-2be8-4bca-a27c-33093af29b49.png)

> Max cape context menu in inventory

![image](https://user-images.githubusercontent.com/28558150/140743738-91d3bc92-cc28-40d8-9d11-3038a1460093.png)

Alexsuperfly:
swapping menu options on equipped items and creating new menu options entirely which send an action to the server are both explicitly prohibited by jagex in their [third party statement ](https://secure.runescape.com/m=news/another-message-about-unofficial-clients?oldschool=1)

the general teleport swap in the plugin should work for the max cape, if not, we can leave this open as a feature request for adding that.

marlenekoh:
I see, I didn't read their third party statement.
The general teleport swap in the plugin doesn't seem to work for the max cape. I'll leave it open as a feature request then, thanks!!

Adam-:
added in #14379

PR #14379, opened 2021-11-16
max cape teleports swap

orange-puff:
This is related to https://github.com/runelite/runelite/issues/14353
The max cape teleport menu entry wasn't being swapped with wear by the plugin because it has a `Teleports` entry rather than a `Teleport` entry.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -66,6 +66,11 @@ public interface MenuEntrySwapperConfig extends Config
 	)
 	String uiSection = "ui";
 
+	enum MaxCapeMode{
+		WEAR,
+		TELEPORTS
+	}
+
 	enum ArdougneCloakMode
 	{
 		WEAR,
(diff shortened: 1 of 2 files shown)


Adam-:
If anything I would expect the `Teleports` to be part of the existing `Teleport` swap?

orange-puff:
Since I don't have a max cape I have no ability to test this. But according to the issue author the original plugin is not swapping the menu entries

Adam-:
My question is: why is "Teleports" not swapped with the existing Teleport item swap, and instead a new option is being added?

orange-puff:
Good point. Updated

orange-puff:
Updated to use the existing `Teleport` swap rather than an additional option

alizan:
Hi, I can see there was an update to the Max Cape being added to teleport item in menu entry swapper as per request in this case. 

However this is not ideal for some people that want to left-click wear for Max Cape, while keeping other teleport items left-click teleport as it was before recent update. Would it be possible to add toggable option for Max Cape similar to other teleport capes, as suggested in [https://github.com/runelite/runelite/issues/14353](url) ?

Nightfirecat:
8944eb9be6d62ec8384e29484dab14c4c2858d9d was just added to master, which would allow you to override the general teleport items swap on the max cape (or any other teleport items/jewellery).

Issue #14250, opened 2021-10-08
Timers Plugin - Group Iron Man Teleport


the-fifth:
Currently the Home teleport timer icon only shows when triggered via the spell book teleport, can the teleport timer be shown/track when the helmet teleport is used? They share the same cooldown.

Nightfirecat:
Fixed in #14842

PR #14384, opened 2021-11-17
Timer plugin support gim helm teleport

orange-puff:
This will create a `TimerTimer` with `GameTimer` enum value `HOME_TELEPORT` after the user successfully uses the `Teleport` on the `Group iron helm` or the `Hardcore group iron helm`. This item's teleport timer is the same as home teleport timer.  This PR is in response to https://github.com/runelite/runelite/issues/14250

This has been tested and appears to be working.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -205,6 +205,7 @@ public final class AnimationID
 	public static final int LEAGUE_HOME_TELEPORT_5 = 8805;
 	public static final int LEAGUE_HOME_TELEPORT_6 = 8807;
 	public static final int RAID_LIGHT_ANIMATION = 3101;
+	public static final int GROUP_IRON_HELM_TELEPORT = 714;
 
 	public static final int CONSTRUCTION = 3676;
 	public static final int CONSTRUCTION_IMCANDO = 8912;
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -205,7 +205,7 @@ public final class AnimationID
 	public static final int LEAGUE_HOME_TELEPORT_5 = 8805;
 	public static final int LEAGUE_HOME_TELEPORT_6 = 8807;
 	public static final int RAID_LIGHT_ANIMATION = 3101;
-	public static final int GROUP_IRON_HELM_TELEPORT = 714;
+	public static final int GENERIC_TELEPORT = 714;
 
 	public static final int CONSTRUCTION = 3676;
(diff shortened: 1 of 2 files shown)


runelite-api/src/main/java/net/runelite/api/AnimationID.java
```diff
@@ -205,6 +205,7 @@
 	public static final int LEAGUE_HOME_TELEPORT_5 = 8805;
 	public static final int LEAGUE_HOME_TELEPORT_6 = 8807;
 	public static final int RAID_LIGHT_ANIMATION = 3101;
+	public static final int GROUP_IRON_HELM_TELEPORT = 714;
```
Nightfirecat, line 208:
```suggestion
	public static final int GENERIC_TELEPORT = 714;
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -493,6 +481,13 @@ public void onMenuOptionClicked(MenuOptionClicked event)
 		}
 
 		TeleportWidget teleportWidget = TeleportWidget.of(event.getParam1());
+		// check for Group Iron Man Helm Teleport
+		if (event.getMenuOption().contains("Teleport")
+				&& event.getMenuTarget().toLowerCase(Locale.ROOT).contains("group iron helm"))
```
Nightfirecat, line 486: This should be checking the menu type and the clicked item ID instead of the target text.

Nightfirecat (changes requested):
This PR is problematic because it assumes the group ironman teleport is unique (it's not), so would restart the timer anytime you used a normal teleport after using the group ironman teleport since they share an animation.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
orange-puff: Do you man with `event.getId()`?

orange-puff: log.info("EventId: " + event.getId());
        log.info("Param0: " + event.getParam0());
        log.info("Param1 : " + event.getParam1());
        log.info("ActionParam: " + event.getActionParam());
Hmm... none of these seem to return the item id when the item is equipped 

Maybe I am missing something but it doesn't look like any of the other fields would return the item id

orange-puff:
I believe the most accurate way to tell if the user just used the gim tely in `onMenuOptionClicked` is by using `getMenuTarget` and I think I've avoided the problematic aspects of the original implementation.

Nightfirecat:
Home teleport timers are now based on vars, which work regardless of teleport method. Implemented here: #14842

PR #14403, opened 2021-11-23
api: Ultimate Ironman account type

Obasill:
Figured I'd add this since it's missing, also it's nice for one of the diaries I'm working on for the quest helper.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/vars/AccountType.java
+++ b/runelite-api/src/main/java/net/runelite/api/vars/AccountType.java
@@ -65,6 +65,16 @@ public enum AccountType
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Hydrox6:
The other functions check for a range of values; this is the same as `== AccountType.ULTIMATE_IRONMAN`

Obasill:
Ah gotcha, however I'm throwing errors unless I'm using `AccountType.ULTIMATE_IRONMAN.isIronman()`
For my if check, will this still work properly? I don't have an UIM account to check.

Hydrox6:
You are probably doing something very *very* wrong

Obasill:
Yeah, 7am and no sleep will do that to you. Thanks for the help.

PR #14405, opened 2021-11-23
Added Swap option for Seed box

Freeburn113:
Added config option and code to make it possible to swap the seedbox "fill" and "check" option


Nightfirecat:

PR #14419, opened 2021-11-26
fixes #13885 Ignore color tag when filtering messages

### LINKED ISSUE #13885: Unable to filter recolored chat using a starts with regex
**Describe the bug**
Recolored chat has a color tag appended to it so trying to filter a message that starts with a phrase for example: `^buy` wont match since the message starts with the color tag.

**To Reproduce**
1. set public chat to a custom color
2. find a spammer, recommend house portal on party world
3. filter a message thats getting spammed like `^join`
4. the chatbox message wont be censored but overhead text will

**Expected behavior**
the message in the chatbox and overhead text to be censored the same way

nlandis1995:
fixes #13885
When filtering messages color tags were not being ignored and resulting in issues when filtering using ^ or $ in the regex. Additionally it was possible mistakenly filter out a message due to the tag matching the filter. This could also make the color go away. To resolve this I added a method in ColorUtils to unwrap the color tag. Censor message now uses this function to unwrap the tag before it does the filtering and if there was a color tag found it puts it back before returning.


Adam-:
Do you have an example of a real in-game message which you are trying to filter but cannot due to this issue?

nlandis1995:
Hi I tried to link to this to the original issue but it didn't seem to work. https://github.com/runelite/runelite/issues/13885.
The example given in the original issue was trying to filter messages that start with anything using ^ in the regex. Since the message starts with the color tag that wouldn't work. The same would be true for ending tags. 

The point would be you might not want to filter a specific word but filter any spam messages that start with the word. If the messages are colored you could not do that. So if their were bots spamming "buy this" you couldn't just filter by ^buy if the message was colored.

Not sure how common it is for people to use ^ and $ in their regex filters but it seems useful against spamming bots with varying usernames.

Another rare but possible occurrence is trying to filter out a number which was in the hex code of the color. Doing that would flag any message of that color and also alter the color since it is altering the tag.

Adam-:
I think that I would rather modify the chat coloring code to work similarly to the Jagex chat coloring code, which applies the color when building the chatbox as opposed to editing the underlying message. This would fix this issue since the color would not be present when it is trying to filter. https://github.com/Adam-/runelite/commit/621f19a3eb26cbf92d01bebf8e76c325da963e02

Adam-:
Incidentally this also probably fixes #14420 ?

nlandis1995:
Yeah if you added the color after the chatFilter callback that would fix this issue. To do that you would just need to refactor a bit. If you didn't want to change the client scripts too much you would need to add the color after the chatFilter callback and add that to the chatMessage manager because the chatfilter is an optional plugin. Doing it at that point though would not fix #14420 because currently the username field is set by the output of procedure 5031 and stored to a local variable that is never overwritten afterwards. So without bigger changes to the client script you would need to do the two fixes at 2 different locations. The username coloring would need to be done right after procedure 5031 while it is still on the stack before the variable is stored, and the actual message would need to be done after chatFilter so that the message is filtered first.

Adam-:
Did you look at that commit I linked? I think it fixes both of these issues. I had Hydrox look at it and he thinks it also fixes #14420

nlandis1995:
Oh sorry missed the link. Yeah that should fix issue #14420 because the username won't be colored at the time procedure 5031 is run.

Adam-:
Okay I've pushed 90f200be531b8a899a605521f3e9840de6f9794d - am closing this since it fixes the underlying issue this is trying to fix.

PR #14420, opened 2021-11-26
fixes #13537 Chatbox behavior fix for colored usernames.

### LINKED ISSUE #13537: Recolouring usernames causes friends to not count as friends, and ignoring someone to not remove existing chat messages
If any of the username recolour options are set under `Chat Color`, the client fails in certain ways. Notably, if a username is recoloured and the user is subsequently ignored, the existing chat messages don't get removed like usual. Another obvious behaviour is if you have someone added as a friend, and either have `Public friend usernames`, `Usernames`, or `Friends chat usernames` (if they're in an FC with you), right clicking their message will give a menu as if they're not on your friends list (no `Message` option, shows `Add friend` and `Add ignore`)

**To Reproduce**
1. Set `Chat Color` -> `Usernames` to something.
2. Go to a high traffic area (eg. W302 GE)
3. Add any one of the spammers there to your ignore list
4. Add them to the ignore list again using the same message, and note that they're already on your ignore list. Without the colour, the message you used to add them the second time would have been hidden.
5. Add another random spammer to your friends list. It does not matter if you or the target have private on.
6. Note that if you right click their message again, it won't have the option to message them, and will instead show options to add them as a friend or to the ignore list.

**Expected behaviour**
Messages with recoloured usernames should act the same as messages without recoloured usernames.

**Screenshots**
This issue was originally reported by `Lazark` in the Discord, and I have confirmed all of the above situations. Below is a gif they provided of the issue.
pcnPpJg79a

 - RuneLite version: 1.7.5

nlandis1995:
fixes #13537
The bug in question made it so that when a username was colored if you were friends with the user and clicked on the message it still showed the menu as if you were not friends. Additionally if you ignored the user after the message was already in the chatbox any messages where the username was colored would not be hidden. This happens because inside the ChatBuilder client script procedure 5031 is run which determines whether a sender is a friend or not. This does not return the correct result if there is a color tag in the username. To resolve this I added a method to ColorUtil to unwrap a color tag. Before the chatbox parent script runs all color tags are removed from the message nodes and stored in a mapping. I added a callback to the ChatBuilder script for just after procedure 5031 runs. At this callback the color tag is returned to the stack variable so that the color is actually displayed. At the end of the chatbox parent script the message nodes get their color tags added back to the username if they had any. This resolves both the issue of the wrong menu showing up and the issue of ignored messages not going away.


Adam-:
fixed via 90f200be531b8a899a605521f3e9840de6f9794d

Issue #9206, opened 2019-06-25
XP Tracker Show both XP gained and XP remaining on canvas


Jbluehdorn:
Currently the on-screen tracker top line and bottom line have separate things that they can show but there are times it would be nice to have the information only currently available on one line. I'm wondering if it would be possible to mix and match those a bit more?

badookey:
I'm happy to look into this! I know it's a bit of an old issue so I understand if you're not available anymore @Jbluehdorn.

I've tried some small adjustments but I wasn't entirely sure what you meant by 'information only currently available on one line'. Let me know if this looks like what you're after. 

![Here's what i've done so far...](https://user-images.githubusercontent.com/18100566/144427484-c64c7cd6-a967-42f7-894b-5ba43705e24d.png)

PR #14437, opened 2021-12-02
Add settings to xp tracker to improve on-screen canvas display options

badookey:
This addresses #9206

Add setting for displaying the progress bar on the on-screen xp tracker,
allowing users to not display the progress bar to reduce on-screen
clutter.

Similarly, added setting to only display one line of the xp tracker
on-screen, allowing users to tweak their display to their liking.


jigalin:
Nice addition, Luke :)

Issue #14445, opened 2021-12-05
Allow players to change the fill color and opacity of Tile Indicators


dakoders:
**Is your feature request related to a problem? Please describe.**
Yes - The tile indicators currently add a dark gray shade to any tile indicator that is enabled (hover, true, destination) and can often make it really difficult to tell if a portal/grasping claws is spawning under the player in Nightmare and Phosani's Nightmare.

**Describe the solution you'd like**
I'd like to be able to select the color and opacity of the tile indicators, so I can choose to either have the tile filled with a light color (like the official client), have it just outlined and the fill transparent, or have both a colored infill and outline.

Hydrox6:
Duplicate of #4188

Adam-:
added in #14447

PR #14447, opened 2021-12-07
Added settings for tile indicator fill colors

JoshJ96:
Commit based on this issue: https://github.com/runelite/runelite/issues/14445

My first commit :) happy to work with you guys

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
@@ -29,27 +29,38 @@ import net.runelite.client.config.Alpha;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
 
 @ConfigGroup("tileindicators")
 public interface TileIndicatorsConfig extends Config
 {
-	@Alpha
-	@ConfigItem(
-		keyName = "highlightDestinationColor",
-		name = "Destination tile",
-		description = "Configures the highlight color of current destination",
+	@ConfigSection(
+		name = "Destination Tile",
+		description = "Destination tile configuration",
+		position = 0
+	)
+	String destinationTile = "destinationTile";
+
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsOverlay.java
```diff
@@ -106,6 +106,9 @@ private void renderTile(final Graphics2D graphics, final LocalPoint dest, final
 			return;
 		}
 
+		graphics.setColor(new Color(fillColor.getRed(), fillColor.getGreen(), fillColor.getBlue(), fillColor.polyAlpha()));
+		graphics.fillPolygon(poly);
```
Adam-, line 110: `renderPolygon` is already doing a `fill` with the polygon with `new Color(0, 0, 0, 50)` - you should not be drawing over it again but instead call the overloaded `renderPolygon` which accepts a fill color.

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
```diff
@@ -55,12 +55,22 @@ default boolean highlightDestinationTile()
 	{
 		return true;
 	}
+	@ConfigItem(
+		keyName = "destinationTileFillColor",
+		name = "Destination tile fill color",
+		description = "Configures the fill color of destination tile",
+		position = 3
+	)
+	default boolean destinationTileFillColor()
```
Adam-, line 64: all of these should probably have `@Alpha`

JoshJ96: Sounds good ill update and do another commit with that

JoshJ96:
All done with commits for tonight. Let me know what you think!

Adam-:
The default behavior should be the existing behavior, which is the fill color is `0,0,0,50`

JoshJ96:
Updated

JoshJ96:
> The default behavior should be the existing behavior, which is the fill color is `0,0,0,50`



> The default behavior should be the existing behavior, which is the fill color is `0,0,0,50`

Updated

PR #14464, opened 2021-12-10
emoji: Add smirking face emoji 😏

donth77:
Add smirking face (😏) to emoji plugin with trigger `:^)`

As one of the original emojis, smirking face deserves a spot in the emoji plugin. It was approved as part of Unicode 6.0 and added to Emoji 1.0. It features a half-smile, raised eyebrows, and eyes looking to the side.


raiyni:
Emojis can now be accessed with their name ie: `:smirk:`

Issue #3684, opened 2018-06-08
Player Indicators - Wild Only Toggle


TheTurtle42:
Would it be possible to add a toggle that allows the player indicator feature to only work in the wild? 

I find the feature very useful for quickly spotting pkers, but can be annoying when not in the wild. I find myself having to change the configuration every time I move between wild and non-wild.

Thank you!

DannysPVM:
Yes it is possible @TheTurtle42. Here is what I've done to the plugin so that it only shows player indicators in the wilderness. Will PR this soon.

_Demo Player Indicators - Wild Only Toggle_
 https://puu.sh/ABd9I.gif

TheTurtle42:
Awesome! Exactly what I had in mind.

Thank you, DannysPVM

DannysPVM:
No problem, glad I was able to help. :)

deathbeam:
Same as a lot of previous similar requests. You can just re-configure the plugins based on what you are doing, adding auto-config option for every niche case seems counter-productive.

picocode1:
so where is the option for this?

Issue #2686, opened 2018-05-13
Hotkeys to disable/enable plugins


Gamer1120:
I would love to be able to set hotkeys to enable/disable certain features. For example: CTRL-ALT-I to disable/enable Idle Notifier.

Qwix:
This is possible with the Plugin Presets plugin.  https://github.com/antero111/plugin-presets/wiki/General-Features#using-plugin-presets

PR #14468, opened 2021-12-12
Enable "in wilderness only" option for highlighting other players 

domien96:
Similar request in #3684 The difference here is that not every player indication will appear whether or not the player is in the wilderness.

I created the option in such a way that clan members, friends, etc still show up despite of being in the wilderness.
The wilderness check is focused just on the "highlight others".
This way, users do not have to turn off the highlight others checkbox every time they enter or leave the wilderness.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/HighlightOthersOptions.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/HighlightOthersOptions.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2019, Jordan Atwood <nightfirecat@protonmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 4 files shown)


Nightfirecat:
As mentioned in #3684, we will not be adding this feature.

domien96:
Can I get some more argumentation for the people who want this feature so they can understand why it is not allowed.

Nightfirecat:
It's not that it isn't allowed, it's that we simply don't want to add this feature.

domien96:
Why?

Adam-:
I think we didn't want to go down into the rabbit hole of having every plugin auto enable/disable itself at every possible opportunity, since there are an infinite number of possible scenarios. Even with this there are many different combinations such as if you are in the wilderness vs if they are in the wilderness, pvp worlds safe zones vs danger zones, pvp worlds wilderness vs danger zones, target worlds, etc. We actually don't know where safe zones are in pvp worlds so it isn't something we could add without mapping them individually anyway. And, then we have to maintain the safe zones....

Or, you can just toggle the plugin when you want to use it by clicking 1 button.

Nightfirecat:
For what it's worth, #2686 would be a much better and more general solution to the issue of players having plugins they'd want enabled only at certain times.

PR #14469, opened 2021-12-12
Add a flag to prevent the new strip-accents feature in Chat Filter Plugin

featurecat:
There was a new feature added 1 month ago to the chat filter that prevents any filter from being able to distinguish between accented characters (eg ë and e are treated as the same). This pull request adds the option to specify a regex flag (?U) that prevents the stripping of accented characters.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
@@ -321,7 +321,8 @@ public class ChatFilterPlugin extends Plugin
 	{
 		String strippedMessage = jagexPrintableCharMatcher.retainFrom(message)
 			.replace('\u00A0', ' ');
-		String strippedAccents = StringUtils.stripAccents(strippedMessage);
+		boolean shouldStripAccents = !strippedMessage.contains("(?U)");
+		String strippedAccents = shouldStripAccents ? StringUtils.stripAccents(strippedMessage) : strippedMessage;
 		assert strippedMessage.length() == strippedAccents.length();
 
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This is something we recognized when adding the feature, but figured there would rarely (never?) be a case where it would impact users. What is the use case here?

featurecat:
The most effective filter I had was recognizing that most bots use the accented characters, so by adding a filter such as `([áàâäãåçéèêëíìîïñóòôöõúùûüßýÿ].*){1,}` I am able to filter out the vast majority of bots (optionally, it can be increased to {2,} or higher to allow some accented characters). This worked very well until the update that prevents detecting accented characters.

featurecat:
e.g. this bot could be filtered out with my filter: 
image

Adam-:
This seems like a valid use case, though I don't want to just make up undocumented fake pattern flags. A toggle for normalizing the characters though seems fine.

runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java
```diff
@@ -321,7 +321,8 @@ String censorMessage(final String username, final String message)
 	{
 		String strippedMessage = jagexPrintableCharMatcher.retainFrom(message)
 			.replace('\u00A0', ' ');
-		String strippedAccents = StringUtils.stripAccents(strippedMessage);
+		boolean shouldStripAccents = !strippedMessage.contains("(?U)");
```
Adam-, line 324: `strippedMessage` is the message from the player and not the configured pattern, so this doesn't work

Adam-:
fixed in b5515075efa35319ede070c3c4f49dcca085c859

Issue #14471, opened 2021-12-12
Freeze timer is removed when seeding


SaintJuniper:
Please check if your issue is not a duplicate by [searching existing issues](https://github.com/runelite/runelite/search?type=Issues)

**Describe the bug**
Using mithril seeds while frozen removes your freeze timer if you're frozen

**To Reproduce**
Get frozen (barrage/zgs) in PvP
Use mithril seeds 

**Expected behavior**
Freeze timer continues to be shown and will continue to tick down because you're still frozen

PR #14475, opened 2021-12-14
Add Mithril Seed use-case to TimersPlugin.java

JoshJ96:
Addresses https://github.com/runelite/runelite/issues/14471

This is what I have so far. I assumed the best way to do this was through the onChatMessage event

I wasn't sure to use message.equals or message.contains as Mithril Seeds are a bit interesting and generate two lines when planted:
image


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -879,14 +887,16 @@ public void onGameTick(GameTick event)
 		{
 			// assume movement means unfrozen
```
Nightfirecat, line 888: This should be updated to mention the mithril seed case

Nightfirecat (changes requested):
Your branch needs rebase also

JoshJ96:
Can u just close this lol I screwed up

PR #14476, opened 2021-12-14
Fix: Freeze timer will stay active when planting a mithril seed

JoshJ96:
Addresses https://github.com/runelite/runelite/issues/14471

Just tested in-game with two accounts at Clan Wars. Looks good :)

Planting a mithril seed generates a two-line chat message as shown below, so I used that to temporarily set a boolean value to true, making the timer not disappear in the situation of movement. The value is then reset later during the game tick event

image


dereklopes:
If you have game chat set to "Filtered" or "Off" does not generate the chat messages mentioned. Does the fix still work in this scenario?

Nightfirecat:
Your chat filter does not affect chat messages firing the event. Go ingame and check for yourself to see.

dereklopes:
> Your chat filter does not affect chat messages firing the event. Go ingame and check for yourself to see.

Cool, that's good to know. Thanks for the info:)

PR #14496, opened 2021-12-18
Fix #14487 - sometimes non-custom left-click inventory item swaps override the custom option.

### LINKED ISSUE #14487: Menu entry swapper overrides manual left click "wear" settings
**Describe the bug**
Some Menu Entry Swapper settings take priority over some (but not all?) manually set left-click options. For example, "Item Swaps --> Teleport item" overrides a manually-set "Wear" option on mythical capes and max capes. It does not override "Use."

**To Reproduce**
1. Enable menu entry swapper.
2. Enable "Item Swaps --> Teleport item."
3. Right click inventory, "Configure Left-click", set a mythical cape's left-click option to "wear" and save.

**Expected behavior**
Manually set left-click options should take precedence. 

**Screenshots**
image
image

 - OS: Arch Linux (Linux Zen 5.15.6)
 - RuneLite version: 1.8.7
 - Launcher version: 2.2.0

**Additional context**
When certain Menu Swapper (or other plugin) settings are enabled, left-click options are changed from default. For example, "Wear" is the default option on the mythical cape, but with "teleport items" in Menu Entry Swapper, the default option becomes "Teleport." Attempting to change it to "Wear" manually doesn't work, presumably because "Wear" is the **actual** default setting, so it is overriden by the plugin.

geheur:
sometimes left-click item swaps that are not the custom left-click swap (e.g. teleport swap) override the custom left-click swap.

There are two parts to this fix:
1) Always return inside `swapMenuEntry` if the entry passed to it is an inventory item entry with a custom left click swap (regardless of the entry's option being the swapped option). This prevents non-custom left-click swaps from running.
2) Do not mark the default left-click option with "* " if the custom left-click swap is not set for that option. This allows the default left-click option to be set as the custom left-click option in the case that there is not currently a custom left-click swap, because otherwise, the option is called "* Wear" which is not recognized as a valid inventory action in `onMenuOptionClicked`.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -29,7 +29,6 @@ package net.runelite.client.plugins.menuentryswapper;
 import com.google.common.annotations.VisibleForTesting;
 import static com.google.common.base.Predicates.alwaysTrue;
 import static com.google.common.base.Predicates.equalTo;
-import com.google.common.base.Strings;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.LinkedHashMultimap;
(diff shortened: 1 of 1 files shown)


Adam-:
added in 406ce7bd6b2ee93ce14a1e9cd1f97da9ee501a5e / 8fa819d035b3a6d0de91396ac8e8ed2b6eaea30e / 0f52e3ff18dcece4ff0ca39148bc2d0a9fb96e55

PR #14498, opened 2021-12-18
Fix wall documentation

Skretzo:
- West and east were mixed up
- Added missing north-west, north-east, south-west, south-east


Skretzo:
Fixed in 45f744fa2b0adcc78af52b2d74eed19838d765dd

PR #14500, opened 2021-12-18
Disable fuzzy search if GE search event consumed

### LINKED ISSUE #14499: Grand Exchange Plugin's 'fuzzy search' breaks bank tags search in the GE
**Describe the bug**

Usually when you search for something like `tag:clue` in the Grand Exchange, it will show results as the items you have tagged.

If you also have the 'fuzzy fallback' or 'fuzzy only' config settings on in the Grand Exchange plugin, this will override the aforementioned behaviour and instead show fuzzy results of the text searched.

It would make more sense to me for features such as tag searching to override fuzzy search.

**To Reproduce**

* Turn on 'Grand Exchange' and 'Bank Tags' plugins
* Turn on one of the fuzzy search config settings for 'Grand Exchange'
* Try to search for one of your bank tags in the Grand Exchange, for example 'tag:clue'
* Get fuzzy results of the text instead

**Expected behavior**
You should get the tag search results instead of the fuzzy results.

**Screenshots**

Example of fuzzy results instead of tag results.

Zoinkwiz:
fixes #14499.

This allows for other plugins, such as Bank Tags, to take priority over fuzzy searching of results. Presumably most cases of the grand exchange search event being consumed will be due to another activity having explicit mutation of the results, which then would mean fuzzy search wouldn't make sense to be applied.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangePlugin.java
@@ -638,7 +638,7 @@ public class GrandExchangePlugin extends Plugin
 
 		GrandExchangeSearchMode searchMode = config.geSearchMode();
 		final String input = client.getVar(VarClientStr.INPUT_TEXT);
-		if (searchMode == GrandExchangeSearchMode.DEFAULT || input.isEmpty())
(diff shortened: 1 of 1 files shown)


Adam-:
This is wrong because it is assuming that the GE plugin subscriber runs after the bank tag plugin subscriber, which is not always the case.

Adam-:
fixed in ce56056e2f89372d69a9af0683cb4266c4ebaa5c

Issue #14505, opened 2021-12-20
Ability to change infobox text to bold


lolplayer509845:
![image](https://user-images.githubusercontent.com/96413024/146734590-e0f8a09a-fd5a-4961-b119-505038ea672f.png)
![image](https://user-images.githubusercontent.com/96413024/146734624-0d5615f8-3758-4d3a-8554-841f3829dc48.png)

Was just requesting if there can be an option added to change the infobox text to bold. Right now you can have your overlay, tooltip, and interface text set to bold, but not the infoboxes. The text for infoboxes can go the "small" size if decreased in size, but does not go past "regular" to "bold". Just wondering if this is an easy addition. Image 1 is to show current infobox text, 2nd image is the requested text for infoboxes. Thanks

Adam-:
added in 791912a7c0ea6fcac6972b6054da4469109947a5

PR #14509, opened 2021-12-21
Ability to change infobox text font

Vanillj:
This addresses https://github.com/runelite/runelite/issues/14505.

Small change so you can select betwen Small, Regular and Bold font, default is the Regular font. The infobox still acts the same and gets smaller when the infobox size gets smaller.


Vanillj:
Is the infobox size limit on the font necessary? When we can just let the user decide what font size to use.

Adam-:
added in 791912a7c0ea6fcac6972b6054da4469109947a5

PR #14511, opened 2021-12-21
screen-markers: Add marker labels

### LINKED ISSUE #8621: Add numbers to on screen markers
Add on screen numbers for screen markers. For instance when having multiple screen markers on for agility clicking, being able to identify which one is next in the sequence can be hard to remember. Being able to number each screen marker (for instance adding the number into the middle of the marker) helps the user identify which marker would be next in the sequence.

Skretzo:
- Adds a toggle to display the marker name as a label above the marker
labelled screen marker  
label toggle
- Fixes incorrect visibility tooltip text

Resolves #8621


Adam-:
Do you have any information on what "Fixes incorrect visibility tooltip text" means?

Skretzo:
When you hover over the visibility icon it will always show "Hide screen marker" when it should alternate between "Show screen marker" and "Hide screen marker"

Adam-:
Is that a bug? Can you split it into a separate commit?

Skretzo:
I consider it a bug, yes, but it has not been noticed or raised as an issue yet. I can attempt to split it into a separate commit.

Skretzo:
Please don't merge this yet. I noticed something weird: notice how the border colour is not showing up in the border colour icon in the above image despite the screen marker being green. Not sure what I did wrong?

Adam-:
added in 260c923abe10a8eda1ea51c8ef035397ababfa9a and 109ad81473be78fba1a7ca1ab49dc140f194413f

PR #14527, opened 2021-12-29
Add spawn object command

Krazune:
This adds a new dev tools command that spawns a RuneLiteObject at the player location. The object is destroyed when the scene is reloaded (like any other RuneLiteObject).

Example:
`::spawnobject 1` - This would spawn a dwarf head.

I keep using the shell to spawn objects but this would be a lot more practical.
PS: I did not check for any exception because no exception is being caught in the other switch cases.


Krazune:
I updated the code to the new behaviour of client.loadModel(). This included a timeout to stop it from trying to load an invalid model forever. The timeout can be changed in case it is considered too rudimentary.

PR #14536, opened 2021-12-31
Roof removal poh fixes

keyosk:
Related to: https://github.com/runelite/runelite/pull/14529

The first [commit](https://github.com/runelite/runelite/pull/14536/commits/3af75e6752deb963dc9da690d5afe4f442b041e7) fixes an issue where POH roof removal overrides weren't initialized on plugin startUp when configured to do so.

The second [commit](https://github.com/runelite/runelite/pull/14536/commits/99b0a18a4c4c159b5a480c5fb2a069babf33469e) moves the `configOverrideRegions.contains(regionID)` conditional out of the inner for loop, as it's parameter did not change across loop iterations.


Hydrox6 (approved):
Really annoyed that I missed this initially :d

The change to `applyRoofOverrides` probably isn't needed; the code is usually only run during scene load (and now config change) and only ever has to loop through a small amount of things, but it's not a bad change to have.

PR #14539, opened 2022-01-01
Handle CMD+Q inputs correctly on MacOS

testing-ongithub:
On MacOS, CMD+Q should result in a soft exit, which can allow for user confirmation, rather than a hard exit. Setting the "apple.eawt.quitStrategy" property to "CLOSE_ALL_WINDOWS" accomplishes this. However, the property has to be set before the _AppEventHandler is initialized and reads that property, which happens when SwingUtilities.invokeLater() is first called in SplashScreen.java.

Setting the "apple.eawt.quitStrategy" property to "CLOSE_ALL_WINDOWS" results in a windowClosing event being sent to each JFrame. If any of them handle that by exiting, the program will exit. Otherwise, the program will continue as if CMD+Q was not entered. The current SplashScreen windowClosing handling is EXIT_ON_CLOSE, so change it to DO_NOTHING_ON_CLOSE.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/SplashScreen.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/SplashScreen.java
@@ -46,6 +46,7 @@ import javax.swing.plaf.basic.BasicProgressBarUI;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.client.ui.skin.SubstanceRuneLiteLookAndFeel;
 import net.runelite.client.util.ImageUtil;
+import net.runelite.client.util.OSType;
 import org.pushingpixels.substance.internal.SubstanceSynapse;
 
 @Slf4j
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/ui/SplashScreen.java
```diff
@@ -131,6 +132,7 @@ protected Color getSelectionForeground()
 		timer.start();
 
 		setVisible(true);
+		setDefaultCloseOperation(DO_NOTHING_ON_CLOSE);
```
keyosk, line 135: Should the previous call on line 76 be updated instead? https://github.com/runelite/runelite/blob/561a28ab8c38336b5998812409979bfb91169df7/runelite-client/src/main/java/net/runelite/client/ui/SplashScreen.java#L76

PR #14550, opened 2022-01-04
rl-client: keep plugin hub jars around for longer

abextm:
if you started 2 clients with different configs at the same time one
client could delete files out from under the other while it was trying
to load them, causing them to fail to load

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
@@ -35,6 +35,8 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
```diff
@@ -167,6 +167,9 @@ private void refreshPlugins()
 				SplashScreen.init();
 			}
 
+			long now = System.currentTimeMillis();
+			long keepAfter = now - 1000 * 60 * 60 * 24 * 3;
```
Adam-, line 171: Using the java 8 time apis is preferred:

```java
				Instant.now()
					.minus(Duration.ofDays(1))
					.toEpochMilli();
```

runelite-client/src/main/java/net/runelite/client/externalplugins/ExternalPluginManager.java
```diff
@@ -189,6 +192,7 @@ private void refreshPlugins()
 					{
 						externalPlugins.add(manifest);
 
+						manifest.getJarFile().setLastModified(now);
```
Adam-, line 195: This only needs to happen when adding to `keep` right - otherwise the file is modified from the download which does a write()

abextm: doing it always makes a race less likely

Adam-: That's fine as long as `setLastModified` fails okay if the file doesn't exist.

abextm: as documented, it returns false and does not throw

PR #13993, opened 2021-08-06
Falador Shield item charges

jasmoreau:
Shows Falador Shield charges once you check or recharge and have none left.

PR #14556, opened 2022-01-06
MenuEntrySwapper Falador Shield

0-08:
Addition of the shield to fix #13993


Paveldin:
Uh, Wrong PR that linked to 13993. It's mean to add the Falador Shield in the MenuEntrySwapper plugin. It adds the "Recharge-Prayer" and "Check" options

Nightfirecat:
8944eb9 has already been added to master and released, which sufficiently covers this use case.

PR #14559, opened 2022-01-07
runenergy: Fix replaced orb text flickering with unlocked FPS

pilino1234:
The client script for updating the orbs could update the text between the latest onGameTick and the frame being drawn which could cause the text to flicker.
This is fixed by setting the run energy orb text before a frame is rendered, inspired by c56c26ff48839d297973e59f4babc9b5b91f2ad2.

Reported by @emerald000

### DIFF
--- a/cache-client/pom.xml
+++ b/cache-client/pom.xml
@@ -29,7 +29,7 @@
 	<parent>
 		<groupId>net.runelite</groupId>
 		<artifactId>runelite-parent</artifactId>
-		<version>1.8.9</version>
+		<version>1.8.10-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>cache-client</artifactId>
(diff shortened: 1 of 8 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyPlugin.java
```diff
@@ -169,7 +170,13 @@ public void onGameTick(GameTick event)
...
+	}
 
+	@Subscribe
+	public void onBeforeRender(BeforeRender beforeRender)
+	{
+		// This text is also set by the client script for updating the orbs,
+		// set the text before every frame is drawn to prevent flicker.
 		if (energyConfig.replaceOrbText())
 		{
 			setRunOrbText(getEstimatedRunTimeRemaining(true));
```
Adam-, line 182: This should probably precompute the `getEstimatedRunTimeRemaining` and just set that each frame

Adam-:
Also I don't think this is related to unlocked fps at all, since unlocked fps doesn't increase the tick rate

PR #14561, opened 2022-01-07
runenergy: Fix flickering replaced run energy orb text

pilino1234:
If a frame is rendered in between the orb text being set by the client script and the plugin's `onGameTick` replacing the text, the text will flicker.
Fixed by setting the run energy orb text before a frame is rendered, inspired by c56c26f.

Reported by emerald000
Continued from #14559 with fixed review comment:
- Only compute the remaining time once per tick since it cannot change more frequently anyway


pilino1234:
Fixed in b593a5167e2eddbe6f5fa8a0bac5f0cbb282ed5a

PR #4036, opened 2018-06-27
itemstats: Add mixed (barbarian) potions

Nightfirecat:
Closes #12057
Closes #12890

PR #14566, opened 2022-01-08
assorted nex changes

Hydrox6:
Quite a lot of things need changing with new potions, wow.

Due to Nex spending the start of the fight non-hostile, a boss respawn timer was left out; a player paying attention to the timer would already have as much time as they needed to escape from the fight if they wanted, or to prepare the correct prayer.

The Item Stats for the ancient brew was incredibly annoying, due to it being able to boost prayer 5% above max. There's probably a better way to do what I did, but I'm not sure what it is.


Adam-:
I cherry picked these without the item stats stuff since I want to look at it later

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/StatBoost.java
```diff
@@ -57,7 +70,14 @@ public StatChange effect(Client client)
 		int calcedDelta = heals(client);
 		if (boost && calcedDelta > 0)
 		{
-			max += calcedDelta;
+			if (capCalculator != null)
```
Nightfirecat, line 73: I'm not a fan of complicating this parent class due to the implementation of CappedStatBoost. Instead, I'd rather contain all of the cap calculation in the `heals()` implementation of the child class.

eg. Use this implementation for CappedStatBoost and revert all the changes to this file.
```java
@Override
{
	final int current = getStat().getValue(client);
	final int max = getStat().getMaximum(client);
	final int delta = deltaCalculator.calculateDelta(max);
	final int cap = capCalculator.calculateDelta(max);

	if (delta + current <= max + cap)
	{
		return delta;
	}

	return max + cap - current;
}```

Nightfirecat (commented):
fwiw I wrote some tests to make sure this worked like I expected, modeled off my mixed potion PR (#4036). Add them if you want, or not; just putting them here in case they're of interest:

```java
@Test
public void testAncientBrew()
{
	final Effect ancientBrew = new ItemStatChanges().get(ItemID.ANCIENT_BREW4);

	assertEquals(4, skillChange(Skill.PRAYER, 99, 99, ancientBrew));
	assertEquals(11, skillChange(Skill.PRAYER, 99, 90, ancientBrew));
	assertEquals(11, skillChange(Skill.PRAYER, 99, 0, ancientBrew));
	assertEquals(2, skillChange(Skill.PRAYER, 50, 50, ancientBrew));
	assertEquals(7, skillChange(Skill.PRAYER, 50, 40, ancientBrew));
}

private int skillChange(Skill skill, int maxValue, int currentValue, Effect effect)
{
	if (effect == null)
	{
		throw new IllegalArgumentException("Applied effect is null");
	}

	when(client.getRealSkillLevel(skill)).thenReturn(maxValue);
	when(client.getBoostedSkillLevel(skill)).thenReturn(currentValue);
	final StatsChanges statsChanges = effect.calculate(client);

	for (final StatChange statChange : statsChanges.getStatChanges())
	{
		if (!statChange.getStat().getName().equals(skill.getName()))
		{
			continue;
		}

		return statChange.getRelative();
	}

	return 0;
}```

PR #14567, opened 2022-01-08
itemcharges: Add support for blood essence

SirGirion:
I believe I've handled the relevant game messages. You can only have one active at a time so don't need any fancy per-item tracking. I did confirm that as of now if you only use one charge it uses "1" instead of "one". This does not handle the use-case of alching the essence before it runs out of charges which would reset it  however I have a feeling Jagex is going to patch that behavior anyways so probably not worth the effort of handling.


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -138,8 +145,10 @@
 	private static final int MAX_AMULET_OF_CHEMISTRY_CHARGES = 5;
 	private static final int MAX_AMULET_OF_BOUNTY_CHARGES = 10;
 	private static final int MAX_SLAYER_BRACELET_CHARGES = 30;
+	private static final int MAX_BLOOD_ESSENCE_CHARGES = 1000;
 
 	private int lastExplorerRingCharge = -1;
+	private boolean didBloodEssenceBreak = false;
```
Vanillj, line 151: You can probably check for the text "You activate the blood essence." and set the charges to 1000, instead of using this boolean variable.

Adam-:
Do you have a log of the chat messages from the essence breaking?

Vanillj:
> Do you have a log of the chat messages from the essence breaking?

I can confirm they are correct.
bcYL2fY

Adam-:
It may be good to add a test for this

PR #14592, opened 2022-01-15
Fixing NPE in slayer plugin when updating the counter.

### LINKED ISSUE #14542: Slayer plugin NPE
Found in a support user's logs, appears to be due to `addCounter()` not guaranteeing that `counter` is set to a non-null value. (stemming from slayer xp gains when the plugin doesn't know the current task, for first-time RuneLite users for instance) Relevant line of code: https://github.com/runelite/runelite/blob/0da504769df58cb2f648d2fcc8048978c7997c5b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java#L645

```
2022-01-02 14:01:45 [Client] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
	at net.runelite.client.plugins.slayer.SlayerPlugin.killed(SlayerPlugin.java:645)
	at net.runelite.client.plugins.slayer.SlayerPlugin.onStatChanged(SlayerPlugin.java:568)
	at net.runelite.client.callback.Hooks.post(Hooks.java:169)
	at client.mz(client.java:29351)
	at client.hf(client.java:7570)
	at client.gg(client.java:3685)
	at client.i(client.java:1843)
	at am.k(am.java:335)
	at am.run(am.java:314)
	at java.base/java.lang.Thread.run(Unknown Source)
```

Doxterpepper:
Fixes #14542 

In an edge case where runelite has no record of a slayer task but a
monster is killed giving slayer XP, don't update the counter because we
don't know anything about the slayer task at hand.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -647,7 +647,7 @@ public class SlayerPlugin extends Plugin
 		// save changed value
 		setProfileConfig(SlayerConfig.AMOUNT_KEY, amount);
 
-		if (!config.showInfobox())
+		if (!config.showInfobox() || Task.getTask(taskName) == null)
 		{
 			return;
 		}
--- a/runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/slayer/SlayerPluginTest.java
@@ -639,6 +639,39 @@ public class SlayerPluginTest
 	}
 
 	@Test
+	public void testXpDropWithNoKnownTask()
+	{
+		final Player player = mock(Player.class);
+		when(player.getLocalLocation()).thenReturn(new LocalPoint(0, 0));
+		when(client.getLocalPlayer()).thenReturn(player);
+		lenient().when(slayerConfig.showInfobox()).thenReturn(true);
+
+		StatChanged statChanged = new StatChanged(
+				Skill.SLAYER,
+				100,
+				2,
+				2
+		);
+
+		slayerPlugin.onStatChanged(statChanged);
+
+		slayerPlugin.setTaskName("");
+		slayerPlugin.setAmount(98);
+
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -647,7 +647,7 @@ void killed(int amt)
 		// save changed value
 		setProfileConfig(SlayerConfig.AMOUNT_KEY, amount);
 
-		if (!config.showInfobox())
+		if (!config.showInfobox() || Task.getTask(taskName) == null)
```
Adam-, line 650: This should be `Strings.isNullOrEmpty(taskName)` instead, since tracking unknown tasks is still valid and something the plugin will do. Even better would be to just nullcheck `counter` inside of `killed` though.

Doxterpepper: It's still unclear how we get to `counter.setCount(amount)` with an empty taskName while amount > 0. It seems it should be impossible to get to this line with an empty taskName but also a non zero amount since we set both through setTask.

If there is an edge case where we're not finding the slayer task name at all in the chat box for whatever reason, maybe it would be even better to just force it to something like "Unknown task" when we find taskName is null or empty? We already support unknwon task names as you said, so that would probably cover this case too.

Adam-: Oh I had not seen the `amount` check, so I agree I do not see how this can be happening then. Maybe we wait and see if someone figures out how to reproduce the issue.

Adam-: There are a couple of suspicious looking writes to amount, like for bracelet of slaughters. You might be able to 1) get a task on mobile 2) proc a slaughter bracelet to set amount = 1 3) get an xp drop - which would then trip it?

Adam-:
this is superseded by 1f6ae893d732cf3a022fe6d9662cd2a42ff2b217

PR #14596, opened 2022-01-16
LootTrackerPlugin: Fix bug with wintertodt crate count going up when not opened.

Hansel34:
Fixes issue:14586

Check that the supply crate is actually opened before updating the loot tracker plugin.


Nightfirecat:
I believe this no longer applies as of Wintertodt loot now being looted from a crate pile.

PR #14512, opened 2021-12-22
clue: Refactor Anagrams into builder model

wayne-li2:
Implementing the next set of @emerald000 changes. First, convert the Anagram clue scroll object creation into builders.

**Why Builders?**

The original method served us well for basic data about each clue scroll. However, the proposal to implement suggested teleports (and auto checking requirements for those teleports) requires a more malleable form of building each clue scroll's properties. Feel free to take a look at [the original commit](https://github.com/emerald000/runelite/commit/e9ac3a4dbd08297f976412cb976a522e83ec2934) as an example.

**Why not do them all at once?**

I would like to, except each clue type is fairly unique in its requirements. Not only that, but for each clue type, I manually (or via scripts I write myself) verify the correctness of the data. I also use the scripts to rebuild the data-set in builder format. This takes time to do, and in a world where new clue scrolls can be added any time, I believe it is prudent to merge structural changes in sooner, rather than later.

**What else has changed in this PR?**

Fix a small bug -- when objectId is -1, return an empty array instead.

We also update some area descriptions for anagram clues. These are purely opinionated, and largely come from emerald000 or myself. Feel free to suggest edits to these descriptions.

PR #14601, opened 2022-01-16
clues: Revert the location points

wayne-li2:
Unfortunately, during the various rebases and cherry-picking of https://github.com/runelite/runelite/pull/14512 I accidentally changed 5 location points to emerald000's location points. While these points are still decent, I have not vetted them in-game, so for now, I prefer reverting them.


wayne-li2:
Checked with Jordan, this is okay.

PR #14634, opened 2022-01-25
ChatCommands: Character Summary

JakeDeakin:
Change Add ChatCommands !quests (Quests Completed)  !ad (Achievements completed) !ct (Combat Tasks Completed)  !col (Collections Logged)

ChatCharacterSummary


Adam-:
Are you sure that there aren't any varps/varbits for these? Using those is preferred over reading them off of the widgets

tylerthardy:
> Are you sure that there aren't any varps/varbits for these? Using those is preferred over reading them off of the widgets

There are definitely combat task varps:
```
3116
3117
3118
3119
3120
3121
3122
3123
3124
3125
3126
3127
3128
```

JakeDeakin:
> Are you sure that there aren't any varps/varbits for these? Using those is preferred over reading them off of the widgets

I honestly have no idea. This is the first time I've done something in Runelite and widgets were just something I knew about and it worked. I will look into varbits/varps.

Adam-:
> > Are you sure that there aren't any varps/varbits for these? Using those is preferred over reading them off of the widgets
> 
> I honestly have no idea. This is the first time I've done something in Runelite and widgets were just something I knew about and it worked. I will look into varbits/varps.

They are much less finicky to use since it doesn't require parsing widgets / typically the widget text is built from them. Use the var inspector to find the ids for them, if they are easily changable

tylerthardy:
> > Are you sure that there aren't any varps/varbits for these? Using those is preferred over reading them off of the widgets
> 
> I honestly have no idea. This is the first time I've done something in Runelite and widgets were just something I knew about and it worked. I will look into varbits/varps.

I found the exact varbits you'll need - these are used for displaying the data in the panel.
image
Total = 33 + 41 + 58 + 56 +9 + 1 = 198
image


I found them by searching the text on that UI panel in the [cs2-scripts repo](https://github.com/RuneStar/cs2-scripts), and tracing through the relevant scripts to find the varbits in [4765](https://github.com/RuneStar/cs2-scripts/blob/master/scripts/%5Bproc,script4765%5D.cs2)

Some great resources:
https://github.com/runelite/runelite/wiki/VarPlayers%2C-VarBits%2C-and-VarClients
https://github.com/runelite/runelite/wiki/Working-with-client-scripts
https://github.com/RuneStar/cs2-scripts

JakeDeakin:
Thanks @tylerthardy Looks like I have a bit of reading to do to wrap my head around this to be able to find the varbits myself.

JakeDeakin:
Updated to change from using widgets to using clientscripts/varps instead.

Adam-:
You should just access the vars directly instead of pulling them off of the scriptvm stack, since that makes no sense.

JakeDeakin:
> You should just access the vars directly instead of pulling them off of the scriptvm stack, since that makes no sense.

I went that route since I keep getting an error any time I try to access the varbits from within the Lookup methods and it works. I can access them within submit, but not lookup. I can access Varplayers just fine though which I have for collections.

For the achievement diaries, I can't find any varbits for the total numbers either, just some hardcoded numbers in the scripts that the other scripts use. is there a way to access the return values of scripts if they aren't stored as varbits or varplayer?

Adam-:
They need to be accessed on the client thread, maybe that is why. Typically you can just wrap it in an invokeLater() I think.

JakeDeakin:
> They need to be accessed on the client thread, maybe that is why. Typically you can just wrap it in an invokeLater() I think.

invokeLater() is working for getting the varbit values thanks.

Is there any reason not to take the values from the scripts for Achievement Diaries where there are 48 varbits for the values that need to be added up or could I just take the output value from the script where they have done the math already? Especially for the one that's just hard coded numbers that aren't stores as varbits at all? Or am I just missing another way of getting the final return value from one of the scripts?

Adam-:
> > They need to be accessed on the client thread, maybe that is why. Typically you can just wrap it in an invokeLater() I think.
> 
> invokeLater() is working for getting the varbit values thanks.
> 
> Is there any reason not to take the values from the scripts for Achievement Diaries where there are 48 varbits for the values that need to be added up or could I just take the output value from the script where they have done the math already? Especially for the one that's just hard coded numbers that aren't stores as varbits at all? Or am I just missing another way of getting the final return value from one of the scripts?

It would be best to run the script directly via `client.runScript(3971)` and then read the return value from `intStack[0]`

Rowan-Mulder:
- The structure of Varbits.java has slightly changed over time.
- Some commands (like !col) in my experience didn't work until refreshing chat manually (chat command would either remain as "!col" or would show up as "Collections Logged: 0/0"). I'm not aware if these only occur in development.
- Some imports aren't used and are best removed.

Other than that, LGTM!

PR #14664, opened 2022-02-04
add fairy ring combination to fairy queen hideout clue step

michaeldipper-sky:
this is useful in leagues and other game modes where fairy tale pt2 might be autocompleted as the travel log won't have the combination


Hydrox6:
Just curious: why? After going to the fairy hideout, an entry is added to the bottom of the list that can be clicked to go to the next ring in the sequence, which is much easier to use than having to enter each step manually.

michaeldipper-sky:
yep just seen that it does get added when you autocomplete! will close :)

PR #14668, opened 2022-02-05
Fix topological sort Concurrent Modification Exception

LootBagger:
This PR has 2 commits. The first is a minimal change which reproduces the concurrent modification exception (CME) I mentioned in #14552 on my machine (openjdk 17 and 11). The second is a fix which prevents the CME from happening by making a new hashset in every iteration in `topologicalSort`.

Code formerly built a directed graph, topologically sorted it then reversed
the sorted list. This is mathematically equivalent to reversing the edge
direction in the directed graph. Changing the edge direction (and removing
the call to Lists.reverse) causes a CMEin the topologicalSort function

Copy pasted log tail including stacktrace:
```
2022-02-04 20:04:16 [main] DEBUG n.r.client.config.ConfigManager - Loading configuration value groundMarker.rememberTileColors: true
2022-02-04 20:04:16 [main] INFO  n.r.client.account.SessionManager - No session file exists
2022-02-04 20:04:16 [main] ERROR net.runelite.client.RuneLite - Failure during startup
java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1597)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1630)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1628)
	at com.google.common.graph.DirectedGraphConnections$2$1.computeNext(DirectedGraphConnections.java:145)
	at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:145)
	at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:140)
	at net.runelite.client.plugins.PluginManager.topologicalSort(PluginManager.java:651)
	at net.runelite.client.plugins.PluginManager.loadPlugins(PluginManager.java:345)
	at net.runelite.client.plugins.PluginManager.loadCorePlugins(PluginManager.java:275)
	at net.runelite.client.RuneLite.start(RuneLite.java:323)
	at net.runelite.client.RuneLite.main(RuneLite.java:257)
2022-02-04 20:04:17 [Client] DEBUG client-patch - Game state changed: UNKNOWN
2022-02-04 20:04:18 [Client] DEBUG client-patch - Game state changed: LOGIN_SCREEN
```

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
@@ -631,11 +631,14 @@ public class PluginManager
 	/**
 	 * Topologically sort a graph. Uses Kahn's algorithm.
 	 *
-	 * @param graph
-	 * @param <T>
-	 * @return
+	 * @param graph - A directed graph
+	 * @param <T>   - The type of the item contained in the nodes of the graph
+	 * @return - A topologically sorted list corresponding to graph.
+	 * <p>
+	 * Multiple invocations with the same arguments may return lists that are not equal.
 	 */
-	private <T> List<T> topologicalSort(Graph<T> graph)
+	@VisibleForTesting
+	static <T> List<T> topologicalSort(Graph<T> graph)
 	{
 		MutableGraph<T> graphCopy = Graphs.copyOf(graph);
 		List<T> l = new ArrayList<>();
@@ -650,7 +653,7 @@ public class PluginManager
 
 			l.add(n);
 
-			for (T m : graphCopy.successors(n))
+			for (T m : new HashSet<>(graphCopy.successors(n)))
 			{
 				graphCopy.removeEdge(n, m);
 				if (graphCopy.inDegree(m) == 0)
--- a/runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/PluginManager.java
@@ -26,7 +26,6 @@ package net.runelite.client.plugins;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
-import com.google.common.collect.Lists;
 import com.google.common.graph.Graph;
 import com.google.common.graph.GraphBuilder;
 import com.google.common.graph.Graphs;
@@ -320,8 +319,7 @@ public class PluginManager
 				continue;
 			}
 
-			Class<Plugin> pluginClass = (Class<Plugin>) clazz;
-			graph.addNode(pluginClass);
+			graph.addNode((Class<Plugin>) clazz);
 		}
 
 		// Build plugin graph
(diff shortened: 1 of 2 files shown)


Adam-:
Does this code throw an error on master under any configurations? This looks like you are both 1) introducing the CME and 2) also fixing the CME separately + also have titled the PR claiming you are fixing a CME.

LootBagger:
I have not made the code trigger a CME in master. In https://github.com/runelite/runelite/pull/14668/commits/7a60ed8e4a11b019ae857a878b7ffe736c6e7993 I made made an optimization to loadCorePlugins so instead of creating a list in reverse order then allocating a new reversed list, it creates the list in the proper order to begin with. This triggered the CME in `topologicalSort`.

The criterion for a buggy `topologicalSort` is that it throws an exception when a valid graph with a valid topological sort is passed to it. That's what happened in this case, so it is buggy. As-is in master, you are modifying a collection while you have an iterator over it. The fact that [successors](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/graph/Graph.html#successors(N)) uses an iterator is not documented, but that's why `java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1628)` is in the stack trace for the CME. Modifying a collection while iterating over it is [unspecified behavior](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--). It is only by chance that master does not trigger the CME - an update to the JDK or guava could trigger a CME. guava uses the lazy-fail `java.util.HashMap` `LinkedHashMap` and `TreeMap` but if guava were updated to use a [synchronized map](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/Collections.html#synchronizedMap(java.util.Map)) master would fail immediately with no changes to runelite code or configuration (I can test this by patching guava if desired)

Adam-:
Can you provide a graph which *on master* causes the code on master to throw a CME?

LootBagger:
How am I supposed to change the graph if I'm on master? Do you want a new test case? If I added a test I wouldn't be on master anymore.

Adam-:
A new test case would be nice, though it isn't required if you could instead provide a graph, I can build a test case. It is however very possible to make a test without modifying the `topologicalSort` method on master just by not modifying it in the commit which adds the test.

Adam-:
I've figured out this gets tripped on master by having one plugin which has multiple dependencies.

Here is a simple test case

```java
    @Test
    public void testTopologicalSort()
    {
        MutableGraph<Integer> graph = GraphBuilder
            .directed()
            .build();

        graph.addNode(1);
        graph.addNode(2);
        graph.addNode(3);

        graph.putEdge(1, 2);
        graph.putEdge(1, 3);

        PluginManager.topologicalSort(graph);
    }
```

This makes sense reversing the digraph would introduce this error since there are several plugins which depend on eg the xptracker.

LootBagger:
This PR has been updated based on advice on discord
> So what you want to do now is, make a commit to fix the cme, add the testcase too for good measure. Put in the commit message is it triggered by having a plugin with multiple deps.
Then make a 2nd commit reversing the digraph and removing the reverse(), which is just an optimization

PR #14675, opened 2022-02-07
Added a condition to check if the barrowsRewardContainer is null before trying to getItems()

HippolyteMith:
Hello I'm a student in IT and this is my first time trying to contribute to a project. I added a test to make sure that the chest is empty before geting the items.


Hc747:
This doesn't address the issue, instead, it propagates it to line 189 where you begin iterating over the items.
An appropriate fix looks like this:
```java
ItemContainer barrowsRewardContainer = client.getItemContainer(InventoryID.BARROWS_REWARD);
if (barrowsRewardContainer == null)
{
    return;
}
Item[] items = barrowsRewardContainer.getItems();
```

HippolyteMith:
Thanks for the help, didnt saw the bug :)

PR #14679, opened 2022-02-08
client: fix mouse tooltip below canvas height

roflmuffin:
When a tall tooltip is rendered using the "Under cursor" tooltip position, the bottom of the window is cut-off. This change removes the `UNDER_OFFSET` value from the canvas height calculation to ensure it is always displayed.

Before:

 

After:


Alexsuperfly:
Are you sure you just don't have your mouse tooltips incorrectly snapped to an overlay position.  We have a [troubleshooting section](https://github.com/runelite/runelite/wiki/Troubleshooting-problems-with-the-client#misaligned-tooltips) for fixing that, even though it was made unsnappable a long time ago, if it was snapped before the fix that still persists and needs to be manually corrected.

roflmuffin:
That has fixed it, thanks! I think I googled all different variations of "runelite mouse tooltips position" and couldn't find anything :(

PR #14680, opened 2022-02-08
GPU Plugin: Refactor count_prio_offset method to reduce branches

### LINKED ISSUE #12952: GPU plugin's 'Compute Shaders' option is causing artifacts
This issue seems to be related/a duplicate of #11351 but that issue has been closed.

**Describe the bug**
Depending on the complexity of the environment, vertices will start to disappear/stretch. I was able to get a few screenshots (added below). I was also able to test and confirm that this issue only occurs while the "Compute Shaders" option is enabled in the GPU plugin. In most situations this issue seems very minor and uncommon but when it does occur it's very noticeable.

**Screenshots**
Stretching vertices 1
Stretching vertices 2
Stretching vertices 3

**Environment**
 - OS: Windows 10 (Version 20H2 - build 19042.685)
 - RuneLite version: 1.6.35
 - Launcher version: 2.1.5 (Using the standalone .jar file and using Java 11.0.9)
 - GPU: Asus TUF RTX 3080 (They exist!)
 - Nvidia Driver Version: 460.89 (Did a fresh driver install.)

If there's anything I can do to help better diagnose this problem let me know. Thank you.

keyosk:
Fixes #12952

Ported from 117's HD Plugin: https://github.com/RS117/RLHD/pull/212

There's some undefined behavior going on. We observed that Nvidia Turing and Ampere cards were not making it through all the switch case statements, and the return wasn't being hit. By reducing the branches and changing this to a for loop, issues were resolved in the HD Plugin. We believe that this patch may be relevant here as well.

Intellij's complaint about the original code in part lead us to this resolution:

missing_return_statement

Edit: Some additional fun reading material about undefined/quirky behavior with conditional branching/switch statements in opengl shaders:

- https://www.peterstefek.me/shader-branch.html
- https://stackoverflow.com/a/34208113

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/priority_render.glsl
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/gpu/priority_render.glsl
@@ -70,44 +70,14 @@ int priority_map(int p, int distance, int _min10, int avg1, int avg2, int avg3) 
 // the given adjusted priority
 int count_prio_offset(int priority) {
   int total = 0;
-  switch (priority) {
-    case 17:
-      total += totalMappedNum[16];
-    case 16:
-      total += totalMappedNum[15];
-    case 15:
-      total += totalMappedNum[14];
-    case 14:
-      total += totalMappedNum[13];
-    case 13:
-      total += totalMappedNum[12];
-    case 12:
-      total += totalMappedNum[11];
-    case 11:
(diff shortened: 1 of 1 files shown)


keyosk:
Closing in favor of https://github.com/Adam-/runelite/commit/bc5beb5304eb0b911538d558dd433bf09cd04b9c

PR #14686, opened 2022-02-11
MenuEntrySwapperPlugin: Added Trade swap for MenuEntries with a non-conformed "Talk-to" option.

### LINKED ISSUE #14591: Menu Entry Swapper NPC Addition Request: Thora the Barkeep in Rellekka
This is not a problem or an issue, just a small request to add another NPC into the Menu Entry Swapper list. 

Solution: adding "Thora the Barkeep" in the Rellekka Longhall Bar to the list of NPCs that allow the "Trade" option to replace the "Talk-To" option. 

This is the best location in RuneScape to buy and deposit Beer in large quantities, and having this feature implemented would be a nice QOL enhancement. Attached image is used for reference for the location of the NPC in question. 

unknown

ProjectileRage:
Some NPCs have a non-conformed "Talk-to" option. Therefore by adding a swap for the non-conformant option, "Talk to", NPCs such as "Thora the Barkeep" will now have a proper Trade menu entry swap for the left-click position.

Resolved #14591


Nightfirecat:
Adam looked through the cache and found there are apparently exactly 9 NPCs ingame with this non-standard option and asked a JMod to check on these cases. In the event this isn't fixed ingame, we could add this.

PR #14691, opened 2022-02-13
Notifications: Set app name when using notify-send on Linux

jplsek:
If a notification is sent while the screen is locked in GNOME, GNOME shows "notify-send" as the app name instead of "RuneLite". Usually a notification will have the title/description, but since it shows in the lock screen, it instead shows the app name for privacy reasons. I assume this may be also the case for other desktop environments / display managers.

To reproduce: start runeline, go to dev tools, press Notification, then immediately lock the screen and wait a few seconds. When looking at the lock screen, it says "notify-send" instead of "RuneLite".

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -314,6 +314,8 @@ public class Notifier
 		commands.add("notify-send");
 		commands.add(title);
 		commands.add(message);
+		commands.add("-a");
+		commands.add(SHELL_ESCAPE.escape(appName));
 		commands.add("-i");
 		commands.add(SHELL_ESCAPE.escape(notifyIconPath.toAbsolutePath().toString()));
 		commands.add("-u");


Adam-:
Is this option commonly available? The latest ubuntu docs do not show it http://manpages.ubuntu.com/manpages/jammy/man1/notify-send.1.html

jplsek:
It shows when running `notify-send --help` and it's been in the code for at least [11 years](https://github.com/GNOME/libnotify/blame/a228f1e84d8a1ebe2a2cc10d19900af719ed44dc/tools/notify-send.c). I'm not sure why they don't have it in their man page.

runelite-client/src/main/java/net/runelite/client/Notifier.java
```diff
@@ -314,6 +314,8 @@ private void sendLinuxNotification(
 		commands.add("notify-send");
 		commands.add(title);
 		commands.add(message);
+		commands.add("-a");
+		commands.add(appName);
```
Adam-, line 318: this needs to be `SHELL_ESCAPE.escape(appName)`

Issue #14314, opened 2021-10-25
Hallowed sack included in Loot Tracker (Hallowed Sepulchre)


Denitsz:
Would love to see Hallowed sacks to be tracked by Loot Tracker as they aren't tracked for now.
(Hallowed sacks bought from the Mysterious Hallowed Goods vendor)
https://oldschool.runescape.wiki/w/Hallowed_sack

PR #14692, opened 2022-02-13
track hallowed sack loot in the loot tracker

Christos-Apostolidis:
Addresses #14314


Adam-:
Do these drop items on the ground under you if you don't have the inventory space to open them?

PR #14715, opened 2022-02-25
Change dev clue command to use a chatbox panel

Cyborger1:
Using a chatbox panel allows for pasting text and loading clues that are too long. In case anyone tries to use it the old way, text will also be kept if entered in the chatbox as command arguments.


Felanbird:
@Cyborger1 Do you mind rebasing this PR, the test file already has this mock/bind now so you just need to drop that and merge upstream into this. I'd like to add this to NFC's TODO pile so I don't have to keep cherry-picking it into my branches 😁

Cyborger1:
@Felanbird done

Felanbird (approved):

Cyborger1:
Not sure why the build is failing though, did I do something wrong?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -191,6 +191,9 @@ public class ClueScrollPlugin extends Plugin
 	@Inject
 	private OverlayManager overlayManager;
 
+	@Inject
+	private ChatboxPanelManager chatboxPanelManager;
+
```
Felanbird, line 196:
```suggestion
```
Looks like this was already added also, I guess I dealt with this conflict too and forgot

PR #14724, opened 2022-03-01
GPU: Prevent skybox from showing while logging in on resizable mode

testing-ongithub:
When playing on resizable mode with both the GPU plugin and Skybox plugin, the skybox can show for a few frames while logging in before the welcome screen appears (see picture below).

This PR stops the GPU plugin from using the selected skybox color until after the first game tick in which the user is logged in, which seems solve the issue.

This PR modifies the GPU plugin rather than the Skybox plugin because the issue doesn't occur without the GPU plugin on. However, if it would be more appropriate, a similar change works in the Skybox plugin too (specifically: `onBeforeRender` in the Skybox plugin could be made to return early when the `hasLoggedIn` variable is false).

skybox-login


Nightfirecat:
If I'm understanding this correctly, the issue is that the skybox plugin is setting the color too early (in `onBeforeRender` prior to the first logged in game tick firing), so this should be fixed there. The GPU shouldn't be managing any state to get the "right" skybox color, we should be providing the right skybox color to begin with.

testing-ongithub:
That makes sense. I've updated the PR to change the Skybox plugin instead of the GPU plugin.

PR #14735, opened 2022-03-08
adding identification of rejuvunation potion

fayccal:
Hi, i added the Item identification of the rejuvenation potion in the plugin.
Can you tell me if it is okay this way?


Adam-:
The long name is too long and it overlaps: 
image

I changed it to just `Rejuv` which looks better

image

fayccal:
Thanks you for the changes, i agree that it is better this way.

Issue #14753, opened 2022-03-16
Cannon Plugin Broken - Cannon Shattered Ornament Kit


jenkins918:
With the new Shattered Relics Cannon ornament kit release the existing cannon runelite plugin no longer works. Ammo is not tracked properly nor is double hit spots
![image](https://user-images.githubusercontent.com/30446221/158659012-6a54d33a-7c77-4f5d-b0dc-45824cce3037.png)

Hydrox6:
Don't need to be assigned; just make a branch on your fork and have a go!
(also, can't assign to people outside of the org anyway)

PR #14759, opened 2022-03-18
Add cannon ornament to fix cannon plugin

DavidLuong98:
Cannon ornament uses a different ID and the cannon plugin did not pick up the ornament kit. I added support for that.

See issue : #14753


Adam-:
I took #14758 instead because it fixes the tracking of the cannonballs too.

Issue #14762, opened 2022-03-19
Item Identification: Add Label for Desert Treasure Diamonds


corrigami:
Is your feature request related to a problem? Please describe.
As a colour-blind player, I find that distinguishing the different diamonds was a bit challenging.

Describe the solution you'd like
Add in a category in the Item Identification plugin for the diamonds.

Describe alternatives you've considered
Not much I can think of, as I struggle with it regardless :( I guess, you don't need to differentiate them for very long during the quest, but I just found it annoying and time consuming having to try them all on every obilisk.

PR #14764, opened 2022-03-19
Add Item id for Desert Treasure diamonds

DavidLuong98:
See issue #14762 

Adding labels for the 4 diamonds.


runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
```diff
@@ -254,6 +254,10 @@
 	DRAGONSTONE(Type.GEM, "Dragon", "DR", ItemID.UNCUT_DRAGONSTONE, ItemID.DRAGONSTONE),
 	ONYX(Type.GEM, "Onyx", "ON", ItemID.UNCUT_ONYX, ItemID.ONYX),
 	ZENYTE(Type.GEM, "Zenyte", "Z", ItemID.UNCUT_ZENYTE, ItemID.ZENYTE),
+	SHADOW_DIAMOND(Type.GEM, "Shadow Diamond", "SHD", ItemID.SHADOW_DIAMOND),
+	BLOOD_DIAMOND(Type.GEM, "Blood Diamond", "BD", ItemID.BLOOD_DIAMOND),
+	ICE_DIAMOND(Type.GEM, "Ice Diamond", "SD", ItemID.ICE_DIAMOND),
```
Skretzo, line 259: Why is the short name `"SD"` for the ice diamond?

DavidLuong98: I had a typo. thanks for catching that. I pushed a fix

Adam-:
These long names you've chosen are probably too long and would overlap the adjacent items. See eg https://github.com/runelite/runelite/pull/14735#issuecomment-1062540681. Test it out.

Nightfirecat:
Why not mark them with the type of diamond they are? I feel like Shadow/Blood/Ice/Smoke would be much more readable and understandable than ShDia/BlDia/IceDia/SmDia

DavidLuong98:
> Why not mark them with the type of diamond they are? I feel like Shadow/Blood/Ice/Smoke would be much more readable and understandable than ShDia/BlDia/IceDia/SmDia

@Nightfirecat I'll change it. At first I thought that since the person is colorblind, it would be useful to explicitly mention that these are diamonds, since other gems have the same shape. Though, no other game contents are Shadow/Blood/Ice/Smoke gems, so what you are saying makes sense.

Issue #14744, opened 2022-03-10
Clue stash build helper


NavierStrks:
Before you have a stash unit built, the clue helper will obviously tell you that when you have that step. However, if you need/want to build the unit during the clue step, as I do, I often have to look up the Construction level requirement and the items required (tier of plank, number of nails (10), hammer and saw (and gold leaf)), specifically the con level and plank tier.

The helper could include those things (Con level and items) required to build the stash.

Thank you!

PR #14772, opened 2022-03-20
Add STASH Unit build requirements to clue scroll overlay

DavidLekei:
Suggested by #14744


CFitzsimons:
My 2c would be that this should probably be a menu option and not enabled by default.  Thoughts?

DavidLekei:
Added a menu option to enable displaying the build requirements (default is disabled). Also fixed multiple of any type of item requirement being counted correctly (ie building requires 10 of any type of nails, but quantity wasn't being checked properly)

PR #14775, opened 2022-03-21
i add the elder potion

EragonRD:
Hello, i add the Elder Potion


Nightfirecat:
Why is only the elder potion added and none of the other Chambers of Xeric potions? Is this actually useful? Chambers potions are largely very distinct colors. The only one I could see confusing this one with is revitalisation, and that's an unlikely scenario given that elders are rarely used in favor of overloads.

EragonRD:
Tkx for the comment, i can add them too very quickly.

There is no problem if you want to talk with me somewhere else.

Nightfirecat:
Added in 1fbd2461889b4666ec90c16760cdc7ddf27cd760

PR #14777, opened 2022-03-22
Add abyssal whip (or) to count as abbysal whip in the emote clue overlay

emielv:
Add abyssal whip (or) to count as abbysal whip in the emote clue overlay


Adam-:
This should probably also be added to the master sherlock step.

emielv:
Added in other commit

PR #14806, opened 2022-03-31
Item Charge: fix binding necklace degrading in gotr

majiru:
When using a binding necklace in crafting combination runes with the Guardians of the Rift minigame, the chat message given is always "You partially succeed to bind..". This modifies the regex to match these messages as well. Since this is also the message sent when the player does not have the necklace equipped, there is an added check to ensure the player has one equipped.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
@@ -83,7 +83,7 @@ public class ItemChargePlugin extends Plugin
 	private static final Pattern BINDING_CHECK_PATTERN = Pattern.compile(
 		"You have ([0-9]+|one) charges? left before your Binding necklace disintegrates\\.");
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -83,7 +83,7 @@
 	private static final Pattern BINDING_CHECK_PATTERN = Pattern.compile(
 		"You have ([0-9]+|one) charges? left before your Binding necklace disintegrates\\.");
 	private static final Pattern BINDING_USED_PATTERN = Pattern.compile(
-		"You bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
+		"You (partially succeed to )*bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
```
Adam-, line 86:
```suggestion
		"You (partially succeed to )?bind the temple's power into (mud|lava|steam|dust|smoke|mist) runes\\.");
```

Adam- (commented):
I think this is probably a Jagex bug, but we can fix it anyway

PR #14819, opened 2022-04-01
Inventory Viewer: Hide inventory viewer when bank interface opens

EliasLahham:
Fixes issue where inventory viewer still shows on the bank interface being open if the configurable setting "Hidden on inventory tab" is checked

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/inventoryviewer/InventoryViewerOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/inventoryviewer/InventoryViewerOverlay.java
@@ -36,6 +36,7 @@ import net.runelite.api.Item;
 import net.runelite.api.ItemComposition;
 import net.runelite.api.ItemContainer;
 import net.runelite.api.VarClientInt;
+import net.runelite.api.widgets.WidgetInfo;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/inventoryviewer/InventoryViewerOverlay.java
```diff
@@ -75,7 +76,8 @@ public Dimension render(Graphics2D graphics)
 			return null;
 		}
 
-		if (client.getVar(VarClientInt.INVENTORY_TAB) == 3 && config.hideIfInventoryActive())
+		if ((client.getVar(VarClientInt.INVENTORY_TAB) == 3 || client.getWidget(WidgetInfo.BANK_CONTAINER) != null)
```
Nightfirecat, line 79: Does this work both in fixed and resizable viewports after you close the bank? The bank widget, in some cases, is hidden rather than nulled after being closed--whether temporarily (until the next game tick) or for longer. This should ensure that is checked.

EliasLahham: I've checked both resizeable, full screen, and fixed and this fix works for all of them

Issue #14814, opened 2022-03-31
Bank plugin ability to search for quantities


raniem:
**Is your feature request related to a problem? Please describe.**
Currently Bank plugin supports searching for items with user defined values for GE or HA prices. In a messy bank there is no way to see only placeholders (items with stacks of 0 in bank) to release unneeded ones (releasing all could potentially mess the bank even further). 

**Describe the solution you'd like**
I would like a feature in which user could search for items in stacks of defined size. E.g. `q = 0`, `q > 1` and so on with same supported patterns as `ge` and `ha` do. 

**Describe alternatives you've considered**
None other than the Vanilla `Release placeholders`-button. 

**Additional context**
This is just something I figured to maybe be useful for some players as I watched my friends bank fill with some random placeholders here and there, but releasing all would have made their bank even messier than it already is.

EliasLahham:
Currently in progress, will update this issue with a PR sometime tonight

pwatts6060:
Nice, someone in my clan asked for this today as as well. Mainly for find placeholders with q=0

raniem:
Hey! Are there any updates on this?

PR #14820, opened 2022-04-02
Bank Search: Quantity search query

EliasLahham:
Closes: #14814

Added bank search query for item quantities.

Note: Treating placeholders as quantity 0 by checking placeholderTemplateId

Examples:

image

image

4/26 edit: Now using "qty" as the search term rather than "q"

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
@@ -82,7 +82,7 @@ public class BankPlugin extends Plugin
 	private static final String SEED_VAULT_TITLE = "Seed Vault";
 
 	private static final String NUMBER_REGEX = "[0-9]+(\\.[0-9]+)?[kmb]?";
-	private static final Pattern VALUE_SEARCH_PATTERN = Pattern.compile("^(?<mode>ge|ha|alch)?" +
+	private static final Pattern VALUE_SEARCH_PATTERN = Pattern.compile("^(?<mode>qty|ge|ha|alch)?" +
 		" *(((?<op>[<>=]|>=|<=) *(?<num>" + NUMBER_REGEX + "))|" +
 		"((?<num1>" + NUMBER_REGEX + ") *- *(?<num2>" + NUMBER_REGEX + ")))$", Pattern.CASE_INSENSITIVE);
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -413,13 +413,21 @@ boolean valueSearch(final int itemId, final String str)
 		final int qty = itemQuantities.count(itemId);
 		final long gePrice = (long) itemManager.getItemPrice(itemId) * qty;
 		final long haPrice = (long) itemComposition.getHaPrice() * qty;
+		final boolean isPlaceholder = itemComposition.getPlaceholderTemplateId() == 14401;
```
Adam-, line 416: use `NullItemID.NULL_14401`

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -278,7 +278,7 @@ else if (event.getScriptId() == ScriptID.BANKMAIN_SEARCH_REFRESH)
 			// vanilla only lays out the bank every 40 client ticks, so if the search input has changed,
 			// and the bank wasn't laid out this tick, lay it out early
 			final String inputText = client.getVar(VarClientStr.INPUT_TEXT);
-			if (searchString != inputText && client.getGameCycle() % 40 != 0)
+			if (!searchString.equals(inputText) && client.getGameCycle() % 40 != 0)
```
Adam-, line 281: I don't think this change is necessary? Is it required for this PR?

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -413,13 +413,21 @@ boolean valueSearch(final int itemId, final String str)
...
 		final long haPrice = (long) itemComposition.getHaPrice() * qty;
+		final boolean isPlaceholder = itemComposition.getPlaceholderTemplateId() == 14401;
 
 		long value = Math.max(gePrice, haPrice);
 
 		final String mode = matcher.group("mode");
 		if (mode != null)
 		{
-			value = mode.toLowerCase().equals("ge") ? gePrice : haPrice;
+			if (mode.equalsIgnoreCase("q"))
```
Adam-, line 423: this might be better as a `switch (mode.toLowerCase())`

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
EliasLahham: I think this was a change intellij requested before I committed. I can change it back

Adam-: This is because intellij doesn't like using reference comparisons when comparing strings, but in this case it is fine.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
Adam-: or even better, just check `!= -1` I think like how ItemManager does.

EliasLahham: Ok, this change + the others have been made.

PR #14823, opened 2022-04-02
bank: add regex search

majiru:
I find myself reaching for regex quite often when searching around in my bank. It's quite nice to have '/^rune/' to find alchs and '/rune$/' to find magic runes.
This adds regex searching through the a '/' prefix and optional '/' suffix.


majiru:
Functionality for this is included in the linked external plugin.

PR #14825, opened 2022-04-02
emojis: add moyai

masonhobbs:
Added the moyai emoji just because I felt like he needed to be in. Triggered via "B|" since its kinda similar to expressionless.

Here's how it looks in-game:
image


masonhobbs:
add moyai 🗿

masonhobbs:
add moyai 🗿

PacosMartini:
add moyai 🗿

masonhobbs:
mark as spam but no hiding from moyai 🗿

masonhobbs:
for real though whats up with the delay this is a simple glance and approve PR

Nightfirecat:
Because we haven't discussed whether we want to add this emoji, simple as that. It's not about whether it's a big change, it's about whether we want this change (which again, we haven't discussed/decided)

masonhobbs:
I don't know what there is to discuss, I mean ye got eggplant in there, what kind of questions come with including this? This was a fun side-project to get a hang of modifying RL source but if its this much red tape, that's a serious headache.

masonhobbs:
one line PR with no updates - exactly my point. Can't even imagine how much trouble it must be to contribute for a complicated PR that requires updates. Dumb.

masonhobbs:
just dont ask for moyai

Issue #14813, opened 2022-03-31
[Menu Entry Swapper] Suggestion to add "Group Previous" as a UI swap option for shift+click on home teleport


ThickPeep:
Would it be possible to have the menu entry swapper have a drop down option for home teleport which would let you choose a different shift+left click option instead of only "outside"? I know this is a bit niche since it only applies to group Irons, but it is a feature that group irons can make a lot of use of, especially when bossing.

EliasLahham:
Looking into

PR #14827, opened 2022-04-03
Menu Entry Swapper: Add configurable house teleport swap options

EliasLahham:
Closes: #14813

Added new picklist in the UI section of the Menu Entry Swapper plugin to control what destination 'teleport to house' should take.

This is needed because Jagex added new right-click teleport options for group ironmen.

The new picklist of options can only be utilized if the 'shift-click teleport spells' checkbox is also checked*

GIM teleport example:
shiftClickGIMtele

Non-GIM teleport example:
nonGimExample

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.menuentryswapper;
 
+import lombok.RequiredArgsConstructor;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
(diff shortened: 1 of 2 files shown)


Adam-:
I do not have a high level gim, but where in the game settings is the ability to set "Group: Choose" or "Group: Previous" as the default left-click option?

EliasLahham:
@Adam- My GIM isn't high level either, but I don't see any in game settings for house teleport. All I see are teleport warning for wilderness teleports and the house teleport inside/outside setting

Adam-:
Why are you doing this double-way swap then? The existing ones are there because there is only one yes/no swap setting and the spells can be set either way in game.

Also I don't know what this home portal swap is doing.

EliasLahham:
I'm a little confused. This change is just so GIM have the option for their 2 unique teleport options which is why I defaulted the new select list to outside since everyone has outside.

If you can point me in the direction of a setting in game that touches the house teleport I'm all ears, because I don't see one other than manually right clicking the house tele

Adam-:
It's a setting in the house settings https://oldschool.runescape.wiki/w/Player-owned_house#Building_a_house

EliasLahham:
Those settings are the same for GIM/reg players. So there's no in game way of customizing that home teleport other than defaulting to inside your own house, if I'm understand where you're coming from.

Adam-:
The existing behavior selects the non-left click option when you hold shift down, which is what this `swapTeleport` method is doing - it adds a swap for both ways (eg `teleport to house` -> `outside` and `outside` -> `teleport to house`) to achieve this. This is because the left click option on the teleport is configurable, so we have to add both swaps. 

You are adding a swap both ways for the GIM teleports too, and I am trying to figure out why. If there is no way to change the left click to the new GIM teleport then I think there are no way what you are doing makes sense.

EliasLahham:
I'll look into a better way of doing it later today, a little busy with work at the moment.

EliasLahham:
I'm starting to get the issue. I just tested what I currently have with that in-game setting set to 'outside' by default and my code doesn't work.

Maybe we add support for another keybind to activate/overwrite the casting option? All the other teleport options have a 'configure option' with the house one only being configurable for inside/outside for both gim players and normal.

tl;dr, not really sure how to go about it now

MichaelLavin778:
@EliasLahham I put out a pull request to your repo with some chagnes that I think should fix your issue (test me jsut in case)

MichaelLavin778:
> @EliasLahham I put out a pull request to your repo with some changes that I think should fix your issue (test me just in case)

hold up, I just realized that my fork is newer that yours, one moment...

EliasLahham:
> Why are you doing this double-way swap then? The existing ones are there because there is only one yes/no swap setting and the spells can be set either way in game.
> 
> Also I don't know what this home portal swap is doing.

Some changes were made to this code by MichaelLavin778 which makes this all a more robust now.

Also idk what that home portal swap code was doing in there, I think I brought it over from an older build, it's gone now.

PR #14831, opened 2022-04-04
item stats: Add missing consumables

emerald000:


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -169,21 +193,58 @@ private void init()
...
+		add(combo(5, food(6), heal(PRAYER, perc(0.1, 2)), boost(MAGIC, perc(0.05, 2)), heal(ATTACK, perc(0.1, -2)), heal(STRENGTH, perc(0.1, -2)), heal(DEFENCE, perc(0.1, -2))), ANCIENT_MIX1, ANCIENT_MIX2);
 
 		// Regular overload (NMZ)
 		add(combo(5, boost(ATTACK, perc(.15, 5)), boost(STRENGTH, perc(.15, 5)), boost(DEFENCE, perc(.15, 5)), boost(RANGED, perc(.15, 5)), boost(MAGIC, perc(.15, 5)), heal(HITPOINTS, -50)), OVERLOAD_1, OVERLOAD_2, OVERLOAD_3, OVERLOAD_4);
 
 		// Bandages (Castle Wars)
 		add(new CastleWarsBandage(), BANDAGES);
 
+		// Bandages (Theatre of Blood entry mode)
+		add(combo(7, food(20), heal(PRAYER, perc(0.25, 5)), heal(RUN_ENERGY, 20), boost(ATTACK, perc(0.15, 4)), boost(STRENGTH, perc(0.15, 4)), boost(DEFENCE, perc(0.15, 4)), boost(RANGED, perc(0.1, 4)), boost(MAGIC, 4)), BANDAGES_25730);
```
Nightfirecat, line 236: Do you have confirmation that the prayer healing is 25% + 5? The wiki has its value (unconfirmed) as 29, which would be accurate for a player with 99 prayer.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -228,11 +289,16 @@ private void init()
 		add(combo(2, heal(HITPOINTS, 11), boost(AGILITY, 5), heal(RUN_ENERGY, 10)), SUMMER_PIE, HALF_A_SUMMER_PIE);
 		add(combo(2, heal(HITPOINTS, 10), boost(FLETCHING, 4)), DRAGONFRUIT_PIE, HALF_A_DRAGONFRUIT_PIE);
 
+		// Teas
+		add(combo(food(3), boost(ATTACK, perc(.02, 2))), CUP_OF_TEA_1978 /* Standard tea */);
```
Nightfirecat, line 293: The attack boost is a static 2 boost according to the wiki, but I can see ingame that this is not accurate from testing ingame and getting a +3 boost. Do you have confirmation that 2% + 2 is the formula?

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -228,11 +289,16 @@ private void init()
...
+		add(combo(food(3), heal(RUN_ENERGY, 5)), NETTLE_TEA, NETTLE_TEA_4240 /* Milky */, CUP_OF_TEA_4242 /* Nettle */, CUP_OF_TEA_4243 /* Milky nettle */);
```
Nightfirecat, line 294: The wiki claims this run energy heal only happens if you're not at full health when drinking the tea. Is that accurate?

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -110,12 +122,16 @@ private void init()
...
 		add(combo(3, food(7), new SimpleStatBoost(STRENGTH, true, perc(.05, 2)), new BoostedStatBoost(ATTACK, false, perc(.02, -3))), PREMADE_BLURB_SP, BLURBERRY_SPECIAL);
 		add(combo(3, food(5), new SimpleStatBoost(STRENGTH, true, perc(.05, 2)), new BoostedStatBoost(ATTACK, false, perc(.02, -3))), PREMADE_DR_DRAGON, DRUNK_DRAGON, PREMADE_CHOC_SDY, CHOC_SATURDAY);
 		add(combo(3, food(5), new SimpleStatBoost(STRENGTH, true, perc(.06, 1)), new BoostedStatBoost(ATTACK, false, perc(.02, -3))), PREMADE_WIZ_BLZD, WIZARD_BLIZZARD);
 		add(combo(3, food(3), new SimpleStatBoost(STRENGTH, true, perc(.04, 1)), new BoostedStatBoost(ATTACK, false, perc(.05, -3))), GROG);
 		add(combo(3, food(1), boost(STRENGTH, perc(.02, 1)), new BoostedStatBoost(ATTACK, false, perc(.06, -1))), BEER, BEER_7740);
+		add(combo(2, food(4), boost(STRENGTH, perc(.04, 2)), heal(ATTACK, perc(.1, -2))), BEER_TANKARD);
+		add(combo(2, food(15), boost(STRENGTH, perc(.1, 2)), heal(ATTACK, perc(.5, -4))), KEG_OF_BEER_3801 /* Non-quest version */);
```
Nightfirecat, line 134: These impact 3 skills, not 2, so the `primaries` argument needs to be changed. (_hitpoints_, strength, and attack) Additionally, the skill drains are based on your boosted level, not your maximum level, so the attack drains should be `BoostedStatBoost`s instead of `heal`s. (see: https://oldschool.runescape.wiki/w/Temporary_skill_drain)

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -136,6 +152,10 @@ private void init()
...
 		add(combo(5, food(1), new SimpleStatBoost(MAGIC, true, perc(.02, 2)), new BoostedStatBoost(ATTACK, false, perc(.05, -1)), new BoostedStatBoost(DEFENCE, false, perc(.05, -1)), new BoostedStatBoost(STRENGTH, false, perc(.05, -1))), WIZARDS_MIND_BOMB, MIND_BOMB1, MIND_BOMB2, MIND_BOMB3, MIND_BOMB4);
 		add(combo(5, food(2), new SimpleStatBoost(MAGIC, true, perc(.02, 3)), new BoostedStatBoost(ATTACK, false, perc(.05, -2)), new BoostedStatBoost(DEFENCE, false, perc(.05, -2)), new BoostedStatBoost(STRENGTH, false, perc(.05, -2))), MATURE_WMB, MIND_BOMBM1, MIND_BOMBM2, MIND_BOMBM3, MIND_BOMBM4);
+		add(combo(9, food(14), boost(STRENGTH, 3), boost(MINING, 1), heal(PRAYER, perc(.06, -1)), heal(AGILITY, perc(.09, -3)), heal(ATTACK, perc(.06, -1)), heal(DEFENCE, perc(.08, -2)), heal(HERBLORE, perc(.06, -1)), heal(MAGIC, perc(.05, -1)), heal(RANGED, perc(.06, -1))), BRAINDEATH_RUM);
+		add(combo(food(2), heal(PRAYER, perc(.04, -2))), BLOODY_BRACER);
+		add(combo(2, food(1), boost(AGILITY, 1), heal(STRENGTH, -1)), ELVEN_DAWN);
+		add(combo(3, boost(RANGED, 4), heal(STRENGTH, perc(.04, -2)), heal(MAGIC, perc(.04, -2))), LIZARDKICKER);
```
Nightfirecat, line 158: Likewise, make sure you're remembering the hp heal when calculating the `primaries` value, and use `BoostedStatBoost` for percentage skill drains here.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -169,21 +193,58 @@ private void init()
...
+		add(combo(food(6), boost(FISHING, 3)), FISHING_MIX1, FISHING_MIX2);
+		add(combo(food(6), boost(RUN_ENERGY, 20)), SUPER_ENERGY_MIX1, SUPER_ENERGY_MIX2);
+		add(combo(food(6), boost(HUNTER, 3)), HUNTING_MIX1, HUNTING_MIX2);
+		add(combo(food(6), boost(STRENGTH, perc(.15, 5))), SUPER_STR_MIX1, SUPER_STR_MIX2);
+		add(combo(food(6), boost(MAGIC, 3)), MAGIC_ESSENCE_MIX1, MAGIC_ESSENCE_MIX2);
+		add(new SuperRestoreMix(), SUPER_RESTORE_MIX1, SUPER_RESTORE_MIX2);
+		add(combo(food(6), boost(DEFENCE, perc(.15, 5))), SUPER_DEF_MIX1, SUPER_DEF_MIX2);
+		add(combo(food(6), boost(RANGED, perc(.10, 4))), RANGING_MIX1, RANGING_MIX2);
+		add(combo(food(6), boost(MAGIC, 4)), MAGIC_MIX1, MAGIC_MIX2);
+		add(combo(3, boost(ATTACK, perc(.20, 2)), boost(STRENGTH, perc(.12, 2)), heal(PRAYER, perc(.10, 0)), new BoostedStatBoost(DEFENCE, false, perc(.10, -2)), new BoostedStatBoost(HITPOINTS, false, perc(-.12, 6))), ZAMORAK_MIX1, ZAMORAK_MIX2);
```
Nightfirecat, line 225: I had mentioned this on Discord, this is not the correct way to handle Zamorak Mix boosts--the hitpoints boost is incorrectly calculated here because the "heal" of the mixed potion is calculated separately from the drain of the zamorak brew, both against the starting boosted hp value. My PR #4036 handles this and has a test case showing the expected behavior. I'm inclined to prefer that implementation for barbarian potions to duplicating all of the potion effects for each entry and creating prayer/super restore/stamina mix classes specifically for those potions.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -169,21 +193,58 @@ private void init()
 		add(combo(3, boost(ATTACK, perc(.15, 5)), boost(STRENGTH, perc(.15, 5)), boost(DEFENCE, perc(.15, 5)), heal(HITPOINTS, -10)), DIVINE_SUPER_COMBAT_POTION1, DIVINE_SUPER_COMBAT_POTION2, DIVINE_SUPER_COMBAT_POTION3, DIVINE_SUPER_COMBAT_POTION4);
 		add(combo(2, boost(RANGED, perc(0.1, 4)), boost(DEFENCE, perc(0.15, 5)), heal(HITPOINTS, -10)), DIVINE_BASTION_POTION1, DIVINE_BASTION_POTION2, DIVINE_BASTION_POTION3, DIVINE_BASTION_POTION4);
 		add(combo(2, boost(MAGIC, 4), boost(DEFENCE, perc(0.15, 5)), heal(HITPOINTS, -10)), DIVINE_BATTLEMAGE_POTION1, DIVINE_BATTLEMAGE_POTION2, DIVINE_BATTLEMAGE_POTION3, DIVINE_BATTLEMAGE_POTION4);
+		add(combo(5, heal(PRAYER, perc(0.1, 2)), boost(MAGIC, perc(0.05, 2)), heal(ATTACK, perc(0.1, -2)), heal(STRENGTH, perc(0.1, -2)), heal(DEFENCE, perc(0.1, -2))),
+			ANCIENT_BREW4, ANCIENT_BREW3, ANCIENT_BREW2, ANCIENT_BREW1);
```
Nightfirecat, line 197: This is not correct as it does not handle the maximum capped value of prayer (only boosts up to 5% above max). I would prefer to take #14566 for this, once it's been completed.

Nightfirecat (commented):
I'd prefer to take ancient brews and mixed potions separately in their respective PRs, but the rest looks pretty good.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
emerald000: I tried at 4 different levels, and they all fit that formula which is similar to prayer potions. I don't have 100 accounts with different levels to test, but I am fairly confidant that it is correct.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
emerald000: It's +2 at level 1 and in the 40s, and +3 at level 99 and in the 50s. I don't remember the exact levels that were tested. That would be the only reasonable formula for those boosts.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
emerald000: I am a bit confused about what the `primaries` argument means. I don't quite understand the effect on the code that parameter has. If you look at the Jug of Wine, `primaries` is set to 1. Yet it changes both hitpoints and attack.

Nightfirecat: It's used to check all the passed `SingleEffect`s to determine the positivity of the resulting `Combo`

emerald000:
I believe I have addressed everything.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
Nightfirecat: Looks like this was removed, perhaps in favor of the entry on L104? I think `CUP_OF_TEA_1978` can be added to it, and the attack boost changed to 2 + 2%.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/special/NettleTea.java
```diff
@@ -0,0 +1,49 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package net.runelite.client.plugins.itemstats.special;
+
+import net.runelite.api.Client;
+import static net.runelite.api.Skill.HITPOINTS;
+import net.runelite.client.plugins.itemstats.StatBoost;
+import static net.runelite.client.plugins.itemstats.stats.Stats.RUN_ENERGY;
+
+public class NettleTea extends StatBoost
```
Nightfirecat, line 32: Rename this class to reflect that this is only the run energy boost (`NettleTeaRunRestore`?) or make it implement `Effect` and return both the hitpoints and optional run energy restoration there.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
emerald000: `CUP_OF_TEA` (id 712) is an unobtainable item. I changed it to `CUP_OF_TEA_1978` which is the obtainable version.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/special/NettleTea.java
emerald000: Class renamed. I feel it's better if you have the class do the least amount of custom work possible.

Issue #14796, opened 2022-03-26
Add Shayzien agility courses to KC chat command


invalidCards:
**Is your feature request related to a problem? Please describe.**
Not really

**Describe the solution you'd like**
Add lookups for `!kc` chat command for the Shayzien Basic and Shayzien Advanced agility courses

**Describe alternatives you've considered**
Typing it out myself like some kind of noob

**Additional context**
Even the werewolf course has a KC command so why not this one :)

LlemonDuck:
Some short names for the shayzien courses were added in ce0ec53:

![Uploading image.png…]()

PR #14834, opened 2022-04-05
Add support for shayzien agility course

Giovds:
Adds the shayzien basic and advanced agility course to the 'boss' list.
This way commands like !kc and !pb will work accordingly. Also moved
prif agility course closer to the other agility related bosses.

image

---

This should resolve issue #14796


Giovds:
Not sure if there are any naming conventions for the abbreviations, so I came up with some basic ones. Any other suggestions/edits are welcome of course.

PR #14840, opened 2022-04-07
Add mouse-over stats for Ancient Brews

### LINKED ISSUE #14838: Add mouse-over text for Ancient Brews.
Ancient Brews can restore prayer just like Super Restores and Prayer Potions. They have variable restoration amounts and could use similar mouse-over text to show how much it'll drain stats and restore prayer, and boost magic.

DavidLuong98:
Closes  #14838

Adds stats for the ancient brew.


Nightfirecat:
This implementation is not right, as it does not take into account the boost cap the potion implements. #14566 implements the boost correctly.

PR #14842, opened 2022-04-08
TimerPlugin: track home and minigame teleports cooldown using varps

superiorser9:
`Varp 892` stores the difference, measured in minutes, between the time home teleport spell was last used and midnight, January 1, 1970 UTC (`System.currentTimeMillis() / 60000`) and the cooldown expires 30 minutes after.

`Varp 888` stores the difference, measured in minutes, between the time minigame teleport was last used and midnight, January 1, 1970 UTC (`System.currentTimeMillis() / 60000`) and the cooldown expires 20 minutes after.

Which means e.g. if the player teleports to home at 01:00:49 (hh:mm:ss) the cooldown will expire at 01:30:00 (not at 01:30:49).

Checking varps is obviously much cleaner and much more accurate.


Nightfirecat:
Needs rebase since 2bc2e867bdc1b07be38f486193faeadfa780845f

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -151,11 +152,12 @@ public class TimersPlugin extends Plugin
 	private int lastPoisonVarp;
 	private int lastPvpVarb;
 	private int lastCorruptionVarb;
+	private int lastHomeTeleport;
+	private int lastMinigameTeleport;
```
Nightfirecat, line 156: Reset these in `shutDown()`

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -875,6 +894,25 @@ private void removeTzhaarTimer()
 		}
 	}
 
+	private void checkTeleport(VarPlayer varPlayer)
+	{
+		int lastTeleport = client.getVar(varPlayer);
+		long lastTeleportSeconds = (long) lastTeleport * 60;
+		boolean homeTeleport = varPlayer.equals(VarPlayer.LAST_HOME_TELEPORT);
```
Nightfirecat, line 901: Rather than writing this to handle only the binary of home teleport or minigame teleport, a switch statement or some other comparison operator should be used to assign a `final GameTimer` variable at the top of this method, eg.

```java
final GameTimer teleport;
switch(varPlayer)
{
	case VarPlayer.LAST_HOME_TELEPORT:
		teleport = HOME_TELEPORT;
		break;
	case VarPlayer.LAST_MINIGAME_TELEPORT:
		teleport = MINIGAME_TELEPORT;
		break;
	default:
		// Other var changes are not handled as teleports
		return;
}
```

Doing this is easily expanded upon if future var-controlled teleports are found, and removes the need for all the ternaries in this method.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -875,6 +894,25 @@ private void removeTzhaarTimer()
...
+
+		Instant teleportExpireInstant = Instant.ofEpochSecond(lastTeleportSeconds).plus(homeTeleport ? 30 : 20, ChronoUnit.MINUTES);
```
Nightfirecat, line 903: I don't think this calculation is needed. From my experience ingame, the teleport duration is not variable between 19-20 or 29-30 minutes. I suspect the transmitted var is epoch minutes rather than epoch seconds to reduce the size of the transmitted int, but is saved as epoch seconds on the server side.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
superiorser9: We'll have to disagree on this one. I tested this a few times (and once again after I read your review) and it seems correct.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -875,6 +896,40 @@ private void removeTzhaarTimer()
...
 
+	private void checkTeleport(VarPlayer varPlayer)
+	{
+		final GameTimer teleport;
+		final int time;
+		switch (varPlayer)
+		{
+			case LAST_HOME_TELEPORT:
+				teleport = HOME_TELEPORT;
+				time = 30;
```
Nightfirecat, line 907: The length of these timers can be gotten from the timer itself, eg. `teleport.getDuration().getSeconds()`, which can then be fed back into the expiry instant via `Instant.ofEpochSecond(lastTeleportSeconds).plus(teleportCooldownSeconds, ChronoUnit.SECONDS)`. That avoids hard-coding the duration, when it's already present in the enum definition.

Nightfirecat (commented):
I still need to test this ingame, but this is looking pretty good.

A thought which has been discussed regarding this plugin is that timers based on vars should probably not be destroyed and recreated whenever its var change sets (or re-sets) the time of the timer, since that causes the infobox to shift around in order in its group. It's out of scope for this PR, but we should address that at some point, perhaps with another `createGameTimer()` overload.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: I can confirm it works the way you describe--it is in fact variable between 19-20 and 29-30 minutes respectively for minigame and home teleports.

PR #14846, opened 2022-04-09
stop accounting for gim teleports as a left click option

MichaelLavin778:
matched up the versions, should be a better pull request now


MichaelLavin778:
sorry i keep messing this up

PR #14855, opened 2022-04-13
Fixed custom notification sounds.

CalebWhiting:
This is relative to my own system (Linux), I'm unaware if this behaviour was different on other platforms.

Previously the first notification was played twice, and subsequent calls were mute.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -471,11 +471,11 @@ public class Notifier
 				return;
 			}
 		}
-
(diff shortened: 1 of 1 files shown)


Nightfirecat:
As I mentioned in Discord, notification sounds (mostly) work for me on Linux, running Plasma under Xorg on a Debian 11 install. For me, the notification sound fired twice the first time I had a notification, then once every time thereafter. I strongly recommend digging deeper to understand _why_ it doesn't work for you, and why this change would fix it for you, especially since it directly contradicts the comment above which says not to do what you're doing.

CalebWhiting:
Reverted the change, however fixed the first notification repeating.
Would you rather I opened a new pull request?

PR #14857, opened 2022-04-13
hiscore: add rifts closed activity

BeirlaenAaron:


Adam-:
added in eb13c0153b5493674373ba3818e11bd900eb8c37

Adam-:
For future reference:

1) we usually make the builder backwards-compatible with the previous response, since occasionally they forget the new hiscore entry to one of the various endpoints
2) the activity icons are 18x18, not 16x16 like the Jagex hiscores use
3) we additionally strip the pngs to reduce their size (notice how mine is <1kb despite being a larger image)

PR #14860, opened 2022-04-14
plugin: Added config option to NpcAggroArea for hiding not working warning

OSRS-Athan:
Added an option to hide the warning that you have to teleport away or enter a dungeon to make the plugin work. This warning takes a lot of screen size and after using the plugin once you know how to fix it.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
@@ -128,4 +128,15 @@ public interface NpcAggroAreaConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "hideWarning",
+			name = "Hide warning",
+			description = "Hides warning for plugin not working.",
+			position = 9
+	)
+	default boolean hideWarning()
+	{
+		return false;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaConfig.java
@@ -130,10 +130,10 @@ public interface NpcAggroAreaConfig extends Config
 	}
 
 	@ConfigItem(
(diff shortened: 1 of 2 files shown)


Hydrox6:
If the warning is on the screen, then the plugin is not working because the user has not calibrated it. This is not an optional nicety; if you don't want to see the warning, either calibrate the plugin or turn the plugin off.

We have already seen the result of being able to hide this, and the result is users being confused as to why the plugin isn't working because they've enabled an option in the config without reading it. As a result, we will not be repeating it.

See #10420

OSRS-Athan:
@Hydrox6 Would it be ok if it only displayed the warning when the plugin is actually not working?

Example:
I currently have
```
Always active = false
NPC Names = [ Gargoyle, Dagannoth ]
```

This should make the plugin only work when I am in proximity of those NPC's.
However even during the time the plugin should not be working it is displaying a warning.

Hydrox6:
That is how the plugin used to work, which also confused users. Users tend to not want to leave the place they just got to to recalibrate the plugin, so them knowing that they need to calibrate it *before* their trip is very useful.

Again, if you don't want to see it, calibrate it or turn the plugin off.

See #14781

OSRS-Athan:
@Hydrox6

Well if you are going on a trip it is very likely you are going to use a teleport of some sort. So it would be a really small portion of players and they still are getting the warning when it matters. Calibrating it every time that warning shows is a much larger hassle. Also turning the plugin off would not be a solution because it stops tracking the aggro zone and when you need to use it it has to calibrate.

If however you still think this its a bad idea could we possible settle with a smaller icon for the warning with this warning as an on hover action?

Hydrox6:
Without the overlay, the plugin silently doesn't work. This is not good UX, especially when it's an identical state to the plugin working, but just not being active. Even if it's "very likely" that a user would accidentally calibrate it, that isn't enough.

That the plugin needs so much calibration is a known bug that I've submitted a fix for, and that bug originated from another bad "treat the symptom, not the cause" fix from 3 years prior.

The overlay is meant to be in your face and highly visible: it's telling you what to do to get the plugin working again. Making it less visible defeats the entire point of it. And once again, it's incredibly simple to get it off the screen: follow its instructions, or disable the plugin.

OSRS-Athan:
I'm sorry I did not see (#14833). That will fix the problem. Thanks!

PR #14892, opened 2022-04-23
Compost Tracking

LlemonDuck:
Implements compost tracking for farming patches using a similar concept as the Examine plugin.

The plugin watches for composting and inspect actions, then stores a pending compost which is finalized by chat messages. In order to distinguish between multiple compost actions queued, the tracker also checks for adjacency to the patch and has a timeout of 30 seconds.

At the time, the plugin does not display the information to the user, since I know there is ongoing discussion about how to do so. Hopefully this will help to drive that forward now that the detection portion is completed.

I think the test cases here are a little messy, but I didn't feel that things like `playerIsBesidePatch` should be extracted to a utility if they aren't used elsewhere.


LlemonDuck:
Following discussions on Discord:

* Rebased off 1edd41733613273ea01432a1bb39a33bbef7cb54 and removed player nullability checks
* Replaced `@Inject`-annotated properties with `@Inject`-annotated lombok `@RequiredArgsConstructor`
* Replaced scene scan for `patchObject` with constant-time tile index usage
* Use `NullObjectID` for patch IDs in `FarmingWorld`
* Use `FarmingTracker#updateData` to clear compost states on grown/dead crops instead of `VarbitChanged` event

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,264 @@
...
+		{
+			return;
+		}
+
+		PendingCompost pc = new PendingCompost(
+			Instant.now(),
+			WorldPoint.fromScene(client, e.getParam0(), e.getParam1(), client.getPlane()),
+			farmingPatch
+		);
+		log.debug("Storing pending compost action [{}] for patch [{}]", pc, farmingPatch);
```
abextm, line 139: redundant; there is a debug already when setting config

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,264 @@
...
+
+	public Compost getCompostState(FarmingPatch fp)
+	{
+		return configManager.getRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, configKey(fp), Compost.class);
+	}
+
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked e)
+	{
+		if (e.isConsumed())
```
abextm, line 97: always check MenuAction before any other params

abextm (commented):
- Of course this will need a UI to be merged. 
- All the new files you added need a license header.
- You have some empty lines with whitespace; formatting w\ IJ will remove these so they should not be present.
- `This is a flower patch. The soil has not been treated. The patch is empty and weeded.`
We can get rid of all the ids by doing something like this <https://github.com/abextm/runelite/commit/a9e83d2e63fc570b593bc729df2cf1ef76cb4f08>

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,264 @@
...
+		int maxY = minY + patchObject.sizeY() - 1;
+
+		// player should be within one tile of these coords
+		// todo corner detection?
+		return playerX >= (minX - 1) && playerX <= (maxX + 1) && playerY >= (minY - 1) && playerY <= (maxY + 1);
+	}
+
+	private void filterTimeouts()
+	{
+		pendingCompostActions.entrySet()
```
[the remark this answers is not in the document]
LlemonDuck: Done, but changed to reference a `PendingCompost#timeout` instead.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
LlemonDuck: I disagree, this one helps to indicate that we are anticipating a compost action, before the compost is applied.

LlemonDuck:
@abextm Updated to use `ObjectComposition#getVarbit` instead, and a few other small changes. Let me know what you think so far. After this, I think all we need is a UI.

LlemonDuck:
Added a ui (distinction between compost types is present in tooltip):

image

In order to place the images overtop of each other, it adds a `JLabel overlayIcon` to `TimeablePanel`, and places both the original `icon` and the new `overlayIcon` into a JLayeredPane. This also requires that it set custom bounds for both labels.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,294 @@
...
+		int playerY = p.getWorldLocation().getY();
+
+		// patch coords
+		int minX = patchObject.getWorldLocation().getX();
+		int minY = patchObject.getWorldLocation().getY();
+		int maxX = minX + patchObject.sizeX() - 1;
+		int maxY = minY + patchObject.sizeY() - 1;
+
+		// player should be within one tile of these coords
+		// todo corner detection?
```
Adam-, line 256: you can't interact with objects from corners, but maybe you can use spells from corners or further away? Have you checked that?

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,294 @@
...
+		return playerX >= (minX - 1) && playerX <= (maxX + 1) && playerY >= (minY - 1) && playerY <= (maxY + 1);
+	}
+
+	private void filterTimeouts()
```
Adam-, line 260: This should be called something else. maybe expirePendingActions

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTabPanel.java
```diff
@@ -132,6 +141,9 @@ public class FarmingTabPanel extends TabContentPanel
 			}
 		}
 
+		AsyncBufferedImage compostImage = itemManager.getImage(ItemID.COMPOST);
+		compostImage.onLoaded(() ->
```
Adam-, line 145: I think I would prefer just building the compost image when the panel is updated, like weeds and the others. Also that way, you can make the image of the compost use the same compost type as is applied to the patch  (even though they items looks very similar there is a small distinction)

runelite-api/src/main/java/net/runelite/api/widgets/WidgetInfo.java
```diff
@@ -505,6 +505,8 @@
 	SPELL_ARCEUUS_HOME_TELEPORT(WidgetID.SPELLBOOK_GROUP_ID, WidgetID.ArceuusSpellBook.ARCEUUS_HOME_TELEPORT),
 	SPELL_KOUREND_HOME_TELEPORT(WidgetID.SPELLBOOK_GROUP_ID, WidgetID.StandardSpellBook.KOUREND_HOME_TELEPORT),
 	SPELL_CATHERBY_HOME_TELEPORT(WidgetID.SPELLBOOK_GROUP_ID, WidgetID.StandardSpellBook.CATHERBY_HOME_TELEPORT),
+
```
Adam-, line 508: remove this

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingTabPanel.java
```diff
@@ -132,6 +141,9 @@ public class FarmingTabPanel extends TabContentPanel
...
+			this.compostOverlayImage = ImageUtil.resizeImage(compostImage, TimeablePanel.OVERLAY_ICON_BOUNDS.width, TimeablePanel.OVERLAY_ICON_BOUNDS.height));
```
Adam-, line 146: Probably the timeable panel should be responsible for resizing this

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,294 @@
...
+	private static final Pattern COMPOST_USED_ON_PATCH = Pattern.compile(
+		"You treat the (\\w+)( patch)? with (?<compostType>ultra|super|)compost\\.");
+	private static final Pattern FERTILE_SOIL_CAST = Pattern.compile(
+		"The (\\w+)( patch)? has been treated with (?<compostType>ultra|super|)compost\\.");
+	private static final Pattern ALREADY_TREATED = Pattern.compile(
+		"This (\\w+)( patch)? has already been (treated|fertilised) with (?<compostType>ultra|super|)compost( - the spell can't make it any more fertile)?\\.");
+	private static final Pattern INSPECT_PATCH = Pattern.compile(
+		"This is an? (\\w+)( patch)?\\. The soil has been treated with (?<compostType>ultra|super|)compost\\..*");
+
+	private static final ImmutableSet<Integer> compostItems = ImmutableSet.of(
```
Adam-, line 86: by convention all of our static final fields are usually uppercase

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
```diff
@@ -0,0 +1,294 @@
...
+	{
+		Instant timeout;
+		WorldPoint patchLocation;
+		FarmingPatch farmingPatch;
+	}
+
+	private static final Duration COMPOST_ACTION_TIMEOUT = Duration.ofSeconds(30);
+
+	private static final Pattern COMPOST_USED_ON_PATCH = Pattern.compile(
+		"You treat the (\\w+)( patch)? with (?<compostType>ultra|super|)compost\\.");
```
Adam-, line 78: Use non-capturing groups if you don't need the values such as `(?: patch)?` - and don't include unnecessary captures in general such as `(\\w+)`

```suggestion
		"You treat the \\w+(?: patch)? with (?<compostType>ultra|super|)compost\\.");
```

etc for the patterns below.

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/CompostTracker.java
LlemonDuck: The spell drags the player toward the patch and places you beside it before activating. I have not tried corners for casting yet.

LlemonDuck: Reconfirmed, can't cast the spell from a distance or diagonally. The player is dragged toward the patch.

LlemonDuck:
Updated to address your feedback @Adam-, and renamed the commit to have header `timetracking: compost tracking`

PR #14921, opened 2022-05-05
[WorldMap] Add strength & ranged requirements to WorldMap AgilityShortcut icons & tooltip 

NinjaLikesCheez:
This change adds strength & ranged level requirements to `AgilityShortcuts` (lifted from the Wiki, might not be inclusive of all agility shortcuts) and uses those requirements in the WorldMap plugin to strike out icons which are not player usable.


ineffyble:
Would be great if this can be merged.

PR #14930, opened 2022-05-08
clues: Fix capitalization for Dark Mage anagram

Cyborger1:
It seems like the capitalization changed at some point, likely after GotR.

image


Nightfirecat:
Looks like the location is also off by one tile based on that screenshot. Could that be corrected here as well?

PR #14963, opened 2022-05-13
clientui: setResizable after setVisible

### LINKED ISSUE #14235: Double-clicking the title bar on macOS still maximizes the window even if "Lock window size" is enabled
"Lock window size" is enabled in RuneLite settings, window size is set to 1280x800 and "Keep game size". Resizing arrows do not appear at the edges of the window, as expected, but double-clicking the window title bar expands the window to fit the screen. Double clicking it again does not restore the original size.

MacBook Air M1
macOS 11.4 Big Sur
RuneLite M1 client (latest version)

LlemonDuck:
Prevents an issue with clients being resizable with toolbar-less means in some environments (e.g. double clicking title bar on macos).

Honestly I'm not entirely sure why this is ignored if the window isn't visible, rather than being propagated to the window manager when the window is shown, but I guess Swing isn't that smart (like always).

I originally found the information on [this SO answer](https://stackoverflow.com/a/14884056) and it seemed to resolve the issue. The original author seems similarly unsure about its internal mechanisms.

Closes #14235

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -488,7 +488,7 @@ public class ClientUI
 			}
 
 			// Update config
-			updateFrameConfig(true);
+			updateFrameConfig(false);
 
 			// Create hide sidebar button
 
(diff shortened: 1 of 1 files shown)


Adam-:
Do you know if this `updateResizable` is still necessary? This might be related to 28dd5f477ceb4399cff8bbff0d6beaa8a78129c9

LlemonDuck:
I needed to move the `setResizable` to after the `setVisible(true)`, but I wanted to still initialize the other config-based frame properties at the existing call-site.

`updateResizable` just links the property to the config as a single call so it doesn't get decoupled some later time.

PR #14965, opened 2022-05-13
Fix switching between GPU plugins

Mark7625:
Since the SceneUploader code is largely the same between different GPU
plugins, there can be scene ID collisions if the plugins are initialized
around the same time then switched off and on a small number of times. Making the plugin not show objects or making the scene upload wrong. nextInt() returns any of the 2^32 possible numbers so we're down to about 1/4.3B chance of colliding. This fixes some weird behaviour when switching from 117,GPU,and other plugins that have GPU, While 117 has this already GPU will also need the fix so it will fix any future issues with gpu plugins or forks


ipkpjersi:
Can confirm this fixes the issues I've experienced when switching between GPU and HD plugins, LGTM.

Nice job.

clivester90:
This has helped me fixed a number of issues as well, Thank you mark!

Nightfirecat:
Added in b9da0435d2d6069ef2de65683f06b4d9bd7324da

PR #14975, opened 2022-05-15
disable left click for placeholders bank button

### LINKED ISSUE #14961: Add 'Disable left-click' for the placeholder toggle
This added config should work the same way as the ones for depositing your inventory or equipment do--when enabled, left-clicking the "Toggle placeholders" button (with the :lock: symbol) should force a minimenu to pop up.

---

### Discussed in https://github.com/runelite/runelite/discussions/14943

<sup>Originally posted by **Memebeams** May 12, 2022</sup>
Currently, the bank plugin allows you to avoid accidental deposits by disabling left-click on deposit buttons:
image

Sometimes, I will remove items from my bank without realizing that the placeholder button was turned off and mess up my bank arrangement. This button would make it harder for dingbats like me to turn placeholders off</div>

orange-puff:
Closes #14961

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bank/BankConfig.java
@@ -102,10 +102,21 @@ public interface BankConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "rightClickPlaceHoldersSetting",
+		name = "Disable left click for placeholders setting",
+		description = "Configures whether the placeholders button will turn on/off on left click",
+		position = 7
+	)
+	default boolean rightClickPlaceHoldersSetting()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "seedVaultValue",
 		name = "Show seed vault value",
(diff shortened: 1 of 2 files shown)


OAGallo (approved):

Nightfirecat (changes requested):
As written, this causes the right-click menu to be forced in places which are not the bank. An easy-to-access example is the scroll wheel zoom button in the settings menu. This should only apply while in the bank.

orange-puff:
I've tested this and it appears to work, but this looks like a cleaner way to handle this. https://github.com/runelite/runelite/pull/14997 Maybe they can reopen their PR instead

Fiffers:
I'll reopen once a separate PR of mine is reviewed, don't want to mix them up.

Nightfirecat:
@Fiffers I definitely recommend learning to use git branches so you're not stuck maintaining only a single PR at a time.

Fiffers:
> @Fiffers I definitely recommend learning to use git branches so you're not stuck maintaining only a single PR at a time.

Sounds good 👍

Nightfirecat:
I agree #14977 was the better approach, I've squashed its commits and pushed 1ca54e0f75a519c3fe3dae3761644656082a31e6

PR #14998, opened 2022-05-21
Added an option to the implings plugin that allows for notifications …

### LINKED ISSUE #14693: Implings - Only highlights or notify if you can catch them
Currently, you can set either a highlight or notification for certain Impling types. However, this needs to be changed when either playing different accounts or if your hunter level changes.

This should be a fairly simple solution - Simply have a dropdown option called "Level Notification Override" with three options, None, Net, and Barehanded.
When any of these are selected, it overrides the set notifications based on your level. 
There could be an additional setting called "Level Highlight override", which has the same effect but for highlighting.

For example, if I'm level 36 and I have every impling set to Notify, setting the notification override to Net will notify me for Baby, Young, Gourmet, and Earth, while ignoring all implings I'm unable to catch.

This would turn the plugin into a "set and forget" style setup, preventing random pings for implings you can't catch, while giving the option for those who hunt implings to inform others.

thomaszee:
…to only be shown when the level requirement is met for the selected catching method.

Close #14693

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingType.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingType.java
@@ -31,18 +31,20 @@ import lombok.Getter;
 @Getter
 enum ImplingType
 {
-	BABY("Baby"),
-	YOUNG("Young"),
-	GOURMET("Gourmet"),
-	EARTH("Earth"),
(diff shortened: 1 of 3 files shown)


thomaszee:
Not too proud of the namings in the config class, but couldn't come up with something better. Feel free to suggest more fitting variable / config names and descriptions.

PR #15000, opened 2022-05-22
Item Charges -- Bracelet of Clay and Castle Wars bracelet added

BearSprayGit:


MasonPMGit:
Sorry about that, the formatting should be fixed now though.

BearSprayGit:
Added support for multiple configs per item
Added explorers ring energy charges
Added explorers ring tele charges for explorers ring 2

PR #15014, opened 2022-05-29
TimerPlugin: Check stamina effect using varbits

superiorser9:
`Varbit 24` holds the info of number of game ticks remaining on stamina effect in intervals of 10;

`Varbit 10385` holds the info of number of game ticks remaining on ring of endurance effect in intervals of 10;

Drinking a stamina potion while wearing the ring of endurance will make `varbit 24` and `varbit 10385` change to 20 and `varbit 10385` will decrease every 10 game ticks until it reaches 0 (assuming the ring wans't unequipped) then `varbit 24` will start decreasing. 
Meaning `varbit 10385` will prioritize over `varbit 24`. 
If the ring was unequipped, `varbit 10385` changes to 0 and `varbit 24` will start decreasing.

It was discussed before that the `setDuration` method isn't the optimal solution, but I don't see any other option without recreating the infobox.

Checking varbits is obviously much cleaner and much more accurate.


runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -47,6 +47,21 @@
 	 */
 	public static final int RUN_SLOWED_DEPLETION_ACTIVE = 25;
 
+	/**
+	 * Stamina effect timer
+	 * Number of game tick remaining on stamina effect in intervals of 10; for a value X there are 10 * X game ticks remaining.
```
Nightfirecat, line 52: typo, I'll commit via GH's interface
```suggestion
	 * Number of game ticks remaining on stamina effect in intervals of 10; for a value X there are 10 * X game ticks remaining.
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -341,6 +342,32 @@ else if (!imbuedHeartTimerActive)
 			checkTeleport(VarPlayer.LAST_MINIGAME_TELEPORT);
 			lastMinigameTeleport = minigameTeleportVarp;
 		}
+
+		if (staminaEffectActive == 1 && lastStaminaEffect != totalStaminaEffect && config.showStamina())
+		{
+			final Duration staminaDuration = Duration.of(10L * totalStaminaEffect, RSTimeUnit.GAME_TICKS);
+
+			if (staminaTimer == null && !staminaDuration.isZero())
```
Nightfirecat, line 350:
```suggestion
			if (staminaTimer == null && totalStaminaEffect > 0)
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -341,6 +342,32 @@ else if (!imbuedHeartTimerActive)
...
+			{
+				staminaTimer = createGameTimer(STAMINA, staminaDuration);
+			}
+			else if (staminaDuration.isZero())
```
Nightfirecat, line 354:
```suggestion
			else if (totalStaminaEffect == 0)
```

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -223,6 +219,11 @@ public void onVarbitChanged(VarbitChanged event)
 		int imbuedHeartCooldownVarb = client.getVarbitValue(Varbits.IMBUED_HEART_COOLDOWN);
 		int homeTeleportVarp = client.getVar(VarPlayer.LAST_HOME_TELEPORT);
 		int minigameTeleportVarp = client.getVar(VarPlayer.LAST_MINIGAME_TELEPORT);
+		int staminaEffectActive = client.getVarbitValue(Varbits.RUN_SLOWED_DEPLETION_ACTIVE);
```
Nightfirecat, line 222: Is there ever a case where this would be `0`, but `lastStaminaEffect` and `totalStaminaEffect` would still be the same value? I don't think we need to check this--I believe `(lastStaminaEffect != totalStaminaEffect && config.showStamina())` is all that's needed to check for stamina boost changes.

superiorser9: I can't think of any logical case where that would happen, but it's checked here: https://github.com/Joshua-F/cs2-scripts/blob/741271f0c3395048c1bad4af7881a13734516adf/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L25
So I'd assume there might be a special case where this would happen... or it's just jagex being jagex...
Up to you, If you think it needs to be removed, it's gone.

PR #15068, opened 2022-06-04
ThreeStepCrypticClue: Handle step resets when scroll parts are lost; propagate requirements of steps to combined clue

### LINKED ISSUE #15067: Multi-step cryptic clue does not show spade requirement
### Type

Incorrect behavior

### Operating System

_No response_

### OS Version

_No response_

### Bug description

Multi-cryptic clues with sub-steps requiring spades don't indicate a spade is needed. Without looking at the code, I don't know whether this is because the multi-cryptic clue does not inherit that trait from its steps, or if it is because the step does not have that trait defined. (or both!) This should be fixed so that all active steps have their various requirements (spade, light source, etc.) propagated up to the multi-cryptic clue.

### Screenshots or videos

_No response_

### RuneLite version

_No response_

### Logs

_No response_

Nightfirecat:
Fixes #15067

See commit messages for info.


Nightfirecat:
After some discussion in Discord, I plan to drop the first commit as we don't want to erroneously re-add the step to the overlay when scroll parts are banked.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/ThreeStepCrypticClue.java
```diff
@@ -80,7 +94,24 @@ public static ThreeStepCrypticClue forText(String plainText, String text)
...
+				continue;
+			}
+
+			return clue;
+		}
+		return null;
+	}
+
+	@VisibleForTesting
+	final List<Map.Entry<CrypticClue, Boolean>> clueSteps;
```
LlemonDuck, line 114: SortedMap?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/ThreeStepCrypticClue.java
```diff
@@ -147,6 +185,36 @@ private boolean checkForPart(final ItemContainer itemContainer, int clueScrollPa
 		return false;
 	}
 
+	private void propagateRequirements(final Map.Entry<CrypticClue, Boolean> clueStep)
```
LlemonDuck, line 188: Personally I prefer not to pass around Map.Entry values since they effectively become unnamed tuples and are harder to read than just unpacking the value into named parameters

LlemonDuck (approved):
looks pretty good!

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/ThreeStepCrypticClue.java
Nightfirecat: We don't want to sort by keys though, we want to maintain insertion order.

Nightfirecat: I opted to change this to a `List<ThreeStepCrypticClueStep>`, with that class being a simple POJO:

```java
@AllArgsConstructor
@Data
class ThreeStepCrypticClueStep
{
	final CrypticClue clue;
	boolean completed;
}
```

LlemonDuck (approved):

PR #15076, opened 2022-06-08
Update Cannon Plugin to use Varplayer Instead of Chat Messages

Bonfire:
VarPlayer `3` holds the player's current cannon ammo count. Previously, this VarPlayer could not be used as it did not provide cannon ammo count. This has since changed and is once again correctly providing the current ammo count. Related tests located in CannonPluginTest have been modified to accommodate these changes and pass accordingly.

I have tested these changes to the best of my abilities.

Disclaimer: This is my first time contributing to the RuneLite repo and my first time writing tests for a plugin. Please let me know if you find any issues or if there are any areas that can be improved. Thank you :)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -30,7 +30,6 @@ import java.awt.Color;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -309,33 +294,14 @@ else if (event.getMenuAction() != MenuAction.GAME_OBJECT_FIRST_OPTION)
...
 
 	@Subscribe
-	public void onProjectileMoved(ProjectileMoved event)
+	public void onVarbitChanged(VarbitChanged varbitChanged)
 	{
-		Projectile projectile = event.getProjectile();
+		cballsLeft = client.getVarpValue(VarPlayer.CANNON_AMMO.getId());
 
-		if (CANNONBALL_PROJECTILE_IDS.contains(projectile.getId()) && cannonPosition != null && cannonWorld == client.getWorld())
+		if (config.showCannonNotifications() && !cannonBallNotificationSent && cballsLeft > 0 && config.lowWarningThreshold() >= cballsLeft)
```
Adam-, line 301: Either check the event index or otherwise avoid hitting a config call per varbit change

Bonfire: I've made it so that the varp value is only obtained if the index matches the CANNON_AMMO id. Thank you for the feedback

Adam-:
added in 3ef90957718198742ac8d30b7bad1a30b67b88db

PR #15089, opened 2022-06-11
worldhopper: Change the US flag icon to differentiate US East vs US West servers

wqian94:
Differentiation is currently done by checking against an immutable List.
It may be smarter to eventually implement a ping-based approach to
accommodate new servers and server relocations. The default fall-back is
the regular US flag icon.

image


wqian94:
No longer needed as of latest client update.

PR #15097, opened 2022-06-12
timers: fix teleport timers showing up even if config is disabled

hex-agon:


hex-agon:
Fixed in 575469a3cc401df32888ae055385a9fc79d5fa8a

PR #15105, opened 2022-06-13
Fishing plugin: Add an option to draw a pie progress indicator on fishing spots.

### LINKED ISSUE #14633: [Request] Fishing spot age
When AFK barbarian fishing and only looking at the screen every couple minutes, it would be nice to know which fishing spot is the "youngest" to be able to maximize AFK time.

Some sort of indicator, perhaps changing the color of the fishing spot from green to red, to indicate its general age. I'm aware there is a range as to when a fishing spot can despawn, but even a couple categories signifying "young", "medium", and "old", would help greatly.

tvanslyke:
This is a QoL change to make fishing even AFKer; players will be able to clearly see which nearby fishing spots are the youngest.
This change extends the existing functionality for minnow fishing spots, in which a pie progress indicator is drawn to show its age, to other fishing spot types.
This change preserves the existing behavior for minnows and keeps the minnow-specific options while adding a new option to enable spot timers for other spots.

Additionally, this change adds an overlay "warning" color option that the timer fades to as the spot ages.
This uses the existing "20% of the remaining time left" as the point for the full warning color, since that's what the minnow options were doing with their hard-coded warning color.

Additionally this change makes the pie indicator smaller, draws it to the bottom-right of the spot icon, and makes it the last overlay to be rendered so that it isn't covered by the spot icon.
Without this change it is difficult to see how much progress is on the timer; making it icon smaller and drawn in the bottom-right also minimizes occlusion of the spot icon.

This change reuses the existing logic with MinnowSpot, but renames it to TrackedFishingSpot to better indicate that it no longer tracks only minnow spots.
This change also adds new debug logging to make it easier to measure fishing spot timers in the future.
For example, this logging was used to show that, despite the claims on the wiki, anglerfish spots appear to last for the same duration as other fishing spots.
Karambwanji should probably be measured at some point since those spots seem to last a very long time (not karambwan spots, which last forever).

Close #14633


runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingSpotOverlay.java
```diff
@@ -191,10 +165,68 @@ else if (spot == FishingSpot.HARPOONFISH && npc.getId() == NpcID.FISHING_SPOT_10
...
+	public Duration getMaxSpotTime(FishingSpot spot)
+	{
+		switch (spot)
+		{
+			case MINNOW:
+				return Duration.ofSeconds(15);
+			default:
+				// 530 ticks, as per the wiki: https://oldschool.runescape.wiki/w/Fishing_spots
+				// The wiki claims that Anglerfish spots last longer than normal spots but this didn't seem to be true in testing.
+				return Duration.ofMinutes(5).plus(Duration.ofSeconds(18));
```
Nightfirecat, line 229: This can use `RSTimeUnit.GAME_TICK` rather than 5m18s.

Nightfirecat:
Also, what behavior does this have for fishing spots which spawn because they enter the area in which they are visible to you, rather than because they move on their own?

PR #15109, opened 2022-06-14
Blast Furnace Red/Green Clickboxes for Pipes, Drive Belt and Cogs

jmfallecker:
Many people use the Blast Furnace as a method to train Crafting. This should help distinguish which pipes are broken (red) and which are fixed (green).


Nightfirecat:
How many is "many"? This feature seems extremely niche and better-suited as a hub plugin.

jmfallecker:
I know quite a few UIMs who use this method to train crafting for free, but I can certainly do a hub plugin rather than this. It's most likely not the majority of the user base who use this. Although, that's also why I defaulted it to "off" so that it wouldn't confuse anyone who uses the plugin for smelting.

Nightfirecat:
Added to the plugin hub in runelite/plugin-hub#2839

PR #15112, opened 2022-06-15
#15110 - Support double ammo mould in SmeltingPlugin

### LINKED ISSUE #15110: smelting plug-in doesn't account for new 8-cannonball mould (giants foundry)
### Type

Incorrect behavior

### Operating System

MacOS

### OS Version

Mac OS Big Sur

### Bug description

A new 8-ball cannonball mould was released as a reward in Giant's Foundry. It takes the same number of ticks to turn two (2) steel bars into eight (8) cannonballs that a normal 4-ball ammo mould takes to turn one (1) steel bar into four (4) cannonballs

However, the plugin only counts 4 at a time, not accounting for the potential that the new ammo mould is in use, which should double the cannonballs smelted. (and I'm unsure whether that should be reflected in the actions per hour).

ADDITIONALLY: I'm unsure if this is a bug, but it may be worth looking into: this may need to be reflected within the XP globes if they use the same variables!

### Screenshots or videos

Screen Shot 2022-06-14 at 6 41 59 PM

### RuneLite version

Runelite version: 1.8.22
Oldschool version: Rev 205
Launcher version: 2.3.0

### Logs

[launcher.log](https://github.com/runelite/runelite/files/8904335/launcher.log)
[client.log](https://github.com/runelite/runelite/files/8904336/client.log)

Thource:
Fix #15110 by checking if there is a double ammo mould in the inventory

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingPlugin.java
@@ -31,6 +31,10 @@ import javax.inject.Inject;
 import lombok.AccessLevel;
 import lombok.Getter;
 import net.runelite.api.ChatMessageType;
+import net.runelite.api.Client;
+import net.runelite.api.InventoryID;
+import net.runelite.api.ItemContainer;
+import net.runelite.api.ItemID;
 import net.runelite.api.MenuAction;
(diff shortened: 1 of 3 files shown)


Adam-:
Added in #15446 - your code is wrong because it doesn't account for making cannonballs out of a single bar

PR #15113, opened 2022-06-15
Change Loot tracker "Collapse / Un-collapse All" button tooltip text

### LINKED ISSUE #15111: Loot tracker "Collapse / Un-collapse All" button has wrong tooltip text
### Type

Incorrect behavior

### Bug description

Hovering the collapse/un-collapse all button in the loot tracker displays the wrong text in its tooltip. When it will collapse all the boxes, it displays "Un-collapse All", and vice versa.

Trentr9:
Change the modal text to be the correct action on the collapse button. After looking at the logic I believe that the this change makes the most sense as the button state should be 'on' if all loot boxes are collapsed.

This should resolve #15111


Adam-:
fixed in #15118

Issue #14885, opened 2022-04-22
[World Switcher] Enable locked rows for favourited worlds


drawkel:
**Is your feature request related to a problem? Please describe.**
When I scroll through the worlds, the marked favourite worlds along with the column descriptors (World / # / Activity / Ping) scroll with the rest of the worlds causing me to have to scroll to the top to select them again.

**Describe the solution you'd like**
I'd like the rows for the worlds to be frozen, this means that when I wish to select a world that requires me to scroll down, the world # / # of people / Activity / Ping alongside the favourited worlds will scroll down with it.
There can be a separator to make it obvious which are the favourite worlds at a glance, but it's up to what you all feel is a better fit.

**Describe alternatives you've considered**
Scrolling back up to select the favourited worlds. Hiding less desirable worlds to have more to click at without scrolling.

**Additional context**
I feel this is pretty understandable, but if necessary, I can draw a visual later on.

PR #15115, opened 2022-06-15
[#14885] World Switcher - Pin favourite worlds to the top of the panel

Thource:
This change splits the World Switcher panel into 2 JScrollPanes - a pane for favourite worlds and a pane for the rest.

The favourite world pane will grow to fit 10 worlds, then a scroll bar appears after that.
The other pane fills the rest of the panel space.
There's a 12 px gap under the favourite worlds pane, so that they're seperated from the rest of the worlds.

preview

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/FixedWidthPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/FixedWidthPanel.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2017, Adam <Adam@sigterm.info>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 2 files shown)


LlemonDuck:
I won't comment on whether this is a user interface we should change, but...

The two sections in your screenshot don't really look visually distinct enough, I'm sure this would confuse some users who don't notice the gap. I think it would help to include the sorting headers in both panels so they visually appear as separate tables.

Thource:
> I won't comment on whether this is a user interface we should change, but...
> 
> The two sections in your screenshot don't really look visually distinct enough, I'm sure this would confuse some users who don't notice the gap. I think it would help to include the sorting headers in both panels so they visually appear as separate tables.

Should the 2 sections be individually sortable or should each header sort both sections?

Thource:
Here's a new screenshot of how it looks with 2 headers: 
image

I feel like a search bar could be useful but probably wants to be implemented in a seperate PR, maybe make an issue for it?

Trentr9:
The panel just looks too busy with two headers there.

Would it be worth making the header stick to the top of the pane and not move when you scroll down, along with the worlds you've set as favs.

Might mean the whole panel needs some reworking though

Thource:
> The panel just looks too busy with two headers there.
> 
> Would it be worth making the header stick to the top of the pane and not move when you scroll down, along with the worlds you've set as favs.
> 
> Might mean the whole panel needs some reworking though

That’s how I had it initially 😆

Trentr9:
> > The panel just looks too busy with two headers there.
> > Would it be worth making the header stick to the top of the pane and not move when you scroll down, along with the worlds you've set as favs.
> > Might mean the whole panel needs some reworking though
> 
> That’s how I had it initially 😆

I've had a quick play around with the plugin layout and came up with this so far

(New on the left, old on the right)
image

The header up the top is 'sticky' so when a user scrolls down the header stays visible, there are a two ways that I think we could go from here, have the world favourites also be 'sticky' below the header so that they are always visible to the user, or not be sticky and just stay up the top like the currentl behaviour.

I think regardless if they are sticky or not they need to have something to differentiate them from non-favourited worlds. Thoughts?

Thource:
> > > The panel just looks too busy with two headers there.
> > > Would it be worth making the header stick to the top of the pane and not move when you scroll down, along with the worlds you've set as favs.
> > > Might mean the whole panel needs some reworking though
> > 
> > 
> > That’s how I had it initially 😆
> 
> I've had a quick play around with the plugin layout and came up with this so far
> 
> (New on the left, old on the right) image
> 
> The header up the top is 'sticky' so when a user scrolls down the header stays visible, there are a two ways that I think we could go from here, have the world favourites also be 'sticky' below the header so that they are always visible to the user, or not be sticky and just stay up the top like the currentl behaviour.
> 
> I think regardless if they are sticky or not they need to have something to differentiate them from non-favourited worlds. Thoughts?

The ticket that this relates to (#14885) specifies pinning both the header and favourite worlds to the top of the panel.

@drawkel what do you think?

drawkel:
Yeah that looks like what I was thinking! I do agree that having both headers being 'sticky' at the top would be a bit bulky. Maybe having the favourited worlds 'stick' on the bottom instead (similar to how the in-game UI's World Switcher looks) with the option to toggle in the settings!

Just a thought, if not, the revised way with the top header sticking already looks awesome :)

Trentr9:
A few bugs to iron out, but I think something like this would be a good improvement.

WorldSwitcherPluginFavExample

drawkel:
Wow, that looks amazing! :)

Trentr9:
I’ve started playing around with sticky world favs.

Will post some screenshots later

PR #15140, opened 2022-06-18
spec-counter: handle thralls and venge conflicts

InfernoStats:
This PR updates the Spec Counter Plugin to:

1. Fix thralls being confused for spec hits - predominantly an issue when a DWH spec didn't land but is reported to have OR a BGS spec lands but is reported as a 1-3.
2. Fix vengeance being confused for spec hits*
3. Transmits 0 hits

* Vengeance can still rarely be confused if it is closer to the original expected damage due to scaling than the true hitsplat.

This obviously is not perfect, the only way to do that would be to get rates for every scaled boss (done for ToB but not CoX), but it gets the plugin closer to being more accurate while eliminating a pretty common issue at Xarpus.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -163,6 +163,27 @@ public class SpecialCounterPlugin extends Plugin
 	@Subscribe
 	public void onGameTick(GameTick event)
 	{
+		if (client.getGameState() != GameState.LOGGED_IN)
+		{
+			return;
+		}
+
+		assert client.getLocalPlayer() != null;
+		int currentRegion = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
+		boolean inInstance = client.isInInstancedRegion();
+
+		// if the player left the region/instance and was fighting boss that resets, reset specs
+		if (currentRegion != previousRegion || (wasInInstance && !inInstance))
+		{
+			if (RESET_ON_LEAVE_INSTANCED_REGIONS.contains(previousRegion))
+			{
+				removeCounters();
+			}
+		}
+
+		previousRegion = currentRegion;
+		wasInInstance = inInstance;
+
 		if (this.specialWeapon == null)
 		{
 			reset();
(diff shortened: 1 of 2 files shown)


Adam-:
> This PR updates the Spec Counter Plugin to:
> 
>     1. Fix thralls being confused for spec hits - predominantly an issue when a DWH spec didn't land but is reported to have OR a BGS spec lands but is reported as a 1-3.
> 
>     2. Fix vengeance being confused for spec hits*
> 
>     3. Transmits 0 hits
> 
> 
>     * Vengeance can still rarely be confused if it is closer to the original expected damage due to scaling than the true hitsplat.
> 
> 
> This obviously is not perfect, the only way to do that would be to get rates for every scaled boss (done for ToB but not CoX), but it gets the plugin closer to being more accurate while eliminating a pretty common issue at Xarpus.

My understanding is this is mostly not possible without understanding the xp scaling, which we don't have. How does this work?

Adam-:
> > How does this work?
> 
> This works by combining both hitsplat and xp information - when we get a spec, use the xp gained and match the hitsplat that best fits the XP drop.
> 
> Common Scenarios on a Non-Scaled Boss (XP Received is in Strength):
> 
>     1. XP Received: 0, Hitsplats: [0] - Since we did not receive any XP for using our spec, the internal damage counter is 0 and we transmit that.
> 
>     2. XP Received: 0, Hitsplats: [0, 2] - Since we did not receive any XP for using our spec, the internal damage counter is 0 and we transmit that. Currently the plugin would pick up the 2 from the thrall and transmit a DWH hit.
> 
>     3. XP Received: 200, Hitsplats: [50, 3] - Since we did receive XP for using our spec, the internal damage counter is set to 50 and we find the distance between each hitsplat. The distance between 50 and 50 is 0 which is less than 47, so we assume that the 50 hitsplat is our true damage.
> 
> 
> Common Scenarios on a Scaled Boss (XP Received is in Strength, Boss is Bloat at 1.975x rate in 5 scale ToB):
> 
>     1. XP Received: 395, Hitsplats: [50] - Since we did receive XP for using our spec, the internal damage counter is set to 395/4 or 98.75. Seeing as we only have one hitsplat, that is the closest to our expected damage and we transmit a 50 BGS.
> 
>     2. XP Received: 395, Hitsplats: [50, 3] - Since we did receive XP for using our spec, the internal damage counter is set to 395/4 or 98.75. Since we have two hitsplats, the closest to our expected damage is 50 (48.75) and not 3 (95.75) and we transmit a 50 BGS.
> 
> 
> The goal of this PR is to ensure that if a DWH is hit or missed, it is _always_ accurate. We will never transmit a missed DWH if we hit and vice versa.
> 
> This PR does not guarantee 100% accuracy If in the rare case a vengeance is recoiled on the same tick that you hit with a BGS or `isDamage` weapon. In that case, the closest to the expected damage is taken and it _may_ be wrong. For example, If Tekton has a scaling rate of 2x and you BGS a 20 (160 xp) but venge a 40, the expected damage is 40 and we'd pick that hitsplat.
> 
> At the end of the day no solution aside from mathing out every rate for every boss in every scale will be perfect, but this PR increases the accuracy of the plugin.

I suspect that the order of thralls/venge/weapon hitsplats is always consistent within a tick, it is probably worth investigating that first as it is a better solution than this.

InfernoStats:
After testing, it appears that thralls and venge both have an onHitsplatApplied event *before* the player's attack hitsplat. This means that the last hitsplat in a game tick on the target will be the player's true hit.

Adam-:
I think this is fixed by 5e038b726c514cfbe4380f019d120178b24edd6d

PR #15159, opened 2022-06-19
Add new `dbtable` support to cache module

Joshua-F:
This adds support for the new data structure they recently added in rev206 to the cache module. This is currently untested fully since at the time of making this there are no dbtables in the cache. I've omitted tests for the time being due to the tests running on an old cache and I'm not sure if we'd want to bump that version to one that would have dbtables in the future.


abextm:
added in 1c648fd54b67b4513033bbcb426ea0460dbcc5f5~5

PR #15162, opened 2022-06-19
blastfurnace: add smiths gloves (i) to ice gloves check

### LINKED ISSUE #15158: Blast Furnace Plugin doesn't recognize "Smith gloves (i)" as an alternative for Ice gloves.
### Bug description

https://oldschool.runescape.wiki/w/Smiths_gloves_(i)

Giant's Foundry introduced the "Smith gloves", which can be combined with "Ice gloves" to become "Smith gloves (i)", which has the same effect as "Ice gloves" for Blast Furnace. 

When you wear "Ice gloves" inside the Blast Furnace and you have bars ready, the Blast Furnace plugin will highlight the bar dispenser green. However, while wearing "Smith gloves (i)", the bar dispenser never changes to green (unless someone throws a bucket of water on the bar dispenser).

crbnFiber:
Fix #15158 by checking if Smiths gloves (i) are equipped


Vanillj:
Might be a good idea to link the issue (#15158) for automatic close later.

Issue #14672, opened 2022-02-06
Change label for Farming patches that give produce recurrently and Farming patches that have been checked


F-Lambda:
**Issue**
Currently, all Farming patches in the Time Tracking plugin are labeled "Done" when the timer is full. However, there are some patches for which this label is ambiguous. The first type of patch for which this is ambiguous is log-giving trees, as it does not indicate whether they have already been checked. The second type of patch is produce-giving patches, such as fruit trees, cacti, and bushes. For them, it is not clear whether they are ready to be checked, or if they've already been checked and have more produce to be harvested.

**Solution**
For all patches that give a *Check* interact prompt when they are fully grown, the text will change to say "Checked" once they have been checked. Futher, for patches that give produce, if they have already been checked and they have available produce, the text will change to say "Harvestable", even if the maximum amount of produce is not available.

**Screenshots**
![Example image](https://i.imgur.com/fSa0dp3.png)
The Catherby fruit tree patch is fully grown but not checked, while the Farming Guild patch has been checked and has some fruit available. Currently, it is ambiguous if the Catherby patch has not been checked yet, or if it has been checked and has all fruit available.

PR #15163, opened 2022-06-19
timetracking: change label for harvestable farming patches

Jallah123:
this fixes: #14672

make a visual distinction between harvestable and done


CheesyCracks:
Any notes on why this was closed?  Looking into doing similar changes.

PR #15168, opened 2022-06-20
Party extended status + overlay

LlemonDuck:
* 1392854455418ce599c01856fb356655a51e536a: Combines CharacterNameUpdate and SkillUpdate into a unified StatusUpdate packet which only transmits values that have changed. Also expands the packet to include stamina, special attack energy, and vengeance status.
* 22e54a87f879db4520208e55e5cffeb9c51a0356: Uses the StatusUpdate packets to display the above values as a player overlay: 
image

For most players, these features will probably remain off, but they are used by PvM groups. However, I have left `statusOverlayVeng` and `statusOverlayRenderSelf`, since I felt that would be a desired default for those that are using vengeance, and shows nothing without it.

Since stamina is less important and could otherwise be spammy, its precision is reduced to the nearest 10% rounded up, to reduce network traffic.


MoreBuchus:
Adding an option to stack that information vertically (minus the venge icon) would be ideal for boosters.

LlemonDuck:
Updated for protobuf party service change

InfernoStats:
Looking forward to this being merged for the venge indicators 👍

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -448,45 +457,93 @@ public void onUserJoin(final UserJoin event)
...
+		final int specEnergy = client.getVar(VarPlayer.SPECIAL_ATTACK_PERCENT) / 10;
+		final boolean vengActive = client.getVarbitValue(Varbits.VENGEANCE_ACTIVE) == 1;
 
 		final Player localPlayer = client.getLocalPlayer();
 		final String characterName = Strings.nullToEmpty(localPlayer != null && client.getGameState().getState() >= GameState.LOADING.getState() ? localPlayer.getName() : null);
 
-		if (localMember != null)
+		boolean shouldSend = false;
+		final StatusUpdate update = new StatusUpdate();
+		if (forceSend || !characterName.equals(lastStatus.getN()))
```
Adam-, line 490: this NPEs

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyStatusOverlay.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+		
+		return null;
+	}
+	
+	private PartyMember findPartyMember(List<PartyMember> partyMembers, Player p)
+	{
+		for (PartyMember partyMember : partyMembers)
+		{
+			if (partyMember.isLoggedIn() && p.getName().equals(partyMember.getDisplayName()))
```
Adam-, line 134: this also NPEs

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -373,18 +377,47 @@ void requestSync()
 	}
 
 	@Subscribe
-	public void onCharacterNameUpdate(final CharacterNameUpdate event)
+	public void onStatusUpdate(final StatusUpdate event)
```
Adam-, line 380: This should probably drop status updates for self since it will race if multiple updates are sent out before one is received

Adam- (changes requested):
the commits should be broken up as
1) consolidate the existing messages into one
2) add the new overlay+ new fields to the existing message

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -448,45 +457,93 @@ public void onUserJoin(final UserJoin event)
...
+		if (shouldSend)
+		{
+			party.send(update);
 		}
 
-		lastHp = currentHealth;
-		lastPray = currentPrayer;
-		lastCharacterName = characterName;
+		// non-null values for next-tick comparison
+		lastStatus = new StatusUpdate(
```
Adam-, line 537: this can be dropped if !shouldSend

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyStatusOverlay.java
```diff
@@ -0,0 +1,173 @@
...
+				continue;
+			}
+
+
+			graphics.setFont(OVERLAY_FONT);
+			if (renderHealth)
+			{
+				double healthRatio = Math.min(1.0, (double) partyData.getHitpoints() / partyData.getMaxHitpoints());
+				Color healthColor = ColorUtil.colorLerp(COLOR_HEALTH_MIN, COLOR_HEALTH_MAX, healthRatio);
+				renderPlayerOverlay(graphics, player, String.valueOf(partyData.getHitpoints()), healthColor, -20, -20);
```
Adam-, line 107: It would be neat to draw these with offsets similar to hitsplats where the offsets depend on number of hitsplats

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -145,9 +149,8 @@ public class PartyPlugin extends Plugin
 	private PartyPanel panel;
 	private NavigationButton navButton;
 
-	private int lastHp, lastPray;
-	private String lastCharacterName = "";
 	private WorldPoint lastLocation;
+	private StatusUpdate lastStatus;
```
Adam-, line 153: this probably needs to be nulled at shutdown or else the plugin doesn't restart correctly

runelite-client/src/main/java/net/runelite/client/plugins/party/messages/StatusUpdate.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+public class StatusUpdate extends PartyMemberMessage
+{
+	
+	// we use boxed fields to null out properties that haven't changed
+	// and minified field names to reduce the payload size
```
Adam-, line 17: you can use `@SerializedName` to achieve this without having to mangle the source as much

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
LlemonDuck: `characterName` is guarded by the above `Strings.nullToEmpty`, and `lastStatus` is guarded by `forceSend`

PR #15169, opened 2022-06-20
checkstyle: no trailing whitespace on empty lines

LlemonDuck:
Expands the existing CheckStyle rule which detects trailing whitespace to also apply to blank lines.

Could also be split out with something like 
```xml
<module name="RegexpSinglelineJava">
	<property name="format" value="^\s+$"/>
	<property name="message" value="Empty lines should not contain whitespace"/>
	<property name="ignoreComments" value="true"/>
</module>
```

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -41,7 +41,7 @@
 			<property name="ignoreComments" value="true"/>
 		</module>
 		<module name="RegexpSinglelineJava">
-			<property name="format" value="[^\s]\s+$"/>
+			<property name="format" value=".*\s+$"/>
 			<property name="message" value="No trailing whitespace"/>
 			<property name="ignoreComments" value="true"/>
 		</module>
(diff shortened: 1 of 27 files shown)


checkstyle.xml
```diff
@@ -41,7 +41,7 @@
 			<property name="ignoreComments" value="true"/>
 		</module>
 		<module name="RegexpSinglelineJava">
-			<property name="format" value="[^\s]\s+$"/>
+			<property name="format" value=".*\s+$"/>
```
Nightfirecat, line 44: Is this regex run against the whole line, or does it match substrings? eg. does `\s+$` work to match whitespace at EOL?

LlemonDuck: This matches substrings. I expanded the original regex to work on both lines with content `[^\s]\s+$` and without content `.*\s+$`.

However, this rule would work if we wanted two separate rules for this instead:
```xml
<module name="RegexpSinglelineJava">
	<property name="format" value="^\s+$"/>
	<property name="message" value="Empty lines should not contain whitespace"/>
	<property name="ignoreComments" value="true"/>
</module>
```

PR #15176, opened 2022-06-21
Party: Auto-join last party session upon logon

redrails:
It can be a little bit tedious to rejoin party when logging in, this should automatically allow to rejoin the previously joined party in the config upon login.


Adam-:
We don't want to have an auto-join party feature because there is limited capacity on the party server and don't want a bunch of people unknowingly sitting in parties that they've fogtten about.

PR #15194, opened 2022-06-22
Remove deprecated enums

tygyh:
Remove unused enum entries


Nightfirecat:
These enum values are still used in a number of hub plugins and should not be removed. See: <https://abextm.github.io/plugin-hub/#WyJeTG5ldC9ydW5lbGl0ZS9hcGkvTWVudUFjdGlvbjsuSVRFTV9VU0VfT05fR0FNRV9PQkpFQ1QiLCIiXQ==>, and search for the various removed entries to see how they're used there.

PR #15195, opened 2022-06-22
Anti drag cox chest

MoreBuchus:
This will update the current anti drag plugin to include the CoX private storage units.

cox chest


runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
```diff
@@ -202,6 +206,19 @@ else if (widgetLoaded.getGroupId() == WidgetID.INVENTORY_GROUP_ID)
 		}
 	}
 
+	@Subscribe
+	public void onItemContainerChanged(ItemContainerChanged event)
+	{
+		//Chambers of Xeric storage units reset delay everytime ItemContainerChanged is fired
+		if (client.getVarbitValue(Varbits.IN_RAID) == 1)
```
Adam-, line 213: Flip these around because it is faster
```suggestion
		if (event.getContainerId() == InventoryID.CHAMBERS_OF_XERIC_PRIVATE_CHEST.getId() && client.getVarbitValue(Varbits.IN_RAID) == 1)
```

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
Adam-: You can also just delete this entirely and add a 2nd check in `onScriptPostFired` for `[proc,raids_storage_private_items](component $component0, component $component1, component $component2)` ID 1607

PR #15207, opened 2022-06-24
Linux : Force frame to front in clientUI.forceFocus()

alexandreLamarre:
Without `frame.toFront()` the `forceFocus()` method focuses the window, but it remains hidden behind other windows (on Arch linux / Manjaro & Ubuntu 18) which doesn't necessarily match what I think `forceFocus()` should do.

For Linux, I believe `clientUI.forceFocus()` implementation details should match the use case of forcing focus. I used as reference : https://stackoverflow.com/questions/641172/how-to-focus-a-jframe , but believe `setVisible(true)` is not necessary in this case.

Let me know if this change in `forceFocus()` on linux matches the intended usecase!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -760,6 +760,7 @@ public class ClientUI
 				WinUtil.requestForeground(frame);
 				break;
 			default:
+				frame.toFront();
 				frame.requestFocus();
 				break;
 		}


Adam-:
Indeed the expected behavior is it brings the window to front. Though I don't have a linux machine to test this.

LlemonDuck:
This did not work on my Linux Mint 20.3 system

joshwc:
Adding:
```
case Linux:
				frame.setVisible(true);
				frame.toFront();
				frame.requestFocus();

				break;
```
to `forceFocus()` solves the issue of the window not being brought to the front. `frame.setVisible(true);` does seem to be necessary, at least in Mint 21.3.

Funkylord:
where did you input this code?

coopermor:
As I told you yesterday, "Unless you want to build the client from source you can't put it anywhere"

But if you click the Files changed tab it shows you the file that was changed and the line number

Funkylord:
Oh I see and I apologize for yesterday my mindset was not in a good place at the time.  So in the files changed tab that's what I copy into the config file on my end on the corresponding line numbers?

coopermor:
You cannot put this into any config. You would have to build the client from source to incorporate the changes.

PR #15212, opened 2022-06-25
special counter: Add special weapon null check

Nightfirecat:
`specialAttackHit()` expects all its arguments to be non-null, so the
passed special weapon must be null-checked before being passed to it.
This fixes a frequent NPE caused by using any special attack weapon not
defined in the SpecialWeapon enum.


Adam-:
fixed by 3bbfdbbfb0ae9d642fa2585a00dd53e656894d90

PR #15249, opened 2022-07-01
Compact party overlay

MoreBuchus:
The following is what this PR would change:
+ Adds an option to display the party info (hp, prayer, stam, spec) with smaller font, stacked, and offset from the player as to avoid being covered by hitsplats/hp bar
+ Changes the color used to identify special attack percent to yellow, as more players associate special attack with energy transfer and having multiple components of the overlay set to green is confusing
+ Changes the overlay layer to under widgets to show hitpoints above hitsplats when using the diamond shape

java_BWfVgM9anm


Adam-:
I think I'd rather just replace the existing overaly with this one, since it looks cleaner, and I think the original overlay was based off of something you provided to us originally anyway? cc @LlemonDuck

LlemonDuck:
I'm fine with replacing it as well, I think this is easier to read. Diff looks good to me as well.

PR #15269, opened 2022-07-04
[cache] ModelLoader renaming fields of decoders

Shadowrs:
Renaming local fields based on public tools which were already renamed. new fields between types new have an underscore _ in the name


Shadowrs:
Found better naming scheme to use to commit in future- closing request

PR #15271, opened 2022-07-05
Add various worldmap nodes to Necropolis/Ullek region

fuzxi:
Let me know if anything needs changing, this is my first PR.

The northern shortcut is listed as stepping stone (singular) because there's only one stone, while the southern one has two.


Nightfirecat:
Added in d0cc01f88c5b87639f26f38e7622c201d2324b4e and 2486bca73a9d02af3d1063520087504b9c5ac5e5. Your points were all off by 1 tile. Be sure to use the world map location tool (particularly the centered blue lines drawn over the map) to get the exact location of world map icons.

PR #15319, opened 2022-07-11
GameTimer.java

raialb:
change timer to 200 seconds for Anti-venom+ for better overlay accuracy


Adam-:
Is the anti-venom timer inaccurate? It currently uses varbits for the timer which should be extremely precise.

PR #15321, opened 2022-07-11
menu swapper: override non-custom item swaps with custom swaps.

geheur:
This is motivated by people wanting to use the non-custom "Teleport" swap while also left-click "Wear"ing the mythical cape.


geheur:
Closing because this can now be done by hub plugins e.g. custom menu swaps, and it has conflicts.

PR #15345, opened 2022-07-14
skill calculator: per-action bonus applicability

LlemonDuck:
Replaces `SkillBonus#isIgnoreBonus` with `SkillBonus#isBonusApplicable(SkillBonus)` in order to support bonuses that affect certain skill actions and not others.

At the moment this only applies to PrayerAction, and I included the Arceuus offering spells as well as the Sacred Bone Burner from #15232.

Closes #15231
Closes #15232


Nightfirecat:
I haven't yet reviewed this, but I wanted to mention that another feature I'd want to eventually see added is the ability to toggle multiple skill bonuses in a sort of checkbox + radio button style UI. (For instance, you can use both zealot robes and chaos altar, with the effects stacking. As of now, prayer is the only skill where this would apply.) I'll likely review this PR with that in mind--that we'll want to be able to apply multiple bonuses in the long term.

LlemonDuck:
That should be reasonably compatible with this PR by keeping a set of active bonuses instead of only a current bonus, then combining the applicable factors per-action.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/RunecraftAction.java
```diff
@@ -74,4 +74,10 @@ public enum RunecraftAction implements ItemSkillAction
 	private final int level;
 	private final float xp;
 	private final boolean ignoreBonus;
+
+	@Override
+	public boolean isBonusApplicable(SkillBonus bonus)
```
Nightfirecat, line 79: Not related to this PR, but I noticed that mind, body, and chaos cores incorrectly do not have `ignoreBonus = true`. Can you add a commit which fixes that as well? Thanks.

Nightfirecat (approved):
LGTM, pending RC core fix

LlemonDuck:
> LGTM, pending RC core fix

Done

PR #15357, opened 2022-07-16
Fix Dev Tools: Tile Location Tooltip in Instanced Regions

damencs:
- properly display the region id depending on whether it's instanced or not

image


Nightfirecat:
Added in ad949e6ff4295231df2b1064d85f1bba0e92ab8b

PR #15372, opened 2022-07-18
Agility: add true tile highlighting to Hallowed Sepulchre projectiles

rislah:
Adds true tile highlighting to Hallowed Sepulchre projectiles


kentor:
Just FYI this can be achieved with Better NPC highlights and using the IDs 9669,9672,9673,9670,9671,9674

I've tried this and found that true tiles don't really help that much

PR #15374, opened 2022-07-19
Timers: Improve tracking

Macweese:
This PR aims to move away from using `chatMessage` events and instead rely on vars.

Includes improvements for
 - Charge spell for God spells
 - Teleblock
 - Bladed God staves
 
Removed the need to keep track of widgets/other indicators for recognizing PvP zones for teleblock purposes; The varbit associated with teleblock does this.

Added support for Staff of Balance as it too has the same protective effect.
Fix bug associated with bladed God staves; re-equipping after unequipping a bladed God staff within the effect duration should re-introduce timer as the protective effect hasn't absolutely waned but is made invalid in absence of the weapon.

I could be mistaken on the clearing of effects on death in `GameTimer`.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -38,10 +38,6 @@ import net.runelite.api.AnimationID;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
-import net.runelite.api.EquipmentInventorySlot;
-import net.runelite.api.InventoryID;
-import net.runelite.api.Item;
-import net.runelite.api.ItemContainer;
 import net.runelite.api.ItemID;
 import static net.runelite.api.ItemID.FIRE_CAPE;
 import static net.runelite.api.ItemID.INFERNAL_CAPE;
@@ -58,7 +54,6 @@ import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.GraphicChanged;
-import net.runelite.api.events.ItemContainerChanged;
 import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.NpcDespawned;
 import net.runelite.api.events.VarbitChanged;
@@ -133,7 +128,6 @@ public class TimersPlugin extends Plugin
 
 	private TimerTimer staminaTimer;
 
-	private int staffProtectionExpireTick = -1;
 	private int lastRaidVarb;
 	private int lastVengCooldownVarb;
 	private int lastIsVengeancedVarb;
@@ -660,7 +654,6 @@ public class TimersPlugin extends Plugin
 
 		if (config.showStaffOfTheDead() && message.contains(STAFF_OF_THE_DEAD_SPEC_MESSAGE))
 		{
-			staffProtectionExpireTick = client.getTickCount() + 100;
 			createGameTimer(STAFF_OF_THE_DEAD, Duration.of(100, RSTimeUnit.GAME_TICKS));
 		}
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/timers/TimersPluginTest.java
```diff
@@ -204,28 +204,28 @@ public void testTransparentChatboxTbRemoved()
 	public void testMageArena2TbFull()
 	{
 		when(timersConfig.showTeleblock()).thenReturn(true);
-		ChatMessage chatMessage = new ChatMessage(null, ChatMessageType.SPAM, "", "<col=c356ef>A Tele Block spell has been cast on you. It will expire in 2 minutes.</col>", "", 0);
-		timersPlugin.onChatMessage(chatMessage);
+		when(client.getVarbitValue(Varbits.TELEBLOCK)).thenReturn(300);
```
Nightfirecat, line 207: There's no need to keep all the teleblock tests now that it doesn't have to parse numerous text variants. You can keep a minimal number to test the various varbit value tiers or just remove them entirely.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -1069,13 +1036,20 @@ public void onItemContainerChanged(ItemContainerChanged itemContainerChanged)
...
-				weapon.getId() != ItemID.TOXIC_STAFF_UNCHARGED))
+		if (weapon != null &&
+				(weapon.getId() == ItemID.STAFF_OF_THE_DEAD ||
+						weapon.getId() == ItemID.TOXIC_STAFF_OF_THE_DEAD ||
+						weapon.getId() == ItemID.STAFF_OF_LIGHT ||
+						weapon.getId() == ItemID.STAFF_OF_BALANCE ||
+						weapon.getId() == ItemID.TOXIC_STAFF_UNCHARGED) &&
+				staffProtectionExpireTick > client.getTickCount())
+		{
+			// Recreate timer with remaining duration if a protective staff is re-equipped within 1 minute since activation
```
Nightfirecat, line 1047: This code existed because our previous understanding of the special effect was wrong--we believed that unequipping the staff caused the effect to end rather than to stop working only when it was unequipped. Instead of removing/recreating the timer, just leave it present at all times regardless of whether a sotd variant is equipped.

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
```diff
@@ -227,6 +227,13 @@
 	 */
 	LAST_HOME_TELEPORT(892),
 
+	/**
+	 * Charge spell
+	 * 0 < Value * 2 = Remaining ticks on charge buff for God spells
```
Nightfirecat, line 232: Wut? Is this meant to convey that values above 0 are double the number of remaining ticks? ie. `10` represents 5 remaining ticks? This is not clear at all.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: We discussed this on Discord and decided we'd like the infobox to remain visible even if a staff is unequipped, though it'd be nice to make the box's image different when a staff is not equipped--partially transparent perhaps. Currently, the infobox API doesn't offer a way to change the image--to achieve that effect, it would need to be removed and added again with a changed image. We are also trying to move away from code which removes and re-adds infoboxes since that causes reordering of other active infoboxes, so some API changes are needed to make this happen the way we'd like.

Nightfirecat:
I've pushed f647188612eb925c0188172b01e76eaedcf9a5af and 461f8471d68cf653fae9e9efb3d32b7dc5e474a7 for charge spell and teleblock duration tracking since those are done fully. I'd like to do that infobox image API work prior to implementing the sotd (and variants) change. If I don't get around to that, [this commit](https://github.com/Nightfirecat/runelite/commit/8f9e660122f7259f03ba91845dedf0a3267e571a) should basically cover the proposed changes from this PR.

Nightfirecat:
Note to self (or for any other devs who might implement this), the wiki is in fact correct, that unequipping a staff of the dead does not end the special attack's effect--it simply does not activate during that minute time unless a staff of the dead (or one of its variants) is equipped at the time of the received melee attack.

PR #15387, opened 2022-07-22
Add autoJoin ConfigItem in the party plugin

JacobOdell:
Small change that adds a "Auto join previous party" checkbox in the party plugin configuration that automatically joins your previous party on client launch.


Adam-:
We don't want to have an auto-join party feature because there is limited capacity on the party server and don't want a bunch of people unknowingly sitting in parties that they've forgotten about.

PR #15390, opened 2022-07-23
Updated PrayerDoseOverlay with 2 more colour options 

Mantic-says-hi:
This is to assist people who are colour blind.
img


Adam-:
The configuration supports Color configuration, which just opens a color picker, so you could just use that to make the start/end colors configurable.

Mantic-says-hi:
> The configuration supports Color configuration, which just opens a color picker, so you could just use that to make the start/end colors configurable.

Yea sure I can change that if you would rather it be that way. 

The goal for this update was to make the plugin more accessible to colour blindness instead of full customization. 

I just figured it might be better from a design perspective just to have a few options that work well than letting a user pick a start/end and making a custom start end might even confuse people/cause them to make goofy start ends.
It might not be intuitive for some people to have one be bright and the other be dull, etc..

PR #15393, opened 2022-07-23
Added duplicate SpriteID's to inform about usecase.

PatrykSitko:
BANK_BACKGROUND and BANK_PIN_BACKGROUND corresponds to: DIALOG_BACKGROUND;
But was unclear during development.


Nightfirecat:
There's no reason to duplicate this sprite definition. Just use the name which is already defined.

PR #15395, opened 2022-07-23
ground-markers: Improve minimap overlay render clipping

Skretzo:
Use the minimap sprite as a graphics clip mask instead of using a static circle cutoff inside the minimap widget rectangle.  
before and after


Skretzo:
Superseded by #19271

Issue #15400, opened 2022-07-25
[Item Identification] Barrows armor 100/75/50/25 is not displayed


MESLewis:
### Type

Incorrect behavior

### Bug description

Barrows armor uses different items for 100/75/50/25 but the item identification plugin has no way to differentiate them.

PR #15403, opened 2022-07-25
Add Barrows Equipment Identifiers

MikeMillar:
* Add Barrows Equipment Identifiers for different levels of degradation.
* Add new identification type "Equipment".
* Add Item Identifcation Config method "showEquipment".
* Configure Item Identification Overlay to show on equipment and inventory equipment.

Issue [#15400](https://github.com/runelite/runelite/issues/15400)


Hydrox6:
I don't think there's a use for this showing what type of armour it is. They're visually distinct enough that just showing the durability tier would be fine (and much smaller both in the UI and in the code)

runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentificationConfig.java
```diff
@@ -325,4 +325,15 @@ default boolean showEnchantedJewellery()
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "showEquipment",
+			name = "Equipment",
+			description = "Show identification on equipement",
```
discomethod, line 332:
```suggestion
			description = "Show identification on equipment",
```

Nightfirecat:
The linked issue was not really clear about how this should be implemented and incorrectly pointed to the Item Identification plugin as where it should be done when the Item Charges plugin makes much more sense.

PR #15407, opened 2022-07-27
Add isDying override for Skotizo's Awakened altars

### LINKED ISSUE #15405: Skotizo's Awakened Altars are hidden by Entity Hider after being downed once
### Discussed in https://github.com/runelite/runelite/discussions/15404

<sup>Originally posted by **jonr1996** July 26, 2022</sup>
Could a fix be added to not hide the Awakened Altars inside of the Skotizo boss room. Currently the altars become unclickable once they've been downed once. </div>

These altars should not be hidden at all, and should have an exception added in `NpcUtil#isDying()`.

ColeSpringer:
Fixes #15405 by adding the four AWAKENED ALTARs to isDying


Nightfirecat:
How exactly do awakened altars work? Do they spawn and remain spawned (but transform into an uninteractable state when dead) for the duration of the fight, or do they despawn and respawn as already-dead NPCs? I'd like to be sure of how they work before putting them under the `// These NPCs hit 0hp but don't actually die` section.

Nightfirecat:
I've added this in 177a720cfc0a40ce71e5bdd2212f3e0ad0912be7. Given that they can continually change forms from Altar -> Awakened Altar -> Altar etc., utilizing the `setDead(false)` code path felt like the better solution here.

PR #15413, opened 2022-07-28
Fix special attack plugin to track npc targets of special attacks to line up with hitsplats

osrscommits:
Some followup from https://github.com/runelite/runelite/pull/15338 and https://discord.com/channels/301497432909414422/419891709883973642/1001959649802727426.

As recap, hitsplats would be misattributed when doing a long-range attack that takes `<weapon_cooldown>` ticks to land followed up a special attack (most common by using a trident on olm mage hand followed by bgs on melee hand). This fixes that issue by tracking what NPC the player is interacting with when the special attack occurs, and ensuring that it only looks at hitsplats on that NPC, filtering out the rest.

I didn't test this in any raids, but reproduced the issue on cyclops in catacombs (long range one with trident, then swap to bgs and spec another with hitsplats on the same tick). Confirmed it repros on base runelite and is fixed with the PR.

I think this will also prevent the infobox from prematurely disappearing, which I found to be a common issue in chambers (I think due to swapping between mage and melee hand, though didn't confirm the actual reason; it had to be something that would cause `removeCounters` to be called, which I suspect was the check with `interactedNpcIndexes`); with this change, we would hit the early return from `lastSpecTarget == null`, and it shouldn't disappear until the hand dies.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -235,7 +235,14 @@ public class SpecialCounterPlugin extends Plugin
 		// invokeLater because the varbit event happens prior to interact changed, so we can't always see what npc
 		// the player is attacking yet.
 		clientThread.invokeLater(() ->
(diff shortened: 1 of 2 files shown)


osrscommits:
Tested this in both chambers and tob, specs showed as normal (nothing should change with this PR in tob, it fixes a few bugs with chambers)

Adam-:
added in ec2d5dcbae591603a32bcefc9011c7c51443c45a

PR #15414, opened 2022-07-28
Sort ground item menu entries by GE price

Barragek0:
Reorders ground item menu entries so that the most expensive items are at the top of the pile, and least expensive at the bottom.
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
@@ -47,6 +47,8 @@ import java.util.Queue;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsPlugin.java
```diff
@@ -354,7 +354,37 @@ public void onClientTick(ClientTick event)
...
+				{
+					final MenuEntry aEntry = a.getEntry();
+					final int aId = aEntry.getIdentifier();
+					final int aQuantity = getCollapsedItemQuantity(aId, aEntry.getTarget());
+					
+					final MenuEntry bEntry = b.getEntry();
+					final int bId = bEntry.getIdentifier();
+					final int bQuantity = getCollapsedItemQuantity(bId, bEntry.getTarget());
+					
+					if (config.sortByGEPrice())
```
Suravarus, line 380: Wouldn't it be better to have this check at the beginning of the sort, and return if the option isn't enabled? If the user doesn't have this option enabled, we're making some unnecessary calculations.

PR #15422, opened 2022-07-31
Adding ignore pets configuration to NPC Indicators

Grantapher:
## Summary

This change adds the option to exclude pets from NPC indicators

- on (Default) - Exclude all pets from NPC Indicators.
- off - Don't exclude any pets. This maintains parity with how NPC indicators currently works.

The reasoning for this is to exclude overloaded npc matches. For example, `* dragon` matches all dragons, but this includes `Prince black dragon`.

## Screenshots

For the actual highlighting, I added `*chaos*,*cano` to my "NPCs to highlight" list and stood next to a random pet chaos elemental with my Smolcano.

#### Off

Off

#### On

On

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsConfig.java
@@ -247,4 +247,16 @@ public interface NpcIndicatorsConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+		position = 14,
+		keyName = "ignorePets",
+		name = "Ignore pets",
+		description = "Excludes pets from being highlighted"
+	)
+	default boolean ignorePets()
+	{
+		return true;
+	}
+
 }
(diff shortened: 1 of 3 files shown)


Mark7625:
https://github.com/runelite/runelite/pull/15412

Done the same here also just noticed tho my build failed so just need to fix that rq

Grantapher:
For changing only rebuild, I don't see how a method like [`NpcIndicatorsPlugin#onNpcSpawned`](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java#L345) would avoid highlighting pets if the change isn't made to that `if`. There is no rebuild called in that method so it doesn't seem like it would ensure that we don't add a pet to the highlights when they spawn in.

As for the pet names, I believe that the change as is would remove pets from the highlights and thus the names as well. I don't see much of a case where pets are being included in highlights but we want to specifically omit their names when we can just omit them wholly.

Adam-:
added in be2581a6b0a1cc12cb896d8de75dcaff6ba9662a

PR #15460, opened 2022-08-08
Use RsProfile configuration in NPC aggro area plugin

geeckon:
Needed for the plugin to work correctly when playing multiple accounts.


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -431,10 +431,10 @@ private void loadConfig()
 
 	private void resetConfig()
 	{
-		configManager.unsetConfiguration(NpcAggroAreaConfig.CONFIG_GROUP, NpcAggroAreaConfig.CONFIG_CENTER1);
```
Adam-, line 434: this should also unset the old config values

PR #15484, opened 2022-08-12
added swaps for 'buy-x' and 'sell-x' because some places such as the …

JordanShackelford:
…nightmare zone use dash instead of space


Adam-:
I'm currently working on system for generic UI swaps which would probably make this unnecessary, however I don't think your PR works because

1) nmz has no "Value" option at all, and
2) you can't sell things to the nmz

Nightfirecat:
This is now possible after Adam's additions to the menu entry swapper plugin which added generic UI swaps.

PR #15488, opened 2022-08-13
GroundItem Plugin: Increase render distance

### LINKED ISSUE #14729: Ground item overlays don't render after a certain distance
**Describe the bug**
After a certain distance, the ground items overlay doesn't render even though I can see the item.
Im cannoning Ice Trolls and some of the trolls run away because i'm out of their range but the cannon can still reach them. This results in me missing high value loot in the corner of my screen when they die far away. 

**To Reproduce**
1) Drop an item
2) See the overlay appears
3) Run away until the overlay disappears but the item is still within range for the client to render it.

**Expected behavior**
Ground item overlay continues to render at all distances that the client makes items available.

**Screenshots**
image

Steven-Posterick:
Increase render distance and use x and y comparisons. 
Fixes #14729

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -62,7 +62,7 @@ import net.runelite.client.util.QuantityFormatter;
 
 public class GroundItemsOverlay extends Overlay
 {
-	private static final int MAX_DISTANCE = 2500;
+	private static final int MAX_DISTANCE = 25 * Perspective.LOCAL_TILE_SIZE;
 	// We must offset the text on the z-axis such that
 	// it doesn't obscure the ground items below it.
 	private static final int OFFSET_Z = 20;
@@ -193,8 +193,9 @@ public class GroundItemsOverlay extends Overlay
 		{
 			final LocalPoint groundPoint = LocalPoint.fromWorld(client, item.getLocation());
 
-			if (groundPoint == null || localLocation.distanceTo(groundPoint) > MAX_DISTANCE
-				|| (onlyShowLoot && !item.isMine()))
+			if (groundPoint == null || Math.abs(localLocation.getX() - groundPoint.getX()) > MAX_DISTANCE
+					|| Math.abs(localLocation.getY() - groundPoint.getY()) > MAX_DISTANCE
+					|| (onlyShowLoot && !item.isMine()))
 			{
 				continue;
 			}


Nightfirecat:
Is there a maximum interaction distance with ground items? Why have you chosen 25 tiles as the render distance for this?

Skretzo:
The interaction distance with ground items is bigger than 25 tiles, and 25 tiles seems to be the vanilla black fog render distance

Steven-Posterick:
Yeah, I used 25 as it was the cap for the GPU plugin. 
The 117 HD plugin looks like it allows a higher distance, so I can also set it to the variable from the GPU config.

Nightfirecat:
What you're doing now does not make sense. (how does 117hd have anything to do with this?)

If removing the distance check would cause item name overlays for invisible items (beyond GPU render distance, or beyond vanilla 25-tile distance) then check the rendered distance and use that as the cap distance.
If it does not cause overlays for invisible items, then remove the check completely instead.

Steven-Posterick:
I've added a commit to use the draw distance. 
What I was bringing up was that even if you set 90 as the draw distance for the GPU plugin, it only allows up to 25 tiles to be drawn. In contrast, the 117HD allows the whole 90 tiles to be drawn.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -192,9 +191,9 @@ public Dimension render(Graphics2D graphics)
 		for (GroundItem item : groundItemList)
 		{
 			final LocalPoint groundPoint = LocalPoint.fromWorld(client, item.getLocation());
-
-			if (groundPoint == null || Math.abs(localLocation.getX() - groundPoint.getX()) > MAX_DISTANCE
-					|| Math.abs(localLocation.getY() - groundPoint.getY()) > MAX_DISTANCE
+			final int localDrawDistance = client.getScene().getDrawDistance() * Perspective.LOCAL_TILE_SIZE;
```
KieranDevvs, line 194: Probably want to declare this outside the loop as it *should* be the same for all items.

KieranDevvs:
Checked out the code, made the changes I suggested:

image

Looks like its working as expected.
@Nightfirecat what's holding this back?

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
Steven-Posterick: Thanks for the suggestion! I've added in a commit for that.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
```diff
@@ -188,13 +187,14 @@ public Dimension render(Graphics2D graphics)
 		final boolean onlyShowLoot = config.onlyShowLoot();
 		final DespawnTimerMode groundItemTimers = config.groundItemTimers();
 		final boolean outline = config.textOutline();
+		final int localDrawDistance = client.getScene().getDrawDistance() * Perspective.LOCAL_TILE_SIZE;
```
Nightfirecat, line 190: There's one other change I would recommend for this, which would be to limit the scene distance to 50, as that's the maximum distance we draw clickboxes for items.

```suggestion
		final int localDrawDistance = Math.min(client.getScene().getDrawDistance(), 50) * Perspective.LOCAL_TILE_SIZE;
```

Nightfirecat (commented):
Works great, needs one small change to be good to go in my eyes.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
KieranDevvs: @Steven-Posterick are you ok to make the change? I would do it but it would require opening a new PR.

Steven-Posterick:
> Works great, needs one small change to be good to go in my eyes.

Looks good, accepted.

runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
Steven-Posterick: > @Steven-Posterick are you ok to make the change? I would do it but it would require opening a new PR.

Sure, accepted it now.

KieranDevvs:
@Nightfirecat this is ready for a 2nd pass, thanks.

jonathangarelick:
@Nightfirecat A Discord user wanted a status update on this. Seems good to go?

mike-ul:
I would love to see this get added soon.

KieranDevvs:
> I would love to see this get added soon.
Its been almost 2 years, don't hold your breath.

PR #15499, opened 2022-08-15
Item Stats: Added option to show item stats only while bank interface is open

Boredska:
Added option to show item stats only while bank interface is open.

Example of what this does:

java_HAchtpwMrd

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatConfig.java
@@ -113,6 +113,16 @@ public interface ItemStatConfig extends Config
 	}
 
 	@ConfigItem(
+			keyName = "showStatsOnlyInBank",
+			name = "Show Stats Only In Bank",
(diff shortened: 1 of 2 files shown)


Nightfirecat:
I can see this being useful for equipment, but this would also apply to consumables, which players typically want shown always. With that in mind I think it makes more sense for users to enable equipment stats on occasion than it does to have an equipment-specific setting to be shown while the bank is open.

Boredska:
I suppose a toggle could be added to "Always show stats on consumables."
Alternatively a change I can think of is to change this boolean option to enum option:

"While not banking"

- Default - show both
- Hide consumables stats
- Hide equipment stats
- Hide both

Which should I do?

Nightfirecat:
Discussed a bit in Discord, but didn't have any conclusions beyond what's already been said here. As an alternative to integrating this as a core change you can always create a local build with these changes, or you could create a hub plugin which toggles this config when visiting/leaving a bank.

Boredska:
Created a plugin hub version of this. https://github.com/runelite/plugin-hub/pull/3193

PR #15515, opened 2022-08-17
npcaggro: half-calibration nonsense

Hydrox6:
Currently it's possible to half-calibrate the plugin, getting only one of the boxes working. If you half calibrate it, log out, then log back in with the plugin active (i.e. in an area with an NPC its tracking, or with Always active on), it'll throw an error every game tick.

The first commit protects against the error directly: `notifyOnce` wasn't being cleared when it should have.
The second commit protects against half-calibrations by not showing the scene overlay and keeping the tutorial overlay visible until both centres are set.

Hopefully this will also make less users confused about the calibration. *Hopefully*

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
@@ -66,7 +66,7 @@ class NpcAggroAreaOverlay extends Overlay
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		if (!plugin.isActive() || plugin.getSafeCenters()[1] == null)
+		if (!plugin.isActive() || plugin.getSafeCenters()[0] == null || plugin.getSafeCenters()[1] == null)
 		{
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaOverlay.java
```diff
@@ -66,7 +66,7 @@ private NpcAggroAreaOverlay(Client client, NpcAggroAreaConfig config, NpcAggroAr
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		if (!plugin.isActive() || plugin.getSafeCenters()[1] == null)
+		if (!plugin.isActive() || plugin.getSafeCenters()[0] == null || plugin.getSafeCenters()[1] == null)
```
Nightfirecat, line 69: I don't think this line should be changed. It is helpful to know when and where the first center is placed even before the second center is calibrated. The warning overlay can (and should) still be shown until the second center is set though.

Hydrox6: Thing is, with how it currently works, it won't always show where the first centre is; it only happen if for some reason, the second is calibrated first. I think making it consistently wait until both are set is better than sometimes showing information to the user which is wrong (since a half-calibration will not save)

PR #15517, opened 2022-08-18
DPS counter: add option to sort by damage

bspammer:
Allow automatic sorting of DPS values while fighting a boss, especially useful when in a decent-sized party so you can see who's going to get the drop.


StephanBijzitter:
Would be great to have this, having around 10 people in your party (for example at nex) makes the list of users with their damage values almost useless; sorting by damage would help a lot to see who's doing well (top of list) and who's doing poorly (bottom of list)

bspammer:
@StephanBijzitter agreed, unfortunately I don't know how to ask for a review

charlie-collard:
@Adam- sorry to tag, are you the right person to ask for review on this?

PR #15524, opened 2022-08-18
timers: use varbit to track remaining overload time

### LINKED ISSUE #12848: overload timer in chambers of xerics will disappear if hp is 0 for a tick
if you do 3hp rope skip in cm chambers you will lose overload timer but enhance timer stays.

https://streamable.com/89cpbq

there is chance that hp dips down to 0 and makes overload timer disappear since game thinks you died.

Enriath:
At some point, varbits appeared that track how many stat refreshes the player has from overloads. This means we use how antipoisons are tracked and make the overload tracking more accurate, and being able to rely on the varbits to clear the timers means they won't disappear when people do advanced cheese in Chambers! This varbit could also be used to add a brand new Overload Stat Restore timer, and I was going to include it here but couldn't figure out an icon to use.

Could close #12848, though that issue also shows the same issue happens with Prayer Enhances. I have a similar change for those nearly ready, though less information exists on the non (+) tiers, making it more difficult to check without actually making some.


Nightfirecat:
@Enriath I've rebased this PR with the post-VarbitChanged API updates. Can you confirm that this is still correct? I'd like to get this in before finishing work on #16074

PR #15525, opened 2022-08-19
Random Events: Add alert for keeping specific herbs for Jekyll

AmrikSD:
Adds a small config item for the Random events plugin:
- allows you to select which reward you want from Jekyll & Hyde
- it then alerts you when you remove it from your inventory.


Nightfirecat:
This feels like it would be better-suited as a hub plugin than a core plugin, as the vast majority of players will not care to keep a specific herb in their inventory for the chance at offering it to Dr. Jekyll

AmrikSD:
> This feels like it would be better-suited as a hub plugin than a core plugin, as the vast majority of players will not care to keep a specific herb in their inventory for the chance at offering it to Dr. Jekyll

That's fair enough - I made the default behaviour to do nothing

Issue #15478, opened 2022-08-11
Tombs of Amascut update checklist


Nightfirecat:
[August 24th](https://twitter.com/OldSchoolRS/status/1550472341435588608) will bring with it the new raid Tombs of Amascut, and with that new content we'll need to make a number of updates to the client to accommodate it. This issue is to be used to track any expected or known features and bugs related to this update.

- [x] Tombs of Amascut regions (and its surrounding regions) should be added to the Discord plugin to display accurate rich presence status (c1d98ebff7c72ac93139bd01eb79a1c5ef028d9f)
- [x] ~~Tombs of Amascut regions may need to be added to the Skybox plugin if a non-black skybox would be appropriate for them~~ (seems to be fine with no changes)
- [x] The dungeon icon for the raid on the world map, and any other added icons, should get the usual World Map plugin treatment with their corresponding names (fd932639f696316f4e17f31f6fd3852a5f390c7f)
- [x] Depending on the implementation of the [Lightbearer](https://oldschool.runescape.wiki/w/Lightbearer), the special attack regeneration indicator of the Regen Meters plugin may need to be updated to continue working correctly (15ca8132a0f8f166f75c928b23d2797749c74f60)
- [x] Any untradeable rewards (or variants of rewards, eg. [Elidinis' ward](https://oldschool.runescape.wiki/w/Elidinis%27_ward)) should have their tradeable variants added to the ItemMapping enum so their prices are accurately shown in the bank and other interfaces (53c4bfedcce7acc3903cf2407a0f45a8d90fbe4d)
- [x] [Stat renewal potions](https://oldschool.runescape.wiki/w/Stat_renewal) will need to be added to the Skill Calculator plugin, and have the restore effect added to the Item Stats plugin and Timers plugin (d083f63f1b07f5711e4d11a35dda06cae5b5365f and 597aa299a5d1dc434b3f6524df43678ace59de32)
  - [x] timers plugin implementation is pending after release
- [x] The Tombs of Amascut login screen background should be added to the Login Screen plugin (#15573)
- [x] ~~Bosses with transformations may need to be handled via NpcUtil to avoid becoming uninteractable via Menu Entry Swapper/Entity Hider plugins~~ (appears to work fine with no changes)
[...]

PR #15556, opened 2022-08-25
RegenMeter: Add support for Lightbearer

Macweese:
Lightbearer ring has a unique mechanic which halves the duration cycle and resets the regeneration cycle for special attacks when equipped or unequipped. This PR adds support for the item's functions to the Regen Meter plugin.

Ref: #15478


Nightfirecat:
Do you have a Lightbearer to test this with? I suspect it does not work the way you've coded it here. I would expect that you cannot equip it just before 25 spec regen ticks elapses to get 10% spec regen faster.

PR #15600, opened 2022-09-01
OpponentInformation: Update boss health bars to use OpponentInformation display style config

QuestingPet:
This updates the health bars for bosses (e.g. CoX, ToA, etc.) to allow for percents (or both HP and percent) based on the OpponentInformation plugin's config. 

This is currently very useful in ToA, where Jagex only shows the hitpoints of each boss, and not their percentages. Having the Opponent Information overlay also show percentages is useful for bosses that proc specials at certain thresholds (lots of them). Right not he RuneLite one *only* shows percentages, and is only visible if the Jagex one is disabled.

ToB uses its own health bar widget, and so is not updated here. It already uses percents by default.

Tested in all 3 raids, and CoX and ToA correctly show hp, percent, or both depending on display style.

https://cdn.discordapp.com/attachments/419891709883973642/1014695648454529194/unknown.png

https://cdn.discordapp.com/attachments/419891709883973642/1014697509416861727/unknown.png


runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -171,4 +186,72 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
...
+		}
+
+		final int currHp = client.getVarbitValue(Varbits.BOSS_HEALTH_CURRENT);
+		final int maxHp = client.getVarbitValue(Varbits.BOSS_HEALTH_MAXIMUM);
+
+		final String text;
+		switch (config.hitpointsDisplayStyle())
+		{
+			case HITPOINTS:
+				text = getHpText(currHp, maxHp);
```
Adam-, line 232: this is the default value so you can just delete this

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -171,4 +186,72 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
...
+				break;
+			case PERCENTAGE:
+				text = getPercentText(currHp, maxHp);
+				break;
+			case BOTH:
+				text = getHpText(currHp, maxHp) + " (" + getPercentText(currHp, maxHp) + ")";
```
Adam-, line 238: `getHpText(currHp, maxHp)` is the current value so you can just delete this and replace it with `widget.getText()`

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -171,4 +186,72 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
...
+				break;
+			default:
+				throw new IllegalStateException("Unexpected HitpointsDisplayStyle: " + config.hitpointsDisplayStyle());
+		}
+
+		widget.setText(text);
+	}
+
+	private String getHpText(int current, int maximum)
```
Adam-, line 247: delete this

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
```diff
@@ -171,4 +186,72 @@ public void onMenuEntryAdded(MenuEntryAdded menuEntryAdded)
...
+			return;
+		}
+
+		clientThread.invoke(this::updateBossHealthBarText);
+	}
+
+	@Subscribe
+	public void onConfigChanged(ConfigChanged event)
+	{
+		if (!event.getKey().equals("hitpointsDisplayStyle"))
```
Adam-, line 204: Its best to check the config group, too

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
QuestingPet: See comment above

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
QuestingPet: See comment above

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java
Adam-: This edge case is probably not worth handling if it requires this much code

PR #15609, opened 2022-09-02
puzzle solver: add setting to increase the amount of steps displayed

NotBlackice:
I know that displaying more than the current 3/4 steps can result in overlap, but I figured it would be nice to give players the option to have more. I had to cap it at 5 because going any higher results in a negative value for the dot size and I was not comfortable changing any of that code as I'm extremely new to this. I _did_ test changing `DOT_MARKER_SIZE - i * 3` from a 3 to a 2 which allows up to 7 steps to be displayed, but wanted to start with just 5 steps for now.


pwatts6060:
Added in https://github.com/runelite/runelite/pull/19357
This PR should be closed

PR #15620, opened 2022-09-05
Add config option to use US floor numbering

ThePharros:
Adds a config option to choose US floor numbering:
image

Also fixed phrasing for "MAJORS LAVA BADS AIR" anagram clue solution.

(The approach may not be optimal and require some refactoring, but the framework may be feasible.)


Sharpienero:
Bumping this - can we see this merged in?

PR #15621, opened 2022-09-05
RunEnergy: Ring of Endurance addition

hugocowan:
**Why is this addition required?**

As of right now no plugin can safely calculate run energy usage when the Ring of endurance is equipped. There is no variable or item change that occurs when the ring does not have its passive ability enabled. With these changes, developers can reliably calculate run energy use while using the Ring of endurance.


hugocowan:
This is fully bugfixed/feature complete. I won't be adding any more to it/make any more commits unless @Adam- or anyone else suggests changes.

hugocowan:
Firstly, thank you for taking a look at this PR! Secondly, your changes look great, thank you for fixing my code - there were some bits that made me slap my forehead a bit there.

I found one thing I thought worth bringing up:

When you tick the Replace orb text with run time left box, the question mark shows up but the tooltip message about checking your Ring of endurance no longer shows. I think that there should still be text in the tooltip to explain the question mark.

image

When it’s unticked, there is no question mark, and the tooltip does show up. I think this is fine.

image

hugocowan:
To fix the above issue I suggest changing this:

```Java
if (config.replaceOrbText())
{
    sb.append("Run Energy: ").append(client.getEnergy()).append('%');
}
else if (plugin.getRingOfEnduranceCharges() == null && plugin.isRingOfEnduranceEquipped())
{
    sb.append("Check your Ring of endurance to get the time remaining.");
}
else
{
    sb.append("Run Time Remaining: ").append(plugin.getEstimatedRunTimeRemaining(false));
}

int secondsUntil100 = plugin.getEstimatedRecoverTimeRemaining();
if (secondsUntil100 > 0)
{
    final int minutes = (int) Math.floor(secondsUntil100 / 60.0);
    final int seconds = (int) Math.floor(secondsUntil100 - (minutes * 60.0));

    sb.append("</br>").append("100% Energy In: ").append(minutes).append(':').append(StringUtils.leftPad(Integer.toString(seconds), 2, "0"));
}
```

to this:

```Java
if (config.replaceOrbText())
{
    sb.append("Run Energy: ").append(client.getEnergy()).append('%');
}
else
{
    sb.append("Run Time Remaining: ").append(plugin.getEstimatedRunTimeRemaining(false));
}

int secondsUntil100 = plugin.getEstimatedRecoverTimeRemaining();
if (secondsUntil100 > 0)
{
    final int minutes = (int) Math.floor(secondsUntil100 / 60.0);
    final int seconds = (int) Math.floor(secondsUntil100 - (minutes * 60.0));

    sb.append("</br>").append("100% Energy In: ").append(minutes).append(':').append(StringUtils.leftPad(Integer.toString(seconds), 2, "0"));
}

if (plugin.getRingOfEnduranceCharges() == null && plugin.isRingOfEnduranceEquipped())
{
    sb.append("</br>Check your Ring of endurance to get the time remaining.");
}
```

Which results in this:

image
image
image

hugocowan:
Made Ring of endurance passive effect only affect run energy calculation when the stamina effect is inactive.
I confirmed the ring actually does reduce run energy by 15% with no stamina potion by testing with a timer in-game.

runelite-client/src/main/java/net/runelite/client/plugins/runenergy/RunEnergyOverlay.java
```diff
@@ -90,6 +91,11 @@ public Dimension render(Graphics2D graphics)
 				sb.append("Run Time Remaining: ").append(plugin.getEstimatedRunTimeRemaining(false));
 			}
 
+			if (client.getVarbitValue(Varbits.RUN_SLOWED_DEPLETION_ACTIVE) == 0 && plugin.getRingOfEnduranceCharges() == null && plugin.isRingOfEnduranceEquipped())
+			{
+				sb.append("</br>Check your Ring of endurance to get the time remaining.");
```
Adam-, line 96: I don't think this makes sense to show `Run time remaining: xyz<br>Check your Ring of endurance to get the time remaining.` like you are doing here since it is self-contradictory. I would probably do what I had prior to your last change.

hugocowan: The way it is now, the `xyz` value will always be a `?` when the message is showing.

This way we always inform the player why there's a `?` instead of a time. The way you had it before would only tell the user why there's a `?` if they had the replaceOrbText set to false.

hugocowan: image

In both of these images there's a `?` and the message `Check your Ring of endurance to get the time remaining.`

What was there prior to my change would only show the message in one of these images, but the `?` both times. I hope this is making sense.

PR #15624, opened 2022-09-07
devtools: highlight current region in locations tool inside instances

LlemonDuck:
When not in instanced regions, the Location devtool highlights the map region including your player as green:

image

This does not currently work in instances since the player region id is not transformed to its corresponding map region.

When in an instance, this PR normalizes the region ID for highlighting to use non-instance regions.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/devtools/LocationOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/devtools/LocationOverlay.java
@@ -64,6 +64,8 @@ public class LocationOverlay extends OverlayPanel
 
 		if (client.isInInstancedRegion())
 		{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/devtools/LocationOverlay.java
```diff
@@ -64,6 +64,12 @@ public Dimension render(Graphics2D graphics)
 
 		if (client.isInInstancedRegion())
 		{
+			WorldPoint instancePoint = WorldPoint.fromLocalInstance(client, localPoint);
```
Adam-, line 67: `fromLocalInstance` can never be null, maybe we can mark it as nonnull

LlemonDuck: removed null check

PR #15649, opened 2022-09-10
Fix Keybind not matching key released events of modifiers when other modifiers are held.

geheur:
to reproduce:
1. set keybind for pinging in Party plugin to "Ctrl" (can be any keybind outside of key remapping plugin, but this one has more noticeable results.)
2. press down both "shift" and "ctrl".
3. release "ctrl"
4. release "shift"
5. you can no longer walk around in-game until you press and release "ctrl" again, only ping.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/Keybind.java
@@ -125,9 +125,17 @@ public class Keybind
 			keyCode = KeyEvent.VK_UNDEFINED;
 		}
 
-		if (e.getID() == KeyEvent.KEY_RELEASED && keyCode != KeyEvent.VK_UNDEFINED)
+		if (e.getID() == KeyEvent.KEY_RELEASED)
 		{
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/config/Keybind.java
```diff
@@ -125,9 +125,16 @@ protected boolean matches(KeyEvent e, boolean ignoreModifiers)
...
+		if (e.getID() == KeyEvent.KEY_RELEASED)
 		{
-			return this.keyCode == keyCode;
+			if (keyCode != KeyEvent.VK_UNDEFINED)
+			{
+				return this.keyCode == keyCode;
+			}
+			else
+			{
+				return this.keyCode == keyCode && (this.modifiers & modifiers) == this.modifiers && ((mf & this.modifiers) == mf);
```
Adam-, line 136: how sure are you about the implicit mf unbox here? IJ warns for it, but I think for it to happen you have to get an VK_UNDEFINED key event which I am not sure happens?

PR #8216, opened 2019-03-15
Added combat levels to the player indicator overlay

Karel-van-Waveren:

PR #15659, opened 2022-09-11
Add option to display combat levels in Player Indicators plugin

### LINKED ISSUE #6567: Show combat level in player indicators overlay

ThePharros:
Closes #6567, supersedes #8216

Example:
image

Currently does not allow combat levels to be shown when the player is in a PvP area, so as to avoid issues with the

> Offers additional information about other players, with the purpose of scouting PvP targets

guideline. This may require additional condition checks should the current implementation not suffice (Currently, combat levels will still be shown if player is in a safe area that borders PvP areas, such as the Wilderness ditch or Ferox Enclave.).


ForkTheSpoonWrecker:
Any chance this will be added? Love the idea.

PR #15671, opened 2022-09-14
Fix BA plugin, which was broken in 668d99

### LINKED ISSUE #15638: [Barbarian Assault] Call change timer breaks after wave completion
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Wave completion inconsistently but frequently breaks the call change timer. Has only started happening since today's (Sept. 8th) updates. The timer will be stuck at whatever it displayed at the end of the previous wave and will not reset upon call change.

### Screenshots or videos

_No response_

### RuneLite version

RuneLite version: 1.8.33
Launcher version: 2.4.2

### Logs

_No response_

geheur:
Thanks @Kamielvf 
https://github.com/runelite/runelite/commit/668d99dfef72c900b1c6c7a5ab121863ddb06252 changed it from checking the old varbit value for being 1 to checking the new varbit value for being 1. As a result, `currentRound` was set to null when the wave started, instead of when the previous wave ended. Due to raciness between the varbit change and the widget load, this resulted in either a null or a properly initialized `currentRound` variable value. So, it was broken some waves and not others, randomly.

Closes #15638

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/barbarianassault/BarbarianAssaultPlugin.java
@@ -155,7 +155,7 @@ public class BarbarianAssaultPlugin extends Plugin
 	@Subscribe
 	public void onVarbitChanged(VarbitChanged event)
 	{
-		if (event.getVarbitId() == Varbits.IN_GAME_BA && event.getValue() == 1)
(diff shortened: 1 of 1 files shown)


Nightfirecat (approved):
Haven't tested, but this LGTM. Looking at 668d99dfef72c900b1c6c7a5ab121863ddb06252, the reason this bug was introduced was because the previous var names were very bad, which led to the code being understood wrong.

PR #15681, opened 2022-09-16
Add ToA supplies to InventoryID

ThePharros:


Nightfirecat:
As mentioned on Discord, we don't add ID definitions for hub plugins. (unused definitions run the risk of bit rot--where its value changes, but is not updated in our code because it is not discovered to be broken) You can use [`Client#getItemContainer(int id)`](http://static.runelite.net/api/runelite-api/net/runelite/api/Client.html#getItemContainer(int)) to get an item container instance from a hub plugin.

PR #15684, opened 2022-09-16
Adding ability to set colours using <col=XXXXX> tags to npc indicator…

jimma-rs:
Adding the ability to specify a specific color when highlighting an npc 

image


Nightfirecat:
We'd much rather have work put into developing reactive config panels to support the ability for multiple NPC highlight groups

Issue #14610, opened 2022-01-18
Add "Zealot's robes" bonus effect to "Skill Calculator"


TiboDeMunck:
**Describe the solution you'd like**
Add the Zealot's robes bonus effect to the Skill Calculator on the prayer tab

**Describe alternatives you've considered**
Calculating it myself

Giovds:
The [wiki](https://oldschool.runescape.wiki/w/Zealot%27s_robes) states the following:
> Zealot's robes are a set of robes that provides a 1.25% chance per piece to prevent [bones](https://oldschool.runescape.wiki/w/Remains#Bones) or [ensouled heads](https://oldschool.runescape.wiki/w/Ensouled_head) from being consumed whilst training [Prayer](https://oldschool.runescape.wiki/w/Prayer) – totaling 5% with the full set equipped.

A couple of things to note:
- The set does not provide an experience bonus. So it would probably follow [similar math as the Chaos Altar](https://oldschool.runescape.wiki/w/Chaos_Temple_(hut)#Mathematical_Proof).
- It works both on bones and ensouled heads. Quickly glancing over the code, the latter seems to require quite a refactor.
- It can be stacked with the Chaos Alter (resulting in a 52.5% chance of not consuming bones).

PR #15703, opened 2022-09-20
adding Zealots Robes to skill calc

fmahmud4:
Used the wiki percentage to add to the skill calculator.
REF: https://oldschool.runescape.wiki/w/Zealot%27s_robes

ISSUE: #14610


Nightfirecat:
This should support enabling both the Zealot's robes and other bone offering bonuses at the same time. Currently the plugin only allows enabling one bonus at once, so that would need to be added. (with the ability to ensure you can't enable incompatible bonuses) On top of that, the bonus offered is doubled when being used with the chaos altar bonus; such logic would also need to be added.
Overall, this feature will require a significant overhaul to how bonuses are handled.

Nightfirecat:
This was added in #17674.

PR #15704, opened 2022-09-20
Hotfix for the Plugin Hub Connection

zacharyrperales:
As of now, non-release versions of RuneLite do not have access to a manifest.js file for the Plugin Hub because there is either not one located in the directory for the specific version of RuneLite that is building or the directory itself is missing or has not been made yet. Instead of attempting to visit the repository directory for the latest version of RuneLite that we build, we visit the manually identified directory on the repository of the latest version of RuneLite which contains a manifest.js file (1.8.33.1) until a proper fix is implemented.


Enriath:
How it works at the moment is intentional. We don't provide builds of pluginhub plugins for snapshot versions. Usually you can pass `-Drunelite.pluginhub.version=<current RL version>` as a VM option to load the plugins built against the release into your build, which will work if there are no ABI changes. This configuration is not supported and if there are ABI changes you would need to build the plugins yourself and place them into the `sideloaded-plugins` directory when running in developer mode in order to have them loaded.

zacharyrperales:
> How it works at the moment is intentional. We don't provide builds of pluginhub plugins for snapshot versions. Usually you can pass `-Drunelite.pluginhub.version=<current RL version>` as a VM option to load the plugins built against the release into your build, which will work if there are no ABI changes. This configuration is not supported and if there are ABI changes you would need to build the plugins yourself and place them into the `sideloaded-plugins` directory when running in developer mode in order to have them loaded.

Whoops, I meant to merge and pull this change into my own master branch. I am still getting familiar with Git and GitHub. 

Thank you for all of the information on this topic.

PR #15706, opened 2022-09-21
agility: add Necropolis obstacle

ThePharros:
Example


runelite-api/src/main/java/net/runelite/api/ObjectID.java
```diff
@@ -22422,6 +22422,7 @@
 	public static final int LION_GUARD = 44584;
 	public static final int WARP = 44585;
 	public static final int WARP_44586 = 44586;
+	public static final int RUBBLE_44595 = 44595;
```
Nightfirecat, line 22425: You should never modify ObjectID; per the comment at the top of the file, it is automatically generated. Reference `NullObjectID.NULL_44595` instead.

Nightfirecat (changes requested):

PR #15721, opened 2022-09-23
Extend party ping feature and add advanced ping menu

kamielvf:
NB: watch with sound

https://user-images.githubusercontent.com/35824069/191918472-af26fa5e-14a7-4d40-9b88-8cd329a2efe7.mp4

preview preview
preview

Extends the existing party ping feature with the following:
- added 4 new pings, each with their own config options
- targets can now be pinged (npcs, objects)
- destination ping can show a line between source player and destination
- made ping duration configurable
- advanced ping menu can be shown after selecting a target and dragging the mouse
- when no colour preference is set in config, ping colour is based on unique hash

11dee6b0175591ac0cecae2673089c011d167c41 fixes an issue where pinging on the same tile/target will stack pings, instead it will make it so that it removes the older pings on the same target before adding a new one (per-player)

148738759029d966f56dca06d3351526935b7b48 also fixes another issue where the ping duration is dependent on FPS. The higher your FPS was, the faster the ping would fade.

This currently will not allow pings to target players, but that can easily be added if desired.


runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPingSelectionOverlay.java
```diff
@@ -0,0 +1,166 @@
...
+			renderRadialHighlight(graphics, baseX, baseY);
+		}
+
+		renderMenuIcons(graphics, centerX, centerY);
+		renderIndicationLine(graphics, mouseStartPosition, mouseCanvasPosition);
+
+		return null;
+	}
+
+	private void renderRadialMenuBackground(final Graphics2D graphics, int baseX, int baseY)
```
Nightfirecat, line 94: This and most of this class's methods can be made `static`.

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPingSelectionOverlay.java
```diff
@@ -0,0 +1,166 @@
...
+		graphics.drawOval(baseX + RADIAL_MENU_THICKNESS, baseY + RADIAL_MENU_THICKNESS, RADIAL_MENU_HOLE_SIZE - 1, RADIAL_MENU_HOLE_SIZE - 1);
+
+		graphics.setColor(OUTSIDE_STROKE_COLOR);
+		graphics.drawOval(baseX - 2, baseY - 2, RADIAL_MENU_SIZE + 3, RADIAL_MENU_SIZE + 3);
+		graphics.drawOval(baseX + RADIAL_MENU_THICKNESS + 1, baseY + RADIAL_MENU_THICKNESS + 1, RADIAL_MENU_HOLE_SIZE - 3, RADIAL_MENU_HOLE_SIZE - 3);
+	}
+
+	private void renderRadialHighlight(final Graphics2D graphics, int baseX, int baseY)
+	{
+		// substract 1 because TARGET ping type is our default and not shown in advanced menu
```
Nightfirecat, line 116: typo: subtract
```suggestion
		// subtract 1 because TARGET ping type is our default and not shown in advanced menu
```

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -320,27 +469,155 @@ public void onGameStateChanged(GameStateChanged event)
...
 	}
 
+	void resetPingMenu()
+	{
+		advancedPingMenuOpen = false;
+		mouseStartPosition = null;
+		pendingPartyPing = null;
+	}
+
+	private boolean pingIsSameSourceAndTarget(PartyPingData data, PartyPingData other)
```
Nightfirecat, line 479: This strikes me as a method which should be an instance method of `PartyPingData` and not a private method in this class.

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -320,27 +469,155 @@ public void onGameStateChanged(GameStateChanged event)
...
+		}
+		
+		PartyPingTargetType targetType = ping.getTargetType();
+		if (targetType != PartyPingTargetType.NPC && (ping.getPoint().getPlane() != client.getPlane() || !WorldPoint.isInScene(client, ping.getPoint().getX(), ping.getPoint().getY())))
+		{
+			return;
+		}
+
+		Integer target = ping.getTarget();
+		Color color = null;
```
Nightfirecat, line 509: This is set to some value (null or user preference) later, it does not need to be initialized here.
```suggestion
		Color color;
```

Nightfirecat (commented):
Very brief review at this point; I've not tested the code

PR #15724, opened 2022-09-25
Fix for explorer ring not updating its charge count on login sometimes.

geheur:
Note: The explorer ring charge tracker should really just read the varbit instead of storing the varbit into the config and reading the config. So, this PR should probably not be accepted. Also, this pr is not correct, I don't really understand why it works because the LOGGED_IN game state happens before the varbit change for the explorer's ring, in cases where the ring is not full and the varbit changed actually fires. It seems to work but it feels dirty.

Fix for explorer ring not updating its charge count on login in some cases (those being that it is currently full and the varbit was last at its full state or it is a fresh login (those are probably the same situation since a value of 0 represents full charge)). Likely caused by a regression in 668d99dfef72c900b1c6c7a5ab121863ddb06252. Previously the varbitchanged event would trigger the explorer's ring charge update on any varbit changing, but after that commit it only checks on the varbit changed event for that specific varbit.

Also updates the value on plugin startup. Previously it would have updated after opening the plugin relatively soon since varbits change a lot but this is now no longer the case as explained above.


geheur:
Closing in favor of #15724.

PR #15731, opened 2022-09-26
notifier: get terminal-notifier from login shell on macOS

ylh:
Per https://github.com/runelite/launcher/pull/111; as long as the user has their shell profile set up for whatever package manager they're using, this should reliably find `terminal-notifier`.


Adam-:
Can you pass the terminal-notifier arguments as arguments to the shell? This avoids having to double-escape them I think. Eg
```sh
sh -c 'foo "$@"' -- arg1 arg2 ...
```

ylh:
✔

PR #15757, opened 2022-10-01
Add Necropolis, Ruins of Ullek, and Tombs of Amascut regions to Discord plugin

ThePharros:
closes #15730


ThePharros:
As per the Discord discussion, the ToA lobby and chest room have been separated into their own Discord areas.

PR #15765, opened 2022-10-04
Implement feature request from discussion #15764 - XP Updater plugin

haardes:
Change to XpUpdaterConfig.java to allow players to set whether or not XP should be updated on world hop.
XpUpdaterPlugin.java check for set config value and correct gamestate.


haardes:
@Adam- bump

Nightfirecat:
I don't think this should be behind a config. Just check for logging out or for world hopping to trigger the updates.

notheowner:
I don't see why a toggle for world hopping could be a bad thing. We had the XP updater not updating on world hops for so long, some people might not even want it updating on hops now. Certain skilling or bossing methods require constant world hopping, it would result in spam updates flooding update tables in certain websites. 
You could argue that: "Hey just disable the plugin when doing those methods". I feel like that would cause more inconvenience overall, people forgetting to turn it on or off etc. rather than a what seems to me "simple" toggle.

Adam-:
The code requires you to have gained 10k xp before it will submit an update, so I don't think it would be spammy.

haardes:
Removed config-option for updating on world-hop

PR #15770, opened 2022-10-06
Minimap: remember zoom

JZomDev:
Sets zoom amount when reaching the login_screen state and saves it to the config to reload the value after loading client again.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/minimap/MinimapConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/minimap/MinimapConfig.java
@@ -38,19 +38,49 @@ public interface MinimapConfig extends Config
 	@ConfigSection(
 		name = "Minimap dot colors",
 		description = "The colors of dots on the minimap.",
-		position = 0
+		position = 3
 	)
 	String minimapDotSection = "minimapDotSection";
 
 	@ConfigItem(
 		keyName = "zoom",
 		name = "Zoom",
-		description = "Enables zooming on the minimap"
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/minimap/MinimapConfig.java
```diff
@@ -35,23 +35,46 @@ public interface MinimapConfig extends Config
 {
 	String GROUP = "minimap";
 
-	@ConfigSection(
-		name = "Minimap dot colors",
```
Nightfirecat, line 39: I don't think this config section needs to be moved for now

runelite-client/src/main/java/net/runelite/client/plugins/minimap/MinimapPlugin.java
```diff
@@ -96,6 +101,11 @@ public void onGameStateChanged(GameStateChanged event)
 			storeOriginalDots();
 			replaceMapDots();
 		}
+
+		if (event.getGameState() == GameState.LOGIN_SCREEN)
```
Nightfirecat, line 105: Adding a minimap zoom changed event to hook into would be a better way to go about this. (would require an internal change though)

raiyni:
Cleaned up relevant feedback. A zoom changed event would require an internal change.

raiyni:
The hub plugin [Default Minimap Zoom](https://runelite.net/plugin-hub/show/default-minimap-zoom) covers this

PR #15778, opened 2022-10-06
clues: add world map hint for music clue scrolls

Juzzed:
Fixes #15777

221006_221834_java_Outgoing

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
@@ -45,7 +45,7 @@ import net.runelite.client.ui.overlay.components.TitleComponent;
 @Getter
 public class MusicClue extends ClueScroll implements NpcClueScroll, LocationClueScroll
 {
-	private final WorldPoint location = new WorldPoint(2990, 3384, 0);
+	private static final WorldPoint LOCATION = new WorldPoint(2990, 3384, 0);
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
```diff
@@ -43,9 +43,9 @@
 
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 @Getter
-public class MusicClue extends ClueScroll implements NpcClueScroll
+public class MusicClue extends ClueScroll implements NpcClueScroll, LocationClueScroll
 {
-	private static final WorldPoint LOCATION = new WorldPoint(2990, 3384, 0);
+	private final WorldPoint location = new WorldPoint(2990, 3384, 0);
```
Nightfirecat, line 48: Why is `static` removed from this constant definition?

Juzzed: I just followed how CoordinateClue.java implemented LocationClueScroll.

Nightfirecat: I'd recommend overriding `getLocation()` as:

```java
@Override
public WorldPoint getLocation()
{
	return LOCATION;
}
```

rather than renaming the constant and relying on the lombok-generated getter to implement it.

PR #15828, opened 2022-10-21
entity hider: add option to hide Thralls

Digiridoo911:
Feature request: https://github.com/runelite/runelite/discussions/15096

Adds the option to hide Thralls


Adam-:
This should use the constants from `NullNpcID` instead of hardcoded values

PR #15831, opened 2022-10-22
adds Moss Giant to Moss giant 

### LINKED ISSUE #12904: Iorwerth Dungeon Moss Giants cause an extra entry in the loot tracker that persists until restart
While farming Mossy Keys, I've noticed that all my kills for the session appear under an entry called "Moss Giant", whereas my previous entries come under "Moss giant". I think I killed a normal moss giant first (once), then only killed the ones in the Iowerth dungeon after that. No loot seems to have been lost, it's just listed under the other name

The weirdness comes down to the Iowerth ones having the capital G in their name. Why Jagex did that I have no idea.

If I were to guess, the issue happens when the LootTrackerService adds the loot to the database, sees that there's already a moss giant entry, and uses that. I'm not entirely sure if it's case-insensitive like that though.

**To Reproduce**
0. Delete any Moss giant loot in your loot tracker, and make sure you're logged in
1. Kill a regular Moss giant (eg. one on Karamja)
2. Restart your client to make sure it's saved on the server
3. When the client starts, note that the loot is under `Moss giant` in the tracker
4. Kill a Moss Giant in the Iowerth Dungeon, notice it's created a new loot entry under `Moss Giant`
5. Restart the client
6. Notice that you now have 2 kills under `Moss giant`, and `Moss Giant` is nowhere to be found.

Ideally, the loot should all come under the same box in the client, maybe by making the Loot Tracker check for existing eventIDs case-insensitively.

BearSprayGit:
_Might_ fix #12904 without checking for old record names and if there's a lowercase match using that name

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -598,15 +598,20 @@ public class LootTrackerPlugin extends Plugin
 	void addLoot(@NonNull String name, int combatLevel, LootRecordType type, Object metadata, Collection<ItemStack> items, int amount)
 	{
 		final LootTrackerItem[] entries = buildEntries(stack(items));
-		SwingUtilities.invokeLater(() -> panel.add(name, type, combatLevel, entries, amount));
+		if (name.toLowerCase() == "moss giant")
+		{
+			name = "Moss giant";
+		}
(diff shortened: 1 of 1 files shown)


Nightfirecat:
We should definitely not be special-casing "Moss giant" versus "Moss Giant" in the loot tracker. If anything, we should (a) not change the current behavior as it's arguably not a bug, or (b) use case-insensitive string comparisons when checking whether a box exists matching the name of the monster which dropped the loot.

PR #15837, opened 2022-10-23
Individual loottracker suppression of warning 

lethalstrain:


Nightfirecat:
What's the use case for this? Where would you need to reset one (or several) but not all entries in the loot tracker? We want the warning to be present as this impacts users being able to view their loot on the logged-in view of runelite.net or on other computers when synced with the Account plugin.

lethalstrain:
For me the use case would be to not have the warning when i reset everything from eg. black demons (i tend to only want to keep a few things open at the same time)

PR #15844, opened 2022-10-24
Use Wise Old Man API v2 in xpupdater plugin

rorro:


psikoi:
Context: We're migrating to a v2 of our API. I'll also change the seasonal domain name when the next league comes around.

PR #15852, opened 2022-10-27
Remove unused list of MUSIC_TRACKS_UNLOCKED from VarPlayer

claytondetke:
Adding values to VarPlayer with the varpIds of the bitfields for music track unlocks that have been added between 2020 and 2022.


claytondetke:
Just to explain why I'm posting this PR out of the blue, I'm currently working on a plugin that requires knowing how many total music tracks the player has unlocked. In order to find that out, I needed to figure out the IDs for the VarPlayer values for the more recent music tracks. I managed to find all 4 of them, and I figured I'd do this in case anyone else needs these values as well.

Adam-:
These varps are there from when we used to have a music unlock plugin. But they are all unused now, so we can probably just remove them instead.

claytondetke:
Whatever happened to that plugin? Also, I'm using those for the plugin I'm working on, so I'd prefer to leave them in if it isn't too much trouble. But if it's important that they be removed, I'll just copy the varpIds to live in my plugin instead.

Nightfirecat:
Jagex added it as a vanilla feature, so we removed that plugin. See #10571

Adam-:
> Whatever happened to that plugin? Also, I'm using those for the plugin I'm working on, so I'd prefer to leave them in if it isn't too much trouble. But if it's important that they be removed, I'll just copy the varpIds to live in my plugin instead.

You can just use `getVarpValue(int)` in your plugin and not have to edit the core VarPlayer.

claytondetke:
Updated to remove all of the apparently unused music unlocked varp values. I will be using their values on the plugin I'm working on personally though. Also, out of curiosity, is there any form of documentation listing what various different varps/varbits represent? If so, should I add the values I found myself to it? If not, that's fine too, I'm just trying to be of help.

Nightfirecat:
That varies from entry to entry. In the past, we were much less strict about new entries being well-documented, nowadays we like it to be clear just from looking at the entry how it works since there's so much variance from var to var. You can see toward the bottom of this file some of the entries have detailed javadoc blocks, despite a lot of the earlier ones not. If you're particularly interested, we wouldn't mind more javadoc to cover existing entries, but that should be a separate commit/PR from this one.

claytondetke:
Ok, cool. So if I were to write detailed comments explaining how the music unlocked bitfields work, would it be okay to add them in? Again, probably a separate commit/PR though. Also, sure I can look into improving the documentation on some of the other varps.

Nightfirecat:
Do you mean as documentation on some other vars or...? We still want these music unlocked var definitions removed as we don't use them. Adding documentation to them wouldn't be sufficient reason for us to keep those definitions.

claytondetke:
Documentation on some other vars. Sorry for the poor wording.

Nightfirecat:
If it's relevant to them, sure

PR #15855, opened 2022-10-27
config: Rebuild ConfigPanel when visible configs change

### LINKED ISSUE #14713: Plugin Panel Refresh / Rebuild
We want to add Graphic Detail Presets to 117 but we need a way to reload the config panel for that plugin or to update individual settings on the panel and show the changes

a call that we can update a whole plugin panel for the settings and the changes to be seen visually or a way for us to update them individually

Reflection, but this is a little Janky and not a good way to do it

aHooder:
When config values are modified by anything other than the config panel itself, changes aren't reflected by the panel. This PR makes the config panel automatically rebuild itself when a config key corresponding to an item in the panel is changed. To avoid breaking component focus when changing a config manually, the panel isn't rebuilt when it itself is directly responsible for the change. Finally, to avoid unnecessarily rebuilding the panel multiple times for near-simultaneous changes, a single rebuild call is scheduled (or rescheduled) to run 10 ms after the last applicable config change.

Potential use cases:
- Implementing presets for a plugin, such as low/medium/high presets for 117 HD.
- Resetting a config to its default value without having to rename the config key.
- Automatically fix conflicting configs.
- Disable a config on failure.

Resolves #14713.


abextm:
You can probably use an awt timer for this instead of the scheduler. LGTM if this is blocking something you need, otherwise I would rather wait for my config panel rewrite, which is a few months out probably

aHooder:
> You can probably use an awt timer for this instead of the scheduler. LGTM if this is blocking something you need, otherwise I would rather wait for my config panel rewrite, which is a few months out probably

It'd be a nice thing to have, though it can wait I suppose. The ideal thing imo would be for each config panel component to update itself on config change, instead of rebuilding the entire panel. Is that something you'd want to include in your rewrite?

aHooder:
Closed in favour of #19117

PR #15874, opened 2022-11-01
Add drawOutline method in ModelOutlineRenderer for projectiles

SamHammersley:


Nightfirecat:
What would this be used for? Presently, this has no uses because it's not possible to mark projectiles. (I can't really think of a good use for this either)

Adam-:
The client uses the projectile velocity to compute the rotation (which itself is computed from `dest - current location`), however you can just use it:

```java
double t = atan2(velocityX, velocityY); // here +y is away from the camera, instead of down like in local space
```

you can then convert it from radians to jau by multiplying it by `2048/(2pi)`

```java
int jau = (int) (t * (2048/(2 * Math.PI)));
```

this gives you a value in the range `-1024` to `1024`, with `0` being north (atan2 returns a value in the range `-pi` to `pi` with 0 being north)

Then you can just offset it by 1024 to get it in the range 0-2048 with 0 being south

```java
jau = jau + 1024 & 2047; // and maybe & 2047 for good measure
```

PR #15876, opened 2022-11-02
Combat Level: Update to work on the Character Summary tab

testing-ongithub:
Before:
before

After:
after


runelite-api/src/main/java/net/runelite/api/ScriptID.java
```diff
@@ -411,4 +411,7 @@
 
 	@ScriptArguments(integer = 9)
 	public static final int WORLDMAP_LOADMAP = 1712;
+
+	@ScriptArguments(integer = 7, string = 1)
+	public static final int CHARACTER_SUMMARY_TEXT_FORMAT = 3948;
```
Nightfirecat, line 416: Let's call this "account summary" text summary to match <https://github.com/Joshua-F/cs2-scripts/blob/afbd80d369facdcbbe01271dbc064f97924f6d0a/scripts/%5Bclientscript%2Caccount_summary_update_combatlevel%5D.cs2> and <https://github.com/Joshua-F/cs2-scripts/blob/afbd80d369facdcbbe01271dbc064f97924f6d0a/scripts/%5Bclientscript%2Caccount_summary_update_totallevel%5D.cs2>
```suggestion
	public static final int ACCOUNT_SUMMARY_TEXT_FORMAT = 3948;
```

runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelPlugin.java
```diff
@@ -153,6 +158,35 @@ public void onConfigChanged(ConfigChanged event)
...
+	@Subscribe
+	public void onScriptPreFired(ScriptPreFired scriptPreFired)
+	{
+		if (scriptPreFired.getScriptId() == ScriptID.CHARACTER_SUMMARY_TEXT_FORMAT)
+		{
+			final String[] stringStack = client.getStringStack();
+			final int stringStackSize = client.getStringStackSize();
+			final String levelText = Text.removeTags(stringStack[stringStackSize - 1]); // first argument
+
+			// This script is used for both total level and combat level, so verify the script is modifying the combat level
```
Nightfirecat, line 170: Are you able to update the widget text after script 3893 (clientscript,account_summary_update_combatlevel) runs rather than needing this check?

runelite-api/src/main/java/net/runelite/api/ScriptID.java
testing-ongithub: Updated, thanks! Would you like me to update the widget and variable names throughout this pull request from "character" to "account" as well?

Note that the reason I originally chose "character" rather than "account" is that when you mouse over the button to change to that tab, the mouseover text is "Character Summary".

runelite-client/src/main/java/net/runelite/client/plugins/combatlevel/CombatLevelPlugin.java
testing-ongithub: Unless I'm misunderstanding something about your suggestion, I don't think that works. If you update the widget text in the scriptPostFired event for the script you mentioned, the text winds up off-centered:
off-centered

On the other hand, by running with the scriptPreFired event for the 3948 script, the text automatically gets centered as seen in this PR's main description.

Nightfirecat:
I'm seeing a bit of flicker for the combat level when opening the character summary page. (you can trigger this by flipping between the character summary tab and any other tab like the quest list) Is that something which can be fixed?

testing-ongithub:
I had previously noticed the flicker and assumed it wasn't something that could be fixed. However, your comment motivated me to look into it again, and I found that there is a second script that _also_ modifies the text (🍝). I've updated this PR to update the combat level text when that script runs too, and that seems to have fixed the flicker.

PR #15879, opened 2022-11-03
Prayer overlay | Customisable Indicator Colours

Mantic-says-hi:
Add ability to change the colour of the 'PrayDoseOverlay' and Pray Flick overlay for colorblindness support and aesthetic reasons.

Since the Prayer dose overlay might be a bit tricky to setup for people initially; I've added 3 new preset colour options along with the default. On-top of this the final option 'Custom' will allow users to create a custom set of colours to fade in and out of.

Colour pickers are below the appropriate places where the settings that enable the colours are to make it intuitive about what option changes what overlay. ( Picture below if you don't understand what I mean )
 
image


Mantic-says-hi:
Visualisation of change made based on feedback in Discord from Adam#8591

image

Issue #15892, opened 2022-11-06
EIGHTH is incorrectly spelled as EIGTH in MenuAction PLAYER_EIGTH_OPTION


Ethan-Vann:
### Type

Other

### OS Version

N/A

### Bug description

EIGHTH is incorrectly spelled as EIGTH in MenuAction PLAYER_EIGTH_OPTION
EIGHTH is the correct spelling

### Screenshots or videos

![bug](https://user-images.githubusercontent.com/22735513/200191762-02409040-51e5-47a9-951c-b68463dab3ab.PNG)

### RuneLite version

1.9.3
Dev

PR #15904, opened 2022-11-08
Change spelling of EIGTH to EIGHTH in MenuAction.java

wtbtucker:
Associated with issue #15892


Nightfirecat:
It's not enough to change only the enum name; you also need to update the name in all classes where it's used (which is causing the compilation failure here). Additionally, since this is an API which is used by some hub plugins (see: https://abextm.github.io/plugin-hub/#WyJeTG5ldC9ydW5lbGl0ZS9hcGkvTWVudUFjdGlvbjsuUExBWUVSX0VJR1RIX09QVElPTiIsIiJd), we must either notify those hub plugins of the new name, or maintain a `@Deprecated` entry with the same ID (and accounting for that duplicate ID entry in the static integer -> MenuAction mapping in the class) to ensure those plugins are still able to build and function correctly.

PR #15917, opened 2022-11-12
menu swapper: Fix duplicate Reset option in bank, Use submenus for ui swaps

Nightfirecat:
Because the bank tags plugin adds an "Edit-tags" menu entry which has a
null widget, this causes the highestOp calculation to be incorrect for
all entries prior to it, as the loop effectively `break`s when some menu
entry with a widget not matching that of the current entry's widget is
encountered. This commit fixes this behavior by moving the widget check
inside the loop body and `continue`ing on a non-match rather than
`break`ing.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
@@ -1089,9 +1089,15 @@ public class MenuEntrySwapperPlugin extends Plugin
 
 					// find highest op from the current menu, post any existing swaps, for inserting Reset
 					int highestOp = 10;
-					for (int i = idx; i >= 0 && entries[i].getWidget() == w; --i)
+					for (int i = idx; i >= 0; --i)
 					{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java
```diff
@@ -1089,57 +1094,105 @@ private void configureUiSwap(MenuOpened event)
...
-
-								chatMessageManager.queue(QueuedMessage.builder()
-									.type(ChatMessageType.CONSOLE)
-									.runeLiteFormattedMessage(message)
-									.build());
-
-								log.debug("Unset ui swap for {}/{}", componentId, menuEntry.getTarget());
+						if (leftClick != null)
+						{
+							submenuTarget = entry.getTarget();
```
Adam-, line 1129: This is wrong because it doesn't set a menu target on the Swap left click/Swap shift click menus if there is no configured swaps.

Adam-:
added in 1b941cc72f2ef0cabf36651413dc153571897049

PR #15918, opened 2022-11-12
ge plugin: convert untradeable/members items for alt-click searching

deadmeu:
Implementing https://github.com/runelite/runelite/discussions/15908

The `getWidget` method is copied from [`WikiPlugin`](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiPlugin.java#L348). Not sure if there is a better way to get the widget (`menuEntry` is of type `RUNELITE`), or maybe it should just be moved into client or another helper class.

I'm only taking the first result of the `ItemMapping.map()` call. It would be nice if we could search against all of the components of an item, but that would require an enhancement to the search. There's an open discussion for adding regex filtering to the search (https://github.com/runelite/runelite/discussions/14991) but that might be a bit overboard.

Also, I've found that sometimes, alt-click searching a Neitiznot Faceguard searches for a Basilisk Jaw, but other times it searches for a Neitiznot Helm. I think the item mapper might be randomly choosing an order at launch?


Adam-:
This doesn't work because it isn't safe to access item manager from the awt event thread. You can see this if you run with assertions `-ea`. Even the current behavior of looping the menus off-thread is technically unsafe, but since it isn't mutating anything it kind of works.

Normally we would use a normal menu entry callback for this type of feature. However the client treats button presses with alt pressed as mouse button 4, which is used to move the camera, which prevents clicking on menus. This is why this uses a mouse listener instead.

I tried working it into the menu system anyway by rewriting the input event to remove the alt mask, but the overlay renderer ended up consuming the mouse event in an attempt to move the inventory, so that won't work.

The best I can figure out is probably https://github.com/Adam-/runelite/commit/6106697b23a60aaac7a78c2bcc858043f0c137ec which just moves the item searching to the client thread. Preferably I think I would like to make this feature have a configurable key bind which is not alt because it interferes with too many other things. 

Your observation of it picking an item at random to lookup seems bad. At the very least it should be deterministic, but for things like the faceguard it has no clear answer as both items it breaks down into are tradable.

Maybe it would be best to just strip the (members) instead.

PR #15929, opened 2022-11-18
Show ground item expiry in whole seconds

alexjurkiewicz:
Previously, expiry was shown in tenths of a second. If players need this much precision, ticks are still available.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsOverlay.java
@@ -366,7 +366,7 @@ public class GroundItemsOverlay extends Overlay
 					final String timerText;
 					if (groundItemTimers == DespawnTimerMode.SECONDS)
 					{
-						timerText = String.format(" - %.1f", despawnTimeMillis / 1000f);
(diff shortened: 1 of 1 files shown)


Enriath:
....but why?

Nightfirecat:
fwiw the existing format matches that of npc respawn indicators, with 0.1 precision. I don't see why this should change.

alexjurkiewicz:
Having item expiry display in tenths of a second adds a lot of visual
noise, I think.

What do you think about a configuration option to choose the precision? The
values could be "1 sec" and "0.1 sec".

Nightfirecat:
Presently, the pie despawn timer only appears when you hold the hotkey for the plugin. I had thought the same was true for the seconds and ticks timers, but evidently those are always shown. I could see an argument for a toggle to show those only when the hotkey is held, but changing the format seems pretty pointless.

alexjurkiewicz:
From Discord:

> Adam
> Today at 11:12 AM
> [...] #15929 is guaranteed to piss off at least several dozen people, most of which will come complain to me
> so i cant do that

Frosty-J:
While I don't use the seconds setting myself and this pull request seems dead in the water, I have to agree. Especially now that clue scrolls can be juggled for 3660.0 seconds at a time, decisecond precision adds more text that doesn't serve much purpose, given the 600ms tick rate.

PR #15938, opened 2022-11-20
Grounditems: Add option to consider item value per inventory slot

### LINKED ISSUE #12891: Ground Items : Value checks shouldn't combine non-stackables.
Firstly, if you have no idea what I'm talking about. let's say you have a pile of items:

10xArrows (1k each)
10xBones (1k each)

You are in the middle of nowhere with only 1 slot, which do you pickup?
Obviosly you pickup the arrows. They only take 1 slot and get you 10k worth.

Let's say you have set "Hide < Value" to 5k.
Right now, both the pile of arrows and the pile of Bones would be highlighted.
I think it should only highlight the arrows, because A pile of bones is only 1k a slot, while the arrows is 10k a slot.

As an example of how this is inconsitent, lets say instead of it being 10 bones, it is 10 unique non-stackable items. Now the items are not highlighted, because "Ground Items" does a value check on each item seperatly, even though value-wise that pile is identical.

The change I would suggest is that when ever a value check is performed, that it only totals up stackables.

I can't think of a situation where you wouldn't want this. If there is though, a checkbox called "Combine non-stackable values" would be clear. (Current behaviour is checked)

Zeprus:
Adds configuration to the Ground Items plugin to consider item value per inventory slot.
If enabled unstackable and unnoted items on the ground won't combine their value to determine hidden or highlighted status.
Enabled on default

Closes #12891

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItem.java
@@ -50,6 +50,7 @@ class GroundItem
 	@Nullable
 	private Instant spawnTime;
 	private boolean stackable;
+	private boolean noted;
 
 	int getHaPrice()
 	{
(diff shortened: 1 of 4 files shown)


Zeprus:
Just realized I created the wrong PR, whoops.

PR #15939, opened 2022-11-21
gpu: add draw distance requires compute shaders

alexjurkiewicz:
Draw distance requires compute shaders to be enabled, so:

1. Move compute shader's config item before draw distance
2. Mention the dependency in draw distance's description


alexjurkiewicz:
> Adam
> Today at 11:17 AM
> idm just mentioning it requires compute shaders to be on.  but I dont think reordering it to the top is a good idea since it is on by default and is a more-advanced option I dont want people to immediately be flicking on/off

raiyni (changes requested):
Revert all of the ordering changes and only change the description per the discussion.

alexjurkiewicz:
done, thanks for the feedback!

PR #15957, opened 2022-11-24
Fix misspelled enum 

### LINKED ISSUE #15892: EIGHTH is incorrectly spelled as EIGTH in MenuAction PLAYER_EIGTH_OPTION

John-Kryspin:
Resolves #15892

After this is shipped we should eventually let these library owners know, and then remove this deprecation, and leave only the properly spelled ENUM

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/MenuAction.java
+++ b/runelite-api/src/main/java/net/runelite/api/MenuAction.java
@@ -231,7 +231,7 @@ public enum MenuAction
 	PLAYER_FIFTH_OPTION(48),
 	PLAYER_SIXTH_OPTION(49),
 	PLAYER_SEVENTH_OPTION(50),
-	PLAYER_EIGTH_OPTION(51),
+	PLAYER_EIGHTH_OPTION(51),
 
 	/**
 	 * Menu action for normal priority child component actions.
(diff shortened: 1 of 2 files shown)


vasama:
It still seems misspelled. I assume the intended spelling is `EIGHTH`?

John-Kryspin:
Yep. That was me working too quickly. Fixed!

Adam-:
Leaving the old one there but unused is not useful. For example, using `MenuAction::getType` will return the new value, not the old value, so any plugin relying on a check for the old value will silently break vs having a compilation error. So it is better to just rename it.

John-Kryspin:
@Adam- thanks Adam, but if we go ahead and change it, are we concerned it will break dependent plugins instantly? I'm not aware of the process for how to roll out breaking changes like what you're proposing. Happy to change the commit, after learning more about the process!

Adam-:
From discord, abex suggested just adding a static field with the old name. Which will preserve backwards compatibility.

John-Kryspin:
What do we think now? Happy to discuss on discord as well. Thanks :)

Adam-:
Just update player indicators to use the name

PR #15968, opened 2022-11-27
Fairyrings: Add custom fairyring tags

### LINKED ISSUE #12388: Suggestion: Fairy Ring Travel Log Nicknames (Tags)
**Would it be possible to add a right click fairy ring code in the travel log>set name (tag)?**

I can't really complain too much the way fairy rings work these days is a massive improvement and I love the travel log + being able to add favourites but I always find myself opening my map and looking for the fairy ring icon for certain areas and getting the code from there, I know it shouldn't be hard to memorise them but some of the descriptions have me stumped I wouldn't think this would be hard to add to the client correct me if I'm wrong but I feel this would be very useful and make things a lot easier.

Examples of how I'd find this useful anyway would be the fairy ring code CIR, I use this atm to get to the farming guild, I am not 100% familiar with Kourend so I get confused, another one when I go fishing for karambwan I am looking for Karamja in the codes descriptions without going back and forth tens of times I wouldn't think to go to the gnome glider as after 15 years of playing on and off I am only just discovering new areas, training methods etc due to playing Ironman.

Not sure how much support this would get and I will eventually learn what is what over time but it would just be really nice to have this option in my honest opinion.

**Some ideas on how this could be implemented (or even toggleable options for multiple);**

1. Change the description of the destination completely underneath the code.
2. Keep the original description but have the nickname in brackets.
3. Have the hover option display the tag.
4. Tag can be searched for in the search bar.

Cheers.

nocommentsgood:
Close #12388

Allow the users to create custom descriptions under FR codes in the travel log, when using a Fairyring. 
Add config items to FairyRingConfig to persist user custom FR descriptions.
Enable fairyring teleport button to read custom FR descriptions.


Felanbird:
Superseded by #16168

PR #15979, opened 2022-12-01
Fix the info plugin side panel from re-adding the same things a secon…

geheur:
…d time when the plugin is restarted

This prevents visual bugs and duplicated listeners or events in the side panel if you restart the info plugin.


Adam-:
fixed in f753cad4c7c239e8388f9ff3ae992be84552291b

PR #15998, opened 2022-12-06
Add Tooltip to Agility Obstacle Examine

John-Kryspin:
When encountering an obstacle highlighted in red, I usually go to the wiki to look it up, to see how many levels away I am from using it. I did not know this detail was shown on the map.

Intuitively, I wished that examining the obstacle would give me that context without having to click away to the wiki or open up the map.

This PR implements that feature, re-using the same function which exists to generate the map tooltip.
image

I also added a config. Not sure why someone would turn this off, but good to give the user an option.

Let me know what you think

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
@@ -38,33 +38,18 @@ import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.Client;
-import net.runelite.api.ItemID;
+import net.runelite.api.*;
+
 import static net.runelite.api.ItemID.AGILITY_ARENA_TICKET;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
```diff
@@ -38,33 +38,18 @@
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.Client;
```
raiyni, line 41: Revert all of these import changes. Your static imports and normal imports should be configured as described in the wiki.

raiyni (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/agility/AgilityPlugin.java
John-Kryspin: Thanks for the review, I applied those settings in the code conventions and re-committed that file

Issue #14760, opened 2022-03-18
world map: Mouse tooltips should be constrained to world map view


Skretzo:
When the world map icon mouse hover tooltip extends beyond the world map view it should become right-aligned to the edge of the world map view, similar to how both the `Mouse tooltips/highlight` plugin and the official osrs world map icon tooltips work.
![image](https://user-images.githubusercontent.com/53493631/159012711-3cf1f565-96be-4174-8b3a-dd37e8e97058.png)

Vandager:
Looking into this.

PR #16009, opened 2022-12-07
worldmap:Update world map tooltip positions to stay within the world map

peterforsling:
Resolves issue #14760

This is a duplicate of my previous PR #15529. I inadvertently did some serious damage to my fork of RuneLite so I nuked it and started it from scratch.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/worldmap/WorldMapOverlay.java
@@ -339,8 +339,6 @@ public class WorldMapOverlay extends Overlay
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


Skretzo:
@raiyni The tooltips still need to be restricted to the viewport, like the vanilla tooltips are

peterforsling:
Based on what @Skretzo is saying, if the tooltips stay within the viewport then the overlay drawing order issue isn't an issue that needs to be resolved, right? Since the tooltips will never overlap with anything outside of the world map viewport.

Skretzo:
The overlay draw order issue also needs to be addressed because of elements inside the world map:
animation

papichickens:
Could i claim this bug? It's for a university project and I'd reckon that I could get a fix.

cdfisher:
> Could i claim this bug? It's for a university project and I'd reckon that I could get a fix.

You can work on it and submit a PR with a fix but I don't think anyone will assign sole responsibility for it.

PR #16017, opened 2022-12-11
idlenotifier: handle non-standard attack options during npc transforms

LlemonDuck:
NPCs which begin combat with a menu option that is not "Attack" are missed during combat detection. This PR uses NpcChanged to detect when an NPC that we are interacting with transforms and gains an Attack option while still being interacted with.

I noticed this at Kraken since combat begins with "Disturb" as a Whirlpool before the NPC transforms into the Kraken, and verified that my change resolved the issue.


LlemonDuck:
I assume so but I don't have a Kraken task to check anymore. If I get one, I'll confirm.

LlemonDuck:
Yes, your commit works for me.

Nightfirecat:
Added in 82fe3f371f9cf3b2a66c7cd022b7daec627613c3

PR #16036, opened 2022-12-16
config: add Warn annotation

Macweese:
Allows selective behaviour to the warning prompt when toggling plugin settings.
For example, warning the user a feature may cause instability only when attempting to enable a feature. The user would on the other hand not have to confirm disabling if it was previously on.

Allows for three types of warnings:
`ALWAYS` - issue a warning prompt regardless of change
`ON_ENABLE` - warn when attempting to enable
`ON_DISABLE` - warn when attempting to disable
By default it's set to always, to be in line with current behaviour.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigItemComparator.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigItemComparator.java
@@ -33,6 +33,7 @@ import java.util.regex.Pattern;
 
 public class ConfigItemComparator<T extends Number & Comparable> implements Comparator<T>
 {
+	@Override
 	public int compare(T a, T b) throws ClassCastException
 	{
 		return a.compareTo(b);


LlemonDuck:
* `@ConfigItem` already has a `warning` field which is shown when enabling the respective option: https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/config/ConfigItem.java#L46
* I don't think there's much of a use case for warnings on disabling config items.
* If this moves forward, you should move the warning text into your new annotation as well.

raiyni:
I'm struggling to see the point of this pr. You removing warnings that could already exist for all types and replacing it with regressions.

It looks like you are attempting to add validation in a spot that's not really needed in core?

xianhuiling2:
you all idiots stupid are OK

PR #16037, opened 2022-12-16
slayer: cancel task on inferno fail

LlemonDuck:
When dying in the Inferno with a Zuk task (and only 1 remaining), your task is cancelled. However, the Slayer plugin still tracks that your task is ongoing.

The plugin rightfully recognizes players *leaving* the Inferno and the Fight Caves, but there is a distinct, less specific message that appears when failing the Inferno: `You no longer have a slayer task.`

image

This commit generalizes the message check to merge the three possible messages (two original, one new) into a regex.

I have not confirmed whether this message appears when failing the Fight Caves. I also do not know for certain whether this message could appear from anything else, but if so it probably has the same effect.


Nightfirecat (approved):

Nightfirecat:
I wonder what message (if any) you get from leaving the fight caves/inferno when you have a task of 2 kills from having master/grandmaster CA tasks completed. If we need to fix that after the fact, we can. I doubt that's something that comes up often.

PR #16044, opened 2022-12-17
Item Charges: Add support for Tombs of Amascut supplies

Jroon94:
| Before | After |
|--------|-------|
|toa_charges_before |toa_charges_after|

~~+ Added toggle for non-potion supplies (Scarabs, bandages, salts)~~


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
```diff
@@ -453,4 +453,16 @@ default boolean braceletOfClayNotification()
...
 	}
+
+	@ConfigItem(
+		keyName = "tombsOfAmascutSuppliesCharges",
+		name = "Tombs of Amascut Supplies Charges",
+		description = "Show Tombs of Amascut supply item charges/uses",
+		position = 32,
+		section = chargesSection
+	)
+	default boolean showTombsOfAmascutSuppliesCharges()
```
Nightfirecat, line 464: Instead of adding a new config for this, these items should use the `showPotionDoseCount()` config instead. I know not every toa item is a potion, but most of them are used like potions, and I don't see a world where users want only some of toa's supplies labeled this way

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargeConfig.java
Jroon94: Yeah, I thought so too. Just figured I'd put it this way first to be safe and see what you guys thought. Have reverted it to assume them as potions

raiyni (approved):

xianhuiling2:
idiot

Jroon94:
@Nightfirecat anything left to do on this? Just checking in since it's been awhile

PR #16056, opened 2022-12-22
Make inventory grid delay tied to anti drag delay

Jbleezy:


Adam-:
added in dd2af32008985a4751808595e875c2ca56548422

Issue #11878, opened 2020-06-11
Exiting Diablo II casts the game into a white screen


AcridBrimistic:
**Describe the bug**
Opening Diablo II and then closing it while RuneLite is open turns the game client into a white screen.

This can be remedied by doing something that causes a screen update (e.g changing resolution, low detail mode, etc.)

**To Reproduce**
1. Open RuneLite and Diablo II concurrently.
2. Close Diablo II.
3. The game client is now a white screen

**Expected behavior**
The game should continue as normal, with nothing unusual happening.

**Screenshots**
![what](https://user-images.githubusercontent.com/18503603/84442442-4dbad700-ac46-11ea-8e4e-76f81d0b4a69.png)

**Environment (please complete the following information):**
 - OS: Arch Linux (X, icewm, no compositor)
 - RuneLite version: 1.6.18.1
 - Launcher version: 2.1.0
 - Diablo II version: 1.13c

Nightfirecat:
Does [disabling hardware acceleration](https://github.com/runelite/runelite/wiki/Disable-Hardware-Acceleration#linux) fix this for you? Also, can you upload your `client.log` and `launcher.log` files from when this occurs?

Adam-:
fixed in f072c82973d58889abcb3df064ac3efb2eee4deb

Issue #3422, opened 2018-05-28
White screen


Popedigoz:
My RuneLite goes white if I Windows + L or (I think) open fullscreen games (did not receive any answer from anyone after asking thrice on Discord)

stevenoctopus:
I just received this on Windows when a program was asking for administrator rights and the popup window hid and dimmed my screens.

sedmicha:
I just encountered this issue after returning from lock screen on Windows 10 while having RuneLite in stretched fixed mode

Popedigoz:
Yeah, I forgot to mention I am also playing RuneLite with fixed stretched mode, (fullscreen) 'Keep aspect ratio' is ticked if it counts.

thgonzalez:
Same as everyone here.

Y2Grs:
For a temporary fix you can disable "Stretched Fixed Mode" and the screen should reappear and then you can just enable the plugin again and you should be fine... Its a little annoying having to do this but hey it might save you some gear if this happens to you while you're in the wild or something

deathbeam:
Looks like this is solveable by disabling hardware acceleration: https://github.com/runelite/runelite/wiki/Disable-Hardware-Acceleration

DankSteelMemes:
Still happens even with that flag there and HW acceleration disabled. 

Disabling Stretch mode like advised does fix the issue but enabling it brings the white screen right back. Unfortunate this won't be fixed...

khogeland:
Disabling HW acceleration isn't a great solution if you're trying to alt tab from a full screen game because it degrades the performance of both

Adam-:
The root cause of this was identified and fixed in f072c82973d58889abcb3df064ac3efb2eee4deb

PR #16058, opened 2022-12-22
client: fix rendering issue

Macweese:
This PR is aimed to fix a rendering issue in the client caused by Stretched Mode which was introduced in f672ef9014d98892ea0fd4c48d3932db173bea2b
For reference to issues see #3422 ,  #11878

Changes include; dismissing any `VolatileImage` which has been restored or is incomplete, addition of a new event to relay information about display changes, improve how the client determines which `GraphicsDevice` it belongs to, fixing VSync in the GPU plugin.

Tested systems
• Windows 10.0.19044.2364, 10.0.19044.2251, 10.0.18362, 7 (Build 7601)
• Ubuntu 20.04 LTS
• MacOS X 10.6

Tested with OpenGL, DD/D3D
Preferably this can be further tested by others.

____

#### Notes

Consider the OS' versions as some are dated and might not be an accurate representation of the userbase.
I found OpenGL manages to retain the bitmap better and continue displaying (correct) information whereas DD/D3D in comparison fails to do the equivalent.

Of possible solutions the proposed introduces minimal connascence, or otherwise serves to inform what I've identified as the problem, should a more comprehensive solution be made.

#### Cause
Cascading failure from changes in VRAM

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
+++ b/runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
@@ -24,7 +24,6 @@
  */
 package net.runelite.client.callback;
 
-import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
@@ -388,15 +387,18 @@ public class Hooks implements Callbacks
 		final Image finalImage;
 		if (client.isStretchedEnabled())
 		{
-			GraphicsConfiguration gc = clientUi.getGraphicsConfiguration();
+			GraphicsConfiguration gc = ClientUI.getGraphicsDevice().getDefaultConfiguration();
 			Dimension stretchedDimensions = client.getStretchedDimensions();
 
 			if (lastStretchedDimensions == null || !lastStretchedDimensions.equals(stretchedDimensions)
-				|| (stretchedImage != null && stretchedImage.validate(gc) == VolatileImage.IMAGE_INCOMPATIBLE))
+				|| (stretchedImage != null && stretchedImage.validate(gc) != VolatileImage.IMAGE_OK))
 			{
-				/*
-					Reuse the resulting image instance to avoid creating an extreme amount of objects
-				 */
+				// Only reuse image if it is unsullied else reconstruct new
+				if (stretchedImage != null)
+				{
+					stretchedImage.flush();
+				}
+
 				stretchedImage = gc.createCompatibleVolatileImage(stretchedDimensions.width, stretchedDimensions.height);
 
 				if (stretchedGraphics != null)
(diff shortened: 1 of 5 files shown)


Adam-:
Can you provide us the steps used to reproduce #11878?

runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
```diff
@@ -521,7 +535,7 @@ private void setupSyncMode()
 			log.info("unsupported swap interval {}, got {}", swapInterval, actualSwapInterval);
 		}
 
-		client.setUnlockedFpsTarget(actualSwapInterval == 0 ? config.fpsTarget() : 0);
```
abextm, line 524: this is wrong; the fps target must be zero when vsync is on

runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -388,15 +387,18 @@ public void draw(MainBufferProvider mainBufferProvider, Graphics graphics, int x
...
 			Dimension stretchedDimensions = client.getStretchedDimensions();
 
 			if (lastStretchedDimensions == null || !lastStretchedDimensions.equals(stretchedDimensions)
-				|| (stretchedImage != null && stretchedImage.validate(gc) == VolatileImage.IMAGE_INCOMPATIBLE))
+				|| (stretchedImage != null && stretchedImage.validate(gc) != VolatileImage.IMAGE_OK))
 			{
-				/*
-					Reuse the resulting image instance to avoid creating an extreme amount of objects
-				 */
+				// Only reuse image if it is unsullied else reconstruct new
```
Adam-, line 396: how is the image reused here if the line below this creates a new image? what is the purpose of flushing the image if that is the case?

runelite-client/src/main/java/net/runelite/client/callback/Hooks.java
```diff
@@ -388,15 +387,18 @@ public void draw(MainBufferProvider mainBufferProvider, Graphics graphics, int x
 		final Image finalImage;
 		if (client.isStretchedEnabled())
 		{
-			GraphicsConfiguration gc = clientUi.getGraphicsConfiguration();
+			GraphicsConfiguration gc = ClientUI.getGraphicsDevice().getDefaultConfiguration();
 			Dimension stretchedDimensions = client.getStretchedDimensions();
 
 			if (lastStretchedDimensions == null || !lastStretchedDimensions.equals(stretchedDimensions)
-				|| (stretchedImage != null && stretchedImage.validate(gc) == VolatileImage.IMAGE_INCOMPATIBLE))
+				|| (stretchedImage != null && stretchedImage.validate(gc) != VolatileImage.IMAGE_OK))
```
Adam-, line 394: This is causing the volatile image to be completely reconstructed each frame on my system (Win 10)

Adam-: I determined that the first revalidate() call on a new image always returns RESTORED, which is what causes this to get in a rebuild loop.

Nightfirecat:
This is fixed in bbd51305e807110e0af6513bf7410e32b0be1980, f072c82973d58889abcb3df064ac3efb2eee4deb, 1254f0f0528406f3d155964dd217d74ac23c4270

PR #16062, opened 2022-12-23
Add double sprite widget identifiers

AshleyThew:
Some dialogue widget windows have two sprites, these are not covered by DIALOG_SPRITE.

Not sure on the naming convention, can make adjustments as requested.

Image of an example from the cooking tutor in Lumbridge -> asking about fish and meat.
image


Nightfirecat:
We don't typically add widget definitions unless they are used in the core client. If you need them for a hub plugin, you can use the `client.getWidget(int packedId)` or `client.getWidget(int groupId, int childId)` methods to get a reference to that widget

PR #16064, opened 2022-12-25
Itemreminders

CalumMagee:
Created a plugin that takes in ZoneId + ItemID pairs in the config and sends notifications if the player is in that zone without that item. Inspired by me forgetting to take my DHL into Olm too many times but I made it configurable so others can use it for whatever they wish as well.


Nightfirecat:
[As was labeled on the associated discussion,](https://github.com/runelite/runelite/discussions/16060) this should be created as a hub plugin rather than an addition to the core of RuneLite. You can [follow the steps in the README](https://github.com/runelite/plugin-hub/blob/master/README.md) to get that done.

PR #16074, opened 2022-12-27
timers: Track more durations using varbits

testing-ongithub:
There is one varbit for regular antifire and another varbit for super antifire.


testing-ongithub:
I updated this PR to use varbits for more things than just antifires. 🙂 

Newly added: Divine potions, magic imbue, and most Arceuus spells.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -67,14 +65,14 @@
...
-	DIVINE_BASTION(ItemID.DIVINE_BASTION_POTION4, GameTimerImageType.ITEM, "Divine Bastion", 5, ChronoUnit.MINUTES, true),
-	DIVINE_BATTLEMAGE(ItemID.DIVINE_BATTLEMAGE_POTION4, GameTimerImageType.ITEM, "Divine Battlemage", 5, ChronoUnit.MINUTES, true),
+	DIVINE_SUPER_ATTACK(ItemID.DIVINE_SUPER_ATTACK_POTION4, GameTimerImageType.ITEM, "Divine Super Attack", true),
+	DIVINE_SUPER_STRENGTH(ItemID.DIVINE_SUPER_STRENGTH_POTION4, GameTimerImageType.ITEM, "Divine Super Strength", true),
+	DIVINE_SUPER_DEFENCE(ItemID.DIVINE_SUPER_DEFENCE_POTION4, GameTimerImageType.ITEM, "Divine Super Defence", true),
+	DIVINE_SUPER_COMBAT(ItemID.DIVINE_SUPER_COMBAT_POTION4, GameTimerImageType.ITEM, "Divine Super Combat", true),
+	DIVINE_RANGING(ItemID.DIVINE_RANGING_POTION4, GameTimerImageType.ITEM, "Divine Ranging", true),
+	DIVINE_MAGIC(ItemID.DIVINE_MAGIC_POTION4, GameTimerImageType.ITEM, "Divine Magic", true),
+	DIVINE_BASTION(ItemID.DIVINE_BASTION_POTION4, GameTimerImageType.ITEM, "Divine Bastion", true),
+	DIVINE_BATTLEMAGE(ItemID.DIVINE_BATTLEMAGE_POTION4, GameTimerImageType.ITEM, "Divine Battlemage", true),
```
Nightfirecat, line 75: These should all have `removedOnDeath = false` since they're created and removed via vars now?

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -134,6 +117,17 @@ public class TimersPlugin extends Plugin
...
+	private TimerTimer divineSuperAttackTimer;
+	private TimerTimer divineSuperStrengthTimer;
+	private TimerTimer divineSuperDefenceTimer;
+	private TimerTimer divineRangingTimer;
+	private TimerTimer divineMagicTimer;
+	private TimerTimer divineSuperCombatTimer;
+	private TimerTimer divineBastionTimer;
+	private TimerTimer divineBattlemageTimer;
 	private TimerTimer buffTimer;
 	private TimerTimer remedyTimer;
```
Nightfirecat, line 132: Note to self, this may be better as a `Map<GameTimer, TimerTimer>` now that there's so many of these fields. Don't make this change right now--I'll add a commit to this PR doing that if it seems useful enough.

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
testing-ongithub: I tested this (using a divine ranging potion, iirc), and the var was set to 0 on death.

testing-ongithub: Oh wait now I get what you're saying (the var being set to 0 will result in timer being removed, so `removedOnDeath = true` is unnecessary). I've made the change 👍

Nightfirecat:
Leaving a note here confirming the javadoc of the varbits in this PR; <https://github.com/Joshua-F/cs2-scripts/blob/master/scripts/%5Bproc,buff_bar_get_value%5D.cs2> is still a good reference for spell duration for the various "Spell activeness" and "Spell cooldown" entries.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -346,6 +448,263 @@ else if (staminaTimer == null)
...
+			}
+			else
+			{
+				magicImbueTimer.updateDuration(duration);
+			}
+		}
+
+		if (event.getVarbitId() == Varbits.DIVINE_SUPER_ATTACK && config.showDivine())
+		{
+			if (client.getVarbitValue(Varbits.DIVINE_SUPER_COMBAT) > 0)
```
Nightfirecat, line 513: Rather than comparing to 0, this should be checking whether the value of scb is equal to super attack--if you pot a super attack after a scb, you'll have equal values for super strength, super defence, and scb, but a greater value for super attack as it will persist a little longer. This should be kept in mind for all potions which are combined in other potions.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
testing-ongithub: The changes in these varbit values don't occur at exactly the same time. As a result, checking whether the divine super attack varbit value equals the divine super combat varbit value within the block for the divine super attack varbit change event like you suggested fails.

Luckily, the varbits for the components of combination potions always seem to change before the varbits for the combination potions themselves (e.g. the divine super attack, strength, and defence varbits will change before the divine super combat potion varbit). Therefore, I was able to add equality checks to the combination potion varbit change event blocks instead.

Nightfirecat:
I found one issue while working on this--recasting the arceuus spells which have boolean "Spell activeness" vars doesn't re-up their timers. So unfortunately we'll need to keep the chat message triggers for those.

Nightfirecat:
I've updated this with my refactoring changes, please fetch your branch back down and test this again.

testing-ongithub:
Wow, amazing job. Changes look good to me and work perfectly!

(I had the fix for the PMD violation typed up and was going to suggest it when I was done testing, but you beat me to it 😅)

Nightfirecat:
I've been using this branch in my daily-driver dev branch since my original rework of the PR, and have found only one bug: the overload timer was not being cleared when leaving CoX, even though the var controlling it changed to 0. The reason? `IN_RAID` is set to `0` before the overload timer's var is updated, so [my code was attempting to remove the nightmare zone overload timer instead.](https://github.com/runelite/runelite/commit/8a71e616af750f453fcb89f953bf2c3653331a51#diff-95a39a9aa148d9cfc428db0c18e4eaf0ab4ad3b38cb577c0d9149310c12f07d6R346) To handle that, [I've simply re-added the code which removes the overload timer when `IN_RAID` changes](https://github.com/runelite/runelite/commit/4a6d4643a6ac96dac730a5fda99170206dc2c37d#diff-95a39a9aa148d9cfc428db0c18e4eaf0ab4ad3b38cb577c0d9149310c12f07d6R186).

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -385,59 +344,23 @@ else if (poisonVarp >= VENOM_VALUE_CUTOFF)
 				nextOverloadRefreshTick = tickCount + OVERLOAD_TICK_LENGTH;
 			}
 
-			if (overloadVarb <= 0)
-			{
-				nextOverloadRefreshTick = -1;
```
Adam-, line 281: The `timers: Fix overload duration after timer is removed` commit is actually fixing a bug introduced here, and not from 62fc017d9629b33ce9d48ec75fbd6abbbacb993f. Previous code would assign nextOverloadRefreshTick = -1 on overloadVarb <= 0

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -1481,4 +1175,33 @@ private void removeGameIndicator(GameIndicator indicator)
 	{
 		infoBoxManager.removeIf(t -> t instanceof IndicatorIndicator && ((IndicatorIndicator) t).getIndicator() == indicator);
 	}
+
+	private void updateVarTimer(final GameTimer gameTimer, final int varValue, final IntUnaryOperator tickDuration)
+	{
+		updateVarTimer(gameTimer, varValue, i -> i == 0, tickDuration);
+	}
+	private void updateVarTimer(final GameTimer gameTimer, final int varValue, final IntPredicate removeTimerCheck, final IntUnaryOperator tickDuration)
```
Adam-, line 1175: +newline

runelite-client/src/main/java/net/runelite/client/plugins/timers/GameTimer.java
```diff
@@ -66,14 +66,14 @@
...
-	DIVINE_SUPER_ATTACK(ItemID.DIVINE_SUPER_ATTACK_POTION4, GameTimerImageType.ITEM, "Divine Super Attack", 5, ChronoUnit.MINUTES, true),
-	DIVINE_SUPER_STRENGTH(ItemID.DIVINE_SUPER_STRENGTH_POTION4, GameTimerImageType.ITEM, "Divine Super Strength", 5, ChronoUnit.MINUTES, true),
-	DIVINE_SUPER_DEFENCE(ItemID.DIVINE_SUPER_DEFENCE_POTION4, GameTimerImageType.ITEM, "Divine Super Defence", 5, ChronoUnit.MINUTES, true),
-	DIVINE_SUPER_COMBAT(ItemID.DIVINE_SUPER_COMBAT_POTION4, GameTimerImageType.ITEM, "Divine Super Combat", 5, ChronoUnit.MINUTES, true),
-	DIVINE_RANGING(ItemID.DIVINE_RANGING_POTION4, GameTimerImageType.ITEM, "Divine Ranging", 5, ChronoUnit.MINUTES, true),
-	DIVINE_MAGIC(ItemID.DIVINE_MAGIC_POTION4, GameTimerImageType.ITEM, "Divine Magic", 5, ChronoUnit.MINUTES, true),
-	DIVINE_BASTION(ItemID.DIVINE_BASTION_POTION4, GameTimerImageType.ITEM, "Divine Bastion", 5, ChronoUnit.MINUTES, true),
-	DIVINE_BATTLEMAGE(ItemID.DIVINE_BATTLEMAGE_POTION4, GameTimerImageType.ITEM, "Divine Battlemage", 5, ChronoUnit.MINUTES, true),
+	DIVINE_SUPER_ATTACK(ItemID.DIVINE_SUPER_ATTACK_POTION4, GameTimerImageType.ITEM, "Divine Super Attack", false),
```
Adam-, line 69: are you intentionally flipping all of these to removeOnDeath=false?

Nightfirecat: Yes, for timers controlled by vars, the var change will remove them on death (assuming the effect is actually ended), or it will persist afterwards if the var is not set to 0.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
Nightfirecat: Agreed, I've `fixup`'d this commit into the var consolidation change, now ace6d2e3ba96aa43d5c7be55a15bdade91c16448

PR #16081, opened 2022-12-29
Added chat color configurations for friend's chat members in public chat

nbokmans:
Added chat colors configurations for friend's chat members in public chat. Greatly helps with visibility when you're in crowded areas.


runelite-client/src/main/java/net/runelite/client/chat/ChatMessageManager.java
```diff
@@ -31,6 +31,7 @@
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import java.awt.Color;
+import java.util.Arrays;
```
iProdigy, line 34:
```suggestion
```

to appease CI (unused import)

PR #16085, opened 2022-12-30
world map: Add missing quest start locations

pilino1234:
Adds 4 quests missing from the world map:

- Beneath Cursed Sands
- Land of the Goblins
- Sleeping Giants
- Temple of the Eye


Adam-:
I've made the plugin load the quest icon locations from the world map data in cef466c13a75cd887bcb9b589f953adefd98f155

PR #16116, opened 2023-01-06
item stats: add Kovac's grog

CreativeTechGuy:
Add Item Stats for [Kovac's Grog](https://oldschool.runescape.wiki/w/Kovac%27s_grog). Info taken directly from the wiki.


Nightfirecat:
I wonder whether the attack/range/magic drain is based on your level rather than a fixed number, but that's such a minor detail (and easily fixed later) that I'm not concerned about it now.

CreativeTechGuy:
In my single test with a non-maxed account, it seems accurate. But yeah easy to change if necessary. :)

PR #16151, opened 2023-01-14
Fixing the colour of "Your damage" in the corp cave

nathanmobbs:
Currently, the indicator to show whether the player can receive loot doesn't scale well as teams exceed 4 or more. Dividing the damage dealt by the number of players gives you the average damage dealt, not the minimum damage required to get the loot. This damage required is often significantly greater than that average damage.

These updates make the indicator turn green if there's a reasonable (i.e. greater than 1 in 200) chance that you'll receive the loot, while also displaying this probability of that for the benefit of the player. This works for any number of players.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
@@ -29,6 +29,7 @@ import java.awt.Dimension;
 import java.awt.FontMetrics;
 import java.awt.Graphics2D;
 import javax.inject.Inject;
+import org.apache.commons.math3.distribution.NormalDistribution;
 import net.runelite.api.Client;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
 import net.runelite.api.NPC;
@@ -82,8 +83,11 @@ class CorpDamageOverlay extends OverlayPanel
 		int totalDamage = corpPlugin.getTotalDamage();
 		int players = corpPlugin.getPlayers().size();
 
-		// estimate how much damage is required for kill based on number of players
-		int damageForKill = players != 0 ? totalDamage / players : 0;
+		// estimate the probability of your damage exceeding everyone else's
+		double mean = totalDamage / players;
+		double sd = Math.sqrt(totalDamage / (0.1 * players));
+		double probabilityHigherThanOne = new NormalDistribution(mean, sd).cumulativeProbability(myDamage);
+		double probabilityHigherThanAll = Math.pow(probabilityHigherThanOne, players - 1);
 
 		NPC core = corpPlugin.getCore();
 		if (core != null)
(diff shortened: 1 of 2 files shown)


raiyni:
Don't close and open new PRs. You can edit your description and push new commits to the same branch.

nathanmobbs:
Noted, my bad.

Adam-:
I think the intent of the current code is to show if it is possible (green) to get loot, or not possible (red). Which I think works because if you have less than the average damage you can't get the loot because everyone else would also have to have had less than the average damage, but the sum of everyone's damage would the be < the total damage.

That said I am not opposed to expanding the feature to use some smarter statistics to determine drop chance. But I do not want to add a new dependency on `commons-math3` just for this feature.

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
```diff
@@ -82,8 +83,11 @@ public Dimension render(Graphics2D graphics)
 		int totalDamage = corpPlugin.getTotalDamage();
 		int players = corpPlugin.getPlayers().size();
 
-		// estimate how much damage is required for kill based on number of players
-		int damageForKill = players != 0 ? totalDamage / players : 0;
+		// estimate the probability of your damage exceeding everyone else's
+		double mean = players != 0 ? totalDamage / players : 0;
+		double sd = Math.sqrt(totalDamage / (0.1 * players));
```
iProdigy, line 88: doesn't the division by 0 concern also carry over to this line? & can you explain how you arrived to this formulation?

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
```diff
@@ -82,8 +83,11 @@ public Dimension render(Graphics2D graphics)
...
+		double probabilityHigherThanOne = new NormalDistribution(mean, sd).cumulativeProbability(myDamage);
```
iProdigy, line 89: Library-less cdf approximation you are free to use: https://gist.github.com/iProdigy/f5cbc9fff300f674790fab8dbac9e46e

nathanmobbs:
Some rationale and assumptions behind the formulas: 

- I ran 10 million simulations in R based on the accuracy (~92%) and max hit (45) of a player in Karil's and a z spear, which approximate well to the a normal distribution as the effect of missing a hit becomes negligible. This bar chart shows how that looks after 4 hits
image
- We can fit a normal distributions to approximate the bar chart after any number of hits. That looks like this for 1, 2, 3, and 4 hits for example
image
- We can trivially work out the mean for any team size by multiplying the mean at any number of hits by the number of players
image
- however, we still don't know the standard deviation at any of these data points with more than one player. By fitting a curve, we can approximate this with very good accuracy as shown in this picture
image
- Now we can eliminate entirely eliminate the "number of hits" variable, which we don't know, by interpolating in excel as a sort of reverse lookup to find the correct standard deviation
- It just happens that the curves for these standard deviations are a simple x^2 with a coefficient of 0.1043 * [number of players]

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
nathanmobbs: It does indeed! Well spotted

iProdigy:
You may be able to simplify your analysis / reduce the amount of approximations by observing:

* Each hit is an independent event (conditional on gear)
* If we model the damage dealt in a single hit as a normal, we get `fit_V1` where mean=20.8, sd=14
* When considering the cumulative damage dealt across multiple blows, we simply need to add the normals. So, `fit_V2 = N(20.8, 14^2) + N(20.8, 14^2) = N(41.6, 19.8^2)` (recall: you add variances when summing normals). This can repeated for `fit_V3`, `fit_V4`, and so on.

runelite-client/src/main/java/net/runelite/client/plugins/corp/CorpDamageOverlay.java
nathanmobbs: Added a comment with the derivation and prevented the division by 0

nathanmobbs:
Replaced `commons-math3` dependency with a MathUtil class

PR #16155, opened 2023-01-14
Boss Timer: Add Muspah Boss Respawn Timer

roelof69:
image


runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/Boss.java
```diff
@@ -59,7 +59,8 @@
 	DUSK(NpcID.DUSK_7889, 5, ChronoUnit.MINUTES, ItemID.NOON),
 	ALCHEMICAL_HYDRA(NpcID.ALCHEMICAL_HYDRA_8622, 25200, ChronoUnit.MILLIS, ItemID.IKKLE_HYDRA),
 	SARACHNIS(NpcID.SARACHNIS, 10, ChronoUnit.SECONDS, ItemID.SRARACHA),
-	ZALCANO(NpcID.ZALCANO_9050, 21600, ChronoUnit.MILLIS, ItemID.SMOLCANO);
+	ZALCANO(NpcID.ZALCANO_9050, 21600, ChronoUnit.MILLIS, ItemID.SMOLCANO),
+	MUSPAH(NpcID.PHANTOM_MUSPAH_12080, 30, ChronoUnit.SECONDS, ItemID.MUPHIN);
```
Enriath, line 63: Should probably use the full name for the enum, to match how the other bosses are listed
```suggestion
	PHANTOM_MUSPAH(NpcID.PHANTOM_MUSPAH_12080, 30, ChronoUnit.SECONDS, ItemID.MUPHIN);
```

Enriath (changes requested):
Confirmed the time, it's perfect other than the enum name imo

Nightfirecat:
Can this be expressed in game ticks instead of seconds? I'd like to eventually update all the boss timers to use ticks instead since that is strictly more accurate to what happens ingame

roelof69:
Stole your work in #16156 

Converted all of the timers to their nearest time in ticks (rounded down) & added Phantom Muspah timer

Enriath:
You really shouldn't do that. Night only wanted the Muspah timer to be in ticks since we've confirmed that now, and the other ones need to be looked over and confirmed at some point.

Nightfirecat:
Please keep this PR limited to just the new boss, we'll need to test the other bosses ingame before confirming their respawn times

Nightfirecat:
Added in 8f7888f3f9

PR #16157, opened 2023-01-14
EmoteClue - updated iorwerth clue to include bowfa as well as crystal bow

### LINKED ISSUE #16144: Add Bow of faerdhinen to Iorwerth camp emote clue
The emote clues to bow in Iorwerth's camp both accept the Bow of faerdhinen instead of a crystal bow. The clue plugin should be updated to support all forms of the bow (normal, corrupted, and corrupted + recoloured).

Proof submitted in Discord, and has been verified by another contributor
image

AFurryCat:
Resolves #16144

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
@@ -29,6 +29,7 @@ import java.awt.Color;
 import java.awt.Graphics2D;
 import java.awt.Polygon;
 import java.util.List;
+import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.Getter;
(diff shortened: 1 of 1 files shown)


AFurryCat:
Resolves #16157

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -93,6 +94,12 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
 		ItemVariationMapping.getVariations(PHARAOHS_SCEPTRE).stream()
 			.map(ItemRequirements::item)
 			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection CRYSTAL_BOW_OR_BOW_OF_FAERDHINEN = any("Crystal Bow or Bow of Faerdhinen",
+		Stream.of(ItemVariationMapping.getVariations(BOW_OF_FAERDHINEN_INACTIVE).stream(), ItemVariationMapping.getVariations(NEW_CRYSTAL_BOW).stream())
```
Nightfirecat, line 98: `ItemVariationMapping.getVariations(NEW_CRYSTAL_BOW)` isn't helpful here because it will include [the inactive crystal bow](https://chisel.weirdgloop.org/moid/item_id.html?23985#23970-24000) which does not work for this clue. I recommend using `Stream.of(CRYSTAL_BOW, CRYSTAL_BOW_24123)` instead.

Additionally, to make the formatting clearer here since we're streaming several inputs, I'd like to see this formatted as follows:

```java
private static final AnyRequirementCollection CRYSTAL_BOW_OR_BOW_OF_FAERDHINEN = any("Crystal Bow or Bow of Faerdhinen",
	Stream.of(
		ItemVariationMapping.getVariations(BOW_OF_FAERDHINEN_INACTIVE).stream(),
		Stream.of(CRYSTAL_BOW, CRYSTAL_BOW_24123))
		.reduce(Stream::concat)
		.orElseGet(Stream::empty)
		.map(ItemRequirements::item)
		.toArray(SingleItemRequirement[]::new));

**Review by Nightfirecat — CHANGES_REQUESTED** [review]

File: runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java (hunk 1 above)
Comment (AFurryCat) [reply runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java:98]: Would it make more sense to keep using the ItemVariationMapping for the Crystal bow and remove the inactive one in the stream? Meets your aim then of using the mapper to reduce overhead going forwards.

Do we know for sure if the inactive bowfa version works, otherwise that’ll need the same treatment? Ticket said all versions so assuming so but if the inactive crystal bow doesnt…?

Comment (Nightfirecat) [reply runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java:98]: [Based on this comment](https://github.com/runelite/runelite/issues/16144#issuecomment-1383226914), I'd think the inactive bowfa is also not usable for this clue. I'd recommend filtering the inactive form like is done in 835c2d2918652e5993b2ca5b2ffc08bf7aaaa23f, but leaving the crystal bow detection the same as it was before, since we know that works as-is and doesn't seem to need any variation mapping.

File: runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java [hunk 2]
```diff
@@ -93,6 +94,15 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
 		ItemVariationMapping.getVariations(PHARAOHS_SCEPTRE).stream()
 			.map(ItemRequirements::item)
 			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection ACTIVE_CRYSTAL_BOW_OR_ACTIVE_BOW_OF_FAERDHINEN = any("Crystal Bow or Bow of Faerdhinen",
```
Nightfirecat, line 97: Slightly less wordy var name. Otherwise lgtm
```suggestion
	private static final AnyRequirementCollection ACTIVE_CRYSTAL_BOW_OR_BOW_OF_FAERDHINEN = any("Crystal Bow or Bow of Faerdhinen",
```

AFurryCat: Update committed and rebased into initial commit.

PR #16164, opened 2023-01-15
Timers: Add Saturated Heart

roelof69:
Replaces the imbued heart overlay with a saturated heart if the player has one in their inventory.

image


Nightfirecat:
We're not currently doing this for any other type of boost which can occur from multiple sources (antivenom/poison come to mind) so I don't think it's really useful or needed here, since both hearts use the same cooldown timer

Nightfirecat:
Not to mention, this wouldn't show the "correct" heart if you were to activate a saturated heart, deposit it into your bank, then log out and back in.

roelof69:
Understood

Nightfirecat:
We discussed this a bit in our Discord's development channel, and decided that changing the timer's tooltip to read "Imbued/Saturated Heart" would be fine, if you'd like to make that change

PR #16168, opened 2023-01-16
Fairyring plugin: add tags

### LINKED ISSUE #12388: Suggestion: Fairy Ring Travel Log Nicknames (Tags)

JZomDev:
closes #12388 (found after writing this up)

I didn't see until after I created this that #15968 existed, I had searched for "fairy ring" and only saw the one where the PR only edits the description and not adds the ability to add your own tags. Whoops.

Anyways:

image
image

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/ScriptID.java
+++ b/runelite-api/src/main/java/net/runelite/api/ScriptID.java
@@ -473,4 +473,7 @@ public final class ScriptID
 
 	@ScriptArguments(integer = 3)
 	public static final int POTIONSTORE_DOSE_CHANGE = 6555;
+
+	@ScriptArguments(integer = 5)
+	public static final int FAIRYRINGS_SORT_UPDATE = 402;
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingConfig.java
@@ -28,9 +28,12 @@ import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
-@ConfigGroup("fairyrings")
+@ConfigGroup(FairyRingConfig.CONFIG_GROUP)
 public interface FairyRingConfig extends Config
 {
+	String CONFIG_GROUP = "fairyrings";
+	String CONFIG_GROUP_TAGS = "fairyringtags";
+
 	@ConfigItem(
 		keyName = "autoOpen",
 		name = "Open search automatically",
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
@@ -313,22 +313,47 @@ public class FairyRingPlugin extends Plugin
 				c.setCode(w);
 			}
 
+			if (favorites != null)
+			{
+				for (Widget w : favorites.getStaticChildren())
+				{
+					if (w.getId() == ComponentID.FAIRY_RING_PANEL_SEPARATOR)
+					{
+						continue;
+					}
+
+					// Favorites widgets are pre-allocated and hidden if the max of 4 favorites isn't reached
+					if (w.getSpriteId() != -1 && !w.isSelfHidden())
+					{
+						codeMap.computeIfAbsent(w.getRelativeY(), k -> new CodeWidgets()).setFavorite(w);
+					}
+					else if (!Strings.isNullOrEmpty(w.getName()) && !w.isSelfHidden())
+					{
+						codeMap.computeIfAbsent(w.getRelativeY(), k -> new CodeWidgets()).setDescription(w);
+					}
+					else if (!Strings.isNullOrEmpty(w.getText()) && !w.isSelfHidden())
+					{
+						codeMap.computeIfAbsent(w.getRelativeY(), k -> new CodeWidgets()).setCode(w);
+					}
+				}
+			}
+
 			codes = codeMap.values();
+		}
+
+		// reset the separator widget
+		Widget separator = client.getWidget(ComponentID.FAIRY_RING_PANEL_SEPARATOR);
+		if (separator != null)
+		{
+			separator.setHidden(true);
+			separator.setOriginalY(3);
 		}
 
 		// Relayout the panel
 		int y = 0;
-
-		if (favorites != null)
-		{
-			boolean hide = !filter.isEmpty();
-			favorites.setHidden(hide);
-			if (!hide)
-			{
-				y += favorites.getOriginalHeight() + ENTRY_PADDING;
-			}
-		}
-
+		CodeWidgets lastFavorite = null;
+		boolean hasFavorites = false;
+		boolean hasNormal = false;
 		for (CodeWidgets c : codes)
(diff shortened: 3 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingConfig.java
```diff
@@ -34,10 +34,22 @@ public interface FairyRingConfig extends Config
...
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "enableFairyTags",
+		name = "Enable Tag searching",
+		description = "Enable the ability to add tags to search fairy ring codes by",
+		position = 2
+	)
+	default boolean enableFairyTags()
```
raiyni, line 51: Personally, I think this can be removed in favor it always being enabled. Other people can provide input on this.

raiyni (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -316,10 +354,13 @@ else if (!Strings.isNullOrEmpty(w.getText()))
 				}
 			}
 
+			String finalFilter = filter;
 			boolean hidden = !(filter.isEmpty()
 				|| Text.removeTags(c.getDescription().getText()).toLowerCase().contains(filter)
 				|| code.toLowerCase().contains(filter)
-				|| tags != null && tags.contains(filter));
+				|| tags != null && tags.contains(filter)
+				|| (enableTagSearching && tags != null && getTags(code).stream().anyMatch(s -> s.contains(finalFilter))
```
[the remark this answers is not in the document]
JZomDev: I had though that a set contains would match on exact tags being met rather than a partial matching, so I am checking the set to see if any of the Strings within the Collection of Strings match partially the way I knew how.

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -339,6 +398,17 @@ else if (!Strings.isNullOrEmpty(w.getText()))
 			if (!hidden)
 			{
 				y += c.getDescription().getHeight() + ENTRY_PADDING;
+
+				if (c.getFavorite() != null && c.getFavorite().getSpriteId() == SpriteID.FAIRY_RING_REMOVE_FAVOURITE)
+				{
+					separator.setHidden(false);
+					lastFavorite = c;
+				}
```
Nightfirecat, line 406: We should only show the separator if both favorite and non-favorite entries are shown. As-is, you can find only favorited entries and have a trailing separator afterwards, which doesn't look good.
Screenshot_20240714_110157

I think this is fixed by un-hiding the separator in the "add padding to the first widget after the separator" case

runelite-api/src/main/java/net/runelite/api/ScriptID.java
```diff
@@ -458,4 +458,7 @@ public final class ScriptID
 
 	@ScriptArguments(integer = 2)
 	public static final int MOTHERLODE_HUD_UPDATE = 1634;
+
+	@ScriptArguments(integer = 1)
```
Nightfirecat, line 462: This isn't the correct number of args

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -68,6 +75,7 @@
 )
 public class FairyRingPlugin extends Plugin
 {
+	public static final String CONFIG_GROUP_TAGS = "fairyringtags";
```
Nightfirecat, line 78: Instead of using a completely separate config group for tags, this can use an additional qualifier, eg. `FairyRingConfig.GROUP + ".tags"`, or something of the like. You can reference the bank tags plugin to see how it manages its config.

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -212,17 +239,17 @@ public void onGameTick(GameTick t)
...
+		if (!fairyRingWidgetOpen && (searchInputBoxOpen || tagInputBoxOpen))
 		{
 			chatboxPanelManager.close();
 		}
 	}
 
-	private void updateFilter(String filter)
+	private void updateFilter(final String filter)
 	{
-		filter = filter.toLowerCase();
```
Nightfirecat, line 225: It seems like all callsites are passing strings which are manually lowercased. Is there some reason for this change besides being able to make the parameter to be `final`?

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -364,4 +434,76 @@ else if (!Strings.isNullOrEmpty(w.getText()))
...
+			clientThread.invokeLater(() -> updateFilter(searchInput.getValue().toLowerCase()));
+		}
+	}
+
+	Collection<String> getTags(String fairyRingCode)
+	{
+		return new LinkedHashSet<>(Text.fromCSV(getTagString(fairyRingCode).toLowerCase()));
+	}
+
+	String getTagString(String fairyRingCode)
```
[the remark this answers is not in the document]
Nightfirecat: I agree, this method can be inlined into `getTags()`.

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
JZomDev: I think in one of the commits I made made it so that filter had to be final, I think I can roll that back now if you'd like.

runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -68,6 +76,8 @@
 )
 public class FairyRingPlugin extends Plugin
 {
+	public static final String CONFIG_GROUP = "fairyrings";
```
Nightfirecat, line 79: We usually define this in the config interface, but it's not really too big a deal where it ends up.

PR #16172, opened 2023-01-17
Filter fairy ring favorites

raiyni:
As is, fairy ring favorites aren't searchable and are completely hidden when a search criteria is entered.

Would also be a precursor for https://github.com/runelite/runelite/pull/16168 to allow searching tags on all entries.


JZomDev:
#16168 now includes these changes in it

PR #16181, opened 2023-01-19
Change construction icon.

gitman247:
Change construction icon to match in game icon, as well as the regular icon in the skill_icons folder.


raiyni:
is this image ran through a png optimizer?

gitman247:
@raiyni no, I took the icon from the osrs wiki page here: https://oldschool.runescape.wiki/w/File:Construction_icon.png

I then modified the file size to 16x16 using the ubuntu default image editing program.

Let me know if there's a preferred PNG optimisation tool I can use and I will re-submit the PR

Nightfirecat:
How does this look on the panel compared to the icon we're using now? I wonder if the original motivation behind the different icon was because it stood out better.

psikoi:
I think the small icons came from the OSRS hiscores

although construction looks surprisingly different to the in-game counterpart 🤔

image

Nightfirecat:
With that in mind, I don't think we're interested in updating this icon or any of the other small icons at this time.

lukeanderson93:
Hmm, I can sort of understand the position you’ve taken, but I think that’s a shame. Everywhere else the regular construction logo is used, all over the osrs wiki and in game. Even within the runelite client, the larger logo in the logo folder is correct.

Can I request that you reconsider? Seems a real shame to be inconsistent with the game/wiki and use this jarring old image simply because the antiquated hiscores have not been updated to the correct logo.

gitman247:
I can sort of understand the position you've taken, but I think that's a shame. Everywhere else the standard construction icon is used, all over the osrs wiki and in game. Even within the runelite client in the larger icon folder the correct one is used:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/resources/skill_icons/construction.png
Can I request you reconsider? Seems a shame to be inconsistent with the icon in game and everywhere in the wiki, just to be consistent with the outdated hiscores that probably were just forgotten to be updated.

Nightfirecat:
I'm a bit confused, how can the icon we're using now be incorrect [when it's literally the icon that Jagex uses for the online hiscores](https://secure.runescape.com/m=hiscore_oldschool/overall.ws)?

Some further context on why the small icons are taken from the hiscores:
Screenshot of Discord chat logs from Psikoi, reading: "agility has different colors, but it's the same shape
they were all designed for 13x13 or whatever, they're not just scaled down versions of the skill icons, which tends to look better at small scales
like the cooking icon has 2 bones instead of 4"
Below the chat are the cooking skill icons from the hiscores and the ingame skill menu, with the aforementioned stylistic differences

gitman247:
@Nightfirecat Yes, technically the current icon is the correct or "official" icon for the hiscores.

Said that, as a player it _feels_ like the incorrect icon. Even if it is used on the hiscores, it's not used anywhere else in game or on the wiki. I'm sure it's pretty confusing for new players, even as an existing player it takes me a second or two to realise what skill it is when looking a player up. The in game icon is used everywhere on the wiki, on quest requirements, diaries etc. To my knowledge this is the only place where this alternative icon is used.

As I said, I can understand your desire to be aligned with the hiscores, but in my head this will always feel like the wrong icon and add more confusion for players. Anyway that's just my 2 cents.

As an aside I modified the hiscores and I can't understand why Jagex redesigned the icon for 16x16, looks fine IMO.

image

PR #16184, opened 2023-01-20
Added new config item for loottracker for the warning for deletes of loots

moojustice1:
Added a new config item for the Loot Tracker to enable/disable the delete warning box that appears when you go to delete a loot tracker.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerConfig.java
@@ -138,4 +138,14 @@ public interface LootTrackerConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+			keyName = "showDeleteWarning",
+			name = "Show warning message when deleting items",
+			description = "Enables/disables the warning message for clearing loot sections."
+	)
+	default boolean showDeleteWarning()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


Nightfirecat:
What's the motivation for this change? Users are generally quite unhappy about accidentally removing loot from their tracker as there's effectively no way to undo such a removal, and it's expected to persist unlike something like the XP tracker.

moojustice1:
Hey, so like you said some users won't like this, which is why the safety warning box is on by default. I am part of the other side of users that can't stand it, so I would uncheck this box. 

Since it's a configuration, it can be changed by the individual, meaning if people want it on, they do not have to change anything. But my friends and I can't stand the warning box, so this change removes it for us, who can configure it.

Skretzo:
What about a compromise?
Currently, when the warning box `This will permanently delete "X" loot` pops up the selected/highlighted button is `No`. Maybe this could be changed to instead select/highlight (tab position) the `Yes` button. That way you can use the `Enter` hotkey to quickly delete the loot when the warning box pops up. Cancelling with the `Escape` hotkey or clicking the buttons normally with the cursor will remain the same.

moojustice1:
That could work, and I can update this PR to reflect that. 

I'd like to understand a bit more about why a configuration item is not okay. In my head, giving the users power over what they want is better than doing a comprimise/workaround, but I'd like to understand from your perspective on why we wouldn't want that.

Alexsuperfly:
there was previously a config option someone PRed for a similar issue they had with the aggro timer which just generated a lot of #support traffic, the message is there for a reason, allowing people to elect to turn it off does not work because people dont know what they are doing.

Nightfirecat:
Like Alex said, it's not about "giving users the power", it's that we also support users of the software on top of creating it, and it is in our best interests not to give users metaphorical guns with which to shoot themselves in the foot with.

moojustice1:
Sweet, thank you everyone for the guidance. I've been a bit busy but I'll add the updates that were mentioned here. Thanks again everyone!

PR #16207, opened 2023-01-25
Fake XP drop on zero hitsplash

### LINKED ISSUE #10658: fake exp drops on 0 hits
can there be a feature to show an exp drop with the number 0 if you splash/hit 0, it would be really nice for things like olm and woox walking vorkath

nocommentsgood:
Close #10658 

When enabled, hitting a 'zero hitsplash' will drop a fake XP drop with value of 2 XP. Enabling this disables fake XP drops on tick value set by user. 

Passing value < 2 to XPDROP_DISABLED script second argument doesn't allow the script to fire, hence the value 2.


Nightfirecat:
As mentioned on Discord and [in my comment on the linked issue thread](https://github.com/runelite/runelite/issues/10658#issuecomment-1404658628) this is not a solution we're interested in adding for this problem.

nocommentsgood:
I will be working on moving this to a hub plugin

Arkhena:
> I will be working on moving this to a hub plugin

Did you ever get around to make this hub plugin?

PR #16226, opened 2023-01-28
cannon plugin: obtain information using vars

Macweese:
This PR migrates the tracking to primarily use `VarbitChanged` events.

Changes include
• ~~API: Correcting `WorldArea.hasLineOfSightTo()` to be bidirectional,~~ [~~linked update~~](https://secure.runescape.com/m=news/small-game-updates-and-betas?oldschool=1)
• Cannon overlay: Only highlighting tiles which the cannon has line of sight to.
• Cannon plugin: Use vars.

I did not implement changes to include tracking of cannon expiration or world through vars because those are a bit trickier and not as straight forward as just reading var values alone.

The math for `VarPlayer 3551`, `VarPlayer 4`

Value for the var is determined by
$var = x128^2 + {y \over 128^2} + z128^4 $

For which the coordinates can be extracted by
$\text{x-coordinate} = \lfloor \frac{var}{128^2} \rfloor$
$\text{y-coordinate} = 128^2 (var \bmod 128^2)$
$\text{z-coordinate} = \lfloor \frac{var}{128^4} \rfloor$

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
+++ b/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
@@ -248,8 +248,8 @@ public final class VarPlayer
 	 * -1 = not displayed (including normal trees or trees grown through farming).
 	 */
 	public static final int BUFF_BAR_WC_GROUP_BONUS = 4007;
-  
-  /**
+
+	/**
 	 * The location of the player's dwarf cannon, represented by a bit-packed int of the following format:
 	 * <p>
 	 * 0zzzxxxxxxxxxxxxxxyyyyyyyyyyyyyy<br>
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
@@ -45,6 +45,7 @@ import net.runelite.api.events.VarbitChanged;
 import net.runelite.client.Notifier;
 import net.runelite.client.callback.ClientThread;
 import net.runelite.client.config.ConfigManager;
+import net.runelite.client.config.Notification;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.events.ConfigChanged;
 import net.runelite.client.game.ItemManager;
@@ -131,7 +132,7 @@ public class CannonPlugin extends Plugin
 		if (client.getGameState() == GameState.LOGGED_IN)
 		{
 			clientThread.invoke(this::setFields);
-			clientThread.invokeLater(this::addCounter);
+			addCounter();
 		}
 	}
 
@@ -141,6 +142,7 @@ public class CannonPlugin extends Plugin
 		cannonSpotOverlay.setHidden(true);
 		overlayManager.remove(cannonOverlay);
 		overlayManager.remove(cannonSpotOverlay);
+		state = Cannon.NULL;
 		cannonWorld = -1;
 		cballsLeft = 0;
 		removeCounter();
@@ -194,18 +196,18 @@ public class CannonPlugin extends Plugin
 
 			// VarPlayer.DWARF_CANNON_LOCATION is fired before VarPlayer.CANNON_AMMO when the cannon is picked up
 			// The location will be null before the cannon ammo is emptied to zero, preventing a false empty notification
-			if (!config.showCannonNotifications() || cannonWorldPoint == null || previousAmmoCount != cballsLeft + 1)
+			if (cannonWorldPoint == null || previousAmmoCount != cballsLeft + 1)
 			{
 				return;
 			}
 
 			if (cballsLeft == 0 && previousAmmoCount == 1)
 			{
-				notifier.notify("Your cannon is out of ammo!");
+				notifier.notify(Notification.ON, "Your cannon is out of ammo!");
 			}
 			else if (cballsLeft <= config.lowWarningThreshold() && previousAmmoCount > config.lowWarningThreshold())
 			{
-				notifier.notify(String.format("Your cannon has %d cannon balls remaining!", cballsLeft));
+				notifier.notify(config.showCannonNotifications(), String.format("Your cannon has %d cannon balls remaining!", cballsLeft));
 			}
 		}
 
@@ -332,9 +334,9 @@ public class CannonPlugin extends Plugin
 		int parts = client.getVarpValue(VarPlayer.DWARF_CANNON_PARTS_ASSEMBLED);
 
 		state = parts == 4
-				? Cannon.COMPLETE
-				: parts == 0
-						? Cannon.NULL
-						: Cannon.INCOMPLETE;
+			? Cannon.COMPLETE
+			: parts == 0
+				? Cannon.NULL
+				: Cannon.INCOMPLETE;
 	}
 }
--- a/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
@@ -42,6 +42,8 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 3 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -413,8 +299,34 @@ private void removeCounter()
...
+
+		varps.add(VarPlayer.DWARF_CANNON_LOCATION.getId());
+		varps.add(VarPlayer.DWARF_CANNON_PARTS_ASSEMBLED.getId());
+
+		for (int varp : varps)
+		{
+			vc.setVarpId(varp);
+			vc.setValue(client.getVarpValue(varp));
+			onVarbitChanged(vc);
+		}
```
Nightfirecat, line 330: This seems... highly questionable. What is the purpose to manually calling `onVarbitChanged()` on plugin startup? It's extremely bad form to be calling an event subscriber directly outside of tests.

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
```diff
@@ -231,6 +231,36 @@
 	 */
 	LAST_MINIGAME_TELEPORT(888),
 
+	/**
+	 * The location of the player's dwarf cannon
+	 * <p>
+	 * X-Coordinate = (Value % 128^4) / 128^2<br>
+	 * Y-Coordinate = Value % 128^2<br>
+	 * Z-Coordinate = Value / 128^4
```
Nightfirecat, line 239: While this is an accurate way to describe it, it's not really the best way to think about these values; the values transmitted are bit-packed so that all three coordinate values are stored in a single variable. `128 ^ 4` (which is `268_435_456`) is better represented via bitwise operations as `1 << 28` and `128 ^ 2` as `1 << 14`. All these divisions and modulo operations are similarly able to be represented as bitwise operations:

* x: `(val & ((1 << 28) - 1)) >> 14`
* y: `val & ((1 << 14) -1)`
* z: `val >> 28`

As such, this javadoc (and the code interfacing it) would make much more sense if it were described with that knowledge in mind.

In this case, you could describe `DWARF_CANNON_LOCATION` as a bit-packed int of the following format:

```java
/**
  * The location of the player's dwarf cannon, represented by a bit-packed int of the following format:
  * <p>
  * 0zzzxxxxxxxxxxxxxxyyyyyyyyyyyyyy<br>
  * (that is, z represented by the first 3 positive bits followed by x and y each represented by 14 bits)
  * <p>
  * Getting the value of each of these coordinates can be accomplished [...]
```

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonOverlay.java
```diff
@@ -129,9 +130,13 @@ private void drawDoubleHitSpots(Graphics2D graphics, LocalPoint startTile, Color
 				int yPos = startTile.getY() - (y * LOCAL_TILE_SIZE);
 
 				LocalPoint marker = new LocalPoint(xPos, yPos);
+
+				WorldArea worldArea = new WorldArea(plugin.getCannonWorldPoint(), 1, 1);
+				boolean hasLineOfSight = worldArea.hasLineOfSightTo(client, WorldPoint.fromLocal(client, marker));
```
Nightfirecat, line 135: I don't understand what this code is for. Does this fix some bug that exists in the cannon plugin now? If so, that should be a separate commit, if not a separate PR entirely.

runelite-api/src/main/java/net/runelite/api/coords/WorldArea.java
```diff
@@ -665,18 +662,15 @@ else if (thisY >= otherY + other.getHeight() - 1)
 		{
 			return false;
 		}
-		return sourceTile.hasLineOfSightTo(targetTile);
+		return sourceTile.hasLineOfSightTo(targetTile) || targetTile.hasLineOfSightTo(sourceTile);
```
Nightfirecat, line 665: Does this cause any change of behavior in its current uses? (looks like only the dev tools' scene overlay uses it)

What game behavior does this code change model compared to the original? I would not expect this code to need any change now.

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -91,48 +87,14 @@
 	private OverlayManager overlayManager;
 
 	private final VarbitChanged cannonAmmoChanged = new VarbitChanged();
+	private final VarbitChanged cannonPartsAssembled = new VarbitChanged();
```
Nightfirecat, line 90: This looks unused?

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonOverlay.java
Nightfirecat: Fixing this is fine, but it's unrelated to tracking cannon from vars; please isolate this change to its own commit.

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -413,8 +295,33 @@ private void removeCounter()
...
+		cannonWorldPoint = getCannonWorldPoint(client.getVarpValue(VarPlayer.DWARF_CANNON_LOCATION));
+		cannonWorld = cannonWorldPoint == null ? -1 : client.getWorld();
+
+		int parts = client.getVarpValue(VarPlayer.DWARF_CANNON_PARTS_ASSEMBLED);
+
+		state = parts == 4
+				? Cannon.COMPLETE
+				: parts == 0
+				? Cannon.NULL
+				: Cannon.INCOMPLETE;
```
Nightfirecat, line 325:
```suggestion
		state = parts == 4
			? Cannon.COMPLETE
			: parts == 0
				? Cannon.NULL
				: Cannon.INCOMPLETE;
```

runelite-client/src/main/java/net/runelite/client/plugins/cannon/CannonPlugin.java
```diff
@@ -413,8 +295,33 @@ private void removeCounter()
 		counter = null;
 	}
 
-	private static WorldArea buildCannonWorldArea(WorldPoint worldPoint)
+	private WorldPoint getCannonWorldPoint(int var)
```
Nightfirecat, line 298: It's rather confusing to overload the getter of `cannonWorldPoint` like this, a different method name would be helpful to disambiguate them. Additionally, the passed value is always `client.getVarpValue(VarPlayer.DWARF_CANNON_LOCATION)` so it would be better to do that call in this method instead of passing the value in.

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -141,7 +102,7 @@ public void testThresholdNotificationShouldNotify()
 		when(config.showCannonNotifications()).thenReturn(true);
 		when(config.lowWarningThreshold()).thenReturn(10);
 
-		cannonAmmoChanged.setValue(30);
+		cannonAmmoChanged.setValue(11);
```
Nightfirecat, line 105: Is this test value changed for some particular reason?

runelite-client/src/test/java/net/runelite/client/plugins/cannon/CannonPluginTest.java
```diff
@@ -168,23 +129,23 @@ public void testThresholdNotificationShouldNotifyOnce()
...
 
-		cannonAmmoChanged.setValue(30);
+		cannonAmmoChanged.setValue(22);
 		plugin.onVarbitChanged(cannonAmmoChanged);
-		cannonAmmoChanged.setValue(10);
+		cannonAmmoChanged.setValue(0);
 		plugin.onVarbitChanged(cannonAmmoChanged);
 
-		verify(notifier, never()).notify("Your cannon has 10 cannon balls remaining!");
+		verify(notifier, never()).notify("Your cannon is out of ammo!");
```
Nightfirecat, line 138: This test change causes the test name to be completely inaccurate. Also, this ends up being exactly the same as `testCannonOutOfAmmo()` below with your changes.

Nightfirecat (changes requested):

PR #16234, opened 2023-02-01
attackstyles: Fix longrange index for weapon type 28

Skretzo:
The `LONGRANGE` attack style in weapon type 28 is at index 3, not 2. This fix stops the plugin overlay from displaying the wrong attack style after silently breaking at https://github.com/runelite/runelite/blob/5021e6db9ff63c6d54481151be6307840bfce6c1/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java#L242


Nightfirecat:
What weapons have type 28?

Skretzo:
Good question. I think it is the Tumeken's shadow, which I don't have, so I tested this with:
`::setvarb 357 28`
`::setvarp 43 3`
and observed that the longrange attackstyle was highlighted in red.

Nightfirecat:
Added in 1940047215. Thank you!

PR #16237, opened 2023-02-02
[Skill Calculator] Set Hallowed Sepulchre Floor Experience to Cumulative.

larsvansoest:
Except for floor 1, repeating a single floor in the Hallowed Sepulchre is impossible. Namely, each floor must be completed to continue to the next. Therefore, the required laps per floor shown on the skill calculator are not intuitive. 

As a solution, this PR lists the floor experience cumulatively. All starting at floor 1, the entries now denote the total experience after completing each floor. Finally, the word `Floor` is replaced by `Lvl` because of the maximum activity name size (`Hallowed Sepulchre Floor 1-4` would show as `Hallowed Sepulchre Floor ...`).

#### Before
image

#### After
image


Nightfirecat:
I think the original intent was to shift-click to select multiple floors when running a specific floor in the sepulchre. I'm not really sure which way would be less confusing at the moment

larsvansoest:
> I think the original intent was to shift-click to select multiple floors when running a specific floor in the sepulchre. I'm not really sure which way would be less confusing at the moment

I wasn't aware of the shift-click functionality. In that case, this change isn't necessary indeed.

PR #16243, opened 2023-02-03
timers: Reset var timers when duration increases

testing-ongithub:
This allows timers based on varbits to turn red at the correct time even when refreshed early.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -1196,6 +1196,12 @@ public class TimersPlugin extends Plugin
 		{
 			varTimers.put(gameTimer, createGameTimer(gameTimer, duration));
 		}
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -1196,6 +1196,11 @@ else if (timer == null)
 		{
 			varTimers.put(gameTimer, createGameTimer(gameTimer, duration));
 		}
+		else if (timer.getDuration().compareTo(duration) < 0)
```
Nightfirecat, line 1199: So, there's one issue with this approach: server ticks do not have a consistent length. Ideally they should fire every 600ms at exactly 600ms intervals but, as we've seen in the past on total level worlds or worlds that are overloaded, tick length varies pretty dramatically. With that in mind, I'd want there to be some leeway so we're not just regularly replacing timers when there's a minuscule time difference between the expected and actual remaining duration.

PR #16246, opened 2023-02-05
Added capability to display damage type on screen. Also renamed all Weapon Type enum entries to be their actual names, and not numbers.

Ben-Colwell:


Ben-Colwell:
Capture

Nightfirecat:
I'm not sure whether we want to display weapon attack style in this plugin (maybe that would be better suited as a hub plugin?) but I had recently worked on updating the weapon type names in https://github.com/Nightfirecat/runelite@weapontype-names, specifically Nightfirecat/runelite@7cb638c6e6a3f37bbff27385fa24d19ca73f510e, which I think would be better for now--that commit uses names [which match those given on the wiki's weapon types page](https://oldschool.runescape.wiki/w/Weapons/Types) and ensures the ID is still visible in the entries (for ease of lookup)

Ben-Colwell:
I figured we should show it in this plugin since it basically already showcases 1/2 of the combat stats i.e. aggressive, accurate etc. It just kinda seems to me, to make sense to include it in here. Especially since everything to do so, is already here. 

I do agree that we should have the int still available like on your fork. Maybe we combine both of them?

The main reason I think we should include the damage style is because there is no easy way to get it using the API and varbits. This would enable anyone to very easily grab that information. Although, maybe it should be added to the api somehow?? I'm really not sure

Nightfirecat:
I mentioned this in Discord but I'll repeat it here for visibility--the intent of the plugin was to hide attack styles which would give xp in styles a player didn't want to train. That has ~nothing to do with the damage type of the attack style, and I don't really see this being something most users would care about. (To me, this seems like a good candidate for a hub plugin)

That said, I think fleshing out `WeaponType` in this way could be worthwhile if we end up making it a public API for the sake of fixing #6479 or something of the sort.

PR #16251, opened 2023-02-05
timers: Re-add message-detection for thrall creation timer

### LINKED ISSUE #16288: Missing Thrall Timer
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

I casted a thrall and I did not get a new thrall timer.

### Screenshots or videos

https://streamable.com/677174

https://streamable.com/ckvvv0
At the beginning of this clip you can see that a previous thrall timer was just ending, so it should be clear that I have Arceuus spells duration enabled in the base Timers plugin: 
image

### RuneLite version

RuneLite version: 1.9.10
Oldschool revision: Rev 211
Launcher version: 2.5.0

### Logs

[client.log](https://github.com/runelite/runelite/files/10774963/client.log)
[launcher.log](https://github.com/runelite/runelite/files/10774964/launcher.log)

testing-ongithub:
In most cases, the var for having a thrall will be set to 0 and then back to 1 when a thrall is created, but that is not always the case.

For example, if you summon a thrall in one ToA room (e.g. the hub room) and then summon another in another room (e.g. one of the paths), the var won't be reset to 0 in between, which will lead to the timer expiring early.

Fixes #16288

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
@@ -98,6 +98,8 @@ public class TimersPlugin extends Plugin
 	private static final String STAFF_OF_THE_DEAD_SPEC_MESSAGE = "Spirits of deceased evildoers offer you their protection";
 	private static final String PRAYER_ENHANCE_EXPIRED = "<col=ff0000>Your prayer enhance effect has worn off.</col>";
 	private static final String SHADOW_VEIL_MESSAGE = ">Your thieving abilities have been enhanced.</col>";
+	private static final String RESURRECT_THRALL_MESSAGE_START = ">You resurrect a ";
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Is the var change fired on the same tick as the chat message? I think it's not really worth keeping the var change detection for this when the chat message also reliably fires when a thrall is summoned.

testing-ongithub:
> Is the var change fired on the same tick as the chat message?

Yes it is, at least as far as I've seen in my testing.

> I think it's not really worth keeping the var change detection for this when the chat message also reliably fires when a thrall is summoned.

I had a similar thought. However, I ultimately kept it in with this PR for consistency with how the Liquid Adrenaline and Shadow Veil timers are created with both var changes and chat messages due to similar issues with the var changes. If the var-based timer creation should be removed here, should it be removed for those too?

testing-ongithub:
Updated based on feedback from discord.

RonnyTommy548:
Any update on this? This is still an issue

PR #16262, opened 2023-02-10
Show users remaining buttons to solve light box

HickeyMW:
Displaying "Solution: unknown" can make it seem like the solver doesn't know how to find the solution. Displaying remaining buttons to press alleviates that problem and makes it clear to the user how to get the solution.


Felanbird (approved):
I believe this PR works exactly as expected

1.9.11:
java_2023-02-21_21-42-04


this PR (that I've updated to current master):
java_2023-02-21_21-44-10

PR #16267, opened 2023-02-12
add option to screenshot wilderness loot chest when opening

di-brian:


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -680,6 +689,12 @@ else if (killType == KillType.COX_CM)
 				clueNumber = null;
 				break;
 			}
+			case WILDERNESS_LOOT_CHEST:
+			{
+				fileName = "Key";
```
Adam-, line 694: this should probably actually be `Loot key` right?

PR #16283, opened 2023-02-17
feat: optionally highlight others in pvp areas

MikeBrasser:
Added a dropdown menu to the highlight others feature of the 'player indicator' plugin.

Previously it was only possible to highlight all the other players. Not dependent on whether you're in a PvP area or not.

Now it is possible to only enable it whilst being in PvP or the wilderness via this new feature.

The dropdown menu holds three enum values:

"Disabled" -> There will be no highlighting of other players.
"Everywhere" -> All other players will be highlighted. Just like it previously did. (Before it was a simple boolean toggle)
"Only in PvP" (NEW) -> Other players will only be highlighted whilst you are in a PvP area.


kylebot0:
Very pog

Adam-:
added in 2c7f0ec1e2d6602209cdd013a084a6b8f4caa1a7

PR #16290, opened 2023-02-18
world switcher: use a sticky table header

jplsek:
Use a sticky table header for the World Switcher.

Before:

[before.webm](https://user-images.githubusercontent.com/2388657/219901216-3193e955-45ce-475e-baa6-859ed1650c72.webm)

After:

[after.webm](https://user-images.githubusercontent.com/2388657/219901226-0bfd6c68-9afc-4c22-88c1-545c95733ced.webm)

(It's been a while since I've done Swing. Let me know if there is a better way to do this to possibly get rid of the `setPreferredSize` and `revalidate`!)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldhopper/WorldSwitcherPanel.java
@@ -37,11 +37,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import javax.swing.JPanel;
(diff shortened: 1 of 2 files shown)


symphonia117:
Did this feature get lost somewhere? This does not appear to be the behavior of the world switcher tab anymore.

Adam-:
It was merged, but then reverted in bbb54ae8cc36ffb4f993e6284bf774154ff94e85. I don't quite recall why, it broke something.

iProdigy:
revert reason: https://github.com/runelite/runelite/issues/18315

Issue #9400, opened 2019-07-19
Slayer Points not updated after spending points


danxgh10:
When you spend slayer points on skipping a task, the info box for slayer doesn't update your points. I assume it's the same for spending points on other things but have not checked.

Serpa88:
I was unable to reproduce the issue, I have tried cancelling my task, but it updated my points properly.
Please include better steps on how to reproduce the issue, if you have anything else to add.
Just for reference the important way the points are updated are when the rewards window is open or after you finish a task. This means after you cancel a task your points are updated in the next game tick as long as the rewards window is still open.

Adam-:
fixed in 1f6ae893d732cf3a022fe6d9662cd2a42ff2b217

Issue #7839, opened 2019-02-13
Slayer plugin tracking unrestricted worlds


TheStonedTurtle:
Slayer plugin is tracking points/task in unrestricted worlds and overwriting the real game data.

![https://i.imgur.com/IRF1OCX.png](https://i.imgur.com/IRF1OCX.png)

ShaunDreclin:
This is also an issue for deadman worlds, and for playing on multiple characters.

Do we have a way of saving character-specific configs?

Adam-:
This probably has not been a problem for awhile, ever since we migrated storage of the pts/streak to rsprofile.

Issue #7333, opened 2019-01-13
Show slayer infobox when you don't have a task


vosdev:
When your task is complete, the infobox disappears, and will stay away until you get a new task.

The infobox holds useful information that I would like access to without having to get a slayer assignment. If I am on a 99 slayer streak, I want to make sure I don't go to Turael. You cannot get this information from the game itself without completing a task.

**Describe the solution you'd like**
When you do not have an active task, show the infobox anyways (slayer gem logo) (Optional). It can show your Pts, Streak, and instead of a task name like `Black Demons` it can say : `You need something new to hunt.`

**Describe alternatives you've considered**
Sticky Note on my monitor with the streak number written down on it.

deathbeam:
You can change the timeout in slayer plugin settings

Nachtmerrie:
I already have the timeout set to 9999 minutes. This does not matter. The infobox is only active when you do not have a slayer task. My feature request is for it to be also active (optional) when you do *not* have a slayer task.

PR #16318, opened 2023-02-24
slayer plugin: Refactor plugin

### LINKED ISSUE #6458: Highlight dwarven guards for slayer dwarf task

Macweese:
This PR migrates the majority of extraction for slayer related information to be done through vars, and adds other minor UX improvements. The result is more accurate task, streak and points information.

Features added
• Ensure `!Task` chat command returns response
• Distinguish overlay color between superiors and regular NPCs

Most of the values used were obtained from the cache, some exceptions are mock fields such as `Location.WILDERNESS` which is not "real". I wasn't able to find any vars associated with wilderness slayer, therefore `RSProfile` is used instead.

Some things which may need to be confirmed, either as there was no information on the wiki or it was unclear: 
• Sandcrabs task - it exists as a task value in the cache, but no information is available on the wiki
• Fremennik task location - it exists as a location value in the cache, but no information is available on the wiki
• Wiki claims all ogre variants are valid towards Mogre tasks

Region IDs may need to be checked. The values are presumed, since there is no clear definition for the bounds of the locations nor was any definition for them found, e.g. as enum or var in the cache. As such, the wiki was used as reference for the general location so the accuracy of those may be questionable.

Resolves #6458, #6793, #7333, #7839, #9400, #10646


Nightfirecat:
Can you break out the commits from #16317? We don't want to review both at the same time.

Adam-:
Is the NPC variation data also pulled from the cache?

Adam-:
I am not able to review PRs that are this large when all or most of the changes are in a single commit. If you want to split this apart I would be happy to look at individual parts of the proposed change closer.

Adam-:
A portion of this was added in 1f6ae893d732cf3a022fe6d9662cd2a42ff2b217, if you want to rebase this.

PR #16337, opened 2023-03-01
Clear the screen markers between profile switching

JZomDev:
onConfigChanged adds the new profile's markers


peterfaulker11:
:shipit:

zguilt:
This doesn't work after all.

JZomDev:
The overlays are cleared but the side panel stays as the old config, not sure what redraws them/what order. Was getting some assert error about the redraw when I tried to refix it. 

It's clear this needs more work though & some one else ought to solve it correctly.

Issue #1882, opened 2018-04-24
Clue Scroll Messages


AzureLight:
I noticed when doing a clue scroll that the tips use US flooring convention while the game/UK uses a different convention. Would it be possible to have an option added to swap from US to UK convention at will (specifically, the Wizard Traiborn medium clue instructed you to go to the 2nd floor whereas in the UK we call it the 1st floor)?

jpabeem:
I can relate, maybe just use the British notation everywhere? See also OSRS wiki's [styleguide.](http://oldschoolrunescape.wikia.com/wiki/RuneScape:Style_guide#Floor_numbering)

jorisrs:
I suggest adding a locale option to the client in general so it will affect any plugin that makes use of it.

deathbeam:
Looks like this was fixed in https://github.com/runelite/runelite/pull/2244 . We should not add toggles and just follow the game naming convention

PR #16351, opened 2023-03-06
Fix text in Donovan the Family Handyman medium clue step

oohwooh:
Donovan the Family Handyman is located on the top floor. https://oldschool.runescape.wiki/w/Donovan_the_Family_Handyman


raiyni:
The clue scroll hints are using the British/UK floor numbering system. See https://github.com/runelite/runelite/issues/1882

PR #16358, opened 2023-03-08
world map: added the ability to hide already completed quests from the world map

Puharesource:
Changes made:
* Added a new config option to the WorldMap plugin, which allows the player to hide the quest icon for already completed quests.
* Added a new rs2asm script `RemoveWorldMapIcon.rs2asm` with ID = `500001`, which was also added to ScriptID


RAKEBangalang:
Did something happen to the plugin? As I do not have a toggle to hide completed quests at all

PR #16374, opened 2023-03-14
Fix race between ExternalPluginManager and PluginManager during profile change.

geheur:
EPM and PM race to do their profile change work, because they do not use subscriber priority and do their work on different threads. This pull request makes sure that EPM runs before PM so that external plugins that are not installed in the new profile cannot be turned on by PM. If they are turned on by PM, it can cause plugins that do not shut down cleanly to cause issues like leftover infoboxes or overlays, and makes the profile change take longer.


Adam-:
With the exception of startup, the profile changes actually always already happen on the executor thread. So, this can probably be fixed just by prioritizing the EPM to run before PM.

geheur:
added in 89bfa828c90a0f5b1fbf8d5948bdd8d3c7139adc

Issue #15302, opened 2022-07-09
Menu Entry Swapper: 'Buy 50' op doesn't work for Nightmare Zone


SaffiCauthon:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Microsoft Windows 10 Home

### Bug description

Apologies if this isn't considered a bug, but it goes against the intended functionality as an end-user.

Most other shops in Old School Runescape use 'Buy ?' where `?` is 1, 5, 10, or 50.
The Nightmare Zone shop uses 'Buy-?' instead, replacing the space with a hyphen.

This makes it such that the Menu Entry Swapper "Shop Buy Shift-Click" does not work as intended.

I believe additional logic will need added [here](https://github.com/runelite/runelite/blob/ec9b3dfb84d1a13552721e7e2108f87977b70400/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperPlugin.java#L332) for the "Buy-50" option in this minimenu.

### Screenshots or videos

![Normal Shop](https://i.imgur.com/6zCQlDI.png)
![Nightmare Zone Shop](https://i.imgur.com/348gdgq.png)

### RuneLite version

RuneLite version: `1.8.26`
Oldschool revision: `Rev 206`
Launcher version: `2.4.2`

JordanShackelford:
I tried setting the second argument to "buy 1" || "buy-1" but that didn't seem to work and I'm not sure how to use regex in this case so I did the unelegant solution of copying and pasting all 8 buy and sell swaps and adding the dash lol

Nightfirecat:
This is now possible via custom UI swaps in the menu entry swapper plugin; I recommend using that instead.

PR #16386, opened 2023-03-16
Add Menu Entry Swaps to NMZ Reward Shop

AdamRWeinstein:
Closes Issue #15302
Updates Menu Entry Swapper to allow change from default "buy-1" to "buy-5", "buy-10", or "buy-50" in NMZ Reward Shop


Nightfirecat:
This is now possible via custom UI swaps in the menu entry swapper plugin; I recommend using that instead.

PR #16401, opened 2023-03-21
Fix NPE in InfoBoxManager caused by profile switch.

geheur:
If a profile switch changes an infobox's layer (aka group) by directly modifying the config, InfoBoxManager does not update its layers map to reflect this change. This invalid state causes it to throw NPE if a plugin attempts to remove an infobox because it cannot find it where the config says it is, resulting in that infobox not being removed and in some cases even being duplicated later on.

This PR fixes this issue by making all changes to the layer map happen as a result of onConfigChanged, to make sure that config changes are always reflected in the InfoBoxManager's layer map.

How to replicate:
1. make 2 "new profile"s
2. in profile 2, detach the slayer task infobox and then disable it in the slayer plugin
3. swap between the two profiles and observe that the infobox never goes away and might even duplicate sometimes
n.b. the item charges infoboxes always duplicate when you do this, the slayer task infobox only does sometimes.

The NPE:
```
	at net.runelite.client.ui.overlay.infobox.InfoBoxManager.removeInfoBox(InfoBoxManager.java:180)
	at net.runelite.client.plugins.slayer.SlayerPlugin.removeCounter(SlayerPlugin.java:638)
	at net.runelite.client.plugins.slayer.SlayerPlugin.onConfigChanged(SlayerPlugin.java:475)
	at net.runelite.client.config.ConfigManager.switchProfile(ConfigManager.java:234)
	at net.runelite.client.plugins.config.ProfilePanel.lambda$switchToProfile$7(ProfilePanel.java:666)
	at net.runelite.client.util.RunnableExceptionLogger.run(RunnableExceptionLogger.java:41)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBoxManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/infobox/InfoBoxManager.java
@@ -105,6 +105,10 @@ public class InfoBoxManager
 		{
 			layers.values().forEach(l -> l.getInfoBoxes().forEach(this::updateInfoBoxImage));
 		}
+		else if (event.getGroup().equals(INFOBOXLAYER_KEY))
+		{
+			moveInfoboxLayer(event.getKey(), event.getOldValue(), event.getNewValue());
+		}
 	}
 
 	@Subscribe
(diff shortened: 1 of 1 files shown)


geheur:
Fixed in 9db98e6380c5ea83714e50b06e891e1dd7a37ce8

PR #16414, opened 2023-03-25
Add "Phantom Muspah" to the list of available slayer tasks

### LINKED ISSUE #16409: !Task doesn't work when on a "Phantom Muspah" task
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

When you're on a "Phantom Muspah" Task and do type and submit `!task` in chat it doesn't change to what task you have where others can see. 

Also it might be due to that in the slayer plugin that the monster "Phantom Muspah" Has not been registered. This is due to the icon being the icon of the `enchanted gem`. 

I've checked with the group members with me that doesn't have the task i have... But for them it seems to work fine. Also checked if the option was disabled in but this wasn't the case.

### Screenshots or videos

image
image

### RuneLite version

Runelite version: 1.9.13.1
Launcher version: 2.5.0

### Logs

[client.log](https://github.com/runelite/runelite/files/11054555/client.log)
[launcher.log](https://github.com/runelite/runelite/files/11054561/launcher.log)

albertofp:
Closes #16409

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
@@ -136,6 +136,7 @@ enum Task
 	NECHRYAEL("Nechryael", ItemID.NECHRYAEL, "Nechryarch"),
 	OGRES("Ogres", ItemID.ENSOULED_OGRE_HEAD, "Mogre", "Ogress", "Skogre", "Zogre"),
 	OTHERWORLDLY_BEING("Otherworldly beings", ItemID.GHOSTLY_HOOD),
+	PHANTOM_MUSPAH("Phantom Muspah", ItemID.ANCIENT_ESSENCE),
 	PIRATES("Pirates", ItemID.PIRATE_HAT, "Pirate"),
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/Task.java
```diff
@@ -136,6 +136,7 @@
 	NECHRYAEL("Nechryael", ItemID.NECHRYAEL, "Nechryarch"),
 	OGRES("Ogres", ItemID.ENSOULED_OGRE_HEAD, "Mogre", "Ogress", "Skogre", "Zogre"),
 	OTHERWORLDLY_BEING("Otherworldly beings", ItemID.GHOSTLY_HOOD),
+	PHANTOM_MUSPAH("Phantom Muspah", ItemID.ANCIENT_ESSENCE),
```
raiyni, line 139: We generally use the pet item for bosses

albertofp: Good point, fixed

PR #16436, opened 2023-03-28
Include missing Revenants for Ghosts plugin

### LINKED ISSUE #16430: ![image](https://user-images.githubusercontent.com/6641961/228094792-520c1ab3-3016-44f8-9240-ab9b77feaa1d.png)
image
Hi while we're here, it looks like the code also doesn't include Revenants as listed in the wiki.

_Originally posted by @brasil221 in https://github.com/runelite/runelite/pull/16427#discussion_r1149903107_

Pkz-E:
Closes #16430 
Not sure if it's possible to reuse the Revenants enum so just duplicated it for now. If there is a way I can fix that :)


Nightfirecat:
I think you can simply add a single string "Revenant" and it will work, since the matcher specifically looks for string boundaries (start/end of string) or spaces around the word. If that does work, please add an entry to the plugin test for task matching.

Pkz-E:
> I think you can simply add a single string "Revenant" and it will work, since the matcher specifically looks for string boundaries (start/end of string) or spaces around the word. If that does work, please add an entry to the plugin test for task matching.

You're right, "Revenant" only does pick up the revenants. I added a test case for "Revenant Imp", not sure if you want all of the revenants checked for these cases.

PR #16441, opened 2023-03-29
kourendlibrary: reset state on new gamemessage

LlemonDuck:
From the blog today:
> Players inside the Arceuus library will now receive a message in their chatbox letting them know when the books have shifted locations.

Unfortunately I forgot to take a screenshot of the chat message before resetting my client. But, this was tested using the JShell.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPlugin.java
@@ -234,7 +234,12 @@ public class KourendLibraryPlugin extends Plugin
 	@Subscribe
 	public void onChatMessage(ChatMessage event)
 	{
-		if (lastBookcaseAnimatedOn != null && event.getType() == ChatMessageType.GAMEMESSAGE)
+		if (event.getType() != ChatMessageType.GAMEMESSAGE)
+		{
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/kourendlibrary/KourendLibraryPlugin.java
```diff
@@ -243,6 +244,10 @@ public void onChatMessage(ChatMessage event)
 				lastBookcaseAnimatedOn = null;
 			}
 		}
+		if (event.getType() == ChatMessageType.GAMEMESSAGE && Text.removeTags(event.getMessage()).startsWith("You hear the shifting of books"))
```
Nightfirecat, line 247: Let's make the gamemessage type check a guard clause, since that's common between these if blocks. Also, getting the exact message would be handy here so we can just `.equals()` the strings.

LlemonDuck:
The message is:
image

I have updated the PR with your feedback.

LlemonDuck:
Following discussion on Discord, I've removed the Text.removeTags call by using .endsWith instead. The chat message does not have a closing color tag:
```
2023-03-29 18:41:56 EDT [Client] DEBUG client-patch - Chat message type GAMEMESSAGE: <col=800000>You hear the shifting of books due to a mysterious force...or are you just hearing things?
2023-03-29 18:41:59 EDT [Client] INFO  n.r.c.plugins.kourendlibrary.Library - Library is now reset
```

PR #16461, opened 2023-04-05
fix(screenmarkers): on Profile change clear screenmarkers and rebuild

peterfaulker11:
On Profile change, clear ScreenMarkers, then rebuild ScreenMarkerPluginPanel.


geheur:
This has already been added in 87125a2382c3a87d8ef53ddaab05649e71cd386b

PR #16483, opened 2023-04-11
Switch config invocation cache key strategy

ranseur92:
Switch config invocation cache key to an immutable pair with a cached hashCode.

Should allow config inheritance with negligible overhead.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/ConfigInvocationHandler.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/ConfigInvocationHandler.java
@@ -39,7 +39,7 @@ class ConfigInvocationHandler implements InvocationHandler
 	private static final Object NULL = new Object();
 
 	private final ConfigManager manager;
-	private final Cache<Method, Object> cache = CacheBuilder.newBuilder()
+	private final Cache<ConfigKeyPair, Object> cache = CacheBuilder.newBuilder()
 		.maximumSize(256)
 		.build();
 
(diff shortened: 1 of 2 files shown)


Nightfirecat:
What is the motivation behind this change? Where are you looking to use config inheritance?

ranseur92:
I believe it's a QOL change, My use-case at the moment, is developing multiple plugins in 1 code base, which share the same config, for different instances. for example, seperate config options for gauntlet and corrupted.

`@ConfigGroup(GauntletConfig.KEY)
public interface GauntletConfig extends Config {

	String KEY = "ironlite-gauntlet";

	@ConfigItem(keyName = "strategy", name = "Strategy", description = "", position = 1)
	default GauntletStrategy strategy() { return GauntletStrategy.REGULAR; }

	@ConfigItem(keyName = "primaryWeapon", name = "Primary Weapon", description = "", position = 2)
	default GauntletWeapon primaryWeapon() { return GauntletWeapon.STAFF; }

	@ConfigItem(keyName = "secondaryWeapon", name = "Secondary Weapon", description = "", position = 3)
	default GauntletWeapon secondaryWeapon() { return GauntletWeapon.BOW; }
`

Then I can re-use that config, like so: `@ConfigGroup(CorruptedGauntletConfig.KEY)
public interface CorruptedGauntletConfig extends Config, GauntletConfig {

	String KEY = "ironlite-corrupted-gauntlet";

}
`

Nightfirecat:
I'd argue that is not a strong use case, as this could be a single plugin instead of multiple, either with options which apply to both, or ConfigSections for each gauntlet variety

ranseur92:
> I'd argue that is not a strong use case, as this could be a single plugin instead of multiple, either with options which apply to both, or ConfigSections for each gauntlet variety

I admit, my particular use-case isn't that strong, was hoping it got implemented so it would clean up the code behind my config. 

I do however feel that there will be other use-cases for it, it's a nice QOL change allowing interface inheritance as one would expect it to work, with negligible overhead.  

but I also understand your thoughts/concerns and it seemingly being pointless, and respect your decision either way.

PR #16488, opened 2023-04-13
KeyRemapping Fix

### LINKED ISSUE #16274: Hot key remap doesnt work in chatbox
### Type

Crash or hang

### Operating System

None

### OS Version

_No response_

### Bug description

when a chatbox isnt chat (ie dialogue or notification about something) hot key remap doesnt map, f keys work tho so its not against rules or whatever right ?

### Screenshots or videos

_No response_

### RuneLite version

newest

### Logs

_No response_

jgalego25:
Closes #16274

Changed the KeyRemapping plugin so that it allows inputs while the dialog box is open.
Since the options in the dialog box only accepts number keys, I changed the plugin to work only if the
remap is a letter.
If the dialog box is not open, the plugin should work as intended (numbers and letters).


Adam-:
Please see my comment on the corresponding issue https://github.com/runelite/runelite/issues/16274#issuecomment-1530725381. Your PR looks like it breaks this important functionality.

PR #15288, opened 2022-07-07
Ring of forging fix

Tatters654:
Fixes #12725

Fix ring of forging having wrong charge value after a ring disintegrates while smelting ore. Related discord discussion https://discord.com/channels/301497432909414422/419891709883973642/994434757553299466

PR #16500, opened 2023-04-18
Fix ROF off by one error

### LINKED ISSUE #12725: Wrong charge count on ring of forging after it melts
**To Reproduce**
1. Equip a ring of forging
2. Smelt 140 iron bars
3. See error

**Expected behavior**
Displayed charge count is 140

**Actual behavior**
Displayed charge count is 139

**Additional context**
This is the same bug we used to encounter with expeditious/slaughter bracelets, where the count would be incorrectly set due to multiple chat messages being received in an unexpected order. See below logs:

```
2020-10-30 21:58:59 [Client] DEBUG client-patch - Chat message type SPAM: You smelt the iron in the furnace.
2020-10-30 21:59:01 [Client] DEBUG client-patch - Chat message type GAMEMESSAGE: <col=7f007f>Your Ring of Forging has melted.</col>
2020-10-30 21:59:01 [Client] DEBUG n.r.c.plugins.examine.ExaminePlugin - Got examine without a pending examine?
2020-10-30 21:59:01 [Client] DEBUG n.r.client.config.ConfigManager - Setting configuration value for itemCharge.ringOfForging to 140
2020-10-30 21:59:01 [Client] DEBUG client-patch - Chat message type SPAM: You retrieve a bar of iron.
2020-10-30 21:59:01 [Client] DEBUG n.r.client.config.ConfigManager - Setting configuration value for itemCharge.ringOfForging to 139
```

olligobber:
Fixes #12725 

The issue is that when a ring breaks, the client sends the message for the ring breaking, followed by a message that the ring has been used. Consequently, we need to set the number of charges one higher when we get the break message, so that when the use message arrives and we decrement the number of charges, we end up with the correct amount.

The test for this has been redesigned to send both the break and use messages, simulating actual client behavior, and tests we end up with the correct number at the end.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
@@ -383,8 +383,9 @@ public class ItemChargePlugin extends Plugin
 				{
 					notifier.notify("Your ring of forging has melted.");
 				}
-
-				updateRingOfForgingCharges(MAX_RING_OF_FORGING_CHARGES);
+				
+				// Add one to compensate for the one about to be subtracted by the next action
+				updateRingOfForgingCharges(MAX_RING_OF_FORGING_CHARGES+1);
 			}
 			else if (chronicleAddMatcher.find())
 			{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
@@ -384,7 +384,7 @@ public class ItemChargePlugin extends Plugin
 					notifier.notify("Your ring of forging has melted.");
 				}
 				
-				// Add one to compensate for the one about to be subtracted by the next action
+				// This chat message triggers before the used message so add 1 to the max charges to ensure proper sync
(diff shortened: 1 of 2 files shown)


Adam-:
You've changed the behavior of the plugin, and the test tests the behavior of the plugin, and so the test is failing. It is possible the test is incorrect and testing the wrong behavior, in which case the test would need to be updated. Offhand I do not know if either the test is right or your change is right.

olligobber:
I have updated the test so that it should be correct, but it is still failing.

Alexsuperfly:
#15288 at least explains why the change is being made and has a correct test

olligobber:
I feel like the comments in my code explain why the change is being made. The way they fixed the test is unsatisfactory, as it does not emulate the actual behaviour of the client, so I opted to do the test this way. If #15288 does this better, then we should merge that instead.

dogdogcatdog:
The test is expecting the value 140, but 141 is returned. Because your change of adding + 1 to `MAX_RING_OF_FORGING_CHARGES`. Though I don't think this is the right solution. A conditional to subtracting in the first place I think is more suitable such as skipping if `RING_OF_FORGING_USED_TEXT` is after `RING_OF_FORGING_BREAK_TEXT`.

olligobber:
I don't know why 141 is returned, if you look at the modified test, the ring is used after it is broken, same as in the client, so it should return 140.

I made this change to imitate the way necklaces of binding are coded, so if we make this change to rings of forging, we should make the same change to necklaces of binding

PR #16505, opened 2023-04-18
woodcutting: Add redwood tree at farming guild

### LINKED ISSUE #7404: Farming Guild redwoods don't have the indicator that Woodcutting Guild ones have

SirGirion:
I tested that this draws the axes in a similar fashion to the woodcutting guild trees. I tested various combinations of turning the plugin on/off and confirming that axes still draw.

In the trees enum, the null object IDs are the 4 choppable spots' imposter objects and then the real object IDs are the corresponding choppable objects.

Closes #7404

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTreesOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTreesOverlay.java
@@ -88,7 +88,8 @@ class WoodcuttingTreesOverlay extends Overlay
 		{
 			if (treeObject.getWorldLocation().distanceTo(client.getLocalPlayer().getWorldLocation()) <= 12)
 			{
-				// Check if this is a redwood tree at the farming patch
+				// Redwood trees at the farming guild are multilocs rather than global objects synced for all players
+				// so we must handle them here rather than relying on GameObjectSpawned events
 				final ObjectComposition treeComp = client.getObjectDefinition(treeObject.getId());
 				if (treeComp.getImpostorIds() != null && Tree.findTree(treeComp.getImpostor().getId()) == null)
 				{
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTreesOverlay.java
```diff
@@ -87,6 +87,15 @@ private void renderAxes(Graphics2D graphics)
...
 			{
+				// Check if this is a redwood tree at the farming patch
+				if (client.getObjectDefinition(treeObject.getId()).getImpostorIds() != null)
+				{
+					final Tree tree = Tree.findTree(client.getObjectDefinition(treeObject.getId()).getImpostor().getId());
+					if (tree != Tree.REDWOOD)
+					{
+						continue;
+					}
+				}
```
Nightfirecat, line 98: I don't think you need to check that the found `tree` is not a redwood, generally checking whether it is null should be sufficient I think?
```suggestion
				final ObjectComposition treeComp = client.getObjectDefinition(treeObject.getId());
				if (treeComp.getImpostorIds() != null && Tree.findTree(treeComp.getImpostor().getId()) == null)
				{
					continue;
				}
```

SirGirion: I'm pretty unfamiliar with how imposter objects work so I was trying to guard against somehow the transformed ID being equal to some other tree enum. If this can't happen then your edit is good (and truthfully was my original error check)

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingTreesOverlay.java
```diff
@@ -87,6 +88,12 @@ private void renderAxes(Graphics2D graphics)
 		{
 			if (treeObject.getWorldLocation().distanceTo(client.getLocalPlayer().getWorldLocation()) <= 12)
 			{
+				// Check if this is a redwood tree at the farming patch
```
Nightfirecat, line 91: This would be better written as a brief outlining that redwood trees at the farming guild work differently than any other tree, because they are multilocs (in order to be instanced for individual players) rather than a global object synced for all players.

SirGirion: Updated, let me know if it should be expanded upon further

PR #16509, opened 2023-04-19
add red keris to special counter

AdemTurer:
Add Keris Partisan of Corruption to special counter for use in other plugins (like party defence tracker)


Nightfirecat:
The Keris special attack is not an indefinite effect, so this would be better-suited for a hub plugin to indicate the duration of the effect.

Nightfirecat:
As discussed on Discord, because of the rather different nature of this special attack, we're not looking to add it to the special counter plugin at this time.

PR #16524, opened 2023-04-21
specialcounter: add accursed sceptre

### LINKED ISSUE #16516: Support special attack counting for the Accursed sceptre
Because the Accursed sceptre and Accursed sceptre (a) have a special attack which lowers defense (and magic), it is well-suited to have its special attack tracked by the plugin.

itsTGF:
Closes #16516 

Adds support for Accursed Sceptre & Accursed Sceptre (a) in the Special Attack Counter plugin. Tested at all ranges (1-9 tiles).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -488,8 +488,7 @@ public class SpecialCounterPlugin extends Plugin
 
 	private int getHitDelay(SpecialWeapon specialWeapon, Actor target)
 	{
-		// DORGESHUUN_CROSSBOW is the only ranged wep we support, so everything else is just melee and delay 1
-		if (specialWeapon != SpecialWeapon.DORGESHUUN_CROSSBOW || target == null)
+		if (target == null)
 			return 1;
 
 		Player player = client.getLocalPlayer();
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -474,8 +474,8 @@ private PlayerInfoDrop createSpecInfoDrop(SpecialWeapon weapon, int hit, int pla
 
 	private int getHitDelay(SpecialWeapon specialWeapon, Actor target)
 	{
-		// DORGESHUUN_CROSSBOW is the only ranged wep we support, so everything else is just melee and delay 1
-		if (specialWeapon != SpecialWeapon.DORGESHUUN_CROSSBOW || target == null)
+		// ACCURSED_SCEPTRE and DORGESHUUN_CROSSBOW are the only ranged wep we support, so everything else is just melee and delay 1
```
Nightfirecat, line 477: Since it's no longer a single special case, this should be a flag on the weapon and be checked here, eg:

```java
if (!specialWeapon.isRanged() || target == null)
```

Or perhaps the whole method for calculating delay should be a field of the weapon instead of written out in this file.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
itsTGF: I've went ahead with just adding an isRanged flag to Special Weapon

Adam-: > Since it's no longer a single special case, this should be a flag on the weapon and be checked here, eg:
> 
> ```
> if (!specialWeapon.isRanged() || target == null)
> ```
> 
> Or perhaps the whole method for calculating delay should be a field of the weapon instead of written out in this file.

Just having an `isRanged` seems worse than the original code because it becomes extremely non-obvious that setting isRanged=true requires also adding a case to the switch here. I would either keep the original code or move the cycle computation to a method on the enum.

Nightfirecat:
I think this commit should be working as intended, and follow good code practice.

JohnLHa:
@Adam- is there any update on this PR? Would love any update to the special counter adding in mage spec weapons.

Adam-:
My only comment would be the distance function is almost certainly wrong. Though maybe it works in practice. I would prefer if we had the real one instead.

Issue #16512, opened 2023-04-20
Motherlode Mine's node timers inaccurate


JShenobi:
### Type

Crash or hang

### Operating System

None

### OS Version

Win10

### Bug description

The circular timers showing when a depleted node in MLM will respawn have not been updated to reflect the new times as of the March 15 2023 update: https://oldschool.runescape.wiki/w/Update:The_Last_of_Poll_78

It appears the new time, at least for downstairs, is only 60s.

### RuneLite version

Runelite Version 1.9.15.2
OSRS Rev 213
Launcher Version 2.1.6

Fillz:
I've been doing some tests, mostly upstairs, in the motherlode mine. It seems, as you said, that the respawn time is 60 seconds most of the time. However, sometimes it is shorter. The shortest respawn time I've observed is around 32 seconds after the ore vein is depleted. During my tests, the respawn time has never been more than 60 seconds.

ShayanImran:
What would be the expected behavior here? From my understanding the timer was hard-coded is that correct?

JShenobi:
Correct, there is no configuration in the plugin for the timers. 

The expected behavior would be the node respawning as the timer-circle fills up completely, instead of respawning and removing the circle when the circle is only a third full.

Fillz:
@JShenobi That is partly correct. I have been looking through the code, and from what I can tell this is how it works:

There are two hard coded values. A max respawn time value (the max amount of game ticks it will take for an ore-vein to respawn) and a min respawn time value. This is because the ore-vein can actually respawn faster than the max-value and according to comments in the code, the min value is based on the developer's observations.

So what happens is that the pie timer you see on top of the rock in-game will start ticking when the ore-vein despawns, and will complete when the max time is reached. BUT! It will re-color the pie timer from yellow to green when the min time is reached (the pie timer has not completed yet at this point), indicating that the ore-vain can potentially respawn during the remaining time window. In the current code, the max time is 277 game ticks (~166 seconds) and the min time is 150 game ticks (90 seconds).

As I wrote in my first comment, it now seems that the max respawn time is 60 seconds and I have observed (as I assumed the first developer did) a minimum time of around 32 seconds. A solution could be to change the max/min timers to these values.

I'll update the code and make a pull request after I've had time to test this more. I encourage others to test this too and report what they find.

Adam-:
[...]

PR #16542, opened 2023-04-26
Mining plugin: Improvement of ore-vein respawn timers in mlm

Fillz:
The Mining plugin now keeps track of spawn events for the ore-veins in mlm and stores the spawn times. It uses these stored spawn times to calculate the respawn time for each individual wall. It is accurate within 5 ticks. 

I also made the pie timer green during the last 5 ticks of the animation, to indicate that it is about to spawn. From my own testing, it seems like the ore-vein respawns either exactly when the timer turns green - or exactly when the timer is done. But maybe it can respawn within that window as well.

This is an improvement on issue #16512, which was first solved by pr #16534.


Adam-:
fixed in 8550230de72fc37e751d70fa066f5afdc1670fd9

PR #16565, opened 2023-05-08
Item charges Plugin: Fix explorer's ring charges bug

### LINKED ISSUE #16550: [Item Charges] Explorer Ring charge count doesn't update on Check
### Type

Incorrect behavior

### Operating System

None

### OS Version

_No response_

### Bug description

Using the check option on the ring doesn't update the charges properly.  Only tested with explorers ring 4, I don't have access to other versions.

How to desync charge count for testing:
Turn off Item Charges plugin.
Use one or more alchemy charges on explorers ring.
Turn on Item Charges plugin.
Item Charges will still show the original number of charges. It properly updates after alchemy is used again.

### Screenshots or videos

image

### RuneLite version

1.9.15.3
2.6.3

### Logs

_No response_

Fillz:
Close #16550

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
@@ -40,6 +40,7 @@ import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.EquipmentInventorySlot;
+import net.runelite.api.GameState;
 import net.runelite.api.InventoryID;
 import net.runelite.api.Item;
 import net.runelite.api.ItemContainer;
@@ -197,6 +198,10 @@ public class ItemChargePlugin extends Plugin
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		if (client.getGameState() == GameState.LOGGED_IN)
+		{
+			clientThread.invokeLater(() -> updateExplorerRingCharges(client.getVarbitValue(Varbits.EXPLORER_RING_ALCHS)));
+		}
 	}
 
 	@Override


runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -197,6 +197,8 @@ ItemChargeConfig getConfig(ConfigManager configManager)
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		// Update explorer ring charges with value retrieved from varbit value
```
Nightfirecat, line 200: What does this comment clarify? This reads exactly the same as the following line, translated to a sentence.

Fillz:
@Nightfirecat code updated. I no longer get the assertionError, so I assume it is now running on the correct thread.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
```diff
@@ -197,6 +197,7 @@ ItemChargeConfig getConfig(ConfigManager configManager)
 	protected void startUp()
 	{
 		overlayManager.add(overlay);
+		this.clientThread.invokeLater(this::updateExplorerRingChargesFromVarbit);
```
Nightfirecat, line 200: We typically don't use `this.` for referencing fields unless they are shadowed, so this can call `clientThread.invokeLater();` without the extra specifier.

Also, since this code is only called here, it can be an anonymous function instead, ex.

```java
clientThread.invokeLater(() ->
{
	updateExplorerRingCharges(client.getVarbitValue(Varbits.EXPLORER_RING_ALCHS));
});

Comment (Fillz) [reply runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java:200]: Very true, I'll fix that.

Comment (Fillz) [reply runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java]: Fixed now.

**Review by Nightfirecat — COMMENTED** [review]
LGTM. Let's wrap this in a game state check so that plugin startup doesn't fire configmanager warnings when not logged in / on the login screen. (getting the game state can run safely off-thread)

```java
if (client.getGameState() == GameState.LOGGED_IN)
{
	// ...
}
```

Fillz:
@Nightfirecat Done.

PR #16569, opened 2023-05-09
GrandExchangePlugin: Add Option for Notifications on Finished Transactions Only

Samuelmm97:
## Context:
This pull request adds a new configuration option to the Grand Exchange plugin that allows users to specify whether they want to receive notifications for all transactions, or only for those transactions that have finished. 

## Changes:

Added a new configuration option onlyFinishedTransactions in the Grand Exchange plugin configuration. This option defaults to false, preserving the current behaviour of the plugin by default.
Updated the onChatMessage function to check the onlyFinishedTransactions configuration option. If it's true, notifications will only be sent for chat messages that start with "Grand Exchange: Finished buying" or "Grand Exchange: Finished selling".

## Benefits:

Users can now reduce the number of notifications they receive from the Grand Exchange plugin by choosing to only be notified when transactions are completed.
This change improves the usability of the Grand Exchange plugin, making it more user-friendly and customizable.

## Testing:

The changes have been tested locally and work as expected. When the onlyFinishedTransactions option is enabled, the user only receives notifications for completed transactions. When it's disabled, the user receives notifications for all transactions.

Please review the changes and let me know if you have any feedback or questions.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
@@ -47,12 +47,23 @@ public interface GrandExchangeConfig extends Config
 	@ConfigItem(
 		position = 2,
 		keyName = "enableNotifications",
-		name = "Enable Notifications",
+		name = "Notify on offer update",
 		description = "Configures whether to enable notifications when an offer updates"
 	)
 	default boolean enableNotifications()
 	{
 		return true;
+	}
+
+	@ConfigItem(
+		position = 2,
+		keyName = "notifyOnOfferComplete",
+		name = "Notify on offer complete",
+		description = "Configures whether to enable notifications when an offer completes"
+	)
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grandexchange/GrandExchangeConfig.java
```diff
@@ -47,14 +47,25 @@ default boolean quickLookup()
...
+		name = "Notify on offer update",
 		description = "Configures whether to enable notifications when an offer updates"
 	)
 	default boolean enableNotifications()
 	{
 		return true;
 	}
 
+	@ConfigItem(
+		position = 2,
```
Samuelmm97, line 59: Just to double check on line 48 the config is also position = 2, is this an issue?
image

Adam-: It orders them alphabetically then which is fine.

PR #16577, opened 2023-05-11
tile indicators: add true tile fadeout.

geheur:
I chose to make this fade out quadratically over the user-defined fadeout time. This looked good to me for times >= 60 client ticks - the tile doesn't appear to fade when moving for multiple ticks, and looks good fading out. Another option is to hold the tile at full opacity for 1 game tick (30 client ticks), which may be important for people who want < 60 client tick fadeout - otherwise you would see the tile fade out on every tick while moving consecutively over multiple ticks, which might be annoying to look at (although it looks fine to me).


Nightfirecat (approved):
LGTM

geheur:
Thanks for the review (no changes requested, oh my! <3). So, what happens next with this PR?

I did add one more commit to move a couple variable definitions closer to where they are used, and added a comment explaining the gamestate subscriber. Hope that's ok.

LlemonDuck:
A lot of users won't know what a client tick is, I think it would be better to configure the fade delay based on milliseconds since reaching the tile. Then also players using uncapped frame rates could specify sub-client-tick timers.

geheur:
@LlemonDuck Sure, I guess it would be good for people playing under 50fps.

As for uncapped frame rates, I don't see how that impacts client ticks, I thought only very low frame rates result in client tick rates other than 50?

LlemonDuck:
It's useful for both what you mentioned above on the low fps end, but also on the high fps end if a user wanted to specify 550ms, they would be unable to as the client tick system would restrict them to either 540ms or 560ms. I doubt it would even be noticeable, but also I think it just makes more sense to tie rendering to wall clock rather than update cycles.

geheur:
I've added the millisecond timing and made the fadeout more accurate when dealing with game ticks that contain more or less than 30 client ticks, so it should now correctly handle server lag and low framerates.

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -62,4 +81,34 @@ protected void shutDown() throws Exception
...
+
+		if (lastPlayerPosition != null)
+		{
+			if (!playerPos.equals(lastPlayerPosition))
+			{
+				lastTickPlayerMoved = client.getTickCount();
+			}
+			else if (lastTickPlayerMoved + 1 == client.getTickCount())
+			{
+				lastTimePlayerStoppedMoving = System.currentTimeMillis();
```
Nightfirecat, line 98: Let's use `System.nanoTime()` to avoid any user system clock update issues this could cause.

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -45,6 +52,18 @@ public class TileIndicatorsPlugin extends Plugin
 	@Inject
 	private TileIndicatorsOverlay overlay;
 
+	@Inject
+	private Client client;
+
+	@Getter
+	private WorldPoint lastPlayerPosition = null;
```
Nightfirecat, line 59: Object fields default to `null` values anyway, so this doesn't need to be explicitly defined.
```suggestion
	private WorldPoint lastPlayerPosition;
```

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -45,6 +52,18 @@ public class TileIndicatorsPlugin extends Plugin
...
+
+	@Getter
+	private int lastTickPlayerMoved = 0;
```
Nightfirecat, line 62: int fields default to 0 already
```suggestion
	private int lastTickPlayerMoved;
```

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -45,6 +52,18 @@ public class TileIndicatorsPlugin extends Plugin
...
+
+	@Getter
+	private long lastTimePlayerStoppedMoving = 0;
```
Nightfirecat, line 65: long fields default to 0 already
```suggestion
	private long lastTimePlayerStoppedMoving;
```

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -62,4 +81,34 @@ protected void shutDown() throws Exception
...
+
+		lastPlayerPosition = playerPos;
+	}
+
+	@Subscribe
+	public void onGameStateChanged(GameStateChanged e)
+	{
+		if (e.getGameState() == GameState.LOGIN_SCREEN)
+		{
+			// When true tile fadeout is enabled, this prevents the true tile from showing up right after logging in.
```
Nightfirecat, line 110: This should also apply to world hops, and possibly disconnects?

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
```diff
@@ -62,4 +81,34 @@ protected void shutDown() throws Exception
...
+			lastPlayerPosition = null;
```
Nightfirecat, line 111: This should also either be cleared on shutdown, or set on startup. As-is, you can toggle the plugin off, move a few tiles, then turn it back on to see up to 1 tick of it displaying in the wrong location.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
geheur: World hops do not cause your character to move so I don't think the tile will show up when you world hop. On a reconnects, it is advantageous for the tile to pop up if your character has moved since disconnecting.

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsPlugin.java
geheur: I decided to clear it on shutdown.

geheur:
@Nightfirecat I've added your changes (besides the world hop/disconnect thing because I think it's ok as is). Thanks for reviewing :D.

geheur:
I have updated the true tile location update to the render method so that is does not update at a slightly different time than the destination tile at high fps: <https://github.com/geheur/Corner-Tile-Indicators/issues/1>

runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsOverlay.java
```diff
@@ -75,18 +83,51 @@ public Dimension render(Graphics2D graphics)
...
+				{
+					lastTimePlayerStoppedMoving = System.nanoTime();
+				}
 			}
 
-			final LocalPoint playerPosLocal = LocalPoint.fromWorld(client, playerPos);
-			if (playerPosLocal == null)
+			lastPlayerPosition = playerPos;
+
+			if (lastPlayerPosition != null)
```
Nightfirecat, line 101: What does this guard against? Your code currently reads:

```java
WorldPoint playerPos = client.getLocalPlayer().getWorldLocation();
if (lastPlayerPosition != null) {}
lastPlayerPosition = playerPosition;
if (lastPlayerPosition != null)      // this line
```

Can player position actually be null? If so, it probably impacts the code in the first `if` block prior to this one. If not, I don't understand what this check is actually doing.

geheur: The old version of the plugin has this check. However, looking at other core runelite plugins shows that `client.getLocalPlayer().getWorldLocation()` is not null checked in `render` (e.g. `ImplingsOverlay`). I'll remove this.

PR #16600, opened 2023-05-17
Add primordial boots to dragon boots requirement emote step

Jatie1:


ahinkle:
Can you show that this actually works?

Jatie1:
Don't have the step now but I did earlier and it worked, it's what inspired me to make this pull req

PR #16603, opened 2023-05-18
feat: 'always on top' for agility overlay

zmanowar:
Adds a configuration toggle that forces the agility overlay to draw on `ALWAYS_ON_TOP`. This is useful for mouse alignment during black-out scenes.

Enabled:
image

Disabled:
image

Black Scene Example:
image


zmanowar:
Feel free to close if it's not inline with the agility plugin -- can create a external plugin if need be!

Adam-:
As a feature this seems fine, though the use of saveOverlay to trick it into rebuilding the layers seems like a hack. If anything I would just remove+add the overlay since that would be enough and seems more straight forward.

PR #16627, opened 2023-05-26
screenshot: Take level up screenshots on chat message, capitalize captured levels

Nightfirecat:
The level up dialog appears at the same time as the associated chat
message is received, and the latter is more reliable. (Level up widgets
tend not to appear when you are attacked by many monsters at once, for
instance)

Most skills are already capitalized in the level up message and dialog;
the recently-added combat message and dialog, however, does not
capitalize the word. For consistency's sake on our end, let's capitalize
it.


Nightfirecat:
While this does fix the issue of screenshots being missed when combat/other interactions cancel the dialog, this would be problematic if multiple levels are gained at the same time. Closing this for the time being, I'll push the capitalization commit separately for now.

PR #16647, opened 2023-06-04
Warning for Changing Attack Styles

calbright98:
Added a chat message which shows when a player's attack style is changed to an unwanted style.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -28,7 +28,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
 import com.google.inject.Provides;
-import java.util.HashSet;
+import java.util.EnumSet;
 import java.util.Set;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -63,8 +63,8 @@ public class AttackStylesPlugin extends Plugin
 {
 	private int equippedWeaponTypeVarbit = -1;
 	private AttackStyle attackStyle;
-	private final Set<Skill> warnedSkills = new HashSet<>();
-	private boolean warnedSkillSelected = false;
+	private final Set<Skill> warnedSkills = EnumSet.noneOf(Skill.class);
+	private boolean warnedSkillSelected;
 	private final Table<WeaponType, WidgetInfo, Boolean> widgetsToHide = HashBasedTable.create();
 
 	@Inject
@@ -123,12 +123,12 @@ public class AttackStylesPlugin extends Plugin
 	}
 
 	@Nullable
-	public AttackStyle getAttackStyle()
+	AttackStyle getAttackStyle()
 	{
 		return attackStyle;
 	}
 
-	public boolean isWarnedSkillSelected()
+	boolean isWarnedSkillSelected()
 	{
 		return warnedSkillSelected;
 	}
@@ -349,7 +349,7 @@ public class AttackStylesPlugin extends Plugin
 		}
 	}
 
-	private void hideWidget(Widget widget, boolean hidden)
+	private static void hideWidget(Widget widget, boolean hidden)
 	{
 		if (widget != null)
 		{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
@@ -32,6 +32,7 @@ import java.util.EnumSet;
 import java.util.Set;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.GameState;
 import net.runelite.api.ScriptID;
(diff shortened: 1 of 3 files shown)


calbright98:
It's saying something is wrong with my configuration but I can't find the issue. It's saying Configuration Guide Error and "Failed to execute goal goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-cli) on project client: There are test failures". Is there something wrong with the plugin?

geheur:
probably need to update the test to mock/bind the chatmessagemanager.

```
Error:  Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.033 s <<< FAILURE! - in net.runelite.client.plugins.attackstyles.AttackStylesPluginTest
Error:  testHiddenWidget(net.runelite.client.plugins.attackstyles.AttackStylesPluginTest)  Time elapsed: 0.028 s  <<< ERROR!
com.google.inject.ConfigurationException: 
Guice configuration errors:

1) No implementation for net.runelite.client.config.ChatColorConfig was bound.
  while locating net.runelite.client.config.ChatColorConfig
    for the 2nd parameter of net.runelite.client.chat.ChatMessageManager.<init>(Unknown Source)
  while locating net.runelite.client.chat.ChatMessageManager
    for field at net.runelite.client.plugins.attackstyles.AttackStylesPlugin.chatManager(Unknown Source)
  while locating net.runelite.client.plugins.attackstyles.AttackStylesPlugin
    for field at net.runelite.client.plugins.attackstyles.AttackStylesPluginTest.attackPlugin(Unknown Source)
  while locating net.runelite.client.plugins.attackstyles.AttackStylesPluginTest

1 error
	at net.runelite.client.plugins.attackstyles.AttackStylesPluginTest.before(AttackStylesPluginTest.java:72)

Error:  testWarning(net.runelite.client.plugins.attackstyles.AttackStylesPluginTest)  Time elapsed: 0.002 s  <<< ERROR!
com.google.inject.ConfigurationException: 
Guice configuration errors:

1) No implementation for net.runelite.client.config.ChatColorConfig was bound.
  while locating net.runelite.client.config.ChatColorConfig
    for the 2nd parameter of net.runelite.client.chat.ChatMessageManager.<init>(Unknown Source)
  while locating net.runelite.client.chat.ChatMessageManager
    for field at net.runelite.client.plugins.attackstyles.AttackStylesPlugin.chatManager(Unknown Source)
  while locating net.runelite.client.plugins.attackstyles.AttackStylesPlugin
    for field at net.runelite.client.plugins.attackstyles.AttackStylesPluginTest.attackPlugin(Unknown Source)
  while locating net.runelite.client.plugins.attackstyles.AttackStylesPluginTest

1 error
	at net.runelite.client.plugins.attackstyles.AttackStylesPluginTest.before(AttackStylesPluginTest.java:72)
```

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -32,18 +32,18 @@
...
 import javax.annotation.Nullable;
 import javax.inject.Inject;
-import net.runelite.api.Client;
-import net.runelite.api.GameState;
-import net.runelite.api.ScriptID;
-import net.runelite.api.Skill;
-import net.runelite.api.VarPlayer;
-import net.runelite.api.Varbits;
+
+import net.runelite.api.*;
```
Nightfirecat, line 36: Please reverse these wildcard import and extra newline changes. [Per our Code Conventions](https://github.com/runelite/runelite/wiki/Code-Conventions#imports) we generally never use them for non-static imports, and even for static imports don't use them frequently.

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
```diff
@@ -277,7 +280,11 @@ private void updateWarning(boolean weaponSwitch)
 				{
 					if (weaponSwitch)
 					{ // NOPMD EmptyIfStmt
-						// TODO : chat message to warn players that their weapon switch also caused an unwanted attack style change
+						final String msg = new ChatMessageBuilder()
+								.append(ChatColorType.HIGHLIGHT)
+								.append("Switching weapons changed your attack style to an unwanted style!")
```
Nightfirecat, line 285: There are ways to have a different attack style selected via means other than switching weapons (running out of runes for autocasting, entering areas which automatically equip items such as  the gauntlet, etc.), so I'd prefer wording which doesn't necessarily reference that. Also, we know what weapon style is unwanted, so we should print the style in this message. imo this should be formatted:

> Your weapon is using a warned attack style: {STYLE}

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/attackstyles/AttackStylesPlugin.java
calbright98: I've been having some trouble implementing this one. Whenever I switch to a staff that has defensive casting selected it seems to change the attack style to casting then defensive casting. Then when I switch to some other attack style such as Accurate with the staff equipped it switches to Accurate then defensive then Accurate again. I'm not sure what's causing these intermediate changes in the attack style. I get similar results when using the monitoring the VarPlayer.ATTACK_STYLE variable.

Nightfirecat: It's possible then that you'll need to set a flag to check this on the next game tick (after all var changes have been processed) rather than alerting upon var change

calbright98: Would it be bad design to check this flag every game tick?

Nightfirecat: It will effectively be the same either way. Either way you'll end up running:

```java
@Subscribe
void onGameTick(GameTick e)
{
	if (combatStyleVarChanged)
	{
		// ...
	}
}
```

or you'll have some field storing the var and comparing the previous to current attack styles, like what you're doing now. Both are effectively equally efficient and work the same way, so that should be fine.

calbright98 (commented):
Fixed in latest commit

PR #16665, opened 2023-06-09
npcaggro: ensure that the npc aggression name match is case insensitive

pataar:
Currently, names in the 'NPC names' config option aren't working if they are written with caps.


Nightfirecat:
Can you add a test case to show this behavior breaking on master and working after your change is applied?

pataar:
Thanks for the feedback. I couldn't find any testfile yet. Is there any or do I have to create a new one?

pataar:
Nvm, issue seemed to be related to something else.

PR #16676, opened 2023-06-15
clientui: Make "request focus force" work on Linux

Limero:
This makes "Request focus->Force" work on Linux. Previously, the window was never switched to. This fix replicates the behavior of other platforms.

There is another [pending PR](https://github.com/runelite/runelite/pull/15207) for fixing this, but I couldn't get it to work without also setting `setVisible`.

Tested and working on Gnome xorg and Wayland. It works when the window is on the same workspace and not in focus.

I couldn't get it working on Sway, but that is most likely because Sway doesn't support the feature.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -783,6 +783,8 @@ public class ClientUI
 				WinUtil.requestForeground(frame);
 				break;
 			default:
+				frame.setVisible(true);
+				frame.toFront();
 				frame.requestFocus();
 				break;
 		}


33KK:
From `sway(5)`:
```
focus_on_window_activation smart|urgent|focus|none
           This option determines what to do when a client requests window activation. If set to urgent, the urgent state will be set for that window. If set to focus, the window will
           become focused. If set to smart, the window will become focused only if it is already visible, otherwise the urgent state will be set. Default is urgent.
```

Limero:
> From `sway(5)`:
> 
> ```
> focus_on_window_activation smart|urgent|focus|none
>            This option determines what to do when a client requests window activation. If set to urgent, the urgent state will be set for that window. If set to focus, the window will
>            become focused. If set to smart, the window will become focused only if it is already visible, otherwise the urgent state will be set. Default is urgent.
> ```

Thanks! However, I can't get it to work. By default it should mark the workspace as "urgent", which doesn't happen either. So it's like if sway doesn't handle these events.

Blonks-afk:
I can confirm that this fix also works on xorg and wayland with Cinnamon/Mutter (Muffin)

Nightfirecat:
At least on KDE6 using xorg, this change does not work for forcing focus.

Issue #16583, opened 2023-05-12
Default Miner ore/vein timer is off


SuperFly1414:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

I have the timer for mine ores and veins turned on, and so far they are working for all the ores, but when using the timer in the motherload mine, it is off. Prior, it was off by being 15ish seconds too slow, now it is about the same but too fast. The timer ends, but you have to wait roughly 15-20 seconds until the vein respawns.

### RuneLite version

RuneLite Ver. 1.9.16
Launcher Version 2.6.3

Fillz:
I'd be interested to know how to recreate this. I could understand if the timers were too slow, that the ore-vein spawns faster than the timer indicates, but not the other way. I never observed that, and I can't see how it could even happen.

The timers were "fixed" by me in pr #16534, but are far from perfect. This was just a correction of the old approximation of the correct timers - which can be way off.

However, I improved the system for displaying ore-vein timers to more sophisticated in pr #16542. I believe that the timers will have an error margin of 0-3 seconds with that solution.

Shadoman75:
I am also seeing this issue, and had to use a seperate plugin for Motherlode Mine while mining. Timer starts Yellow, goes green about halfway, and stays such until completion, and the timer never extends past that point. Its consistantly 45 seconds after the completion of the timer, that it finally respawns the ore for me.

Adam-:
From testing upstairs and downstairs at the MLM, I very consistently see the veins respawn at 100 ticks +/- maybe 3. The current limits are min=53 max=100 and so this causes it to sort of work since the max time corresponds to when it respawns. I am certainly not able to see any variance anything close to the 79 ticks from #16681,

Fillz:
Hi @Adam-
I did quite a bit of testing on this in april and I am pretty sure I figured out how it works.

Obviously, the current method of respawn timers is flawed and probably off. But I have improved the method of detecting respawn times in PR #16542, so please have a look when you get the time.

To prove that the respawn time can be much shorter than 60 seconds, here are some steps to recreate that scenario:

1. Mine a fresh ore vein
2. As soon as you mine the first pay dirt, stop mining and set a timer for around 35 seconds
3. When 35 seconds has passed, mine the last pay dirt (it should deplete at this point)
4. Start a timer to see the respawn rate.

[...]

PR #16681, opened 2023-06-16
Mining: Update Respawn Timer for Ore Vein at Motherload Mine

codestafa:
Closes https://github.com/runelite/runelite/issues/16583

Before updating:

MLM Before

After updating:

MLM After


Adam-:
fixed in 8550230de72fc37e751d70fa066f5afdc1670fd9

PR #16696, opened 2023-06-21
[MacOS] Closing the side panel doesn't resize the game. #16503

adam-a:
Fixes: https://github.com/runelite/runelite/issues/16503

Tested on Ventura 13.4 --

The problem was within `ContainableFrame.java:contractBy`

At the top exists a check `isFullScreen()` which was returning `false` because the `getExtendedState()` call was returning state `0` (NORMAL -- aka not fullscreen) even when we were in full-screen mode on MacOS.

I'm not really sure why this happens. The extended state wasn't being set correctly on Mac devices and I assume its some Apple integration issue with Java Swing. I'm not an expert on this stuff.

Anyways,

What I did to fix it added a full-screen listener to the window which sets the extended state accordingly. Currently, this is only being done when the client is running on MacOS!

**Demonstration:**

https://github.com/runelite/runelite/assets/12455749/67fdab50-753a-456a-a12b-235650588627

PR #16699, opened 2023-06-22
Move FullScreenAdapter to top level to fix regression

adam-a:
See regression in https://github.com/runelite/runelite/pull/16696


abextm:
fixed in 5d20caf9bb5aced1a9f0f1045dc49c5f62176418

PR #16704, opened 2023-06-24
Complete branch coverage for ColorUtil

neeerp:
Was playing around with the IntelliJ coverage reporter and thought I'd do something helpful in the process. This PR achieves 100% branch coverage in `ColorUtil.java`.

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
@@ -199,7 +199,7 @@ public class ColorUtilTest
 		String INVALID_FORMAT = "ffffff";
 
 		assertEquals(Color.WHITE, ColorUtil.fromString(WHITE_MAX_ALPHA));
-		assertEquals(new Color(0xffffff, true), ColorUtil.fromString(WHITE_ZERO_ALPHA));
+		assertEquals(ColorUtil.colorWithAlpha(Color.WHITE, 0), ColorUtil.fromString(WHITE_ZERO_ALPHA));
 		assertNotEquals(Color.WHITE, ColorUtil.fromString(WHITE_ZERO_ALPHA));
 
 		assertNull(ColorUtil.fromString(TOO_LARGE));
@@ -216,6 +216,7 @@ public class ColorUtilTest
 		assertEquals(Color.BLACK, ColorUtil.fromHex("0x0"));
 		assertEquals(Color.BLACK, ColorUtil.fromHex("#0"));
 
+		assertEquals(ColorUtil.colorWithAlpha(Color.BLACK, 0), ColorUtil.fromHex("0x00000000"));
 		assertEquals(Color.WHITE, ColorUtil.fromHex("0xFFFFFFFF"));
 
 		INVALID_COLOR_HEXSTRING_LIST.forEach((string) -> {
--- a/runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
@@ -168,7 +168,8 @@ public class ColorUtilTest
 			assertTrue(ColorUtil.isHex(hex));
 		});
 
-		INVALID_COLOR_HEXSTRING_LIST.forEach((string) -> {
+		INVALID_COLOR_HEXSTRING_LIST.forEach((string) ->
+		{
 			assertFalse(ColorUtil.isHex(string));
 		});
 	}
@@ -186,7 +187,8 @@ public class ColorUtilTest
 			assertFalse(ColorUtil.isAlphaHex(hex));
 		});
 
-		INVALID_COLOR_HEXSTRING_LIST.forEach((string) -> {
+		INVALID_COLOR_HEXSTRING_LIST.forEach((string) ->
+		{
 			assertFalse(ColorUtil.isAlphaHex(string));
 		});
 	}
@@ -220,7 +222,8 @@ public class ColorUtilTest
 		assertEquals(ColorUtil.colorWithAlpha(Color.BLACK, 0), ColorUtil.fromHex("0x00000000"));
 		assertEquals(Color.WHITE, ColorUtil.fromHex("0xFFFFFFFF"));
 
-		INVALID_COLOR_HEXSTRING_LIST.forEach((string) -> {
+		INVALID_COLOR_HEXSTRING_LIST.forEach((string) ->
+		{
 			assertNull(ColorUtil.fromHex(string));
 		});
 	}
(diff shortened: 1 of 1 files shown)


runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -123,8 +126,66 @@ public void colorLerp()
 	public void colorToHexCode()
 	{
 		COLOR_HEXSTRING_MAP.forEach((color, hex) ->
-		{
-			assertEquals(hex, ColorUtil.colorToHexCode(color));
-		});
+			assertEquals(hex, ColorUtil.colorToHexCode(color)));
```
Nightfirecat, line 129: While we don't have CI enforcing it in tests, we'd prefer it maintains the same code style as the sources, which requires a block for lambdas.

eg.

```java
map.forEach((k, v) ->
{
	// ...
});
```

rather than

```java
map.forEach((k, v) ->
	/* ... */ );
```

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -123,8 +126,66 @@ public void colorLerp()
...
+	public void colorToAlphaHexCode()
+	{
+		COLOR_ALPHA_HEXSTRING_MAP.forEach((color, hex) ->
+			assertEquals(hex, ColorUtil.colorToAlphaHexCode(color)));
+	}
+
+	@Test
+	public void isHex()
+	{
+		COLOR_HEXSTRING_MAP.forEach((_color, hex) ->
```
Nightfirecat, line 142: Iterate over `COLOR_HEXSTRING_MAP.values()` instead of key-value pairs if the key isn't used

Nightfirecat: It would also be helpful to iterate over `COLOR_ALPHA_HEXSTRING_MAP` and test the expected result for those values, and vice-versa for the `isAlphaHex()` test following this.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -123,8 +126,66 @@ public void colorLerp()
...
+
+		assertNull(ColorUtil.fromHex("0x"));
+		assertNull(ColorUtil.fromHex("########"));
+		assertNull(ColorUtil.fromHex("0xABCDEFGH"));
+	}
+
+	@Test
+	public void fromObject()
+	{
+		assertEquals(Color.RED, ColorUtil.fromObject(""));
```
Nightfirecat, line 189: This test isn't useful. We don't want to guarantee that an empty string will yield `Color.RED`. It would be much more helpful to assert the result for a variety of objects is not null.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -123,8 +126,66 @@ public void colorLerp()
...
+		assertEquals(Color.BLACK, ColorUtil.fromHex("#000000"));
+		assertEquals(Color.BLACK, ColorUtil.fromHex("000000"));
+
+		assertEquals(Color.BLACK, ColorUtil.fromHex("0x0"));
+		assertEquals(Color.BLACK, ColorUtil.fromHex("#0"));
+
+		assertEquals(Color.WHITE, ColorUtil.fromHex("0xFFFFFFFF"));
+
+		assertNull(ColorUtil.fromHex("0x"));
+		assertNull(ColorUtil.fromHex("########"));
```
Nightfirecat, line 182: A case which tests the result from something like `"##ABCDEF"` (double hash) would be good.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
```diff
@@ -123,8 +126,66 @@ public void colorLerp()
...
+	@Test
+	public void fromString()
+	{
+		String WHITE_MAX_ALPHA = "-1";
+		String WHITE_ZERO_ALPHA = "0xffffff";
+		String TOO_LARGE = "0xffffffff";
+		String INVALID_FORMAT = "ffffff";
+
+		assertEquals(Color.WHITE, ColorUtil.fromString(WHITE_MAX_ALPHA));
+		assertEquals(new Color(0xffffff, true), ColorUtil.fromString(WHITE_ZERO_ALPHA));
```
Nightfirecat, line 162: The expected value is much clearer to read if it is constructed instead as `ColorUtil.colorWithAlpha(Color.WHITE, 0)`

Nightfirecat (changes requested):

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
neeerp: I see. I had IntelliJ auto-suggest turning Lambdas with statement blocks into expressions. I'm using both the RuneLite CheckStyle and the IntelliJ formatter config, but it seems neither codifies this. 

Do you think this is something that should be added to CheckStyle? I could look into that.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
neeerp: Done both suggestions; also added `INVALID_COLOR_HEXSTRING_LIST` constant with an assortment of strings that should fail both checks.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
neeerp: I've replaced this test with what you've suggested.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
neeerp: That didn't occur to me! I've done that now.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
neeerp: I've created a `INVALID_COLOR_HEXSTRING_LIST` constant that contains such an example, and am iterating over all thsoe examples in these tests now.

runelite-client/src/test/java/net/runelite/client/util/ColorUtilTest.java
Nightfirecat: I wouldn't mind that being added in another PR, but that's best left separate from these changes.

neeerp: Sure, another time then.

Nightfirecat:
Added in 5a38611df30cc55a2770a526c61c98ee5e1141e4

PR #16718, opened 2023-06-29
Keyremapping prompt customizations

macica2:
This update allows users to toggle the prompt text of the Key Remapping plugin between "Press Enter to Chat..." and "*". It also lets the user change the color of this prompt text, which currently is black on opaque chat boxes or white on transparent ones.

This was created based on a user request from [this Reddit comment](https://www.reddit.com/r/2007scape/comments/14f8q32/have_a_good_idea_for_a_rl_plugin_ill_make_it_for/jp08j7b/?context=3). The user mentioned that when he first suggested the idea, he got pushback because users might get confused if they completely hide the prompt text and later can't remember why they can't type. Because of this, I simply made it a toggle instead of a free form text field or a checkbox to hide it entirely. This was years ago, so if we aren't as concerned with that anymore, I can change it to a text field.

This is my first update for RuneLite so please let me know if I should change anything!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingConfig.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.keyremapping;
 
+import java.awt.Color;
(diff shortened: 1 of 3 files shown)


macica2:
Closing since this will be created as a pluginhub plugin instead.

PR #16721, opened 2023-06-29
Add support for Forestry events and additional tree tracking

Infinitay:
# Forestry Events
Added support for various Forestry events in the form of `HintArrow` overlays or tile highlights and labels to help make the events easier or as a QoL

## Rising Roots
- Will display a `HintArrow` on the glowing root
### Additional Comments
- Maybe support for highlighting the regular roots would be nice because they tend to spawn a few seconds before the glowing root

## Flowering Tree
- Will display a `HintArrow` and highlight the two active flowers to be pollinated

### Implementation
1. Checks to see if a player successfully pollinates a flower bush
2. Highlight the respective pollinated bush
- We only keep track of two active bushes

### Potential Issues
- Because I changed how we're tracking the flowering bushes, I had to modify how spawn notification's are sent. That's why there also a _workaround_ with a variable to track whether or not an event was already sent. I haven't tested this because it was getting late. Please test to make sure there are no issues, otherwise I will check when I wake up.
	- Also, this may cause issues where two of the same Flowering Tree event spawns right next to one another. If they are both active and close enough, the second event's spawning won't trigger a notification. Currently, notifications are sent on spawn, and cleared once the bee NPC despawns.

## Struggling Sapling
- Will display a `HintArrow` and highlight on the necessary ingredients and their order
- Will display a label on the tile indicating the order in which it should be added to the mulch

### Implementation
1. Get all of the spawned related events' `GameObject`s such as `Rotting leaves`, etc
	- Used to reference the GameObject and to highlight it
2. Listen to the chat to see which ingredient _the sapling seems to love_" and at what order

### Potential Issues
- As you may have noticed in a TODO comment, there was an issue where the last step wasn't tracked. Though originally it was because I didn't realize there are multiple id's for `Wild mushrooms`. It should be resolved.
- Rarely it wouldn't track ingredients properly. I believe it was because I wasn't properly clearing the variables. With little testing, it seems resolved, but further testing would be appreciated, and a second eye on how I handled clearing the variables in the code.

## Leprechaun
- Will display a `HintArrow` above the leprechaun

# Additional Changes
## # Players Chopping a Tree
- Will display a label on the base of the tree with the respective number

### Implementation
1. Have a map for tree that will keep track of how many players are chopping a tree
2. Have a map for `Player`s that will keep track of which tree a player is chopping
- I use two maps here to potentially speed up code at the cost of allocating an additional map for players. This way instead of counting all the players chopping the tree, we can simply reference the size of the set in the tree map.
4. When a `Player` is chopping a tree, find the closest tree they are facing* and track it accordingly
5. When a `Player` has _spawned_ and they are chopping a tree, find the closest tree they are facing* and track it accordingly
	- Primarily for when a player logins in or walks to a new area where a player is already chopping down a tree, it'll track it.
	- Also helps in cases when you change planes and players get despawned, but the `GameObject` stays. For example, if you go to the top of the bank at Seer's Village, the trees will still be rendered.
6. Loop through all the trees within the tree map and display the respective label of the number of people chopping that given tree
	- Color coded where it'll stay green once it reaches the maximum number of players required for the maximum bonus

_*Closest tree is calculated by distance and orientation of the player_

### Potential Issues
- I've seen some times where the overlay stays. I've also seen times where it automatically fixes itself.
- There might be some issues when first loading into a region and people are already are chopping. Although, this might be resolved. It'd be nice if someone could test it to make sure they don't have the same error.
- I'm reconsidering the signature for the treemap. I think it would be time and space efficient to instead simply use `Set<Integer>` (set of player ids) instead of `Set<Player>`

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -179,9 +179,9 @@ public final class AnimationID
 	public static final int MAGIC_ENCHANTING_AMULET_2 = 720; // emerald, jade, dragonstone
 	public static final int MAGIC_ENCHANTING_AMULET_3 = 721; // ruby, topaz, onyx, zenyte
 	public static final int MAGIC_ENCHANTING_BOLTS = 4462;
-	public static final int BURYING_BONES = 827;
+	public static final int BURYING_BONES = 827; // Also used for Struggling Sapling Forestry event
 	public static final int USING_GILDED_ALTAR = 3705;
-	public static final int LOOKING_INTO = 832; // Generic animation used for filling water vessels, Shades of Mort'ton, etc.
+	public static final int LOOKING_INTO = 832; // Generic animation used for filling water vessels, Shades of Mort'ton, pollinating flower bushes
 	public static final int DIG = 830;
 	public static final int DEMONIC_GORILLA_MAGIC_ATTACK = 7225;
 	public static final int DEMONIC_GORILLA_MELEE_ATTACK = 7226;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingConfig.java
@@ -171,7 +171,7 @@ public interface WoodcuttingConfig extends Config
 		position = 20,
 		keyName = "highlightGlowingRoots",
 		name = "Highlight glowing roots",
-		description = "Highlights glowing roots during Rising Roots events",
+		description = "Highlights glowing roots during the Rising Roots event",
 		section = forestrySection
 	)
 	default boolean highlightGlowingRoots()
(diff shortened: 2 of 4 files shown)


Infinitay:
Changes since [`29b4136` (#16721)](https://github.com/runelite/runelite/pull/16721/commits/29b4136351e0dcbff15d50cf553fceba327c5616)

# Forestry Events
- Removed all usages of `HintArrow`
- Displays all forestry event overlays regardless of having an active woodcutting session or not
	- Brief discussion on [Discord](https://discord.com/channels/301497432909414422/419891709883973642/1124230559741263962)

## Leprechaun
- Now highlights the leprechaun by overlaying a convex hull of the Npc

https://github.com/runelite/runelite/assets/6964154/7ca673fc-57c8-49d5-9af7-ca5bd50ea889


# Players Chopping a Tree
- Refactored to use `Integer` as value for the tree map
	- Instead of storing a set of players, now we're just counting, hopefully some minor memory improvement
- Using `Direction` via `Angle#getNearestDirection` instead of a custom function

# Known issues
## Players Chopping a Tree
- Edge case when a player quickly switches the tree their chopping without canceling their animation

## Struggling Sapling
- Edge case when multiple instances of this event spawns nearby, it will not render the labels on the appropriate tiles
	- For example, if you're at position A at the current sapling event and another sapling event spawns at position B
		- Tiles will be highlighted appropriately
		- Object abbreviation will be labeled appropriately
		- The ingredient step could be mislabeled, instead of labeling the objects at position A where we are, it'll label it on the object at position B
			- I'm not sure if the steps will be the same for that spawn too. If so, we can just apply the label to all of the same object label

## Flowering Tree
- Not confirmed, but I assume it will have the same issue as the struggling sapling when multiple instances of the event are spawned

Infinitay:
Seeing as everything has already been refactored and implemented, I will go ahead and close this. I made a plugin for tracking the number of players on trees and fixed all* of the issues with it. If you want it feel free to download the plugin via the plugin hub when available.

https://github.com/Infinitay/tree-count-plugin

PR #16731, opened 2023-06-30
clues: Support Forestry outfit

Felanbird:
- [X] Full Set
- [ ] Mix & Match

Mentioned in passing to me that a full set does work, but is not labeled as such on the clue.

Asked them to test mix&match if they get the step again, but very likely to work as long as Jagex doesn't randomly grief.


Felanbird:
Received confirmation that mix&match works as expected :+1:

PR #16733, opened 2023-07-01
random events: handle kiss the frog events & add tests

cwjoshuak:
- Adds a set of frog npcs that will be used to hide menu options for.
- Doesn't change initial set of npcs, so notifications aren't affected.
- Modifies menu option logic to remove menu options only when you don't have a random event, but now works across all npcs to hide instead of only one.
- Adds a getter to `currentRandomEvent` so it can be used in tests.
- Adds tests. I referenced some code in MenuEntrySwapper and couple of others.

Not the greatest with tests, if needs more coverage or changes let me know.


cwjoshuak:
frog ids
image

hidden frogs

https://github.com/runelite/runelite/assets/15935660/3b506531-ee16-4243-ab0a-c88370c74192

PR #16749, opened 2023-07-05
config: Add default profile linking to config profiles

Nightfirecat:
This commit adds the ability to mark config profiles as being the
default for a set of rsprofiles, meaning that profiles will change
automatically on rsprofile change. This will likely resolve most cases
for players wanting to use the `--profile` launcher argument, and make
it easier to use game mode-specific, or account-specific settings, as it
can be made automatic.


runelite-client/src/main/java/net/runelite/client/config/ConfigProfile.java
```diff
@@ -48,6 +50,10 @@ public class ConfigProfile
 	@Getter(AccessLevel.PACKAGE)
 	@Setter(AccessLevel.PACKAGE)
 	private long rev;
+	@Getter
+	@Setter
+	@Nullable
+	private Collection<String> defaultForRsProfiles;
```
abextm, line 56: this should be typed more specifically, as this will sometimes be a LinkedHashSet and sometimes be an ArrayList

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
```diff
@@ -221,22 +232,18 @@ private void onProfileChanged(ProfileChanged ev)
 			return;
 		}
 
-		SwingUtilities.invokeLater(() ->
-		{
-			for (ProfileCard card : cards.values())
-			{
-				card.setActive(false);
-			}
+		SwingUtilities.invokeLater(this::reload);
```
abextm, line 235: Why does this need to degrade to a full reload now?

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
```diff
@@ -221,22 +232,18 @@ private void onProfileChanged(ProfileChanged ev)
...
+	private void onRuneScapeProfileChanged(RuneScapeProfileChanged ev)
+	{
+		if (!active)
+		{
+			return;
+		}
 
-			card.setActive(true);
-		});
+		SwingUtilities.invokeLater(this::reload);
```
abextm, line 246: Is there a reason you are running this on the EDT?

runelite-client/src/main/java/net/runelite/client/config/ConfigManager.java
```diff
@@ -1473,6 +1474,34 @@ private void onPlayerChanged(PlayerChanged ev)
...
+	private void onRuneScapeProfileChanged(RuneScapeProfileChanged ev)
+	{
+		final String rsProfileKey = getRSProfileKey();
+		ConfigProfile switchToProfile = null;
+		try (ProfileManager.Lock lock = profileManager.lock())
+		{
+
+			for (final ConfigProfile profile : lock.getProfiles())
+			{
+				final Collection<String> getDefaultForRsProfiles = profile.getDefaultForRsProfiles();
```
abextm, line 1487: var named `get`

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
```diff
@@ -666,6 +724,74 @@ private void switchToProfile(long id)
...
 
+	private void unsetRsProfileDefaultProfile()
+	{
+		setRsProfileDefaultProfile(-1);
+	}
+
+	private void setRsProfileDefaultProfile(long id)
+	{
+		boolean switchProfile = false;
+		try (ProfileManager.Lock lock = profileManager.lock())
```
abextm, line 735: Profile lock should only happen on the executor thread. This should probably live in ProfileManager, not in here

abextm (changes requested):
This breaks the layout so that it flows off the right side of the panel


The link toggle can be come enabled without a rsprofile set, ie while on the title screen, which makes no sense.

Another interesting quirk, which is likely out of scope for this pr, is that when using the Jagex Launcher the client knows what displayname you will log into, but it doesn't know which account hash, and thus rsprofile. So it will likely be confusing to some users that it waits until login to select the correct profile, and does not select it on startup.

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
Nightfirecat: `(create|delete|rename|switchTo)Profile` and `handleDrag` don't create a lock on the executor thread, though I do see `importProfile` and `cloneProfile` do, would you want all of these to run on the executor thread also?

I'm not really sure how this functionality could or should move to ProfileManager... Profile changes trigger no events, so even if all the logic around setting and unsetting a profile as default for some rsprofile was moved into ProfileManager, there wouldn't be a good way to react to the changes it would enact. As-is, `setRsProfileDefaultProfile()` effectively does the same work as `switchToProfile()`, albeit on a different ConfigProfile field.

Nightfirecat:
I've corrected the buttons making the panel overly wide, though I'd be open to a better way to address it than setting minimum/preferred size, if there's some more idiomatic way to do so. I also decided to just hide the link button until we have an rsProfile available to link to. (Adam pitched another idea being to open an error dialog when trying to link to nothing; I think this way would be easier on #support, as we can just tell users to log in first)

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
abextm: I think I am remembering something that I wanted to change but didn't, not sure. This is fine

abextm (approved):

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
```diff
@@ -660,6 +752,77 @@ private void switchToProfile(long id)
...
+					changed = true;
+				}
+
+				if (changed)
+				{
+					lock.dirty();
+				}
+				reload(lock.getProfiles());
+			}
+			finally
```
Adam-, line 816: should this be in finally?

runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
```diff
@@ -660,6 +752,77 @@ private void switchToProfile(long id)
...
+						profile.setDefaultForRsProfiles(new ArrayList<>());
+					}
+					profile.getDefaultForRsProfiles().add(rsProfileKey);
+					switchProfile = true;
+					changed = true;
+				}
+
+				if (changed)
+				{
+					lock.dirty();
```
Adam-, line 812: it is safe to call .dirty() multiple times if you want, so you can just call it each place that it dirties the profiles.

Adam-:
Clicking the "link" button generates a warning when the profile is already in use:
`[pool-2-thread-1] WARN  n.r.client.config.ConfigManager - switching to already-active profile!`

PR #16756, opened 2023-07-06
Change forestry basket bank use option to empty

DominickCobb-rs:
Add option to change "Use" to "Empty" when in bank for Forestry Basket


Nightfirecat:
Use left-click or shift-click remapping in the Menu Entry Swapper plugin for this

DominickCobb-rs:
It would just change the text from Use to Empty, not change the default option as Menu Entry Swapper does. Does Menu Entry Swapper provide functionality already to just change the text of a specific option without re-ordering being involved? I wouldn't mind trying to add it. Might just be better off as a plugin-hub plugin if neither of these two are a fit for it.

PR #16764, opened 2023-07-09
party: allow custom colors

LlemonDuck:
Currently, when using the Party plugin your display color is deterministically chosen by your player name. This can lead to some collisions, like in this case with my friend:

This PR allows users to specify their own colors, for more personal identity, to avoid collisions, or even as an accessibility feature for colorblind users.

image


runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPlugin.java
```diff
@@ -546,6 +556,11 @@ private void checkStateChanged(boolean forceSend)
 			shouldSend = true;
 			update.setVengeanceActive(vengActive);
 		}
+		if (forceSend || memberColor != lastStatus.getMemberColor())
```
LlemonDuck, line 559:
```suggestion
		if (forceSend || !memberColor.equals(lastStatus.getMemberColor()))
```

PR #16766, opened 2023-07-09
chat color: simplify config item text.

geheur:
It is not needed to specify "(transparent)" for every transparent chat config item, because these config items are under a section named "transparent".


raiyni:
Maybe the descriptions should keep the transparent moniker? I do like the idea of trying to slim down config names though.

geheur:
Pushed a commit for that. Thank you vim macros :D qq

PR #16767, opened 2023-07-09
added option for including tax with GE price

Pheubel:
Includes an option for the highlight GE value to include the tax put on the item.

## Reasoning:
The purpose of this is to give better insight to a player from a selling perspective, such as on a slayer task. For example, the player receives a drop that goes for 60k on the GE and also alchs for 60k, normally the plugin would display a loss due to the cost of the nature rune. But by taking the GE tax into account the actual amount that the player will make when selling it to the GE will be `60000 - floor(60000 * 0.01) -> 60000 - 600 = 59400`. For a player that micromanages the liquidation of their drops, it might now be more interesting to high alch the item instead of selling it on the GE.

## Limitation:
the current way I implemented the GE tax is very simple and does not take into account the items exempt from GE tax. However, when I look at the list of [exempt items](https://oldschool.runescape.wiki/w/Category:Items_exempt_from_Grand_Exchange_tax), i think that the impact is minimal. It could be made more precise if a look up for this list of items existed so that applying the GE tax can be skipped for these items, but to my knowledge there isn't one.

resolves: https://github.com/runelite/runelite/discussions/15002

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemprices/ItemPricesConfig.java
@@ -96,4 +96,15 @@ public interface ItemPricesConfig extends Config
 	{
 		return true;
 	}
+
+	@ConfigItem(
+		keyName = "includeTax",
+		name = "Include GE tax",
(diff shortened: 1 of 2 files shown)


raiyni:
1. Tax is capped at 5m per item 
2. I'm not sure how useful this really is if you know it's always 1% and you have the ability it sell it for +- the avg GE price?

PR #16801, opened 2023-07-19
groundmarkers: Fix minimap stroke

CoreyForsyth:
The stroke is set in the renderPolygon method, along with drawing the polygon


raiyni:
What issue is this fixing?

Adam-:
It looks like this was double-drawing the polygon. Playing around with it locally I think it is better to just drop the `renderPolygon` call instead. `renderPolygon` additionally does an alpha fill, but the minimap is so small it isn't really noticeable so it would be best to not do it.

CoreyForsyth:
Updated to remove the `renderPolygon` call. Thank you

Adam- (changes requested):
you are missing the setColor() call

CoreyForsyth:
image
Before -> After at default zoom

Adam- (approved):

PR #16804, opened 2023-07-19
Add PMD Supression for Music Plugin Changes

emabrey:
Resolve the unhelpful PMD errors breaking the build despite being a minor issue. This patch simply adds a `SuppressWarning` annotation to the two problematic methods which were altered within the `net.runelite.client.plugins.music.MusicPlugin` class in commit 56055a07d48f1da72c2ad1c906cba594432577f8. The suppression approach should be temporary, so a TODO comment was added to make it easier to find the annotations for a more permanent fix in the future.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java
@@ -308,6 +308,7 @@ public class MusicPlugin extends Plugin
 	}
 
 	@Subscribe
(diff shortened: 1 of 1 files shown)


Nightfirecat:
We elected to revert the commit which caused the PMD errors in the first place, done here: c7956bfff3fcc470dba2276c36c5865f1992719f. We may yet remove the feature as today's update broke it in ways which are seemingly not so easily fixed, but if we do we'll do so in a way which doesn't break CI.

Issue #16806, opened 2023-07-19
Skill Calculator overwriting current experience with base level experience clicking out of level inputs


mattwright324:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

When trying to use the Skill Calculator the field for Current Experience will change from your actual current experience to the base experience for that level often wildly affecting action counts listed.

It seems to be triggered by the cursor first being in a level input then clicking to another input despite not changing the level input value.

### Screenshots or videos

https://github.com/runelite/runelite/assets/7717432/f4e06779-92cf-4c9d-abf0-2dc16f622de0

### RuneLite version

RuneLite version: 1.10.8.2
Oldschool revision: Rev 215
Launcher version: 2.1.3

PR #16812, opened 2023-07-20
#16806

mcaterino-sudo:
Fixes unintended behavior described in issue.

Added a check for the focus listener to not run if the current level does not change, so the current experience will not reset.
The action listener is unchanged so pressing 'enter' will still reset current experience to the base experience for that level.


Nightfirecat (changes requested):
Remove the random cache files added in this PR

PR #16827, opened 2023-07-22
Removed several smells across several files 

TristanJDuke:
This was school project work using Designite software to produce smell reports and and then squash them


jonathangarelick:
Please review the [Code Conventions](https://github.com/runelite/runelite/wiki/Code-Conventions). Brace placement is always on the next line.

PR #16842, opened 2023-07-27
grounditems: add broken arrow/glass to default hidden

jonathangarelick:
These two items are displayed by default (via "don't hide untradeables"). They are not quite useful for anything and clutter the screen, especially at Ferox Enclave.


jonathangarelick:
Actually, we might want to show broken glass, as it is a [quest item](https://oldschool.runescape.wiki/w/Broken_glass). Broken arrow has no use afaik.

PR #16858, opened 2023-07-30
DT2: add dt2 bosses to kc command

goldcodeo:
From my understanding the chatplugin calls runelite api to check database for player -> kc mapping, and just needs the boss names added. Let me know if I am missing anything, and I will gladly update.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -2465,6 +2465,23 @@ public class ChatCommandsPlugin extends Plugin
 			case "pm":
 				return "Phantom Muspah";
 
+			// Desert Treasure 2 bosses
+			case "the leviathan":
+			case "leviathan":
+				return "The Leviathan";
+
+			case "vardorvis":
+				return "Vardorvis";
+
+			case "duke sucellus":
+			case "duke":
+			case "sucellus":
+				return "Duke Sucellus";
+
+			case "the whisperer":
+			case "whisperer":
+				return "The Whisperer";
+
 			default:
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -2465,6 +2465,23 @@ private static String longBossName(String boss)
 			case "pm":
 				return "Phantom Muspah";
 
+			// Desert Treasure 2 bosses
+			case "the leviathan":
```
Nightfirecat, line 2469: This is covered by the default case; `WordUtils.capitalize()` would transform "the leviathan" to "The Leviathan" already. The same is true for the other entries. The case matching here is only for shortnames/nicknames, so "leviathan", "duke", "sucellus", and "whisperer" are all good additions.

geheur:
Did you check if this works in-game? For me, `!kc leviathan` already works. The killcount is stored under `leviathan` and not `the leviathan`. So it seems to me like those mappings should be reversed.
```
killcount.rsprofile.B4HbbEBQ.duke\ sucellus=1
killcount.rsprofile.B4HbbEBQ.leviathan=1
killcount.rsprofile.B4HbbEBQ.whisperer=1
killcount.rsprofile.B4HbbEBQ.vardorvis=1
```
I have only killed the quest version of the bosses though.

Adam-:
Indeed this looks backwards. I also have:
```
killcount.rsprofile.<profile>.duke\ sucellus=8
killcount.rsprofile.<profile>.leviathan=1
killcount.rsprofile.<profile>.whisperer=3
killcount.rsprofile.<profile>.vardorvis=10
```

Nightfirecat:
Added in 5a2c3c349621f3c7d5e1b26e6d49fabc1e426dbe

PR #16863, opened 2023-07-31
Add PluginMessage event

jocopa3:
A simple way for plugins to communicate over the `EventBus` while allowing listeners to know the source plugin and filter by `messageName`. While it's already possible to do this by e.g. posting/subscribing to String, HashMap, etc., providing a proper object for EventBus communication simplifies the problem and makes it possible to identify the event source.

This makes no attempt to verify objects in the data map are visible to other classloaders.

### Example Usage

**Event Posting:**
```java
Map<String, Object> data = new HashMap<>();
data.put("location", new WorldPoint(x, y, plane));
eventBus.post(new PluginMessage(this, "example:doThing", data));
```

**Subscriber:**
```java
@Subscribe
public void onPluginMessage(PluginMessage message)
{
    String messageName = message.getMessageName();
    if (!messageName.startsWith("example:"))
    {
        return;
    }

    String command = messageName.substring(messageName.indexOf(':') + 1);
    if (command.equals("doThing"))
    {
        WorldPoint location = (WorldPoint) message.getMessageData().get("location");
        doThing(location);
    }
}
```

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/events/PluginMessage.java
+++ b/runelite-client/src/main/java/net/runelite/client/events/PluginMessage.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2023, jocopa3
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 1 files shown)


Adam-:
added in 0b12333eafcb9ae1b394849fd65b0526a4e3a8f9

PR #16917, opened 2023-08-13
inventorytags: Add colors of equipped items to Inventory Tag submenu

rugg0064:
Before this, the sub-menu to add tags when you shift-click on an item only pulled up colors of tagged inventory items, and not colors of tagged equipped items. Now it will consider both inventory and equipped item tags.


runelite-client/src/main/java/net/runelite/client/plugins/inventorytags/InventoryTagsPlugin.java
```diff
@@ -183,7 +185,10 @@ public void onMenuOpened(final MenuOpened event)
 					.setTarget(entry.getTarget())
 					.setType(MenuAction.RUNELITE_SUBMENU);
 
-				for (Color color : invColors())
+				Set<Color> invEquipmentColors = new HashSet<>();
+				invEquipmentColors.addAll(invColors());
```
Adam-, line 189: It is okay to just `getTagsFromItemContainer(INVENTORY)` and `getTagsFromItemContainer(EQUIPMENT)` here

PR #16941, opened 2023-08-18
logback: configure programatically

CoreyForsyth:
Configures logback via code instead of parsing XML on startup.
This should save a modest amount of time on startup, ~50-100ms.

Simple test for a before/after:

```
public class Test
{
	public static void main(String[] args)
	{
		long s = System.currentTimeMillis();
		LoggerFactory.getLogger(Test.class);
		System.out.println(System.currentTimeMillis() - s);
	}
}
```


LlemonDuck:
The loss in ease-of-configuration here is not worth the marginal startup benefits.

PR #16962, opened 2023-08-25
Corner only tile indicators for tile indicators plugin.

geheur:
image

Adds a checkbox to the config for each of the 3 tiles that the plugin can indicate, which causes that tile to be highlighted with corners only instead of the entire outline. This reduces the visual clutter of the tile indicators while still clearly marking the tile.

based on #16857


geheur:
Is this something that could make it into core or would this be better off as a hub plugin?

geheur:
closed in favor of putting this on the plugin hub.

PR #16980, opened 2023-09-02
Update EntityHiderPlugin.java

harbingerofme:
Hide others without showing ignores in entity hider.
This is being abused in dmm as shown [here](https://youtube.com/clip/UgkxbNhVJqNgiL48TAxvLQkgHwEl9Q3sD5Fm?si=SBJTM7jDSl8zRfxI)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
@@ -210,12 +210,15 @@ public class EntityHiderPlugin extends Plugin
 			{
 				return !hideClanMembers;
 			}
+			if (drawingUI ? hideOthers2D : hideOthers)
+			{
+				return false;
+			}
 			if (client.getIgnoreContainer().findByName(player.getName()) != null)
 			{
-				return !hideIgnoredPlayers;
+				return !hideIgnoredPlayer;
 			}
-
-			return !(drawingUI ? hideOthers2D : hideOthers);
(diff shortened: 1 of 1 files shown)


runelite-client/src/test/java/net/runelite/client/plugins/entityhider/EntityHiderPluginTest.java
```diff
@@ -72,7 +72,6 @@ public void before()
 	{
 		Guice.createInjector(BoundFieldModule.of(this)).injectMembers(this);
 
-		when(client.getIgnoreContainer()).thenReturn(ignoreNameableContainer);
```
Nightfirecat, line 75: I'd advise changing this to a lenient stub so future tests can't run into a case where they'd need to stub this just to avoid NPEs.

Nightfirecat (approved):
Change generally LGTM, and I agree with the reasoning behind it

runelite-client/src/test/java/net/runelite/client/plugins/entityhider/EntityHiderPluginTest.java
harbingerofme: Resolved in 79579366117a37bd2c2245ed34ad81c84a3f3db3

fuzxi:
I strongly support this change. It makes no sense that Entity Hider can be used this way when using Player Indicators to highlight ignored players/offline friends is on the [list of rejected features](https://github.com/runelite/runelite/wiki/Rejected-or-Rolled-Back-Features) (bullet point 3).

PR #16985, opened 2023-09-03
Added tile border width option to the player indicators plugin

orange-puff:
This PR adds a config option to the Player Indicators plugin that will control the width of the tile if the user has selected `Draw tiles under players` rather than defaulting to `new BasicStroke(2)`

This is more inline with `Object Markers`, e.g. https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/objectindicators/ObjectIndicatorsConfig.java#L107-L117

**Old:**
old

**New:**
new

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/playerindicators/PlayerIndicatorsConfig.java
@@ -230,6 +230,17 @@ public interface PlayerIndicatorsConfig extends Config
 
 	@ConfigItem(
 		position = 11,
+		keyName = "tileBorderWidth",
+		name = "Tile border width",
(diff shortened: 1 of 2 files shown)


Nightfirecat:
This change in isolation isn't objectionable, however this would mean that player indicator tiles would not match the style of every other tile indicator used in the client. I'd be much more interested in making a global toggle for tile highlight width than restricting it to a single plugin like this.

orange-puff:
The `Tile Indicators` plugin actually also allows the user to set the border width https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java#L146-L156

I am not necessarily opposed to adding something global, but can you point me to an example of some global toggle that exists right now so I can use that as an example? Also, would I be expected to migrate any plugin that can create tiles as part of its functionality as part of this PR?

Issue #16819, opened 2023-07-21
Desert Treasure II - The Fallen Empire update checklist


Nightfirecat:
### Type

Other

### Operating System

None

### To do:
- [ ] **Boss Timers plugin:** Add respawn timers for the new bosses
- [ ] **Chat Commands plugin:** Add `!kc` shortnames for new bosses
- [ ] **Clues plugin:** Add the [Ring of shadows](https://oldschool.runescape.wiki/w/Ring_of_shadows) as an alternative to the Ring of visibility in any clues where it's needed
- [ ] **Discord plugin:** Add or update regions involved with the quest and bosses
- [ ] **Hiscores plugin:** Add icons for the new bosses
- [x] **Quests API file:** Add The Fallen Empire entry (cca72dec9f4a585b0f1cbd01f4e7b8ce3ff6a1ca)
- [ ] **Regeneration Meter:** Update to work with the unique "special attack" mechanic of the [Soulreaper axe](https://oldschool.runescape.wiki/w/Soulreaper_axe)
- [ ] **Skybox plugin:** Add colors for new regions as needed
- [ ] **Slayer plugin:** Add icons for boss tasks
- [ ] **World Map plugin:** Add quest start point

### To check and implement if applicable:
- [ ] **Chat Commands plugin:** Confirm that kill count and personal best times are tracked properly for the new bosses
- [ ] **ItemMapping:** Add blood torva -> normal torva mapping (if it passes the poll)
- [ ] **Loot Tracker plugin:** Support new boss loot tracking (if it doesn't already work)
- [ ] **NPCUtil:** Add isDying support for new bosses (if it doesn't already work)
- [ ] **World Map plugin:** Add tooltips to any other newly-added icons (dungeons, etc.)

### RuneLite version

.

goldcodeo:
Do we need to wait for the game update to start working on there, or are some of the possible, now ?

Nightfirecat:
There isn't really anything that's possible in advance. At minimum, most items require the associated NpcID/ItemID entries to be added to runelite-api, or knowing the IDs of regions ingame. Those won't be added until day of release.

Whiteaj36:
Can confirm that Sanguine Torva is needed to add to the mapping, or at least needs its own item entry into ItemID.java. If there's an easy way to get the item ID i can post a PR up. But not sure how to get that item ID.

geheur:
There are item ids in ItemID.java

Birger9:
[...]

PR #16992, opened 2023-09-04
Plugin/soulreaper axe in regenmeter plugin

OmbudRov:
Added Soulreaper Axe to the Regeneration Meter Plugin and added a VarpValue for Soul Stacks

Part of #16819


OmbudRov:
i forgot to mention this but i'm holding onto the axe if anyone wants it to test the code

Nightfirecat:
I can confirm the varplayer index and that it drains by 1 every 10 ticks. This code seems like it would not handle the case where a player reaches 5 soul stack and continues attacking--it would indicate that it is still draining. (I'm not sure there's a good solution for this case since no further var changes would be fired)

Issue #17004, opened 2023-09-09
The Miscellania favor reminder should update at midnight GMT like the rest of the dailies


Nightfirecat:
[Per Jagex's 2023-09-06 game update](https://secure.runescape.com/m=news/dmm-tweaks-upcoming-launcher-changes--more?oldschool=1):

> Those of you who are currently regents in the Kingdom of Miscellania will be glad to know that your subjects now attend to your various whims at exactly midnight each day, instead of whatever time you initially completed the quest.

This means we can more reliably update the active coffer funds and favor amount at midnight GMT instead of assuming it happens 1 day after it was last changed.

sufferfish:
I'd be willing to take a stab at this.

sufferfish:
> I'd be willing to take a stab at this.

I feel stupid. I'm unsure how to test my changes since I haven't completed the necessary quest...

acrinklaw:
Hey @sufferfish, how did you implement it? You could always mock things out to test out the logic, especially since it seems like it would need minimal changes. I can also checkout your branch and test it for you in game if you'd like, I just didn't see it in your fork

sufferfish:
Hey @acrinklaw I've pushed my changes to my fork. Feel free to check them out. Upon further review I don't think it's completed but I was hoping to check that the changes I currently made didn't bork the game. @Birger9 feel free to work on this though! I wouldn't want to keep it from being done in a timely manner.

Birger9:
@sufferfish I have completed the task and I have tested it. My code should work. I will open a PR for this issue. 
Sorry if you feel like have stolen this issue from you, I'm sorry if you feel like that. I have pressure from my University to finish a number of issues from an open source project within a certain time frame.

PR #17026, opened 2023-09-15
Update miscellenia favor GMT time

Birger9:
Fix for this issue: The Miscellania favor reminder should update at midnight GMT like the rest of the dailies #17004

Link: https://github.com/runelite/runelite/issues/17004

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
@@ -28,6 +28,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.inject.Provides;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
@@ -214,7 +215,7 @@ public class KingdomPlugin extends Plugin
 
 	private int estimateCoffer(Instant lastChanged, int lastCoffer)
 	{
-		int daysSince = (int) Duration.between(lastChanged, Instant.now()).toDays();
+		int daysSince = getNumbersOfDaysPassed(lastChanged);
 		int maxDailyWithdrawal = isRoyalTroubleCompleted() ? MAX_WITHDRAWAL_ROYAL_TROUBLE : MAX_WITHDRAWAL_BASE;
 		int maxDailyThreshold = maxDailyWithdrawal * 10;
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/kingdomofmiscellania/KingdomPlugin.java
```diff
@@ -227,13 +228,25 @@ private int estimateCoffer(Instant lastChanged, int lastCoffer)
...
+		int daysSince = getNumbersOfDaysPassed(lastChanged);
 		float dailyPercentage = isRoyalTroubleCompleted() ? APPROVAL_DECREMENT_ROYAL_TROUBLE : APPROVAL_DECREMENT_BASE;
 
 		lastApproval -= (int) (daysSince * dailyPercentage * MAX_APPROVAL);
 		return Math.max(lastApproval, 0);
 	}
 
+	private int getNumbersOfDaysPassed(Instant lastChanged)
+	{
+		Instant midnight = lastChanged.plus(1, ChronoUnit.DAYS).truncatedTo(ChronoUnit.DAYS);
```
Adam-, line 240: I think this reduces to:
```java
	private static int getNumbersOfDaysPassed(Instant lastChanged)
	{
		lastChanged = lastChanged.truncatedTo(ChronoUnit.DAYS);
		var now = Instant.now().truncatedTo(ChronoUnit.DAYS);
		return (int) Duration.between(lastChanged, now).toDays();
	}
```
?

Birger9: Hi! That is correct. I'm not used with pull requests. Do I add your changes in my fork and commit to the pull request? Or can you do something directly from Github to approve you change?

Adam-: You can update the PR. I think your code is adding an extra day (the `+ 1`), do know why that is? Is it required? My code is not.

Birger9: The extra one was required for my code but not yours, I double checked your code and changed the value of lastChanged and the code returned the right number of days that have passed,

I added the extra one because if last changed is 9/16/2023 5PM then midnight variable will be 9/17/2023 Midnight and Instant.Now() could be 9/17/2023 4PM. Then if I understand it correctly then (int) Duration.between(lastChanged, Instant.now()).toDays() would return 0 which is incorrect. As it have not been 24 hours but a day has still passed. I might be incorrect in that assumption but when I tested my code a bit it worked.

Birger9: One thing, should not var now = Instant.now().truncatedTo(ChronoUnit.DAYS) be Instant now instead? There is no need to have it a var?

PR #17028, opened 2023-09-15
Idle Notifier: Add ent chopping

### LINKED ISSUE #12898: Ent idle notifier
**Describe the bug**
The idle notifier doesn't notify when done chopping an ent trunk

**To Reproduce**
Kill an ent and chop it's trunk while the idle notifier plugin is on

**Expected behavior**
Notify when done chopping

 - OS: Ubuntu (pretty sure it happens on Windows as well)
 - RuneLite version: 1.6.33
 - Launcher version: 2.1.3

I also noticed that the loot tracker plugin doesn't work for ent trunks, is this intended?

BBeeaarr:
Add ent chopping to idle notification plugin
Add ent chopping IDs
Closes #12898


Adam-:
these anim ids are re-used with canoe making, so this would cause the notifier to trigger from that. Which it probably should not. Also this requires the 2h axes now. I am not aware of a good way to differentiate between canoes and ents.

Nightfirecat:
We discussed this further and decided it's probably fine if the idle notifier fires for a player going idle after starting to chop a canoe.

PR #17029, opened 2023-09-15
Bank Tags: Add option to tag all variations of items by default

jeremybahadirli:
Reference discussion: #17027

This PR adds a configuration option to the Bank Tags plugin to tag all variations of items by default. When enabled, dragging an item onto a Tag Tab tags all variations of the item, while holding Shift tags only the single variation. Therefore, enabling this option effectively inverts the default behavior.


Nightfirecat:
I don't think we want this behavior. Even if we decide we do want to make this possible, we would probably make it default to tagging variants without a config.

christianlegge:
@Nightfirecat I'm not sure you understand this change. The shift behavior already exists and it defaults to tagging only the specific item without variants. The change only adds a config option to reverse the behavior, which you said should be how it works anyway.

jeremybahadirli:
@Nightfirecat @christianlegge
Correct. The current behavior is that normally single variations are tagged, and holding Shift tags all variations. All this PR does is add a toggle to reverse the default behavior.

PR #17055, opened 2023-09-20
timers: add mark of darkness timer

### LINKED ISSUE #17051: No timer for Mark of Darkness
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

10

### Bug description

Mark of Darkness just changed into a self-buff, but is not tracked as one.

### Screenshots or videos

_No response_

### RuneLite version

image

### Logs

_No response_

Macweese:
Add timer for mark of darkness. I wasn't able find any useful vars for the spell, so this relies on chatmessage events instead.

Fixes #17051


runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -829,6 +833,11 @@ else if (message.endsWith(WARD_OF_ARCEUUS_MESSAGE))
 			{
 				createGameTimer(WARD_OF_ARCEUUS, Duration.of(magicLevel, RSTimeUnit.GAME_TICKS));
 			}
+			else if (message.endsWith(MARK_OF_DARKNESS_MESSAGE))
+			{
+				createGameTimer(MARK_OF_DARKNESS, Duration.of(magicLevel, RSTimeUnit.GAME_TICKS));
+				createGameTimer(MARK_OF_DARKNESS_COOLDOWN, Duration.of(magicLevel - 10, RSTimeUnit.GAME_TICKS));
```
Nightfirecat, line 839: It looks like you're creating the cooldown timer twice--once here and once in the cooldown block? I imagine this line should be removed.

runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java
```diff
@@ -846,6 +855,20 @@ else if (client.getVarbitValue(Varbits.COMBAT_ACHIEVEMENT_TIER_MASTER) == 2)
...
 		}
 
+		if (config.showArceuusCooldown())
+		{
+			final int magicLevel = client.getRealSkillLevel(Skill.MAGIC);
+			if (message.endsWith(MARK_OF_DARKNESS_MESSAGE))
+			{
+				createGameTimer(MARK_OF_DARKNESS_COOLDOWN, Duration.of(magicLevel - 10, RSTimeUnit.GAME_TICKS));
+			}
+			else if (message.endsWith(MARK_OF_DARKNESS_EXPIRED_MESSAGE))
```
Nightfirecat, line 865: Does the expired message block need to be gated behind a config check at all? I don't think so. (Also when, if ever, does it fire when the timers aren't naturally expiring on their own?)

PR #17063, opened 2023-09-22
Add Filter Violators config option to chat filter plugin

IanBurgan:
Related to this discussion where a similar feature is requested: https://github.com/runelite/runelite/discussions/17021

The feature adds a configuration option that when enabled would filter out all messages from a user after one of their messages is found to match an existing filter.

### Implementation notes:
So this PR deviates slightly from the suggested feature and doesn't use a count of violations. There doesn't seem to be an easy way to accurately count how many times a user sent a message that got filtered without accidentally double counting messages. It's possibly noteworthy that the ignored users only exist in memory and won't persist. Modifying the config with the `configManager` is awkward because there is already a list of patterns used to ignore users that would ideally be used but they're patterns not usernames and the UI wouldn't immediately show config changes. If using a separate list of users in the config, storing the usernames would possibly really pollute the config so I think it's best avoided.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterConfig.java
@@ -175,4 +175,15 @@ public interface ChatFilterConfig extends Config
 	{
 		return false;
 	}
+
+	@ConfigItem(
+			keyName = "filterUsersMessagesAfterViolation",
+			name = "Filter Violators",
+			description = "Filter a player's future messages if they violate a configured filter",
+			position = 14
+	)
+	default boolean filterUsersMessagesAfterViolation()
+	{
+		return true;
+	}
 }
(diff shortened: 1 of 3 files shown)


raiyni:
>  There doesn't seem to be an easy way to accurately count how many times a user sent a message that got filtered without accidentally double counting messages

You would use the `onChatMessage` event to store a cache count much like duplicate messages

IanBurgan:
> > There doesn't seem to be an easy way to accurately count how many times a user sent a message that got filtered without accidentally double counting messages
> 
> You would use the `onChatMessage` event to store a cache count much like duplicate messages

I might be misunderstanding something about how `onChatMessage` works, but it appears this would still lead to issues. The duplicate message caches invalidates the messages here when the game state changes: https://github.com/runelite/runelite/blob/48be242fad0c58dcaf57a4f87c7cc5e404efee40/runelite-client/src/main/java/net/runelite/client/plugins/chatfilter/ChatFilterPlugin.java#L149-L159

I think this would cause trouble for accurately counting messages. Maybe the inaccuracy is fine if a "count" is preferred? Or maybe I'm not understanding when messages come through

PR #17067, opened 2023-09-24
clues: add ring of shadows to cryptic clue text

YvesW:
Since the ring of shadows does the same thing post-quest, I've added it to the cryptic clue text.


Nightfirecat:
If we're updating these clues, it might also be worth updating the emote clue(s) which need a Ring of visibility to mention the Ring of shadows also.

YvesW:
> If we're updating these clues, it might also be worth updating the emote clue(s) which need a Ring of visibility to mention the Ring of shadows also.

Added!

PR #17084, opened 2023-10-04
Make config section toggle more resistant to mouse movement

MosheBenZacharia:
Previously the config panel section toggles would only expand/collapse if you clicked down and up without moving your mouse a single pixel. This makes it behave closer to a normal button.


MosheBenZacharia:
> Nightfirecat — Yesterday at 11:26 AM
> iirc we discussed it in here briefly after it was opened, and thought it would be better if it was a button or some other element rather than a clickable label

closing since according to this seems like they don't want this to be fixed until this is changed to be a jbutton

PR #17096, opened 2023-10-09
Add ability to bind backspace in chat

vikke1234:
In order to be able to have better ergonomics, add ability to bind the backspace.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsConfig.java
@@ -210,4 +210,14 @@ public interface ChatCommandsConfig extends Config
 	{
 		return Keybind.NOT_SET;
 	}
+	@ConfigItem(
+			position = 22,
+			keyName = "clearCharacter",
+			name = "Backspace",
+			description = "Enable hotkey to press backspace"
+	)
+	default Keybind clearCharacter()
+	{
+		return Keybind.NOT_SET;
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatKeyboardListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatKeyboardListener.java
@@ -58,8 +58,8 @@ public class ChatKeyboardListener implements KeyListener
 	{
 		if (chatCommandsConfig.clearSingleWord().matches(e))
 		{
-			int inputTye = client.getVarcIntValue(VarClientInt.INPUT_TYPE);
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsConfig.java
```diff
@@ -210,4 +210,14 @@ default Keybind clearChatBox()
 	{
 		return Keybind.NOT_SET;
 	}
```
Birger9, line 212: Add space between line 212 and 213

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatKeyboardListener.java
```diff
@@ -58,8 +58,8 @@ public void keyPressed(KeyEvent e)
 	{
 		if (chatCommandsConfig.clearSingleWord().matches(e))
```
Birger9, line 59: @Override
	public void keyPressed(KeyEvent e) {
		int inputType = client.getVarcIntValue(VarClientInt.INPUT_TYPE);
		String input = inputType == InputType.NONE.getType()
			? client.getVarcStrValue(VarClientStr.CHATBOX_TYPED_TEXT)
			: client.getVarcStrValue(VarClientStr.INPUT_TEXT);

		if (chatCommandsConfig.clearSingleWord().matches(e)) {
			handleClearSingleWord(e, inputType, input);
		} else if (chatCommandsConfig.clearChatBox().matches(e)) {
			handleClearChatBox(e, inputType);
		} else if (chatCommandsConfig.clearCharacter().matches(e)) {
			handleClearCharacter(e, inputType, input);
		}
	}

	private void handleClearSingleWord(KeyEvent e, int inputType, String input) {
		if (input != null) {
			e.consume();

			// Remove trailing spaces
			while (input.endsWith(" ")) {
				input = input.substring(0, input.length() - 1);
			}

			int idx = input.lastIndexOf(' ') + 1;
			final String replacement = input.substring(0, idx);

			clientThread.invoke(() -> applyText(inputType, replacement));
		}
	}

	private void handleClearChatBox(KeyEvent e, int inputType) {
		e.consume();
		clientThread.invoke(() -> applyText(inputType, ""));
	}

	private void handleClearCharacter(KeyEvent e, int inputType, String input) {
		e.consume();
		if (input != null && input.length() > 0) {
			final String replacement = input.substring(0, input.length() - 1);
			clientThread.invoke(() -> applyText(inputType, replacement));
		}
	}
	
	I think you should refactor your code to this for the following reasons:
	
	1. Code reusability as I have created separate methods. The code is easier to understand now, I think atleast.
	2.  We reuse inputType and input to reduce the fetch to once instead of multiple times.

PR #17109, opened 2023-10-12
Add option to disable npc aggression message regarding calibration

grcatterall:


Nightfirecat:
Duplicate of https://github.com/runelite/runelite/pull/14860, we're not interested in adding such an option.

SuperNerdEric:
Sorry to bump this again, but I also find this a bit obnoxious with how much real estate it takes up. If you disable it because you aren't actively using it you are **guaranteed** that you will have to calibrate again when you **do** want it. But if you've been playing for a while it's probably going to end up getting calibrated for when you need it.

It would be nice to have an option to make the message much more compact with abbreviated instructions. You could even have a little warning icon like this with a tooltip that gives you the full instructions for calibration.
image

Alternatively you could make it always be calibrating even when the plugin is disabled, just not show the lines or overlay. That way when you turn it on it would mostly likely just work if you have been playing for a while.

Adam-:
Previously when we had the option to remove it, it caused an endless amount of support requests from people who enabled the option and then didn't know why the plugin wasn't working. It also can't calibrate while the plugin is disabled because the plugin code isn't running to calibrate. I wouldn't be opposed to something like make the unaggro infobox show a ? with the information which is currently in the overlay instead in a tooltip.

PR #17112, opened 2023-10-14
party: tie ping duration to system time

PortAGuy:
Currently, the ping duration is tied to the current fps. This commit solves this by instead decreasing the ping alpha by checking the elapsed time of when the ping was first created, and the current time.

(Clips are taken at 165fps)
Before:

https://github.com/runelite/runelite/assets/83369117/fe64a31b-5516-4c4f-8b25-3a8f9ffff7fa

After:

https://github.com/runelite/runelite/assets/83369117/67aa6aa3-d64a-408b-9a02-cf2856b41f8f

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/party/data/PartyTilePingData.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/party/data/PartyTilePingData.java
@@ -38,4 +38,5 @@ public class PartyTilePingData
 	private final WorldPoint point;
 	private final Color color;
 	private int alpha = 255;
+	private final long creationTime = System.currentTimeMillis();
 }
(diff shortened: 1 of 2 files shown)


PortAGuy:
Forgot to include a clip of the current behaviour at 50fps. Here is that clip:

https://github.com/runelite/runelite/assets/83369117/c7a6d4e8-7aef-4450-8465-fca0782ada23

runelite-client/src/main/java/net/runelite/client/plugins/party/data/PartyTilePingData.java
```diff
@@ -38,4 +38,5 @@ public class PartyTilePingData
 	private final WorldPoint point;
 	private final Color color;
 	private int alpha = 255;
+	private final long creationTime = System.currentTimeMillis();
```
Nightfirecat, line 41: Let's use `System.nanoTime()` instead of `System.currentTimeMillis()` to avoid the potential footgun of a user's system clock adjusting mid-ping, which could leave a ping visible for much longer than intended.

PR #17117, opened 2023-10-17
Opponent Info plugin: add toggle to only show boss health overlay

bradysauter:
Allows the Opponent Info plugin's custom health overlay to be toggled off, so that only the in-game setting 'boss health overlay' is affected.

Use case is I want to have both exact health amounts and percentage displayed on boss health bars (below), but I do not want the custom overlay displayed for non-boss monsters or players.
 image

This is the overlay that is being made toggleable (example of an NPC and a player): 
image
image

This overlay is unaffected, can still be toggled on/off independent of the new config:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
@@ -65,13 +65,13 @@ public interface OpponentInfoConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "onlyShowBossHealthOverlay",
-		name = "Only show boss health overlay",
-		description = "Hides this plugin's additional health overlay; only modifies the in-game setting 'boss health overlay'",
+		keyName = "showOpponentHealthOverlay",
+		name = "Show opponent health overlay",
+		description = "Shows a health bar overlay when a boss health overlay is not present",
 		position = 4
 	)
-	default boolean onlyShowBossHealthOverlay()
+	default boolean showOpponentHealthOverlay()
 	{
-		return false;
+		return true;
 	}
 }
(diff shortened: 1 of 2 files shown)


Adam-:
The comment you are editing says it already does the behavior you are adding?

```java
// The in-game hp hud is more accurate than our overlay and duplicates all of the information on it,
// so hide ours if it is visible.
```

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
```diff
@@ -63,4 +63,15 @@ default boolean showOpponentsInMenu()
...
 		return false;
 	}
+
+	@ConfigItem(
+		keyName = "onlyShowBossHealthOverlay",
+		name = "Only show boss health overlay",
+		description = "Hides this plugin's additional health overlay; only modifies the in-game setting 'boss health overlay'",
+		position = 4
+	)
+	default boolean onlyShowBossHealthOverlay()
```
Adam-, line 73: This should be inverted to "Show opponent info overlay" and default to true.

bradysauter:
> The comment you are editing says it already does the behavior you are adding?
> 
> ```java
> // The in-game hp hud is more accurate than our overlay and duplicates all of the information on it,
> // so hide ours if it is visible.
> ```

The small health overlay is already hidden when there is a boss health bar visible, yes. However, there is no way to toggle off the small health overlay when interacting with players or non-boss NPCs.

I am adding the ability to always hide the small health overlay, while retaining the ability to customize the boss health overlay.

runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoConfig.java
bradysauter: Yes I think that is clearer, changed.

Issue #17111, opened 2023-10-13
Dragon pickaxe special doesn't get filtered in "other players area sounds" in music plugin


Stickaplex:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

If you go to your in-game sound settings and mute area sounds you will no longer hear dragon pickaxe special attack. If you unmute the in-game but mute other players area sounds in the music plugin, you will hear everybody's special attack.

So either for some reason all special attack sounds are classed as your area sounds or it is being ignored in someway. Either way it feels like it should be muted through this option.

Tbh not entirely sure if this is a bug, but it feels like it is.

### RuneLite version

RL version: 1.10.14
Launcher version: 2.6.7

bchesnut19:
Looks like Dragon pickaxe special uses found_gem with sound ID 2655, appears it's a sourceless sound

Adding sound ID to [SOURCELESS_PLAYER_SOUNDS ](https://github.com/runelite/runelite/blob/e9edf591de0a4ed169aff6b461318857284e6373/runelite-client/src/main/java/net/runelite/client/plugins/music/MusicPlugin.java#L102) corrects the issue, but also filters sound effect from user + any gem found

PR #17127, opened 2023-10-21
Fix for filtering other players Dragon Pickaxe spec sounds

bchesnut19:
#17111


raiyni:
This ends up muting gem rock sounds.

bchesnut19:
It would just mute them when "Mute other peoples area sounds" toggled on, correct? Feel like that's worth the tradeoff, the current bug very noticeable when using it at shooting stars

PR #17130, opened 2023-10-22
Draft: Show account session name on logout tooltip

Zarthus:
## What does this do and why?

If you have multiple instances of runelite running it's not quite clear what you're signing out of. This adds the username of the session to the signout in the top right.

| Before | After |
|--------|--------|
| rl1 | |
 

The reason it's still a Draft is because this was a very quick POC and I did not validate if it works.  I still need to test it but haven't got time to setup the full project and run it locally yet.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/account/AccountPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/account/AccountPlugin.java
@@ -83,10 +83,14 @@ public class AccountPlugin extends Plugin
 			.onClick(this::loginClick)
 			.build();
 
+		final String signOutSuffix = sessionManager.getAccountSession() != null
+				? " (" + sessionManager.getAccountSession().getUsername() + ")"
+				: "";
+
 		logoutButton = NavigationButton.builder()
 			.tab(false)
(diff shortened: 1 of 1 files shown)


geheur:
You're not supposed to use multiple runelite accounts, instead you are supposed to use multiple profiles.

Zarthus:
@geheur I suspected as much (which was half the reason for opening the draft). Do you think it makes sense to list the currently used profile name in the top bar instead?

geheur:
Personally, no. I don't see how that would be useful.

Zarthus:
Let's close this then. I was having trouble properly navigating multiple accounts and getting them all setup correctly, but that's probably self induced.

geheur:
https://github.com/runelite/runelite/wiki/General-Features#profiles might be helpful if you haven't seen it before.

Zarthus:
@geheur - is there a way to pass `--profile` from the Jagex launcher? I'm struggling because jagex launcher does not remember which profile belongs to which account.

geheur:
That's not possible at the moment although it is planned. Also the discord is a better place for discussion like this.

PR #17131, opened 2023-10-22
Add 'Escape Crystal' varbits

Xylot:
Add the varbits for the 'Escape Crystal':

Inactivity teleport/time
ring of life teleport


Nightfirecat:
We do not add varbit definitions for data which is not used by the core client, as it's subject to bit rot. You can call `client.getVarbitValue()` from a hub plugin without the need for a defined field here.

Issue #17152, opened 2023-10-30
No Screenshots for achieving lvl 99 with level-up dialogs disabled


Nightfirecat:
### Discussed in https://github.com/runelite/runelite/discussions/17151

<div type='discussions-op-text'>

<sup>Originally posted by **iLight-24** October 30, 2023</sup>
Hi,
I achieved 2 99's last week but when I tried to see the screenshots in my folder it was missing.
All other lvls screenshot was there and the 99 missing.
</div>

skillingdev:
I could take this if no one else is working on it.

Adam-:
fixed in #17156

PR #17156, opened 2023-10-31
Fix screenshots for level 99 messages with level-up dialogs disabled

skillingdev:
#17152

When level-up dialogs are disabled, we check chat messages to capture screenshots for level ups. However, the chat message displayed for level 99 level ups is slightly different than non-99s, meaning we weren't taking screenshots for 99 level ups.

This PR adds a pattern to capture the 99 message, and checks for it after checking for the non-99 pattern.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -104,8 +104,7 @@ public class ScreenshotPlugin extends Plugin
 	private static final int CORRUPTED_GAUNTLET_REGION = 7768;
 	private static final Pattern NUMBER_PATTERN = Pattern.compile("([0-9]+)");
 	private static final Pattern LEVEL_UP_PATTERN = Pattern.compile(".*Your ([a-zA-Z]+) (?:level is|are)? now (\\d+)\\.");
-	private static final Pattern LEVEL_UP_MESSAGE_PATTERN = Pattern.compile("Congratulations, you've just advanced your ([a-zA-Z]+) level. You are now level (\\d+)\\.");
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Can you add a test case for this behavior?

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -105,6 +105,7 @@ public class ScreenshotPlugin extends Plugin
 	private static final Pattern NUMBER_PATTERN = Pattern.compile("([0-9]+)");
 	private static final Pattern LEVEL_UP_PATTERN = Pattern.compile(".*Your ([a-zA-Z]+) (?:level is|are)? now (\\d+)\\.");
 	private static final Pattern LEVEL_UP_MESSAGE_PATTERN = Pattern.compile("Congratulations, you've just advanced your ([a-zA-Z]+) level. You are now level (\\d+)\\.");
+	private static final Pattern LEVEL_UP_MESSAGE_PATTERN_99 = Pattern.compile("Congratulations, you've reached the highest possible ([a-zA-Z]+) level of 99.");
```
Nightfirecat, line 108: You could instead update the original level up regex so that we don't need to do a separate regex match check.

```
Pattern.compile("Congratulations, you've (?:just advanced your ([a-zA-Z]+) level\\. You are now level (\\d+)|reached the highest possible ([a-zA-Z]+) level of (99))\\.");
```

skillingdev:
I'll add a test & use a single regex pattern shortly.

skillingdev:
The latest commit unifies the message patterns for non-99 and 99 level-ups. It does make the capture groups a bit more involved. I can revert or use explicit if statements if people prefer that over the ternary operator.

Adam-:
You can name the capture groups and then reference them by name, to make this less convoluted.

PR #17157, opened 2023-10-31
Fishing Plugin: Ensure `setLastFishCaught` is called when catching Karambwanjis

### LINKED ISSUE #17087: Karambwanji Fishing Issue
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

When fishing karambwanji with a full inventory of shrimp the plugin stops displaying the fishing.
I'd imagine this is because your inventory doesn't really change while you fish them?

### Screenshots or videos

image
This is an image of the panel when you do not have a full inventory yet.

### RuneLite version

runelite: 1.10.14
launcher: 2.6.1
osrs: Rev 217

### Logs

_No response_

skillingdev:
Fixes #17087 (see issue for my reproduction + screenshot)

The message displayed while catching Karambwanjis is slightly different from the other fishing related messages, meaning that it was possible to get into a state where the fishing overlay wouldn't appear even though you were actively fishing for Karambwanjis. This was happening because `lastFishCaught` was never set (and the overlay requires `lastFishCaught` to be set in order to appear).

This PR fixes that issue. I can now approach a Karambwanji spot with a full inventory, start fishing, and see the fishing overlay appear as soon as the first XP drop appears.

~Notably, the "Caught fish" message isn't quite right - it is only taking into account actions taken, but one action can result in a variable amount of karambwanjis being caught at once (depending on fishing level). Fixing that is a larger change so I've left it out for now, but I'm happy to give it a shot if people would like it done.~ I've received confirmation that we are intentionally tracking actions [here](https://github.com/runelite/runelite/issues/17087#issuecomment-1787494910).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
@@ -85,7 +85,7 @@ public class FishingPlugin extends Plugin
 	private static final int TRAWLER_SHIP_REGION_NORMAL = 7499;
 	private static final int TRAWLER_SHIP_REGION_SINKING = 8011;
 	private static final int TRAWLER_TIME_LIMIT_IN_SECONDS = 314;
-	private static final Pattern CAUGHT_KARAMBWANJI_MESSAGE = Pattern.compile("You catch (\\d+) Karambwanji.");
+	private static final Pattern CAUGHT_FISH_MESSAGE = Pattern.compile("You catch (?:a|some|\\d+ Karambwanji\\.)");
 
 	private Instant trawlerStartTime;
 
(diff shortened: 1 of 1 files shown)


raiyni:
Can you provide the message?

skillingdev:
Sure - it's included in my screenshot here: https://github.com/runelite/runelite/issues/17087#issuecomment-1787494910

image

"You catch 15 Karambwanji." (although that'd be different if my fishing level was different :D )

MosheBenZacharia:
(You catch .*) should work for all three cases, that is what I use in my plugin.

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
```diff
@@ -83,6 +85,7 @@ public class FishingPlugin extends Plugin
 	private static final int TRAWLER_SHIP_REGION_NORMAL = 7499;
 	private static final int TRAWLER_SHIP_REGION_SINKING = 8011;
 	private static final int TRAWLER_TIME_LIMIT_IN_SECONDS = 314;
+	private static final Pattern CAUGHT_KARAMBWANJI_MESSAGE = Pattern.compile("You catch (\\d+) Karambwanji.");
```
Nightfirecat, line 88: Let's adapt this to match all standard catch messages.
```suggestion
	private static final Pattern CAUGHT_FISH_MESSAGE = Pattern.compile("You catch (?:a|some|\\d+ Karambwanji\\.)");
```

This change lets us simplify the condition where it's used to:

```java
if (message.equals("Your cormorant returns with its catch.") || m.find())
```

Adam-:
added in 064c907fa7a070df98775721506a1059bac9da5a. I prefer to add tests for the chat message parsing code, and also avoid regex matching if possible due to the overhead.

PR #17169, opened 2023-11-06
Ground items - fix bug where you cannot remove items via the overlay [-] or [+] button if it has a quantity.

geheur:
e.g. if you have `coins<10`.


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/ItemThreshold.java
```diff
@@ -94,4 +94,18 @@ boolean quantityHolds(int itemCount)
...
+	public static boolean matchesEntry(String entry, String itemName)
+	{
+		int index = entry.indexOf('<');
+		if (index == -1)
+		{
+			index = entry.indexOf('>');
+		}
+		if (index != -1)
+		{
+			entry = entry.substring(0, index);
```
Birger9, line 107: Hi! Is there ever a chance that the variable entry ever contains trailing whitespaces? If so, one should add .trim() to row 107 as otherwise equalsIgnoreCase without trimming will lead to an incorrect result

raiyni: No https://github.com/runelite/runelite/blob/f8420ab53559574662546ca8595831f2bb2d8786/runelite-client/src/main/java/net/runelite/client/util/Text.java#L60

Adam-:
This behavior I think was intentional, since you cannot add quantity values with the +/- buttons, it doesn't allow you to then remove them. This is consistent with the wildcard matches too which do not allow +/- to work.

PR #17182, opened 2023-11-09
Add: Plugin Profile Account Linking

DoinkOink:
Adds the ability to link Runescape Usernames to specific plugin profiles. When an account is logged in and a profile is linked to that RSN it will automatically load that profile.


raiyni:
this was added in https://github.com/runelite/runelite/commit/3f4bb46931e54db2d281221663ed84fce4311dbe ?

PR #17194, opened 2023-11-15
First working version of gender neutral replacement plugin

ace-english:
Created plugin. Offers replacements for gendered terms in NPC dialogue


Felanbird:
I believe plugins are basically never added to core anymore, so this should be added as a [hub plugin](https://github.com/runelite/plugin-hub/#readme)

ace-english:
> I believe plugins are basically never added to core anymore, so this should be added as a [hub plugin](https://github.com/runelite/plugin-hub/#readme)

thanks, I appreciate the pointer

PR #17200, opened 2023-11-17
Fix rune pouch overlay for 4th/5th runes in leagues

testing-ongithub:
For the leagues rune pouch extension, the 4th and 5th runes are set in the 5th and 6th rune pouch varbits, respectively.

This pull request updates the `renderGrid` method to allow skipping empty runes, using similar logic to what's in the `renderList` method.

Before:
 pouch-before
pouch-before2

After:
pouch-after
pouch-after2


Adam-:
I think @abextm has a similar change somewhere for this that he hasn't pushed yet

abextm:
fixed in ba8bb5d5c3ea3219f4bd478e923613ecf0b5b80b because I already tested that, though its basically identical

Issue #17049, opened 2023-09-19
[Loot Tracker] Other Actions in Grubby Chest area adds to Grubby Chest Count


krickit:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Steps to reproduce:

1. "Open Grubby Chest" while holding a Grubby Key or Right click option "Check Grubby Chest"
2. Attempt various actions (eat food, fletch, certain teleports, etc.)

Unexpected effect: The Loot Tracker counter for Grubby Chest increases without opening the chest

Tested:

- Triggering: House tab, Standard Spellbook teleports, Ectophial, Fletching Willow logs, Bury Bones, Eat Mackerel
- Not triggering: Khardest Memoirs teleport out, Xeric's Talisman Teleport out, decanting potions.

Not tested: Not already having a Grubby Chest loot tracked. 

Additional Notes:
Changed to safe mode and tested again, behaviour still occurs

### Screenshots or videos

![image](https://github.com/runelite/runelite/assets/40974473/b4f89c51-9429-4955-af1d-eb1c67383575)

### RuneLite version

1.10.12.2
Rev 216
2.6.3

### Logs

[client.log](https://github.com/runelite/runelite/files/12657894/client.log)
[launcher.log](https://github.com/runelite/runelite/files/12657900/launcher.log)

mizzu:
I couldn't replicate it with "Open" chest, but it was definitely an issue with the "Check" option. I created a pull request with some more info in #17209.

PR #17209, opened 2023-11-19
#17049 check chest look tracker bug

mizzu:
When checking a chest, onChatMessage in LootTrackerPlugin.java would get called and create the inventory snapshot to get ready for the loot change because the "Open" and "Check" options would use the same starting text (e.g. "You have opened the Grubby Chest"). Any inventory changes after this would cause the LootTracker to increment whatever chest was checked.

I added a check when the "Check" option menu is clicked on to grab the name of the chest that was checked, and compare that to the next chest reward type chat message to make sure it was a chest that was clicked. Tested with the Grubby Chest and Larran's Big Chest.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -958,7 +958,8 @@ public class LootTrackerPlugin extends Plugin
 
 			final int regionID = client.getLocalPlayer().getWorldLocation().getRegionID();
 
-			if(!lastChestChecked.isEmpty() && CHEST_EVENT_TYPES.get(regionID).equals(lastChestChecked)) {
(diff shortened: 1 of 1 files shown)


Adam-:
This type of issue is sort of all over the plugin, and is just inherit with how it works. Eg shade chests, which do check for Open, can break just by not having a key to open the chest when you click it:

```java
		else if (isObjectOp(event.getMenuAction()) && event.getMenuOption().equals("Open") && SHADE_CHEST_OBJECTS.containsKey(event.getId()))
		{
			onInvChange(collectInvAndGroundItems(LootRecordType.EVENT, SHADE_CHEST_OBJECTS.get(event.getId())));
		}
```

The ground snapshot code, which is for tracking Whisperer loot, has a "timeout" concept which I think might be better here. While it doesn't solve the underlying problem of triggering the loot tracker and making an inv change at once, which is unsolvable (and easily done a bunch of different ways, such as with eg nests), it would avoid this issue of accidentally having an inv change queued when actually there are no imminent inv changes.

PR #17219, opened 2023-11-23
woodcutting: add group bonus

YvesW:
Adds a woodcutting group bonus that uses the varbit that is now transmitted.

Example:
https://streamable.com/ufv4e3

Works properly in the WC guild:
https://github.com/runelite/runelite/assets/7929021/efb8dbbb-f1b5-4bc6-8fef-e93d1c82608b

Works properly when chopping farming trees:
https://github.com/runelite/runelite/assets/7929021/8c4e7bd5-8f7e-4e17-a4f6-6688ab6fb34c

Thanks to Rugg for posting the text of the enhanced client in ``#runelite``, which allowed me to easily search for the right CS2 script.

Things to take into consideration, purely preference/opinions of the maintainers:
- Alternatively, e.g. itemManager.getImage(ItemID.RUNE_AXE) could be used as icon instead of the WC icon.
- One could opt to not add the infobox if the ``groupBonus`` is 0.
- The color could be changed from white to green when the boost is >0


YvesW:
Switched to an approach that also uses ``Varplayer.BUFF_BAR_WC_GROUP_BONUS``. All bugs should be stamped out now. Videos are updated.

raiyni:
I think if this hides the 0 value it will be good. Doesn't make sense to show the infobox if you aren't getting the bonus I think.

YvesW:
The infobox should now not be added when the bonus is 0, and the infobox will be removed/re-added when everyone else stops/starts cutting the tree you're cutting, respectively.
Tests: https://youtu.be/YMMcCW6sX08

PR #17222, opened 2023-11-24
Monkey woodcutting fixes

### LINKED ISSUE #547: The woodcutting overlay thinks I'm not woodcutting when I'm a monkey
I'm not sure if there's actually a way to solve this, since the overlay is using the player's animation to figure out if there's any woodcutting going on, but it doesn't look like there's any animation for woodcutting while you're a monkey.
2018-02-08-000109_555x507_scrot

YvesW:
Fixes monkey woodcutting:
- idle notifier (closes #547)
- woodcutting overlay when chopping trees as a monkey

The following videos were recorded before I split the varp changes into a separate commit, so the group bonus is not included in this PR anymore.
Before:
https://streamable.com/eo2rjo

After:
https://streamable.com/to6dh2

A separate commit is used for the VarPlayer changes so I could also use that commit in https://github.com/runelite/runelite/pull/17219

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
+++ b/runelite-api/src/main/java/net/runelite/api/VarPlayer.java
@@ -240,4 +240,12 @@ public final class VarPlayer
 	 * Assigned slayer task location. The mapping of value to name can be found in {@link EnumID#SLAYER_TASK_LOCATION}
 	 */
 	public static final int SLAYER_TASK_LOCATION = 2096;
+
+	/**
+	 * Determines whether the woodcutting group bonus should be displayed on the buff bar or not.
+	 * 96 = displayed (including the woodcutting guild).
+	 * 0 = not displayed (after login until cutting a tree except normal trees or trees grown through farming).
+	 * -1 = not displayed (including normal trees or trees grown through farming).
+	 */
+	public static final int BUFF_BAR_WC_GROUP_BONUS = 4007;
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -52,6 +52,7 @@ import net.runelite.api.events.GraphicChanged;
 import net.runelite.api.events.HitsplatApplied;
 import net.runelite.api.events.InteractingChanged;
 import net.runelite.api.events.NpcChanged;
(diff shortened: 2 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
```diff
@@ -103,6 +104,7 @@ public class IdleNotifierPlugin extends Plugin
 	private Instant sixHourWarningTime;
 	private boolean ready;
 	private boolean lastInteractWasCombat;
+	private static final int buffBarNotDisplayed = -1;
```
Adam-, line 107: By convention, static final fields are always UPEPRCASE. Or just inline the -1 since it does not really matter.

runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingOverlay.java
```diff
@@ -40,6 +41,7 @@
 class WoodcuttingOverlay extends OverlayPanel
 {
 	private static final String WOODCUTTING_RESET = "Reset";
+	private static final int buffBarDisplayed = 96;
```
Adam-, line 44: see above comment

Adam- (changes requested):

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
```diff
@@ -240,4 +240,11 @@ public final class VarPlayer
 	 * Assigned slayer task location. The mapping of value to name can be found in {@link EnumID#SLAYER_TASK_LOCATION}
 	 */
 	public static final int SLAYER_TASK_LOCATION = 2096;
+
+	/**
+	 * Determines whether the woodcutting group bonus should be displayed on the buff bar or not.
+	 * 96 = displayed (including the woodcutting guild).
+	 * -1 = not displayed (including normal trees or trees grown through farming).
```
Nightfirecat, line 247: varp 4007 is also 0 on login, and until you begin woodcutting

Nightfirecat (commented):
LGTM besides the minor javadoc note

runelite-api/src/main/java/net/runelite/api/VarPlayer.java
YvesW: Thank you very much for the review! I've amended the javadoc.

PR #17223, opened 2023-11-24
ui: Fix frame out of bounds on OS X

Klievan:
Fixes: https://github.com/runelite/runelite/issues/12711

When swapping resolutions or when running an external monitor on MacOS, the client can come into a state where it becomes out of bounds and is unable to be moved unless adapting to a higher resolution or by turning on safe-mode to reset the bounds of the window. I am unaware if this issue also exists on Windows.

To reproduce:

1. Set window resolution to 3840 x 2160
2. Resize the window so that it covers a major part of your screen
3. Close RuneLite
4. Set the display resolution to 1920 x 1080
5. Open RuneLite and observe that the client has become unusable. Plugins panel is inacessible, topbar is not shown so can not be resized. Gameclient only shows partially.

The client seems to detect if a window is out of bounds, but only moves it's position. From my testing, frame. getOwner() returns null in this case, so this doesn't solve much. The fix does the following

1. After moving the frame, check if it is still out of bounds
2. If it is still out of bounds, this means that the window is bigger than the display. Find an appropriate display to place the client on by checking for the first display that overlaps with the Runelite frame.
3. If an appropriate display is found, grab it's dimensions. If it is running MacOS, account for the top-bar to prevent the window from sliding under.
4. Apply the new bounds to the frame.

This fix has been tested in the following situations:
1. Moving from a high resolution to a lower resolution on the same screen.
2. Moving from a high resolution external display to a lower resolution display.


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -566,6 +566,28 @@ public void show()
...
+						if (gc == null)
+						{
+
+							gc = findDisplayFromIntersect(frame.getBounds());
+							if (gc != null)
+							{
+								Rectangle adjustedBounds = gc.getBounds();
+
+								// Adjust for macOS top bar
+								if (OSType.getOSType() == OSType.MacOS)
```
Adam-, line 581: I am not really a fan of hard coding the menu bar offsets, have you tried using the screen insets instead? Such as:

```java
var insets = Toolkit.getDefaultToolkit().getScreenInsets(gc);
bounds.y += insets.top;
bounds.height -= (insets.bottom + insets.top);
```

Also maybe offset x too. Or alternatively, I would not be opposed to just resetting the client size to the safe mode size.

Issue #17137, opened 2023-10-25
Time Tracking Plugin doesn't track seaweed protection


TheHappyPittie:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

11

### Bug description

Time tracking plugin tracks most things but doesn't track if you've paid for protection for seaweed.  I've attached pics of the calquat tree that has it working correctly and my seaweed patches (which i've double checked are protected) are not showing that they're protected. 

### Screenshots or videos

![image](https://github.com/runelite/runelite/assets/114444106/ade9e8e3-66d8-4950-8e1a-6168b9d8f6d7)
![image](https://github.com/runelite/runelite/assets/114444106/52211766-c145-4399-8127-5e3380e79d21)

### RuneLite version

1.10.14.3
2.6.7

Birger9:
I can take a look at this issue :)

Adam-:
added in 1a6ede8902d8220941eae2fda21d326db853a061

Issue #16909, opened 2023-08-11
Cactus Farming Patch in Farming Guild does not show as protected


Hitpointler:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Cactus patch in Farming Guild does not show that the cactus seed is protected, when in actual fact it is. This has been this way since I can remember but been too lazy to report. No other bug that I have identified with RL in 30m farming exp. 

### Screenshots or videos

https://gyazo.com/b590ee61b19693fb521ece719061c0c3

https://gyazo.com/f24d794a5223b4cae8b5765ee8675816

### RuneLite version

RuneLite version: 1.10.9.4
Launcher version 2.6.7

### Logs

Fuck knows what any of this means, I just want a picture of a little basket next to one of my cactus patches tbh with you

Birger9:
This is a similar bug as https://github.com/runelite/runelite/issues/17137 so I will take a look at both of these issues

Adam-:
added in 1a6ede8902d8220941eae2fda21d326db853a061

PR #17234, opened 2023-11-27
time tracking payment support for multi patch farmers

Birger9:
Fix so that the time tracking plugin supports when a farmer protects two patches. This is a fix for issues: https://github.com/runelite/runelite/issues/16909 and https://github.com/runelite/runelite/issues/17137

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/PaymentTracker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/PaymentTracker.java
@@ -24,6 +24,7 @@
  */
 package net.runelite.client.plugins.timetracking.farming;
 
+import java.util.Objects;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import lombok.AccessLevel;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/PaymentTracker.java
```diff
@@ -90,6 +100,48 @@ private static String configKey(FarmingPatch fp)
...
+	{
+		if (event.getScriptId() != ScriptID.KEYINPUT_MATCHED)
+		{
+			return;
+		}
+
+		int pressedKey = client.getIntStack()[1];
+		int componentID = client.getIntStack()[0];
+
+		if ((pressedKey == 1 || pressedKey == 2) && widgetActive)
```
Adam-, line 133: I don't see what `widgetActive` is doing here. Why can't I delete it and have the code work better because it then doesn't require me to keypress Continue on some previous npc dialog?

runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/PaymentTracker.java
```diff
@@ -90,6 +100,48 @@ private static String configKey(FarmingPatch fp)
 		return fp.configKey() + "." + TimeTrackingConfig.PROTECTED;
 	}
 
+	@Subscribe
+	public void onMenuOptionClicked(MenuOptionClicked menuOptionClicked)
+	{
+		switch (menuOptionClicked.getMenuAction())
+		{
+			case NPC_THIRD_OPTION:
+				chosenPatch = 1;
```
Adam-, line 109: This should definitely either check the option is Pay (North/South/whatever) here, or maybe check the NPC is a farmer (or both?). This will get set any time you use op3 on any npc.

Adam-:
This has been added in 1a6ede8902d8220941eae2fda21d326db853a061. Additionally your pr:

1) does not work when clicking through the dialog to pay
2) does not work at Hodisious allotment due to it being North East/South West
3) does not work at the farming guild cactus or bush patch

PR #17235, opened 2023-11-27
clues: allow confining clues to regions

### LINKED ISSUE #5231: Clue plugin indicates wrong npc with identical name
Elite clue "Stop Crying!. Talk to the head"
Wrong head mourner highlighted in client.
This is correct hear mourner: http://oldschoolrunescape.wikia.com/wiki/Head_mourner
Highlighted one is currently: http://oldschoolrunescape.wikia.com/wiki/Head_mourner_(West_Ardougne)

In the npc id they are slightly different HEAD_MOURNER_5310 and HEAD_MOURNER_4258

Windows 10
RuneLite 1.4.15

Macweese:
This PR brings the ability for implementors of `NpcClueScroll` to confine the NPC marking to region IDs. This prevents false positive associations, between NPCs and clues by the clue scroll plugin, when displaying hints and overlays.

Currently the clue scroll plugin will, for certain clues, mark NPCs indiscriminately of whether they are linked to the clue or not. This is mainly due to using string comparisons, where the plugin will solicit for any NPCs with the same name and treat all matches as the intended target. This results in e.g. all NPCs named "Bartender" being marked, when in fact _a specific_ Bartender is sought.

To solve this, I played around with different solutions such as using NPC IDs/NPC compositions instead of string matching, and while they work, I ultimately opted for this approach since it entails a lesser change, avoiding shotgun surgery, for the same result. Additionally, some clues require area based value checks anyway[^1]. In my opinion though, I think using NPC IDs instead of strings is superior for these purposes.

The current changes only encompass `CrypticClue` clues, I skimmed through the other types but didn't find any clues which may infer restrictions to NPCs by variance in the same way.

Changes
• Add optional regions to the `NpcClueScroll` interface
• Amend respective `CrypticClues`

Fixes #5231

[^1]: For example the clue "_In a town where wizards are known to gather, search upstairs in a large house to the north._" involves the generic NPC "Man", but is restricted to Yanille.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -58,7 +58,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 {
 	static final List<CrypticClue> CLUES = ImmutableList.of(
 		new CrypticClue("Show this to Sherlock.", "Sherlock", new WorldPoint(2733, 3415, 0), "Sherlock is located to the east of the Sorcerer's tower in Seers' Village."),
-		new CrypticClue("Talk to the bartender of the Rusty Anchor in Port Sarim.", "Bartender", new WorldPoint(3045, 3256, 0), "The Rusty Anchor is located in the north of Port Sarim."),
+		new CrypticClue("Talk to the bartender of the Rusty Anchor in Port Sarim.", "Bartender", new WorldPoint(3045, 3256, 0), new int[]{12082}, "The Rusty Anchor is located in the north of Port Sarim."),
 		new CrypticClue("The keeper of Melzars... Spare? Skeleton? Anar?", "Oziach", new WorldPoint(3068, 3516, 0), "Speak to Oziach in Edgeville."),
 		new CrypticClue("Speak to Ulizius.", "Ulizius", new WorldPoint(3444, 3461, 0), "Ulizius is the monk who guards the gate into Mort Myre Swamp. South of fairy ring CKS."),
 		new CrypticClue("Search for a crate in a building in Hemenster.", CRATE_357, new WorldPoint(2636, 3453, 0), "House northwest of the Ranging Guild. West of Grandpa Jack."),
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -353,58 +354,70 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 	}
 
-	private CrypticClue(String text, String npc, WorldPoint location, String solution, boolean requiresLight)
+	private CrypticClue(String text, String npc, int objectId, WorldPoint location, int[] regions, String solution)
 	{
-		this(text, npc, location, solution);
+		this(text, npc, objectId, location, regions, solution, null);
+	}
+
+	private CrypticClue(String text, String npc, WorldPoint location, int[] regions, String solution)
```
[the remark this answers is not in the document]
Skretzo: A new class to instead specify a clue NPC through a combination of name, ID and/or region IDs could also be useful, but I understand that you wanted to keep this change small.

Adam-:
It might be better to use `WorldArea` to represent the npc regions instead of a list of region ids.

Adam-:
> It's a far more precise solution, no doubt, though I felt it was "messier" if that makes sense. The determining factor was that there are no instances of clues whose NPCs and regions coincide (as of now).

If the NPCs and regions don't coincide wouldn't WorldArea be better because the areas are not region constrained?

Adam-:
> The determining factor was that there are no instances of clues whose NPCs and regions coincide (as of now).

I interpreted the "NPCs and regions coincide" to mean where the NPCs cleanly fit into regions (or region boundaries).

This looks like instead you mean there are not two clues who have both the same npc *and* region? Having two clues with the same npc and region do not matter though because the clue matching is performed via text matching.

Adam-:
> > Having two clues with the same npc and region do not matter though because the clue matching is performed via text matching.
> 
> I don't understand this, I don't believe this is related. These changes affect clues with NPCs who share names, not two clues who share a name/NPC. As mentioned [here](https://github.com/runelite/runelite/pull/17235#issue-2012061918):

As far as I can tell whether or not two clues have the same region + npcs or not is completely irrelevant to anything being proposed here, and also to the existing implementation. It has no effect on either region vs world area. So I don't understand why you are using it as a "determining factor" in whether to use world area or region id.

Adam-:
> I intended to refer to _clues_ with the word coincide, not the clues' attributes of NPCs and regions. Meaning: there are no two clues, whose attributes of both NPCs and regions, are the same.
> 
> > This looks like instead you mean there are not two clues who have both the same npc and region?
> 
> Yes, that is what I meant to say.

I think the confusion here is actually you mean: there are no two clues with distinct npcs with the same name and region.

Having two clues with the same npc and region, regardless of your interpration of "npc" to mean either "npc name" or "npc id", is never a problem.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -58,7 +58,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 {
 	static final List<CrypticClue> CLUES = ImmutableList.of(
 		new CrypticClue("Show this to Sherlock.", "Sherlock", new WorldPoint(2733, 3415, 0), "Sherlock is located to the east of the Sorcerer's tower in Seers' Village."),
-		new CrypticClue("Talk to the bartender of the Rusty Anchor in Port Sarim.", "Bartender", new WorldPoint(3045, 3256, 0), "The Rusty Anchor is located in the north of Port Sarim."),
+		new CrypticClue("Talk to the bartender of the Rusty Anchor in Port Sarim.", "Bartender", new WorldPoint(3045, 3256, 0), new int[]{11569}, "The Rusty Anchor is located in the north of Port Sarim."),
```
Adam-, line 61: 11569 is 2880,3136, which is Musa point and is wrong.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -70,26 +70,26 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 		new CrypticClue("Identify the back of this over-acting brother. (He's a long way from home.)", "Hamid", new WorldPoint(3376, 3284, 0), "Talk to Hamid, the monk at the altar in the PvP Arena."),
-		new CrypticClue("In a town where thieves steal from stalls, search for some drawers in the upstairs of a house near the bank.", "Guard", DRAWERS, new WorldPoint(2611, 3324, 1), "Kill any Guard located around East Ardougne for a medium key. Then search the drawers in the upstairs hallway of Jerico's house, which is the house with pigeon cages located south of the northern East Ardougne bank."),
+		new CrypticClue("In a town where thieves steal from stalls, search for some drawers in the upstairs of a house near the bank.", "Guard", DRAWERS, new WorldPoint(2611, 3324, 1), new int[]{10547}, "Kill any Guard located around East Ardougne for a medium key. Then search the drawers in the upstairs hallway of Jerico's house, which is the house with pigeon cages located south of the northern East Ardougne bank."),
```
Adam-, line 80: there are guards outside of 41,51 that work for this too, such as by the north bank in 41,52. I would probably include the whole ardy area.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -117,11 +117,11 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 		new CrypticClue("Even the seers say this clue goes right over their heads.", CRATE_14934, new WorldPoint(2707, 3488, 2), "Search the crate on the Seers Agility Course in Seers Village"),
 		new CrypticClue("Speak to a Wyse man.", "Wyson the gardener", new WorldPoint(3026, 3378, 0), "Talk to Wyson the gardener at Falador Park."),
-		new CrypticClue("You'll need to look for a town with a central fountain. Look for a locked chest in the town's chapel.", "Monk" , CLOSED_CHEST_5108, new WorldPoint(3256, 3487, 0), "Search the chest by the stairs in the Varrock church. Kill a Monk in Ardougne Monastery to obtain the key."),
+		new CrypticClue("You'll need to look for a town with a central fountain. Look for a locked chest in the town's chapel.", "Monk" , CLOSED_CHEST_5108, new WorldPoint(3256, 3487, 0), new int[]{12086}, "Search the chest by the stairs in the Varrock church. Kill a Monk in Ardougne Monastery to obtain the key."),
```
Adam-, line 124: 12086 is 47,54 which is the edgeville monastery

Adam- (changes requested):
it may be worth using a builder at some point, the constructor overloads are starting to get a little much

Adam-:
added in 3084c56929b0f0e19e66366c2a6242d2157ec6e6

PR #17243, opened 2023-11-30
raids: Fix reloading raids from inside

vikke1234:
The Varbit `RAID_STATE` changes to 5 when you
reload the raid. Checking for this and resetting
the plugin now gives the expected behavior.


Adam-:
fixed in fb98120b41560ba7c0c6195424b65f8e262836ec

vikke1234:
@Adam- it seems that if you start two raids in one FC, https://github.com/runelite/runelite/commit/fb98120b41560ba7c0c6195424b65f8e262836ec breaks

Adam-:
Hello. Can you provide an explanation for what you mean by two raids in one FC? Do you have steps to reproduce it?

vikke1234:
It turns out that when you reload a raid, it actually leaves the party.  So if on another account you are by the board you can make another raid in the same friends chat. This raid works just as well as the original one, but you don’t need to change friends chats.

Adam-:
Can you provide steps to reproduce this? Is it just to reload the raid on account 1 and then make a new raid on account 2?

vikke1234:
Yes

Issue #17256, opened 2023-12-06
Inferno ground items despawn timer incorrect


emiljensen2:
### Type

Incorrect behavior

### Operating System

None

### OS Version

Windows 10

### Bug description

The despawn timer for items dropped in the inferno is based on them despawning in 30 minutes, while in reality they despawn after 3 hours.

### RuneLite version

rl 1.10.17.1
rev 218
launcher 2.6.8

emiljensen2:
I'm afraid I don't have a direct jagex source or a video, I tried to find a clip on sabfas's twitch channel or 420streamer, but best I could find is it being written on screen.
https://www.twitch.tv/sabfas/clip/GleamingInnocentGazelleLitty?featured=false&filter=clips&range=all&sort=time
or just people writing it.
https://www.reddit.com/r/2007scape/comments/ifcmbu/inferno_bug/
https://www.reddit.com/r/2007scape/comments/hp66d1/til_chins_disappear_upon_dying_what_bad/fxnqo0s/
or inferno channel in we do raids.

emiljensen2:
![image](https://github.com/runelite/runelite/assets/55587410/66775551-52ca-4aa0-a117-ed58444aef05)

emiljensen2:
I've tweeted at ash, he'll probably reply in the morning.

emiljensen2:
Confirmed 3 hours
https://twitter.com/JagexAsh/status/1732808272627871776

PR #17266, opened 2023-12-09
fix ground items inferno despawn timer

thecage555:
Made change to GroundItemsOverlay.java to amend wrong despawn timer in the Inferno as described in issue #17256


zachtyll (approved):
LGTM.

Nitpick: It would be great if we could use the name [INFERNO_REGION_ID](https://github.com/runelite/runelite/blob/b7d88347ea32611b262690320914e18516286825/runelite-client/src/main/java/net/runelite/client/plugins/timers/TimersPlugin.java#L131), but it is clear enough and the other ID's in this file don't use ID as a suffix.

PR #17290, opened 2023-12-18
crowdsourcing plugin: Fix NullReferenceException in CrowdsourcingDialogue sanitize function

fox091:
Add a guard clause for a null username before attempting to sanitize the dialogue message. If the username is null, explicitly return null instead of the message, as it cannot be sanitized.

Change all uses of sanitize() to only store data in CrowdsourcingManager if the sanitized message is not null.

If username is null, a NullReferenceException will be thrown when attempting to use the username as the pattern for replacement, causing a client crash.

This fixes an inconsistent crash I have been getting for quite a while, with the following stack trace:
```
java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "this.pattern" is null
	at java.base/java.util.regex.Pattern.<init>(Pattern.java:1428)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1069)
	at java.base/java.lang.String.replaceAll(String.java:2944)
	at net.runelite.client.plugins.crowdsourcing.dialogue.CrowdsourcingDialogue.sanitize(CrowdsourcingDialogue.java:57)
	at net.runelite.client.plugins.crowdsourcing.dialogue.CrowdsourcingDialogue.onChatMessage(CrowdsourcingDialogue.java:142)
	at net.runelite.client.callback.Hooks.post(Hooks.java:194)
	at dh.ah(dh.java:48047)
	at ob.at(ov.java:19)
	at client.jt(client.java:46947)
	at client.hm(client.java:3033)
	at client.bp(client.java:1161)
	at bm.an(bm.java:394)
	at bm.ov(bm.java)
	at bm.run(bm.java:33867)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

I added a test for `onChatMessage` that reproduces the problem.  I also added some test cases to cover all of the `onChatMessage` and `sanitize` code.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/dialogue/CrowdsourcingDialogue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/dialogue/CrowdsourcingDialogue.java
@@ -25,7 +25,9 @@
 
 package net.runelite.client.plugins.crowdsourcing.dialogue;
 
+import java.util.ArrayList;
 import javax.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.Client;
 import net.runelite.api.events.ChatMessage;
@@ -35,9 +37,11 @@ import net.runelite.api.widgets.Widget;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.plugins.crowdsourcing.CrowdsourcingManager;
 
+@Slf4j
 public class CrowdsourcingDialogue
 {
 	private static final String USERNAME_TOKEN = "%USERNAME%";
+	private static final String NON_BREAKING_SPACE = " ";
 
 	@Inject
 	private Client client;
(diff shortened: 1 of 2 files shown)


Adam-:
Can you run in debug mode and figure out what chat message is triggering this?

fox091:
Sure!  At the rate it's been crashing lately, I should hopefully find out pretty quickly.  I crashed 4 times today during 1 run of the fight caves, which drove me to diving into it.

Can I build the client with debug mode on permanently, or does it have to be through the `--debug` flag?  I'm running through the [Bolt launcher](https://github.com/Adamcake/Bolt) and I'm not sure how to change the command line args.

Adam-:
You have to either pass `--debug`, or run `--configure` and enable debug there. This won't be causing a crash and at worst will just log an exception. So it might be a red herring if you are actually getting crashes.

fox091:
Oh, interesting...  The timing always lined up for the client crashing, so I just assumed it was the cause.

When you say `run --configure and enable debug there`, do you mean run the build RuneLite jar with `--configure`?

Adam-:
Yes, `java -jar RuneLite.jar --configure`

fox091:
Awesome, thanks!  I got it running in debug.  It's getting late, so I will run it tomorrow evening and report back when I find something.

fox091:
I ran for the full 6 hours before getting nerd logged with no crash.  Will report back once I get another crash.

fox091:
I'm now confident that the NullReferenceException is in fact a red herring.

I just got a crash VERY similar to the ones I usually do.  The client completely freezes before closing after 5-10 seconds.  I'm not quite sure how to read the jvm_crash log, but I _think_ the 117 HD plugin is the culprit.  I have yet to find out which message ends up with the player name being null.

I know plugins on the plugin hub are not relevant here, but would you like me to drop my logs here to verify @Adam- ?

In the mean time, I'll keep running my local build to determine what the message is, if I ever manage to get a null player name again soon.

PR #17304, opened 2023-12-25
Fix exported and renamed bank tags containing more items than necessary.

geheur:
Currently it adds all items ids that are in other bank tags that could be included in the target bank tag if those items were in the bank. e.g., if you have a specific clue scroll in one bank tag and variation tagged clues in another, when you export the second tag, it will contain not only the variation tag clue scroll id, but also the specific clue scroll id in the first bank tag.

Tested export/import, renaming bank tags, and ge plugin search, which all use the changed method.


geheur:
This is not a cosmetic issue, as it can cause a variation id to show up in exported tags that aren't in the tag. This will cause the tag to import incorrectly. e.g. a tag with an uncharged glory will contain both uncharged glory and all glory variants when exported.

PR #17314, opened 2023-12-29
Music Plugin: Reapply ambient sounds on shutdown

JZomDev:
When muteAmbientSounds is enabled the game state should be reset to Loading to reload the Ambient Sounds that were cleared.


Adam-:
fixed in 52f66163c0e6045f290955d34a56bc74f082b4b2

PR #17317, opened 2023-12-30
clues: rename rune heraldic items in clue steps

Felanbird:
In https://togithub.com/runelite/runelite/pull/6811 this was attempted to be changed to `Rune (h) helm/shield` but was never rebased

In https://github.com/runelite/runelite/pull/8783 this was attempted to be changed to `Rune helm/shield (h)` and also to use a range of items but was closed after 1 month

So I've gone ahead and done both a name change & changing the items to a `AnyRequirementCollection`

Notes: 
Do we still want this change? It was back in 2019 but I personally think it would be less confusing for users to be able to search what is listed `Rune helm (h` on the ge and be given the correct items.

Rune helms (h) are variations from the normal rune full helm, and contain the trimmed and gold variation. Since these variations are as old as time itself, I figured changing it is not something we'd want to do, so I just went ahead and filtered the non heraldic helmets. 

I'll open a PR that moves all the `AnyRequirementCollection`'s to their own file, after this and https://togithub.com/runelite/runelite/pull/17273 are merged

Closes #6811


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -103,6 +103,15 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
 			.orElseGet(Stream::empty)
 			.map(ItemRequirements::item)
 			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection ANY_RUNE_HERALDIC_SHIELD = any("Rune shield (h)",
+		ItemVariationMapping.getVariations(RUNE_SHIELD_H1).stream()
+			.map(ItemRequirements::item)
+			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection ANY_RUNE_HERALDIC_HELM = any("Rune helm (h)",
+		ItemVariationMapping.getVariations(RUNE_FULL_HELM).stream()
+			.filter(itemId -> itemId != RUNE_FULL_HELM && itemId != RUNE_FULL_HELM_G && itemId != RUNE_FULL_HELM_T)
```
Nightfirecat, line 112: In my opinion this is a bug in our variation mapping, and should be fixed at the source (ie. https://github.com/abextm/runelite-cache-code-updater/issues/42) instead of this clunky workaround.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -103,6 +103,15 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
 			.orElseGet(Stream::empty)
 			.map(ItemRequirements::item)
 			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection ANY_RUNE_HERALDIC_SHIELD = any("Rune shield (h)",
+		ItemVariationMapping.getVariations(RUNE_SHIELD_H1).stream()
+			.map(ItemRequirements::item)
+			.toArray(SingleItemRequirement[]::new));
+	private static final AnyRequirementCollection ANY_RUNE_HERALDIC_HELM = any("Rune helm (h)",
+		ItemVariationMapping.getVariations(RUNE_FULL_HELM).stream()
```
Nightfirecat, line 111: Assuming the below comment re: item variation mapping is addressed separately, the following change should also happen.
```suggestion
		ItemVariationMapping.getVariations(RUNE_HELM_H1).stream()
```

PR #17330, opened 2024-01-02
Add option to flash taskbar

SuperNerdEric:
Add the option for notifications to flash the task bar. This is different from requesting focus as it will not change what window is currently focused, just flash the taskbar.

This makes it easier to see at a glance which RuneLite windows have a notification without stacking up multiple tray notifications if you are playing on many accounts at once.

I have tested on Windows 10, but I am not totally clear of the behavior of [requestWindowUserAttention](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/awt/Taskbar.html#requestWindowUserAttention-java.awt.Window-) on other platforms.

flash taskbar
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -208,6 +208,11 @@ public class Notifier
 			mouseLastPressedMillis = client.getMouseLastPressedMillis();
 		}
 
+		if (runeLiteConfig.notificationFlashTaskbar())
+		{
+			clientUI.flashTaskbar();
+		}
+
 		log.debug(message);
 	}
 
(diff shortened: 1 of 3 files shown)


Adam-:
This would probably be a good addition to the Request focus setting. The "request" setting of it will either have this behavior or will bring the window to focus, but it depends on some external factors we don't really understand.

SuperNerdEric:
> This would probably be a good addition to the Request focus setting. The "request" setting of it will either have this behavior or will bring the window to focus, but it depends on some external factors we don't really understand.

Would it be appropriate to add a drop down option here called "Flash"?
image

Adam-:
Probably "Task bar" would be a better name

SuperNerdEric:
I moved it over to the Request focus options. I'm not sure if you still want the existing "Request" option as that one seems to behave differently, but I left it in.

image

PR #17344, opened 2024-01-07
cryptic clues: fix description of medium clue step solution

pkhermouch:
Noticed this was off while doing this clue step during Leagues. [Source](https://oldschool.runescape.wiki/w/Donovan_the_Family_Handyman)


Adam-:
This is actually technically correct currently, and is using UK floor numbering instead of American floor numbering. But changing it to top seems much better as it is unambiguous.

PR #17385, opened 2024-01-16
xp updater: add runetracker support

Limero:
This adds support for RuneTracker to the xp updater. I have built and tested it locally and it works

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xpupdater/XpUpdaterConfig.java
@@ -44,6 +44,17 @@ public interface XpUpdaterConfig extends Config
 
 	@ConfigItem(
 		position = 2,
+		keyName = "runetracker",
+		name = "RuneTracker",
+		description = "Automatically updates your stats on runetracker.org when you log out"
+	)
+	default boolean runetracker()
+	{
+		return false;
+	}
+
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Are you associated with runetracker at all? All information I have points to this not being an active project, and we have not heard of interest from players (or from runetracker developers/maintainers) for this feature.

Limero:
> Are you associated with runetracker at all? All information I have points to this not being an active project, and we have not heard of interest from players (or from runetracker developers/maintainers) for this feature.

I'm not affiliated with RuneTracker. I have been using the service since 2014 and it's still my preferred tracker, and I know other people who also use it. Especially people who have years of history on the graphs and exp records.

It would convenient if Runelite automatically updated this site, as I currently do this manually. It might not be an active project (although, the RS3 part saw an update Aug 2023) but until it stops working or shuts down, this feature is useful for some people.

Adam-:
I know some of the people who operate it, I dropped them a message about it and will see what they think.

deltacore47:
Hey folks - 

I am the main developer for RuneTracker ([proof](https://runetracker.org/proof.html)) and it is still chugging along in maintenance mode for both OSRS and RS3. I'm not opposed to this integration if there is interest.

That said, it would be better to call this page (replace Cowman_133 with user to lookup) to prompt tracker stats to update: https://rscript.org/lookup.php?type=stats07&user=Cowman_133

PR #17389, opened 2024-01-17
Item variations: consolidate cosmetic variants and add Dinh's bulwark

### LINKED ISSUE #17303: Trailblazer reward item variations not present.
### Type

Incorrect behavior

### Operating System

None

### OS Version

_No response_

### Bug description

The "Blazing blowpipe" should be considered a variation of the toxic blowpipe. It is a cosmetic variant, just like the serpentine helms are.
Same for the "Dinh's blazing bulwark" and the dinh's bulwark.

Twisted ancestral/regular ancestral, as well as sanguine torva and regular torva should also be variations.

### Screenshots or videos

_No response_

### RuneLite version

1.10.18-SNAPSHOT/b7d88347ea32611b262690320914e18516286825

### Logs

_No response_

timleafy:
Resolves #17303 .


geheur:
That file is autogenerated so manually changing it is not the right way to do this.

https://github.com/abextm/runelite-cache-code-updater/blob/master/src/main/java/net/runelite/cache/codeupdater/apifiles/ItemVariationsUpdate.java

raiyni:
As mentioned, this should be fixed in the updater if it's not picking up these items.

timleafy:
My bad—thanks for pointing me to the updater.

PR #17390, opened 2024-01-17
soulreaper axe in regenmeter plugin

OmbudRov:
Added Soulreaper Axe to the Regeneration Meter Plugin and added a VarpValue for Soul Stacks
Part of #16819 

Copy of #16992 but i didnt mess up this time

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterOverlay.java
@@ -97,6 +97,14 @@ class RegenMeterOverlay extends Overlay
 						(int) DIAMETER, (int) DIAMETER);
 				}
 			}
+			if (!plugin.wearingSoulreaperAxe)
+			{
+				renderRegen(g, ComponentID.MINIMAP_HEALTH_ORB, plugin.getSpecialPercentage(), SPECIAL_COLOR);
+			}
+			else if (config.showSoulStack())
+			{
+				renderRegen(g, ComponentID.MINIMAP_HEALTH_ORB, plugin.getSoulStackPercentage(), SPECIAL_COLOR);
+			}
 
 			renderRegen(g, ComponentID.MINIMAP_SPEC_ORB, plugin.getSpecialPercentage(), SPECIAL_COLOR);
 		}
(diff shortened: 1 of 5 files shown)


OmbudRov:
Added code to handle the case where a player reaches 5 soul stack and the plugin didnt show the decay

Nightfirecat:
Mentioned this in Discord, but this would reset the delay if your thrall landed damage while at 5 soul stack. I think tracking the axe attack animation is probably the most reliable way to handle the max charge case.

OmbudRov:
alright, i think i've fixed everything and changed the max stack check to the attack animation

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
```diff
@@ -185,8 +202,38 @@ else if (currentHP > maxHP)
...
+	{
+		Player local = client.getLocalPlayer();
+
+		if (event.getActor() != local)
+		{
+			return;
+		}
+
+		int animId = local.getAnimation();
+		if (animId == 10172 && client.getVarpValue(VarPlayer.SOUL_STACK) == 5)
```
Nightfirecat, line 230: This animation ID should go into the AnimationID api class.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterOverlay.java
```diff
@@ -97,8 +97,14 @@ public Dimension render(Graphics2D g)
 						(int) DIAMETER, (int) DIAMETER);
 				}
 			}
-
-			renderRegen(g, ComponentID.MINIMAP_SPEC_ORB, plugin.getSpecialPercentage(), SPECIAL_COLOR);
+			if (!plugin.wearingSoulreaperAxe)
```
Nightfirecat, line 100: This would cause a player wearing the axe with `config.showSoulStack()` to have nothing rendered. I think this check should instead be:

```java
if (!plugin.wearingSoulreaperAxe || !config.showSoulStack())
{
	// ...
}
else
{
	// ...
}
```

OmbudRov: i'm not sure if i follow correctly, but the situation you're describing shows the stack decay without the changes.

runelite-client/src/main/java/net/runelite/client/plugins/regenmeter/RegenMeterPlugin.java
OmbudRov: added this

PR #17426, opened 2024-01-29
timers: add scurrius food pile timer

jamieluckett:
Had a quick look for a varbit but couldn't see anything (though am very new to this world) so using a chat message 

Edit: Check now uses the varbit


YvesW:
Here you got the cs2 code for the buff bar:
```
case buff_scurrius_food_pile_cooldown_776 :
	if (%varcint1188 > 0) {
		return(~script5942(%varcint1188, %varbit9581, 100));
	}
```

The varbit might work like the GWD altar one for which NFC has made a cool PR here: https://github.com/runelite/runelite/pull/16493
Haven't touched scurrius yet though, so make sure to check it ingame and potentially clone https://github.com/Joshua-F/cs2-scripts and ctrl+shift+f through it (or check potentially relevant scripts like 5942) if you can't figure it out. For example, the WC group bonus uses a Varp to determine if it should be shown on the buff bar, and then the varbit determines the number on the buff bar.

Also make sure to fix the checkstyle violation.

jamieluckett:
@YvesW Thanks for the reply, have pushed a commit to fix the code styling - will also look into that CS2 script

jamieluckett:
@YvesW Cheers for leading me on to the CS2 scripts to investigate the varbit - have tested changes now with some logging in/out and my new changes seem to be working great using varbit 9581 👍

PR #17452, opened 2024-02-05
Added Some Mappings to VarPlayer

Legacy35:
added  PREV_ATTACK_STYLE And AUTO_RETALIATE .


Nightfirecat:
We don't add entries to this file unless they are going to be used in the core client.

PR #17453, opened 2024-02-05
Add Per-Item Highlights to Ground Items

### LINKED ISSUE #12274: Allow the ground items plugin to create multiple groups of color-coded items
Being able to tag ground items with multiple colors would make identifying item types at a glance much simpler. At the moment you can only highlight items in one group or color-code by value. Attached is a concept I made in Photoshop to show what it could be used for.
image

### LINKED ISSUE #14233: Custom Ground Marker Fill Colors - Parity with Screen Markers and NPC Indicators

bheisenberg:
Offers the player more ground item highlight options by enabling them to set a custom highlight color for individual items!

**Instructions**
To set a custom highlight color for an item:

- Shift + Right Click an already highlighted ground item
- Navigate to the "Set highlight color" submenu

Within the submenu, the player can select from the following options:
- "Pick" to choose a new color from the color picker
- "Set color" to choose from one of the top 5 most used highlight colors
- "Reset" to return to the default highlight color and remove the config data for the item

**Images**
image
_A few items that players might want to differentiate with highlights_
image
_A look at the "Set highlight color" submenu. The current color is always excluded from the list_
image
_The color picker always defaults to the current color_
image
_Lootbeam support!_
image
_Valued items + button now displays correctly_

fixes #14233, fixes #12274

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemHighlight.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemHighlight.java
@@ -1,0 +1,19 @@
+package net.runelite.client.plugins.grounditems;
+
+import lombok.Getter;
+
+import java.awt.*;
+
+@Getter
+public class GroundItemHighlight
+{
+	private final String name;
+	private final Color color;
+
+	public GroundItemHighlight(String name, Color color)
+	{
+		this.color = color;
+		this.name = name;
+	}
+
+}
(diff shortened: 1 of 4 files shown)


bheisenberg:
This is currently implemented to only work on highlighted items. I have a branch available that makes highlights visible for all non-hidden items. That approach can be implemented into this PR if requested.

Adam-:
The way that this stores and checks the color overrides seems slow. Since it is doing an O(n) check over each override per item per frame. And also it is doing string comparisons of the item names instead of using item ids like how it should be. I think it should be possible to store the highlight color (and maybe also the hidden/highlight boolean) in GroundItem and cache it. And then store the overrides in the config keyed by the item id.

Adam-:
added in bd05ac363c074f9d11469111932f9538d6393e34

Issue #17218, opened 2023-11-23
Idle notifier: not notifying for uncharged crystal felling axe


YvesW:
### Type

Bug

### Operating System

None

### Bug description

A user reported that idle notifier did not work when cutting redwoods with his uncharged crystal felling axe. Idle notifier was set to 1000ms delay, animations were enabled. Notifications while focused were enabled and game notification messages were enabled. As can be seen below, no notification is fired in safe mode.

https://discord.com/channels/301497432909414422/442492468307427330/1177071256760238080
Direct link to the file:
https://cdn.discordapp.com/attachments/442492468307427330/1177071255778775060/Replay_2023-11-23_13-13-26_-_Trim.mp4

He noted that it started working after charging his axe. https://github.com/runelite/runelite/commit/ff88db263565fa62b25a0a1df0c95fd3e7778a0f skips an id, 10073, so I expect that the animation id (very likely 10073) is just missing. If that is the case, it should also be added to e.g. the woodcutting plugin.

### RuneLite version

-

YvesW:
Ost has confirmed that the axe does indeed use a different animation ID (10073; see image below) that is missing!
Next questions if someone ever has those tools:
- What's the animation ID for the normal crystal axe when it's uncharged (charged: 8324)?
- What's the animation ID for the crystal harpoon when it's uncharged (charged: 8336)?
- What's the animation ID for the crystal pickaxe when it's uncharged when...
normal mining? (charged: 8347)
MLM mining? (charged 8345)
crashed star mining? (charged 8350)

This is potentially also a problem for uncharged infernal tools! Thus, if someone has one of those, it's also worth it to check the animation IDs.

![image](https://github.com/runelite/runelite/assets/7929021/a8294201-1301-440b-83bb-831b46f1a359)

lexigorse:
This can be closed now, correct?

YvesW:
I kept it open because I was still wondering if the animation ID for the normal crystal axe (uncharged) and crystal harpoon (uncharged) were correct, but I'm fine with closing it since we haven't received any reports about those.

PR #17458, opened 2024-02-06
idle notifier: add crystal felling axe (inactive) animation

YvesW:
The crystal felling axe (inactive) uses a different animation id than the charged one for some reason, as outlined in: #17218
Please also refer to that issue for a picture of the animation id.


Adam-:
This needs to be added to at least: wc, wintertodt, idle notifier plugins

YvesW:
Apologies, I should have used ctrl+B instead of rushing out a PR before I had to leave. Fixed it now.
Obviously feel free to squash the commits if you prefer to.

PR #17462, opened 2024-02-06
bank: Support group storage with search hotkey

PhraZier:


runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -132,6 +133,26 @@ public void keyPressed(KeyEvent e)
...
+					clientThread.invoke(() ->
+					{
+						Widget searchButton = client.getWidget(ComponentID.GROUP_STORAGE_SEARCH_BUTTON);
+						if (searchButton == null || searchButton.isHidden())
+						{
+							return;
+						}
+
+						Object[] searchToggleArgs = searchButton.getOnOpListener();
+						// the search toggle script requires 1 as its first argument
```
Adam-, line 149: what? why?

PhraZier: This depends on script 5275, where the script returns early when "opindex0 ! 1".

Adam-: Ah, the argument is event_op:

```
if_setonop("shared_bank_search_toggle(event_op, $component6, $component0, $component7, $component8, $int25, $component1, $comsubid22, $component2)", $component6);
```

What you have is wrong because you are overwriting event_op on the listener, which would make other ops not work correctly. Even if there are no other ops currently, this is a bad idea because it will break quietly later if other ops are ever added. 

This type of thing is good to document with comments. Such as `this calls [clientscript,shared_bank_search_toggle]` and `the first argument is event_op and we need to spoof op1`.

PhraZier: I have changed it to use a copy of array based on your suggestion in discord, instead of directly overwriting event_op on the listener.

runelite-client/src/main/java/net/runelite/client/plugins/bank/BankPlugin.java
```diff
@@ -132,6 +134,33 @@ public void keyPressed(KeyEvent e)
...
+							return;
+						}
+
+						Object[] searchToggleArgs = searchButton.getOnOpListener();
+						if (searchToggleArgs == null)
+						{
+							return;
+						}
+
+						searchToggleArgs = Arrays.copyOf(searchToggleArgs, searchToggleArgs.length);
```
Adam-, line 155:
```suggestion
						searchToggleArgs = searchToggleArgs.clone();
```

PR #17482, opened 2024-02-10
Added infobox for unaggro calibration

DapperMickie:


Adam-:
I would prefer to have just the infobox for the uncalibrated message instead of an option between both the infobox and the overlay panel

DapperMickie:
@Adam- Alexsuperfly and ron had a discussion in the discord channel and said that having the infobox only, would still result in people asking why the plugin is not working. That's why I changed it to being optional. As you can see in my "added infobox for calibration" commit, it was exactly like you just said, the infobox only.

Adam-:
I think that it should work like this comment details https://github.com/runelite/runelite/pull/17109#issuecomment-1875960335. As long as you always show the uncalibrated infobox that seems like it would be ok.

DapperMickie:
So reverting back to my previous commit should be fine then?

PR #17483, opened 2024-02-10
Screenshot: Embed P3 color profile on macOS screenshots

jeremybahadirli:
Display-P3-vs-sRGB

Every currently-manufactured Apple device uses a wide-gamut (P3) display.

The OS is responsible for mapping colors to the display's color space. For instance, on a P3 display, when the client contains the color #FF0000, it is mapped to far-right corner of the _Display P3_ region in the illustration above. This is the display's maximum red value.

Currently on macOS, screenshots are saved with no color profile. macOS interprets image files with no color profile as sRGB. This means that when viewing a RuneLite _screenshot_, the same #FF0000 color is displayed as the rightmost corner of the _sRGB_ region. Therefore, on macOS, RuneLite screenshots appear slightly desaturated compared to the game client.

This PR implements embedding the "Display P3.icc" color profile when screenshots are saved on macOS. This utilizes Apple's `sips` - scriptable image processing system - which is standard on all macOS installations.

## Examples
Digital Color Meter is set to "Display native values", meaning it is reading colors relative to the display's color space. All color samples are taken from the #FF0000 region in the sidebar.

### Game Client
#FF0000 is correctly rendered as the display's maximum red value:
Real Game

---

### Screenshot (no color profile)
#FF0000 is incorrectly rendered as the maximum red value of the _sRGB_ color space, which is below the display's maximum red value:
Screenshot - No Profile

---

### Screenshot (Display P3 profile)
#FF0000 is correctly rendered, and is consistent with the appearance of the game client:
Screenshot - P3 Profile


abextm:
The default color space of awt applications is supposed to be sRGB, so the screenshot using that is correct. If anything the bug is that java (and rlawt) does not actually do this, but even if that was fixable, I don't think mac users would be happy to have the saturation of the application lowered.

jeremybahadirli:
While I agree in theory, in practice I'm of the opinion that the most important thing is for screenshots to match the game client.

abextm:
embedding a random color profile into the screenshot will just make it not match for everyone who does not have the same display, which is most people.

PR #17492, opened 2024-02-13
woodcutting: prevent division by zero, wait until 3 logs cut for hourly rate

lhoenig:
A division by zero occurs in the calculation of the number of logs/bark cut per hour in the Woodcutting plugin if 0 < `elapsed` < 1ms for `elapsed` the time since the creation of the `WoodcuttingSession`, because `elapsed` is not zero, thus the condition is true, but the calculation uses the value converted to milliseconds, which is zero on the first action (a few thousand nanoseconds).

This results in the following visible bug:
Screenshot 2024-02-13 at 12 32 04

This PR fixes this by only updating `logsPerHr`,`barkPerHr`,  if at least one millisecond has passed since the creation of the `WoodcuttingSession`, i.e. on the second action, since the session is created after the first log/bark.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingSession.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/woodcutting/WoodcuttingSession.java
@@ -70,9 +70,10 @@ class WoodcuttingSession
 		bark += num;
 
 		Duration elapsed = Duration.between(start, Instant.now());
-		if (!elapsed.isZero())
+		long elapsedMs = elapsed.toMillis();
+		if (elapsedMs > 0)
 		{
-			barkPerHr = (int) ((double) bark * Duration.ofHours(1).toMillis() / elapsed.toMillis());
+			barkPerHr = (int) ((double) bark * Duration.ofHours(1).toMillis() / elapsedMs);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Alternatively, this could be fixed by using the same implementation as the xp tracker, which iirc limits per-hour calculations to a minimum of 1 minute duration. See: #2271

lhoenig:
> Alternatively, this could be fixed by using the same implementation as the xp tracker, which iirc limits per-hour calculations to a minimum of 1 minute duration. See: #2271

Agreed, another alternative could be to do it like the Fishing plugin, which only sets (displays) the per-hour value after three actions in a session. As I wouldn't expect the logs/hr estimates to be so wildly inaccurate as in the case of the XP tracker, we wouldn't need to make the player wait a whole minute.

lhoenig:
I added the three logs cut condition and also left the 1 ms check in, so we under no circumstances divide by zero.

For the bark I am not sure if such a three-barks-gained condition makes sense, as it is gained in variable amounts I think? I don't have first-hand experience with Forestry so cannot really judge.. it seems like a good idea to keep the 1ms check for the bark as well though as I think it might be possible to gain some without first felling a tree in some situations, i.e. at the very start of a woodcutting session.

PR #17494, opened 2024-02-14
Fix searching in view all tag tabs

vikke1234:
Currently when viewing all available tags (by right clicking the + in the upper left corner and pressing `view all tags`) searching the bank breaks, this is due to it only closing the tag if there's an "active" tag. The view all tag tabs is a special type of tag and needs to be checked separately.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
@@ -245,7 +245,7 @@ public class TabInterface
 		}
 		else if (event.getScriptId() == ScriptID.BANKMAIN_SEARCH_TOGGLE)
 		{
-			if (activeTab != null)
+			if (activeTab != null || tagTabActive)
 			{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/banktags/TagManager.java
```diff
@@ -122,6 +130,10 @@ boolean findTag(int itemId, String search)
 
 		Collection<String> tags = getTags(itemId, false);
 		tags.addAll(getTags(itemId, true));
+		if (strict)
+		{
+			return tags.stream().anyMatch(tag -> tag.equals(Text.standardize(search)));
+		}
 		return tags.stream().anyMatch(tag -> tag.startsWith(Text.standardize(search)));
```
Josh-Renwick-99, line 137: This could probably be changed to be a bit more user friendly, no? 

```suggestion
		return tags.stream().anyMatch(tag -> tag.contains(Text.standardize(search)));
```

Josh-Renwick-99:
Looks good other than the comments left in.

Issue #16943, opened 2023-08-18
Loot tracker doesn't track PNM loot


YvesW:
### Type

Incorrect behavior

### Operating System

None

### Bug description

Reported by discord user lck.alex
Tested in safe mode and recorded on video
I suspect Jagex might have messed with the (P)NM drops like they did with the DT2 bosses, but not sure and I personally haven't tested it ingame

### Screenshots or videos

https://github.com/runelite/runelite/assets/7929021/8b7dddd7-e004-454e-a8b0-d66287a727ad

### RuneLite version

RL 1.10.10.2
Launcher 2.6.8

### Logs

[client.log](https://github.com/runelite/runelite/files/12383776/client.log)
[launcher.log](https://github.com/runelite/runelite/files/12383777/launcher.log)

Slayerlouver:
got the same exact problem. Randomly show some loot. Even though Bigbones also should go in collection log every time, it just doesn't mostly.

PR #17502, opened 2024-02-16
fix nightmare loot tracking

umer-rs:
Fixes https://github.com/runelite/runelite/issues/16943

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/LootManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/LootManager.java
@@ -80,7 +80,7 @@ public class LootManager
 	private WorldPoint krakenPlayerLocation;
 
 	private NPC delayedLootNpc;
-	private int delayedLootTickLimit;
+	private int delayedLootTick;
 	private List<WorldArea> delayedLootAreas;
 
 	@Inject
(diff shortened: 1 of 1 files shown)


Adam-:
Why are you changing the wall beast area logic? I believe it works and has no issues. And what is the significance of this 3 tick loot spawning window? Why not 1 tick? What we have now uses the entire duration specified in `delayedLootTickLimit` starting from the npc change, which I am not really opposed to. Though the 15 tick nightmare death is rather exaggerated.

umer-rs:
> Why are you changing the wall beast area logic? I believe it works and has no issues. And what is the significance of this 3 tick loot spawning window? Why not 1 tick? What we have now uses the entire duration specified in `delayedLootTickLimit` starting from the npc change, which I am not really opposed to. Though the 15 tick nightmare death is rather exaggerated.

> Why are you changing the wall beast area logic? I believe it works and has no issues.
It is bugged. If you kill it with a weapon that has a delay, and run away, it doesn't work.

> And what is the significance of this 3 tick loot spawning window?
The 2 tick loot window is the exact tick the loot spawns on, and 1 tick leeway after that. If we allow for checking every tick, dropping items will proc the `processDelayedLoot`. I'm not opposed to making it only work for 1 specific tick though.

Adam-:
> > Why are you changing the wall beast area logic? I believe it works and has no issues.
> 
> It is bugged. If you kill it with a weapon that has a delay, and run away, it doesn't work.

This type of information is good to include in either a comment, the commit message, or the PR description. Ideally, this should be in a separate commit. 

> 
> > And what is the significance of this 3 tick loot spawning window?
> 
> The 2 tick loot window is the exact tick the loot spawns on, and 1 tick leeway after that. If we allow for checking every tick, dropping items will proc the `processDelayedLoot`. I'm not opposed to making it only work for 1 specific tick though.

```java
		if (delayedLootNpc != null && --delayedLootTickLimit < 2)
		{
			processDelayedLoot();
			// Handles cases where the npc dying is not caught by the client.
			if (delayedLootTickLimit < 0)
			{
				clearDelayedLootNpc();
			}
		}
```
This is 3 tick because of delayedLootTickLimit=1, delayedLootTickLimit=0, delayedLootTickLimit=-1

Adam-:
@umer-rs I have squashed the commits and added some comments. Can you review that it still looks ok? (Particularly for Duke and NM)

umer-rs:
Code changes look good. Will test in game tonight just for a sanity check.

umer-rs:
👍 Looks good to me in game. Killed the 3 mobs affected and it works as expected.

PR #17507, opened 2024-02-19
Prevents the roof removal plugin from getting a null value on start

### LINKED ISSUE #17504: Roof Removal NPE on client startup
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

10

### Bug description

1. Enable `Roof Removal` plugin
2. Restart the client
3. Login
4. See Roof Removal isn't working

### Screenshots or videos

RuneLite_2024-02-17_16-59-22

### RuneLite version

image

### Logs

```
2024-02-17 17:09:37 EST [Client] ERROR n.r.client.callback.ClientThread - Exception in invoke
java.lang.NullPointerException: Cannot invoke "net.runelite.api.Scene.setRoofRemovalMode(int)" because the return value of "net.runelite.api.Client.getScene()" is null
        at net.runelite.client.plugins.roofremoval.RoofRemovalPlugin.lambda$startUp$0(RoofRemovalPlugin.java:110)
        at net.runelite.client.callback.ClientThread.lambda$invoke$0(ClientThread.java:49)
        at net.runelite.client.callback.ClientThread.invokeList(ClientThread.java:119)
        at net.runelite.client.callback.ClientThread.invoke(ClientThread.java:101)
        at net.runelite.client.callback.Hooks.tick(Hooks.java:218)
        at client.mz(client.java:27319)
        at client.by(client.java)
        at bm.ap(bm.java:400)
        at bm.ka(bm.java)
        at bm.run(bm.java:46151)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

R-Dson:
Moved getScene() inside of the if-case on start to make sure it's not null and making sure the value is set first, before we call performRoofRemoval().
Closes #17504

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/roofremoval/RoofRemovalPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/roofremoval/RoofRemovalPlugin.java
@@ -106,8 +106,8 @@ public class RoofRemovalPlugin extends Plugin
 			if (client.getGameState() == GameState.LOGGED_IN)
 			{
 				performRoofRemoval();
+				client.getScene().setRoofRemovalMode(buildRoofRemovalFlags());
 			}
-			client.getScene().setRoofRemovalMode(buildRoofRemovalFlags());
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


Adam-:
I think that this would get rid of the exception, however, roof removal would still not work on startup as described in #17504 since the roof removal mode is not being applied. Probably this should be moved to performRoofRemoval() or so.

Adam-:
fixed in fc54812af179308909c0b6b15ca6f7e08d422851

PR #17519, opened 2024-02-22
screenshot: Add option to disable death screenshots in LMS

cdfisher:
Adds an option to ignore deaths in Last Man Standing even when the "Screenshot deaths" option is enabled.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
@@ -180,10 +180,22 @@ public interface ScreenshotConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "ignoreLMSDeaths",
+		name = "Ignore deaths in Last Man Standing",
+		description = "Configures whether or not the Screenshot Deaths option ignores deaths in Last Man Standing.",
+		position = 11,
+		section = whatSection
+	)
+	default boolean ignoreLMSDeaths()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "friendDeath",
 		name = "Screenshot Friend Deaths",
 		description = "Configures whether or not screenshots are automatically taken when friends or friends chat members die.",
-		position = 11,
+		position = 12,
 		section = whatSection
 	)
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -298,7 +300,11 @@ public void onActorDeath(ActorDeath actorDeath)
 			Player player = (Player) actor;
 			if (player == client.getLocalPlayer() && config.screenshotPlayerDeath())
 			{
-				takeScreenshot("Death", SD_DEATHS);
+				final int regionID = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
```
Josh-Renwick-99, line 303: Consider an early return here instead, more concise this way

```suggestion
final int regionID = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
if (config.ignoreLMSDeaths() && LAST_MAN_STANDING_REGIONS.contains(regionID)) {
    return;
}

takeScreenshot("Death", SD_DEATHS);
```

runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotConfig.java
```diff
@@ -179,11 +179,23 @@ default boolean screenshotPlayerDeath()
 		return false;
 	}
 
+	@ConfigItem(
+		keyName = "ignoreLMSDeaths",
+		name = "Ignore deaths in Last Man Standing",
```
Josh-Renwick-99, line 184:
```suggestion
		name = "Ignore Last Man Standing Deaths",
```

Nightfirecat:
Discussed this in Discord a bit--we'd like this change, but I think it actually doesn't go far enough. I think all safe death pvp minigames should be included, not just LMS, meaning it should also include castle wars, soul wars, fight pits, clan wars, and Emir's arena.

Nightfirecat:
Somehow I did not notice until now that this PR was specific to death screenshots, this should probably also add a toggle for pvp kill screenshots as those get similarly noisy from safe pvp minigame kills.

PR #17528, opened 2024-02-26
scriptId: add NOTIFICATION_DISPLAY_INIT

Antimated:
Adds the NOTIFICATION_DISPLAY_INIT ScriptID.

Referencing the following cs2 file.
https://github.com/Joshua-F/cs2-scripts/blob/master/scripts/%5Bclientscript,notification_display_init%5D.cs2


Nightfirecat:
We don't add ID definitions to the api unless they're used in the core client, as we otherwise cannot verify their accuracy.

PR #17529, opened 2024-02-26
clientui: apply custom borders when maximizing

### LINKED ISSUE #17499: Grey line above taskbar
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Dear,

Could you check, I've got no idea if this is a setting or a bug, it started with the last update to RL:

After startup I get this weird line in the bottom of my client, now usually it wouldn't bother me but I always get scared
my screen is broken because it pixels my chat sometimes. (Line can be found on the first screenshot link 1st image.)
Now, I can fix this by minimizing the client to taskbar & then clicking it open again, the line dissapears.
(1st screenshot 2nd image)

https://i.imgur.com/PXu1tIR.png

I also have a test of both when testing in game chat, you can clearly see the on 2nd screenshot link (1st image) there's something wrong with the "e" in this case, it's with all letters all the time if I don't apply the fix. (2nd image is after fix)

https://i.imgur.com/a5l1DJB.png

### Screenshots or videos

_No response_

### RuneLite version

RL version: 1.10.22
Launcher version: 2.5.0

### Logs

_No response_

aHooder:
Before the window is made visible, window state change events aren't guaranteed to be fired. This causes RuneLite's custom window chrome borders to not be removed when restoring a maximized window on client startup. It's fixed by manually reapplying borders when restoring the maximized window state for the "Remember client position" setting.

The state change event not being guaranteed when the window isn't visible is mentioned here: https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/Frame.html#setExtendedState(int)

~~I couldn't find any open issues about this.~~

Closes #17499

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -656,6 +656,10 @@ public class ClientUI
 				if (configManager.getConfiguration(CONFIG_GROUP, CONFIG_CLIENT_MAXIMIZED) != null)
 				{
 					frame.setExtendedState(JFrame.MAXIMIZED_BOTH);
+					// According to the documentation of JFrame#setExtendedState, if the frame isn't visible, a window
+					// state change event isn't guaranteed to be fired. Since RuneLite's custom chrome borders rely on a
+					// state change listener, borders need to be applied manually when maximizing prior to setVisible
+					applyCustomChromeBorder();
 				}
 			}
 			else


aHooder:
Here's a clip of the issue, first on runelite master, then with this PR. Thanks to @Jin-Jiyunsun 

https://github.com/runelite/runelite/assets/831317/2a197340-0934-4a5e-845d-4a43cce06cb5

aHooder:
Closes https://github.com/runelite/runelite/issues/17499

DevilTechniques:
How do we apply fix for this issue? doesn't seem to be working for me yet?

Felanbird:
> How do we apply fix for this issue? doesn't seem to be working for me yet?

It will be fixed when the next version of RuneLite comes out

DevilTechniques:
> > How do we apply fix for this issue? doesn't seem to be working for me yet?
> 
> It will be fixed when the next version of RuneLite comes out

Hey thank you for the update, how often do RL updates happen? Just FMI.

PR #17549, opened 2024-03-03
Add configurable colors for Forestry events.

jsqw:
### Current Behavior
- The Woodcutting plugin primarily uses a bright green color for most events.

### Proposed Changes
- Introduce an option for users to configure their own colors for various Forestry events
- This feature enhances user experience by allowing personalization and supports better accessibility

### Example
Below is an example of how the color customization interface looks within the plugin settings

Example of Color Customization in Woodcutting Plugin


Rubix147:
Yes please, good for us colorblind users.

DolfZiggler:
I think this is a really smart idea. The colours enhance the playability and promote a much more fulfilling user-end experience.

PR #17550, opened 2024-03-03
Add Mappings for WorldPoints to the World Map WorldPoints

Zoinkwiz:
## Problem

The World Map in the game does not accurately represent all locations in their true location, usually as a means of grouping locations which are physically (in terms of WorldPoint) far away, or providing a view of multiple planes at once.

This means that if you make use of the `WorldMapOverlay.mapWorldPointToGraphicsPoint`, it will provide the WorldPoint that the location is represented on the WorldMap, but not the actual WorldPoint of the location.

This is problematic for a few reasons:

1. Certain WorldPoints between locations can overlap with one another on the World Map. This is notable for the Kalphite Lair's shortcut, which is positioned to overlap with the true position of the Chasm of Tears.

  Screenshot 2024-03-02 191620

Without a means of determining if a location is in the current map view, it's not possible to address this issue.

2. The World Map Dev Tools can be useful for quickly checking the WorldPoint of a location, but currently it can mislead by stating the incorrect location. 
3. For locations which are at a different WorldPoint on the World Map to their real position, you need to define two separate WorldPoints. In my experience in the Quest Helper, this is an often overlooked or misunderstood step, which adds confusion and results in helpers which are less helpful than they should be.

## Changes made

Due to these points, I've created this PR to create a mapping of locations which are shifted for the World Map, and make use of this mapping when rendering in the core World Map Plugin and Dev Tools Plugin.

in the Dev Tools Plugin, this mapping has two changes in terms of the end user:
1. The World Point shown in the bottom-left should map to the real location, including plane, so long as the focused point is a walkable spot and not the black void.
2. There is a cyan square added around the central point of the map. Its purpose is to mark the 'true' WorldPoint of the focused point on the map. This means you get scenarios like the Screenshot below, where the cyan square is placed in floor 0 of the Ancient Caverns, with the map focus being on floor 1.

Screenshot 2024-03-02 194917

For the World Map Plugin, it should allow for any WorldPoints which are re-defined as their true WorldPoint to now correctly determine the WorldPoint mapping, which area they are in, and map to their World Map position only for their area. For example, after this change the Kalphite Dungeon shortcut no longer appears in the Chasm of Tears after being set to its true location, but continues to map correctly for the Kalphite Dungeon.

Screenshot 2024-03-02 191646

## Future changes / still needed

In order to ensure this works correctly going forwards, existing WorldPoints defined so as to map correctly to the World Map should be re-defined to be their true position. I've adjusted a token number of these occurences, but it should be done on a widespread basis.

The actual mappings have at present been defined by myself through a mixture of the [WorldMapDumperTest](https://github.com/runelite/runelite/blob/master/cache/src/test/java/net/runelite/cache/WorldMapDumperTest.java), as well as manually configuring each area not included in this. Ideally, there should be a script which can go through each Archive, extract out the chunks and limitations of each mapping, and then generate the mapping file itself. I imagine this could be run in addition to all the ItemID, ObjectID, and so on objects so that this wouldn't need future adjustment.

## Considerations

Currently this makes the assumption that this should be the default behaviour for WorldMapLocationOverlay, however it would be a change in the prior behavior, which might result in some WorldPoints defined to work on the World Map now getting mapped incorrectly. Would this be best made as an option on WorldMapPoint which needs to be enabled actively?

If there are other things that'd need doing (such as fully combing through and changing all of the existing usages of the WorldMapPoint), let me know and I can try to do it!


Zoinkwiz:
I've made use of the cache directly for generating with https://gist.github.com/Zoinkwiz/8b364e4c79cd269ae4073919cff0ba9b, and adjusted the layout and usage to be more similar to how its defined in the cache rather than the shifting method that was used previously.

PR #17557, opened 2024-03-07
sbs timer disable functionality

sonnypb:


Nightfirecat:
Added in 1bc367f6daaced4c01f343b179c291fcc7d117ef, alongside a test to ensure we don't miss this check for other timers on ConfigChanged.

PR #17565, opened 2024-03-09
Party: Add option to automatically join the previous party on startup of the RuneLite client

Niels-Evenblij:
Add an option to the Party plugin to automatically join the previous party on startup of the RuneLite client


Adam-:
We don't want to have an auto-join party feature because there is limited capacity on the party server and don't want a bunch of people unknowingly sitting in parties that they've forgotten about.

Issue #17575, opened 2024-03-13
MacOS: Game doesn't render at native resolution


khogeland:
### Type

GPU

### Operating System

MacOS

### OS Version

Sonoma

### Bug description

The 3D rendering does not use the native screen resolution and instead depends on your system UI scaling, resulting in a much lower quality game view than is possible. These screenshots are taken with the GPU and Stretched Mode plugins enabled.

How it should look (this is with my system resolution set to native):

![Screenshot 2024-03-13 at 1 54 55 PM](https://github.com/runelite/runelite/assets/1495099/26840c2b-8142-4a87-813d-2bcad55a83e6)

How it actually looks (with the default high-DPI system resolution):

<img width="933" alt="Screenshot 2024-03-13 at 1 55 34 PM" src="https://github.com/runelite/runelite/assets/1495099/0ea2ace9-9dfe-4106-b803-0e457d533d03">

### RuneLite version

2.6.13

Adam-:
added in 6dcbb05a40a937cfd864cb2efbb70b3a01395803

PR #17577, opened 2024-03-14
Fix MacOS retina support

khogeland:
Turns the viewport scaling back on for MacOS (once the rlawt fix is in)

Depends on https://github.com/runelite/rlawt/pull/2
Fixes https://github.com/runelite/runelite/issues/17575


Adam-:
How does this affect the screenshot code in Hooks.java:475 (this only runs with GPU off, so maybe nothing?) or ImageCapture.java:116 (this runs regardless I think)

khogeland:
> How does this affect the screenshot code in Hooks.java:475 (this only runs with GPU off, so maybe nothing?) or ImageCapture.java:116 (this runs regardless I think)

I don't think this affects anything that doesn't operate on the AWT/GL context. Screenshots are all high quality and well-aligned with or without the GPU plugin.

khogeland:
Hold on, this still isn't rendering right.

Adam-:
Please don't close and reopen PRs but instead just update this existing PR. If you want to mark it as a draft to show it isn't ready that is fine.

khogeland:
Oh, ok, I see what happened! Maven overwrite my monkey-patched rlawt jar so I was looking at the old behavior, that confused me. Yes, specifically changing the resolution to one _with a different scaling factor_ causes it to misbehave, but it fixes itself when toggling the GPU plugin.

Adam-:
added in 6dcbb05a40a937cfd864cb2efbb70b3a01395803

Issue #17586, opened 2024-03-20
Varlamore meta issue


Adam-:
- [x] new farming patches (14b058400787847103e744c925af7a260a79649f)
- [x] new fairy rings: ajp (#17587)
- [x] new clues (#17590)
- [x] gpu hillskew textures are broken (cdfc401ab6df5d2f97c84f86a896fa1394626133)
- [x] roof removal at hunters guild and bazaar is broken f5c04bf327ea7c72ac155d2e3342884670e23f69
- [x] new hiscores (f533dc70182a2e71c7b641fa25dfd8388d287a2a)
- [x] loottracker (#17593)
    - [x] lunar chest
    - [x] fortis colosseum
    - [x] loot sacks
- [x] new login screen (#17605)
- [x] new "doom" hitsplat type (#17620)
- [x] item stats for new consumables (#17607)
- [x] weapon categories changed? attack style plugin is broken? ffe16454ba894d673db91e2d1fd8d3c4c265b2c7
- [x] several new hunter traps (#17604)
- [x] discord plugin regions (6b793fc0e5095c7f63eeeda4ae091d691e23c7b3)

Adam-:
```
{{Clue info
|id = 967
|text = The cryptic reveals where to search:<br><br>Search the bookshelf with a skull on it inside the Burrow.
|notes = Search the bookshelf near in the Hunter Guild.
|map = {{Map|1569,9460|mtype=pin|plane=0|mapID=-1}}
}
```

```
{{Clue info
|id = 972
|text = The emote reveals <br><br>Yawn in the Fortis Grand Museum.<br>Equip an emerald necklace, blue skirt, and turqoise gnome robe top.
|items = Equip an emerald necklace, blue skirt, and turqoise gnome robe top.
|notes = Perform the [[yawn]] emote near inside the Fortis Grand Museum
|map = {{Map|1712,3163|mtype=pin|plane=0|mapID=-1}}
```

```
{{Clue info
|id = 969
|text = The coordinate reveals where to dig:<br><br>0.42N 27.35W
|notes = Dig north of Ortus Farm
The clue has an additional challenge:

Kill an [[Armadylean guard]].
|map = {{Map|1557,3183|mtype=pin|plane=0|mapID=-1}}
}}
```

```
{{Clue info
|id = 365
|text = The emote reveals <br><br>Bow within the temple in Civitas illa Fortis.<br>Equip the full set of Sunfire Fanatic armour.
[...]

PR #17587, opened 2024-03-20
worldmap: add AJP fairy ring

Felanbird:
✔️ on #17586 
image


Adam-:
this needs to be added to the fairy ring plugin too

PR #17590, opened 2024-03-20
clues: add Valmore clues

Felanbird:
Adds all of the clues currently crowdsourced from the wiki

notes:
~~As of right now, the easy clue `Search the Bookshelf with a skull on it inside the Burrow.` does NOT highlight the bookshelf, as it is a decoration instead of a game object. That being said it does put the arrow in front of the correct bookshelf.~~ probably fixed by NFC by the time this is merged

~~the WorldPoint for `SOUTH_EAST_OF_THE_BAZAAR` is just made up because I can't find the STASH unit~~ thanks gamma
~~don't know what the new emote SpriteID is - or which doubleAgent it is~~
~~don't know what the correct enemies are for the new HotCold tasks~~


Adam-:
link to #17586

Issue #17597, opened 2024-03-21
Fishing plugin overlay false positive


Macweese:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10.0.19045.4170

### Bug description

Catching butterflies both barehanded and with a butterfly net causes the fishing plugin overlay to be drawn.

Triggered by catching Ruby harvest, Snowy knight, Black warlock.
Does **not** occur when catching from Sunlight Moth, Moonlight Moth.

Caused by similar chatmessages.

### Screenshots or videos

https://github.com/runelite/runelite/assets/50101641/4b851044-20f7-4bb7-bbdb-1b6147afe269

### RuneLite version

RuneLite version 1.10.25
Launcher version 2.6.1
OSRS rev 220

PR #17598, opened 2024-03-21
fishing: prevent false positive session

Macweese:
Adding a space after 'a' prevents chat messages when catching butterflies from passing the check. Butterfly message: "You catch and..."
Fishing message: "You catch a ..."
Fixes https://github.com/runelite/runelite/issues/17597

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
@@ -33,6 +33,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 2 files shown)


Felanbird:
image

Felanbird:
np 👍 both of them were a collective effort in the discord I just am the one not too lazy to comment

runelite-client/src/main/java/net/runelite/client/plugins/fishing/FishingPlugin.java
```diff
@@ -86,6 +87,8 @@ public class FishingPlugin extends Plugin
 
 	private Instant trawlerStartTime;
 
+	private static final Pattern FISHING_CATCH_REGEX = Pattern.compile("(You catch (a|an) )|(You catch some)|(Your cormorant returns with its catch.)|(You catch .* Karambwanji)");
```
Adam-, line 90: this is just `You catch (a|an|some) |Your cormorant returns with its catch.|You catch .* Karambwanji`. You don't need to add unnecessary capture groups for everything.

PR #17603, opened 2024-03-21
widget: make moons of peril overlay movable

MESLewis:
supersedes #17599 without depreciated changes and with more accurate name


Adam-:
ref #17586

PR #17607, opened 2024-03-22
item stats: add varlamore consumables

Macweese:
This PR adds new consumables, from the Varlamore update, to the item stats plugin.
#17586 

Consumables added:
- Cooked Bream
- Cooked Moss Lizard
- Eclipse wine
- Moon-lite
- Moonlight Moth / Mix
- Moonlight Potion
- Sunbeam ale
- Sunlight Moth / Mix
- Sun-shine
- Trapper's Tipple

I tried my best to check the effects, though since some depend on players' certain skill level it's hard to test it thoroughly. Data was also used from the OSRS Wiki for the consumables' effect calculations.

~~Update:
The formula for moonlight potion is not accurate for defence skill; converting to draft.~~

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -35,6 +35,7 @@ import static net.runelite.client.plugins.itemstats.Builders.*;
 import net.runelite.client.plugins.itemstats.delta.DeltaPercentage;
 import net.runelite.client.plugins.itemstats.food.Anglerfish;
 import net.runelite.client.plugins.itemstats.food.CookedBream;
+import net.runelite.client.plugins.itemstats.food.CookedMossLizard;
 import net.runelite.client.plugins.itemstats.potions.Ambrosia;
 import net.runelite.client.plugins.itemstats.potions.AncientBrew;
 import net.runelite.client.plugins.itemstats.potions.MixedPotion;
@@ -115,6 +116,7 @@ public class ItemStatChanges
 		add(food(maxHP -> (int) Math.ceil(maxHP * .05)), WATERMELON_SLICE);
 		add(food(perc(.1, 1)), COOKED_SWEETCORN, SWEETCORN_7088 /* Bowl of cooked sweetcorn */);
 		add(new CookedBream(), COOKED_BREAM);
+		add(new CookedMossLizard(), COOKED_MOSS_LIZARD);
 		add(combo(food(1), boost(DEFENCE, perc(.02, 1))), CABBAGE_1967 /* Draynor Manor */);
 		add(combo(food(8), heal(RUN_ENERGY, 5)), PAPAYA_FRUIT);
 		add(combo(food(3), boost(ATTACK, perc(.02, 2))), CUP_OF_TEA_1978 /* Standard tea */);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -34,6 +34,7 @@ import static net.runelite.api.ItemID.*;
 import static net.runelite.client.plugins.itemstats.Builders.*;
 import net.runelite.client.plugins.itemstats.delta.DeltaPercentage;
 import net.runelite.client.plugins.itemstats.food.Anglerfish;
+import net.runelite.client.plugins.itemstats.food.CookedBream;
 import net.runelite.client.plugins.itemstats.potions.Ambrosia;
 import net.runelite.client.plugins.itemstats.potions.AncientBrew;
 import net.runelite.client.plugins.itemstats.potions.MixedPotion;
@@ -113,6 +114,7 @@ public class ItemStatChanges
 		add(food(maxHP -> (int) Math.ceil(maxHP * .06)), STRAWBERRY);
 		add(food(maxHP -> (int) Math.ceil(maxHP * .05)), WATERMELON_SLICE);
 		add(food(perc(.1, 1)), COOKED_SWEETCORN, SWEETCORN_7088 /* Bowl of cooked sweetcorn */);
+		add(new CookedBream(), COOKED_BREAM);
 		add(combo(food(1), boost(DEFENCE, perc(.02, 1))), CABBAGE_1967 /* Draynor Manor */);
 		add(combo(food(8), heal(RUN_ENERGY, 5)), PAPAYA_FRUIT);
 		add(combo(food(3), boost(ATTACK, perc(.02, 2))), CUP_OF_TEA_1978 /* Standard tea */);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -37,6 +37,7 @@ import net.runelite.client.plugins.itemstats.food.Anglerfish;
 import net.runelite.client.plugins.itemstats.potions.Ambrosia;
 import net.runelite.client.plugins.itemstats.potions.AncientBrew;
 import net.runelite.client.plugins.itemstats.potions.MixedPotion;
+import net.runelite.client.plugins.itemstats.potions.MoonlightPotion;
 import net.runelite.client.plugins.itemstats.potions.PrayerPotion;
 import net.runelite.client.plugins.itemstats.potions.SaradominBrew;
 import net.runelite.client.plugins.itemstats.potions.StaminaPotion;
@@ -55,6 +56,13 @@ public class ItemStatChanges
 	{
 		init();
 	}
+
+	public static final SingleEffect ATTACK_POT = boost(ATTACK, perc(.10, 3));
+	public static final SingleEffect STRENGTH_POT = boost(STRENGTH, perc(.10, 3));
+	public static final SingleEffect DEFENCE_POT = boost(DEFENCE, perc(.10, 3));
+	public static final SingleEffect SUPER_ATTACK_POT = boost(ATTACK, perc(.15, 5));
+	public static final SingleEffect SUPER_STRENGTH_POT = boost(STRENGTH, perc(.15, 5));
+	public static final SingleEffect SUPER_DEFENCE_POT = boost(DEFENCE, perc(.15, 5));
 
 	private void init()
 	{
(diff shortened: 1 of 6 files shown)


DominickCobb-rs:
I've been pouring over the Moonlight potion effects all week and updating the wiki regularly. Pulling this from the code I had written prior to your PR, I've amended it to include the correct formula for the prayer restoration. Here is the code:

```@Override
{
	int herblore = HERBLORE.getValue(client);
	int prayer = PRAYER.getMaximum(client);
	return (int) Math.max(prayer*.25,herblore*.3)+7;
}```

**Comment by DominickCobb-rs** [comment]
But also yes if the herblore level drops below 38 the prayer restored is 0, I forgot that part.

File: runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/MoonlightPotion.java [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+						attackBoost.setStat(stat);
+						return attackBoost.effect(client);
+					}
+					if (stat == STRENGTH)
+					{
+						strengthBoost.setStat(stat);
+						return strengthBoost.effect(client);
+					}
+					return null;
+				}))
```
Nightfirecat, line 93: Why are the prayer restore, defense boost, and attack/strength boosts all assembled in completely different ways here? `restoreChange` is not reused, so the variable can be inlined. Both the attack and strength boosts already have stats set because they're initialized as `SimpleStatBoost`s, so the `setStat()` calls are no-ops. `OFFENSIVE_MELEE_STATS` is similarly not reused, so I'm not really sure what purpose it serves here.

Anyway, the only potential purpose I see to using a Stream here is to exclude the prayer restore if the player's herblore level is less than 38, but even then it's probably fine to list "+0 Prayer" for this boost, in which case this whole stream-to-array process can be replaced by a simple `StatChange[]` var declaration.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/MoonlightPotion.java
```diff
@@ -0,0 +1,138 @@
...
+	STRENGTH_POTION(STRENGTH, 12, perc(.10, 3)),
+	SUPER_ATTACK_POTION(ATTACK, 45, perc(.15, 5)),
+	SUPER_DEFENCE_POTION(DEFENCE, 66, perc(.15, 5)),
+	SUPER_STRENGTH_POTION(STRENGTH, 55, perc(.15, 5)),
+	NONE(null, 0, max -> 0),
+	;
+
+	private final Stat stat;
+	private final int levelRequirement;
+	private final DeltaCalculator deltaCalculator;
```
Nightfirecat, line 130: Most of these effects are already defined in ItemStatChanges. (some/all are even defined as variables for reuse in mixed potion definitions) If possible, I'd like to make those definitions static variables which can be referenced here instead of being duplicated.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/MoonlightPotion.java
```diff
@@ -0,0 +1,138 @@
...
+			.max(Comparator.naturalOrder()).orElseThrow());
+		return changes;
+	}
+
+	private DeltaCalculator getMeleeEffect(Client client, Stat stat)
+	{
+		return Arrays.stream(MeleeEffect.values())
+			.filter(t -> t.getStat() == stat && client.getBoostedSkillLevel(Skill.HERBLORE) >= t.getLevelRequirement())
+			.max(Comparator.comparingInt(MeleeEffect::getLevelRequirement))
+			.orElseGet(() -> MeleeEffect.NONE).getDeltaCalculator();
```
Nightfirecat, line 109: `.orElseGet(() -> MeleeEffect.NONE)` can be `.orElse(MeleeEffect.NONE)` instead.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -352,6 +365,19 @@ private void init()
...
 		add(combo(boost(ATTACK, perc(.15, 5)), boost(STRENGTH, perc(.15, 5)), boost(DEFENCE, perc(.15, 5)), boost(RANGED, perc(.15, 5)), boost(MAGIC, perc(.15, 5)), heal(PRAYER, perc(.25, 8))), POTION_OF_POWER1, POTION_OF_POWER2, POTION_OF_POWER3, POTION_OF_POWER4);
 
+		// Moths & Butterflies
+		final Effect rubyHarvest = boost(ATTACK, perc(.15, 4));
+		final Effect snowyKnight = heal(HITPOINTS, perc(.15, 4));
+		final Effect blackWarlock = boost(STRENGTH, perc(.15, 4));
+		final Effect sapphireGlacial = boost(DEFENCE, perc(.15, 4));
+		final SunlightMoth sunlightMoth = new SunlightMoth(.2, 6);
+		add(rubyHarvest, RUBY_HARVEST, RUBY_HARVEST_MIX_1, RUBY_HARVEST_MIX_2);
+		add(snowyKnight, SNOWY_KNIGHT, SNOWY_KNIGHT_MIX_1, SNOWY_KNIGHT_MIX_2);
```
Nightfirecat, line 375: Multi-bite/use food items (such as cakes and pies) should list the amount each use heals, not the total. This would incorrectly list a 1-dose Snowy knight mix as healing twice as much as it does.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
```diff
@@ -352,6 +365,19 @@ private void init()
 		add(combo(heal(HITPOINTS, perc(.15, 1)), heal(RUN_ENERGY, 100)), BANDAGES_25202);
 		add(combo(boost(ATTACK, perc(.15, 5)), boost(STRENGTH, perc(.15, 5)), boost(DEFENCE, perc(.15, 5)), boost(RANGED, perc(.15, 5)), boost(MAGIC, perc(.15, 5)), heal(PRAYER, perc(.25, 8))), POTION_OF_POWER1, POTION_OF_POWER2, POTION_OF_POWER3, POTION_OF_POWER4);
 
+		// Moths & Butterflies
+		final Effect rubyHarvest = boost(ATTACK, perc(.15, 4));
+		final Effect snowyKnight = heal(HITPOINTS, perc(.15, 4));
```
Nightfirecat, line 370: The wiki seems to have a lot of conflicting information about how much this actually heals. Has this been tested?

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/special/SunlightMoth.java
```diff
@@ -0,0 +1,81 @@
...
+	{
+		StatsChanges changes = new StatsChanges(0);
+
+		SimpleStatBoost calcRestore = new SimpleStatBoost(null, false, perc(percentRestored, delta));
+		SimpleStatBoost calcHeal = new SimpleStatBoost(HITPOINTS, false, perc(0, 8));
+
+		changes.setStatChanges(Stream.concat(
+			Stream.of(new Stat[]{HITPOINTS})
+				.filter(stat -> stat.getValue(client) < stat.getMaximum(client))
+				.map(stat -> calcHeal.effect(client)),
```
Nightfirecat, line 67: Generally we do not omit hitpoints healing when a player is at or beyond the amount which can be healed by food items; let's not filter this out even if it would heal 0.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
capslock13: There was a hotfix that changed it shortly after release of the mixes which changed it which could leave to mixed info. Just tested: Snowy Knight Mix restores 8 and Snowy Knight restores 15. I believe this is a flat amount not percent but my alt doesn't have members currently to test so could only verify on an account with 99 hp

DominickCobb-rs: @capslock13 you can currently test by changing your stats on beta worlds.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/food/CookedMossLizard.java
```diff
@@ -0,0 +1,41 @@
...
+import static net.runelite.client.plugins.itemstats.stats.Stats.COOKING;
+import static net.runelite.client.plugins.itemstats.stats.Stats.HUNTER;
+
+public class CookedMossLizard extends FoodBase
+{
+	@Override
+	public int heals(Client client)
+	{
+		int cooking = COOKING.getValue(client) / 3;
+		int hunter = HUNTER.getValue(client) / 2;
```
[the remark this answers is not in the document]
capslock13: Tested by bringing spicy stews in. Healed 31 at 93/93 cooking 74/74 hunter; after boosting to 98/93 and 77/74 the lizard healed 32. However I think the formula for amount healed is wrong and further testing is needed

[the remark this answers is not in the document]
c13-c: Looking at it again it seems that the only values it was wrong for (e.g. 1, 1 -> 7 healed) are below the level required to cook/catch the lizards and shouldn’t matter.

YvesW:
The wiki people, mostly @``depressiontm``, have done more testing on the beta worlds! The formulas seem to be the following:
- Cooked moss lizard
  ```floor(min(boosted cooking level / 3, boosted hunter level / 2))```
- Cooked bream
  ```max(7, floor(min(boosted cooking level / 3, boosted fishing level / 3)))```

tylerwgrass:
> The new hunter foods which have split healing (on consumption & delayed heal and any additional effects), how should they be added as a food? The full expected heal, or only the initial, or is a new special implementation wanted for this?

Right now these foods show up like this using 50hp as an example:

```
+12 (62)
+ 8 (58)
```

This is a bit misleading and I would find something like this more accurate:
```
+12 (62)
+ 8 (70)
```

which would require a new implementation. 

The ToA items which have a similar effect over a longer duration instead show what the heal amount is per proc, and then the total that is healed.

Config for the silk dressing:
```
add(combo(heal(HITPOINTS, 5), heal(HITPOINTS, 5 * 20)), SILK_DRESSING_1, SILK_DRESSING_2);
```

burkeg:
What's still left on this PR? Would we be willing to merge a partial version of this with the items that are confirmed accurate and add the rest in a separate follow-up?

burkeg:
> What's still left on this PR? Would we be willing to merge a partial version of this with the items that are confirmed accurate and add the rest in a separate follow-up?

@Nightfirecat @Macweese any comment?

Nightfirecat:
I've not yet taken the time to re-review this both due to personal time commitments outside of the project, and due to other project work having taken more of my time and attention. I do plan to give this another proper look over sometime soon.

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/MoonlightPotion.java
```diff
@@ -0,0 +1,109 @@
...
+}
+
+@Getter
+enum StatEffect
+{
+	ATTACK_POTION(ATTACK, 3, ItemStatChanges.ATTACK_POT),
+	DEFENCE_POTION(DEFENCE, 30, ItemStatChanges.DEFENCE_POT),
+	// Misnomer: Moonlight potion gives the effect of smelling salts on defence but requires the level for,
+	// and activates the var/effect timer for—albeit without the drain prevention from—Divine Super Defence potion
+	DIVINE_SUPER_DEFENCE_POTION(DEFENCE, 70, new SimpleStatBoost(DEFENCE, true, ItemStatChanges.SALT_BOOST)),
```
Nightfirecat, line 82: Is this correct, or is the wiki correct? The wiki cites the 70+ prayer defence boost as `level * 1.2 + 7`

runelite-client/src/test/java/net/runelite/client/plugins/itemstats/ItemStatEffectTest.java
```diff
@@ -247,6 +247,176 @@ public void testAmbrosia()
...
+		assertEquals(26, skillChange(Skill.PRAYER, 42, 0, moonlightPotion));
+		assertEquals(4, skillChange(Skill.PRAYER, 42, 38, moonlightPotion));
+
+		when(client.getBoostedSkillLevel(Skill.HERBLORE)).thenReturn(77);
+		assertEquals(30, skillChange(Skill.PRAYER, 77, 18, moonlightPotion));
+
+		when(client.getBoostedSkillLevel(Skill.HERBLORE)).thenReturn(80);
+		assertEquals(6, skillChange(Skill.ATTACK, 80, 91, moonlightPotion));
+		assertEquals(7, skillChange(Skill.STRENGTH, 92, 103, moonlightPotion));
+		assertEquals(9, skillChange(Skill.DEFENCE, 75, 89, moonlightPotion));
```
Nightfirecat, line 287: Was this a real difference observed, or is this created to fit the formula given by the moonlight potion implementation? (see comment there)

Nightfirecat (changes requested):
All LGTM besides these couple of comments

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/MoonlightPotion.java
coopermor: I'll confirm the wiki is correct. At 71 herblore and 70 defence I get a +21 defence boost which match the formula `level * 1.2 + 7`

Issue #17600, opened 2024-03-21
Implings plugin: "Show Spawn locations" option does not work for Essence Implings


Jinxer42:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Go to poro poro and with highlight Essence implings turned on and spawn locations toggled on. The plugin appears to work for all other imlpings correctly except for Essence imps.

### Screenshots or videos

![image](https://github.com/runelite/runelite/assets/25554393/453fd0a4-c9de-495a-af33-2e5902175994)

### RuneLite version

RuneLite version: 1.10.25
Launcher Version: 2.6.9

### Logs

[launcher_2024-03-20.0.log](https://github.com/runelite/runelite/files/14694577/launcher_2024-03-20.0.log)
[client_2024-03-20.0.log](https://github.com/runelite/runelite/files/14694578/client_2024-03-20.0.log)

Adam-:
Do essence implings have spawns? The plugin seems to think they do not. If they do, and you could collect the coordinates for them, that would be helpful.

Jinxer42:
> Do essence implings have spawns? The plugin seems to think they do not. If they do, and you could collect the coordinates for them, that would be helpful.

According to the wiki they do have set spawns (5 of them).  Some rough testing confirms this.  I tested 1 spot for 10 spawns and 7 of the 10 occurred on the exact same tile, while three occurred in tiles to the north (see image below).  This is comparable to rates i have experienced with eclectic imp spawns (although I never actually took data on those).

You will have to forgive me. But how do I get the coordinates for the tiles, so I can send them to you?

![image](https://github.com/runelite/runelite/assets/25554393/7688ab48-b8a6-42c7-ad58-3de7dba222dd)

mario026:
Added the 5 spawn points for the essence impling

https://github.com/runelite/runelite/pull/17613

Jinxer42:
> Added the 5 spawn points for the essence impling
> 
> #17613

Thanks much! I look forward to it.

PR #17613, opened 2024-03-23
Implings plugin: add 5 spawns for the essence impling

mario026:
Updated the impling spawn with the 5 spawns for the essence impling.

the essence impling with ID: 1639 has a random spawn location.

the essence impling with the ID: 1649 is connected to the 5 spawn points

fixes: https://github.com/runelite/runelite/issues/17600

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingSpawn.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/implings/ImplingSpawn.java
@@ -82,7 +82,14 @@ enum ImplingSpawn
 	SPAWN_ECLECTIC1(new WorldPoint(2567, 4319, 0), ImplingType.ECLECTIC),
 	SPAWN_ECLECTIC2(new WorldPoint(2591, 4340, 0), ImplingType.ECLECTIC),
 	SPAWN_ECLECTIC3(new WorldPoint(2591, 4295, 0), ImplingType.ECLECTIC),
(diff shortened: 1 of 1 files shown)


Jinxer42:
I think 4 of the 5 spawns for Essence imps are labeled incorrectly (with some closer than others).  I have tested each location 10 times and only 1 of the locations ever had an essence imp spawning at the label position (see image below for an example). I can upload screen shots of all the locations if that would be helpful.  Let me know if there is anything else I can do.

image

Adam-:
Ok, I'll revert it until someone is able to figure out where they spawn at.

mario026:
Hee @Adam-. I thought it tested it well. I see that there is a other PR open for this issue. However i still had a question regarding the labels. Because the implings didnt spawn 100% at the same spot. How does the label renderen know which one to render if i add 5 postions like you showed for 1 impling, so like 25 spots if all implings have 5 spawn location in one area. 

is there something in an event that holds the new spawn location for the impling that was just catched? and if that matches with one of the coded location the label will be rendered?

Nightfirecat:
NPCs can walk on the same tick they spawn, meaning they can appear to have spawned on a tile adjacent to where they actually spawn. You can see the NPC indicators plugin handles this case: https://github.com/runelite/runelite/blob/f890298298cdc99923ce1836252c00cb7c8e1426/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java#L761-L774

mario026:
Aah Thanks for the information. I don't think i will try to fix it because of the open pr. if that pr doesnt solve it i'm more then willing to try again. or an other entry level issue

PR #17620, opened 2024-03-24
hitsplat: add doom and burn hitsplats

mo-pi:
Adds Doom and Burn hitsplats introduced in Varlamore: Part One
#17586


Adam-:
Usually new hitsplats have max and me/other variants, are there none for these?

mo-pi:
There only seems to be one hitsplat for both, I tested the burn in game and it shows no me/other variant and no max variant. Doom always hits for a max and can only be viewed on yourself due to it being used only in the colosseum.

PR #17625, opened 2024-03-25
Add support for Calcified rocks to the mining plugin

lezed1:
Add respawn timers to [Calcified rocks](https://oldschool.runescape.wiki/w/Calcified_rocks). I testing the respawn time a few times, and it seems to range from 50 to 60 ticks. I made the timer be 60 ticks as the worst-case scenario and made sure the timers are removed if the rock respawns early.

Here is an example of the respawn timers
Screenshot 2024-03-24 231159

And after two of the rocks have respawned.
Screenshot 2024-03-24 231215


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -355,12 +339,6 @@ public void onWallObjectSpawned(WallObjectSpawned event)
 			case GOLD_VEIN: // Arzinian gold vein
 			case GOLD_VEIN_5990: // Gold vein
 			case GOLD_VEIN_5991: // Gold vein
-			{
```
Adam-, line 358: why is this being removed? isnt it important?

lezed1: Oops, I didn't mean to delete that. I originally added the new object IDs here, but later realized they are wall game objects, not wall objects - and then I deleted more than my new code!

lezed1:
I'm not sure what to do about [this error](https://github.com/runelite/runelite/actions/runs/8414713314/job/23038793276?pr=17625#step:5:3109):
> ```Error:  Failed to execute goal on project client: Could not resolve dependencies for project net.runelite:client:jar:1.10.25.3-SNAPSHOT: Could not find artifact net.runelite:client-patch:jar:1.10.25.3-SNAPSHOT in runelite (https://repo.runelite.net/) -> [Help 1]```

PR #17626, opened 2024-03-25
item identification: add jars and mixes

### LINKED ISSUE #18782: [Item Charges] Moonlight Potion doses not shown
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Win10

### Bug description

When enabling the `Potion doses` setting in the Item Charges plugin, [Moonlight Potions](https://oldschool.runescape.wiki/w/Moonlight_potion) do not show their number of doses in the inventory.

### Screenshots or videos

Image

Image

### RuneLite version

RuneLite version: 1.11.1
Launcher version: 2.7.2

### Logs

[client.log](https://github.com/user-attachments/files/18914400/client.log)
[launcher.log](https://github.com/user-attachments/files/18914401/launcher.log)

Macweese:
This PR adds jars and mixes for butterflies and moths, and potion mixes to both the item identification plugin and the item charges plugin. For the item identification plugin; a separate config option was added for butterfly and moth consumables.

I was not sure if there is a naming convention for the abbreviations which should be followed, so I improvised with the new item names.

Fixes #18782


runelite-client/src/main/java/net/runelite/client/plugins/itemidentification/ItemIdentification.java
```diff
@@ -360,6 +360,58 @@ enum ItemIdentification
 	DWARF_WEED_POTION(Type.POTION, "Dwarf", "D", ItemID.DWARF_WEED_POTION_UNF),
 	TORSTOL_POTION(Type.POTION, "Torstol", "TOR", ItemID.TORSTOL_POTION_UNF),
 
+	// Barbarian Potions
```
Nightfirecat, line 363: Can we interleave barbarian potions in with regular potions so that it's obvious the abbreviations match?

Nightfirecat: Interleaving is inserting something in a regular, alternating form. In this case, the first example (attack, attack mix, strength, strength mix, etc.) is what I had in mind, which matches your most recent push, so that looks good.

runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemWithCharge.java
```diff
@@ -146,6 +162,7 @@ enum ItemWithCharge
 	DEFENCE2(POTION, DEFENCE_POTION2, 2),
 	DEFENCE3(POTION, DEFENCE_POTION3, 3),
 	DEFENCE4(POTION, DEFENCE_POTION4, 4),
+	DEFENCE_MIX_1(POTION, DEFENCE_MIX1, 1),
```
Nightfirecat, line 165: I'm fixing it now, but FYI you missed the 2-dose defence mix.

PR #17630, opened 2024-03-25
Add Tonalztics spec to spec counter plugin

capslock13:
Adds tonalztics to the spec counter plugin, after testing the spec always applies the hitsplat 2t after speccing regardless of distance, which means it will never queue before a mage or range thrall (which will always be at least 2t later and if they're both 2t the npc action is always processed first) so the last hitsplat logic still works, except for melee thralls (which is also true for dorgeshuun crossbow and accursed scepter, which both additionally can be incorrect on mage and range thralls if you bow/cast from max distance)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -70,6 +70,7 @@ import net.runelite.client.party.PartyService;
 import net.runelite.client.party.WSClient;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
+import static net.runelite.client.plugins.specialcounter.SpecialWeapon.TONALZTICS_OF_RALOS;
 import net.runelite.client.ui.overlay.OverlayManager;
 import net.runelite.client.ui.overlay.infobox.InfoBoxManager;
 import net.runelite.client.util.ImageUtil;
@@ -99,7 +100,9 @@ public class SpecialCounterPlugin extends Plugin
 	private int hitsplatTick;
 	// most recent hitsplat and the target it was on
 	private Hitsplat lastSpecHitsplat;
+	private Hitsplat secondToLastSpecHitsplat;
 	private NPC lastSpecTarget;
+	private int specialAttackHits = 0;
 
 	private final Set<Integer> interactedNpcIndexes = new HashSet<>();
 	private final SpecialCounter[] specialCounter = new SpecialCounter[SpecialWeapon.values().length];
@@ -163,6 +166,7 @@ public class SpecialCounterPlugin extends Plugin
 		specialWeapon = null;
 		lastSpecTarget = null;
 		lastSpecHitsplat = null;
+		secondToLastSpecHitsplat = null;
 		removeCounters();
 		overlayManager.remove(playerInfoDropOverlay);
 		wsClient.unregisterMessage(SpecialCounterUpdate.class);
@@ -185,12 +189,25 @@ public class SpecialCounterPlugin extends Plugin
 		{
 			if (lastSpecHitsplat.getAmount() > 0)
 			{
-				specialAttackHit(specialWeapon, lastSpecHitsplat, lastSpecTarget);
+				specialAttackHits++;
+			}
+			if (specialWeapon == TONALZTICS_OF_RALOS && secondToLastSpecHitsplat != null
+				&& secondToLastSpecHitsplat.getAmount() > 0)
+			{
+				specialAttackHits++;
+			}
+
+			if (specialAttackHits > 0)
+			{
+				int hit = specialWeapon == TONALZTICS_OF_RALOS ? specialAttackHits : getHit(specialWeapon, lastSpecHitsplat);
+				specialAttackHit(specialWeapon, hit, lastSpecTarget);
 			}
 
 			specialWeapon = null;
 			lastSpecHitsplat = null;
+			secondToLastSpecHitsplat = null;
 			lastSpecTarget = null;
+			specialAttackHits = 0;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


degen19:
Has this been looked at yet? It is currently used in CMs and this would be valuable information to have presented via the plugin. It will be similarly important after the changes next week.

Nightfirecat:
It hasn't; I (nor the other devs, that I'm aware of) don't have this weapon to test/verify this change.

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -184,13 +187,32 @@ public void onGameTick(GameTick event)
 		if (lastSpecHitsplat != null && specialWeapon != null && lastSpecTarget != null)
 		{
 			if (lastSpecHitsplat.getAmount() > 0)
+			{
```
Adam-, line 190: This logically reduces to

```java
            if (lastSpecHitsplat.getAmount() > 0)
            {
                tonalHitsCount++;
            }
            if (specialWeapon.equals(SpecialWeapon.TONALZTICS_OF_RALOS) && secondToLastSpecHitsplat != null
                && secondToLastSpecHitsplat.getAmount() > 0)
            {
                tonalHitsCount++;
            }

            if (tonalHitsCount > 0)
            {
                specialAttackHit(specialWeapon, lastSpecHitsplat, lastSpecTarget);
            }
```

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -470,6 +493,10 @@ private void removeCounters()
 
 	private int getHit(SpecialWeapon specialWeapon, Hitsplat hitsplat)
 	{
+		if (specialWeapon.equals(SpecialWeapon.TONALZTICS_OF_RALOS))
```
Adam-, line 496: Pull getHit() out of specialAttackHit() instead of hacking around it here.

Adam- (changes requested):
additionally please to try avoid unnecessarily reformatting irrelevant code.

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialWeapon.java
```diff
@@ -54,7 +54,12 @@ public enum SpecialWeapon
 		(distance) -> 46 + distance * 10,
 		(c) -> 0
 	),
-	;
+	TONALZTICS_OF_RALOS(
+		"Tonalztics of Ralos",
+		new int[]{ItemID.TONALZTICS_OF_RALOS},
+		false,
+		(distance) -> 50, //The hitsplat is always applied 2t after spec regardless of distance
+		(c) -> 0);
```
Nightfirecat, line 62: Does the special effect apply twice if each hitsplat hits? The wiki's information on this weapon doesn't seem to clarify this.

Nightfirecat (commented):
I can confirm from ingame testing that the hitsplat apples 2 server ticks after special energy drains. :+1:

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialWeapon.java
c13-c: Yes, it reduces by 10% of the target’s magic level per hitsplat that is > 0

capslock13:
The logic has been reduced to what @Adam- suggested and getHit() has been pulled out of specialAttackHit()

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialWeapon.java
Nightfirecat: (yes, it does)

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -286,13 +302,14 @@ public void onHitsplatApplied(HitsplatApplied hitsplatApplied)
 		// venge or thralls.
 		if (hitsplatTick == client.getTickCount())
 		{
+			secondToLastSpecHitsplat = lastSpecHitsplat;
 			lastSpecHitsplat = hitsplat;
 		}
 	}
 
 	private void specialAttackHit(SpecialWeapon specialWeapon, Hitsplat hitsplat, NPC target)
```
Adam-, line 310: replace the `hitsplat` parameter here with `int hit` and modify the call site to compute the hit

runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
```diff
@@ -184,13 +188,25 @@ public void onGameTick(GameTick event)
 		if (lastSpecHitsplat != null && specialWeapon != null && lastSpecTarget != null)
 		{
 			if (lastSpecHitsplat.getAmount() > 0)
+			{
+				specialAttackHits++;
+			}
+			if (specialWeapon.equals(TONALZTICS_OF_RALOS) && secondToLastSpecHitsplat != null
```
Adam-, line 194:
```suggestion
			if (specialWeapon == TONALZTICS_OF_RALOS && secondToLastSpecHitsplat != null
```

Adam-:
> The logic has been reduced to what @Adam- suggested and getHit() has been pulled out of specialAttackHit()

This is just not true because getHit() is still called from specialAttackHit()

Adam- (changes requested):

capslock13:
Correctly removed getHit() from specialAttackHit this time and changed .equals per @Adam- request. Additionally cleaned up the commits.

PR #17650, opened 2024-03-29
worldmap: fix quest icons paint cutoff

Macweese:
This PR fixes an issue with the quest map icons, in which they are only partially drawn.

The explanation for the numbers 25 and 5 is because the original images are 15x15 pixels, and the new ones 20x20 pixels. Attempting to center these inevatbly causes misalignment. However, with a size of 25x25 and an offset of 5 pixels the icons will align with the original without sacrificing their size. 

It is the smallest possible size.

<br>
<details>
  <summary>Before & After</summary>
  
  image
  
</details>
<br>
Added for future reference
<details>
  <summary>Math</summary>
 $$\newline$$ 

   \$a\doteq\space\text{size image a}\$
   \$b\doteq\space\text{size image b}\$
   \$c\doteq\space\text{required new size}\$

   $$\displaylines{c\cases{\min\limits_{c}f(c|b) \space = |a-b|+\frac{a-b}{2}+\left|\frac{a+b}{2}\right|\space , \space \text{offset}\equiv\left((\frac{a+b}{2}+\left|\frac{a-b}{2}\right|\right)-\left(\frac{a-b}{2}+\left|\frac{a+b}{2}\right|\right))(\text{mod}2) \cr c-(\frac{a+b}{2}-\left|\frac{a-b}{2}\right|)=\frac{\text{offset}}{2}}}$$

</details>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
@@ -107,44 +107,46 @@ public class WorldMapPlugin extends Plugin
 
 	static
 	{
+		// Original size of world map icons
+		final int worldMapIconSize = 15;
 		//A size of 17 gives us a buffer when triggering tooltips
-		final int iconBufferSize = 17;
-
+		final int iconOffset = 1;
+		final int iconBufferSize = worldMapIconSize + iconOffset * 2;
 		// Quest icons are a bit bigger than regular icons
 		// A size of 25 aligns the quest icons when converting the world map point to pixel coordinates
 		// The new quest icons must be offset by 5, for a size of 25, to align when drawing on top of the original icon
-		final int questIconBufferSize = 25;
 		final int questIconOffset = 5;
+		final int questIconBufferSize = worldMapIconSize + questIconOffset * 2;
 
 		BLANK_ICON = new BufferedImage(iconBufferSize, iconBufferSize, BufferedImage.TYPE_INT_ARGB);
 
 		FAIRY_TRAVEL_ICON = new BufferedImage(iconBufferSize, iconBufferSize, BufferedImage.TYPE_INT_ARGB);
 		final BufferedImage fairyTravelIcon = ImageUtil.loadImageResource(WorldMapPlugin.class, "fairy_ring_travel.png");
-		FAIRY_TRAVEL_ICON.getGraphics().drawImage(fairyTravelIcon, 1, 1, null);
+		FAIRY_TRAVEL_ICON.getGraphics().drawImage(fairyTravelIcon, iconOffset, iconOffset, null);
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
```diff
@@ -125,15 +128,15 @@ public class WorldMapPlugin extends Plugin
 
 		NOT_STARTED_ICON = new BufferedImage(questIconBufferSize, questIconBufferSize, BufferedImage.TYPE_INT_ARGB);
 		final BufferedImage notStartedIcon = ImageUtil.loadImageResource(WorldMapPlugin.class, "quest_not_started_icon.png");
-		NOT_STARTED_ICON.getGraphics().drawImage(notStartedIcon, 4, 4, null);
+		NOT_STARTED_ICON.createGraphics().drawImage(notStartedIcon, questIconOffset, questIconOffset, null);
```
Nightfirecat, line 131: Why is `getGraphics()` changed to `createGraphics()` for these updated lines?

runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
```diff
@@ -110,8 +110,11 @@ public class WorldMapPlugin extends Plugin
 		//A size of 17 gives us a buffer when triggering tooltips
 		final int iconBufferSize = 17;
```
Nightfirecat, line 111: This code is generally quite old, and these constants aren't the most helpful at explaining why they are the size they are. Vanilla world map icons are 15x15, so this size is chosen to give them an additional 1-pixel buffer for tooltips. Quest icons need to be 25x25 (which is 5 pixels of additional space on each edge) to accommodate the 20x20 images (the standard 15x15 quest icon with 5 pixels on the right and bottom edges) being bottom-right aligned.

I think I'd prefer if we update this whole static block to center around a `final int worldMapIconSize = 15;` and some variables which indicate how much it should be expanded for "standard" icon overrides (fairy travel, nope icon, mining sites, rooftop courses) and for quest icon overrides.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
Adam-: They look to be the same 
image

runelite-client/src/main/java/net/runelite/client/plugins/worldmap/WorldMapPlugin.java
Nightfirecat: I don't know of any reason offhand. I'm not sure anyone has looked into amending the world map scripts to achieve this. Assuming it is possible to achieve the same effect, I agree it would likely work better to take this approach, though that can happen in a separate PR if it's found to be feasible.

PR #17658, opened 2024-03-31
ammo: dizana's quiver tracking

LlemonDuck:
Track Dizana's quiver ammo and show it in an infobox. 

The quiver ammo is tracked in two varps: 4141 for the amount, and 4142 for the item id, so it is slightly more involved than just reacting to inventory events as the plugin is already doing.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoPlugin.java
@@ -24,7 +24,9 @@
  */
 package net.runelite.client.plugins.ammo;
 
+import com.google.common.collect.ImmutableSet;
 import java.awt.image.BufferedImage;
+import java.util.Set;
 import javax.inject.Inject;
 import net.runelite.api.Client;
 import net.runelite.api.EquipmentInventorySlot;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoPlugin.java
```diff
@@ -90,16 +95,26 @@ public void onItemContainerChanged(ItemContainerChanged event)
...
+	public void onGameTick(GameTick event)
+	{
+		checkQuiver();
 	}
 
-	private void checkInventory(final Item[] items)
+	private void checkInventory(ItemContainer equipment)
 	{
+		final Item cape = equipment.getItem(EquipmentInventorySlot.CAPE.getSlotIdx());
+		this.isWearingQuiver = cape != null && itemManager.getItemComposition(cape.getId()).getName().startsWith("Dizana");
```
Nightfirecat, line 110: I'd rather use ItemVariationMapping to check for the cape being the quiver or quiver max cape explicitly here.

runelite-client/src/main/java/net/runelite/client/plugins/ammo/AmmoPlugin.java
```diff
@@ -90,16 +95,26 @@ public void onItemContainerChanged(ItemContainerChanged event)
...
 		}
 
-		checkInventory(event.getItemContainer().getItems());
+		checkInventory(event.getItemContainer());
+	}
+
+	@Subscribe
+	public void onGameTick(GameTick event)
+	{
+		checkQuiver();
```
Nightfirecat, line 104: Not that it's particularly costly to do this every tick, but is there a reason this doesn't happen on var change?

Nightfirecat (approved):

dillydill123:
What's blocking this PR?

PR #17661, opened 2024-03-31
Add auto rejoin functionality to party plugin

Drumstepp:
Add the option for players to opt into automatically rejoining their previous party upon login. This is useful for groups of friends that frequently play together, and don't want to remember to rejoin their party every time they launch the game. 

Summary of changes: 

- Adds a config option for the user to opt into automatically rejoining their previous party. 
- Adds a hidden config option to track if automatic party rejoining should be skipped. Upon the user manually leaving the party through the leave button, this flag is set to true. Upon joining a party again, this flag is set back to false.
- Adds checks on the `onGameStateChanged` event to check 1. If the game state is `LOGGED_IN` 2. If the player is not currently in a party 3. If the player has a previous party ID 4. If the player has opted into automatically rejoining their previous party 5. If automatically rejoining the previous party has not been disabled due to the player manually leaving the party. 
  - If all checks are successful, a call is made to `party.changeParty()` with the user's previous party ID.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyConfig.java
@@ -72,11 +72,22 @@ public interface PartyConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "autoJoinPreviousParty",
+		name = "Auto join previous party",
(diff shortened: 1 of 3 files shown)


cwjoshuak:
https://github.com/runelite/runelite/pull/15176#issuecomment-1161837688

PR #17662, opened 2024-04-01
timers: sectionize and add varlamore buffs

YvesW:
This PR:
- Renames timers to timers & buffs
- Sectionizes the config
- Adds the curse of the moons buff (#17586 unlisted)
- Adds the colosseum doom buff (#17586 unlisted)
- Makes vengeance into a buff counter (added later at the request of NFC)
- Adds a timer for the moonlight potion (added later since varbit wasn't initially available/known). Please refer to https://github.com/runelite/runelite/pull/17662#issuecomment-2067964626 for info about this commit

I've also included footage of ingame tests below. If you'd like me to split up the sectionizing into a separate PR, change/drop a commit, or perform other ingame tests etc., please let me know.

<details>
 <summary>Colosseum Doom tests</summary>

- Infobox persists between waves and gets removed when the player dies: https://youtu.be/zW9vs9m8DiQ
- Infobox gets removed while dying to doom: https://youtu.be/mHv0mvBWkUE
- Infobox gets removed when leaving the arena: https://youtu.be/FEEeGlXtDrg
- Infobox gets removed when teleporting out: https://youtu.be/aWUR-A_mcJs
- The varbit apparently does not get reset to 0 when teleporting out, so I added the regionId code because of that. This is before I added that regionId code: https://youtu.be/7pjB3K7_7wQ
  
</details>

<details>
<summary>Curse of the Moons tests</summary>
  
- Full run: https://youtu.be/XfM9v1dxRoQ
- Eclipse moon infobox gets removed when teleporting out: https://youtu.be/m78s1E7CIWI
- Eclipse moon infobox gets removed when leaving: https://youtu.be/P-86THsJp6k
- Eclipse moon infobox gets removed when dying: https://youtu.be/r9dhxFMGLIM
- Blue moon infobox gets removed when dying + shows off the ``joint lock up`` at 18 stacks: https://youtu.be/GFHiiJC8wKI
  - EHC's infobox text also changes color at 18 stacks, so this is not against the statement.
- Blue moon infobox gets removed when leaving: https://youtu.be/VH5dyxp7Fl4
- Blue moon infobox gets removed when tping out: https://youtu.be/nAqkg-6EcoU
  
</details>

Alternatively I could use <https://abextm.github.io/cache2/#/viewer/sprite/1361> as icon, which the official client does IIRC, but these icons make a lot more sense to me than a random yellow sprite.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -892,4 +892,14 @@ public final class Varbits
 
 	public static final int SPELLBOOK = 4070;
 	public static final int SPELLBOOK_SUBMENU = 9730;
+
+	/**
+	 * The amount of Curse of the Moons stacks received when fighting the Blue Moon or Eclipse Moon.
+	 * The varbit value remains 0 when fighting the Blood Moon.
+	 * When fighting the Blue Moon, the player's joints will lock up at 18 stacks, which causes their next attack to be
+	 * canceled and 18 stacks to be removed.
+	 * When fighting the Eclipse Moon, the stacks increase the chance of a player's attack glancing off the shield of
+	 * the Eclipse Moon. Glancing attacks reduce the player's max hit by two times the flat armour of the Eclipse Moon.
+	 */
+	public static final int CURSE_OF_THE_MOONS = 9853;
 }
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -90,6 +90,15 @@ public final class Varbits
 	public static final int DIVINE_BATTLEMAGE = 13665;
 
 	/**
+	 * Moonlight potion timer.
+	 * When at least 70 herblore, the moonlight potion's defense effect will be removed when this timer runs out.
+	 * If the player drinks a dose of moonlight potion while already under its effects, desync between
+	 * Varbits.MOONLIGHT_POTION and Varbits.DIVINE_SUPER_DEFENCE can occur, with the latter being 1 tick greater.
+	 * In case of desync, the moonlight defence effect will be removed once Varbits.DIVINE_SUPER_DEFENCE becomes 0.
+	 */
+	public static final int MOONLIGHT_POTION = 10029;
+
+	/**
 	 * Ring of endurance effect timer, stamina duration extended from using the ring of endurance
 	 * Number of game ticks remaining on ring of endurance effect in intervals of 10; for a value X there are 10 * X game ticks remaining.
 	 * Unequipping the ring of endurance will cause this to change to 0.
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -902,4 +902,9 @@ public final class Varbits
 	 * the Eclipse Moon. Glancing attacks reduce the player's max hit by two times the flat armour of the Eclipse Moon.
 	 */
 	public static final int CURSE_OF_THE_MOONS = 9853;
+
+	/**
+	 * The amount of Doom stacks received in the Fortis Colosseum.
+	 */
+	public static final int COLOSSEUM_DOOM = 9801;
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/BuffCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/BuffCounter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Tyler <https://github.com/tylerthardy>
+ * Copyright (c) 2024, YvesW <https://github.com/YvesW>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -25,32 +25,35 @@
 package net.runelite.client.plugins.timersandbuffs;
 
 import java.awt.Color;
+import lombok.AccessLevel;
 import lombok.Getter;
-import net.runelite.client.plugins.Plugin;
-import net.runelite.client.ui.overlay.infobox.InfoBox;
-import net.runelite.client.ui.overlay.infobox.InfoBoxPriority;
+import net.runelite.client.ui.overlay.infobox.Counter;
 
-public class IndicatorIndicator extends InfoBox
+@Getter(AccessLevel.PACKAGE)
+class BuffCounter extends Counter
 {
-	@Getter
-	private final GameIndicator indicator;
+	private final TimersAndBuffsPlugin plugin;
+	private final GameCounter gameCounter;
 
-	IndicatorIndicator(GameIndicator indicator, Plugin plugin)
+	BuffCounter(
+		TimersAndBuffsPlugin plugin,
+		GameCounter gameCounter,
+		int count)
 	{
-		super(null, plugin);
-		this.indicator = indicator;
-		setPriority(InfoBoxPriority.MED);
+		super(null, plugin, count);
+		this.plugin = plugin;
+		this.gameCounter = gameCounter;
 	}
 
 	@Override
 	public String getText()
 	{
-		return indicator.getText();
+		return gameCounter.isShouldDisplayCount() ? Integer.toString(getCount()) : "";
 	}
 
 	@Override
 	public Color getTextColor()
 	{
-		return indicator.getTextColor();
+		return gameCounter.getColorBoundaryType().shouldRecolor(getCount(), gameCounter.getBoundary()) ? gameCounter.getColor() : Color.WHITE;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
@@ -36,6 +36,7 @@ import net.runelite.api.SpriteID;
 @AllArgsConstructor
 enum GameCounter
 {
+	COLOSSEUM_DOOM(SpriteID.COLOSSEUM_DOOM, GameTimerImageType.SPRITE, "Doom"),
 	CURSE_OF_THE_MOONS_BLUE(ItemID.BLUE_MOON_HELM, GameTimerImageType.ITEM, "Curse of the Moons (Blue Moon)", ColorBoundaryType.GREATER_THAN_EQUAL_TO, 18, Color.RED),
 	CURSE_OF_THE_MOONS_ECLIPSE(ItemID.ECLIPSE_MOON_HELM, GameTimerImageType.ITEM, "Curse of the Moons (Eclipse Moon)"),
 	VENGEANCE_ACTIVE(SpriteID.SPELL_VENGEANCE_OTHER, GameTimerImageType.SPRITE, "Vengeance active", false),
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Tyler <https://github.com/tylerthardy>
+ * Copyright (c) 2024, YvesW <https://github.com/YvesW>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -25,32 +25,54 @@
 package net.runelite.client.plugins.timersandbuffs;
 
 import java.awt.Color;
+import java.util.function.BiPredicate;
 import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Getter;
 import net.runelite.api.SpriteID;
 
 @Getter(AccessLevel.PACKAGE)
-enum GameIndicator
+@AllArgsConstructor
+enum GameCounter
 {
-	VENGEANCE_ACTIVE(SpriteID.SPELL_VENGEANCE_OTHER, GameTimerImageType.SPRITE, "Vengeance active");
+	VENGEANCE_ACTIVE(SpriteID.SPELL_VENGEANCE_OTHER, GameTimerImageType.SPRITE, "Vengeance active", false),
+	;
 
-	private final String description;
-	private String text;
-	private Color textColor;
 	private final int imageId;
 	private final GameTimerImageType imageType;
+	private final String description;
+	private final ColorBoundaryType colorBoundaryType;
+	private final int boundary;
+	private final Color color;
+	private final boolean shouldDisplayCount;
 
-	GameIndicator(int imageId, GameTimerImageType idType, String description, String text, Color textColor)
+	GameCounter(int imageId, GameTimerImageType idType, String description, ColorBoundaryType colorBoundaryType, int boundary, Color color)
 	{
-		this.imageId = imageId;
-		this.imageType = idType;
-		this.description = description;
-		this.text = text;
-		this.textColor = textColor;
+		this(imageId, idType, description, colorBoundaryType, boundary, color, true);
 	}
 
-	GameIndicator(int imageId, GameTimerImageType idType, String description)
+	GameCounter(int imageId, GameTimerImageType idType, String description, boolean shouldDisplayCount)
 	{
-		this(imageId, idType, description, "", null);
+		this(imageId, idType, description, ColorBoundaryType.NO_BOUNDARY, 0, Color.WHITE, shouldDisplayCount);
+	}
+
+	GameCounter(int imageId, GameTimerImageType idType, String description)
+	{
+		this(imageId, idType, description, ColorBoundaryType.NO_BOUNDARY, 0, Color.WHITE, true);
+	}
+
+	@AllArgsConstructor
+	enum ColorBoundaryType
+	{
+		GREATER_THAN_EQUAL_TO((count, boundary) -> count >= boundary),
+		LESS_THAN_EQUAL_TO((count, boundary) -> count <= boundary),
+		NO_BOUNDARY((count, boundary) -> false);
+
+		final BiPredicate<Integer, Integer> shouldRecolorPredicate;
+
+		boolean shouldRecolor(int count, int boundary)
+		{
+			return shouldRecolorPredicate.test(count, boundary);
+		}
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
@@ -29,12 +29,15 @@ import java.util.function.BiPredicate;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import net.runelite.api.ItemID;
 import net.runelite.api.SpriteID;
 
 @Getter(AccessLevel.PACKAGE)
 @AllArgsConstructor
 enum GameCounter
 {
+	CURSE_OF_THE_MOONS_BLUE(ItemID.BLUE_MOON_HELM, GameTimerImageType.ITEM, "Curse of the Moons (Blue Moon)", ColorBoundaryType.GREATER_THAN_EQUAL_TO, 18, Color.RED),
+	CURSE_OF_THE_MOONS_ECLIPSE(ItemID.ECLIPSE_MOON_HELM, GameTimerImageType.ITEM, "Curse of the Moons (Eclipse Moon)"),
 	VENGEANCE_ACTIVE(SpriteID.SPELL_VENGEANCE_OTHER, GameTimerImageType.SPRITE, "Vengeance active", false),
 	;
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
@@ -587,6 +587,11 @@ public class TimersAndBuffsPlugin extends Plugin
 				updateVarCounter(CURSE_OF_THE_MOONS_BLUE, event.getValue());
 			}
 		}
+
+		if (event.getVarbitId() == Varbits.COLOSSEUM_DOOM && config.showColosseumDoom())
+		{
+			updateVarCounter(COLOSSEUM_DOOM, event.getValue());
+		}
 	}
 
 	@Subscribe
@@ -788,6 +793,11 @@ public class TimersAndBuffsPlugin extends Plugin
 		{
 			removeVarCounter(CURSE_OF_THE_MOONS_BLUE);
 			removeVarCounter(CURSE_OF_THE_MOONS_ECLIPSE);
+		}
+
+		if (!config.showColosseumDoom())
+		{
+			removeVarCounter(COLOSSEUM_DOOM);
 		}
 	}
 
@@ -1077,6 +1087,8 @@ public class TimersAndBuffsPlugin extends Plugin
 					config.tzhaarStartTime(null);
 					config.tzhaarLastTime(null);
 				}
+				// Varbits.COLOSSEUM_DOOM is not set to 0 when teleporting out of the Colosseum.
+				removeVarCounter(COLOSSEUM_DOOM);
 				break;
 			case LOGIN_SCREEN:
 				// fall through
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
@@ -24,7 +24,7 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package net.runelite.client.plugins.timers;
+package net.runelite.client.plugins.timersandbuffs;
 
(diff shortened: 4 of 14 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
```diff
@@ -0,0 +1,69 @@
...
+	;
+
+	private final int imageId;
+	private final GameTimerImageType imageType;
+	private final String description;
+	private final ColorBoundaryType colorBoundaryType;
+	private final Color color;
+	private final int boundary;
+
+	GameCounter(int imageId, GameTimerImageType idType, String description, ColorBoundaryType colorBoundaryType, Color color, int boundary)
```
Nightfirecat, line 48: Replace with `@AllArgsConstructor`

runelite-api/src/main/java/net/runelite/api/Varbits.java
```diff
@@ -886,4 +886,14 @@ public final class Varbits
 	 * 1 = active
 	 */
 	public static final int SPELLBOOK_SWAP = 3617;
+
+	/**
+	 * The amount of Curse of the Moons stacks received when fighting the Eclipse Moon or Blue Moon.
```
Nightfirecat, line 891: Does this not update when fighting the Blood Moon?

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/BuffCounter.java
```diff
@@ -0,0 +1,60 @@
...
+		this.gameCounter = gameCounter;
+	}
+
+	@Override
+	public Color getTextColor()
+	{
+		switch (gameCounter.getColorBoundaryType())
+		{
+			case LESS_THAN_EQUAL_TO:
+				return getCount() <= gameCounter.getBoundary() ? gameCounter.getColor() : Color.WHITE;
```
Nightfirecat, line 54: These could be made into Predicate methods of the BoundaryType enum fields to simplify this getTextColor implementation

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -1044,6 +1080,13 @@ public void onGameTick(GameTick event)
 			}
 		}
 
+		final int regionID = WorldPoint.fromLocalInstance(client, client.getLocalPlayer().getLocalLocation()).getRegionID();
```
Nightfirecat, line 1083: Are there any teleports which leave you in the same region? (does the dueling ring land you inside that region, or just outside of it?) If not, we could remove this counter on GameState -> LOADING instead of checking each tick.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -76,19 +76,20 @@
 import net.runelite.client.game.SpriteManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
-import static net.runelite.client.plugins.timers.GameIndicator.VENGEANCE_ACTIVE;
-import static net.runelite.client.plugins.timers.GameTimer.*;
+import static net.runelite.client.plugins.timersandbuffs.GameCounter.*;
+import static net.runelite.client.plugins.timersandbuffs.GameIndicator.VENGEANCE_ACTIVE;
```
Nightfirecat, line 80: Can Vengeance be made into a BuffCounter?

runelite-api/src/main/java/net/runelite/api/Varbits.java
Nightfirecat: Regardless of that, can this javadoc be expanded upon? The wiki is not the most helpful for understanding what these values might be, or what they do. This would also help to understand why we change the color for the eclipse moon buff counter when >=18.

YvesW: The varbit does indeed not update when fighting the Blood Moon ([getVarbitValue every GameTick while getting hit by the blood moon](https://youtu.be/aB2cXuqKnmU)).
The color is changed at >= 18 Curse of the Moon stacks when fighting the **Blue** Moon, because the player's joints will lock up at 18 stacks, which causes their next attack to be canceled and 18 stacks to be removed.
I've added this info to the javadoc. Please let me know if you'd like me to expand on it further or if you'd like to see a different change.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/BuffCounter.java
YvesW: Apologies, I don't have a lot of experience with using Predicates. I tried using Predicate [here](https://github.com/YvesW/runelite/commit/2c57d1cac0ba6bd1d3fd41d7af442ff07ec25d86) but failed. In the end I used BiPredicate, but I doubt that is what you meant.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
YvesW: That's a good shout! I tested it and created https://github.com/runelite/runelite/pull/17727
It turns out that the westernmost tiles are outside the region, while the other ones are in the same region as the colosseum. Since the regionID for the instanced and non-instanced version of the Colosseum is the same, I added an ``isInInstancedRegion()`` check.
I assume teleporting to the same regionID but going from the instanced to the non-instanced version would not proc ``GameState.LOADING``, based on your commment? I sadly don't have the colosseum teleport unlocked and I can't think of any other similar regions with a teleport, so I can't test it out myself.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/BuffCounter.java
Nightfirecat: No, that looks correct--I hadn't thought about it needing two args instead of one.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
```diff
@@ -0,0 +1,82 @@
...
+	CURSE_OF_THE_MOONS_ECLIPSE(ItemID.ECLIPSE_MOON_HELM, GameTimerImageType.ITEM, "Curse of the Moons (Eclipse Moon)"),
+	VENGEANCE_ACTIVE(SpriteID.SPELL_VENGEANCE_OTHER, GameTimerImageType.SPRITE, "Vengeance active", false),
+	;
+
+	private final int imageId;
+	private final GameTimerImageType imageType;
+	private final String description;
+	private final ColorBoundaryType colorBoundaryType;
+	private final Color color;
+	private final int boundary;
```
Nightfirecat, line 50: Can we maybe reorder these a bit? I'd want them to more closely resemble the mental thought of "if this number matches this boundary type, it should be so-and-so color". In this case, maybe the following?

```java
private final ColorBoundaryType colorBoundaryType;
private final int boundary;
private final Color color;
```

And I'd want to change the first constructor to match. (swap the `color` and `boundary` parameter positions)

runelite-api/src/main/java/net/runelite/api/Varbits.java
Nightfirecat: This looks better. Do we know enough about the eclipse moon's stacks to add a similar description for it?

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -76,19 +76,19 @@
...
-	name = "Timers",
-	description = "Show various timers in an infobox",
-	tags = {"combat", "items", "magic", "potions", "prayer", "overlay", "abyssal", "sire", "inferno", "fight", "caves", "cape", "timer", "tzhaar", "thieving", "pickpocket", "hunter", "impling", "puro"}
+	name = "Timers & Buffs",
+	description = "Show various timers and buffs in an infobox",
+	tags = {"combat", "items", "magic", "potions", "prayer", "overlay", "abyssal", "sire", "inferno", "fight", "caves", "cape", "timer", "tzhaar", "thieving", "pickpocket", "hunter", "impling", "puro", "buff"}
 )
 @Slf4j
-public class TimersPlugin extends Plugin
+public class TimersAndBuffsPlugin extends Plugin
```
Nightfirecat, line 91: Changing the plugin class name would also change the plugin enabled config; to address this, the plugin descriptor needs `configName = "TimersPlugin"` to ensure it remains disabled for users who toggled it off.

YvesW:
For some reason I can't respond to the comment (maybe because it's marked as ``outdated``), so I'll respond here instead.
Re: "This looks better. Do we know enough about the eclipse moon's stacks to add a similar description for it?":

Originally the ``Eclipse Moon`` wiki page contained information about the Curse of the Moon effect. However, when Spine moved that info to the ``Moons of Peril/Strategies#Eclipse Moon`` page, he removed the info about the stacks, because he was not sure what a "glancing blow" actually did. Since I did not want to add potentially incorrect information to the javadoc, I decided against adding it before.
I see now that the actual effect has been added to the ``Moons of Peril/Strategies#Eclipse Moon`` page since then. I've added it to the javadoc!

YvesW:
As discussed before, I've added the moonlight potion's timer.
The moonlight potion is a special one though. See the [wiki](https://oldschool.runescape.wiki/w/Moonlight_potion#Formulas) for its effects. The prayer restore and attack/strength boost are not relevant for the timers (no divine timer).
- When <70 herblore, it will work as a normal or super att/str/def potion. No divine effects/timers.
- When >= 70 herblore:
  - The moonlight potion will act as a super attack and super strenght potion. These are non-divine; they keep draining over time, also after the potion's varbit/divine effect has ran out. Not relevant for this PR.
  - The defence buff is relevant though.
    - Ingame effect: its defence buff is bigger than that of a super defence potion. Its timer/varbit will tick down like a divine, after which your defence will be reset to its base level. Interestingly enough, unlike divines, the buff itself does drain over time.
    - Varbits: when a player takes a sip of the potion, ``Varbits.MOONLIGHT_POTION`` (10029) and ``Varbits.DIVINE_SUPER_DEFENCE`` will be set. You will even receive the ``Your divine potion effect is about to expire.`` and ``The effects of the divine potion have worn off.`` chatbox messages when it runs out.

Please note that by drinking a moonlight potion while already under the effect of one, you can get **desync** between the ``Varbits.MOONLIGHT_POTION`` and ``Varbits.DIVINE_SUPER_DEFENCE``, where the latter is 1 tick greater. This would cause both a moonlight potion and divine def timer to show up, and the moonlight potion timer would end 1 tick too soon. Thus, I've opted to implement a workaround.
I was not able to recreate this desync with divine SCBs; I suspect it's unique for the moonlight potion.

Video of the desync before the workaround:

https://github.com/runelite/runelite/assets/7929021/f0713e0f-b873-4380-aa0e-6d9d595e2cbc

Desync visible in the var inspector:
moonlight_potion_desync_inspector


No desync:
```
8431 changed to 0 tick: 514
10029 changed to 0 tick: 514
DEFENCE boosted lvl: 99 level: 99 at tick: 514
```

Desync:
```
10029 changed to 1 tick: 1070
8431 changed to 1 tick: 1071
10029 changed to 0 tick: 1071 (<- moonlight timer varbit runs out at tick 1071)
ATTACK boosted lvl: 113 level: 99 at tick: 1071 (<- this is the normal stat drain)
DEFENCE boosted lvl: 120 level: 99 at tick: 1071 (<- this is the normal stat drain)
STRENGTH boosted lvl: 113 level: 99 at tick: 1071 (<- this is the normal stat drain)
8431 changed to 0 tick: 1072 (<- super def timer varbit runs out at tick 1072)
DEFENCE boosted lvl: 99 level: 99 at tick: 1072 (<- moonlight potion def effect runs out at tick 1072)
```

---

I also tested some interactions, partially because the moonlight/super def desync made me somewhat suspicious. Nothing really unexpected, except maybe when leaving the dungeon while under the effects of a divine SCB and moonlight potion. If you want to change this, please let me know, but I doubt it's worth it.
- Divine SCB & moonlight interaction
  - Drinking a divine SCB after moonlight potion
    - Varbits: moonlight varbit keeps ticking down, divine att/str/def/scb get set to new value.
    - Ingame stats: def buf remains >super def until it has slowly ticked down to divine def levels.
    - Timers: both the moonlight and divine SCB timers are shown as expected.
  - Drinking a moonlight potion after a divine SCB
    - Varbits: super attack/str/SCB varbits keep ticking down, the divine def/moonlight potion get set to a new value. The above described desync can of course happen when drinking multiple sips of the moonlight potion.
    - Stats: the def buff remains >super def until the moonlight potion times out, att/str get set to base level after the divine SCB times out (as expected).
    - Timers: divine SCB and moonlight timers are shown as expected (only tested after desync workaround).
- When leaving the dungeon
  - When under the effect of the moonlight potion, both the moonlight varbit and super def varbit get set to 0. All stats are removed.
  - While also under the effect of a divine SCB, the moonlight and super defense varbits get set to 0. The divine super att/str/SCB varbits keep ticking. All stats get reset to base level. If you drink a regular SCB, your attack and strength get divine boosted (remain at 118, boost disappears when divine SCB timer runs out). Your defense gets normal super defence boosted (not divine; normal drain occurs).
- When dying, all stats are removed and the SCB effect is gone.


All testing has been done on the beta worlds.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -477,7 +476,8 @@ else if (nextSuperAntifireTick - tickCount <= 0)
 		{
 			if (client.getVarbitValue(Varbits.DIVINE_SUPER_COMBAT) > event.getValue()
 				|| client.getVarbitValue(Varbits.DIVINE_BASTION) > event.getValue()
-				|| client.getVarbitValue(Varbits.DIVINE_BATTLEMAGE) > event.getValue())
+				|| client.getVarbitValue(Varbits.DIVINE_BATTLEMAGE) > event.getValue()
+				|| client.getVarbitValue(Varbits.MOONLIGHT_POTION) > event.getValue())
```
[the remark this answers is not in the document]
Nightfirecat: Also, similarly to how drinking a moonlight potion removes existing divine super defence potion infoboxes, drinking a divine super defence potion (or having this var update from any of the other divine potions which provide its effect) should remove the moonlight potion's infobox, since it preserves the defence boost beyond the 5 minutes provided by the moonlight potion.

That would should be added after this guard clause, eg.

```java
if (client.getVarbitValue(Varbits.MOONLIGHT_POTION) < event.getValue())
{
	removeVarTimer(MOONLIGHT_POTION);
}
```

Nightfirecat (commented):
Besides super defence potion infobox not removing moonlight potion infobox, and that weird lobby doom counter bug, the behavior looks good now.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameCounter.java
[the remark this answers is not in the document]
YvesW: Some of the overloaded constructors are somewhat superfluous now in that vengeance commit. Normally I'd be adding those in the relevant commits that follow after the vengeance commit. That'd also require moving ColorBoundaryType to another commit though, which requires changing BuffCounter etc. The way I've done it rn, GameCounter is basically only touched to add the enum entry, as you requested.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -1059,6 +1111,8 @@ public void onGameStateChanged(GameStateChanged gameStateChanged)
 					config.tzhaarStartTime(null);
 					config.tzhaarLastTime(null);
 				}
+				// Varbits.COLOSSEUM_DOOM is not set to 0 when teleporting out of the Colosseum.
+				removeVarCounter(COLOSSEUM_DOOM);
```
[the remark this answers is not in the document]
YvesW: Hmmm, I don't really understand how this can happen unless they changed the way it behaves on death/leaving significantly after I tested it here: https://youtu.be/mHv0mvBWkUE. Arguably, that was with the regionId check that I removed as a response to https://github.com/runelite/runelite/pull/17662#discussion_r1563674085, that ran every gametick.
GameState.LOADING is procced after both leaving and dying at least: https://youtu.be/_FpmBZCNBmo

Did you happen to run in debug mode, so we can maybe see if e.g. the varbit changed to 0 after the GameState changed? In that case we might have to revert to the regionId check?

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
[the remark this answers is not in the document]
YvesW: Regarding https://github.com/runelite/runelite/pull/17662#discussion_r1677028493:
With ``one tick`` you are referring to how the divine super def seems to send the varbitchanged event before the moonlight potion? I.e. on that gametick after the super def var value has --'d, before the moonlight potion var value has --'d, super def var == moonlight potion var?
That'd leave the moonlight potion timer 1 gametick incorrect though, no? What do you think of changing it to ``>=`` as you suggested, and just removing:
```java
if (client.getVarbitValue(Varbits.DIVINE_SUPER_DEFENCE) == moonlightValue)
{
	removeVarTimer(DIVINE_SUPER_DEFENCE);
}
```
(i.e. not completely removing the +1 hack, just the part that removes the super def timer every tick).
I've implemented this change for now.

YvesW:
I rebased my branch onto master before making any changes, then created a branch from this varlamore-buff-bar branch (before making any changes) and pushed it (https://github.com/YvesW/runelite/tree/varlamore-buff-bar-backup-14-july). This with the aim that you could more easily compare my changes compared to your last review.
I'm either doing something wrong, or https://github.com/YvesW/runelite/compare/varlamore-buff-bar-backup-14-july...YvesW:varlamore-buff-bar looks to be incorrect though. It looks like it's comparing with master instead of the branch I created.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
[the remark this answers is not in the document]
Nightfirecat: That's fine, however the comment in the `MOONLIGHT_POTION` block isn't really very useful now; it would be helpful to explain that without that code, the timer would be 1 tick shorter than it is meant to be.

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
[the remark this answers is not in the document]
Nightfirecat: I can see in my logs game state goes from LOADING to LOGGED_IN, after which point a counter with count=0 is added. It looks like the ordering of checks inside `updateVarcounter()` is wrong; the removal check (in this case, checking `varValue == 0`) should come before creating a new counter if the existing counter is null. That way it will match the behavior of `updateVarTimer()`, which we know to work correctly.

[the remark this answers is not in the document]
YvesW: Gosh darnit, I should have noticed that when I wrote the original code, apologies. It didn't come up in testing as the counters were never removed before the varValue was being set to 0, except the Doom counter, but I was removing that counter every tick so I didn't notice.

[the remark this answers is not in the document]
YvesW: Just had the option to test it; the infobox properly gets removed without getting erroneuously re-added when going to the underground area on the latest commit! https://youtu.be/Srja461wWLk

Can also be seen in the debug logs:
```2024-07-25 13:02:27 CEST [Client] DEBUG com.weaponcharges.DialogTracker - dialog changed: DialogState{OPTIONS, text='Are you sure you wish to leave?', options=[Yes., No.]} previous: DialogState{NO_DIALOG} (game tick: 4935)
2024-07-25 13:02:29 CEST [Client] DEBUG com.weaponcharges.DialogTracker - option selected: "Yes." DialogState{OPTIONS, text='Are you sure you wish to leave?', options=[Yes., No.]}
2024-07-25 13:02:29 CEST [Client] DEBUG com.weaponcharges.DialogTracker - dialog changed: DialogState{NO_DIALOG} previous: DialogState{OPTIONS, text='Are you sure you wish to leave?', options=[Yes., No.]} (game tick: 4938)
2024-07-25 13:02:30 CEST [Client] DEBUG c.d.o.f.util.ColosseumStateTracker - varb 9788 = 0, no modifier selected?
2024-07-25 13:02:31 CEST [Client] DEBUG injected-client - Game state changed: LOADING
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Removed InfoBoxes for filter net.runelite.client.plugins.timersandbuffs.TimersAndBuffsPlugin$$Lambda$4315/0x0000000801509c40@557be474 from net.runelite.client.ui.overlay.infobox.InfoBoxOverlay@75e760d8
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.o.ObjectIndicatorsPlugin - Marking object lp@7998cbb0 due to matching ObjectPoint(id=50748, name=Bank chest, regionId=7316, regionX=13, regionY=29, z=0, borderColor=java.awt.Color[r=255,g=255,b=0], fillColor=null, hull=null, outline=null, clickbox=null, tile=null)
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.o.ObjectIndicatorsPlugin - Marking object lp@693ad846 due to matching ObjectPoint(id=50751, name=Entrance, regionId=7316, regionX=18, regionY=35, z=0, borderColor=java.awt.Color[r=255,g=255,b=0], fillColor=null, hull=null, outline=null, clickbox=null, tile=null)
2024-07-25 13:02:31 CEST [Client] DEBUG injected-client - Game state changed: LOGGED_IN
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7059
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7060
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7061
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7315
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7316
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.p.g.GroundMarkerPlugin - Loading points for region 7317
2024-07-25 13:02:31 CEST [Client] DEBUG injected-client - Replacing archive 12/89
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.f.util.ColosseumStateTracker - Colosseum state change ColosseumState(inLobby=false, inColosseum=true, waveNumber=3, waveStarted=false) => ColosseumState(inLobby=true, inColosseum=false, waveNumber=1, waveStarted=false)
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.LeftClickBankAll]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.loot.LootHider]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.loot.LootTotalDisplay]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.modifiers.ModifierOverlay]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.modifiers.ModifierSelectNamesManager]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.timetracking.SplitsOverlay]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.timetracking.SplitsTracker]
2024-07-25 13:02:31 CEST [Client] DEBUG c.d.o.fortis.module.ComponentManager - Disabling FortisColosseum plugin component [com.duckblade.osrs.fortis.features.waves.WavesOverlay]
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Removed InfoBoxes for filter net.runelite.client.plugins.npcunaggroarea.NpcAggroAreaPlugin$$Lambda$4370/0x0000000801540840@754b0f4d from net.runelite.client.ui.overlay.infobox.InfoBoxOverlay@75e760d8
2024-07-25 13:02:31 CEST [Client] DEBUG n.r.c.u.o.infobox.InfoBoxManager - Adding InfoBox Timer(startTime=2024-07-25T11:02:31.327699300Z, endTime=2024-07-25T11:12:31.327699300Z, duration=PT10M)
2024-07-25 13:03:23 CEST [Client] DEBUG injected-client - Chat message type FRIENDSCHATNOTIFICATION: You have left the channel.
```

PR #17664, opened 2024-04-01
agility: add shortcuts

Macweese:
This PR adds several missing agility shortcuts, and amends others such as the Ungael (Vorkath entrance/exit) which do not require agility.

Other changes include adding test, removing trailing spaces and sorting the entries. Other obstacles which I didn't include since I wasn't sure if it'd be considered overboard were:
<details>
  <summary>The (pass) Tree entrance to Lletya</summary>
  
image
  
</details>

<details>
  <summary>Viyeldi caves rocks</summary>
  
image
  
</details>

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
@@ -113,7 +113,9 @@ public enum AgilityShortcut
 		// Nature Grotto bridge
 		BRIDGE_3522,
 		// Rocky Ledges Viyeldi cave
-		ROCKY_LEDGE, ROCKY_LEDGE_2960, ROCKY_LEDGE_2961, ROCKS_2962, ROCKS_2963, ROCKS_2964),
+		ROCKY_LEDGE, ROCKY_LEDGE_2960, ROCKY_LEDGE_2961, ROCKS_2962, ROCKS_2963, ROCKS_2964,
+		// Abandoned Mine
+		CART_TUNNEL_29332, CART_TUNNEL_29333),
 	KARAMJA_GLIDER_LOG(1, "Log Balance", new WorldPoint(2906, 3050, 0), A_WOODEN_LOG),
 	WEISS_BROKEN_FENCE(1, "Shortcut", null, /* base id */ NullObjectID.NULL_46815)
 	{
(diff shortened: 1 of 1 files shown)


Felanbird:
these 2 should probably be snuck into this PR

```
	NATURE_GROTTO_BRIDGE_NORTH(1, "Bridge", new WorldPoint(3440, 3331, 0), BRIDGE_3522),
	NATURE_GROTTO_BRIDGE_SOUTH(1, "Bridge", new WorldPoint(3440, 3329, 0), BRIDGE_3522),
```
image

raiyni:
Needs rebased

runelite-client/src/main/java/net/runelite/client/plugins/agility/Obstacles.java
```diff
@@ -73,10 +73,10 @@ class Obstacles
 		// Dorgesh-Kaan
 		CABLE_22569, CABLE_22572, LADDER_22564, JUTTING_WALL_22552, TUNNEL_22557, PYLON_22664,
 		CONSOLE, BOILER_22635, STAIRS_22650, STAIRS_22651, STAIRS_22609, STAIRS_22608,
-		// Pollniveach
+		// Pollnivneach
 		BASKET_14935, MARKET_STALL_14936, BANNER_14937, GAP_14938, TREE_14939, ROUGH_WALL_14940,
 		MONKEYBARS, TREE_14944, DRYING_LINE,
-		// Rellaka
+		// Relleka
```
Nightfirecat, line 79: That is not how you spell Rellekka.

runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
Nightfirecat, line 1: Basically every entry in this file which references a null object ID is not working properly because it will highlight all variations (non-interactible and otherwise) for multilocs. The poison wastes grapple tree, for instance, would be highlighted even if the quest hadn't been started. The viyeldi rope rock would be highlighted even with no long rope attached to it. To highlight these objects properly, you should be checking that the transformed id also matches the expected non-null object id, like is done in `WEISS_BROKEN_FENCE`.

Nightfirecat (changes requested):

runelite-client/src/main/java/net/runelite/client/game/AgilityShortcut.java
Nightfirecat: If we can identify all the vars (and the appropriate values) which control these multilocs, that would be fine. I don't think we want the same conditional highlighting for agility shortcuts requiring diary/other unlocks--ideally those would be highlighted, but colored yellow like a shortcut which needed a higher level. Let's plan on updating those separately.

PR #17671, opened 2024-04-03
Add fortis colosseum glory value to varplayer enum

Drumstepp:
VarPlayer 4130 tracks the personal glory value of the player and is updated upon finishing a colosseum run. 

image


Felanbird:
We don't add vars to this file unless they're being used in a core plugin. If you need it for a hub plugin use `client.getVarbitValue()`

Drumstepp:
> We don't add vars to this file unless they're being used in a core plugin. If you need it for a hub plugin use `client.getVarbitValue()`

@Felanbird understood, thanks for the reply. I was following the guidance here: https://github.com/runelite/runelite/wiki/VarPlayers,-VarBits,-and-VarClients#using-a-var-in-a-plugin. Should this be updated to specify it is for core plugins only?

Nightfirecat:
We went ahead and removed that section since it's only really supposed to apply to core plugins. You'll be able to use the beforementioned API for your hub plugin without having any API definition created for it.

PR #17568, opened 2024-03-10
Add forester's campfire to firemaking skill calculator

DapperMickie:
Fixes https://github.com/runelite/runelite/issues/17383

PR #17674, opened 2024-04-03
Skill bonus system overhaul

### LINKED ISSUE #14116: Implement more thorough Skill Calculator bonus system
[See my comment here for more detail.](https://github.com/runelite/runelite/pull/13702#issuecomment-907966832) In short, we need bonuses to be able to be applied globally, to selective items, and/or exclusive to other bonuses. (eg. define certain bonuses which cannot be stacked with others)

### LINKED ISSUE #14610: Add "Zealot's robes" bonus effect to "Skill Calculator"

### LINKED ISSUE #17383: Forester's campfire in firemaking skill calculator
### Discussed in https://github.com/runelite/runelite/discussions/17375

<sup>Originally posted by **xunra** January 14, 2024</sup>
Could be done with SkillBonus like the Pyromancer outfit but instead of 1.025 use 0.33333333. I don't think there are any other negative bonuses like that but it should be easily understandable for users.</div>

DapperMickie:
Fixes #14116
Fixes #14610
Fixes #17383

So far I've made the bonuses stackable.
I added 1 example (firemaking) so you see what'll be necessary if you want to make bonuses stackable.

I will be adding to this PR.


DapperMickie:
Added a way to exclude bonuses for actions.
Added an example (prayer) so you know what needs to be done to exclude a bonus for an action.

DapperMickie:
Fixes: https://github.com/runelite/runelite/issues/17383
Closing: https://github.com/runelite/runelite/pull/17568 
Because this PR will do the same.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/SkillBonus.java
```diff
@@ -44,4 +44,15 @@ public interface SkillBonus
...
 	 */
 	float getValue();
+
+
+	/**
+	 * Gets the list of skill bonuses this skill bonus can be stacked with.
+	 *
+	 * @return List of stackable skill bonuses
+	 */
+	default SkillBonus[] getCanBeStackedWith()
```
Nightfirecat, line 54: I suppose this is mostly to manage prayer bonuses, since there's a rather complex set of bonuses--some of which can be stacked with others, others which can't--which need to be managed? (Mostly zealot robes as they don't affect the sacred bone burner, consuming bones with the sinister offering spell, cremating shades, or scattering ashes)

I worry this implementation could be error-prone as ideally this logically is defining a BiMap (because if A can be stacked with B, B should also be able to be stacked with A), and this API is very verbose and prone to typos. For instance, for PrayerBonus, this would need to be something like the following: (formatting condensed because it is substantially longer when following the project's style guidelines)

```java
enum PrayerBonus implements SkillBonus {
	LIT_GILDED_ALTAR(/* ... */) {
		@Override getCanBeStackedWith() { return [ZEALOT_ROBES]; }
	},
	ECTOFUNTUS(/* ... */) {
		@Override getCanBeStackedWith() { return [ZEALOT_ROBES]; }
	},
	CHAOS_ALTAR(/* ... */) {
		@Override getCanBeStackedWith() { return [ZEALOT_ROBES]; }
	},
	MORYTANIA_DIARY_3_SHADES(/* ... */),
	BONECRUSHER(/* ... */),
	SINISTER_OFFERING(/* ... */),
	DEMONIC_OFFERING(/* ... */),
	SACRED_BONE_BURNER(/* ... */),
	ZEALOT_ROBES(/* ... */) {
		// NOTE: What happens if we add a new entry above which can be stacked
		//       with zealot robes and forget to update this line??
		@Override getCanBeStackedWith() { return [LIT_GILDED_ALTAR, ECTOFUNTUS, CHAOS_ALTAR]; }
	}
}
```

Hence, I wonder if this would be better implemented as a static instance method returning `BiMap<SkillBonus, SkillBonus>`

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/SkillAction.java
```diff
@@ -96,4 +96,9 @@ default boolean isBonusApplicable(SkillBonus bonus)
 	}
 
 	boolean isMembers(final ItemManager itemManager);
+
+	default SkillBonus[] getExcludedSkillBonuses()
```
Nightfirecat, line 100: Is there a reason for this not to be `Set<SkillBonus>`? That would avoid unnecessary Stream uses (can just use `Collection#contains()` calls instead) and is semantically more accurate. (can't exclude a skill bonus multiple times--whatever that would mean?) I'd be open to changing the shape of `PrayerBonus.BONE_BONUSES` if that's the reason this returns an array.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/SkillBonus.java
Nightfirecat: This is mostly ramble, but this does run into an issue where, let's say down the line some skill has 10 possible bonuses all of which stack; even this solution would require `(n ** 2 + n) / 2` entry listings to achieve. It's possible this is a problem far enough out that it's not enough of a concern to deal with now though; I suspect we're never going to reach rs3 levels of stacked xp bonuses.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -386,10 +403,27 @@ private void calculate()
 			int actionCount = 0;
 			int neededXP = targetXP - currentXP;
 			SkillAction action = slot.getAction();
+
+			SkillBonus[] excludedBonuses = action.getExcludedSkillBonuses();
+			if (excludedBonuses.length > 0)
+			{
+				if (Arrays.stream(excludedBonuses).anyMatch(excludedBonus -> currentBonus.stream().anyMatch(currentBonus -> excludedBonus == currentBonus)))
+				{
+					slot.setText("N/A for current bonuses");
```
Nightfirecat, line 412: I think instead we would want something more subtle. If a player selects Zealot robes and looks at scattering ashes for instance (I understand zealot robes will be added in a separate PR, but that is probably the main reason we want this enhancement) we should just show the normal text and xp value. Maybe we could add an icon and/or tooltip to indicate some subset of selected bonuses will not apply to it.

Nightfirecat (commented):
I mostly have API considerations here.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
DapperMickie: Yeah I was mostly trying to make it obvious that this action doesn't get any multiplier when using this bonus. I'll look into the icon / tooltip suggestion.

DapperMickie:
Changed around some things because of NFC comments, would love to get feedback on these changes though because I'm not 100% sure this is what they meant.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
```diff
@@ -84,15 +86,15 @@ class SkillCalculator extends JPanel
...
+	private final Map<SkillBonus, JCheckBox> bonusCheckBoxes = new HashMap<>();
 	private final IconTextField searchBar = new IconTextField();
 
 	private CalculatorType currentCalculator;
 	private int currentLevel = 1;
 	private int currentXP = Experience.getXpForLevel(currentLevel);
 	private int targetLevel = currentLevel + 1;
 	private int targetXP = Experience.getXpForLevel(targetLevel);
-	private SkillBonus currentBonus = null;
+	private final List<SkillBonus> currentBonus = new ArrayList<>();
```
Nightfirecat, line 97: `currentBonus` -> `currentBonuses`. Also a `Set` rather than `List` perhaps, since uniqueness should be guaranteed, and ordering is not important.

Nightfirecat:
I think I've led you astray with my earlier comment--at the time I wasn't aware that Java didn't support default static methods (and therefore doesn't support overriding a static method) so my idea of a solution was completely off-base with how it could be implemented. I've pushed a commit which reverses course in that respect, and does some additional cleanup. (arrays -> sets, improvement of action definition API, etc.) I think from here, most of what's needed to be ready is to add the appropriate bonus stacking definitions to PrayerBonus.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/SkillCalculator.java
DapperMickie: Would I just do 
```
private final Set<SkillBonus> currentBonus = HashSet.newHashSet(0);
```
since 
```
private final Set<SkillBonus> currentBonus = EnumSet.noneOf(SkillBonus.class);```
isn't possible?

Nightfirecat (commented):
Per Discord comments, this needs a history cleanup which I will work on tomorrow. I think besides that, this is correct and working now.

PR #17684, opened 2024-04-05
Add setting to cannon plugin to change the counter text size

Aelthien:
Added setting to cannon plugin to change the counter text size.


Nightfirecat:
The font for this overlay is already controlled via RuneLite settings, specifically "Dynamic Overlay Font". I don't think it needs a separate control.

Aelthien:
Doesn't the dynamic font change all text though?

Nightfirecat:
Yes, that's the idea. We don't want one-off configurations for individual plugins' font settings.

PR #17696, opened 2024-04-08
cache: transform map label position into world space

christoabrown:
Noticed some labels in underground areas were weird. Turns out the map element positions are not always world coordinates.

Before
image

After
Screenshot 2024-04-07 202332

### DIFF
--- a/cache/src/main/java/net/runelite/cache/definitions/WorldMapCompositeDefinition.java
+++ b/cache/src/main/java/net/runelite/cache/definitions/WorldMapCompositeDefinition.java
@@ -25,6 +25,7 @@
 package net.runelite.cache.definitions;
 
 import lombok.Data;
+import net.runelite.cache.region.Position;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


christoabrown:
Added a couple changes to this. I changed the map icon rendering so that they render on the correct plan and also added the intermap link icons in there. I also added parsing for the intermap link locations from the client script.

This image shows the blue arrow intermap link icon and the icons for godwars which were being rendered in the void are now placed on the correct plane.
Screenshot 2024-04-08 183831

cache/src/main/java/net/runelite/cache/region/Position.java
```diff
@@ -51,6 +51,18 @@ public static Position fromPacked(int packedPosition)
 		return new Position(x, y, z);
 	}
 
+	// Unpack a position from an int in a client script
+	public static Position fromScriptInt(int scriptInt)
```
Adam-, line 55: why is this different from `fromPacked`?

cache/src/main/java/net/runelite/cache/definitions/WorldMapCompositeDefinition.java
```diff
@@ -36,4 +38,35 @@ public class WorldMapCompositeDefinition
 	public final Set<MapSquareDefinition> mapSquareDefinitions = new HashSet<>();
 	public final Set<ZoneDefinition> zoneDefinitions = new HashSet<>();
 	public final List<WorldMapElementDefinition> worldMapElementDefinitions = new ArrayList<>();
+
+	public Position calculateWorldOffset(Position position)
+	{
+		int squareX = position.getX() / 64;
+		int squareZ = position.getY() / 64;
+		int zoneX = (int)(((position.getX() / 64.0) - squareX) * 8.0);
```
Adam-, line 46: Is this something you copied from the client code? It mathematically is just `position.getX() & 7`

cache/src/main/java/net/runelite/cache/region/Position.java
christoabrown: Turns out it is the same. Got rid of it

cache/src/main/java/net/runelite/cache/definitions/WorldMapCompositeDefinition.java
christoabrown: Changed it. Did not copy, was just guessing what 'zone' meant and came up with that

christoabrown: Looking at this again, I believe they are different. `1052 & 7` = 4, but `(int)(((1052 / 64.0) - ((int)(1052/64))) * 8.0)` = 3

Adam-: Oh I misread this. It is `(x & 63)/8`. Where is this reference code from you are using?

christoabrown: Yep that works, thanks. I don't have a reference for this since, as far as I know, there are no 'source' positions for this stuff as it was created for the world map which is laid out differently for multi level stuff. This is just an effort to figure out where this stuff is located for the source squares/zones.

Adam-:
This seems okay now. Can you rebase this and fix the commit history? Ideally I think this should be in:

- fix rendering on the wrong plane
- fix label positions in underground areas
- add map links

Issue #17699, opened 2024-04-08
Moons of Peril !kc command is inconsistent with similar commands


swirle13:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10 Pro 22H2

### Bug description

The `!kc` command for Moons of Peril is inconsistent with other multi-boss commands, e.g. barrows. The command doesn't match the name of the boss in the collection log, "Moons of Peril" and instead is the name of the object, "Lunar Chest".

I would expect the command `!kc Moons of Peril` or `!kc moons of peril` to work, but they do not. `!kc lunar chest` is the command that works.

Barrows command for kc does not behave this way. `!kc barrows` returns the proper kill count, even though the collection log states "Barrows Chests opened", similar to Moons of Peril collection log stating "Lunar Chests opened".

Also, "!kc barrows chest" doesn't work, as the collection log entry is "Barrows chests", and thus "!kc barrows chest**s**" _does_ work, but the Moons of Peril command is flipped and is incorrect: "!kc lunar chests" does not work while "!kc lunar chest" does.

I would expect the behavior to allow the name of the boss, as displayed in the collection log, "Moons of Peril", as well as have abbreviations added for it, e.g. "moons", "perilous moons", etc.

### Screenshots or videos

The following screenshot is the output of the following commands:

```
!kc barrows chest
!kc barrows chests
!kc lunar chest
!kc lunar chests
```
![image](https://github.com/runelite/runelite/assets/9778474/539a3ab1-2011-4887-a169-d9b08ab2ab8e)

### RuneLite version

Runelite version: 1.10.26.4
Launcher version: 2.6.10

### Logs

NA

swirle13:
That was so fast, thank you so much!

Would the wiki be able to be updated to reflect these changes? https://github.com/runelite/runelite/wiki/Chat-Commands

Felanbird:
Yes I'm already doing it, but it won't get merged until after the next RuneLite version since these changes aren't live yet

PR #17700, opened 2024-04-09
Add lunar chest variants for !kc command

DapperMickie:
Fixes: https://github.com/runelite/runelite/issues/17699


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -2497,6 +2497,15 @@ private static String longBossName(String boss)
 			case "vard awakened":
 				return "Vardorvis (awakened)";
 
+			// lunar chest variants
+			case "lunar chests":
+			case "perilous moons":
+			case "perilous moon":
+			case "moons of peril":
+			case "mop":
+			case "lc":
```
Nightfirecat, line 2506: I understand these are abbreviations for Moons of Peril and Lunar Chest, but I don't think anyone uses them. I'm inclined to stick to just the lines above and remove these.
```suggestion
```

Issue #17555, opened 2024-03-06
Update Ground Items despawn timer for player-dropped clue scrolls


Nightfirecat:
From [today's update post](https://secure.runescape.com/m=news/miscellaneous-quality-of-life-changes?oldschool=1):

> Clue Scrolls dropped by a player now last one hour on the ground rather than two minutes. This does not change existing behaviour for Clue Scrolls dropped by NPCs or other entities.

We will need to update the timer specifically for clue scrolls which the player drops.

Felanbird:
notes (just checking in case they had forgotten geodes or something):

beginner scroll, master scroll, beginner casket, beginner clue bottle, beginner clue nest, and easy geode all seem to have been updated correctly
![java_2024-03-07_18-32-47](https://github.com/runelite/runelite/assets/41973452/072954e0-fc38-4e1f-b0d7-2532788bcb3b)
![RuneLite_2024-03-06_08-33-43](https://github.com/runelite/runelite/assets/41973452/bc8d9a5b-070e-4c52-957b-1ae259887cd6)
geheur tested master scroll

Adam-:
fixed in 312776c5fc979365ce2e4218ee9d7b289414038a

PR #17701, opened 2024-04-09
Added despawn time for dropped clues

DapperMickie:
Fixes: https://github.com/runelite/runelite/issues/17555


Felanbird:
I was meaning to test this before but since you opened this PR I went ahead and did, and it seems the addition of the 1hour despawn _did_ bypass the 30minute instance despawn

DapperMickie:
So add the same check for instances should be fine?

PR #17710, opened 2024-04-10
overlaypanel: clear on start render

LlemonDuck:
The ordering of when the panel children are cleared may have caused an OOM crash for users today when one of the fortis-colosseum plugin overlays threw an error while rendering.

If an overlay throws while rendering, the panel component children are never cleared. After enough failed renders in a row, these elements would never be cleared, and therefore never garbage collected. Also, a failed render could render components multiple times if the exception is not thrown on a later render call.


Nightfirecat:
Fixed in 857b193179385d740806193207a7cf6f2a6c1d42

PR #17716, opened 2024-04-11
item charges: add quetzal whistle charges

Felanbird:
some1 in the discord asked for this, the dialog for getting more charges tells you that you got `SOME` charges, which is not useful

only tested with a perfect whistle


Felanbird:
Closing as I only wanted to implement this simply, I do agree we should respect the fact that you can have all types of whistle, but I won't be the one to add that functionality.

PR #17717, opened 2024-04-11
Bump PMD to version 7.0.0

crammond:
I've been playing around with upgrading RuneLite to Java 21. Part of that work would require upgrading PMD to version 7.0.0, which was a big enough change on its own that felt like it needed its own PR.

### DIFF
--- a/runelite-client/pom.xml
+++ b/runelite-client/pom.xml
@@ -448,17 +448,17 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-pmd-plugin</artifactId>
-				<version>3.16.0</version>
+				<version>3.22.0</version>
 				<dependencies>
 					<dependency>
 						<groupId>net.sourceforge.pmd</groupId>
 						<artifactId>pmd-core</artifactId>
-						<version>6.44.0</version>
+						<version>7.0.0</version>
 					</dependency>
 					<dependency>
 						<groupId>net.sourceforge.pmd</groupId>
 						<artifactId>pmd-java</artifactId>
-						<version>6.44.0</version>
+						<version>7.0.0</version>
 					</dependency>
 				</dependencies>
 				<configuration>
--- a/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextMenuInput.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextMenuInput.java
@@ -56,7 +56,7 @@ public class ChatboxTextMenuInput extends ChatboxInput implements KeyListener
 	private String title;
 
 	@Getter
-	private List<Entry> options = new ArrayList<>();
+	private final List<Entry> options = new ArrayList<>();
 
 	@Getter
 	private Runnable onClose;
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -1879,6 +1879,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 		return new int[]{text.hashCode()};
 	}
 
+	@SuppressWarnings("PMD.UnusedPrivateMethod")
 	private static WorldPoint getViggoraLocation(ClueScrollPlugin plugin)
 	{
 		int varb = plugin.getClient().getVarbitValue(Varbits.VIGGORA_LOCATION);
--- a/runelite-client/pmd-ruleset.xml
+++ b/runelite-client/pmd-ruleset.xml
@@ -44,7 +44,7 @@
 	<rule ref="category/java/bestpractices.xml/UnusedPrivateMethod">
 		<properties>
 			<property name="ignoredAnnotations"
-					  value="net.runelite.client.eventbus.Subscribe|com.google.inject.Provides"/>
+					  value="net.runelite.client.eventbus.Subscribe,com.google.inject.Provides"/>
 		</properties>
 	</rule>
 	<rule ref="category/java/bestpractices.xml/UseTryWithResources"/>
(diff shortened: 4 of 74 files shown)


Adam-:
I've looked into this before, but decided to not update because they made the immutable field check no longer work with Lombok (I guess due to make the new Java record types?). But since we use Lombok so heavily it is just a loss of functionality. I'd probably rather just remove the immutable field check than go around and annotate everything as PMD ignore though. 

I am not familiar with this pmd-compat6 dep, do you know what that is for?

crammond:
> I've looked into this before, but decided to not update because they made the immutable field check no longer work with Lombok (I guess due to the new Java record types?). But since we use Lombok so heavily it is just a loss of functionality. I'd probably rather just remove the immutable field check than go around and annotate everything as PMD ignore though.

[PMD's states their reasoning here](https://pmd.github.io/2022/11/26/PMD-6.52.0/#modified-rules). It's an unfortunate change considering use cases like Lombok. I'd be happy to modify this PR to remove that rule if there is still interest in the upgrade, though.

> I am not familiar with this pmd-compat6 dep, do you know what that is for?

This was added specifically to support `maven-pmd-plugin` until they update to version 7.0.0 themselves. There's already  a PR to do that here, https://github.com/apache/maven-pmd-plugin/pull/144, so the need for `pmd-compat6` will go away quickly.

Adam-:
I guess I don't mind the update / I am sort of indifferent to it. I see you mentioned moving to Java 21, but to do that we first have to update the launcher to 21 and then push launcher updates out. Updating the launcher version is a pretty major thing, is somewhat irreversible, and tends to uncover obscure Java bugs and also break our code in places we are depend on 11 specific functionality. It is not something I want to sink a bunch of time in unless I have a compelling reason to update.

Adam-:
https://github.com/pmd/pmd/pull/4474 looks like they made it ignore fields annotated with `@Getter` or `@Setter` in 7.0.0. Do you know why you are still having to annotate those fields as ignored?

crammond:
> I see you mentioned moving to Java 21, but to do that we first have to update the launcher to 21 and then push launcher updates out. Updating the launcher version is a pretty major thing, is somewhat irreversible, and tends to uncover obscure Java bugs and also break our code in places we are depend on 11 specific functionality. It is not something I want to sink a bunch of time in unless I have a compelling reason to update.

Thanks for your thoughts on this. I mentioned it in this PR to gauge interest, and it is understandable that you would not want to push for the upgrade without a very good reason, which I admittedly do not have. I also see in the launcher that you are currently depending on JDKs with 32bit Windows support for which there [does not seem to be a temurin offering](https://adoptium.net/temurin/releases/?arch=x86&version=21) at this time.

> they made it ignore fields annotated with `@Getter` or `@Setter` in 7.0.0. Do you know why you are still having to annotate those fields as ignored?

Ah, great catch. I was following [their migration guide](https://pmd.github.io/pmd/pmd_userdocs_migrating_to_pmd7.html) which suggested going to 6.55.0 first before 7.0.0, and I think as a result 6.55.0 reported these erroneously and I fixed them before bumping to 7.0.0. Let me redo the version bump in one go and see what changes that ends up requiring.

crammond:
I believe that is 27 less use cases of `@SuppressWarnings("PMD.ImmutableField")` than I had before, great catch!

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextMenuInput.java
```diff
@@ -56,7 +56,7 @@ private static final class Entry
 	private String title;
 
 	@Getter
-	private List<Entry> options = new ArrayList<>();
+	private final List<Entry> options = new ArrayList<>();
```
Adam-, line 59: are all of these types of changes (setting final on getter/setter fields) then also unnecessary? Ideally I'd like the minimum amount of change that keeps pmd happy.

runelite-client/pom.xml
```diff
@@ -448,17 +448,32 @@
...
 					</dependency>
 					<dependency>
 						<groupId>net.sourceforge.pmd</groupId>
 						<artifactId>pmd-java</artifactId>
-						<version>6.44.0</version>
+						<version>7.0.0</version>
+					</dependency>
+					<dependency>
+						<groupId>net.sourceforge.pmd</groupId>
+						<artifactId>pmd-javascript</artifactId>
```
Adam-, line 470: ?

runelite-client/pom.xml
```diff
@@ -448,17 +448,32 @@
...
+						<version>7.0.0</version>
+					</dependency>
+					<dependency>
+						<groupId>net.sourceforge.pmd</groupId>
+						<artifactId>pmd-jsp</artifactId>
```
Adam-, line 475: ?

Adam- (commented):
There are also several // NOPMD in the code, some of them can be removed because they have been fixed upstream I think. Please check those.

runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxTextMenuInput.java
Adam-: Is the ImmutableFieldRule flagging it? Since it really looks like it shouldn't https://github.com/pmd/pmd/commit/66ddd5617f0c7ea71ed71e13b083b0555b6ad0a0#diff-f0d84f4b43ab1c1e2fbc8b8ee09c4c0c3ee2c5a7b981d99fae749652821ec07aR64

runelite-client/pom.xml
crammond: I meant to comment or add more info in the commit message, but [maven-pmd-plugin depends on these](https://github.com/apache/maven-pmd-plugin/blob/a3ac53c954b00527d17603c8179d6555b14124d9/pom.xml#L174-L183), and I think with the API changes from 6 to 7 `pmd-compat6` needs all of these updated and available to work properly. Their docs for [upgrading PMD at runtime](https://maven.apache.org/plugins/maven-pmd-plugin/examples/upgrading-PMD-at-runtime.html) also show upgrading these packages too.

crammond:
> There are also several // NOPMD in the code, some of them can be removed because they have been fixed upstream I think. Please check those.

I was able to remove a couple of `// NOPMD`s in the code without changes. I also checked old `@SuppressWarnings("PMD.<Rule>")`s but they were still needed without other changes.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -1879,6 +1879,7 @@ public int[] getConfigKeys()
 		return new int[]{text.hashCode()};
 	}
 
+	@SuppressWarnings("PMD.UnusedPrivateMethod")
```
crammond, line 1882: Possibly https://github.com/pmd/pmd/issues/4861

runelite-client/pom.xml
Adam-: It looks like maven-pmd-plugin has updated to pmd 7 now as of Apr 18, does this mean it is no longer necessary to include all of these irrelevant dependencies?

Adam-: > It looks like maven-pmd-plugin has updated to pmd 7 now as of Apr 18, does this mean it is no longer necessary to include all of these irrelevant dependencies?

It seems like it is no longer required.

Adam-:
It seems I do not have permission to push to this PR, so I have rebased this, updated to pmd 7.2.0, and pushed it directly to master. Thanks.

PR #17728, opened 2024-04-14
mining: Add support for Cam Torum Mine

cwjoshuak:
Changes:

- Adds regex for calcified rocks so that it properly displays the mining session overlay
- Port cam torum mine plugin (https://github.com/runelite/plugin-hub/pull/5840#issuecomment-2052460395) into the Mining plugin 
- Added config section into mining plugin for cam torum

(New) default Mining Config:
image

Notes:
ObjectID for waterfall rocks is not generated yet, therefore is hardcoded.


Nightfirecat:
- Why is there a distance config? Is it useful to only highlight within 10/configurable number of tiles?
- Why do the clickbox and outline use different colors? Use one config like basically everything else does. (ex. agility clickboxes)
- We don't want the menu swapping behavior in core (that can already be achieved via the menu entry swapper plugin)
- All the ID files are up-to-date, so there's no missing IDs. 51493 is the id of a null object, so you'd need to look in `NullObjectID` to find it rather than `ObjectID`.
- I'm on the fence about whether we'd want the notification behavior. I suppose having it behind a config is fine.
- What is `inCamTorumMiningRegion` for? Do these objects appear in other places? I don't understand what undesired behavior this is preventing

cwjoshuak:
I will remove the distance config and menu swapping. I will also modify the clickbox to follow the agility plugin.

The region check is in line with all other plugins such as herbiboar, tears of guthix, MLM, etc. All these plugins have region checks. We can remove this region check for Cam Torum if unnecessary.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningRocksOverlay.java
```diff
@@ -54,25 +60,39 @@ class MiningRocksOverlay extends Overlay
...
 	}
 
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
 		List<RockRespawn> respawns = plugin.getRespawns();
-		if (respawns.isEmpty())
+		Map<TileObject, Tile> camTorumStreams = plugin.getCamTorumStreams();
+
+		if (!respawns.isEmpty())
```
Nightfirecat, line 81: I don't think either of these empty checks are necessary as the only thing the `handle` methods do are loops over these collections, which is effectively free for empty collections.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -120,6 +134,10 @@ public class MiningPlugin extends Plugin
 	private final List<RockRespawn> respawns = new ArrayList<>();
 	private boolean recentlyLoggedIn;
 
+	@Getter(AccessLevel.PACKAGE)
+	private final Map<TileObject, Tile> camTorumStreams = new HashMap<>();
```
Nightfirecat, line 138: Why is this a `Map<TileObject, Tile>` instead of a `Set<DecorativeObject>`? The only thing the tile is referenced for is its worldpoint, which can be found from `TileObject#getWorldLocation()`.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -223,6 +253,43 @@ public void onGameTick(GameTick gameTick)
...
+		}
+		camTorumLastNotificationTick = client.getTickCount();
+
+		boolean alreadyMiningStream = false;
+		boolean inCamTorumMine = false;
+		WorldPoint wp = client.getLocalPlayer().getWorldLocation();
+		for (Map.Entry<TileObject, Tile> entry : camTorumStreams.entrySet())
+		{
+			Tile tile = entry.getValue();
+			int distanceToStream = Math.abs(wp.getX() - tile.getWorldLocation().getX()) + Math.abs(wp.getY() - tile.getWorldLocation().getY());
```
Nightfirecat, line 275: This formula is rather confusing, since it resembles an incorrectly-implemented version of  `WorldPoint#distanceTo(WorldPoint other)`. A comment explaining its purpose would help clarify this.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -223,6 +253,43 @@ public void onGameTick(GameTick gameTick)
...
+			if (distanceToStream <= 10)
+			{
+				inCamTorumMine = true;
+			}
+			if (distanceToStream == 1 && pickaxe != null && pickaxe.matchesMiningAnimation(client.getLocalPlayer()))
+			{ // if stream is directly adjacent to player and player is in mining animation
```
Nightfirecat, line 281: This comment should not be needed if `distanceToStream` is better explained.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningRocksOverlay.java
```diff
@@ -27,17 +27,23 @@
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.awt.Shape;
 import java.time.Instant;
 import java.util.List;
+import java.util.Map;
 import javax.inject.Inject;
+
```
Nightfirecat, line 35: Remove empty line (will commit via GitHub interface)
```suggestion
```

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningRocksOverlay.java
```diff
@@ -116,13 +136,28 @@ public Dimension render(Graphics2D graphics)
...
+	}
+
+	public void handleCamTorumWaterStreamHighlights(Graphics2D graphics, Map<TileObject, Tile> streams)
+	{
+		streams.forEach((object, tile) ->
+		{
+			Shape objectClickbox = object.getClickbox();
+			if (objectClickbox != null)
+			{
+				Color overlayColor = config.camTorumWaterOverlayColor();
```
Nightfirecat, line 148: This fetches config N times per frame, instead of once per frame. Define the color var outside of the loop instead.

runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningRocksOverlay.java
```diff
@@ -116,13 +136,28 @@ public Dimension render(Graphics2D graphics)
...
 		}
-		return null;
+	}
+
+	public void handleCamTorumWaterStreamHighlights(Graphics2D graphics, Map<TileObject, Tile> streams)
+	{
+		streams.forEach((object, tile) ->
+		{
+			Shape objectClickbox = object.getClickbox();
+			if (objectClickbox != null)
```
Nightfirecat, line 146: Use a guard clause instead:

```java
if (objectClickbox == null)
{
	return;
}

// continue rendering here
```

runelite-client/src/test/java/net/runelite/client/plugins/mining/MiningPluginTest.java
```diff
@@ -149,4 +165,30 @@ private void testMessage(String gameMessage)
...
+	@Test
+	public void testCamTorumWaterNotify()
+	{
+		when(client.getLocalPlayer()).thenReturn(mock(Player.class));
+		when(client.getLocalPlayer().getWorldLocation()).thenReturn(new WorldPoint(1501, 9538, 1));
+
+		GameStateChanged gameStateChanged = new GameStateChanged();
+		gameStateChanged.setGameState(GameState.LOGGED_IN);
+		miningPlugin.onGameStateChanged(gameStateChanged);
+		when(miningConfig.camTorumNotifyWaterSpawn()).thenReturn(true);
```
Nightfirecat, line 178: Is this intending to test enabling the config after logging in? If so, some kind of `verify()` call should be made before changing the config value.

Nightfirecat:
> The region check is in line with all other plugins such as herbiboar, tears of guthix, MLM, etc. All these plugins have region checks. We can remove this region check for Cam Torum if unnecessary.

The herbiboar region check is used for checking menu clicks which would start tracking--probably not strictly necessary since the objects which can be clicked should have unique object IDs--but is useful. The ToG region check is probably not needed. The MLM region checks are useful as certain overlays (the panels showing number of ores mined, etc.) should not be displayed outside of MLM; region checks are the only reasonable way to accomplish that. None of these cases necessarily apply here, since the object IDs used are new and only used in this area, so we wouldn't encounter them spawning elsewhere in the game. I think it's safe to remove it like you have.

> I've hardcoded a distance of 10 to the closest stream for the notification. Because the world region of the mine spans partly into a public area (see last screenshot), players may receive notifications if this is enabled and they are standing at the edge of the region. In my testing, 10 is the sweet spot for receiving notifications for one section of the mine.

From your screenshot, that looks like it would mean a stream spawning on one side wouldn't notify you if you were on the other side? (visually, the left-most and right-most miners in your screenshot are >10 tiles from each other) I think this would be better solved by defining a WorldArea which encompasses the mining area, and only sending notifications if the player is standing in that area.

PR #17734, opened 2024-04-16
chat commands: add commonly used aliases

YvesW:
This PR...
- Adds the ``temp`` alias for ``Tempoross``.
  - Currently Tempoross only has somewhat obscure aliases like ``fishtodt``. Everyone I know refers to it as ``temp``.
- Adds the ``graar`` alias for ``General Graardor``.
  - This is probably the second most used alias for him (after ``bandos``). Feel free to correct me if I'm crazy.
- Removes the ``gauntlet`` alias for ``Gauntlet``.
  - An alias that's the same as the longBossName except for its capitalization is superfluous.


Skretzo:
I looked at the wiki page visit analytics for a few aliases:
|Alias|Views|
|:-|-:|
|`bando`|298|
|`bandos`|17268|
|`graardor`|3542|
|`graar`|106|
|`garage door`|206|

so I don't think `graar` is needed.

PR #17755, opened 2024-04-18
spellbook: reinstate spellbook behavior on shutdown

Macweese:
This PR is intended to fix issues with the spellbook plugin, which include:
- ~~clearing the plugin's menuoptions from the interface tab on shutdown~~
- clearing spells' modified ops to that of their original state
- restoring spellbook behavior on shutdown whilst in reordering state (dragging, opacity, etc)

To my understanding this issue happens in part because the script callback doesn't get run on shutdown.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/spellbook/SpellbookPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/spellbook/SpellbookPlugin.java
@@ -107,6 +107,9 @@ public class SpellbookPlugin extends Plugin
 
 	private boolean reordering;
 
+	private int spellbookEnumId;
+	private int[] spells;
+
 	@Provides
 	SpellbookConfig getConfig(ConfigManager configManager)
 	{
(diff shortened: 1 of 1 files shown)


Adam-:
I do not understand what you are trying to fix. For example, `clearing the plugin's menuoptions from the interface tab on shutdown` The plugin's menu options get removed from the interface tab on shutdown. Here is a video showing that:

https://github.com/runelite/runelite/assets/309739/f289e0b6-a775-47b7-94e8-37175dfb0a2a

Adam-:
> That's strange, on my machine the Enable / Disable reordering option remained even after shutdown, hence my itention to fix it... are the other issues also non-existant? For me, the following issues were encountered:
> 
>     * disabling the plugin whilst in reordering mode would still permit dragging spells
> 
>     * re-enabling the plugin after disabling whilst in reordering mode would put the plugin into reordering mode, (as if continuing from left off, instead of resetting)
> 
>     * hide/unhide actions persist after shutdown

Try reproducing it on the release.

PR #17765, opened 2024-04-20
menu-entry-swapper: support teleport submenus on hunter cape

typical-whack:
Support teleport submenus on hunter cape in the menu entry swapper plugin.

image


Adam-:
We could probably get away with the hunter cape, since it is a subset of the max cape, but I do not think we can for jewellery

image

Felanbird:
this needs to also update the setting description

Nightfirecat:
No. The game jam content featured is not live in the game, and is not guaranteed to be added. Chances are if it is added, we would just remove our feature since it would not be needed.

PR #17775, opened 2024-04-21
TextComponent: Fix bug where <col> tags could cause an exception to be thrown.

geheur:
The code expects all parts of the colored string to start with a <col> tag, so this fix adds one if it isn't present. You can easily see the bug by setting a clue note to `something<col=ff0000>something`.


Adam-:
fixed in 43588dd7e28519ee68cc294e3b78968e5be0e318

PR #17794, opened 2024-04-26
ground items: Add hide others loot feature

Winter:
Adds a toggleable option to hide other players loot in the ground items plugin. 

Useful for ironmen who can't pickup other players drops.

In the past the only way to differentiate ground items was that there was no loot timer but since 312776c, other players loot shows despawn times now.

Behaves similiar to the `Only show loot` option with respect to the area being reloaded and the plugin thinking it's another players loot drop.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -362,10 +362,21 @@ public interface GroundItemsConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "hideOthersLoot",
+		name = "Hide others loot",
+		description = "Hides items dropped by or intended for other players",
+		position = 25
+	)
+	default boolean hideOthersLoot()
+	{
+		return false;
+	}
+
+	@ConfigItem(
 		keyName = "onlyShowLoot",
 		name = "Only show loot",
 		description = "Only shows drops from NPCs and players",
-		position = 25
+		position = 26
 	)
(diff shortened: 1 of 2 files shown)


Adam-:
I am adding item ownership types to the API soon, and then renaming the "Only show loot" to only show items which are owned by you, which should solve this.

Winter:
> I am adding item ownership types to the API soon, and then renaming the "Only show loot" to only show items which are owned by you, which should solve this.

Thanks for the reply Adam. I'll close this PR in that case.

Adam-:
Is there something you are confused about?

Winter:
> Is there something you are confused about?

Oh! My mistake.. I misread your statement. When you said "which should solve this" I thought you were referring to the PR as a whole, not the issue with the the client losing track of the item ownership. 

I'll re-open the PR and we can re-evaluate once the item ownership is resolved. 

Is there a issue/PR for tracking the item ownership?

Winter:
> I'm casually observing this conversation because my Discord was left on #github so I feel compelled to jump in for no reason and make a guess here:
> 
> A) Your initial understanding of Adam's comment was correct and B) the "confused" emoji was used not because you were "confused" by anything but because it made a :/ face, as in, "guess I wasted my time"
> 
> Fair to say?

Just updating the thread as per what transpired in the Discord in-case anyone wanders into this PR in the future; 

I had mistakingly reacted with a "confusing" emoji on my phone and then I've read Adam's response asking if I was confused about something and then I've re-read his message to mean that there was going to be a later change that might immediately resolve the reloading issue.

Sorry for an inconvenience caused!

I'll close this PR again.

PR #17809, opened 2024-05-04
AntiDrag: Ask key remapper to start before antidrag

connor-richards:
On startup of antidrag, check for the key remapper and start it if possible. Fixes an issue where the antidrag "disable on control press" does not work when ctrl is rebound, requiring the user to toggle antidrag plugin off and on again.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/antidrag/AntiDragPlugin.java
@@ -45,6 +45,8 @@ import net.runelite.client.input.KeyListener;
 import net.runelite.client.input.KeyManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.plugins.PluginInstantiationException;
+import net.runelite.client.plugins.PluginManager;
 
 @PluginDescriptor(
 	name = "Anti Drag",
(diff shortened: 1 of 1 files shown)


Adam-:
Can you describe in more detail what the issue is you are trying to fix?

connor-richards:
> Can you describe in more detail what the issue is you are trying to fix?

Of course!
If you use the feature in the antidrag plugin that cancels the timeout when you press ctrl, and use the key remapper to remap the control key to a letter, then the remapped key press will send the wrong keycode to the antidrag plugin, resulting in the "if keycode == control" statement failing. In my case, I remap control to "n", and I was seeing that the antidrag plugin was getting the "n" keycode rather than the control key.
The antidrag plugin seems to always start before the key remapper, I assume this is related to alphabetical order of their names. This order of startup results in the antidrag plugin registering for keypress events before the key remapper does, and so we see the wrong keypress come through for the control key. As a user, my workaround has been to toggle the antidrag plugin off and back on every time I start the client. This forces the antidrag plugin to re-register for those key presses and get the right keycode.
The fix I have in my pull request simply starts the key remapper before antidrag so that they register themselves with the key manager in a guaranteed order. If you feel that this problem is better addressed in the KeyManager, please let me know. I am new to your code, so what I've suggested may just be a band-aid fix.
Thanks!

connor-richards:
@Adam- Any update on this?

Adam-:
I've not had time to look into this yet.

connor-richards:
@Adam- is this something you are not comfortable merging in? Thanks.

Adam-:
Having plugins start/stop each other is pretty bad UX, and not something we do anywhere else, and usually the problem can be better solved another way.

connor-richards:
Thanks for the update!

PR #17813, opened 2024-05-06
Add Civitas Teleport spell (disabled) sprite to SpriteID enum

jarromie:
`SpriteID.SPELL_CIVITAS_ILLA_FORTIS_TELEPORT_DISABLED = 417;`

image


Nightfirecat:
We don't typically add sprite definitions to that class unless they are used, as we would have no way to notice if the sprite changed from under us. If this is for a hub plugin, you can use a plain int for whatever methods you would use this definition for.

Issue #17799, opened 2024-04-27
Idle Notifier: Custom Notifications missing for non-checkbox enabled notifications


evilgenious448:
### Type

Other

### Operating System

Windows

### OS Version

10

### Bug description

The custom notification feature is only applied to notifications within Idle Notifier that are enabled by a checkbox, while those enabled by a numbered value > 0 do not receive this functionality.

It would be great if this was applied to all notifications, not just to those enabled by a checkbox.

Key missing ones include:
Hitpoints threshold, Prayer Threshold, Low Energy Threshold, High Energy Threshold, Oxygen Threshold and Spec Threshold.

### Screenshots or videos

![image](https://github.com/runelite/runelite/assets/32331499/07934992-c040-4353-827b-4e62ad332b05)

### RuneLite version

RL version: 1.10.28
Launcher Version 2.6.1

DapperMickie:
I'm thinking of 2 solutions, 1 which requires almost no work and 1 which requires a bit of work.

The "no-work" solution would be to just add a checkbox for each of these notifications, although that might be filling up the config a bit.

The other one would be to have a Notification<T> using the same attributes that already exist to make the config look the same, it would just have a cog to the left of the input field.

Adam-:
There is also all of the impling notifications which are sort of in the same boat. I'm not sure the best solution for it.

DapperMickie:
I feel like the Notification<T> would be the 'cleanest' solution, but it also raises some questions. 

Like how do we determine on/off state for these, I know they are currently done programmatically by just using the value, but would we want to have that same functionality?
Do we add the on/off checkbox in the panel behind the cog wheel?

DapperMickie:
If we were to 'hide' the value field behind the cog wheel, we could simplify the ImplingMode enum to:
```java
enum ImplingMode
{
	HIGHLIGHT,
	NOTIFY
}
```
And use the `Notification.OFF` as default/replacement of `ImplingMode.NONE`

DapperMickie:
If we were to do that, I feel like we should always want to show the cog wheel though.
To make it a bit more obvious to people that there are more configuration options for this config.

DapperMickie:
Another option would be to translate a `Notification<T>` into a config section, with the checkbox on top and the value field under it.

DapperMickie:
@Adam- have you had the chance to think about this one?

PR #17827, opened 2024-05-09
Add customizable notifications for implings

DapperMickie:
As discussed here:
https://github.com/runelite/runelite/issues/17799#issuecomment-2081159199


Adam-:
I don't think this makes sense because there is still impling mode NOTIFY despite a separate notification option existing. This means you could enable the notification but still not receive notifications due to not setting the *other setting* to notify. Also the comment you have linked to is just me saying I don't know how this should work.

DapperMickie:
Added a checkbox for highlighting the implings.
Added a Notification for implings (instead of the customizableNotification that didn't make much sense).
Added a migration flow for configs.

PR #17830, opened 2024-05-10
Enable key remapping in dialogs with a "Click to continue" option

### LINKED ISSUE #16274: Hot key remap doesnt work in chatbox

animatic-ps:
Closes #16274

Dialogs with a "Click to continue" option don't accept any input other than a mouse click or Space/re-mapped space key.
Therefore it is safe to enable F-key remapping in these types of dialogs.

This MR adds 3 IDs for the "Click to Continue" message NPC, Player, and Sprite dialogs (like received clue scroll dialogs) and checks if one of these is visible while deciding whether to enact a remapped F key.

I think this is safer than the existing MR against this issue as it won't interfere with any dialogs that are expecting alphanumeric input.

This will be really useful for being able to use remapped F-keys to open inventory/gear to access teleports while the received clue dialog is still open.

Short clip of it in action working on 3 types of "Click to continue" dialogs but still remaining disabled for option dialogs:

https://github.com/runelite/runelite/assets/107762963/e7d4080f-0053-4e41-8c64-d59a9031e202

### DIFF
--- a/runelite-api/src/main/interfaces/interfaces.toml
+++ b/runelite-api/src/main/interfaces/interfaces.toml
@@ -258,6 +258,7 @@ sprite2=3
 id=231
 head_model=2
 name=4
+continue=5
 text=6
 
 [dialog_option]
@@ -266,10 +267,12 @@ options=1
 
 [dialog_player]
 id=217
+continue=5
 text=6
 
 [dialog_sprite]
 id=193
+continue=0
 sprite=1
 text=2
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -99,7 +99,7 @@ class KeyRemappingListener implements KeyListener
 			// In addition to the above checks, the F-key remapping shouldn't
 			// activate when dialogs are open which listen for number keys
 			// to select options
-			if (config.fkeyRemap() && !plugin.isDialogOpen())
+			if (config.fkeyRemap() && (plugin.isClickToContinueVisible() || !plugin.isDialogOpen()))
 			{
 				if (config.f1().matches(e))
 				{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingPlugin.java
@@ -163,6 +163,13 @@ public class KeyRemappingPlugin extends Plugin
 		return client.getWidget(ComponentID.DIALOG_OPTION_OPTIONS) != null;
 	}
 
+	boolean isClickToContinueVisible()
+	{
+		return client.getWidget(ComponentID.DIALOG_PLAYER_CONTINUE) != null ||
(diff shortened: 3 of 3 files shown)


raiyni:
relevant comment https://github.com/runelite/runelite/issues/16274

animatic-ps:
> relevant comment #16274

Yeah I did see that comment. Remapping being disabled when a dialog expects input or a numeric option selection makes sense; dialogs with "click to continue" have neither so will be safe to remap on.

Adam-:
I think it might be better to include all of the cases when it should remap instead of trying to exclude when it should not. Which I think is: `isOptionsDialogOpen() || INPUT_TYPE != NONE`

animatic-ps:
> I think it might be better to include all of the cases when it should not remap instead of trying to exclude when it should. Which I think is: `isOptionsDialogOpen() || INPUT_TYPE != NONE`

Ah cheers, didn't know about that input type var, I'll check it out tomorrow

animatic-ps:
> I think it might be better to include all of the cases when it should not remap instead of trying to exclude when it should. Which I think is: `isOptionsDialogOpen() || INPUT_TYPE != NONE`

I updated the logic I _think_ it works the way you were implying here (check for cases where you should NOT remap), I couldn't find any variable or logic tied directly to input but I think options dialogs, decant, make-x and input dialogs (which are already not remapped on) covers everything 🤞

geheur:
This could noticeably impact people pre-selecting dialog inputs.

e.g. if you have 1 remapped to an fkey:
talk to a banker
press space and then immediately hold 1
this change makes it so that the dialog option that comes up next is not selected.

animatic-ps:
> This could noticeably impact people pre-selecting dialog inputs.
> 
> e.g. if you have 1 remapped to an fkey: talk to a banker press space and then immediately hold 1 this change makes it so that the dialog option that comes up next is not selected, because the dialog options are not yet visible when you pressed 1.
> 
> This of course only matters if you're holding the key down, not if you're spamming it.

Hmm good point; I'm struggling to think of a way around that while still enabling non-input dialogs to allow re-mapping. I personally find it more annoying that I can't access my F keys while I have a dialog open (in particular the dialog that says you've received a new clue/casket).

Worth noting what you describe is already sort of an issue with the plugin for things like blast furnace and other make-x interfaces (if you hold 1 down before the dialog pops up and you have 1 re-mapped then it won't receive the input). Actually things like these, as well as accidentally changing tab by holding a re-mapped number too long after a dialog closes are the reasons why I never map the number keys to F keys with this plugin anymore because it causes too many annoyances.

I do think the issue you describe is reasonably niche and it wouldn't take long for a player to learn to just wait for the continue to go away before holding the key down (doesn't actually save any time regardless)

---

If you think this is a blocker for this pull request then it might just have to be closed (unless you can think of a solution I can try out!)

 It's a shame because its something that annoys me daily not being able to F-key with simple dialogs open but recognise my use case isn't the only one

Marcarrian:
Is there any news on this? I would really like to see this implemented in some form. I personally use QWERT instead of F keys and it would greatly improve my gameplay. Even if it was configurable instead of using complex logic, that would be a huge help.

PR #17851, opened 2024-05-16
changed scoutOverlayInRaid and ccDisplay to true by default

bobmclovin:


bobmclovin:
idk what im doing but those two checkboxes should be enabled by default :)

LlemonDuck:
We decided to change the default for `scoutOverlayInRaid` but not the other. Thanks for the PR!

PR #17855, opened 2024-05-17
Make use of CommandExecuted case insensitive

SuperNerdEric:


runelite-client/src/main/java/net/runelite/client/plugins/devtools/DevToolsPlugin.java
```diff
@@ -316,7 +316,7 @@ public void onCommandExecuted(CommandExecuted commandExecuted)
 	{
 		String[] args = commandExecuted.getArguments();
 
-		switch (commandExecuted.getCommand())
+		switch (commandExecuted.getCommand().toLowerCase())
```
SuperNerdEric, line 319: Not sure of the preferred way to do this one since it's a switch statement

PR #17866, opened 2024-05-20
refactor: use jagex naming conventions where known for some cache config types

ultraviolet-jordan:


Adam-:
While I appreciate trying to use the real names for things, this would probably introduce a lot of confusion for us (and consumers of this such as the wiki) as eg. noone refers to items as "objects". We also have the client API which consistently refers to items and objects instead of objects and locs and this would make the cache lib inconsistent with it. This also is a breaking change to our cache updater & http-service which depend on this. So overall I think the cons outweigh the pros on this, but thanks for the PR.

PR #17888, opened 2024-05-27
Bring up the overlay as soon as the smelting animation starts.

Speeddymon:
This is in prep for adding support for smelting jewellery to the overlay.


Adam-:
I am not too sure why this is required, if it is needed for tracking jewellery crafting then include it as a commit in a PR for that.

Speeddymon:
It's a separate feature. If I can't figure out the jewellery crafting (I'm a noob with Java but taking some course) then I'd like this to go in anyway.

PR #17904, opened 2024-05-31
spec counter: add Seercull

rdutta:
Adds seercull spec to spec counter


Adam-:
added in 7f1660f002a970fd31e38e6d0b2d0af12bfc3aec (had to rebase it due to the elder maul changes)

PR #17907, opened 2024-05-31
[Boosts Information plugin] Allow alerting on fully drained stats

autumn-smellegy:
Currently with this plugin you cannot get notified when your stats fully drain, only when the drain to level + 1. Setting the threshold to 0 disables the feature. This new param will allow the boost threshold to be set to 0 without it disabling the feature, meaning you can get notified when a stat returns to its base level.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
@@ -127,7 +127,7 @@ public interface BoostsConfig extends Config
 	@ConfigItem(
 		keyName = "boostThreshold",
 		name = "Boost threshold",
-		description = "The threshold at which boosted levels will be displayed in a different color. A value of 0 will disable the feature.",
+		description = "Number of levels above your base level at which boosted levels will be displayed in a different color.",
 		position = 12
 	)
 	default int boostThreshold()
@@ -138,11 +138,11 @@ public interface BoostsConfig extends Config
 	@ConfigItem(
 		keyName = "notifyOnBoost",
 		name = "Notify on boost threshold",
-		description = "Configures whether or not a notification will be sent for boosted stats.",
+		description = "Configures whether or not a notification will be sent when boosted stats drain to the boost threshold.",
 		position = 13
 	)
 	default Notification notifyOnBoost()
 	{
-		return Notification.ON;
+		return Notification.OFF;
 	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsPlugin.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import static net.runelite.api.MenuAction.RUNELITE_OVERLAY_CONFIG;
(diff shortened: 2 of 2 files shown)


Adam-:
I don't think this makes sense because it requires checking two different checkboxes (Enable boost threshold and Notify on boost threshold) to enable the notification.

autumn-smellegy:
> I don't think this makes sense because it requires checking two different checkboxes (Enable boost threshold and Notify on boost threshold) to enable the notification.

@Adam- 
They serve distinct purposes - enable boost threshold causes the color of the boosted level to change, and notify on boost threshold sends a notification when the boost threshold is hit.

Can you think of a better way to allow notifications when boosts fully drain? It's a pretty useful piece of functionality that is missing from this plugin.

Adam-:
You could remove the != check, but would need something to handle config migrations I think. When the != 0 check was made there was no other way to disable the notification

autumn-smellegy:
> You could remove the != check, but would need something to handle config migrations I think. When the != 0 check was made there was no other way to disable the notification

@Adam- 
Oh that does seem like a simpler solution :) I don't know what you mean by config migrations though

autumn-smellegy:
Update the PR based on your feedback @Adam-

Adam-:
I am concerned about blanket enabling notifications for everyone like this. Since it is a behavior change where it previously was not sending notifications but now it is.

autumn-smellegy:
> I am concerned about blanket enabling notifications for everyone like this. Since it is a behavior change where it previously was not sending notifications but now it is.

@Adam-
What if we change the default value of the notification option to disabled?

Adam-:
Changing the default only affects new profiles/new users. So that would be good to do, but additionally I think you would have to migrate the config by disabling the notification of boost threshold == 0

autumn-smellegy:
> Changing the default only affects new profiles/new users. So that would be good to do, but additionally I think you would have to migrate the config by disabling the notification if boost threshold == 0

@Adam- 
So you mean something that would do a one-off check if boost threshold == 0 and disable the notification? But then after that if the notification is re-enabled it wouldn't continue to do that? I'm not exactly sure how to implement that.

Adam-:
Here is an example from invtags https://github.com/runelite/runelite/commit/65e902699ce889a5dafc490586d02b039901fd40#diff-863393ac8b89325de9799409ccbb395a30cae2d4e7a6ecbf564234ff3e7967d0R137

autumn-smellegy:
> Here is an example from invtags [65e9026#diff-863393ac8b89325de9799409ccbb395a30cae2d4e7a6ecbf564234ff3e7967d0R137](https://github.com/runelite/runelite/commit/65e902699ce889a5dafc490586d02b039901fd40#diff-863393ac8b89325de9799409ccbb395a30cae2d4e7a6ecbf564234ff3e7967d0R137)

@Adam- 
Ok thanks. I've added this change and got it to pass the tests. Not entirely sure it's exactly how it should be though.

autumn-smellegy:
What comes to mind now is that if a user doesn't have boostThreshold set to 0, then the first time they set it to 0 and restart the client their notification setting will be turned off. This might not be a desired outcome. Should we just mark the config as migrated regardless of the boostThreshold setting, but only disable the notification if boostThreshold is set to 0 when we first check?

YvesW:
Just migrate the config, set a key so you know it's migrated, after that never migrate the config again because you have set the key. Seems like you are already doing that? There is no ``if (boostThreshold == 0)`` check for ``configManager.setConfiguration("boosts", "migrated", "1");`` so you cannot run into the problem you're thinking of?

autumn-smellegy:
> Just migrate the config, set a key so you know it's migrated, after that never migrate the config again because you have set the key. Seems like you are already doing that? There is no `if (boostThreshold == 0)` check for `configManager.setConfiguration("boosts", "migrated", "1");` so you cannot run into the problem you're thinking of?

Oh lol you're right. The code I have now already does what I thought should happen in my above comment. I just moved the debug line about changing the setting inside that conditional.

Issue #16242, opened 2023-02-03
All hotkey-able actions render chat unusable if window focus is lost directly after press


1Defence:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

This bug is most noticeable when using multiple clients and has been present for years I've noticed all hot keyable actions using the HotkeyListener will render the entire chat unusable until the key in question is pressed again.

It seems to be replicable by pressing a hotkey and clicking on a different window(a different client in my case) to lose focus.

https://github.com/runelite/runelite/blob/4b41ae64e3fa450ae0eff3fe6413c36a0b658b8e/runelite-client/src/main/java/net/runelite/client/util/HotkeyListener.java#L67

I believe this never gets set to false if focus is lost while the key gets pressed, so potentially releasing it upon focus loss would solve this.

### RuneLite version

RuneLite version: 1.9.9
Launcher version: 2.4.2

PR #17912, opened 2024-06-02
KeyListener: Release held keys when client focus is lost

1Defence:
Adds overridable focusLost to keyListener and implements it in both ClientUI and KeyManager.

Resolves https://github.com/runelite/runelite/issues/16242

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/input/KeyListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/input/KeyListener.java
@@ -30,4 +30,8 @@ public interface KeyListener extends java.awt.event.KeyListener
 	{
 		return false;
 	}
+
+	default void focusLost()
+	{
+	}
 }
--- a/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
@@ -58,10 +58,10 @@ import java.awt.desktop.QuitStrategy;
 import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
 import java.awt.event.KeyEvent;
-import java.awt.event.KeyListener;
 import java.awt.event.MouseEvent;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.awt.event.WindowFocusListener;
 import java.awt.image.BufferedImage;
 import java.time.Duration;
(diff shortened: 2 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
```diff
@@ -377,6 +378,22 @@ public void windowClosing(WindowEvent event)
 					}
 				}
 			});
+
+			frame.addWindowFocusListener(new WindowFocusListener()
```
Adam-, line 382: this is registered before keyListeners is setup, is that a problem?

runelite-client/src/main/java/net/runelite/client/util/HotkeyListener.java
```diff
@@ -49,6 +49,13 @@ public boolean isEnabledOnLoginScreen()
 		return enabledOnLoginScreen;
 	}
 
+	@Override
+	public void focusLost()
+	{
+		isPressed = false;
+		isConsumingTyped = false;
```
Adam-, line 56: this isn't firing the hotkeyReleased() at all, so it wouldn't cause plugins to release their hotkey without them overriding focusLost()?

runelite-client/src/main/java/net/runelite/client/input/KeyListener.java
```diff
@@ -30,4 +30,8 @@ default boolean isEnabledOnLoginScreen()
 	{
 		return false;
 	}
+	default void focusLost()
```
Adam-, line 33: \n

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/util/HotkeyListener.java
1Defence: > this isn't firing the hotkeyReleased() at all, so it wouldn't cause plugins to release their hotkey without them overriding focusLost()?

you'd mentioned spurious depress events, maybe I misunderstood but I'd assumed that was related to the hotkeyReleased();

runelite-client/src/main/java/net/runelite/client/ui/ClientUI.java
1Defence: if its initially null rather than empty it could, I've fixed that in the recent commit as it makes more sense regardless

PR #17919, opened 2024-06-03
Update to SpecialCounterPlugin to support Tekton spec misses

jaayjee:
Adding a condition to call specialAttackHit when a spec from Elder Maul, DWH or BGS is missed on Tekton as the defence is drained by 5% on a miss from EM/DWH and -10 defence on a BGS miss.

Sending the 0 will support the Party Defense Tracker Plugin which is currently already prepared to handle this condition.


jaayjee:
@Adam- https://oldschool.runescape.wiki/w/Tekton#Strategy
I also have tested this myself with monster examine for all 3 weapons.
https://secure.runescape.com/m=news/easter-2017-holiday-event?oldschool=1
https://secure.runescape.com/m=news/wilderness-slayer?oldschool=1

jaayjee:
Removed the 0 counter being added to the infobox as discussed with @Adam- on Discord

Adam-:
added in #17950  / c099fdf2c470e3892ed57689f1d93972d028f415

Issue #17890, opened 2024-05-28
Idle Notifier 5 Hour 40 Minute unwanted notification bug


WonderfulDeath:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Regarding the "Idle Notifier" plugin and it's bug-every time the log out timer that i have enabled to replace the report button reaches 5 hours and 40 minutes in the current session played i get a notification from specifically this plugin(i don't think this is related to the plugin that replaces the report button, it's just how i found out about this bug), i do not get this unwanted notification when i have this plugin disabled, this notification happens regardless if have the "Idle Logout Notifications" enabled or not- as long as the plugin is turned on i will get this notification at 5 hours 40 minutes played every single time. for more information i also enabled chatbox notifications that give explanations for what it's notifying me for and it's something along the lines of "you're going to be logged out!" notification in the chatbox so i suspect it's related to the idle logout notifications part of the plugin always being active, i even enabled the idle logout notification plugin but turned all of it off in the little cog next to it so it wouldn't notify me but that hasn't worked either- it will notify me regardless as long as the plugin is enabled

### Screenshots or videos

i could give this in an addendum- but i'm unsure of how helpful it would be- also i would need at least 5 hours and 40 minutes to demonstrate it in a screenshot

### RuneLite version

RuneLite Version: 1.10.31.2 Launcher Version: 2.6.7

### Logs

[client.log](https://github.com/runelite/runelite/files/15463246/client.log)
[launcher.log](https://github.com/runelite/runelite/files/15463251/launcher.log)

Speeddymon:
Hello, could you clear out the client.log and launcher.log files completely and then run a session until the sound happens, then exit the client and upload the logs again? The current logs attached span over a month, and it's impossible to know with all of the profile switching and errors from other plugins when the issue occurred to try to narrow down a source.

As a suggestion, create a profile will all other plugins except for this one turned off and use that for a session to see if you can reproduce the issue with only this plugin enabled.

Demonstrating the issue by turning off this plugin only reveals part of the problem. Because this plugin generates chat messages, its possible another plugin is picking up the chat message near the 6 hour mark from this plugin and playing a sound as a result. Turning off everything but this and then still being able to reproduce the issue demonstrates that this plugin is the actual source of the sound.

Hope this helps!

Alexsuperfly:
are you saying you want the 6hr logout notification to be behind a config option so you can turn it off?

[...]

PR #17933, opened 2024-06-06
Add 6 hour log notification configuration

tylerwgrass:
Addresses #17890

### Overview
Adds a configurable option to the Idle Notifier plugin for how many minutes prior to the 6 hour logout the user wants to be notified, with a value of 0 disabling the notification.

For example, a value of 20 will alert the user 20 minutes before forced logout (or 5 hours 40 mins of login time).

### Testing
- Set value to 359 mins and verified alert appeared after 1 min logged in


tylerwgrass:
alternative added in e4d42a625cd8fbcefe147cd9e19478e793abc8a3

PR #17945, opened 2024-06-10
gpu: regions: add Keldagrim, Abyss, Temple of Ikov

cdfisher:
Hide unrelated maps for Keldagrim, as mentioned in #support on Discord this morning.
Additionally adds GPU regions for the Abyssal Nexus and Abyss, Abyssal Area (fairy ring DIP), and the Temple of Ikov.


cdfisher:
Just noting:
Looking at [Mej's map](https://mejrs.github.io/?m=-1&z=2&p=0&x=2915&y=10207&layer=Grid), it seems like 44, 160 and 45, 160 would be included in the region for Keldagrim, but I'm not even seeing them load in at all in 1.10.31.3, whether GPU is on (with or without "Hide unrelated maps" enabled) or off.

Simply having `R 44 158 45 160` was causing things from the edges of the tunnels in 43, 160 to load, so it may be possible that something like

```
// keldagrim
n
R 44 158 45 159
r 45 160
m 44 160
C 4 0 7 7
```

would be better, but since I'm less familiar with that part of the regions format I'm 1) not certain it's right and 2) not sure if there would a better way to accomplish that since I'm not actually able to get those areas to load in as it is.

Adam-:
> Just noting: Looking at [Mej's map](https://mejrs.github.io/?m=-1&z=2&p=0&x=2915&y=10207&layer=Grid), it seems like 44, 160 and 45, 160 would be included in the region for Keldagrim, but I'm not even seeing them load in at all in 1.10.31.3, whether GPU is on (with or without "Hide unrelated maps" enabled) or off.

You might not have the xteas cached for them? Try to get closer so they show up in getRegions()

cdfisher:
Turns out looking at the map for the right game would have been helpful.

LlemonDuck (approved):

PR #17948, opened 2024-06-11
MapImageDumper: Use world map color blending

MESLewis:
Use full HSL ground blending to produce map images that closely resemble the in game world map and [the official world map by Jagex](https://www.runescape.com/oldschool/world-map). Also correct the scaling of sprites and tile shapes to match those sources.

Old color blending:
old color blending
 
Jagex world map image:
Jagex world map

In game world map:
in game

PR color blending:
new color blending

### DIFF
--- a/cache/src/main/java/net/runelite/cache/MapImageDumper.java
+++ b/cache/src/main/java/net/runelite/cache/MapImageDumper.java
@@ -79,10 +79,7 @@ public class MapImageDumper
 	private static final int MAP_SCALE = 4; // this squared is the number of pixels per map square
 	private static final int BLEND = 5; // number of surrounding tiles for ground blending
 
-	private static int[] colorPalette = JagexColor.createPalette(JagexColor.BRIGHTNESS_MIN);
-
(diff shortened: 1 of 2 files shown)


cache/src/main/java/net/runelite/cache/MapImageDumper.java
```diff
@@ -639,6 +633,661 @@ else if (shape == 1)
...
+	{
+		TILE_SHAPE_2D = new byte[8][4][];
+		genTileShapes1();
+		genTileShapes2();
+		genTileShapes3();
+		genTileShapes4();
+		genTileShapes5();
+		genTileShapes6();
+		genTileShapes7();
+		genTileShapes8();
```
allenkinzalow, line 683: These methods all look pretty similar. Is there a way to combine them based on the `TILE_SHAPE_2D` index?

Pazaz: Jagex has the logic broken out the same way, it's used to draw the smooth edges of overlay shapes (think paths and edges of islands). Looks to be taken right out of the client but stored into that result array so it isn't computed over and over.

In the reaally old mapview applet (2004-2008) it looks like
```
drawMap(...) {
    if overlay {
        if shape {
            drawSmoothEdges(...);
        }
    }
}

drawSmoothEdges(...) {
    // inline convertTileRotation() equivalent
    // followed by inline genTileShapes() that draw directly to the image buffer
}
```

cache/src/main/java/net/runelite/cache/models/JagexColor.java
```diff
@@ -119,6 +142,58 @@ public static int HSLtoRGB(short hsl, double brightness)
 		return rgb;
 	}
 
+	public static int HSLtoRGBFull(int hsl)
```
Adam-, line 145: is this different from `HSLtoRGB(hsl, 1d)`..?

cache/src/main/java/net/runelite/cache/MapImageDumper.java
```diff
@@ -580,53 +591,36 @@ else if (shape == 1)
...
-											int p3 = tileShapes[tileRotations[rotIdx++]] == 0 ? underlayRgb : overlayRgb;
-											int p4 = tileShapes[tileRotations[rotIdx++]] == 0 ? underlayRgb : overlayRgb;
-											pixels[drawX + 0][drawY + i] = p1;
-											pixels[drawX + 1][drawY + i] = p2;
-											pixels[drawX + 2][drawY + i] = p3;
-											pixels[drawX + 3][drawY + i] = p4;
+											for (int ix = 0; ix < MAP_SCALE; ++ix)
+											{
+												int p = TILE_SHAPE_2D[shape - 1][rotation][idx++] == 0 ? underlayRgb : overlayRgb;
+												setPixels(pixels, drawX + ix, drawY + iy, p);
```
Adam-, line 608: are you sure about this shape-1? it already does -1 above, so it looks like it would negative index if shape==1

cache/src/main/java/net/runelite/cache/MapImageDumper.java
```diff
@@ -580,53 +591,36 @@ else if (shape == 1)
...
+											}
 										}
 									}
 									else
 									{
+										//What is this?
```
Adam-, line 614: what is this?

MESLewis: Just a note that I never figured out exactly what that code branch was used for. I've done some testing and its for drawing partial overlays, such as the brown on islands around tutorial island. I've updated the comment.

samszotkowski:
Hello, thank you both for getting this merged. I just want to comment a very very small bug regarding sprite placement (which is only halfway even a bug because Jagex's own maps have similar issues):

It appears that certain types of sprites, when they would cross over a horizontal map square boundary, have their northern half drawn on the northern side, then their northern half drawn again on the southern side.

RuneLite (tree cut in half and upper half duplicated):
image

In game (tree just cut in half, not duplicated):
image

RuneLite + some extra processing that doesn't affect sprites:
image

In game minimap, correct location of red dot:
image

PR #17954, opened 2024-06-12
xptracker: add one-tick delay to tracker initialization

### LINKED ISSUE #17910: Exp Tracker tracking all xp from creation of account as new leading to massive xp gained and per hour values on every client restart.
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

This happens on every first log in of the day for me running RuneLite version 1.10.31.2, Oldschool version Rev 222, and Launcher version 2.6.4. I used attack as an example in the screenshots to show that the tracker is just taking all xp that I've ever gained and counting it as all gained upon login but only for the first log in of every client restart. All skills are included in the tracker but only showing the first three. Resetting the xp manually works alright but the current behavior is unexpected.

### Screenshots or videos

image
image

### RuneLite version

RuneLite version 1.10.31.2, Oldschool version Rev 222, and Launcher version 2.6.4.

### Logs

_No response_

cdfisher:
Adds a one tick delay to XP Tracker initialization to prevent the tracker from erroneously showing an account's total as having been gained if logging into high-population areas. 

This should close #17910

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/xptracker/XpTrackerPlugin.java
@@ -132,6 +132,7 @@ public class XpTrackerPlugin extends Plugin
 	private boolean fetchXp; // fetch lastXp for the online xp tracker
 	private long lastXp = 0;
 	private boolean initializeTracker;
+	private boolean waitToInit;
 
 	private final XpPauseState xpPauseState = new XpPauseState();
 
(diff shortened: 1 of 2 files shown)


raiyni:
It looks like this happens due to delayed servers packets maybe. Would it make more sense to check `client::getRealSkillLevels > 0` in case it takes more than 1 tick?

raiyni:
Added in https://github.com/runelite/runelite/commit/265f801ef43060ca17b647644bafaccb713c56db

PR #18001, opened 2024-07-01
agility: track ticket dispensers even if hint arrow disappears

StephanBijzitter:
* The hint arrow can disappear when running from corner to corner in the agility arena
* Whenever the hint arrow is present, cache its location for future use
* Use a varbit to track when the active ticket dispenser changes
* Use a varbit to track ticket reward eligibility
* Move several config options into a new agility arena section
* Highlight the active ticket dispenser if the player still needs to touch it


StephanBijzitter:
Checkstyle seemed to bug out on me a bit, giving me one error in IntelliJ for `AgilityPlugin.java`:
```
Checkstyle: '{' at column 41 should be on a new line.
```
This shows up on row `298` :
```
        if (!isInAgilityArena()) // row 298
	{
		return;
	}
```
I hope that's just on my end, because it doesn't make much sense to me. 🤷‍♂️

StephanBijzitter:
Anyway, I'm currently greenlogging the brimhaven agility arena and noticed some issues, all related with the hint arrow that loves to disappear. This pull request fixes all of that and adds a highlight to the active pillar in case the hint arrow decides to not show itself.

raiyni (changes requested):
It's unnecessary to move these, you also don't need to renumber the positions

PR #18014, opened 2024-07-05
Make it easier to expand/close config sections.

geheur:
Currently you must mouse up on the exact same pixel as you mouse down, unless you are clicking the tiny ">" or "v" button. This makes it hard to open or close these sections.


Nightfirecat:
I'd much rather the sections work similar to other controls, which is that if you press down and release while only staying in the bounds of the section header, it would toggle. Otherwise (if you were to release after dragging the mouse outside of its bounds) it would have no effect. You can see with config item checkboxes, that is their behavior.

geheur:
I think activating it on mouse pressed makes it easier to click, since, especially with a clickable area that is so skinny, it is easy to move your mouse out of it before releasing the mouse button. Also, I can't think of a reason why someone would want it to activate only after they release the mouse button.

Furthermore, the buttons on the runelite sidebar activate on mouse pressed as well, showing a precedent for this design decision.

PR #18020, opened 2024-07-09
Convert party names to lower case upon text entry

gftabor:
There is a long-standing issue where OSRS capitalizes the first letter of a message, so users join parties with incorrect capitalization.

https://github.com/TheStonedTurtle/party-panel/issues/27

The current solution is all users just agree to always use all lower case party names, but this is very confusing. We should just standardize that in Runelite by converting everything to lower case.

Then when this is merged I can propagate that change to https://github.com/TheStonedTurtle/party-panel/pull/35

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPanel.java
@@ -33,6 +33,7 @@ import java.awt.Toolkit;
 import java.awt.datatransfer.Clipboard;
 import java.awt.datatransfer.StringSelection;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import javax.swing.BoxLayout;
 import javax.swing.JButton;
@@ -159,7 +160,7 @@ class PartyPanel extends PluginPanel
 					null,
 					null,
 					"");
-				s = s.toLowerCase()
+				s = s.toLowerCase(Locale.US)
 
 				if (s == null)
 				{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPanel.java
```diff
@@ -159,6 +159,7 @@ class PartyPanel extends PluginPanel
 					null,
 					null,
 					"");
+				s = s.toLowerCase()
```
Adam-, line 162: `generatePassphrase` uses `token.toLowerCase(Locale.US)` so probably the locale should be specified here, too.

runelite-client/src/main/java/net/runelite/client/plugins/party/PartyPanel.java
```diff
@@ -159,6 +159,7 @@ class PartyPanel extends PluginPanel
...
 
 				if (s == null)
```
Adam-, line 164: ?

gftabor: > ?

Haha mostly writing python right now. Missing semicolon is embarrasing.

Adam-: I was trying to identify the null pointer check after the dereference

gftabor: Ahhh, ya I hadn't made that connnection.

PR #18021, opened 2024-07-10
#17831: Fix Fortis Colosseum Death Screenshot

### LINKED ISSUE #17831: Screenshot plugin deaths don't work on Doom deaths
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

10

### Bug description

The Doom mechanic in colo kills you outright when you hit x amount of stacks (based on your modifier level), but is a death where you don't actually hit 0hp, so ActorDeath is never fired and screenshots aren't taken.

The chat message `You have been doomed!` is sent, and a custom death animation triggers when you die, (also varbit 9801 tracks doom stacks) so one of these might need to be used to handle said deaths.

Originally reported: https://github.com/pajlads/DinkPlugin/issues/472

### Screenshots or videos

_No response_

### RuneLite version

1.10.28.1

### Logs

_No response_

jaybyrrd:
Overview:

From the issue: 
> The chat message `You have been doomed!` is sent, and a custom death animation triggers when you die, (also varbit 9801 tracks doom stacks) so one of these might need to be used to handle said deaths.

- Updates the screenshot plugin to correctly trigger a death screenshot in Fortis Colosseum.
- Should resolve #17831

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
@@ -417,6 +417,11 @@ public class ScreenshotPlugin extends Plugin
 			}
 		}
 
+		if (chatMessage.equals("You have been doomed!") && config.screenshotPlayerDeath())
+		{
+			takeScreenshot("Death", SD_DEATHS);
+		}
+
 		if (chatMessage.startsWith("Your Lunar Chest count is"))
 		{
 			Matcher m = NUMBER_PATTERN.matcher(Text.removeTags(chatMessage));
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -417,6 +417,11 @@ public void onChatMessage(ChatMessage event)
 			}
 		}
 
+		if (chatMessage.equals("You have been doomed!") && config.screenshotPlayerDeath())
```
Nightfirecat, line 420: Isn't this message colored ingame? If so, this would not match. Was this tested?

Issue #17736, opened 2024-04-16
NPC aggression timer constantly showing it needs calibration 


Elite-Mercenary-Raven:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Win10

### Bug description

Most of the time when I log in, the NPC aggression plugin shows it's lost calibration. 

I only play on this PC. Currently I'm only playing on one account, but I've had multiple in the past and I've probably got some really old configurations from being a long time RL user. Maybe that has something to do with the issue. 

Is there anything I can do to fix this for the time being? I need the plugin on because I like to AFK at Rune Dragons. 

The upgrade from a text interface to an infobox is a lot nicer and I was happy to see that implemented! Is it also possible that a toggle could be added in "advanced options" that automatically hides the loss of calibration infobox when not near a relevant NPC (either if their name is in the text entry field, or for other people, when near a slayer task mob)? The calibration loss issue has me toggling the plugin on and off still, because I don't like extra stuff on my screen if I'm not immediately planning on AFKing at Rune Dragons. Outside of fixing the bug, it seems like loss of calibration is going to be common for people who play both mobile and PC; so a toggle to remove the infobox when it's not useful would be a nice QOL change that helps keep things tidy for a lot of people. Maybe the best way to do that would be to offer a timer option, where you can set the number of seconds it's visible when you log in (perhaps 30s by default), and a 0 disables it entirely. 

...While I'm at it, another great upgrade to the plugin would be to add a cooldown timer to the "hide when out of combat" option. Currently it disappears immediately upon being out of combat, but that functionality doesn't represent a realistic scenario. It would be way more useful if I could set say, a 45 second timer for it. That would let me see the area lines for a while even if a Rune Dragon doesn't manage to attack me right away, and I could immediately differentiate between aggression loss and no mobs having line of sight. 

**The first thing I'm really wondering though is if I can fix the calibration issue, which I guess shouldn't happen if I'm only playing on this PC.** 

Thank you!

### RuneLite version

RL ver: 1.10.26.4
Launcher ver: 2.6.3

cdfisher:
This isn't a bug, see [here](https://github.com/runelite/runelite/wiki/NPC-Aggression-Timer#how-it-works) for how to address it.

DeBoogie:
Do not exit the game by closing the window, but log out properly first.

Elite-Mercenary-Raven:
[...]

PR #18031, opened 2024-07-12
Fixes NPC aggression timer config loss when switching accounts

jaybyrrd:
Progresses #17736 which describes that when changing accounts you lose npc aggression data. 

- Updates `NpcAggroAreaPlugin` to use RSProfileConfiguration
~- Fixes an NPE that can occur when endTime is null~ could not get to reproduce again. 
- Removes the erroneous resetConfig() on login now that profiles are account based. 

Tested by:

1. Login to first account and calibrate so you get npc aggression info
2. Logout
3. Login to second account and calibrate so you get npc aggression info
4. Logout
5. Log back into first account and observe npc aggression info
6. Logout
7. Log back into second account and observe npc aggression info

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
@@ -47,7 +47,6 @@ import net.runelite.api.Perspective;
 import net.runelite.api.coords.LocalPoint;
 import net.runelite.api.coords.WorldArea;
 import net.runelite.api.coords.WorldPoint;
-import net.runelite.client.events.ConfigChanged;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.GameTick;
 import net.runelite.api.events.NpcSpawned;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -338,7 +339,7 @@ public void onGameTick(GameTick event)
 	{
 		WorldPoint newLocation = client.getLocalPlayer().getWorldLocation();
 
-		if (active && notifyOnce && Instant.now().isAfter(endTime))
+		if (active && notifyOnce && endTime != null && Instant.now().isAfter(endTime))
```
Nightfirecat, line 342: What code path yields `notifyOnce = true` and `endTime = null`? As best I can tell, `notifyOnce` is only ever set to true inside `createTimer()`, and `endTime` has its value set there also.

Nightfirecat (commented):
This change LGTM besides the question regarding the possible NPE.

P.S. If this NPE is real, please split it to a separate commit so the rsprofile changes are on their own.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -239,6 +239,7 @@ private void createTimer(Duration duration)
 
 		BufferedImage image = itemManager.getImage(ItemID.ENSOULED_DEMON_HEAD);
 		infoBoxManager.addInfoBox(new AggressionTimer(duration, image, this));
+		infoBoxManager.removeIf(UncalibratedInfobox.class::isInstance);
```
Nightfirecat, line 242: This is fine for normal operation on a single account, but would lead to the infobox not reappearing when logging out of one account and logging in to another account which didn't have their areas set up.

PR #18048, opened 2024-07-15
Master Clue  Step - Chaos Temple Coordinate Clue spawns a Brassican Mage

tomaspinho:


Nightfirecat:
Is that area one where you could dig inside or outside of a multi-combat area? If so, the player could spawn either depending on whether they're standing in multi. (we have a `BRASSICAN_OR_WIZARDS` enemy entry for that case)

Felanbird:
As NFC said this is almost certainly an OR, based on the few spots you could maybe dig outside the temple

tomaspinho:
Hi folks, I dug where the digging spot shows, which is inside the temple, in front of the altar - coordinate `2951, 3820`. The whole Chaos Temple are seems to be multi, so maybe it's the OR!

Adam-:
The area was changed to singles in the update last week on accident and then hotfix-patched back to multi, so this is probably a side effect of that. If it is multi it always spawns the 3 mages and it will probably fix itself in the update next week.

tomaspinho:
Oh, that explains it then! I’ll close this.

PR #18065, opened 2024-07-18
Added "Bag full of gems" to Loot Tracker.

MrNiknoc:
There are three different ItemIDs for the item "Bag full of gems". The wiki page only has percentages from a moderately sized sample of bags from one of the three sources. I asked in the OSRS Wiki Discord server if this was something that could be crowdsourced, as well as if the three bags should have separate drop tables, and it was suggested that this might be functionality worth adding to Loot Tracker.

Here's the wiki page for Bag full of gems: [Link to wiki page.](https://oldschool.runescape.wiki/w/Bag_full_of_gems)

I've added that functionality. The three different bags will now show up in Loot Tracker as "Bag full of gems (Dusuri)", "Bag full of gems (Percy)", and "Bag full of gems (Belona)".

Bagfullofgemsloottracker

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -266,6 +266,10 @@ public class LootTrackerPlugin extends Plugin
 
 	private static final String WINTERTODT_SUPPLY_CRATE_EVENT = "Supply crate (Wintertodt)";
 
+	private static final String BAG_FULL_OF_GEMS_EVENT = "Bag full of gems (Percy)";
+	private static final String BAG_FULL_OF_GEMS_24853_EVENT = "Bag full of gems (Belona)";
+	private static final String BAG_FULL_OF_GEMS_25537_EVENT = "Bag full of gems (Dusuri)";
+
 	// Soul Wars
 	private static final String SPOILS_OF_WAR_EVENT = "Spoils of war";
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -266,6 +266,10 @@ public class LootTrackerPlugin extends Plugin
 
 	private static final String WINTERTODT_SUPPLY_CRATE_EVENT = "Supply crate (Wintertodt)";
 
+	private static final String BAG_FULL_OF_GEMS_EVENT = "Bag full of gems (Percy)";
```
Adam-, line 269: these should be named eg `BAG_FULL_OF_GEMS_EVENT_PERCY`, `BAG_FULL_OF_GEMS_EVENT_BELONA` etc. Don't hard code ids into names.

MrNiknoc: Would `BAG_FULL_OF_GEMS_PERCY_EVENT` be preferred over `BAG_FULL_OF_GEMS_EVENT_PERCY` to adhere to convention?

Adam-: either is fine

MrNiknoc: Fixed with most recent commits 😁

PR #17869, opened 2024-05-20
ItemMapping: add Varlamore and Leagues 4 items

cdfisher:
Adds the following items:

-Tonalztics of Ralos
-Ring of the elements
-Trailblazer reloaded blowpipe ornament kit
-Blazing blowpipe
-Trailblazer reloaded bulwark ornament kit
-Dinh's blazing bulwark
-Blue moon armour
-Eclipse moon armour
-Blood moon armour
-Echo boots

Resolves #17868

PR #18077, opened 2024-07-22
Item Mapping: 

airgo32:
Added Moons of Peril armor, Echo boots, and demonbane weapons to ItemMapping. Useful for seeing approximate GE value of charged/damaged equipment.

Echo boots example:
image


runelite-client/src/main/java/net/runelite/client/game/ItemMapping.java
```diff
@@ -252,6 +263,10 @@ public enum ItemMapping
 	ITEM_TUMEKENS_SHADOW(TUMEKENS_SHADOW_UNCHARGED, TUMEKENS_SHADOW),
 	ITEM_PHARAOHS_SCEPTRE(PHARAOHS_SCEPTRE_UNCHARGED, true, 1L, PHARAOHS_SCEPTRE),
 	ITEM_VENATOR_BOW(VENATOR_BOW_UNCHARGED, VENATOR_BOW),
+	ITEM_ECHO_BOOTS(GUARDIAN_BOOTS, ECHO_BOOTS),
+
+	// Demonbane weapons
+	ITEM_TORMENTED_SYNAPSE(TORMENTED_SYNAPSE, EMBERLIGHT, SCORCHING_BOW, PURGING_STAFF),
```
Nightfirecat, line 269: I'm not sure whether it would be helpful to most players to display this price on these items, since the only way to get the synapse back from these is to lose it in pvp combat.

YvesW:
FYI besides the synapse this is a dupe of https://github.com/runelite/runelite/pull/17869 iirc.

runelite-client/src/main/java/net/runelite/client/game/ItemMapping.java
airgo32: Makes sense, I wasn't positive those item changes would pass, but figured I'd add them in case someone deemed them appropriate.

cdfisher:
> FYI besides the synapse this is a dupe of #17869 iirc.

If it's wanted I can add the synapse to that PR but it seems like it may not be something we want.

Nightfirecat:
I don't think we want to display the synapse value at this time, though maybe we'll change our mind in the future. As for the other additions, those have been added in #17869.

PR #18079, opened 2024-07-22
Update Telekinetic room to show one move ahead.

geheur:
I'm not sure on the cyan square. But this speeds up the room a lot.


geheur:
fixed the line endings. This file uses CRLF for some reason.

PR #18094, opened 2024-07-27
skybox: Improve default skybox color

### LINKED ISSUE #7481: Adjust the default Skybox color to a less vibrant blue
Comparison of default on the left and suggested on the right: 
skybox

jeremybahadirli:
Closes #7481

This PR changes the default overworld skybox color from #8AD2DF to #B9D6FF, which is less cyan and gives a more realistic haze/fog appearance.

Before:

After:


raiyni:
Can you provide screenshots without 117?

jeremybahadirli:
117 HD is disabled in the screenshots. Skybox and GPU plugins are enabled.

Nightfirecat:
I agree this is an improvement, though I'd like to see what a couple other colors might look like. Searching online yields a number of results that lie somewhere between these two colors. One easily-replicated example would be the CSS colors `lightskyblue` ( #87cefa) and `skyblue` ( #87ceeb). Alternatively, what does using a color somewhere between the current and proposed look like? Let's say, a perfect average between them, like  #a1d4ef?

jeremybahadirli:
Here are the alternative colors. I chose #B9D6FF because it is the default "117 HD Blue" color from the 117 HD plugin ([source](https://github.com/117HD/RLHD/blob/0d689f11d753a307125019f8adaa21c96752a9d3/src/main/java/rs117/hd/config/DefaultSkyColor.java#L36)). Of the proposed colors, it is still my favorite because with the limited render distance in OSRS, I feel the skybox acts more as a haze/fog layer, rather than clear blue sky. I am definitely still open to suggestions though.

#87CEFA:


#87CEEB:


#A1D4EF:

PR #18100, opened 2024-07-28
clues: add felling axes to Enchanted Valley step

Thource:


Felanbird (approved):
Sarnie said in the [forestry p2 reddit post](https://old.reddit.com/r/2007scape/comments/17g121v/forestry_part_two/k6dfplk/) that this works so I'll believe him

PR #18113, opened 2024-07-31
Update ClanRank.java

ql4p:
Updated enum to support new ranks added in https://secure.runescape.com/m=news/qol-improvements--further-deadman-armageddon-tweaks?oldschool=1


Adam-:
Hi, I decided to make this a value class instead 00d6ffc43965963d6b890c2af5b2bf88e5f10c20 - which is more simple and gives easy room for additional expansion later.

PR #18125, opened 2024-08-04
cluescrolls: Improve note for accessing wilderness resource area

tylerwgrass:
Clues requiring you to enter the wilderness resource area have a note that you will require 7500 coins or less if you have diaries completed.
Updated this to be dynamic based on diary completion

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -1912,11 +1912,13 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 		}
 	}
 
+	@SuppressWarnings("PMD.UnusedPrivateMethod")
 	private static String getGoldOreResourceAreaSolution(ClueScrollPlugin plugin)
 	{
 		return "Dig between the three gold ores in the Wilderness Resource Area. ".concat(getResourceAreaCost(plugin));
 	}
 
+	@SuppressWarnings("PMD.UnusedPrivateMethod")
 	private static String getSpeakToPilesSolution(ClueScrollPlugin plugin)
 	{
 		return "Speak to Piles in the Wilderness Resource Area.".concat(getResourceAreaCost(plugin));
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -929,7 +929,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 			.text("One of several rhyming brothers, in business attire with an obsession for paper work.")
 			.location(new WorldPoint(3186, 3936, 0))
 			.npc("Piles")
-			.solutionProvider(CrypticClue::getSpeakToPilesSolution)
+			.solutionProvider((plugin) -> "Speak to Piles in the Wilderness Resource Area." + getResourceAreaCost(plugin))
 			.build(),
 		CrypticClue.builder()
 			.text("Search the drawers on the ground floor of a building facing Ardougne's market.")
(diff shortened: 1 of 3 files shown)


raiyni:
Files require license headers and looks like it needs formatted with the format on the wiki https://github.com/runelite/runelite/wiki/Code-Conventions

tylerwgrass:
I've updated the approach, next time I won't open a PR at 4am 🥴

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -1899,4 +1911,38 @@ private static WorldPoint getViggoraLocation(ClueScrollPlugin plugin)
...
+	{
+		return "Speak to Piles in the Wilderness Resource Area.".concat(getResourceAreaCost(plugin));
+	}
+
+	private static String getResourceAreaCost(ClueScrollPlugin plugin)
+	{
+
+		if (plugin.getClient().getVarbitValue(Varbits.DIARY_WILDERNESS_ELITE) == 1)
+		{
+			return "No entry fee required.";
```
Felanbird, line 1932: is this wording needed? I think it's fine to add nothing about the coins if you have the elite diary done

Felanbird (changes requested):
there's 2 coordinate clues inside the resource area as well, which I assume makes this a bit more complicated

https://github.com/runelite/runelite/blob/5d1c50d47ca8504571f68c476ba14515062331e5/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java#L211
https://github.com/runelite/runelite/blob/5d1c50d47ca8504571f68c476ba14515062331e5/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java#L249
(apparently I forgot to label this master clue in 0923364872ac80e40a3ca44bb1dc7818347d10ad)

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
tylerwgrass: I'm fine with removing it, I left it because I think it's possible someone might not know/forget the elite diary removes this requirement and think they should grab some and "not know how much they need".

tylerwgrass:
> there's 2 coordinate clues inside the resource area as well, which I assume makes this a bit more complicated

If the approach towards these clues seems okay I can try to address the other ones.
I only did these since I had the actual clue to verify against

Felanbird:
I don't have the answer for the approach, but you can run these commands in the [shell](https://github.com/runelite/runelite/wiki/Using-the-client-developer-tools) to trigger the clue helper

```
// inject(net.runelite.client.eventbus.EventBus.class).post(new CommandExecuted("clue", new String[]{"24 degrees 18 minutes north, 23 degrees 22 minutes east"}));
// inject(net.runelite.client.eventbus.EventBus.class).post(new CommandExecuted("clue", new String[]{"24 degrees 07 minutes north, 23 degrees 22 minutes east"}));
// inject(net.runelite.client.eventbus.EventBus.class).post(new CommandExecuted("clue", new String[]{"One of several rhyming brothers, in business attire with an obsession for paper work."}));
// inject(net.runelite.client.eventbus.EventBus.class).post(new CommandExecuted("clue", new String[]{"More resources than I can handle, but in a very dangerous area. Can't wait to strike gold!"}));
```

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
Nightfirecat: Generally I don't think it's good to add more information than is needed; users can add clue notes if they wish, but I don't think it is necessary to mention a diary benefit that players would probably already know about.

Nightfirecat:
> there's 2 coordinate clues inside the resource area as well, which I assume makes this a bit more complicated

I think that would be better suited to a separate PR at this time, since I imagine it would be more easily achieved with a couple other quality of life features for CoordinateClue. (change to use builders like CrypticClue? remove the completely-unnecessary `lightRequired` boolean on CoordinateClueInfo? create some common source of clue data so this resource area method can be reused? etc.)

Nightfirecat:
Added in bd2895e02ae147007a56c69b3ffc0ed30d6ae20c

PR #17670, opened 2024-04-02
clues: move RequirementCollections to their own file

Felanbird:
Moves a lot of clue item variation code to it's own file, commits are split up so what was actually done can be seen, and not a giant pile of code removed and 1 new file created.

Codestyle of the new file (might)? be a little messy since I just blindly copied code from the other files, if there's something we want to changeup, review it. (e.g. maybe the barrows one should be split like this instead of `all(any(...`)
```
all(
	any(...
```

Supersedes:
Closes #6811
Closes #17317
Closes #17423

PR #18134, opened 2024-08-07
clues: Count Ghommal's Avernic defenders for Warriors' Guild emote clue

DustinKieler:
- Removes the old Warriors' Guild emote step (used to be push up emote).
- Updates the current step (Raspberry) to count Ghommal's avernic defenders for the dragon defender item requirement.


Felanbird:
avernic change will be handled by #17670

PR #18148, opened 2024-08-09
Add Nightmare Pb team size chat commands

jb1361:
This PR does the following:

- Add Chat commands for Nightmare team sizes IE: !pb nm 1
- Add more tests to test different team sizes in Nightmare
- Updated the matchPb method in ChatCommandsPlugin to check if there is a team size when you get a kill on a boss
~- Add Override to the longBossName method to support other base commands (lvl and kc) without having add a bunch of cases~

I feel like the Tests for this Plugin are a bit weak. I understand that they are testing the values that are put into the profile config but don't test that the command matches up to a command. 

Are my worries justified? 

If they are I would like to add a new tests to verify that the command actually matches up to the config value. Is there any context that I may not be aware of that would make this not possible and or extremely difficult?

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/ProfilePanel.java
@@ -392,7 +392,8 @@ class ProfilePanel extends PluginPanel
 				{
 					// characters commonly forbidden in file names
 					return CharMatcher.noneOf("/\\<>:\"|?*\r\n\0$")
-						.retainFrom(in);
+						.retainFrom(in)
+						.substring(0, 30);
 				}
 			});
 


jb1361:
Spoke with Adam in the discord, he wants the following:

- !pb nightmare to return the best time of any team size
- A good explanation on where/how the pb's for these team sizes are being saved
- Add a test that tests longBossName, essentially test the actual command and verify the correct data is returned from the profile (this is my personal want)

jb1361:
@Adam- I updated the Plugin to make `!pb nightmare` show the best time of any team size.

To answer your question on how the time is saved, the matchPb called during the New Pb regex does this. Nightmare is an edge case and is the only boss that has a team size. Nightmare also sends the team size and pb after the killcount message so we have to get this information during the matchPB method.

matchPb

I have also made the plugin update the existing best pb when the Adventure log Counters is loaded, I did this because your best time may not be the best time of any team size.

adventure-log-nightmare-pb


### Old behavior:
pb-not-best

### New Behavior:
pb-best

The reason why the kc and lvl command was broken before was because `!pb nightmare` returned the longBossName `Nightmare 6+ players` since `!pb nightmare` now returns `Nightmare` I have undid the changes I made to that method.

jb1361:
Work from further discussion:

* Split up nightmare tests to better reflect what happens in game. Add kc messages before team size - pb/kill time
* Add null check in matchPb to future proof and ensure that no pb's are missed

jb1361:
* Add cases for null teamSize to cover bugs and ensure that the in game message is always the source of truth
* Split out nightmare no new pb tests
* Added tests for AMoxliatl and Hueycoatl

PR #18167, opened 2024-08-15
bank: Display GE/HA price on item sprite via hotkey

KurtisWhiteman:
Extends Bank Plugin to allow Grand Exchange and/or High Alchemy price on Items in bank when holding a hotkey
Colour of Grand Exhange and High Alchemy text is configurable.
Hotkey is also configurable.

Preview:
java_T7LNmVMU1P


KurtisWhiteman:
Discussed with Devs, will make a separate plugin

PR #18177, opened 2024-08-20
Wiki crowd source lockpick messages

samszotkowski:
This adds the ability to track skilling success rates for "pick-lock > object" situations where the message is "You manage/fail to pick the lock," such as [Door (Yanille Dungeon)](https://oldschool.runescape.wiki/w/Door_(Yanille_Dungeon)) and [Door (Chaos Druid Tower)](https://oldschool.runescape.wiki/w/Door_(Chaos_Druid_Tower)) and some others. I include a column tracking whether the player has a lockpick or not in case that affects success rate in cases where the lockpick is optional.

This also fixes a minor bug where for pickpockets it was checking whether the player had the inventory version of max cape rather than the worn one.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/thieving/CrowdsourcingThieving.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/crowdsourcing/thieving/CrowdsourcingThieving.java
@@ -91,8 +91,7 @@ public class CrowdsourcingThieving
 			return false;
 		}
 
-		return inventoryContainer.contains(ItemID.LOCKPICK) ||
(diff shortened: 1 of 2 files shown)


samszotkowski:
Cook is aware of this - [wiki discord](https://discord.com/channels/177206626514632704/424846377621651456/1275558723061944410)

samszotkowski:
Guess we'll be looking to add this to plugin hub crowdsourcing instead. I'll do a separate PR for the max cape fix.

Adam-:
FWIW I did bring this up with Cook - he told me he needed to look at something first, and he never got back to me.

PR #18180, opened 2024-08-21
Add varrock armour smelting tracking

EvanMPutnam:
A buddy of mine realized that during our smelting grind we were not tracking the number of bars that we smelted with our Varrock platebody.  This commit (which is my first time contributing to this community 😄 ) adds that functionality to the smelting tracker.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingPlugin.java
@@ -103,6 +103,13 @@ public class SmeltingPlugin extends Plugin
 			}
 			session.increaseBarsSmelted();
 		}
+		else if (event.getMessage().startsWith("The Varrock platebody enabled you to smelt your next ore")) {
+			if (session == null)
+			{
+				session = new SmeltingSession();
+			}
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/smelting/SmeltingPlugin.java
```diff
@@ -103,6 +103,14 @@ public void onChatMessage(ChatMessage event)
 			}
 			session.increaseBarsSmelted();
 		}
+		else if (event.getMessage().startsWith("The Varrock platebody enabled you to smelt your next ore"))
```
Adam-, line 106: You can just add this to the if above
```java
if (event.getMessage().startsWith("You retrieve a bar of") || event.getMessage().startsWith("The Varrock platebody enabled you to smelt your next ore"))
```
as the code is the same.

EvanMPutnam: Works for me!  Will get an update out.

EvanMPutnam: Done :-)

PR #18182, opened 2024-08-21
bank tags: Add alphabetic sorting option for bank tag tabs

StephanBijzitter:
Fixes #16923

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsConfig.java
@@ -76,6 +76,17 @@ public interface BankTagsConfig extends Config
 	}
 
 	@ConfigItem(
+			keyName = "sortTabsAlphabetically",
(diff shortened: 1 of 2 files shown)


raiyni:
1. This is misleading with dragging and dropping
2. I think onus is on the user to organize their tabs

StephanBijzitter:
Yeah dragging and dropping gets reset by this, but I wanted to keep this minimal (at least until the first review) before going all-out.

As for the user sorting it themselves... that's a massive pain in the ass with the current behaviour of switching tabs, rather than inserting between tabs. Let's say you got 100 tabs and create a new one, which you want at the tob, you'd have to drag&drop 100 times

raiyni:
It follows the Swap/Insert setting of the bank

StephanBijzitter:
> It follows the Swap/Insert setting of the bank

Oh. My. God.

Yeah then it's probably a better idea to make a third party plugin with multiple sorting options that just rewrites `banktags.tagtabs`. Even with inserting between tabs, sorting them at the click of a button is still nice for everyone who already has a large amount of tabs

PR #18187, opened 2024-08-23
MapImageDumper: Fix drawing of sprites on chunk boundary

MESLewis:
Correctly draw sprites that start outside of the current draw region. Fixes issue noted in https://github.com/runelite/runelite/pull/17948#issuecomment-2251829928

### DIFF
--- a/cache/src/main/java/net/runelite/cache/MapImageDumper.java
+++ b/cache/src/main/java/net/runelite/cache/MapImageDumper.java
@@ -1507,7 +1507,6 @@ public class MapImageDumper
 							//What is offsetY?
 							int objSizeOffset = Math.max(2, object.getOffsetY());
 							int drawY = (drawBaseY + (Region.Y - objSizeOffset - localY)) * MAP_SCALE;
-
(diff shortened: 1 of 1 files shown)


samszotkowski:
Here's a bit of a before and after for reference.

Before:


After:

samszotkowski:
If I may recommend one small change, if we change line 1509 to:
`int drawY = (drawBaseY + (Region.Y - objSizeOffset - localY) + object.getOffsetY()) * MAP_SCALE;`
And line 1839 to:
`blitIcon(dst, x, y, sprite, scale);`

Then we get these before/afters (note the red dots, ladders, fountains):






This will make most of our sprites 1 tile off from jagex's website map, but likely where they "should" be, and it fixes the Tatty toy's red dot in Lumbridge general store.

samszotkowski:
Just noticed the bridges get pretty messed up on these so I'll have to take a closer look, my bad

samszotkowski:
I think I've found a better solution, using the object's sizes rather than their offsetY property.
https://github.com/MESLewis/runelite/pull/1

PR #18192, opened 2024-08-25
mining: start session on animation start

Alexsuperfly:
stars and rune essence were not starting mining sessions because there is no "you have gained an item" message for them.  This now instantly starts a mining session whenever you start mining a rock, which allows the mlm plugins session to be removed and use just the mining plugins one now.


runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
```diff
@@ -151,25 +147,10 @@ public void onAnimationChanged(final AnimationChanged event)
 			return;
 		}
 
-		int animId = local.getAnimation();
-		if (animId == AnimationID.DENSE_ESSENCE_CHIPPING)
```
Adam-, line 155: Did you check this handles dense essence ok?

Alexsuperfly: i did check it but i was just running around to all the rocks in the same session, when i just logged in next to it and tried to start a session, you're right, it didn't work correctly.  I put the [dense essence session start](https://github.com/runelite/runelite/compare/83f8076494050f8e9a4739bf2f4ed1692a8c4c54..0aff6bfd6e7efeb632351a3f9c145eec9b017a41) back now.

PR #18199, opened 2024-08-29
agility: fix wilderness course

Felanbird:
~~this is probably correct but I'm not bonded so I didn't test it, if someone else wants to~~

image
image
image


tylerwgrass:
Ran each ending and LGTM.

PR #18209, opened 2024-08-30
itemstats: add attack range

tylerwgrass:
related PRs:

[scraper](https://github.com/runelite/runelite-wiki-scraper/pull/22)
[api]()

attack-range-info


Infinitay:
May I suggest instead using a property name of "Attack Range" or "Weapon Range" instead to make it more clear whether the value is a stat bonus or _meta_ data?

PR #18237, opened 2024-09-05
Add QoL changes to the script inspector

vikke1234:
This PR adds the following changes:

- Copy args for script to clipboard
- Add to black/whitelist using enter
- Remove from black/whitelist using delete
- Remove multiple entries from black/whitelist at once

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
@@ -29,16 +29,26 @@ import java.awt.BorderLayout;
 import java.awt.Component;
 import java.awt.Dimension;
 import java.awt.FlowLayout;
+import java.awt.Toolkit;
+import java.awt.datatransfer.Clipboard;
+import java.awt.datatransfer.StringSelection;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
 import java.awt.event.AdjustmentEvent;
 import java.awt.event.AdjustmentListener;
 import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.swing.BorderFactory;
 import javax.swing.DefaultListModel;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
```diff
@@ -376,6 +422,42 @@ private void addScriptLog(ScriptTreeNode treeNode)
...
+				@Override
+				public void mouseClicked(MouseEvent e)
+				{
+					if (e.getButton() == MouseEvent.BUTTON3)
+					{
+						int row = tree.getClosestRowForLocation(e.getX(), e.getY());
+						tree.setSelectionRow(row);
+						JPopupMenu popup = new JPopupMenu();
+						JMenuItem item = new JMenuItem("Copy args");
+						item.addActionListener(new ActionListener()
```
Nightfirecat, line 436: The `ActionListener` can be replaced by a lambda, eg.

```java
item.addActionListener(actionEvent ->
{
	DefaultMutableTreeNode node = // ...
});
```

runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
```diff
@@ -257,6 +283,26 @@ public void adjustmentValueChanged(AdjustmentEvent e)
...
+				{
+					try
+					{
+						jSpinner.commitEdit();
+						addToSet(jSpinner);
+					}
+					catch (ParseException ex)
+					{
+						// ignore
+					}
```
Nightfirecat, line 301:
```suggestion
					catch (ParseException ignored)
					{}
```

runelite-client/src/main/java/net/runelite/client/plugins/devtools/ScriptInspector.java
```diff
@@ -376,6 +422,42 @@ private void addScriptLog(ScriptTreeNode treeNode)
...
+							public void actionPerformed(ActionEvent e)
+							{
+								DefaultMutableTreeNode node = (DefaultMutableTreeNode) tree.getLastSelectedPathComponent();
+								// example entry: 69  -  378.62 args: [69, 0, 24772672]
+								Pattern argPattern = Pattern.compile("args: \\[(.*)]");
+								Matcher matcher = argPattern.matcher(node.toString());
+								if (matcher.find())
+								{
+									Clipboard clipboard = Toolkit.getDefaultToolkit().getSystemClipboard();
+									clipboard.setContents(new StringSelection(matcher.group(1)), null);
```
Nightfirecat, line 448: Is there any reason we're formatting it as: `2282, 11927571, 11927572, 11927573, 11927574, 11927575` instead of `2282  -  182.18 Logout.SATISFACTION_5SCALE args: [2282, 11927571, 11927572, 11927573, 11927574, 11927575]`? The latter seems better, imo.

Nightfirecat (changes requested):

vikke1234:
Thanks for the comments.

The idea behind not including extra text (iirc) was at the time I wrote it I was working a lot with the scripts and I wanted to easily be able to copy the args for the scripts and paste them into the call in the plugin code. Not including the `2282  -  182.18 Logout.SATISFACTION_5SCALE args: [ ... ]` made this significantly easier and likely you don't need the extra information anyway.

I do think there's an argument to be made to be able to copy that as well though.

I'll try to fix the other comments asap.

Issue #18236, opened 2024-09-05
Display dynamic resource area entry cost in coordinate clues landing inside


Nightfirecat:
Similar to the implementation in #18125, we should display a dynamic resource area cost based on the wilderness diary levels the player has completed. Because coordinate clues are structured differently, there are likely other clean-ups and related changes which could happen during this effort, such as:

* ~~Using builders to create these clues--this would mean we could get rid of some convoluted ctors for the clues~~ (done in #18240)
* ~~Remove the apparently unnecessary `lightRequired` field. I don't recall whether this was needed when I added it originally, but it certainly isn't necessary now~~ (done in #18240)
* Maybe create some common class and/or package for re-used clue data. Depending on how it is organized, this may or may not include things like item requirement collections used in emote clues, but would certainly be useful for this resource area solution suffix method.

The coordinate clues needing this behavior added are listed in this comment: https://github.com/runelite/runelite/pull/18125#pullrequestreview-2217835539

Nightfirecat:
Added in 921d9003e0a6eda61cc2eec22487971891a9b7de

Issue #17616, opened 2024-03-23
Clues: "NO OWNER" anagram clue has wrong location


jsnells1:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

The world location for the elite anagram clue "NO OWNER" is pointing to the wrong location.

### Screenshots or videos

![image](https://github.com/runelite/runelite/assets/18359208/407c0a50-bdfc-4316-9c9d-6865b2a040dc)
![image](https://github.com/runelite/runelite/assets/18359208/5bfb8db1-302f-468c-bbc0-e79f364e7fd9)

### RuneLite version

RL Version: 1.10.25.2
Launcher Version: 2.6.4

jsnells1:
I want to add that in Lletya, the NPC is properly indicated. This is only an issue on the world map.

PR #18240, opened 2024-09-06
clues: clean up CoordinateClue

tylerwgrass:
Addresses first two bullets in #18236:
- Replaces constructors with builder
- Replaces `lightRequired` with a check for the firepit varbit presence

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
@@ -24,16 +24,15 @@
  */
 package net.runelite.client.plugins.cluescrolls.clues;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import java.awt.Color;
 import java.awt.Graphics2D;
-import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import lombok.AccessLevel;
+import lombok.Builder;
 import lombok.Getter;
-import lombok.NonNull;
-import lombok.ToString;
 import net.runelite.api.Varbits;
 import net.runelite.api.annotations.Varbit;
 import net.runelite.api.coords.LocalPoint;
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
```diff
@@ -55,206 +54,186 @@
 public class CoordinateClue extends ClueScroll implements LocationClueScroll
 {
 	@Getter
+	@Builder
```
Nightfirecat, line 57: At this point, CoordinateClue itself should use a builder pattern and incorporate the fields from CoordinateClueInfo.

tylerwgrass:
I've flattened the class.

I tried test loading a few clue scrolls by using associated coordinates listed on the wiki but they don't seem to match what the currently mapped points are, not sure what that is about.

For example I tried loading the clue for coordinates `"03 degrees 09 minutes south, 43 degrees 26 minutes east"`
which should be here
200px-03 09S_43 26E_map

but my client shows this location prior to this commit (and doesn't show at all within this commit because it just fetches from the coord map)
image

not sure if i'm just testing incorrectly

Felanbird:
I can replicate triggering that step manually, shows the map location wrong on 1.10.51 too, idk why, when Jagex unlocked the map #17616 was reported within 4 hours so I assume this isn't actually wrong?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
```diff
@@ -266,22 +735,32 @@ private CoordinateClueInfo(@Nonnull String directions, Enemy enemy, boolean ligh
...
 	private final WorldPoint mirrorLocation;
 
-	public CoordinateClue(String text, WorldPoint location, WorldPoint mirrorLocation)
+	@Builder
+	private CoordinateClue(
+		String text,
+		String directions,
+		Enemy enemy,
+		WorldPoint location,
+		Integer lightSourceVarbitId
```
Nightfirecat, line 744:
```suggestion
		@Varbit Integer lightSourceVarbitId
```

Nightfirecat:
Possibly incorrect coordinate of that clue aside, as that's unrelated to this PR, this looks pretty good now. I've pushed a commit which changes the `CLUES` construction to more what I envisioned (a list, which is streamed and collected to a map, so the location does not need to be entered twice for every clue)

PR #18241, opened 2024-09-06
loottracker: Muddy Chest world metadata

Skretzo:
World metadata is necessary for the crowdsourcing to distinguish F2P and P2P worlds for finding unknown rare loot drop rates.

image


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -1008,6 +1008,13 @@ public void onChatMessage(ChatMessage event)
 				return;
 			}
 
+			if (regionID == 12348) // Muddy Chest crowdsourcing needs F2P/P2P world
```
raiyni, line 1011: This should be defined as a constant like other regionIds in the class

raiyni (changes requested):

PR #18249, opened 2024-09-09
wiki dps: update wiki dps button position on client resize

airgo32:
Fixes a bug which causes the "View DPS" button on the bank equipment screen to be misaligned with the client is resized.

To replicate the bug, open the bank, toggle the equipment screen, and resize the client window.

### Before:
DPS-button-before
### After:
DPS-button-after

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiDpsManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/wiki/WikiDpsManager.java
@@ -291,6 +291,26 @@ class WikiDpsManager
 		text.setAction(0, "View DPS on OSRS Wiki");
 		text.setOnOpListener((JavaScriptCallback) ev -> onClick.run());
 
+		// listener which updates button height if window is resized
+		if (screen == Screen.BANK_EQUIPMENT)
+		{
+			text.setOnTimerListener((JavaScriptCallback) ev ->
+			{
+				int setBonusY = setBonus.getOriginalY() + parent.getHeight() / 2 - setBonus.getHeight() / 2;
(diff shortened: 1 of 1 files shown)


Adam-:
Typically we rebuild interfaces when they get invalidated, instead of trying to adjust around the changes, since it is easier, usually requires less code, and is less bug prone. Bank tags uses this to detect bank resizes (maybe it is not applicable since this is not the bank but the equipment screen):
```java
		else if (event.getScriptId() == ScriptID.BANKMAIN_SIZE_CHECK && enabled)
		{
			// from [proc,bankmain_size_check] - resize tab layer when the bank is resized
			final int[] intStack = client.getIntStack();
			final int intStackSize = client.getIntStackSize();
			Widget w = client.getWidget(intStack[intStackSize - 5]);
			int width = intStack[intStackSize - 4];
			int height = intStack[intStackSize - 3];
			if (w.getWidth() != width || w.getHeight() != height)
			{
				log.debug("Bank resize!");
```

airgo32:
So in this case, it would be better to remove and re-add the button when the window is resized?

Adam-:
I would kind of expect the vanilla code to remove the button for you, if it doesn't then I am not sure what it is doing. Do you know how vanilla handles the resize?

airgo32:
No, I don't know the specifics on resizing in vanilla. Since this button is added by RL and gets its position defined relative to the Set Bonus button, that's what I used to update the DPS button.

Adam-:
What is mean is, how does the Set bonus button move when the window is resized?

Adam-:
I've fixed this in 9070e8711087d43a04c4c8913bf8e78cf09a1903

Issue #18258, opened 2024-09-11
screenshot boss kills no longer working over one thousand kc


Alexsuperfly:
### Bug description

boss kill messages now have a thousands place comma separator which [the regex](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java#L99) fails to match

### Screenshots or videos

a kill message auto screenshotted before the update today
![image](https://github.com/user-attachments/assets/704c0186-c974-46c2-9893-8a9e0579253c)

a kill message manually screenshotted after the update today
![image](https://github.com/user-attachments/assets/ca507080-fbfc-4c83-b25d-33955a2b81d4)

416c616e:
This applies to more than just bosses, Barrows also reports a comma as well. Perhaps expand the changes as I presume this is the new norm?

PR #18273, opened 2024-09-14
screenshot: remove commas from integer parses

416c616e:
Updating logic to replace commas from screenshot plugin integer parses, inline with #18258 and 246283549444d1067e9c4785c6534d71407e7e2c


Adam-:
You have to actually change the `NUMBER_PATTERN` to capture `,`, or else this commit just doesn't do anything.

Adam-:
It would be good to throw a few , in some of the test cases for the plugin so that it is tested.

PR #18285, opened 2024-09-16
Skill calc: Add extended anti-venom and misc. 1&2 dose pots

evilgenious448:
- Add miscelaneous 1-tickable 1 & 2 dose potions such as Stamina, Anti-venom, Extended Antifire and Extended Super Antifire.
- Add Extended Anti-venom+ 1, 2, 3 & 4 dose potions


Adam-:
Is there a reason why you are adding 1/2 dose potions? There is never any reason to train using them since you can decant them into (4) and it doesn't take any additional 2nds to make the (4) dose variants.

evilgenious448:
No particular reason for the 1/2 doses. Mainly, I just saw that for the potions where you could create them in any number of doses (1-4), we had 3 and 4 does in the skill calc, while for all other potions we always used the correct dosages that were permitted (usually 3). So I thought it could be useful for some people to identify that certain potions could be made in 1 or 2 doses, or to help them do certain calculations using the smaller dose amounts. But happy to not have that added if you think it is unnecessary and would clutter the skill calc.

Adam-:
I've added the new extended antivenom here b26987de3a33f7592d55e552dabee4cc5229ba19.

PR #18306, opened 2024-09-22
clues: Add Item IDs to Clues; Identify tiers Easy to Elite Using ID

TheLope:
1. ItemID values have been added to each clue type
2. Config settings have been updated
a. An additional configuration option to auto-identify easy through elite clues has been provided.
3. When the above setting is enabled, Easy, Medium, Hard, and Elite clue scrolls are automatically identified based on Item Id when they enter the inventory.
This manifests in two common scenarios:
a. Clue is picked up from the floor:
java_4VaNmIPRwi
b. A step is completed, and the next step is automatically identified:
4NKbI0n


TheLope:
Added warning to Auto-identify toggle to caveat Beginner and Master functionality.
Perhaps this removes the need for
1. Beginner/Master overlay placeholder
2. Advanced section hidden by default

raiyni:
Wouldn't it make sense to tie this into already existing clue classes so we don't maintain multiple versions of the hint?

TheLope:
> Wouldn't it make sense to tie this into already existing clue classes so we don't maintain multiple versions of the hint?

Very fair, I did not include the MapClues in the ClueText class, because the IDs were already elements there. Not sure if there is a reason, we would not want to divide the rest of them among the other classes.

For some classes like MusicClue, it looks like text is dynamically generated and there wouldnt be a 1-1 mapping of ID to the text.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -958,768 +1207,894 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 			.build(),
 		CrypticClue.builder()
+			.itemId(ItemID.CLUE_SCROLL_EASY_19822)
 			.text("I would make a chemistry joke, but I'm afraid I wouldn't get a reaction.")
 			.location(new WorldPoint(2932, 3212, 0))
 			.npc("Chemist")
 			.solution("Talk to the Chemist in Rimmington")
 			.build(),
 		CrypticClue.builder()
-			.text("Show this to Hazelmere.")
```
TheLope, line 1478: No longer obtainable

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
```diff
@@ -89,172 +93,172 @@ private CoordinateClueInfo(@Nonnull String directions, Enemy enemy, boolean ligh
...
+		.put(new WorldPoint(3039, 3960, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_2739, "Wilderness. Pirates' Hideout.", ZAMORAK_WIZARD))
+		.put(new WorldPoint(2987, 3963, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_2725, "Wilderness. West of Wilderness Agility Course.", ZAMORAK_WIZARD))
+		.put(new WorldPoint(3189, 3963, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_2729, "Wilderness. North of Resource Area, near magic axe hut.", ZAMORAK_WIZARD))
+		.put(new WorldPoint(2341, 3697, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_23167, "North-east of the Piscatoris Fishing Colony bank.", SARADOMIN_WIZARD))
+		.put(new WorldPoint(3143, 3774, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_23168, "In level 32 Wilderness, by the black chinchompa hunting area.", ZAMORAK_WIZARD))
+		.put(new WorldPoint(2970, 3913, 0), new CoordinateClueInfo(ItemID.CLUE_SCROLL_HARD_23169, "Frozen Waste Plateau, south-west of Wilderness Agility Course.", ZAMORAK_WIZARD))
+		.put(new WorldPoint(1410, 3611, 0), new CoordinateClueInfo(-1, "Lake Molch dock west of Shayzien Encampment.", SARADOMIN_WIZARD))
```
TheLope, line 191: Unobtainable, I suggest removal

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -958,768 +1207,894 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 			.build(),
 		CrypticClue.builder()
+			.itemId(ItemID.CLUE_SCROLL_EASY_3507)
 			.text("Search the drawers in the upstairs of a house in Catherby.")
 			.location(new WorldPoint(2809, 3451, 1))
 			.objectId(ObjectID.DRAWERS_350)
 			.solution("Perdu's house in Catherby.")
 			.build(),
 		CrypticClue.builder()
+			.itemId(-1)
```
TheLope, line 1710: I was not able to find this clue in the wiki

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
```diff
@@ -337,6 +341,19 @@ public void makeWorldOverlayHint(Graphics2D graphics, ClueScrollPlugin plugin)
 		}
 	}
 
+//	public static CoordinateClue forItemId(int itemId)
```
TheLope, line 344: Not sure how to handle this using CoordinateClueInfo

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -2294,7 +2294,7 @@ public Collection<Integer> getNpcRegions()
 	@Override
 	public int[] getConfigKeys()
 	{
-		return new int[]{text.hashCode()};
+		return new int[]{text.hashCode() + itemId};
```
TheLope, line 2297: This was changed due to duplicate Sherlock steps. Interested in a better way to handle those if this is not sufficient.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
```diff
@@ -341,18 +342,26 @@ public void makeWorldOverlayHint(Graphics2D graphics, ClueScrollPlugin plugin)
...
-//		{
-//			if (clue.itemId == itemId)
-//			{
-//				return clue;
-//			}
-//		}
-//
-//		return null;
-//	}
+	public static CoordinateClue forItemId(int itemId)
```
TheLope, line 344: Works but I hope there is a suggestion on how to do this in a more straightforward way.

TheLope:
Narrowed scope of PR to focus on Clue IDs

TheLope:
> Wouldn't it make sense to tie this into already existing clue classes so we don't maintain multiple versions of the hint?

Resolved

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -434,6 +452,38 @@ public void onItemContainerChanged(final ItemContainerChanged event)
...
+		clueItemIdsInInventory.removeIf(clue -> !inventoryContainer.contains(clue));
+
+		// Add new clues
+		Item[] inventory = inventoryContainer.getItems();
+		Set<Integer> cluesFound = new HashSet<>();
+
+		for (Item item : inventory)
+		{
+			int invItemId = item.getId();
+			if (findClueScroll(invItemId) != null)
```
TheLope, line 478: This lookup is very inefficient. All I need to know is if the ItemID is a Clue Scroll

TheLope: Resolved while accepting that solution will populate clueItemIdsInInventory with Beginner, Master, Challenge, and Veos Clues. The null check on line 459 should cover these.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -115,133 +116,134 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
...
+		new EmoteClue(ItemID.CLUE_SCROLL_HARD_12550, "Cheer at the top of the agility pyramid. Beware of double agents! Equip a blue mystic robe top and any rune heraldic shield.", "Agility Pyramid", AGILITY_PYRAMID, new WorldPoint(3043, 4697, 3), DOUBLE_AGENT_108, CHEER, item(MYSTIC_ROBE_TOP), any("Any rune heraldic shield", item(RUNE_SHIELD_H1), item(RUNE_SHIELD_H2), item(RUNE_SHIELD_H3), item(RUNE_SHIELD_H4), item(RUNE_SHIELD_H5))),
+		new EmoteClue(CLUE_SCROLL_MASTER, "Dance in Iban's temple. Beware of double agents! Equip Iban's staff, a black mystic top and a black mystic bottom.", "Iban's temple", WELL_OF_VOYAGE, new WorldPoint(2011, 4712, 0), DOUBLE_AGENT_141, DANCE, any("Any iban's staff", item(IBANS_STAFF), item(IBANS_STAFF_U)), item(MYSTIC_ROBE_TOP_DARK), item(MYSTIC_ROBE_BOTTOM_DARK)),
```
Nightfirecat, line 124: While the plugin does prevent lookups for beginner and master clues, it feels strange to associate multiple clues to the same ID here when it is not possible to do such a lookup on item ID alone. I'd prefer some constant (with value `-1`) were used here instead.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java
```diff
@@ -115,133 +116,134 @@ public class EmoteClue extends ClueScroll implements LocationClueScroll
...
-		new EmoteClue("Panic at Al Kharid mine.", "Al Kharid mine", null, new WorldPoint(3303, 3271, 0), PANIC),
-		new EmoteClue("Spin at Flynn's Mace Shop.", "Falador", null, new WorldPoint(2950, 3387, 0), SPIN),
-		new EmoteClue("Salute by the Charcoal Burners. Equip a Farmer's strawhat, Shayzien platebody (5) and Pyromancer robes.", "Charcoal Burners", CHARCOAL_BURNERS, new WorldPoint(1714, 3467, 0), SALUTE, any("Farmer's strawhat", item(FARMERS_STRAWHAT), item(FARMERS_STRAWHAT_13647)), item(SHAYZIEN_BODY_5), item(PYROMANCER_ROBE)));
+		new EmoteClue(ItemID.CLUE_SCROLL_HARD_12544, "Beckon on the east coast of the Kharazi Jungle. Beware of double agents! Equip any vestment stole and a heraldic rune shield.", "Kharazi Jungle", NORTHEAST_CORNER_OF_THE_KHARAZI_JUNGLE, new WorldPoint(2954, 2933, 0), DOUBLE_AGENT_108, BECKON, any("Any stole", item(GUTHIX_STOLE), item(SARADOMIN_STOLE), item(ZAMORAK_STOLE), item(ARMADYL_STOLE), item(BANDOS_STOLE), item(ANCIENT_STOLE)), any("Any heraldic rune shield", item(RUNE_SHIELD_H1), item(RUNE_SHIELD_H2), item(RUNE_SHIELD_H3), item(RUNE_SHIELD_H4), item(RUNE_SHIELD_H5))),
```
Nightfirecat, line 119: Nit-pick: `ItemID.*` is already imported in this file; clue scroll identifiers can be used without the class qualifier.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
Nightfirecat, line 1: The change to this file makes music clue matching more fragile. Please ensure the plugin is still able to identify clues which ask for other songs--this could happen because Jagex adds new music clues with new map areas, or perhaps they could associate a new clue with pre-existing music.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
```diff
@@ -36,10 +37,27 @@ public interface ClueScrollConfig extends Config
...
+		description = "Configures whether or not to display hint arrows for clues.",
+		position = 0
 	)
 	default boolean displayHintArrows()
 	{
 		return true;
 	}
+
+	@ConfigSection(name = "Advanced", description = "Advanced options", position = 1, closedByDefault = true)
+	String advancedSection = "Advanced";
```
Nightfirecat, line 49: I don't think we need a new config section for this.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
```diff
@@ -36,10 +37,27 @@ public interface ClueScrollConfig extends Config
...
+
+	@ConfigItem(
+			keyName = "autoIdentifyClues",
+			name = "Auto-identify clue scrolls",
+			description = "Automatically identify clue scrolls when added to inventory",
+			warning = "Does not work for Beginner and Master clues due to their shared Item IDs",
```
Nightfirecat, line 55: I don't know that this limitation is significant enough to merit bringing up a pop-up window to warn users of it.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
```diff
@@ -36,10 +37,27 @@ public interface ClueScrollConfig extends Config
 	@ConfigItem(
 		keyName = "displayHintArrows",
 		name = "Display hint arrows",
-		description = "Configures whether or not to display hint arrows for clues."
+		description = "Configures whether or not to display hint arrows for clues.",
+		position = 0
```
Nightfirecat, line 41: I think it's fine to let these config items arrange themselves alphabetically.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -434,6 +452,39 @@ public void onItemContainerChanged(final ItemContainerChanged event)
...
+
+		// Add new clues
+		Item[] inventory = inventoryContainer.getItems();
+		Set<Integer> cluesFound = new HashSet<>();
+
+		for (Item item : inventory)
+		{
+			int invItemId = item.getId();
+			// If paramID 623 returns a dbrow, the invItemId is a valid Clue Scroll
+			if (client.getItemDefinition(invItemId).getIntValue(623) != -1)
```
Nightfirecat, line 479: This param should become a documented constant in ParamID.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -682,7 +733,7 @@ else if (event.getGroupId() == InterfaceID.CLUESCROLL)
 					}
 					else
 					{
-						log.info("Unknown clue text: {}", clueScrollText.getText());
+						log.info("Unknown clue scroll (id {}) for '{}'", clueItemId, clueScrollText.getText());
```
Nightfirecat, line 736:
```suggestion
						log.info("Unknown clue scroll (id {}) for text '{}'", clueItemId, clueScrollText.getText());
```

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/AnagramClue.java
```diff
@@ -383,12 +429,14 @@ public class AnagramClue extends ClueScroll implements NpcClueScroll, ObjectClue
 			.answer("2")
 			.build(),
 		AnagramClue.builder()
+			.itemId(ItemID.CLUE_SCROLL_MASTER)
```
Nightfirecat, line 432: Same feedback as emote clues, it seems odd to me to associate clues with real item IDs if we intentionally don't want to match them.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -61,129 +62,248 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 			.npc("Sherlock")
 			.solution("Sherlock is located to the east of the Sorcerer's tower in Seers' Village.")
 			.build(),
 		CrypticClue.builder()
+			.itemId(ItemID.CLUE_SCROLL_ELITE_12114)
+			.text("Show this to Sherlock.")
+			.location(new WorldPoint(2733, 3415, 0))
+			.npc("Sherlock")
+			.solution("Sherlock is located to the east of the Sorcerer's tower in Seers' Village.")
+			.build(),
```
Nightfirecat, line 77: For ease of use here, perhaps associating a set of IDs instead of duplicating the clue many times over is the better approach.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
TheLope: This was mostly an attempt to reduce support traffic for users which may become confused when Beginner and Master clue are not "Auto-identified"

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
```diff
@@ -100,7 +157,7 @@ public static MusicClue forText(String text)
 		if (m.find())
 		{
 			final String song = m.group(1);
```
TheLope, line 159: @Nightfirecat So basically if we hit this line, the song exists. We could then look for an existing MusicClue using that song, and if it doesn't exist, then just create a new MusicClue?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
TheLope: Yeah I can look at having itemId accept a list of IDs, since 12113-12127 are all sherlock clues.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
TheLope: My understanding is that param 623 points to a 'clueScrollRow', which will point to a 'clueSolutionRow'. So this is a good way to check if something is a clue.
Any suggestions would be appreciated for what the constant would be called in https://github.com/runelite/runelite/blob/182b8d7822d088b4206acffbf8546aa8df22cc21/runelite-api/src/main/java/net/runelite/api/ParamID.java
Would this be included in this PR or separate?

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MusicClue.java
TheLope: Resolved by checking if forSong returns, null, if so creating a new MusicClue w/ itemId -1

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
TheLope: Added new ParamID. Would like to give it a better name though

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
TheLope: itemId in CrypticClue now expects Set.of

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -1012,501 +1173,574 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
...
 			.build(),
 		CrypticClue.builder()
+			.itemId(Set.of(ItemID.CLUE_SCROLL_EASY_3507))
 			.text("Search the drawers in the upstairs of a house in Catherby.")
 			.location(new WorldPoint(2809, 3451, 1))
 			.objectId(ObjectID.DRAWERS_350)
 			.solution("Perdu's house in Catherby.")
 			.build(),
 		CrypticClue.builder()
-			.text("Search a crate in the Haymaker's arms.")
```
TheLope, line 1392: I was not able to find this clue in the wiki. Removed
https://oldschool.runescape.wiki/w/The_Haymaker%27s_Arms

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -1930,7 +2217,7 @@ public Collection<Integer> getNpcRegions()
 	@Override
 	public int[] getConfigKeys()
 	{
-		return new int[]{text.hashCode()};
+		return new int[]{text.hashCode() + itemId.stream().mapToInt(Integer::intValue).sum()};
```
TheLope, line 2220: This was changed due to duplicate Sherlock steps. Interested in a better way to handle those if this is not sufficient.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -435,6 +454,39 @@ public void onItemContainerChanged(final ItemContainerChanged event)
 		}
 	}
 
+	private void setActiveClue(Integer itemId)
```
Nightfirecat, line 457: As a point of semantics, `Integer` isn't the right type; item IDs are `int`s.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -435,6 +454,39 @@ public void onItemContainerChanged(final ItemContainerChanged event)
...
+		{
+			clueItemId = itemId;
+			updateClue(clueScroll);
+		}
+	}
+
+	private Set<Integer> getClueItemIdsInInventory(ItemContainer inventoryContainer)
+	{
+		// Remove clues no longer in inventory
+		clueItemIdsInInventory.removeIf(clue -> !inventoryContainer.contains(clue));
```
Nightfirecat, line 471: This seems like a weird place for this side effect to occur; I'd expect it to be called directly within `onItemContainerChanged`. Moreover, this seems like it is necessary only because the plugin is storing held clues instead of the complete inventory? I think this problem goes away if the plugin instead saves a copy of the whole inventory and gets the difference between them on item container change--checking only the difference (the additions, at that) for any new clues entering the inventory.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
Nightfirecat: Please make this change in a separate commit from the PR's behaviors

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollConfig.java
Nightfirecat: Respectfully, this is an overuse of the warning feature. The core client uses config warnings exactly 3 times:
1. GPU plugin > Compute Shaders (requires plugin restart to take effect, notifies user of specific GPU feature requirements to work at all)
2. RuneLite > Enable custom window chrome (requires client restart)
3. RuneLite > Enable tray icon (can impact notification behavior; requires client restart)

This feature has much less impact, and requires less user action, than any of the above, thus doesn't rise to the level of needing a modal warning.

Further, we already get user reports of differences in clue behavior for beginner and master clues for things such as inventory tags, so this would be one more thing the developers/support folk would be aware to inform users of.

I think it would be fine to add this caveat to the description on a new line. (use `<br>` to add that)

Nightfirecat:
One more note, in the same way we're logging the ID for completely unknown clues, it might be good to add logging for known clues with unknown IDs. (such as music clues and coordinate clues since they can be identified without an ID)

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
TheLope: Refactored

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -453,6 +453,11 @@ private void setActiveClue(int itemId)
 			clueItemId = itemId;
 			updateClue(clueScroll);
 		}
+		else if (itemId != ItemID.CLUE_SCROLL_BEGINNER && itemId != ItemID.CLUE_SCROLL_MASTER)
```
TheLope, line 456: > One more note, in the same way we're logging the ID for completely unknown clues, it might be good to add logging for known clues with unknown IDs. (such as music clues and coordinate clues since they can be identified without an ID)

@Nightfirecat I don't think this is quite what you had in mind, but may be sufficient event though its only called when autoIdentifyClues is enabeld

Nightfirecat:
FYI I am testing this on my dev branch now; I'll come back to review this further after some ingame testing.

Nightfirecat:
One issue is that removing the clue overlay (via shift+right-click -> reset) has it reappear the next game tick.

TheLope:
> One issue is that removing the clue overlay (via shift+right-click -> reset) has it reappear the next game tick.

Thanks I hasn't considered this. Will look to migrate, let me know if you have an idea how to do so

Nightfirecat:
The added code in `onGameTick` is definitely the cause; I'm not entirely sure at a glance why it's there, so the solution is probably to remove it

TheLope:
> The added code in `onGameTick` is definitely the cause; I'm not entirely sure at a glance why it's there, so the solution is probably to remove it

That does the work of setting the active clue, but could probably just move it to onItemContainerChanged

TheLope:
> One issue is that removing the clue overlay (via shift+right-click -> reset) has it reappear the next game tick.

Resolved

Nightfirecat:
This all has been working fine in my testing over the past couple of weeks, so I'll do another code pass soon. Before that happens, can you resolve the merge conflicts by switching the item references from api.ItemID to api.gameval.ItemID?

If it helps, abex wrote an IntelliJ IDE script which assists in migrating those references: https://gist.github.com/abextm/e9fd4ba19fe90f921e81abab468e9c0a

TheLope:
> This all has been working fine in my testing over the past couple of weeks, so I'll do another code pass soon. Before that happens, can you resolve the merge conflicts by switching the item references from api.ItemID to api.gameval.ItemID?
> 
> If it helps, abex wrote an IntelliJ IDE script which assists in migrating those references: https://gist.github.com/abextm/e9fd4ba19fe90f921e81abab468e9c0a

Will do thank you

TheLope:
> This all has been working fine in my testing over the past couple of weeks, so I'll do another code pass soon. Before that happens, can you resolve the merge conflicts by switching the item references from api.ItemID to api.gameval.ItemID?
> 
> If it helps, abex wrote an IntelliJ IDE script which assists in migrating those references: https://gist.github.com/abextm/e9fd4ba19fe90f921e81abab468e9c0a

Should be taken care of. Happy to rebase again to squash commits if desired.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -432,6 +441,64 @@ public void onItemContainerChanged(final ItemContainerChanged event)
...
+		if (!newClueItemIds.isEmpty())
+		{
+			return newClueItemIds.iterator().next();
+		}
+
+		// Check if latest clue still in inventory
+		if (clueItemIds.contains(clueItemId))
+		{
+			return latestClueItemId;
+		}
```
Nightfirecat, line 500: I read this as happening if:

* Since the last inventory change, no new clues have entered your inventory _and_
* The list of clues in the inventory still contains the clue for which the solver is active

So in that case, why is it helpful to maintain the value of `latestClueId` when it will not be used? I don't really understand what this is trying to solve.

TheLope: This may be a remnant of some logic since removed. Will give it a proper review but could be no longer needed.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
Nightfirecat: As I understand it, this is not "handled", per se, in master, since these steps all have the same solution (to talk to Sherlock), therefore should use the same clue notes.

Nightfirecat:
I've squashed most of the history and cleaned up some bits of code here and there. I'll be testing off this branch to verify the changes are all working: https://github.com/Nightfirecat/runelite/commits/identify-clues

Once that's confirmed, I'll look at adding the toggleable identification behavior (no identification, identify when no solver is active, and always identify)

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
TheLope: You're correct yeah, I wasnt considering that it would become the "officially", set clue and clueId

Nightfirecat:
I'd love feedback on [the proposed change to the automatic identification config option](https://github.com/Nightfirecat/runelite/commit/bf8c093c323e4a95694345c3fbb5d092be808a6b#diff-747d598f2e91f8dbf153e4e83e1409ef061afa97be7a6f7096501637ef37ad64), specifically on the naming. The wording is currently small enough to prevent the text from being clipped, however it is not the most clear. (regardless of the wording we go with, I'll likely expand on it in the description for clarity)

TheLope:
> I'd love feedback on [the proposed change to the automatic identification config option](https://github.com/Nightfirecat/runelite/commit/bf8c093c323e4a95694345c3fbb5d092be808a6b#diff-747d598f2e91f8dbf153e4e83e1409ef061afa97be7a6f7096501637ef37ad64), specifically on the naming. The wording is currently small enough to prevent the text from being clipped, however it is not the most clear. (regardless of the wording we go with, I'll likely expand on it in the description for clarity)

Could have the config option be "Show solution". Would "solution" be all encompassing in terms of all the overlays shown? With the drop-downs being
1. On read -- current plugin behavior
2. On pickup -- always update overlay for new clues entering inv
3. If inactive on pickup -- if no overlay is currently shown

Description could then clarify "on pickup" options won't work for beginner/master clues

Would be nice to think of another term than "pickup" though, since it'll also apply to:
1. New steps for currently held clues
2. Clues withdrawn from the bank/rewards
3. If we get stackable clues -- clues opened from a scroll box.

Nightfirecat:
One more thought I just had: with clue IDs associated to clues, we should also be able to display the clue solver sooner when clicking clues like we do with map clues, rather than waiting for the clue to show onscreen. I'll make the necessary changes for that to work as well.

TheLope:
@Nightfirecat Looks to be working well! The on-click identification is awesome. Glad this can be an improvement even for those that don't want to use on-pickup.
Speaking of which, even though I personally think folks should use it, I'm not sure setting `On Pickup` as the default option is best since users uninterested may find it disruptive. Then again could be a good way to introduce users to the feature, and hope they understand how to revert in case they don't want.

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/ClueScrollPlugin.java
```diff
@@ -352,7 +352,7 @@ public void onMenuOptionClicked(final MenuOptionClicked event)
 				|| itemComposition.getName().startsWith("Treasure scroll"))
 			{
 				clueItemId = itemComposition.getId();
-				updateClue(MapClue.forItemId(clueItemId));
+				updateClue(findClueScroll(clueItemId));
```
TheLope, line 355: Looks like this should be safe, as any potential misidentification would be corrected once the interface opens

PR #18307, opened 2024-09-22
Fix: Expeditious bracelet notify message

Kashif300:
this commit fixes the incorrect notify message for the expeditious bracelet when running out of charges. The expeditious bracelet notify message for running out of charges has two letters that are capitalized where they should not be.
Screenshot 2024-09-22 153921


raiyni:
Is there a problem with this? Or is it just grammar? Currently, it uses the same grammar as BRACELET_OF_SLAUGHTER_BREAK_TEXT

PR #18311, opened 2024-09-23
feat: add commands for nightmare scales

nasifra:
Additional chat commands for Nightmare, to allow players to easily access times for specific scales.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -2299,6 +2299,31 @@ public class ChatCommandsPlugin extends Plugin
 			case "nmare":
 			case "the nightmare":
 				return "Nightmare";
+			case "nm 1":
+			case "nm solo":
+			case "nightmare 1":
+				return "Nightmare solo";
+			case "nm 2":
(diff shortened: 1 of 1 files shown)


Adam-:
I don't think that this works because there is no code to actually track this. Have you tested this?

jb1361:
This is a duplicate of: https://github.com/runelite/runelite/pull/18148

Adam-:
It's not clear to me why this works. Can you add a test?

jb1361:
> It's not clear to me why this works. Can you add a test?

It will break other nightmare commands as it is.

Adam-:
> > It's not clear to me why this works. Can you add a test?
> 
> It will break other nightmare commands as it is.

How will it break the other nightmare commands?

jb1361:
> > > It's not clear to me why this works. Can you add a test?
> > 
> > 
> > It will break other nightmare commands as it is.
> 
> How will it break the other nightmare commands?
Its been a little while since I made my PR (https://github.com/runelite/runelite/pull/18148) but I remember it had to do with getting the longBossName having the team size in it. When i submitted that PR I stated 
-  Add Override to the longBossName method to support other base commands (lvl and kc) without having add a bunch of cases

jb1361:
@nasifra Would you want to collab on this? I Spoke with Adam in the discord and he had some issues with my PR as well.

PR #18312, opened 2024-09-24
chatcommands: fix gotr above 1000 & clean up the use of replaceAll

JZomDev:
Fixes the issue a user reported, their log shows that the value coming in was "1,841". 
[client (4).log](https://github.com/user-attachments/files/17106892/client.4.log)

This also removes the uses of replaceAll with replace for instances that regex is not used.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -578,21 +578,21 @@ public class ChatCommandsPlugin extends Plugin
 		if (matcher.find())
 		{
 			int floor = Integer.parseInt(matcher.group(1));
-			int kc = Integer.parseInt(matcher.group(2).replaceAll(",", ""));
+			int kc = Integer.parseInt(matcher.group(2).replace(",", ""));
 			setKc("Hallowed Sepulchre Floor " + floor, kc);
 		}
 
 		matcher = HS_KC_GHC_PATTERN.matcher(message);
 		if (matcher.find())
 		{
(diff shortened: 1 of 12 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
```diff
@@ -623,7 +623,7 @@ else if (result.equals("were defeated"))
 		matcher = GUARDIANS_OF_THE_RIFT_PATTERN.matcher(message);
 		if (matcher.find())
 		{
-			int kc = Integer.parseInt(matcher.group(1));
+			int kc = Integer.parseInt(matcher.group(1).replaceAll(",", ""));
```
iProdigy, line 626:
```suggestion
			int kc = Integer.parseInt(matcher.group(1).replace(",", ""));
```
to avoid regex overhead

iProdigy (approved):

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
JZomDev: The Huntor Rumour, Hollowed Sep and Grand Hallowed Coffin both use replaceAll, I just kept on the tradition. 

I do not think anybody is going to get 1,000,000 in any of these 3 fwiw. Just copied what others were doing.

iProdigy: re: multiple commas - `String#replace(String, String)` doesn't stop at the first occurrence

PR #18329, opened 2024-09-25
agility: add colossal wyrm courses

Macweese:
This PR adds the colossal wyrm agility courses (basic, advanced) to the agility plugin.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -920,4 +920,20 @@ public final class Varbits
 	public static final int BUFF_GOADING_POTION = 11294;
 
 	public static final int BUFF_PRAYER_REGENERATION = 11361;
+
+	/**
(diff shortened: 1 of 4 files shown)


dylanstanfield (commented):
Looks like the first course with the same regionId & end points location. What are your thoughts about modifying getCourse to add the ability to differentiate the two wrym courses based on xp gain - just to keep things in line with how they worked before instead of adding bespoke tracking for this?

dylanstanfield:
on further look - i see why you went with the approach you did.

quick question: does this progression checker account for the first few obstacles being the same for both courses? since we only set the course once in the session does this mean we will only ever display one or the other?

i think if we did want to keep things more in line with how they work for most courses though, we might try to track this as a single course with multiple totalXp options - then just display both options in the panel.

dylanstanfield:
@Macweese i think the difference with the wyrm course that shayzien doesn't have is that shayzien can still be split separate records based on the region id

PR #18330, opened 2024-09-25
skill calculator: add colossal wyrm courses

Macweese:
This PR adds the colossal wyrm agility courses to the skill calculator.

I chose not to append "course" to the description because it'd not fit in the card.


evilgenious448:
I would suggest having the Termites as the icon for both images, as you don't really pick up a bone squirrel. Many of the existing courses reuse the same icon such as the rooftops (marks of grace), and hallowed sepulchre (ring of endurance).

I think Shay having the separate icons for basic and advanced courses is the exception rather than the rule. 

Just my personal opinion, feel free to disagree 😄

evilgenious448:
Personally, I would make Termites for both and also change Shayzien to be the Shayzien banner. But when it comes to changes like these I like to check in with @Nightfirecat and see what he thinks.

PR #18335, opened 2024-09-26
Skill Calculator: Varlamore Part 2

evilgenious448:
- Add Huasca Farming Action
- Add Huasca Herblore Action
- Add Prayer Regeneration Potion Herblore Action
- Add Goading Potion Herblore Action
- Add Mastering Mixology Potions


evilgenious448:
@Nightfirecat Prior to moving onto the herbs and paste, I wanted to check with you regarding the making of potions. I have added the potions and the XP rates you get per potion. Rates are sourced from this table and are accurate:
image

However, as show in the video [here](https://cdn.discordapp.com/attachments/1288913931057889302/1288953569092833330/Recording_2024-09-26_160016.mp4?ex=66f70f0e&is=66f5bd8e&hm=f5508b33a51f17a074150c33f20e503afdb470b18a593ab97a3255e9d51c4291&), during the gameplay of the minigame, you are able to snag yourself some additional xp by performing certain actions such as spam clicking a machine or clicking at the right time. The XP is 14 per potion. Thus players who are more focused on the game can increase their XP. Is this something that we want to implement as a skilling bonus? Because a player who is sweaty and will be opting to get these extra points would likely want to factor them into their calculations.

If you think we should add it as a bonus, what is the best way to do it? Because it is a flat 14xp and each potion gives a different amount of XP, we would be unable to do a multiplier like we have been for other bonuses such as `Lumberjack Outfit (102.5%)`. It would have to be (+14) or (+14xp). I am not sure how you think is best but open to feedback.

Nightfirecat:
For the time being, I think it would be best to hold off any further mixology actions/bonuses, since I have a feeling the minigame may be updated in the coming week for balancing changes. Even if that is not the case, it would be better to handle any further changes separately from what's already here.

PR #18341, opened 2024-09-27
timers buffs: remove superfluous time and unit values

YvesW:
var timers do not require time and unit values - they get those from the varbit value.
Only pure GameTimers might require time and unit values in the GameTimer enum (assuming they're static timers like shadow veil, but not e.g. death charge).

Goading potion on master: https://youtu.be/cUcYLRMiqMU
Goading potion using this PR: https://www.youtube.com/watch?v=wmYAD7kIB3I
DFS using this PR: https://youtu.be/A2QbIoV3rKg
Mark of darkness cooldown using this PR: https://www.youtube.com/watch?v=JBu0zBkdVLg
For purging staff fixes, please refer to my WGS PR.

Additionally also changed removeOnDeath to false for all var timers.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameTimer.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameTimer.java
@@ -91,14 +91,14 @@ enum GameTimer
 	MARK_OF_DARKNESS_COOLDOWN(SpriteID.SPELL_MARK_OF_DARKNESS_DISABLED, GameTimerImageType.SPRITE, "Mark of Darkness cooldown", true),
 	PICKPOCKET_STUN(SpriteID.SKILL_THIEVING, GameTimerImageType.SPRITE, "Stunned", true),
 	FARMERS_AFFINITY(ItemID.GRAIN, GameTimerImageType.ITEM, "Farmer's Affinity", false),
-	SMELLING_SALTS(ItemID.SMELLING_SALTS_2, GameTimerImageType.ITEM, "Smelling salts", true),
-	MENAPHITE_REMEDY(ItemID.MENAPHITE_REMEDY4, GameTimerImageType.ITEM, "Menaphite remedy", true),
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/GameTimer.java
```diff
@@ -97,8 +97,8 @@ enum GameTimer
 	SILK_DRESSING(ItemID.SILK_DRESSING_2, GameTimerImageType.ITEM, "Silk dressing", 100, GAME_TICKS, true),
 	BLESSED_CRYSTAL_SCARAB(ItemID.BLESSED_CRYSTAL_SCARAB_2, GameTimerImageType.ITEM, "Blessed crystal scarab", 40, GAME_TICKS, true),
 	SPELLBOOK_SWAP(SpriteID.SPELL_SPELLBOOK_SWAP, GameTimerImageType.SPRITE, "Spellbook Reset", 120, ChronoUnit.SECONDS, false),
-	GOADING(ItemID.GOADING_POTION4, GameTimerImageType.ITEM, "Goading potion", 6, ChronoUnit.MINUTES, true),
-	PRAYER_REGENERATION(ItemID.PRAYER_REGENERATION_POTION4, GameTimerImageType.ITEM, "Prayer regeneration", 8, ChronoUnit.MINUTES, true),
+	GOADING(ItemID.GOADING_POTION4, GameTimerImageType.ITEM, "Goading potion", true),
+	PRAYER_REGENERATION(ItemID.PRAYER_REGENERATION_POTION4, GameTimerImageType.ITEM, "Prayer regeneration", true),
```
Nightfirecat, line 101: These should be changed to have `removeOnDeath` be false, as they're entirely controlled by var updates. If dying removes these timers, they would be removed as a result of vars being set to 0 after dying.

YvesW: My bad, that makes a lot of sense. I assumed Adam had tested it and for some reason the varbits would incorrectly persist while the effect ended at death. Turns out, the prayer regeneration even persists after death! <https://youtu.be/i50oLo3WN4o>

Issue #18128, opened 2024-08-05
timers: incorrect mark of darkness duration with purging staff equipped


YvesW:
### Type

Incorrect behavior

### Bug description

This got changed with the update of 24 July:

> Mark of Darkness will now last 5 times longer if the Purging Staff is equipped when cast.

https://secure.runescape.com/m=news/deadman-armageddon--while-guthix-sleeps-tweaks?oldschool=1

Also check what happens with the buff once you unequip the staff.

### RuneLite version

RL master, commit https://github.com/runelite/runelite/commit/3e31b10e3bd54da01efe42031b2b145157e7c9aa

cdfisher:
Just linking a relevant discussion from the other day: #18123

Flexo013:
> Also check what happens with the buff once you unequip the staff.

The buff duration is decided upon casting. Unequipping the staff has no effect.

YvesW:
> > Also check what happens with the buff once you unequip the staff.
> 
> The buff duration is decided upon casting. Unequipping the staff has no effect.

Don't worry, I tested that extensively when writing the fix. You can see my comments in the linked PR.

PR #18343, opened 2024-09-27
timers buffs: add WGS timers and buffs

YvesW:
Adds the WGS timers and buffs:
- Stone of Jas empowerment
  - Please DM me for Stone of Jas empowerment testing videos. Alternatively, if it's deemed too niche, the commit can be dropped. I assumed Jagex added it to the EHC buff bar with a reason though.
- Burn damage accumulated and next burn hit
  - EHC buff bar video: https://youtu.be/60e66gQ9Z5s
  - Test on RL: https://www.youtube.com/watch?v=dI0H-s5mYrE
- Adds smouldering heart & smouldering gland (TD consumables)
  - Unlike normally, I was not able to test the these, as they're relatively rare drops from TDs. The varbits are from script 5923, and 25 & 4 are from struct 965 and 966 respectively (param 1541, from script 5923).
  -  The wiki notes that teleporting away removes the effect from the smouldering gland -> ``removedOnDeath`` = false because  right now the timer is entirely controlled by var updates. I strongly suspect Jagex will set the varbit to 0 when teleporting away (like the Stone of Jas empowerment), among others based on the CS2 code, so EHC displays the buff correctly.
     - If this would not be the case, we'd have to check the regionIDs for the cave, as TDs are in >1 region, so just removing it on GameState.LOADING would not work.  
  - The wiki notes ``its effects carry over upon death`` for the smouldering heart ->  removedOnDeath = false.
- Fixes Mark of Darkness duration and cooldown timers when wearing the Purging Staff.
  - Tl;dr: If the staff is equipped when casting, the duration will be 5x. It can be unequipped (even banked) while the mark is active. Equipping it with a 1 minute mark active obviously does not prolong the duration.
  - Test: https://youtu.be/SQFkLgK8JTY and https://youtu.be/jAg2dU6IHog
  - Fixes https://github.com/runelite/runelite/issues/18128

Force pushes below are mostly to fix merge conflicts every time a new varbit, timer, or sprite gets added.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
@@ -996,7 +996,7 @@ public class TimersAndBuffsPlugin extends Plugin
 			}
 			else if (message.endsWith(MARK_OF_DARKNESS_MESSAGE))
 			{
-				createGameTimer(MARK_OF_DARKNESS, Duration.of(magicLevel, RSTimeUnit.GAME_TICKS));
+				createGameTimer(MARK_OF_DARKNESS, Duration.of(getMagicLevelMoD(magicLevel), RSTimeUnit.GAME_TICKS));
 			}
 			else if (message.contains(RESURRECT_THRALL_MESSAGE_START) && message.endsWith(RESURRECT_THRALL_MESSAGE_END))
 			{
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -615,6 +615,31 @@ else if (nextSuperAntifireTick - tickCount <= 0)
 			updateVarTimer(MOONLIGHT_POTION, moonlightValue, IntUnaryOperator.identity());
 		}
 
+		if (event.getVarbitId() == Varbits.STONE_OF_JAS_EMPOWERMENT && config.showOverload())
```
Nightfirecat, line 618: Instead of hijacking overload/prayer enhance timers, let's make one new config to cover Tormented demon buffs for Stone of Jas/Smouldering heart/Smouldering gland. (Technically the Stone of Jas empowerment is not related, but you only get the buff before fighting some tds, so it's close enough)

Nightfirecat (commented):
LGTM otherwise, everything has been tested and works.

PR #18359, opened 2024-09-30
CoordinateClue: Fix Ortus Farm and Proudspire enemies

Zoinkwiz:
A Bandosian Guard can spawn here, so this adds it as an option.

Screenshot 2024-09-30 142904

Edit: Seems only Isle of Souls should make use of individual enemies, so this now also fixes the proudspire enemy to be both spawns.


Felanbird (approved):

Nightfirecat:
Added in 086efe99fda6678f998b38ea79497f480cbe8709

PR #18365, opened 2024-10-01
Smooth color interpolation between item value tiers and Stackable item value scaling for coloring

daqniel:
# Summary

Two QoL and eye-candy I thought would be nice and simple additions to the Ground Items plugin.
- Toggleable via plugin settings.
- Leaves existing behavior fully intact if not enabled.
- Little to no extra maintenance needed if implemented.

#### Disclaimer
I'm new to Runelite development and not the freshest on Java so there may be some syntactical oddities to iron out. I also erred on the side of verbose code to keep things readable but if code size is a concern I can slim things down. If this functionality is something that the maintainers think fits well into the Ground Items plugin then we can make whatever adjustments needed.

# Smooth Item Value Color Interpolation

Functionality enabled with default item price and color values.
image

### Implementation Notes
* I added a small non-linearity in the color blending towards the high-end. Colors will blend until 70% of the higher-tier color and then snap to the new tier all at once when an item's value crosses the threshold. This is something subjective that I think helps maintain the meaning in setting your own price brackets. It also helps keep loot beams looking good without needing to implement the same behavior for them. 
    - If a tier is "green" then while the color will shift towards the next tier color as value increases, it'll still be a "greenish" color until that value crosses to the next tier entirely (and the loot beam with it).
    - The specifics of this can be adjusted or it can be removed entirely or made toggleable if desired.

## Why
I found myself pretty annoyed while doing some slayer tasks and wishing there was a visual distinction between the garbage items and the stackable/alchable items I actually want to pick up. I didn't want to mess with highlighting specific drops or changing my "Low value" price often so I thought something like this would be nice to have.

With the default value and color settings, see before and after:

This makes it a lot easier to pick out the stuff I want to pick up while still letting the higher value drops stand out by blending into the next color tier (like the dragon med).

## Possible Issues
- Not integrated with loot beams currently. I think it's debateable whether it'd be better to keep discrete lootbeam color tiers or have them matched to text color.
- Might work better if the default item color was a bit darker than #FFFFFF but that's something the user can customize to their liking.

# Stackable Item Value Scaling
image

## Why
For the same reason as above, I found myself wanting the smaller-but-valuable stackables (runes, coins, darts/javelin heads, etc.) to stand out more than the other garbage items. These drop in low-valued amounts but take up only one inventory slot so you often want to pick them up throughout a trip.

See before (no scaling+color smoothing) and after (2.0 scaling factor+color smoothing)

 

## Possible Issues
- A bit more subjective in terms of utility. Some people wouldn't like this to be on or would like their own custom scaling factor tailored to their personal setup.
- Current implementation just scales the stack value linearly, so it'll have a more noticeable effect on higher value stacks than lower value ones (e.g. a 9m cash stack will be colored like a 13.5m cash stack if the scaling factor is 1.5). The implementation could be adjusted to use some form of logarithmic scaling instead.
- Not integrated with loot beams currently (should be simple to add though). A high value stackable drop might have a color tier that doesn't match its lootbeam tier.
- Not as useful on it's own without smooth color interpolation enabled

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
@@ -137,7 +137,7 @@ public interface GroundItemsConfig extends Config
 	@ConfigItem(
 		keyName = "deprioritizeHiddenItems",
 		name = "Deprioritize Menu Hidden Items",
-		description = "Depriotizies the menu options for items which are hidden, requiring a right click to pick up.",
+		description = "Deprioritizes the menu options for items which are hidden, requiring a right click to pick up.",
 		position = 5
 	)
 	default boolean deprioritizeHiddenItems()
@@ -189,6 +189,9 @@ public interface GroundItemsConfig extends Config
 		return PriceDisplayMode.BOTH;
 	}
 
+
+
+
 	@ConfigItem(
 		keyName = "itemHighlightMode",
 		name = "Item Highlight Mode",
@@ -233,12 +236,28 @@ public interface GroundItemsConfig extends Config
 		return 0;
 	}
 
+	@ConfigItem(
+		keyName = "scaleStackableColoringPrice",
+		name = "Adjust stackable item value",
+		description = "Adjusts the price value used to determine coloring for stackable items by a specified factor",
+		position = 14
+	)
+	default boolean scaleStackableColoringPrice() { return false; }
+
+	@ConfigItem(
+			keyName = "stackableColoringPriceScaleFactor",
+			name = "Stackable item value factor",
+			description = "Configures the factor by which stackable item price values are adjusted for coloring",
+			position = 15
+	)
+	default double stackableColoringPriceScaleFactor() { return 1.25; }
+
 	@Alpha
 	@ConfigItem(
 		keyName = "defaultColor",
 		name = "Default items",
 		description = "Configures the color for default, non-highlighted items",
-		position = 14
+		position = 16
 	)
 	default Color defaultColor()
 	{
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/grounditems/GroundItemsConfig.java
```diff
@@ -356,18 +381,30 @@ default Color insaneValueColor()
...
+	{
+		return false;
+	}
+
 	@ConfigItem(
 		keyName = OWNERSHIP_FILTER_MODE,
 		name = "Ownership filter",
 		description = "Show all items, takeable items, or only your or your group's drops.",
-		position = 25
+		position = 25;
```
InfernoStats: This should not have a semicolon

daqniel: Accidental typo when I was resolving the merge conflict

daqniel:
Field example!
image

raiyni:
I don't really understand the point of scaling stackable items if you are already lerping the value between colors.

daqniel:
> I don't really understand the point of scaling stackable items if you are already lerping the value between colors.

The goal was to make small yet desirable stackables still stand out visually compared when compared to trash items with the same value. For example, I might pick up 2k worth of coins or blood runes on the ground but I wouldn't pick up an adamantite bar cause it takes up inventory space. So having this option to scale up stackables means the 2k stackable on the ground can stand out visually against the addy bar when it otherwise wouldn't.

It's something I find useful and nice to have but I also acknowledge it's subjective and if bloating the config menu is a concern I can remove it.

raiyni:
You can already use qty highlights for that

daqniel:
What are qty highlights? I've looked through the config menu for something like that but I'm not sure what you're referring to.

raiyni:
Ex: in your item highlights `blood rune>50`

daqniel:
Ah ok, didn't know that was a thing. Still, I think there's value in having an option that just works everywhere without having to manually configure each item. I've done some slayer and other content with this functionality enabled on my own compiled client and it's worked quite well without any extra configuration necessary.

daqniel:
I removed the value scaling on stackables as it's not the main focus of this PR.

PR #18368, opened 2024-10-01
Update: Attack Styles Default Plugin

Varietyz:
Added screen flashing option for toggled warned style.
Flashes the screen red for idication.


Varietyz:
image

Adam-:
I think it would be better to make an optional notification that can fire for using a warned style. That way it can use any of the available notification settings, one of which is flash, instead of having to code a bunch of 1-off notification options into the plugin.

Varietyz:
> I think it would be better to make an optional notification that can fire for using a warned style. That way it can use any of the available notification settings, one of which is flash, instead of having to code a bunch of 1-off notification options into the plugin.

Hey adam, this only flashes when a warned style is selected, it does not otherwise
i made barely any changes, only added a config option, slight implementation to plugin and  the flashoverlay class.
Not sure what u mean with the 1-off notification options

Nightfirecat:
Look at the idle notifier plugin and how the individual notification types are implemented. (eg. Hitpoints notification, Prayer notification, etc.)

Varietyz:
> Look at the idle notifier plugin and how the individual notification types are implemented. (eg. Hitpoints notification, Prayer notification, etc.)

All I did was add the flash class, implement 10 total lines into the main plugin file and added one config option.

This now flashes the screen red when an attack style becomes selected that was marked for warning.

I did not touch any other notifications or configs whatsoever.

So i'm not sure to what in my implementation you are referring to since the goal is to just have the screen flash red when a warned style is selected and a config to enable this flashing from happening or not.

I'm sorry, if i'm missing the point you're trying to tell me.

The plugin still functions as it did before, so whoever is the coder of this plugin has implemented the notification system.

Nightfirecat:
We're saying that rather than adding a bespoke screen flashing function, this change should be implemented using the Notification config system

evilgenious448:
The "notification overrides" that were added to Runelite in version [1.10.27 ](https://runelite.net/blog/show/2024-04-17-1.10.27-Release/) gives you a lot of customisation options for notifications beyond just a screen-flash. A number of plugins that fire notifications such as "idle notifier" utilise this system and NFC is suggesting that you implement notification overrides rather than simply a separate flash config, as it already comes bundled with a more customisable notification system.

Varietyz:
I've seen some people asking for a feature like this, mainly pure's, it's why i've committed it to the branch instead of releasing as a stand alone plugin.

If you'd rather have me bring it out as a standalone plugin, i'll do the necessary.
Just thought it would have migated some work and extra processing by submitting it this way.

I honestly haven't thought of any other notifications since it already has most of them?

PR #18384, opened 2024-10-04
cluescrolls: remove tags for emote clues before matching

SirGirion:
Noticed at least two emote clue scrolls were no longer solving due to Jagex randomly adding <br> tags. Given that there are multiple clues and they may end up removing the tags at a later date (or adding new ones), I opted to remove the tags from the clue text rather than change the text we have stored.

Some example debug log lines indicating that the clues failed to resolve are:
`2024-10-04 17:52:44 EDT [Client] INFO  n.r.c.p.cluescrolls.ClueScrollPlugin - Unknown clue text: Salute in the centre of the mess hall.<br>Beware of double agents! Equip a rune halberd<br>rune platebody and an amulet of strength.`
`2024-10-04 14:20:44 EDT [Client] INFO  n.r.c.p.cluescrolls.ClueScrollPlugin - Unknown clue text: Cheer at the top of the agility pyramid.<br>Beware of double agents! Equip a blue mystic robe<br>top and any rune heraldic shield.`


Adam-:
Both of these clues failed to match due to removing a `,`, not due to the tags. The tags are already removed when matching. One of these was already fixed earlier in 7c72a32da0e9cee38ff45bd6f3c70683af9366a0

Issue #3090, opened 2018-05-21
Clue coordinates - Make the scroll icon flash on the world map


MalakaJones:
I feel like making the clue scroll icon flash on the world map would make it easier to spot. Plus it's a relatively simple QoL improvement.

I'm talking about these icons:

![image](https://user-images.githubusercontent.com/15826583/40332984-359f0b3c-5d24-11e8-9005-efc4280cfee3.png)

Edit: Just realized you can click on the edge of the map and it brings you over it. Still would be nice.

Honestidade:
Instead of this or in addition to it, having the world map automatically take you to the clue location would be nice.(Possibly with a button added onto the clue scroll to open the map @ the clue location similar to osb)

PR #18397, opened 2024-10-06
clues: highlight world map icon to make it easier to see

alexanderhenne:
Solves https://github.com/runelite/runelite/issues/3090

https://github.com/user-attachments/assets/012380bf-1bcf-4378-9119-3e6a8f8d62cd


alexanderhenne:
@Macweese It's a circle to be generic for all types of map points. But specifying a highlight image instead might be worth it to improve the visuals.

One question, doesn't the client tick vary depending on your FPS? Or is that a different tick variable?

Issue #18379, opened 2024-10-04
[Timers & Buffs] Moonlight potion setting is inconsistent, incorrectly spawns additional timer


fuzxi:
### Type

Incorrect behavior

### Operating System

None

### Bug description

With 70+ Herblore, the Moonlight potion in the Moons of Peril dungeon acts as a Super Defence potion. Drinking it with the Moonlight potion and Divine potion timers enabled in the Timers & Buffs plugin _sometimes_ spawns two timers, one for the Moonlight potion and one for the Super Defence (note that these are the same effect and the divine timer shouldn't be spawned at all - it's like a divine Super Combat potion spawning three separate timers for divine Attack, Strength, and Defence potions).
 Sometimes if the Moonlight potion timer has been disabled and then re-enabled, drinking a Moonlight potion will instead only spawn the divine Super Defence timer.

Sometimes if the Moonlight potion timer has been disabled, drinking a Moonlight potion will only spawn the divine Super Defence timer. 
Sometimes if the Moonlight potion timer has been disabled, drinking a Moonlight potion will spawn no timer at all.

### RuneLite version

RuneLite version: 1.10.38.1
Launcher version: 2.6.10

Nightfirecat:
Do you know any way to reproduce these behaviors consistently?

fuzxi:
Both timers will reliably spawn the first time you drink a Moonlight potion with both timers enabled for that instance of the client. Meaning, close RuneLite, re-open it, log in, sip potion, two timers. This is unexpected behavior already as outlined above.
After toggling the Moonlight potion setting off and on, and the first boost expires on its own (5 minutes), only the Moonlight potion timer will spawn until the client is restarted (correct behavior).
I haven't found a way to reliably reproduce the other issues yet.

fuzxi:
The issue where there's no timer at all seems to occur when you have the Moonlight timer setting enabled and drink a Moonlight potion on the same tick that your last dose was about to expire. The Moonlight timer shows up for a tick reading 5:00, then disappears. Drinking another dose in this state will spawn only a divine Super Defence timer.

Nightfirecat:
Thanks for the additional info. We'll give this some testing and see what we can do

YvesW:
Thank you for the precise reproduction steps; I was able to reproduce both problems because of that! Turns out the second bug is an interesting edge case with the order in which the varbits are transmitted in that exact scenario.
[...]

PR #18399, opened 2024-10-06
timers buffs: fix moonlight potion issues

YvesW:
- Fixes the divine super defence timer showing up next to the moonlight timer, when drinking the first moonlight potion of the session
  - Re-introduces the full moonlightValue +1 hack in all its glory!
  - Video test: https://youtu.be/rIRFAuIrn-I
- Fixes the moonlight potion instantly being removed and a divine super defence being added, when you sip a moonlight potion on the exact tick it ran out (i.e. the moonlight varbit hits 0 and you click on the potion)
  - The problem occurs because in this very specific scenario, the varbit "desync" that we already knew about happens, but after the first tick, the transmission order of ``Varbits.DIVINE_SUPER_DEFENCE`` and ``Varbits.MOONLIGHT_POTION`` is also switched! See below for a more in-depth explanation.
  - Video test: https://youtu.be/ZoRTZ6MF6Dw
     - You can also tell when playing the end of the video back at slow speed that the moonlight buff is actually lost when ``Varbits.DIVINE_SUPER_DEFENCE`` hits 0.

Doesn't add any unit tests because I could not figure out how to use e.g. ``infoBoxManager.getInfoBoxes().size()`` properly in tests, as I have a lack of experience with unit tests due to my dislike for them. Jut adding ArgumentCaptors for InfoBoxes (``addInfoBox`` call) and Predicates (``removeIf`` call) felt pretty useless and brittle. Feel free to add useful unit tests.

Fixes https://github.com/runelite/runelite/issues/18379

### Double timers
- The double timers (moonlight & divine super defence) on a fresh client boot are caused by https://github.com/runelite/runelite/blob/3cc0c4c967360512b9c1ba849e4e8ff827a299ba/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L483 being ``0 >= 500`` since ``Varbits.DIVINE_SUPER_DEFENCE`` updates before ``Varbits.MOONLIGHT_POTION``.
  - Other divine potions, such as the divine battlemage, remove the divine super defence timer for that reason. E.g. https://github.com/runelite/runelite/blob/3cc0c4c967360512b9c1ba849e4e8ff827a299ba/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L554-L557
  - A divine super def timer does not pop up after the first moonlight potion of the session.
     - The first time the divine super def GameTimer is still null, and ``updateVarTimer(Varbits.DIVINE_SUPER_DEFENCE)`` will take the second branch, creating a new GameTimer https://github.com/runelite/runelite/blob/3cc0c4c967360512b9c1ba849e4e8ff827a299ba/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L1306-L1311
     - After the first moonlight potion, it'll take the third branch, which does not create a new GameTimer https://github.com/runelite/runelite/blob/3cc0c4c967360512b9c1ba849e4e8ff827a299ba/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L1312-L1316
     I assume when the divine super def runs out, ``InfoBoxManager::cull`` removes the infobox since ``Timer::cull`` is true at that point. The GameTimer does not get removed however, since ``updateVarTimer(DIVINE_SUPER_DEFENCE, 0)`` is never called. I've not looked into this further than this.

### Only the divine super defence showing up
I was able to recreate only the divine super defence showing up when drinking a moonlight potion after it just expired on master. This seems to be caused by how interesting the varbit behaves...

We were already aware of the "normal" desync that can happen when you drink a moonlight potion while already under its effect:
Normal desync

As you can see in tick 1352, ``Varbit.DIVINE_SUPER_DEFENCE`` gets updated, then ``Varbits.MOONLIGHT_POTION`` twice. From tick 1353 onwards ``Varbit.DIVINE_SUPER_DEFENCE`` gets updated followed by ``Varbits.MOONLIGHT_POTION``. The former is 1 greater than the latter. This is the desync we already knew,

However, it turns out that in the edge case of drinking a moonlight potion right as the varbit hits 0 (probably while not already having the above "normal" desync), another type of desync can happen!
I was printing varbit changes in VarbitChanged at the time I reproduced the issue on master:
```
client.getTickCount() Varbit varbitId changed to varbitChanged.getValue()
163 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 2
163 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 2

164 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 1
164 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 1

165 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 0
165 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 0

166 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 500
166 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 500
166 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 499

167 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 498
167 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 499

168 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 497
168 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 498

169 Varbit 10029 (Varbits.MOONLIGHT_POTION) changed to 496
169 Varbit 8431 (Varbits.DIVINE_SUPER_DEFENCE) changed to 497
```
As you can see, in tick 166 we get a divine super def and 2 moonlight updates, as expected with the known desync. However, afterwards we keep getting ``Varbits.MOONLIGHT_POTION`` updates before ``Varbits.DIVINE_SUPER_DEFENCE``!
This means that e.g. in tick 167:
- When VarbitChanged fires for Varbits.DIVINE_SUPER_DEFENCE https://github.com/runelite/runelite/blob/3cc0c4c967360512b9c1ba849e4e8ff827a299ba/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L483-L491
  - The early return does not happen: 498 >= 499 -> false
  - The MOON_LIGHTPOTION vartimer gets removed: 498 < 499 -> true
- When VarbitChanged fires for Varbits.MOONLIGHT_POTION fires, the code in this PR that removes the divine super def vartimer does not fire for that tick.
  - ``moonlightValue++`` does not proc: 500 == 498 + 1 -> false
  - Since the first tick of varbit updates is still divine super def -> moonlight potion -> moonlight potion, this should not matter for fixing the first issue. However, this will cause the timer to be incorrect.

This sounded so insane to me that it sounded fake, so I managed to catch it on video later on: https://youtu.be/sjnJWn_APYQ
(this is my feature branch after ``timers buffs: correctly remove divine super defence timer when drinking moonlight potion`` but before writing any other code, i.e. no 2nd/3rd commit).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
@@ -611,6 +611,12 @@ public class TimersAndBuffsPlugin extends Plugin
 			{
 				moonlightValue++;
 			}
+			// Varbits.DIVINE_SUPER_DEFENCE updates before Varbits.MOONLIGHT_POTION. Thus, the early return that
+			// prevents creation of the divine super defence timer does not work for the first game tick
+			if (client.getVarbitValue(Varbits.DIVINE_SUPER_DEFENCE) == moonlightValue)
+			{
+				removeVarTimer(DIVINE_SUPER_DEFENCE);
+			}
 
 			updateVarTimer(MOONLIGHT_POTION, moonlightValue, IntUnaryOperator.identity());
 		}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
@@ -479,13 +479,19 @@ public class TimersAndBuffsPlugin extends Plugin
 				|| client.getVarbitValue(Varbits.DIVINE_BATTLEMAGE) > event.getValue()
 				// When drinking a dose of moonlight potion while already under its effects, desync between
 				// Varbits.MOONLIGHT_POTION and Varbits.DIVINE_SUPER_DEFENCE can occur, with the latter being 1 tick
-				// greater
-				|| client.getVarbitValue(Varbits.MOONLIGHT_POTION) >= event.getValue())
+				// greater due to Varbits.MOONLIGHT_POTION ticking down twice in its first tick.
+
+				// When the user drinks a dose of moonlight potion at the exact tick the potion runs out, i.e. right
+				// after Varbits.MOONLIGHT_POTION == 0, the first tick behaves exactly as the above described desync.
+				// However, after the first tick, Varbits.MOONLIGHT_POTION is suddenly transmitted BEFORE
+				// Varbits.DIVINE_SUPER_DEFENCE every tick. Thus, when the latter is transmitted, the
+				// Varbits.MOONLIGHT_POTION value might be 1 less than the Varbits.DIVINE_SUPER_DEFENCE value.
+				|| client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 >= event.getValue())
 			{
 				return;
 			}
 
-			if (client.getVarbitValue(Varbits.MOONLIGHT_POTION) < event.getValue())
+			if (client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 < event.getValue())
 			{
 				removeVarTimer(MOONLIGHT_POTION);
 			}
@@ -607,12 +613,19 @@ public class TimersAndBuffsPlugin extends Plugin
 			int moonlightValue = event.getValue();
 			// Increase the timer by 1 tick in case of desync due to drinking a dose of moonlight potion while already
 			// under its effects. Otherwise, the timer would be 1 tick shorter than it is meant to be.
-			if (client.getVarbitValue(Varbits.DIVINE_SUPER_DEFENCE) == moonlightValue + 1)
+
+			// When the user drinks a dose of moonlight potion at the exact tick the potion runs out, i.e. right after
+			// Varbits.MOONLIGHT_POTION == 0, the first tick behaves exactly as 'normal' desync. However, after the
+			// first tick, Varbits.MOONLIGHT_POTION is suddenly transmitted BEFORE Varbits.DIVINE_SUPER_DEFENCE every
+			// tick. Thus, when the former is transmitted, there might be a difference of 2 between the varbit values.
+			final int divSupDefValue = client.getVarbitValue(Varbits.DIVINE_SUPER_DEFENCE);
+			if (divSupDefValue == moonlightValue + 1 || divSupDefValue == moonlightValue + 2)
 			{
 				moonlightValue++;
 			}
-			// Varbits.DIVINE_SUPER_DEFENCE updates before Varbits.MOONLIGHT_POTION. Thus, the early return that
-			// prevents creation of the divine super defence timer does not work for the first game tick
+			// Varbits.DIVINE_SUPER_DEFENCE updates before Varbits.MOONLIGHT_POTION (at least during its first tick).
+			// Thus, the early return that prevents creation of the divine super defence timer does not work for the
+			// first game tick.
 			if (client.getVarbitValue(Varbits.DIVINE_SUPER_DEFENCE) == moonlightValue)
 			{
 				removeVarTimer(DIVINE_SUPER_DEFENCE);


runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -479,13 +480,19 @@ else if (nextSuperAntifireTick - tickCount <= 0)
...
+				|| client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 >= event.getValue())
 			{
 				return;
 			}
 
-			if (client.getVarbitValue(Varbits.MOONLIGHT_POTION) < event.getValue())
+			if (client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 < event.getValue())
 			{
 				removeVarTimer(MOONLIGHT_POTION);
 			}
```
Nightfirecat, line 498: While slightly unrelated to the changes at hand in this PR, I'm trying to remember why I recommended adding this before. What actually happens when a divine super defence potion is drank before a moonlight potion ends? If I recall correctly, it causes the defense boost to drain normally for the duration of the divine effect (down to normal boost levels, I assume), then removes the boost at the end of it.

Would you mind confirming that behavior? I'll want to either remove this (if both timers should be preserved) or add some code to the moonlight timer branch to avoid re-adding it on further var updates.

Nightfirecat:
Besides the one outstanding question above, I think this is in a good place now. @YvesW can you do some re-testing here--and check on the aforementioned question--and confirm it's working as expected?

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
YvesW: Your original quote is https://github.com/runelite/runelite/pull/17662#discussion_r1677034740
> Also, similarly to how drinking a moonlight potion removes existing divine super defence potion infoboxes, drinking a divine super defence potion (or having this var update from any of the other divine potions which provide its effect) should remove the moonlight potion's infobox, since it preserves the defence boost beyond the 5 minutes provided by the moonlight potion.

To be frank, it had been a couple months since I had worked on this when you requested the change (I probably didn't have all the very specific interactions still in my head), it was a quite direct request (i.e. not a question), and it didn't seem like it'd e.g. break anything significantly, so I probably didn't review your requested changes as critically and test them as extensively as I should have at the time. My apologies.


The behaviour you currently describe is indeed what I seem to remember, and it's also what I wrote down in one of the posts of my previous PR, I think. I re-tested it:
Drinking a divine super def after a moonlight potion using this feature branch:
- The defence buff ticks down to divine super def levels (118) and gets removed when the divine super def ends. Video of test: https://youtu.be/qiPKAjWhjvw
  - It really ticks down to 118. It's not like the def level gets set to 118 when the moonlight timer ends. In the video, you can see my def is boosted to 119 while the moonlight effect has already worn off.
- The moonlight infobox is currently not properly removed when the super divine def is consumed after the moonlight potion, since the moonlight timer is immediately being readded (the early return when ``Varbits.MOONLIGHT_POTION`` updates is missing). This could be seen in the previous video, alternatively with varbits: https://youtu.be/oXqQERuRMcM


I don't really mind either...
- showing both the moonlight timer and divine super defence (removing the if statement you mentioned), or...
- actually removing the moonlight timer by adding an early return when ``Varbits.MOONLIGHT_POTION`` updates, and adding the test you wrote.

For now I've pushed a commit that does the former. Feel free to let me know if you prefer the latter instead!
I'll do more extensive re-testing to confirm everything is still working as expected soon(tm).

YvesW:
Turns out the config commit you wrote can be problematic.
While the desync described in this PR is happening:
- When ``showMoonLightPotion`` is toggled to false, a divine super defence timer is added
  - Previously this timer was not added since there was no config check, so it'd just return early
- When ``showMoonLightPotion`` is then toggled to true again, the moonlight timer is added again (expected behaviour)
- The divine super defence timer should then be removed, but it is not, as ``removeVarTimer(DIVINE_SUPER_DEFENCE)`` does not get triggered because of the desync combined with the changed transmission order:
https://github.com/YvesW/runelite/blob/b7c6665de5d9541d55a6fa05a78f4dcd4697fd4d/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L622-L628
It will check e.g. ``if (497 == 498)`` (497 is 496++ in this example).

I actually thought about this when writing the original code for this PR, and updated the ``moonlightValue`` if statement that determines the ``moonlightValue``, as the timer value would otherwise be incorrect:
https://github.com/YvesW/runelite/blob/b7c6665de5d9541d55a6fa05a78f4dcd4697fd4d/runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java#L618-L621

However, I didn't change it for the ``removeVarTimer(DIVINE_SUPER_DEFENCE)`` if statement, as I assumed it'd never happen because I'd already prevented the divine super defence from being added in this scenario (the early return). For that reason I for example added ``(at least during its first tick)`` in one of my code comments. The config toggle changed this however.

------------------------------
Video of the problem: https://youtu.be/C3MCu2inumA
Interestingly, this video also displays that you can't get the desync + varbit transmission order problem after a 'normal' desync that happens when drinking a moonlight sip while already under its effects.
  - I drank a moonlight sip while ``Varbits.MOONLIGHT_POTION`` was 0 (``Varbits.DIVINE_SUPER_DEFENCE`` was 1), which just caused the 'normal' desync without a change in transmission order.
  - I drank a sip of moonlight potion while ``Varbits.DIVINE_SUPER_DEFENCE`` was 0. ``Varbits.MOONLIGHT_POTION`` had already become 0 the tick before due to 'normal' desync. This just caused the normal moonlight potion varbit transmission without desync or varbit transmissions changes (i.e. drinking moonlight potion while being unpotted).
------------------------------

I've now fixed this. I'll add a ``testMoonlightDesyncOrderToggle()`` test, and do some more extensive tests soon(tm).

runelite-client/src/main/java/net/runelite/client/plugins/timersandbuffs/TimersAndBuffsPlugin.java
```diff
@@ -488,16 +487,11 @@ else if (nextSuperAntifireTick - tickCount <= 0)
 				// However, after the first tick, Varbits.MOONLIGHT_POTION is suddenly transmitted BEFORE
 				// Varbits.DIVINE_SUPER_DEFENCE every tick. Thus, when the latter is transmitted, the
 				// Varbits.MOONLIGHT_POTION value might be 1 less than the Varbits.DIVINE_SUPER_DEFENCE value.
-				|| (showMoonlightPotion && client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 >= event.getValue()))
+				|| (config.showMoonlightPotion() && client.getVarbitValue(Varbits.MOONLIGHT_POTION) + 1 >= event.getValue()))
```
Nightfirecat, line 490: The order of these checks should be flipped; `getVarbitValue()` is more or less free, calls to config are not.

YvesW:
Should be all good now!

No problems found during the following tests: https://youtu.be/_bxJ-fi_tkU
- Drinking a moonlight potion, and toggling config
- Drinking a moonlight potion while under its effects, and toggling config ('normal' desync)
- Drinking a moonlight potion when it just times out, and toggling config (desync + varb transmission order changes)
- Drinking a divine super defence during the desync + varb transmission order changes related to this PR, plus toggling config
- Drinking a moonlight potion followed by divine super def, and toggling config

Nightfirecat:
My experience running this code in my dev branch for the past several months is that it is working as expected. With that said, [can the last commit in this PR](https://github.com/runelite/runelite/pull/18399/commits/cbd916e9e500553d14333d69f078d8e44715de15) have a test case added for it? (and possibly be squashed with the previous commit?)

After that, and the needed gameval updates, I think this will be good to go.

PR #18405, opened 2024-10-09
Improvements to ChatboxItemSearch

YeahImRose:
Current goals:

- [X] - Allow extra rows of items to be displayed
- [X] - Allow keyboard navigation of extra rows
- [x] - Allow mouse wheel scrolling

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java
+++ b/runelite-client/src/main/java/net/runelite/client/game/chatbox/ChatboxItemSearch.java
@@ -55,7 +55,8 @@ public class ChatboxItemSearch extends ChatboxTextInput
 	private static final int ICON_HEIGHT = 32;
 	private static final int ICON_WIDTH = 36;
 	private static final int PADDING = 6;
-	private static final int MAX_RESULTS = 24;
+	private static final int MAX_RESULTS = 120;
+	private static final int RESULTS_PER_LINE = 12;
 	private static final int FONT_SIZE = 16;
 	private static final int HOVERED_OPACITY = 128;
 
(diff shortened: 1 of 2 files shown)


YeahImRose:
All my initial goals on this are done. My only hangups are as follow:

- I'm unsure if the scroll bar visible in the normal chat menu should be added (don't want to make design choices when I don't know the systems that well)
- I'm unsure on the MAX_RESULTS value. It seems fine to me at 120, but I don't know if we want it lower or higher or customizable per instance of the class

Otherwise all good for me!

raiyni:
This should be trying to build scrollbars like any other jagex interface probably.

YeahImRose:
> This should be trying to build scrollbars like any other jagex interface probably.

Probably, but that's why I was kinda doubting haha. Do you (or anyone else) happen to have any examples of a scrollbar actually being implemented in a widget? I see RuneLiteScrollBarUI, but I don't see any uses of it, and haven't been able to find any other examples of a plugin or something using the Jagex-style scrollbar.

raiyni:
I believe it would just take calling the `[proc,scrollbar_vertical]`  script with the right parameters

raiyni:
I liked the idea of reusing the GE item search for this, there's more to look into for it but I think I like it.


https://github.com/user-attachments/assets/7f7d238e-8f2e-455d-8220-995660387aab

YeahImRose:
I really do like the way that one looks. I've kinda been running up against a brick wall trying to get the scrollbar to apply to a widget that isn't a layer. Not sure if it's even possible to be honest, because every message I've read in the discord about it makes me think it's not.

raiyni:
I believe https://github.com/raiyni/runelite/commits/gelike-search/ covers all usage

YeahImRose:
> I believe https://github.com/raiyni/runelite/commits/gelike-search/ covers all usage

Unfortunately, implementing it that way prevents use cases where you want non-GE items, which was possible with the old implementation
image

raiyni:
It supports non-tradable items just fine


https://github.com/user-attachments/assets/fc887d4c-dfc7-463c-b8cf-41ffaf81994b

YeahImRose:
Using the exact same plugin code which made my earlier screenshot, just swapping in your version of ChatboxItemSearch does not include the same results. This is what I'm talking about
image

raiyni:
looks fine for me except collapsing the same item models, maybe try just building my branch directly as I don't know what you copy and pasted

image

YeahImRose:
Ah yep, building straight from yours does work! I'd be a-ok with it. Should I just close this and you open another PR or is there a way to merge yours in easily?

YeahImRose:
Ah yeah, seems like there's no way to change source. I'll just close this and let you make another https://stackoverflow.com/a/42381583

PR #18406, opened 2024-10-09
Update: Re-order Sidebar (Drag/Drop)

Varietyz:
Added functionality to drag and drop the components in the sidebar. Saving the order or shutdown and on drag/dropping, ensuring the user always has his preferred layout available. Saved to %user%/.runelite/sidebar/sidebar_order.txt i went for a txt file so users can change their order via the file aswel.

This has long been requested!
first_tab_reorder_test


cdfisher:
Given that we quite strongly discourage any manual editing of one's config, I don't think changing the order by editing the underlying file is necessarily desirable.

Varietyz:
> Given that we quite strongly discourage any manual editing of one's config, I don't think changing the order by editing the underlying file is necessarily desirable.

this was up for 6 years.
how is this not a good contribution?

cdfisher:
> > Given that we quite strongly discourage any manual editing of one's config, I don't think changing the order by editing the underlying file is necessarily desirable.
> 
> this was up for 6 years. how is this not a good contribution?

That's not what I said. I'm responding to

> i went for a txt file so users can change their order via the file aswel.

because this is something we specifically discourage.

Varietyz:
> > > Given that we quite strongly discourage any manual editing of one's config, I don't think changing the order by editing the underlying file is necessarily desirable.
> > 
> > 
> > this was up for 6 years. how is this not a good contribution?
> 
> That's not what I said. I'm responding to
> 
> > i went for a txt file so users can change their order via the file aswel.
> 
> because this is something we specifically discourage.

would you prefer a .data file? :)

fourtwentycoding:
This is so unbelievably needed. Its insane I can't drag an icon in 2024.

Varietyz:
> This is so unbelievably needed. Its insane I can't drag an icon in 2024.

its quite the undertaking and alot of trial and error implementing it.

I'm having a bit less time to work on it as i would like to have. i'm still working on it and i feel like i submitted this PR a little too hasty, so i've closed it till i have a proper implementation according to runelite's standards :D

adamk33n3r:
One potential issue I see here, aside from all the unnecessary formatting changes (and images?), is adding a new plugin will just add it to the end of the sidebar, ignoring the plugin's set `priority`. Personally, I would be fine with this change. But it is a "breaking" change in that plugins can no longer choose their position. I think it should be the user's choice where the nav button is anyway. I'm trying to think of a way to keep both. It would probably have to be saved as a diff of sorts. So, by default, it behaves the same as now. But then if you drag an icon, only it will be saved as something like `after:Configuration`. And maybe right clicking will reset it.

I see it's been a couple months since activity. I may take a stab at it.

Varietyz:
> One potential issue I see here, aside from all the unnecessary formatting changes (and images?), is adding a new plugin will just add it to the end of the sidebar, ignoring the plugin's set `priority`. Personally, I would be fine with this change. But it is a "breaking" change in that plugins can no longer choose their position. I think it should be the user's choice where the nav button is anyway. I'm trying to think of a way to keep both. It would probably have to be saved as a diff of sorts. So, by default, it behaves the same as now. But then if you drag an icon, only it will be saved as something like `after:Configuration`. And maybe right clicking will reset it.
> 
> I see it's been a couple months since activity. I may take a stab at it.

I haven't been able to work on it past few weeks due to complications..
If you would need anything from me, dont hesitate to reach out.
I still have the by now out-dated client files that hold the rough base of it. if it could serve as any help.

Maybe we can work on it together when things are looking better for me again. (which i hope is rather soon..)

adamk33n3r:
> I haven't been able to work on it past few weeks due to complications.. If you would need anything from me, dont hesitate to reach out. I still have the by now out-dated client files that hold the rough base of it. if it could serve as any help.

I did take a stab at it a little bit, and had a conversation in the discord about potential solutions to respect plugins default priority here https://discord.com/channels/301497432909414422/419891709883973642/1317985884343238696 it got a little more complicated than I originally thought so I put it back down. 

> Maybe we can work on it together when things are looking better for me again. (which i hope is rather soon..)

I would for sure be down for helping when you get back around to it! You can message me on discord if you want.

propagating:
> > I haven't been able to work on it past few weeks due to complications.. If you would need anything from me, dont hesitate to reach out. I still have the by now out-dated client files that hold the rough base of it. if it could serve as any help.
> 
> I did take a stab at it a little bit, and had a conversation in the discord about potential solutions to respect plugins default priority here https://discord.com/channels/301497432909414422/419891709883973642/1317985884343238696 it got a little more complicated than I originally thought so I put it back down.
> 
> > Maybe we can work on it together when things are looking better for me again. (which i hope is rather soon..)
> 
> I would for sure be down for helping when you get back around to it! You can message me on discord if you want.

Is there a reason a simpler solution is not viable? 

From the outside it seems like a fairly easy way to make this work is to have our 'default priorty' and a 'user priority' that is created and set when a user moves an icon in the list. So changing the order in the list is just a change in priority, and when a new plugin is installed it puts itself in the list based on its default priority as it doesn't have an assigned user priority yet, and the user can then choose to move it or not. Alternatively, the user can set an option for things without user priorities to be at the top or bottom of the plugin list, where they are then sorted by their default priority in that area. Identical default priorities can be sorted according to plugin names, ids, or whatever jvm decides is the proper order.

This feature would still be hugely beneficial in many ways and I would love to take a stab at its implementation, I'm not great at Java, but there are enough tools out there to help translate and convert skills between programming languages.

hauber794:
Would love to see this implemented!

PR #18442, opened 2024-10-16
What Pet is that Anyway? - Overlay for !pets command - Sololegends

sololegends:
# What Pet is that Anyway?

Adds an on-hover tooltip for everyone's `!pets` command so you can see what the heck those little icons actually are.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsOverlay.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2024, Ben <runelite@sololegends.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsOverlay.java
```diff
@@ -0,0 +1,154 @@
...
+			if (messageLines.getBounds().contains(line.getBounds())
+					&& line.getText() != null
+					&& line.getText().matches(".*?Pets: \\([0-9]+\\) <img=[0-9]+>.*"))
+			{
+				// Get the bounds
+				Rectangle bounds = line.getBounds();
+				// Get mouse position
+				Point mousePos = client.getMouseCanvasPosition();
+				// If mouse in message bounds
+				if (bounds.contains(mousePos.getX(), mousePos.getY()))
```
Nightfirecat, line 96: Instead of running bounds checking every frame a pets chat message exists in a player's chat history (visible or otherwise), this could also be gated on checking that the hovered widget is a chat message and checking the hovered message's text.

sololegends: Did a bunch of measuring: 

Limitation of the menu method is you can't hover your own pets which would make for a weird UX. 
However it obviously is faster. 

Compared against my optimized is visible checking it is: 
## Profiling
*72 samples averaged out continuously over 2 ticks at 60fps. All samples with a full chatbox and multiple pets commands*

### Only Others' Pets (onMenuEntryAdded)
Idle: ~0.5ms
Active: ~5.3ms

### Optimized Visibility Checking
*Includes manual pet icon id parsing w/o regex*
Idle: ~1ms
Idle (mouse in chat box): ~8.5ms
Active: ~9.2ms

Delta from empty to full chat box is ~2ms

Delta in JVM instructions from the original bound check and the new visible checking is ~147 instructions down to ~36

PR #18449, opened 2024-10-17
Nex drop rate

chrihb:


Adam-:
You're in the wrong place and instead are looking for the pluginhub https://github.com/runelite/plugin-hub/blob/master/README.md

Issue #14032, opened 2021-08-17
Option for "Tile Indicators" to have a "Only Show True Tile while moving" 


Semicent12:
Basically have an added setting for the plugin "Tile Indicators" to have the True Tile option to only show while your character is moving, but never while standing still.

PR #18483, opened 2024-10-30
tileindicators: Option to only show true tile while moving

Skretzo:
Adds a config option to only highlight the true tile while the player is moving. This means that no tile will be drawn under the player when the player is standing still, which can be a bit annoying when admiring your mole slippers.

Implements #14032

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/tileindicators/TileIndicatorsConfig.java
@@ -167,12 +167,24 @@ public interface TileIndicatorsConfig extends Config
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


geheur:
Is this different from "Corner Tile Indicators" on the plugin hub?

Skretzo:
> Is this different from "Corner Tile Indicators" on the plugin hub? It has an option to hide the true tile when not moving.
> 
> Or #16577?

Would be nice to have this feature in core, but 16577 is better. Although it would be nice if the fadeout can be set to 0 milliseconds instead of a [minimum of 600](https://github.com/runelite/runelite/pull/16577/files#diff-c84ac5c1830761faaa2210f37001d776283b9acc6ee0e6e8d742138221757efaR229).

Skretzo:
Closing in favour of #16577

PR #18500, opened 2024-11-07
hiscore: Use cache icons in hiscore panel

Nightfirecat:
These images from cache will match those used in the mobile client's hiscore panel.

image  
image (1)  
image (2)


runelite-client/src/main/java/net/runelite/client/plugins/hiscore/HiscorePanel.java
```diff
@@ -328,31 +332,13 @@ private JPanel makeHiscorePanel(HiscoreSkill skill)
...
-
-		String skillName = (skill == null ? "combat" : skill.name().toLowerCase());
-		String skillIcon = directory + skillName + ".png";
-		log.debug("Loading skill icon from {}", skillIcon);
-
-		label.setIcon(new ImageIcon(ImageUtil.loadImageResource(getClass(), skillIcon)));
+			// Icons are all 25x25 or smaller, so they're fit into a 25x25 canvas to give them a consistent size for
+			// better alignment. Further, they are then scaled down to 20x20 to not be overly large in the panel.
+			final BufferedImage scaledSprite = ImageUtil.resizeImage(ImageUtil.resizeCanvas(sprite, 25, 25), 20, 20);
+			label.setIcon(new ImageIcon(scaledSprite));
```
Adam-, line 340: off-thread EDT updates like this are not safe I think.

PR #18515, opened 2024-11-11
Ground Items plugin: hide stacks of coins under amount

SamuelDev:
I wanted a setting in the ground items plugin to hide stacks of coins under a given amount, but not all items under that amount.

I have added a new config for coin stack size and added an additional check to the hide logic for coins specifically.

I tested these changes in game after building the client.

I put the settings at `5` coins
image

And it can be seen that the coin stack of 4 is hidden, but the stack of 5 is not
image


JZomDev:
You can already do coins<5 in the `hide itrems` section.

image

raiyni:
This already exists in the config using `coins<NUMBER` on items.

PR #18519, opened 2024-11-12
Refactor CipherClue file

tygyh:


Nightfirecat:
This class was refactored to use a builder pattern, so this isn't needed

PR #18529, opened 2024-11-15
Reaper plugin

RdymentForks:


tylerwgrass:
New plugin development should follow the steps listed [here](https://github.com/runelite/plugin-hub).
Also this is not likely to be approved as-is as it doesn't look like it does anything useful.

RdymentForks:
> New plugin development should follow the steps listed [here](https://github.com/runelite/plugin-hub). Also this is not likely to be approved as-is as it doesn't look like it does anything useful.

sorry, I am trying to make a plugin for a class project, and having some problems with git, I am apparently missing some steps in my process. again, I apologize for my mistake.

PR #18533, opened 2024-11-16
itemstats: update total heal for hunter meats

### LINKED ISSUE #18505: [Item Stats] Hunter Food Color Incorrect
### Type

Incorrect behavior

### Operating System

None

### OS Version

_No response_

### Bug description

Normally, item stats shows tooltip text for healing/potions as yellow when you will waste part of the heal/recovery (shark has yellow +20 if you have 90/99 hp, as you will only heal +9). This handled incorrectly for hunter food.

If I want to heal with cooked pyre fox (+11/+8)  with 99 hp, once I go below 88 hp, both values will show green. However, I believe the correct behavior should be to only show both values as green once you go below 80 hp, as otherwise the full value of the heal will not be realized.

Examples of current behavior:

- 99 hp: both white
- 95 hp: both yellow
- 90 hp: 11 yellow, 8 green (this should still show both as yellow)
- 85 hp: 11 yellow, 8 yellow (**this** range is the primary issue.**  It should show either 11 green, 8 yellow, or both yellow)
-  75 hp: both yellow (this is correct)

### Screenshots or videos

Screenshot of 86/99 hp incorrectly showing both values as green:
image

### RuneLite version

Runelite v1.10.43
Launcher v2.6.3

### Logs

_No response_

tylerwgrass:
fixes #18505 

Also noticed `COOKED_WILD_KEBBIT`, `COOKED_LARUPIA`, and `COOKED_BARBTAILED_KEBBIT` were missing
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -126,12 +126,15 @@ public class ItemStatChanges
 		add(range(food(5), food(8)), LEAN_SNAIL_MEAT);
 		add(range(food(7), food(9)), FAT_SNAIL_MEAT);
 		add(range(food(7), food(10)), SPIDER_ON_STICK_6297, SPIDER_ON_SHAFT_6299);
-		add(combo(food(8), food(6)), COOKED_GRAAHK);
-		add(combo(food(9), food(8)), COOKED_KYATT);
-		add(combo(food(11), food(8)), COOKED_PYRE_FOX);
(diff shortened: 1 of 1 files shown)


Felanbird:
I do not like this approach, while it does make it so the numbers respect our existing color scheme (green full value, yellow wasted value) it makes the food appear as if it heals for _way_ more than it does, which is objectively incorrect.
image
moonlight antelopes just don't heal for 40

We definitely need the plugin to visually display
```
+14 (77) // display as +14 and calc as +14
+12 (89) // display as +12 but actually calc as +26
```

capSAR273:
Any updates on this PR?

cdfisher:
> Any updates on this PR?

There was feedback that the author hasn't addressed so it's unlikely to get merged unless that happens

PR #18537, opened 2024-11-17
chat commands: add chests

Breina:
Documentation PR: https://github.com/runelite/wiki/pull/131

* Larran's small chest
* Larran's big chest
* Brimstone chest
* crystal chest

---
Also removed some duplicate code in unit tests, and added a missing double backslash to the pattern for bird eggs.


Adam-:
Ideally this would be structured like: 1commit to deduplicate code in the tests, 1 to add the chests, 1 to escape the ..

PR #18546, opened 2024-11-19
chat commands: change !cmb to check local levels instead of hiscores

### LINKED ISSUE #18525: !cmb showing wrong details
### Type

Crash or hang

### Operating System

None

### OS Version

Windows 11 Latest

### Bug description

Wrong Details in !cmb

### Screenshots or videos

https://i.imgur.com/IeHHwVL.jpeg

### RuneLite version

Latest

### Logs

_No response_

lithream:
Issue: !cmb command shows wrong values for combat skills that are relatively low (e.g. 45 def shows as 1) and computes the wrong combat level.

Cause:  net.runelite.client.plugins.chatcommands.ChatCommandsPlugin.combatLevelLookup queries the hi-scores for combat skill levels, which doesn't work when the combat skills are too low to show up on the hi-scores.

Fix: Get real skill levels from the local client instead of the hi-scores and use them to compute the combat level.

Fixes #18525 

Example with low cb (levelled str up to 2 to make sure):
image

Example with high cb:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -1687,13 +1687,13 @@ public class ChatCommandsPlugin extends Plugin
 			return;
 		}
 
-		int attack = levels[net.runelite.api.Skill.valueOf("ATTACK").ordinal()];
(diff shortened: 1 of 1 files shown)


raiyni:
This will not work for other players

lithream:
@raiyni Sorry, you're right. Do you think it makes sense to have the function check locally when the command is issued by the player and query the hiscores when issued by other players?

PR #18554, opened 2024-11-22
overlays: add separate background colors for infoboxes and tooltips

sariyamelody:
as we already have separate font sizes for them, further allow users to customize infoboxes and tooltips (i.e. users may want transparent infoboxes/overlays but backgrounds for tooltips)

image


sariyamelody:
Should have fixed the merge issue with main! Let me know if there's any issues.

PR #18559, opened 2024-11-25
skillcalculator: Add multiplier input

tylerwgrass:
Adds a multiplier input to the skill calculator, bound to [1,100].

No Multiplier:
image

With Multiplier:
image

With Multiplier + Bonuses:
image

"Experience Multiplier" gets cut off in the label, so couldn't be consistent there with the label name

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/UnitFormatter.java
@@ -25,10 +25,7 @@
 package net.runelite.client.plugins.config;
 
 import java.text.ParseException;
-import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


Nightfirecat:
Per feedback from Discord, let's cap this at 32x maximum, since multipliers larger than that do not exist.

Also, we should add a `x` suffix, eg. `2x` (reference config units for how this can be achieved), and add a spinner so it can be increased or decreased without typing.

tylerwgrass:
I had to make the formatter factory public to use here, and didn't try too hard to refactor.
LMK if that's an issue

Nightfirecat:
Added in b0c7cb6401aba08078d35078cd1940904734a454 and e249333917767ab3d114a2d197aede752c095699

PR #18565, opened 2024-11-27
Add Stardust to Skill Calculator

Kolin63:


Kolin63:
* Add Stardust to Skill Calculator
* Listed at bottom because although the minimum mining requirement is 10, it can be 90

Kolin63:
Stardust is now listed after Limestone so that the level requirements are in order.

PR #18567, opened 2024-11-27
bank tags: support vials in potion storage

Zoinkwiz:
This PR is intended to make it so that empty vials can also be taken out from potion storage when in a bank tag. As empty vials seem to be unique in how they work in the potion storage, I've added unique methods for handling it.

I've made an assumption that the position for the vials widget in the potion storage children will always be the sum of all potions * 5 + 4, to try and make this more future-proofed. If there's a better way to determine where it'll be placed that'd be better, otherwise the main time this assumption will be challenged is when a new potion gets added to the game.

To allow for vials to be removed without having already opened the potion storage interface, I've added 5 new widgets to prepareWidgets. I've assumed I need to make the non-relevant widgets (line 292-295) be the correct WidgetType, but if they don't need to be then that can be simplified to a loop.


Zoinkwiz:
Thanks! Rebased it.

PR #18570, opened 2024-11-28
Update woodcutting leprechaun event to use new notification system

testing-ongithub:
I wanted to customize this event's notification, but I couldn't without the change in this PR. 😄


Adam-:
That is odd, it must have got missed when I converted them all earlier.

PR #18590, opened 2024-12-06
v2: Users can now toggle the 6-hour (5 hour 40 minute) logout notification on/off as needed

Hosea-MP:
Fixes issue #17890


Adam-:
Your PR doesn't work if the notification setting is anything other than ON. I've added it instead in e4d42a625cd8fbcefe147cd9e19478e793abc8a3.

PR #18594, opened 2024-12-08
interface-styles: Add Mobile interface style

Cyborger1:
Adds Mobile styled buttons, tweaked from the game's sprites cache to fit the expected widget sizes and shapes as much as possible.

If you feel this is not fit for Core, that's fine, I may repurpose this for Resource-Packs instead.

I wanted to submit this here first because I feel the changes that are being made are more in line with what interface styles does.

image
image
image
image


Nightfirecat:
I think this would be a decent fit here. Which sprite(s) needed to be modified from the cache to fit the desktop client?

Cyborger1:
Off the top of my head, I don't remember which exactly, and I don't think I have any intermediate files anymore, but I'll try my best.

For the buttons (1130, 1180 and 1181), I used 5767 and 5769, but had to resize them (40x40 down to 33x36).

For the little minimap orb (1438), I took 5794 and sized it down from 34x34 to 30x30.

For the hp/prayer/etc orbs, (1071 and 1072), I took 5791 and 5792, then I edited them best I could so that the label part on the left was lower down to match the original design.

For the minimap frame (1177), I took 5832 directly, but just had to pad an extra transparent pixel to the left and top of it, so the resolution goes from 181x165 to 182x166. **_Although, I don't remember if that padding is necessary?_** I don't see the other interface styles doing this, so I'm a little confused.

The four corner buttons (1026 to 1029), those I had to manually create from the design patterns of the regular square buttons (5767 and 5769).

The various bottom/top frames for the inventory stones (1032, 1036, 1173, 1174) are combinations of the above.

The remaining chat buttons (3051-3055, 3057 and 3058) are modified from 5779, 5780 and 5781.

3051 to 3055 are just those, but resized from 58x24 to 56x22, while 3057 and 3058 are resized to 79x22.

I got 3053 and 3054 from using the color palettes from 5770 and 5771.

image

Cyborger1:
I now remember why the padding is necessary, the actual widget size is 182 by 166, so giving it a 181x165 does this, duplicating some pixels. That's why I added the padding.
image

Interestingly, I noticed the 2010 interface style avoids this duplicated pixel issue by just snipping off some pixels, which does look slightly off. For a different PR, it may be worth fixing that.
image

PR #18609, opened 2024-12-12
Timers and buffs wealthy citizens pickpocketing

suoarski:
Updated the code for TimersAndBuffs to include wealthy citizens pickpocketing. The timer is triggered when a wealthy citizen begins to get distracted, and ends when it is possible for them to get distracted again. This allows the user to have a small section (top left) of the game window visible whilst doing other things on their PC, and still know when a wealthy citizen is about to or is being currently distracted.

This is my first time developing for Runelite, and the first time in 10 years to be using Java. Any feedback is welcomed.


suoarski:
Here's how the new feature can be used to help AFK:

efwgfwfwefwfweew

Perhaps we could use a better icon for this? I was thinking a picture of a wealthy citizens face with an Urchin (Leo) face calling, and a red background. If I were to do this, I'd probably use generative AI for the icon. Eventually I was also thinking of making a few more timers for the stealing valuables minigame. 

Just wanna make sure these ideas get approved before I spend too much time working on them.

Issue #18573, opened 2024-11-29
Can't change left click for prayer potions


synphony5159:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

*Edited to be more clear on what I was experiencing*
I cannot change the left click option to "drink" for prayer potions. When I turn on "Customizable left-click" it automatically sets my prayer potions to "use", and I cannot change "use" to "drink". This only happens for prayer potions

### Screenshots or videos

![Screenshot 2024-11-29 170633](https://github.com/user-attachments/assets/53b23634-d58c-475c-8eb5-11f17896e02a)

### RuneLite version

RL version 1.10.45 launcher 2.6.13

tylerwgrass:
Did you try clicking "Reset"

synphony5159:
> Did you try clicking "Reset"

That did work for me, thank you! I'm not sure if what I experienced was a bug or not, but reset does fix it

Nightfirecat:
You had changed the left-click option to 'Use', so getting the default ('Drink') is only possible by resetting the left-click option. This is expected behavior.

PR #18619, opened 2024-12-17
MenuEntrySwapper: improve "Reset" option

0hterm:
Improved the "Reset" option to include the name of the default command in parentheses.

Also addresses issue #18573 

image


Nightfirecat:
I'm not at all opposed to this, in fact, I think the idea is very good (I briefly explored doing the same thing in this commit last year: https://github.com/Nightfirecat/runelite/commit/db4d372557428a24525dc914477b0fcd28929585), but this should apply to all menu entry swaps, not only items, and probably should use a different color to distinguish the action (Reset) from the default action (Drink, in your example).

See below for how my code looks when run:
image

Adam-:
Why is this setting swaps for items which have no swaps? This shouldn't be modifying any existing config.

Also the "Reset" option intentionally unsets it to it takes the Jagex default. This allows Jagex to change the default left or shift click on items at a later point and have it work.

PR #18621, opened 2024-12-20
Update clue text to match the correct cave.

martyj:
Referencing: https://oldschool.runescape.wiki/w/Gu%27Tanoth

The only way to get onto the island is by taking a specific cave entrance (the broken bridge leading to a small island is not passable). The relevant cave can be found in the Skavid Caves, south-west of Gu'Tanoth.


TheLope:
You changed it to "south-west of Feldip Hills (AKS)", but as you described it, the cave is south-west of Gu'Tanoth

Adam-:
The cave is indeed north-west of feldip hills

image

martyj:
I stand corrected, thank you.

PR #18623, opened 2024-12-21
npc indicators menu color

mathbum:
enable finer control for npc indicator recoloring tagged npc menu entries. 
now you can choose name, level or both instead of just on vs off.

https://github.com/runelite/runelite/discussions/18622

example set to "Name":
image


runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -315,18 +315,73 @@ public void onMenuEntryAdded(MenuEntryAdded event)
...
+		String cleanedTarget = Text.removeTags(target);
+		if (style == NpcIndicatorsConfig.MenuHighlightStyle.BOTH)
+		{
+			return ColorUtil.prependColorTag(cleanedTarget, color);
+		}
+
+		String name, level;
+		if (target.contains("  (level-"))
+		{
+			int levelStartIndex = cleanedTarget.lastIndexOf('(');
```
mathbum, line 351: this is necessary in the case that there is a level but there is no color tag on it. otherwise it will override the original 'name' color tag

Adam-: It isn't possible to have a level with no color tag

mathbum: oh well that simplifies it :) thanks

mathbum: though this PR does make it possible (if 'both' is selected)

mathbum: Just to be clear monster hp plugin will add an extra color tag to the level. This means if it's used with Npcindicators menu highlight, it will override the "remaining health" color part of the level. 

Arguably this is a bug, but I'm fine either way because it would work still, and at least the color is something chosen by the user.

Adam-:
added in 5346024bd8f06d7d5d170f29837f3f0906703e1a

PR #18635, opened 2024-12-28
fix npcidicators interaction with 'monstermenuhp' bug

mathbum:
The current code makes assumptions about the structure of the 'target'. this means it is brittle and might not worth with other plugins, like some monstermenuhp settings. 

For example, if it is set to 'level' and npcindicators is set to 'name' the user reasonably expects that these two plugins could both work on different parts of the menu. But it doesn't, this code will make it so there is inconsistent behavior based upon the current hp % of the npc.

here is an example monstermenuhp target:
<col=ffff00>Ram<col=ff00>  (lev<col=00ff00>el-2)

note:
1) when monstermenuhp is set to name, and npcindicator is level, current code should work as expected.
2) if monstermenuhp is set to both then the hp colors will inconsistently partially work, this isn't unreasonable to the user since intended behavior based upon user settings is unclear.
3) if npcindicators is set to both, then it will reasonably override monstermenuhp.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
@@ -360,11 +360,15 @@ public class NpcIndicatorsPlugin extends Plugin
 		//<col=ffff00>Tool Leprechaun
 		//<col=ffff00>Ram<col=ff00>  (level-2)
 		String name, level;
-		if (target.contains("  (level-"))
+		int levelStartIndex = target.indexOf("(");
+		if (levelStartIndex > -1)
 		{
-			int c = target.lastIndexOf('<');
-			name = target.substring(0, c);
-			level = target.substring(c);
+			// in case other plugins (like monster menu hp) add tags in middle of level or name, or more than 2 tags total
+			int monsterEndIndex = target.substring(0, levelStartIndex).lastIndexOf('<');
+			monsterEndIndex = monsterEndIndex > -1 ? monsterEndIndex : levelStartIndex;
+
+			name = target.substring(0, monsterEndIndex);
+			level = target.substring(monsterEndIndex);
 		}
 		else
 		{
--- a/runelite-client/src/test/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPluginTest.java
@@ -169,14 +169,14 @@ public class NpcIndicatorsPluginTest
 		npcIndicatorsPlugin.onNpcSpawned(new NpcSpawned(npc));
 
 		TestMenuEntry entry = new TestMenuEntry();
-		entry.setTarget("<col=ffff00>Ram<col=ff00>  (level-2)");
+		entry.setTarget("<col=ffff00>Ram<col=ff00>  (lev<col=00ff00>el-2)");
 		entry.setIdentifier(MenuAction.NPC_FIRST_OPTION.getId());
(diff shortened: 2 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java
```diff
@@ -360,11 +360,15 @@ private static String recolorMenuTarget(String target, NpcIndicatorsConfig.MenuH
...
-		if (target.contains("  (level-"))
+		int levelStartIndex = target.indexOf("(");
+		if (levelStartIndex > -1)
 		{
-			int c = target.lastIndexOf('<');
-			name = target.substring(0, c);
-			level = target.substring(c);
+			// in case other plugins (like monster menu hp) add tags in middle of level or name, or more than 2 tags total
+			int monsterEndIndex = target.substring(0, levelStartIndex).lastIndexOf('<');
+			monsterEndIndex = monsterEndIndex > -1 ? monsterEndIndex : levelStartIndex;
```
mathbum, line 368: this line is exclusively for robustness, i have know of no cases where it is possible for it to be -1, but it seemed reasonable to put a check in here just in case. i can remove if preferred.

mathbum:
@Adam- 
I came up with a pretty simple change to make it more robust and play better with monstermenuhp.

Adam-:
If this is a real problem, it means that monster menu hp runs before npc indicators. However, if it did, it would mean that the release wouldn't work because it calls removeTags() on the entire menu target. But we also know that the release works with monster menu hp. How is this possible?

mathbum:
> If this is a real problem, it means that monster menu hp runs before npc indicators. However, if it did, it would mean that the release wouldn't work because it calls removeTags() on the entire menu target. But we also know that the release works with monster menu hp. How is this possible?

yes, i believe monster hp runs first. where does it call 'removeTags' on the whole target?  "both" does that but "name" and "level" only call "removeTags" on those specific parts, otherwise this wouldnt even work with normal npc menu coloring (i.e. low lvl npcs in green text)

Adam-:
> > If this is a real problem, it means that monster menu hp runs before npc indicators. However, if it did, it would mean that the release wouldn't work because it calls removeTags() on the entire menu target. But we also know that the release works with monster menu hp. How is this possible?
> 
> yes, i believe monster hp runs first. where does it call 'removeTags' on the whole target? "both" does that but "name" and "level" only call "removeTags" on those specific parts, otherwise this wouldnt even work with normal npc menu coloring (i.e. low lvl npcs in green text)

https://github.com/runelite/runelite/blob/runelite-parent-1.10.46.1/runelite-client/src/main/java/net/runelite/client/plugins/npchighlight/NpcIndicatorsPlugin.java#L327

Adam-:
Both npc indicators and MMH have the same priority level (0) on the subscribers, so the order they run in isn't defined (it depends on I think the plugin start order). In practice I think this means that the core would run first since it might load first but I'm not sure. It would probably be best to lower the priority of MMH so that it runs after the core recolors it.

I am hesitant to take your approach to level parsing because there are a number of legitimate npcs with `(` in their names: https://abextm.github.io/cache2/#/viewer/npc/%5C(

mathbum:
> Both npc indicators and MMH have the same priority level (0) on the subscribers, so the order they run in isn't defined (it depends on I think the plugin start order). In practice I think this means that the core would run first since it might load first but I'm not sure. It would probably be best to lower the priority of MMH so that it runs after the core recolors it.
> 
> I am hesitant to take your approach to level parsing because there are a number of legitimate npcs with `(` in their names: https://abextm.github.io/cache2/#/viewer/npc/%5C(

ah okay. ill look into the priority change and didn't know about npcs with "(" MMH uses "(" as a delim so i figured it was safe

PR #18641, opened 2025-01-01
Config panel `@Units` fixes

LlemonDuck:
there are a couple bugs going on here:
* for integer fields annotated with both `@Range` and `@Units`, the UnitFormatter does not validate that the value is within the allowed range and so allows a write to the config with an out of bounds value. If the config panel is refreshed, this is obfuscated from the user by a `constrainToRange` call in `ConfigPanel#createIntSpinner`
* for double fields annotated with `@Units`, only integer values can be saved since the present of a decimal point will throw an exception in `UnitFormatterFactory#stringToValue`.

To fix these, we can add a min and max to the UnitFormatter, and create a corresponding class for double values.


Adam-:
fixed in 582782d250cbe8e45a0c7e5f281e5bdcdf25d664

PR #18652, opened 2025-01-07
droppartychest: add initial version as a core plugin

pgroenbaek:
This is a very simple plugin that adds total chest value to the title of the drop party chest interfaces in the party room and clan hall, like this:

DropPartyChest

It is quite similar to what the bank plugin does to the bank interface title. Just without HA value.

I originally submitted this is a [PR in the plugin-hub repo](https://github.com/runelite/plugin-hub/pull/7180), but YvesW suggested it would fit better as a core plugin. So here it is.

My reasoning for making this is that there is no easy way to see this otherwise in the standard UI.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/droppartychest/DropPartyChestConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/droppartychest/DropPartyChestConfig.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2025, Peter Grønbæk Andersen <peter@grnbk.io>
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


Adam-:
This is probably better as a pluginhub plugin. They probably meant adding it to the existing bank plugin, although I am not sure this fits there since this isn't really storage (eg like the seed vault).

pgroenbaek:
I'm okay with any of the options in terms of where to add it. Whatever the consensus is. I agree with you about it perhaps not really fitting into the bank plugin.

I just think this sort of functionality should be available in the chest UI.

pgroenbaek:
Closing this as it was added to the plugin hub.

PR #18661, opened 2025-01-11
runenergy: Update Run Energy Formulas

coopermor:
8 January 2025 Run Energy Changes
Energy Regeneration was changed from (Agility / 6) + 8 to (Agility / 10) + 15
Energy Drain was changed to include the agility level. (60 + (67 * Weight / 64)) * (1 - Agility/300)

The graceful boost check was looking for the inventory max-cape id rather than the worn max cape which is unrelated to the game update.


coopermor:
At 0 weight and 99 agility with stamina
Energy lost per tick Floor(60+(67*0/64))*(1-99/300)) = 40
Stamina effect reduces to 30% = 12
10,000 energy / 12 energy per ticks * 0.6 ticks/sec = 500 seconds

At 0 weight and 99 agility with ring of endurance
Energy lost per tick Floor(60+(67*0/64))*(1-99/300)) = 40
Ring of endurance reduces to 85% = 34
10,000 energy / 34 = 295 ticks
295 ticks * 0.6 sec/tick = 177 seconds = 2 minutes 57 seconds

PR #18665, opened 2025-01-13
Skill Calculator: Add ability to refresh calculator without switching tabs

Unknowncall:
Instead of having to switch tabs to update the current XP, you can simply reclick the same tab to refresh the tab.


Unknowncall:
Made it so only the XP updates and not the target level.

Adam-:
Your changes seems like forceReload=true always, since I would expect the currentCalculator != calculatorType check to be false when refreshing the calculator. That seems wrong and maybe points to 08128f6d400b2a9290f2b17d515994c4e316618b being wrong.

Unknowncall:
> Your changes seems like forceReload=true always, since I would expect the currentCalculator != calculatorType check to be false when refreshing the calculator. That seems wrong and maybe points to [08128f6](https://github.com/runelite/runelite/commit/08128f6d400b2a9290f2b17d515994c4e316618b) being wrong.

It does seem like forceReload is always true. During my testing I did not have to change forceReload. I looked through the code and the only place modifying the forceReload to true is in the reloadCurrentCalculator which is only called when the world is changed.

Adam-:
> > Your changes seems like forceReload=true always, since I would expect the currentCalculator != calculatorType check to be false when refreshing the calculator. That seems wrong and maybe points to [08128f6](https://github.com/runelite/runelite/commit/08128f6d400b2a9290f2b17d515994c4e316618b) being wrong.
> 
> It does seem like forceReload is always true. During my testing I did not have to change forceReload. I looked through the code and the only place modifying the forceReload to true is in the reloadCurrentCalculator which is only called when the world is changed.

It looks very broken currently, the reloadCurrentCalculator() ends up not triggering the setOnSelectEvent() callback. It's possible this used to work and was broken with the flatlaf update maybe. This should be fixed instead of just adding additional code on top of it like what you have.

Unknowncall:
> > > Your changes seems like forceReload=true always, since I would expect the currentCalculator != calculatorType check to be false when refreshing the calculator. That seems wrong and maybe points to [08128f6](https://github.com/runelite/runelite/commit/08128f6d400b2a9290f2b17d515994c4e316618b) being wrong.
> > 
> > 
> > It does seem like forceReload is always true. During my testing I did not have to change forceReload. I looked through the code and the only place modifying the forceReload to true is in the reloadCurrentCalculator which is only called when the world is changed.
> 
> It looks very broken currently, the reloadCurrentCalculator() ends up not triggering the setOnSelectEvent() callback. It's possible this used to work and was broken with the flatlaf update maybe. This should be fixed instead of just adding additional code on top of it like what you have.

I'll go through a resolve the issue.

Unknowncall:
I created a new branch, going to open a new PR when I have the code. I will use your code from above.

PR #18667, opened 2025-01-13
Fix custom notification sound playing twice the first time after client launch

### LINKED ISSUE #14914: Custom notification sounds play twice after being loaded
### Discussed in https://github.com/runelite/runelite/discussions/14910

<sup>Originally posted by **BuzzfeedBlue** May  1, 2022</sup>
When you use a custom wav file for alert notifications (in the main runelite plugin) the first time a sound is loaded, it will play twice, and until logout, it will then only play once. If you stay logged in, play the sound (twice), and then load a new file (without logging out), it will still play twice because it has to "load for the first time" the new sound

The default system alert doesnt have this issue</div>

tomfulghum:
Calling `clip.loop(1)` would cause the clip to be played twice the first time a notification sound should be played. This happens because after a loop finishes, the clips position is set to the last frame. When a new loop with count 1 (meaning repeat once) is started, the clip is instantly reset to the first frame, which counts as the first loop, causing the clip to audibly only be played once.

Setting the frame position to 0 and the loop count to 0 fixes this problem.

In my limited tests using `clip.start()` instead of `clip.loop(0)` worked just as well, but I don't know if the issue described in the comment above the changed lines is still relevant.

Closes #14914

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/Notifier.java
+++ b/runelite-client/src/main/java/net/runelite/client/Notifier.java
@@ -525,7 +525,7 @@ public class Notifier
 		// Using loop instead of start + setFramePosition prevents the clip
 		// from not being played sometimes, presumably a race condition in the
 		// underlying line driver
-		clip.loop(1);
+		clip.loop(0);
 	}
 
 	private boolean tryLoadNotification()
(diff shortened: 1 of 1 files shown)


Adam-:
@abextm can you check this? it seems you were the one originally with this issue.

PR #18674, opened 2025-01-15
ItemMapping: add leagues V kits and tome of earth

sil3xz:
Adds:

- Echo venator bow
- Echo virtus
- Tome of earth


Nightfirecat:
This PR does not also map the kitted items to the unkitted items—so an item with a kit on would have been considered to only include the price of the kit, and not the base item. I've written the needed changes in fc90582d86c826066a1efe30591a1ab5c1d1db17

PR #18684, opened 2025-01-21
Woodcutting Plugin: Leprechaun Forestry Event Improvements 

ACarnesi:
Added an Indicator for easily identifying End of Rainbow during forestry events at a glance
Added an optional counter for Leprechauns Luck charges

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -970,6 +970,8 @@ public final class Varbits
 	public static final int COMBAT_TASK_MASTER = 12889;
 	public static final int COMBAT_TASK_GRANDMASTER = 12890;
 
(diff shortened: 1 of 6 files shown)


Adam-:
I've made a number of changes to this:

- Don't add luck infobox on login - most players aren't woodcutting, so having an on-by-default infobox added on login if they happen to have luck is a bad idea. I've made it only add the infobox if there is a session active.
- Similarly I've made it remove the infobox when the woodcutting session expires, otherwise it never gets deleted.
- Remove the infobox on plugin shutdown and also when turning the Show leprechaun's luck option off..
- Changed the infobox to just a normal counter with a tooltip instead of using a subclass purely for tooltip generation

ACarnesi:
That all makes sense. I figured there'd be some changes since I haven't ever really worked in this space before. Thanks taking the time to make the improvements!

PR #18689, opened 2025-01-23
Adding Castle Wars interface ID information and a couple overlays

HelixSpiral:
This PR adds in some basic functionality for Castle Wars that can be extended in the future.

- Most Interface ID information
- A couple overlays for the "main" interfaces


runelite-api/src/main/interfaces/interfaces.toml
Nightfirecat, line 1: Please remove the definitions which are not being used in core; these are subject to bit rot as they would not be verified after game updates.

HelixSpiral: Done

runelite-api/src/main/interfaces/interfaces.toml
```diff
@@ -129,6 +129,24 @@ coffer=2
 id=598
 container=2
 
+[castlewars_lobby]
+id=131
+time_remaining=2
+
+[castlewars_saradomin_game]
```
Nightfirecat, line 136: The "game" suffix is redundant, particularly since "game_timer" is one of the children.
```suggestion
[castlewars_saradomin]
```

runelite-api/src/main/interfaces/interfaces.toml
```diff
@@ -129,6 +129,24 @@ coffer=2
...
+id=58
+castle_and_rock_group_status=3
+game_timer=16
+score_box=18
+leave_waiting_room=24
+
+[castlewars_zamorak_game]
```
Nightfirecat, line 143:
```suggestion
[castlewars_zamorak]
```

runelite-api/src/main/interfaces/interfaces.toml
HelixSpiral: Fixed

runelite-api/src/main/interfaces/interfaces.toml
HelixSpiral: Fixed

PR #18698, opened 2025-01-29
Crowdsourcing: raw jubbly cooking success rates

samszotkowski:
Crowdsource plugin was only tracking spam messages, but jubbly is not filtered.

image


samszotkowski:
Just realized same issue is in the cooking plugin too.

samszotkowski:
Nevermind, someone asked Mod Ash instead so we don't need to crowdsource it.

PR #18720, opened 2025-02-06
sidebar: add plugin reordering

tylerwgrass:
- move by right clicking icon and dragging
  - plugin location updated on mouse release to reduce flickering
- config plugin effectively pinned at the top in pos 0
  - may be a bug if there exists a plugin that would be sorted above it in the existing comparator (is that allowed?)
- plugins are keyed by the base64 encoding of the last two segments of their class path (i.e. `Base64("grandexchange.GrandExchangePanel"`)
- saved to config as list of these keys at the current priority
- new plugins added to bottom
  - disabled plugins or those that fail to load will be added back to bottom when reenabled


huntercavazos:
Would love to see this implemented

Issue #18722, opened 2025-02-06
Huasca Herb Missing from Item Identification plugin


verafeiyd:
### Type

Incorrect behavior

### Operating System

None

### OS Version

Windows 10

### Bug description

nothing really else to say

### Screenshots or videos

![Image](https://github.com/user-attachments/assets/4ef6c2db-913b-424f-b585-e1bbfcd431eb)

### RuneLite version

newest launcher/runelite version, updated today with 2/5/25 update

170iqalbi:
Hey can I be assigned this?

Felanbird:
> Hey can I be assigned this?

There's already a PR making this change, it's just that the title for the newest version of the PR doesn't say it does that, but it does.

Nightfirecat:
Added in #18724

PR #18723, opened 2025-02-06
item identification: add huasca herb and seed

DeBoogie:
Add huasca herb and seed to the item identification plugin with "HU" abbreviation. #18722


Nightfirecat:
Every other short name is either 1 character long or 3. This should use "HUA" (and possibly rename Harralander to "HAR") instead of the 2-character "HU" abbreviation.

DeBoogie:
> Every other short name is either 1 character long or 3. This should use "HUA" (and possibly rename Harralander to "HAR") instead of the 2-character "HU" abbreviation.

I tried to consider for a moment whether to use HU or HUA, because there were other (non-Herb) items in the code with shorter abbreviations, but it turned out that they were items with a two words, which is why their abbreviation is two characters long.

> Also, this should label unfinished huasca potions in the unfinished potions section

The suggestion regarding Unfinished Potion has been noted.

DeBoogie:
I will create according branch for each fix.

PR #18724, opened 2025-02-06
item identification: update herb abbreviations and add huasca potion

DeBoogie:
Related to #18722


Nightfirecat:
This looks better. The unfinished harralander potion entry will also need the "H" -> "HAR" treatment. Then, I think this is good to go.

Issue #18654, opened 2025-01-08
[Item Prices] Crystal felling axe has incorrect price


fuzxi:
### Type

Incorrect behavior

### Operating System

None

### Bug description

The Item Prices plugin shows the crystal felling axe as being worth 1310gp on the GE. It's untradeable but can be reverted to a dragon felling axe and a crystal tool seed, which combined are currently worth nearly 29m. The crystal pickaxe does not have the same issue and shows the expected price.

### Screenshots or videos

![2025-01-08-07-13-58](https://github.com/user-attachments/assets/19f8d0a6-fcd2-41a1-b275-4442f44f2ddf)
![2025-01-08-07-14-12](https://github.com/user-attachments/assets/0fdb1d7e-95f0-4638-a3a6-4f74fa3aa757)

### RuneLite version

RuneLite version: 1.10.46.1
Launcher version: 2.6.10

Nightfirecat:
Fixed in 8c61b6b51b0b5deed273a290f746e6d74e3ee103

PR #18725, opened 2025-02-06
item prices: Only show GE price for tradeable items

DeBoogie:
Previously the GE price would be shown for untradeable items. This change ensures GE prices are only shown when the item is actually tradeable. Contributes #18654


Felanbird:
This is the wrong approach, we intend to show the GE value for untradables when they can be converted to a tradable item.

The linked issue wants us to expand our item mapping to include crystal felling axe(s)

Nightfirecat:
Fixed in 8c61b6b51b0b5deed273a290f746e6d74e3ee103

PR #18731, opened 2025-02-08
Fixed a Unit Test

Alluneve:
String.format() will initialize a default format with a Locale, the default locale depending on the setup of the test's host machine, AKA your Windows environment.

this means that if you build this in many countries, you'll get a locale like eg: fr_BE.
and fr_be on String.format("%05.2f", seconds) will return

org.junit.ComparisonFailure: 
Expected :0:03.60
Actual   :0:03,60

this fix made `mvn clean install` actually compile with running tests, running and working tests is nice.

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -386,4 +386,6 @@ public final class AnimationID
 	public static final int HUEYCOATL_DEATH = 11679;
 
 	public static final int SLAYER_TOWER_WATER_BUCKET_DRIP = 3558;
+
+	public static final int KEY_MASTER_IDLE = 4519;
 }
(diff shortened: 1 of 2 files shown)


Nightfirecat:
How are you running this test? We already instruct surefire to use en_US locale here 481b3a5d938fbffaf9aa729fe460939080e6aec0, which should already be picked up when tests are run

Alluneve:
Mvn clean install & a direct run of the specific offending unit test with my version of intelij ultimate.

I was mostly grumbling that the build failed before I ended up reading the documentation that you direct people to build skipping tests.

Nightfirecat:
I think this may have been closed in error because of a force-push to your master branch? Last I recall from discussion in discord we hadn't figured out why this is an issue for you

Alluneve:
Oh sorry, yeah, force updated master to runelite local.

It's still a build problem with locale though

PR #18736, opened 2025-02-08
Added the help command to print out all registered commands

Alluneve:
!help will print out in chat as a game message, all registered !commands.


runelite-client/src/main/java/net/runelite/client/chat/ChatCommandManager.java
```diff
@@ -174,4 +176,8 @@ private static String extractCommand(String message)
 
 		return message.substring(0, idx);
 	}
+
+	public List<String> getCommands(){
+		return new ArrayList<>(commands.keySet());
```
Nightfirecat, line 181: This looks like it would not have any kind of guarantee for what order the keys are returned. I would think it best for us to sort them, either here or in the chat commands plugin.

runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsConfig.java
```diff
@@ -231,4 +231,15 @@ default Keybind clearChatBox()
...
 		return Keybind.NOT_SET;
 	}
+
+	@ConfigItem(
+			position = 22,
+			keyName = "help",
+			name= "help command",
+			description = "prints out a list of commands available"
+	)
+	default boolean help()
```
Nightfirecat, line 241: This is unused

Nightfirecat (changes requested):
Another more serious issue is that this would display commands which may not be usable because they were disabled in config. (disabling commands does not unregister their commands, it only forces an early return in the command handler)

PR #18738, opened 2025-02-09
Loot tracker: Record last menu option as metadata for Royal Titans

samszotkowski:
Distinguish between "Loot" and "Take pages" option in metadata to figure out desiccated pages drop rates. Both get lumped together currently.

image


Nightfirecat:
Added in 18ce6717feb80bfef2c958b4cb83ed6109ebdd12

PR #18745, opened 2025-02-11
client: add @injectable AudioPlayer

riktenx:
Adds new `@Inject`able `AudioPlayer` w/ APIs to play sound from various inputs.

The intent here is to provide a dedicated sound playback channel to prevent caller/plugin misuse of audio streams or lines.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/audio/AudioPlayer.java
+++ b/runelite-client/src/main/java/net/runelite/client/audio/AudioPlayer.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2025, Rikten X <riktenx@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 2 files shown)


riktenx:
I never experienced the lockup effect described in #11165, but in my own testing on Ubuntu 24.04 (in a VM) this seems to work fine:
* I can mash the new "test audio" button and don't experience a lockup
* If I _remove_ the line event close from AudioPlayer, I can mash the test audio button and it WILL lock up after some extended mashing (maybe 15-20 seconds or so at ~8 presses per second) - the torches on the login screen will animate, but nothing responds to mouse clicks whatsoever (including the window manager buttons)

runelite-client/src/main/java/net/runelite/client/audio/AudioPlayer.java
```diff
@@ -0,0 +1,158 @@
...
+			if (event.getType() != LineEvent.Type.STOP)
+			{
+				return;
+			}
+
+			synchronized (this) // We cut off the old line, and this one becomes the new "retainer".
+			{
+				if (prevLine != null)
+				{
+					prevLine.close();
```
Adam-, line 136: It seems like this could close the previous clip before it is done playing? Does just closing the clip here, after STOP event, not work correctly?

riktenx: Think we've mostly discussed this in #development, but just for posterity, having self-closing clips means the control slider dies with the clip itself on certain platforms/software (e.g. pavucontrol).

PR #18783, opened 2025-02-21
item identification: add moonlight potion

### LINKED ISSUE #18782: [Item Charges] Moonlight Potion doses not shown

Macweese:
This PR adds the Moonlight Potion to the item identification plugin. I wasn't sure about setting the abbreviations (e.g. if there's any established standard); specifically the shortform "M" being shared with Magic Potion. Though since Moonlight Potions are exlusively available within Neypotzli I figured the shared initialism to be tolerable.

Resolves #18782


Nightfirecat:
I agree, the shared abbreviation is probably fine. If we decide otherwise later on, we'd probably need to change magic potions to "Mag", since I don't know how moonlight potions could reasonably be shortened further while still making any sense. ("Mo" and "Moo" both seem unhelpful at best)

Nightfirecat:
It'll need to be in a separate PR one way or another, perhaps adding it to your other item charges PR would be best?

PR #18788, opened 2025-02-23
menuentryswapper: add reset option to birdhouse swap

0xMOPI:
Adds the new Reset option to the birdhouse swap

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
@@ -227,14 +227,14 @@ public interface MenuEntrySwapperConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "swapBirdhouseEmpty",
+		keyName = "swapBirdhouse",
 		name = "Birdhouse",
-		description = "Swap Interact with Empty for birdhouses on Fossil Island.",
+		description = "Swap Interact with Empty or Reset for birdhouses on Fossil Island.",
 		section = objectSection
 	)
-	default boolean swapBirdhouseEmpty()
+	default BirdhouseMode swapBirdhouse()
 	{
-		return true;
+		return BirdhouseMode.EMPTY;
 	}
 
 	@ConfigItem(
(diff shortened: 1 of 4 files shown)


Adam-:
Are you not able to use the normal shift right click -> swap setting on birdhouses? That is the preferred way to do this - this option only exists because it predates it.

0xMOPI:
You can but you need to do it for each type of wood birdhouse, which isn't necessarily a problem.

runelite-client/src/main/java/net/runelite/client/plugins/menuentryswapper/MenuEntrySwapperConfig.java
```diff
@@ -227,14 +227,14 @@ default boolean swapBank()
 	}
 
 	@ConfigItem(
-		keyName = "swapBirdhouseEmpty",
+		keyName = "swapBirdhouse",
```
Nightfirecat, line 230: Since the key name is being changed, we should migrate config values. (eg. `true` -> `EMPTY`, and `false` -> `Interact`)

fishermans-friend:
Any update on this? I was just about to go implement it myself.

JZomDev:
> Any update on this? I was just about to go implement it myself.

You can do this swap without this PR being added by right clicking the game object and adding the swap. 

https://github.com/runelite/runelite/wiki/Menu-Entry-Swapper#customizable-inventory-and-equipped-item-swaps

I think you'd have to do this per type of bird house when you go from yew -> magic -> redwood but it works today without this merged.

fishermans-friend:
> > Any update on this? I was just about to go implement it myself.
> 
> You can do this swap without this PR being added by right clicking the game object and adding the swap.
> 
> https://github.com/runelite/runelite/wiki/Menu-Entry-Swapper#customizable-inventory-and-equipped-item-swaps
> 
> I think you'd have to do this per type of bird house when you go from yew -> magic -> redwood but it works today without this merged.

I know you can, and I do, but you shouldn't have to. It makes no sense to have `Empty` be default when `Reset` does the same thing and more. If spaghetti is a concern, you don't need to have the option for both imo, you can just replace the current `Empty` option with `Reset`.

Adam-:
The preferred way to do this is to use the shift right click method to swap the menus. The existing setting is there just because it predates that.

Issue #18675, opened 2025-01-15
Prayer Enhance (+) duration is wrong


harbingerofme:
### Type

Incorrect behavior

### Operating System

Windows

### Bug description

Using a prayer enhance at non level 99 prayer is always wrong (and looking at the code and wiki it's wrong for 99 prayer too)
Wiki suggests a duration of 288 seconds `floor(99/2+31) activations * 6 ticks per activation = 480 ticks` which implies about 288 seconds.
However at 92 prayer, this becomes 277 seconds which makes the timer wrong for a solid 13 seconds.

### RuneLite version

RuneLite version: 1.10.47
Oldschool revision: Rev 288
Launcher version 2.6.7

Nightfirecat:
Related to #16169

Issue #16169, opened 2023-01-17
Track more timers using variables


Nightfirecat:
### Type

Other

### Operating System

None

### Bug description

Over time, [thanks to variables transmitted for the enhanced client's buff bar](https://github.com/Joshua-F/cs2-scripts/blob/master/scripts/%5Bproc,buff_bar_get_value%5D.cs2) we've been able to transition a number of our Timers plugin's timers to be updated via these vars, making them less error-prone and more dependable. In some cases, they become a lot more useful because their state can be persisted between sessions. (eg. logging in with some amount of anti-poison effect remaining)

There remain a handful of timers which could be created or updated based on those found in the linked script, namely:
- [ ] **Update:** [Use varbit 5417 (and 5419?)](https://github.com/Joshua-F/cs2-scripts/blob/1202119848818ef90e38d488f64504ecf491d845/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L28-L39) to update the prayer enhance potion timer
- [ ] **Update:** [Use varbit 4099](https://github.com/Joshua-F/cs2-scripts/blob/1202119848818ef90e38d488f64504ecf491d845/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L80-L83) to update the god wars dungeon altar timer
- [ ] 	**New (?):** [Using varbit 13137](https://github.com/Joshua-F/cs2-scripts/blob/1202119848818ef90e38d488f64504ecf491d845/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L200-L203), create a desert heat timer
- [ ] **New:** [Using varplayer 3347 and 3078](https://github.com/Joshua-F/cs2-scripts/blob/1202119848818ef90e38d488f64504ecf491d845/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L224-L227), create a Forinthry surge timer
- [ ] **Update:** [Use varbit 13672](https://github.com/Joshua-F/cs2-scripts/blob/1202119848818ef90e38d488f64504ecf491d845/scripts/%5Bproc%2Cbuff_bar_get_value%5D.cs2#L324-L331) to update the cannon decay and break timers

[...]

PR #18793, opened 2025-02-24
timers: track prayer enhance effect using varbits

Macweese:
This PR migrates the tracking of the Prayer Enhance effect to rely on vars.

Very limited testing was done in game to see if this works. I only had the opportunity to use accounts with 99 prayer, but from what I was able to observe it worked as expected.

I did not add a test for this because the duration varies greatly depending on the effect/potion tier and the player's prayer level. And as the only accounts I had the ability to test this with, all had 99 prayer, I can only for certain know what a case with 99 prayer should produce. I am also not sure of the formula stated on the OSRS Wiki, since it doesn't 100% align with the results in game, so I refrained from relying on that. This [model](https://www.desmos.com/calculator/yqcnj7fjsy) is an approximation of the effect for the Prayer Enhance (+) potion, the other versions have different formulas when determining the periods and period lengths.

<details>
  <summary>Extra</summary>
  
Additional varbit which I didn't add but may be relevant/of interest for future reference:
```java
/**
 * The tier of prayer enhance potion the player was last affected by.
 *  0 = None
 *  1 = Prayer enhance (-)
 *  2 = Prayer enhance
 *  3 = Prayer enhance (+)
 */
 public static final int COX_PRAYER_ENHANCE_TIER = 12397;
```
</details>

Ref #16169 #18675

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -897,6 +897,22 @@ public final class Varbits
 	 */
 	public static final int COX_OVERLOAD_REFRESHES_REMAINING = 5418;
 
+	/**
+	 * How many remaining prayer points the player will restore through the prayer enhance effect, in Chambers of Xeric.
+	 * <p>
+	 * This value is determined by the player's prayer level and will decrement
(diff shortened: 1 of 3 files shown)


Nightfirecat:
This needs rebase since gameval definitions exist now, though it looks good at a glance.

Issue #18801, opened 2025-02-27
The player's pet is hidden when "Hide attackers" is enabled


Jordan-Cousins:
### Type

Incorrect behavior

### Operating System

Linux

### OS Version

Ubuntu 22.04.5 LTS

### Bug description

When "Hide attackers" is enabled, the plugin also hides your pet. I tested with a few different pets to ensure it wasn't just a single ID issue.

### Screenshots or videos

https://github.com/user-attachments/assets/e319eb92-6d1e-424e-8f73-6f282517ca1c

### RuneLite version

RuneLite version: 1.11.2
Launcher version: 2.6.3

Felanbird:
The plugin defines `attackers` as anyone interacting with you, it's more-or-less just bad naming but I'm not sure what good naming is, as `hide interacters` is more-correct but confusing, and if it said `hide interactions` I would expect to hide any NPC that interacted with you, even if they stop.

Jordan-Cousins:
I figured it was something like that, since it follows the same tracking of an NPC attacking you.

While a bit kludgy, it would *technically* be possible to modify the [line where the drawing is determined for NPC "attackers"](https://github.com/runelite/runelite/blob/5776d89df0a0885624e0df7d147873faf88302dd/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java#L248) and do a check against known non-attacker IDs like pets. Since there's already another kludge in that block, I don't think it would be the worst thing in the world, but since I don't think those IDs are already enumerated anywhere, and they're not sequential, it would have to be manually updated each time a new pet was added.

Thanks for the explanation! Doesn't seem super feasible for how minor of a "bug" it is (if the goal is to actually be attackers, versus the issue just being a nomenclature thing about interacters), but it would be cool if it was changed some day!

PR #18804, opened 2025-02-28
entity hider: Hide own pet separately from attackers

cdfisher:
Changes Entity Hider to ignore the local player's own pet when hiding attackers.

Adds a separate "Hide own pet" config option so the previous behavior can be retained. (I've spent enough time in #support to know *someone* will want that)

Related issue: #18801 

Also includes a couple typographical fixes in Entity Hider config item descriptions as a bonus.


OzanKurt:
Yes please!

PR #18810, opened 2025-03-01
cluescrolls: include viggora location in solution text

abextm:
If he is in the edgeville dungeon the arrow won't show on the overworld map, which is confusing.


runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
```diff
@@ -1954,6 +1954,22 @@ private static WorldPoint getViggoraLocation(ClueScrollPlugin plugin)
 		}
 	}
 
+	private static String getViggoraLocationString(ClueScrollPlugin plugin)
+	{
+		int varb = plugin.getClient().getVarbitValue(Varbits.VIGGORA_LOCATION);
+		switch (varb)
+		{
+			case 1:
+				return "the Rogue's Castle";
```
Nightfirecat, line 1963: "the" could be moved into the clue's text instead of being in each case here, but that's not a big deal

Nightfirecat (approved):

runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
abextm: If he ever moves in the future the location might not include a `the`

PR #18812, opened 2025-03-02
NPC aggression timer: clear aggro lines when disengaging from flagged NPCs

samszotkowski:
Currently, once the area lines turn on they always stay active with this combination of settings:

image

What I would expect with these settings is for the overlay to disappear when I am far away from any NPCs I've named + my current slayer task.

I've changed one line so that this now happens. Now upon teleporting away and stuff like that (game state > LOGGED_IN), it checks if it's supposed to continue showing lines.

It appears that Adam has gone back and forth on this line a handful of times (see commits around Feb 2024) so maybe he will have some thoughts on this.


samszotkowski:
P.S. my current workaround for this bug is to toggle either "always active" or "show on slayer task" on and off again, that makes it call `scanNpcs()`

samszotkowski:
Fixes #18819

Adam-:
I've been trying to move this plugin away from the "scan all npcs all the time" approach. Can you think of a better solution?

PR #18844, opened 2025-03-16
#14760 Fix Mouse Highlight Overlay Compilation Errors

Swatts96:
This PR resolves compilation issues in MouseHighlightOverlay.java related to:

1. Type Mismatch: Converted java.awt.Rectangle to net.runelite.api.geometry.RectangleUnion.Rectangle to match expected types.
2. Method Error: Replaced .contains(int, int) with a manual boundary check, ensuring compatibility with RectangleUnion.Rectangle.

- Changes Made:
- Updated minimap widget bounds conversion to use new Rectangle(x, y, width, height).
- Replaced contains() with manual boundary checking for mouse position.
- Ensured the highlight does not appear when the mouse is over the minimap.
- 

Testing:
- Successfully compiles without errors.
- Expected behavior: The overlay properly highlights elements except when inside the minimap.

Notes:
If further issues arise, we may need to check RectangleUnion.Rectangle for additional helper methods.
Open to any feedback on better handling of bounds conversion.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mousehighlight/MouseHighlightOverlay.java
@@ -112,13 +112,14 @@ public Dimension render(Graphics2D graphics)
 		Rectangle minimapBounds = new Rectangle(awtBounds.x, awtBounds.y, awtBounds.width, awtBounds.height);
 
 		// If the mouse is inside the minimap, don't display the highlight
-		if (mousePos.getX() >= minimapBounds.x1 &&
-			mousePos.getX() <= minimapBounds.x2 &&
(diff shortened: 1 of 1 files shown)


Nightfirecat:
This code does not fix the bug described in #14760.

Also, did you write this code yourself? I don't understand at all why `net.runelite.api.geometry.RectangleUnion.Rectangle` is used here.

Swatts96:
This is my first contribution; I was attempting to align the tooltip behavior with how other UI elements, like the OSRS world map tooltips, function when extending beyond their respective views.

As for the use of RectangleUnion.Rectangle, I was originally trying to reference the minimap bounds in relation to the tooltip display. I think that java.awt.Rectangle may be better choice here since getBounds() already returns it. I'll look a little deeper into the project and re-assess before trying to merge a commit again.

PR #18865, opened 2025-03-25
Add configurable infobox overlay color

### LINKED ISSUE #13585: Infobox Overlay with its own Color Options
Currently there is only one option for "Overlay Color" and it changes everything. The interface, infobox etc.

image

A separate option for **Infobox Overlay for Color and Opacity** would be perfect. It allows further customization.

aaronenberg:
Closes #13585.

This adds support for configuring the overlay color specifically for infoboxes, separately from other types of panels.


aaronenberg:
I see this change is also being done as part of #18554.

PR #18871, opened 2025-03-27
Added Low Alchemy to Item Prices plugin

eX-C0n:
Tested and working as expected.
By default, low alchemy is set to disabled and must be enabled manually for those who wish to have it.
And this time around, removed the extra whitespace that was added.

image


Dynrothe:
Would love to see this added! 😃

PR #18897, opened 2025-04-06
item stats: update castle wars brew

Macweese:
Since a recent update to the Castle Wars minigame, the stat changees for Castle Wars Brews are no longer accurate. This PR intends on making the necessary adjustments to provide accurate stat boost information for the potion.

Through testing ingame I was able to confirm (at level 99) that the potion behaves all at once as a:
- Super restore
and benefits from the holy wrench effect.
- Super strength
- Super attack
- Super defence
- Ranging potion
- Imbued heart
- Stamina potion

Tests were also added.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/CastleWarsBrew.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/CastleWarsBrew.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2025 Macweese
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/CastleWarsBrew.java
```diff
@@ -0,0 +1,72 @@
...
+			boost(MAGIC, perc(.10, 1)),
+			boost(RANGED, perc(.10, 4)),
+			heal(RUN_ENERGY, 20)
+		);
+
+		StatsChanges changes = new StatsChanges(0);
+		changes.setStatChanges(Stream.of(
+				Stream.of(superRestore.calculate(client).getStatChanges()),
+				Stream.of(boosts.calculate(client).getStatChanges()))
+			.reduce(Stream::concat)
```
LlemonDuck, line 63: just use Stream.concat in the first place

LlemonDuck (changes requested):

PR #18912, opened 2025-04-10
Moons of Peril boss death status varbits

SuperNerdEric:
I did a few rounds in different orders to verify that these Varbits correlate to these Moons of Peril bosses and they all reset to 0 when the chest is opened. I plan on using them in my combat-logger plugin, but I figured someone else might find them useful.

image

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/Varbits.java
+++ b/runelite-api/src/main/java/net/runelite/api/Varbits.java
@@ -944,9 +944,9 @@ public final class Varbits
 	 * Moons of Peril boss death status
 	 * 0=alive 1=dead
 	 */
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Firstly, you can use `getVarbitValue()` with a plain int in your hub plugin. Secondly, [today's client update includes gameval definitions directly from cache](https://github.com/runelite/runelite/commit/501f22bf20b9c2e9ad9cee50366535d6a12c3416) which includes varbit names, so `Varbits` is considered deprecated, and gameval.VarbitID should be used instead.

SuperNerdEric:
Of course the day I decided to finally figure out which Varbits these are 🙃

PR #18921, opened 2025-04-12
fix mark of darkness timer when purging staff equiped 

eititan:
Fixes Mark of Darkness duration and cooldown timers when wearing the Purging Staff.
 - If the staff is equipped when casting, the duration will be 5x. It can be unequipped (even banked) while the mark is active. 

Took most of this code from stale PR here: https://github.com/runelite/runelite/pull/18343 and aims to only fix MoD issue rather than adding other timers for While Guthix Sleeps.

Fixes https://github.com/runelite/runelite/issues/18128
Discussion https://github.com/runelite/runelite/discussions/18123


Nightfirecat:
Added in #18343 (I'd already been testing/reviewing that prior to this PR, sorry for the delay on my part)

PR #18929, opened 2025-04-15
mes: correctly apply shift deposit swaps to deposit boxes and GIM storage

cdfisher:
Currently, regardless of the in-game "Bank Item Options" setting being on or off, the MES setting "Bank deposit shift-click" is being applied incorrectly in the following cases:

For deposit boxes:
If the deposit box is set to deposit-1, the MES shift-deposit swap is never applied
If the deposit box is set to any other value and MES is set to deposit-all, the menu entry is swapped to deposit-x instead, and when MES is set to deposit-x, the entry is swapped to deposit-all.

For GIM group storage:
If the MES shift-deposit swap is set to deposit-1, the swap is not applied.

This PR resolves the above behavior.
I tested it with "Bank Item Options" both on and off, and checked at the MLM and Barbarian Assault deposit boxes. 
The group storage change was tested by Zoinkwiz as I don't have access to a GIM account.


cdfisher:
Updated to include a fix for GIM group storage as well.

PR #18941, opened 2025-04-23
overlay: Add custom font support

DustyRealm:
Introduce a custom FontType and configuration options to the Overlay settings. All .ttf and .otf files in .runelite/fonts/ will be automatically loaded. Additionally, all system fonts are available for use.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/config/FontStyle.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/FontStyle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Tanner <https://github.com/Reasel>
+ * Copyright (c) 2025, Hamish <https://github.com/DustyRealm>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -24,45 +24,21 @@
  */
 package net.runelite.client.config;
 
-import lombok.RequiredArgsConstructor;
 import lombok.Getter;
-import net.runelite.client.ui.FontManager;
-import javax.swing.text.StyleContext;
+import lombok.RequiredArgsConstructor;
 import java.awt.Font;
 
-
 @RequiredArgsConstructor
-public enum FontType
+public enum FontStyle
 {
-	REGULAR("Regular", FontManager.getRunescapeFont()),
-	BOLD("Bold", FontManager.getRunescapeBoldFont()),
-	SMALL("Small", FontManager.getRunescapeSmallFont()),
-	CUSTOM("Custom", FontManager.getRunescapeSmallFont());
+	PLAIN("Plain", Font.PLAIN),
+	BOLD("Bold", Font.BOLD),
+	ITALIC("Italic", Font.ITALIC);
 
 	@Getter
 	private final String name;
 	@Getter
-	private final Font defaultFont;
-
-	/**
-	 * @deprecated This method is deprecated because handling CUSTOM font types requires a {@code RuneLiteConfig}.
-	 * Use {@link #getFont(RuneLiteConfig)} instead.
-	 */
-	@Deprecated
-	public Font getFont()
-	{
-		return getFont(null);
-	}
-
-	public Font getFont(final RuneLiteConfig config)
-	{
-		if (this == CUSTOM && config != null)
-		{
-			return StyleContext.getDefaultStyleContext()
-				.getFont(config.customFont(), config.customFontStyle().getStyle(), config.customFontSize());
-		}
-		return defaultFont;
-	}
+	private final int style;
 
 	@Override
 	public String toString()
--- a/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
@@ -424,6 +424,43 @@ public interface RuneLiteConfig extends Config
 	}
 
 	@ConfigItem(
+		keyName = "customFont",
+		name = "Custom font",
+		description = "Place TTF/OTF fonts in .runelite/fonts/ and enter the font family name here, e.g., Serif. All system fonts are also available.",
+		position = 45,
+		section = overlaySettings
+	)
+	default String customFont()
+	{
+		return FontType.CUSTOM.getDefaultFont().getFamily();
+	}
+
+	@ConfigItem(
+		keyName = "customFontStyle",
+		name = "Custom font Style",
+		description = "Configures the custom font style.",
+		position = 46,
+		section = overlaySettings
+	)
+	default FontStyle customFontStyle()
+	{
+		return FontStyle.PLAIN;
+	}
+
+	@Range
+	@ConfigItem(
+		keyName = "customFontSize",
+		name = "Custom font size",
+		description = "Configures the custom font size.",
+		position = 47,
+		section = overlaySettings
+	)
+	default int customFontSize()
+	{
+		return 12;
+	}
+
+	@ConfigItem(
 		keyName = "sidebarToggleKey",
 		name = "Sidebar toggle key",
 		description = "The key that will toggle the sidebar (accepts modifiers).",
--- a/runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
@@ -29,9 +29,15 @@ import java.awt.FontFormatException;
 import java.awt.GraphicsEnvironment;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Stream;
 import javax.swing.text.StyleContext;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.client.RuneLite;
 
+@Slf4j
 public class FontManager
 {
 	@Getter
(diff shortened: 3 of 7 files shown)


runelite-client/src/main/java/net/runelite/client/config/FontType.java
```diff
@@ -27,19 +27,42 @@
...
+
+	/**
+	 * @deprecated This method is deprecated because handling CUSTOM font types requires a {@code RuneLiteConfig}.
+	 * Use {@link #getFont(RuneLiteConfig)} instead.
+	 */
+	@Deprecated
+	public Font getFont()
+	{
+		return getFont(null);
+	}
```
DustyRealm, line 55: I've made the assumption that plugins could be using `getFont()` so we need to be careful about removing it. This approach should mean that any remaining uses will continue to work, they just won't make use of the custom font.

I'm also all ears if there's a better way to get access to the `RuneLiteConfig` without passing it in :raised_hands:

Adam-:
I am not opposed to supporting custom fonts. I think the ideal solution would allow choosing a different custom font for each of these 
image
and not supporting only 1 custom font globally.

I am also curious why you are removing the registerFont() calls on the fallback fonts?

runelite-client/src/main/java/net/runelite/client/ui/FontManager.java
```diff
@@ -54,31 +60,25 @@ public class FontManager
...
 			// bold
-			Font boldFont = Font.createFont(Font.TRUETYPE_FONT, inRunescapeBold)
-				.deriveFont(Font.BOLD, 16);
+			Font boldFont = Font.createFont(Font.TRUETYPE_FONT, inRunescapeBold);
 			ge.registerFont(boldFont);
 
 			runescapeBoldFont = StyleContext.getDefaultStyleContext()
-				.getFont(boldFont.getName(), Font.BOLD, 16);
-			ge.registerFont(runescapeBoldFont);
+				.getFont(boldFont.getFamily(), Font.BOLD, 16);
```
DustyRealm, line 81: This change isn't strictly necessary for my change, but it shifts the font loading / registering to be more aligned with how I understand fonts to work.

runelite-client/src/main/java/net/runelite/client/config/RuneLiteConfig.java
```diff
@@ -423,6 +423,43 @@ default Color overlayBackgroundColor()
...
+	@ConfigItem(
+		keyName = "customFontSize",
+		name = "Custom font size",
+		description = "Configures the custom font size.",
+		position = 47,
+		section = overlaySettings
+	)
+	default int customFontSize()
+	{
+		return 12;
```
DustyRealm, line 460: Effectively defaults to a smaller `RuneScape Small`

DustyRealm:
> I am not opposed to supporting custom fonts. I think the ideal solution would allow choosing a different custom font for each of these.

Agreed, do you have any idea of how that UI should function? Would we be happy with 3 copys of the 3 new config options? I was going for a minimally invasive change initially, definitely open for discussion :relaxed: 

> I am also curious why you are removing the registerFont() calls on the fallback fonts?

From my understanding it's unnecessary; registering the underived result from `createFont` into the GE makes it available. Then we just need to set the constants with our styled / sized font. I _think_ all the extra calls were there erroneously.

Adam-:
> From my understanding it's unnecessary; registering the underived result from createFont into the GE makes it available. Then we just need to set the constants with our styled / sized font. I think all the extra calls were there erroneously.

I tested it and it seems to work okay with just the first regsiterFont(). The StyleContext.getFont() call is to cause a composite font to be created (in FontUtilities.getCompositeFontUIResource) so that it can fallback to I think the "dialog" font for glyphs that don't exist. We used to rely on this several places, but since fallback fonts don't work on macos, we do not anymore. Although some plugins still do.

> Agreed, do you have an idea of how that UI should function?

In addition to the regular/bold/fonts you could list all of the fonts in the fonts/ directory? The custom notification sound works sort of like this currently.

DustyRealm:
> The StyleContext.getFont() call is to cause a composite font to be created..

Ah brilliant, thanks for the concrete "why" of the StyleContext part. Thanks for validating on your machine too -

> In addition to the regular/bold/fonts you could list all of the fonts in the fonts/ directory?

Yea I like this. Ideally the builtin fonts wouldn't be a special case at all, and just get loaded in like the others (although maybe still sourced from the package).

I was thinking making the dropdowns show available fonts would be good, but didn't want a mega list. Maybe we _only_ make `RuneScape` / `RuneScape Small` / `RuneScape Bold` and all fonts in `.runelite/fonts` available. Or maybe it's actually fine as long as we have system fonts at the bottom of the list. I'll have a play around!

DustyRealm:
image
image
image

Have pushed a more robust solution, what do we think?
Allows selection of the built-in runescape fonts, system fonts, and custom fonts.

Are there any considerations around removing `FontType` or changing config names / types like I am here?

runelite-client/src/main/java/net/runelite/client/config/FontStyle.java
```diff
@@ -24,22 +24,20 @@
...
 import lombok.Getter;
-import net.runelite.client.ui.FontManager;
-
+import lombok.RequiredArgsConstructor;
 import java.awt.Font;
 
 @Getter
 @RequiredArgsConstructor
-public enum FontType
+public enum FontStyle
```
Adam-, line 33: you could probably just delete this (though I realize it is used as an enum for the panel). In awt, plain/bold/italics are actually a bitmask (eg so you can have bold+italic), so you may want to consider a JList instead.

Adam- (changes requested):

Adam-:
This seems more or less fine, I've left some comments on some of the bigger problems.

> Are there any considerations around removing FontType or changing config names / types like I am here?

These plugins use it:
image

So I probably wouldn't rename it without a good reason.

DustyRealm:
Thanks for the review :raised_hands: good comments, in agreement with all of them :+1: 

Ah yeap, I don't break any functionality in `FontType`, so don't need to delete it in the creation of `FontStyle`. Good tip on `style` actually being a bitmask though, I might drop `FontStyle` entirely and replace it with a couple of checkboxes.

Can I ask how to pull in  / check if there are any plugins using code I'm changing? Doesn't look like it's just a `git clone --recurse-submodules` on the plugin-hub project.

runelite-client/src/main/java/net/runelite/client/config/RuneLiteFont.java
```diff
@@ -0,0 +1,77 @@
...
+import lombok.NoArgsConstructor;
+import lombok.With;
+import net.runelite.client.ui.FontManager;
+
+@ConfigSerializer(RuneLiteFontSerializer.class)
+@AllArgsConstructor
+@NoArgsConstructor(force = true)
+@Getter
+@With
+public class RuneLiteFont
```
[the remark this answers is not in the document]
DustyRealm: I started with `FontConfig` which I also didn't like ..definitely open to ideas, nothing right-sized coming to me just yet.

runelite-client/src/main/java/net/runelite/client/plugins/config/FontConfigPanel.java
```diff
@@ -0,0 +1,256 @@
...
+
+@AllArgsConstructor
+final class TruncatedTitleCaseListCellRenderer extends TitleCaseListCellRenderer
+{
+	private final int maxLength;
+
+	@Override
+	public Component getListCellRendererComponent(JList<?> list, Object value, int index, boolean isSelected, boolean cellHasFocus)
+	{
+		Component component = super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
```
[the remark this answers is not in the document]
DustyRealm: Sweet, it's not a lot of logic to duplicate, and if I replace the `FontStyle` enum with a couple of checkboxes I can ditch the enum part of the logic entirely in my renderer - agreed :+1:

runelite-client/src/main/java/net/runelite/client/config/FontStyle.java
DustyRealm: Great tip, I think checkboxes for `Bold`, `Italic` would be a good fit here to ditch the enum entirely. I only had it for easy option selection / style `int` mapping :+1:

DustyRealm:
Okay, changes pushed to address all the comments apart from the one about a better name for `RuneLiteFont` ..very keen for ideas there :sweat_smile:

I'm keen to keep the config key names the same, but change the function names e.g `fontType()` to `dynamicOverlayFont()`. I'm wondering if the approach there is to keep the old config items there under a new key with `hidden = true`? (if it's an issue). 
```
	@Deprecated
	@ConfigItem(
		keyName = "legacyFontType",
		name = "Dynamic overlay font",
		description = "Configures what font type is used for in-game overlays such as player name, ground items, etc.",
		position = 30,
		section = overlaySettings,
		hidden = true
	)
	default FontType fontType()
	{
		return FontType.SMALL;
	}
```

Curiously RuneLiteFont also provides a `getFont()` function that returns a `Font` so if left the same, calling code wouldn't _necessarily_ need to be updated .. is it a correct vs convenient call?

DustyRealm:
Keen to not muddy this PR with the idea, but I think it'd be nice if we could define these second level panels using the same `@ConfigItem` syntax.

Feature-wise to support this fonts panel, I think we would just need to add an `options` parameter (potentially a `java.util.functions.Supplier`) on ConfigItem that would turn `String` config options into dropdown menus.

Perhaps I could look at that next and refactor both the notification and font panels to use it. Solution might be slightly different from that, but just a bit too much boilerplate in adding these (although that does come with complete flexibility).

Adam-:
I've pushed a few commits here https://github.com/Adam-/runelite/tree/pr_18941 since it doesn't allow me to push to this pr for some reason (did you turn off the setting for that?)

> Can I ask how to pull in / check if there are any plugins using code I'm changing? Doesn't look like it's just a git clone --recurse-submodules on the plugin-hub project.

We use this tool to check the apis plugin access https://abextm.github.io/plugin-hub/

> I'm somewhat keen to keep the config key names the same, but change the function names e.g fontType() to dynamicOverlayFont(). I'm wondering if the approach there is to keep the old config items there under a new key with hidden = true? (if it's an issue).

You have to keep the config key names the same, or else it will wipe peoples font settings / also the backwards compatibility I just had you add would be useless. Just keep the config names and don't change them. Changing the function names is okay, but it might break plugins. It looks to be limited to cox-scouter-external which uses interfaceFontType(). 

> Keen to not muddy this PR with the idea, but I think it'd be nice if we could define these second level panels using the same @ConfigItem syntax. 

I wouldn't mind a way to let plugins add custom panels to the config. It's something that has been requested for awhile. I am not sure what it would look like though. And certainly not as part of this pr.

Adam-:
added in b9ea1f8ff5fbeac7594e8dc6bdbf31ec1023fd43

Issue #17863, opened 2024-05-19
Idle Notifier does not notify when auto-retaliate stops after 20 minutes afk


ebartram1:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

Idle Notifier is activated with all boxes checked. After 20 minutes of being attacked such as in NMZ or splashing, auto-retaliate automatically turns off and the character stops attacking while continuing to be attacked. The character turns to face the attackers but does not attack them back. I believe that Idle Animation Notifications should trigger a notification in this scenario. 

### RuneLite version

runelite: 1.10.30
oldschool: rev 221
launcher: 2.7.1

lexigorse:
It does look like there's a bit to cover some scenarios of whether you are in or out of combat. I wonder if there's something specific to NMZ that doesn't trigger it. I might look into it if I have time.

Snippet that I found if someone beats me to it

`// We are in combat, so now we have to check for the timer that knocks you out of combat
		// I think there are other conditions that I don't know about, because during testing I just didn't
		// get removed from combat sometimes.
		final long lastInteractionAgo = System.currentTimeMillis() - client.getMouseLastPressedMillis();
		if (lastInteractionAgo < COMBAT_WARNING_MILLIS || client.getKeyboardIdleTicks() < COMBAT_WARNING_CLIENT_TICKS)
		{
			notifyIdleLogout = true;
			return false;
		}`

ebartram1:
> It does look like there's a bit to cover some scenarios of whether you are in or out of combat. I wonder if there's something specific to NMZ that doesn't trigger it. I might look into it if I have time.

I just tested 20 minutes of splashing outside of NMZ and the same problem happens where the player stops splashing but no notification happens.

tylerwgrass:
The [combat timer](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java#L470-L482) only checks if the player was hitsplat and doesn't check they are also attacking.

Possible solution is to track both the player and the target NPC hitsplats with countdowns.

PR #18942, opened 2025-04-23
implement feature request: auto-retaliate idle notification

hawolt:
as requested in issue #17863 this PR adds a notification for when you stop auto retaliating

in addition to the feature I also went ahead and replaced the remaining deprecated calls in the class

a description of the logic to achieve this feature:

we use an Instant that references the time combat was entered, this is achieved by tracking clicked Menu Options to set or unset an instant based on the Option, in addition to this since combat can also be invoked by any agressive NPC and it can also be set by having an interacting NPC target when no Instant is set yet, we should not override this value as it is required to properly track that 20 minutes have elapsed. with this Instant as a reference and the current idle time we can dispatch a notification after combat has been engaged and 60_000 idle ticks (20 minutes) have passed in addition to atleast 20 minutes since combat has been invoked. we need a reference to the timer as when 20 minutes pass and auto-retaliation stops onInteractingChanged is continously called with the NPC that is attacking you, followed with the same event with a null since the client will not retaliate and we want to avoid continously sending the same notification whilst idle

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -492,6 +492,7 @@ public class IdleNotifierPlugin extends Plugin
 	public void onGameStateChanged(GameStateChanged gameStateChanged)
 	{
 		lastInteracting = null;
+		lastCombatStarted = null;
 
 		GameState state = gameStateChanged.getGameState();
 
@@ -913,7 +914,6 @@ public class IdleNotifierPlugin extends Plugin
 			{
 				lastInteract = null;
 				lastInteracting = null;
-				lastCombatStarted = null;
 
 				// prevent animation notifications from firing too
 				lastAnimation = -1;
@@ -1063,6 +1063,7 @@ public class IdleNotifierPlugin extends Plugin
 
 		// Reset interaction idle timer
 		lastInteracting = null;
+		lastCombatStarted = null;
 		if (gameState == GameState.LOGIN_SCREEN || local == null || local.getInteracting() != lastInteract)
 		{
 			lastInteract = null;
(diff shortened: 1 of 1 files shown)


hawolt:
I see a test has failed, I'll update that one as well

hawolt:
I have adjusted the code slightly to behave identical without causing the test to fail as I am unable to figure out how to mock time, I also realized that the Notification for the idle logout which requires 19 minutes to pass also has no test currently, I assume the reason is because mocking Instant does not appear to be working on this Mockito version

Kyahzworld (approved):

rsbfox:
You cannot use getMouseIdleTicks() because mouse movement alone does not reset auto-retaliate idle timer. 

Clicks have to be in game too. Clicking on RL itself isn't enough. I assume getMouseLastPressedMillis() should work.

hawolt:
> You cannot use getMouseIdleTicks() because mouse movement alone does not reset auto-retaliate idle timer.
> 
> Clicks have to be in game too. Clicking on RL itself isn't enough. I assume getMouseLastPressedMillis() should work.

I have re-used the same thing used for the idle logout, shoudnt the idle logout be incorrect as well then?

```
	private boolean checkIdleLogout()
	{
		// Check clientside AFK first, because this is required for the server to disconnect you for being first
		final int idleClientTicks = Math.min(client.getKeyboardIdleTicks(), client.getMouseIdleTicks());
```

I don't mind changing it but I would adjust this at the same time if my assumption is correct

rsbfox:
Mouse movement does prevent idle logout yes it's a client side timer. The retaliate timer requires mouse click or keyboard input. I assume because those are sent to server.

The timer also starts when just starting combat by auto-retaliating so I think lastCombatStarted logic needs some changes to work reliably.

Easy nightmare zone with some absorbs is best way to test. It's how I noticed what counts. Then noticed wiki mentions the same. https://oldschool.runescape.wiki/w/Idle#Combat_timers

rsbfox:
Top timer was started when last clicked client. Bottom timer was when the mobs spawned and I started fighting back.

image

So the combat idle timer is actually a client input idle timer. So no need to check time combat started at all.

I think plugin should just check if auto-retaliate is turned on then check if last keyboard or mouse click was over 20 mins ago whenever player is attacked. Plus some logic to prevent extra notifications till next client input.

PR #18944, opened 2025-04-25
Widget inspector: Update for gamevals

crab-pancake:
Update widget inspector devtool to use exposed config names

Not sure if the getDeclaredFields block is necessary: pretty sure (but didn't check) that there is an inner class for every field


crab-pancake:
just did a basic ctrl-F of the InterfaceID class and there does appear to be the same number of _0000 ids

runelite-client/src/main/java/net/runelite/client/plugins/devtools/WidgetInspector.java
```diff
@@ -335,9 +334,21 @@ static String getWidgetName(int componentId)
 			//until it's actually needed.
 			try
 			{
-				for (Field f : ComponentID.class.getDeclaredFields())
+				for (Field field : InterfaceID.class.getDeclaredFields())  // is this block unnecessary?
```
Adam-, line 337: This would assign the interface ids to any component with id 0 which is not totally correct I think.

PR #18948, opened 2025-04-27
Implemented !log command to chatcommands plugin

touchtouch105:
Implemented !log command to chatcommands plugin


touchtouch105:
Also to note, using this change would allow for the quantity of pets to show, !log pets results in the following

image

PR #18949, opened 2025-04-27
slayer: use inital count VarPlayer

ldavid432:
get the initial task amount from the varp instead of just storing it locally

The varp is always available, but does not reset after completing a task however, that's not really an issue as there is not overlay if you don't have a task.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -258,15 +258,10 @@ public class SlayerPlugin extends Plugin
 	{
 		switch (event.getGameState())
 		{
-			// client (or with CONNECTION_LOST, the server...) will soon zero the slayer varps.
-			// zero task/amount so that this doesn't cause the plugin to reset the task, which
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -258,15 +258,10 @@ public void onGameStateChanged(GameStateChanged event)
...
 		{
-			// client (or with CONNECTION_LOST, the server...) will soon zero the slayer varps.
-			// zero task/amount so that this doesn't cause the plugin to reset the task, which
-			// would forget the initial amount. The vars are then resynced shortly after
 			case HOPPING:
 			case LOGGING_IN:
 			case CONNECTION_LOST:
-				taskName = "";
-				amount = 0;
-				loginFlag = true; // to reinitialize initialAmount and avoid re-adding the infobox
```
ldavid432, line 269: Not sure if removing this is correct... it seems like it was added to avoid clearing the initial amount which we don't need to avoid anymore

Nightfirecat:
Added in c84b694d99e50d73fbdb5e80f6183a02a7212c19

PR #18950, opened 2025-04-27
cluescrolls: correct enemy value for 05.13N 04.16W master step

cdfisher:
The master coordinate clue step "05 degrees 13 minutes north, 04 degrees 16 minutes west" has the player dig right near a multicombat boundary so can spawn either ancient wizards or a Brassican mage, though the overlay only mentions the mage.

Reported by Shush in the wiki Discord server.


cdfisher:
Attaching this image that was shared with the initial report here since it was part of a Discord discussion that's already long-buried. It turns out there are two tiles that are multicombat in an area that's otherwise singles.
coordinate-clue-mages

PR #18954, opened 2025-04-29
refactor deprecated API usages

hawolt:
noticed a lot of deprecated usages so I went ahead and fixed up as many as I could


hawolt:
had to rollback changes for a few files as I could not figure out how to properly mock to fix the tests

PR #18956, opened 2025-05-01
fix client thread blocking from plugin hub interactions

hawolt:
when opening the plugin hub on older computers the client stops taking any user input until the blocking operations are done, this is caused by invoking SwingUtilities.invokeLater, which even when called within a Runnable in an Executor will be executed differently. In addition every interaction with the search bar fired a seperate search, each blocking the client thread from taking any user input yet again for every key stroke, by introducing a debouncer this freeze can be optimized/minimized by firing less filter searches.

I checked for an existing debouncer within the project and could only find debounce logic within the WikiSearch, which I refactored to use the dedicated debouncer which is added as part of this PR


hawolt:
fixed by adam 381a4679d4de16a054104ea08add98e4d01f6c7e

PR #18958, opened 2025-05-01
account: correctly parent JOptionPane

### LINKED ISSUE #18933: Always on top takes priority over displaying the sign out confirmation dialog menu
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

If you have always on top enabled and sign out of RuneLite, the sign out confirmation dialog will be forced to the background, causing the client to stop accepting inputs with no way to close the dialog menu.

### Screenshots or videos

_No response_

### RuneLite version

RuneLite Version: 1.11.6.1
Launcher Version: 2.7.2

### Logs

_No response_

LlemonDuck:
Without a parent component, the JOptionPane may render underneath the client window if alwaysOnTop is set. This can make the JOptionPane uninteractable while also stealing focus from the frame.

Since the Account plugin doesn't have any owned components, we don't have other options for the parent. I've exposed the ContainableFrame instance from ClientUI, but if we don't want to do that then we'll need another way to provide a parent in non-Swing contexts.

I've also fixed one incorrect use in ProfilePanel, which has its own context available.

There is another instances of null-parenting in LinkBrowser, but since it is in a static context *and* is an extensively used API option, it needs a bit more consideration.

Closes #18933

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/account/AccountPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/account/AccountPlugin.java
@@ -38,6 +38,7 @@ import net.runelite.client.events.SessionOpen;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.ui.ClientToolbar;
+import net.runelite.client.ui.ClientUI;
 import net.runelite.client.ui.NavigationButton;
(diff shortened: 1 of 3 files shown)


jptrs93:
Would love to see this merged. We are having similar issues with window ordering in our plugin and would like to make use of the exposed ContainableFrame from the ClientUI.

PR #18984, opened 2025-05-09
banktags: Fix potion storage's potions to have duplicate-item & remove-layout in an active layout

### LINKED ISSUE #18983: [Bank Tags] Potion Storage Potions are missing bank tag management options
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Potions that appear in a Bank Tag layout (core) are missing the following options:

- Remove tag (x)
- Duplicate Item
- Edit Tags (x)

If the item is in the bank itself, then these options appear, however if the potions are not found in the bank and are only in the potion storage then these options do not show.

### Screenshots or videos

In the following image, the Prayer Potions are found ONLY in potion storage, thus are missing the options:
Image

In the following image, the Divine Super Combat potion is found both in Potion Storage AND in the bank, thus all options appear:
Image

### RuneLite version

RL Version: 1.11.7
Launcher Version 2.7.1

### Logs

_No response_

JZomDev:
Switching from Examine to Withdraw-1 & detect if an item is in potion storage allows Duplicate-Item and Remove-layout to be added, because Examine is not on a potion coming from potion storage. 

Fixes #18983

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
@@ -771,7 +771,7 @@ public class TabInterface
 	{
 		if ((activeOptions & BankTagsService.OPTION_ALLOW_MODIFICATIONS) != 0
 			&& event.getActionParam1() == InterfaceID.Bankmain.ITEMS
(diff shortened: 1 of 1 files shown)


Adam-:
added in b9c6ac741a0b130f0930760626068a80c96e8505

PR #18993, opened 2025-05-14
itemcharges: update binding necklace used pattern to match new message

### LINKED ISSUE #18992: Item charges not updating automatically
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

Item charges in this case binding necklaces not updating automatically, not even in safe mode as you can see in the provided video.

Is there a way to fix this since it has worked properly in the past, thank you.

### Screenshots or videos

[[](url)](https://streamable.com/7hx9s1)

### RuneLite version

1.11.8
2.7.1

### Logs

[client.log](https://github.com/user-attachments/files/20214259/client.log)
[launcher.log](https://github.com/user-attachments/files/20214258/launcher.log)

cdfisher:
Fixes #18992

The game message for making combo runes changed slightly with today's game update. This updates the regex to match either the new message or the old one, in case the change gets reverted.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemcharges/ItemChargePlugin.java
@@ -46,12 +46,12 @@ import net.runelite.api.ItemContainer;
 import net.runelite.api.events.ChatMessage;
 import net.runelite.api.events.GameStateChanged;
 import net.runelite.api.events.ItemContainerChanged;
-import net.runelite.api.events.ScriptCallbackEvent;
 import net.runelite.api.events.VarbitChanged;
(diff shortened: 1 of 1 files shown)


hawolt:
I have also made a PR towards this branch to track using the less error prone VarpId https://github.com/cdfisher/runelite/commit/4d5a752c5d96dd8154d8d48cd0a64af303351224 https://github.com/cdfisher/runelite/pull/2

cdfisher:
Updated to use the relevant varplayer instead of chat messages.

- Varp updates immediately to the maximum value of 16 when the necklace is destroyed, so it's no longer necessary to check the destroy widget
- Varp doesn't reset to 16 immediately when the last charge is used, only when a new necklace is equipped, so the check for the break message is retained.

Issue #18135, opened 2024-08-07
[Key Remapping] Choosing dialogue options with number keys doesn't work with the plugin enabled


TomBtw:
### Type

Crash or hang

### Operating System

Windows

### OS Version

Version	10.0.19045 Build 19045

### Bug description

Pressing the number keys (eg. 1-5) to choose dialogue options while having that key rebound to control (mine is 4) does not work. 

### Screenshots or videos

![image](https://github.com/user-attachments/assets/186a6d65-ea39-45dc-970c-2b5db2dbedfa)

### RuneLite version

RuneLite version: 1.10.35
Launcher version: 2.6.7

PR #18996, opened 2025-05-15
Fix the bug in the Key Mappings plugin with dialog options

jbentleyh:
Fix the bug in the Key Mappings plugin where if you are in a dialog with options to choose, then we ignore the ctrl mapping when it is mapped to a valid integer speech option

Original Issue: [https://github.com/runelite/runelite/issues/18135](https://github.com/runelite/runelite/issues/18135)

Smoke test passes:
Screenshot 2025-05-14 at 10 16 18 PM

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -168,7 +168,7 @@ class KeyRemappingListener implements KeyListener
 			}
 
 			// Allow numeric keys to select dialog options, ignoring CTRL mapping
-			if (plugin.isDialogOpen() && plugin.isOptionsDialogOpen() && config.control().matches(e) && Character.isDigit(e.getKeyChar()))
+			if (plugin.isOptionsDialogOpen() && config.control().matches(e) && Character.isDigit(e.getKeyChar()))
 			{
 				mappedKeyCode = e.getKeyCode();
 			}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/keyremapping/KeyRemappingListener.java
@@ -162,15 +162,9 @@ class KeyRemappingListener implements KeyListener
 				mappedKeyCode = KeyEvent.VK_SPACE;
 			}
 
-			if (config.control().matches(e))
+			if (!plugin.isOptionsDialogOpen() && config.control().matches(e))
 			{
 				mappedKeyCode = KeyEvent.VK_CONTROL;
-			}
-
-			// Allow numeric keys to select dialog options, ignoring CTRL mapping
-			if (plugin.isOptionsDialogOpen() && config.control().matches(e))
-			{
-				mappedKeyCode = e.getKeyCode();
 			}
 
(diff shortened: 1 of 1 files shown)


Adam-:
I do not understand why this is not just:

```java
			if (config.control().matches(e) && !plugin.isOptionsDialogOpen())
			{
				mappedKeyCode = KeyEvent.VK_CONTROL;
			}
```

jbentleyh:
@Adam- You're right I can simplify the if-statement and did. I think you did not mean to add the `!` as we do need to confirm we are in a game state that has dialog options open. 

If we assign `mappedKeyCode = KeyEvent.VK_CONTROL`, that sets the `mappedKeyCode` to `17`, which is not in-range to be a valid speech option and does not behave as you think. (https://docs.oracle.com/javase/7/docs/api/constant-values.html#java.awt.event.KeyEvent.VK_CONTROL). So, we just need to get the actual keyCode of the pressed key.

Adam-:
> @Adam- You're right I can simplify the if-statement and did. I think you did not mean to add the `!` as we do need to confirm we are in a game state that has dialog options open.
> 
> If we assign `mappedKeyCode = KeyEvent.VK_CONTROL`, that sets the `mappedKeyCode` to `17`, which is not in-range to be a valid speech option and does not behave as you think - I confirmed in-game as well. (https://docs.oracle.com/javase/7/docs/api/constant-values.html#java.awt.event.KeyEvent.VK_CONTROL). So, we just need to get the actual keyCode of the pressed key.

I don't really understand anything you are saying at all. Can you describe what you think you are trying to fix?

The purpose of this feature is to remap keys to ctrl for toggling run/walk mode, which your code breaks.

jbentleyh:
@Adam- 

Suppose you click an NPC and a dialog box is open with 3 options to choose from and I want to select an option using a number key instead of clicking it.

If CTRL is bound to the number 2, then when I press 2 while the dialog box is open, dialog option 2 is not selected and nothing happens because 2 is mapped to CTRL.

This code is suppose to allow you select option 2, even though it's bound to CTRL. 

Broadly, this does not break the toggle walk/run - I guess you could say it breaks it if the user has a dialog box with options present on screen? If you think that even in that scenario we should not override CTRL, then this issue should be closed.

jbentleyh:
> The code I gave you changes the behavior to fix this.
> 
> >
It does not.. How could it possibly fix that when you have a `!` in front: `!plugin.isOptionsDialogOpen()`

I tested your code in-game and it does not allow me to select option 2 when CTRL is mapped to 2.



> I am saying that it breaks toggling walk/run when outside of a dialog- which is the normal place where people use it.

How is that possible when the first part of the if-statement is checking if you are currently in a dialog with options: `plugin.isOptionsDialogOpen()`

I've tested my changes thoroughly in-game and nothing you're saying is true.

Adam-:
I see the confusion. My first comment was asking why you aren't changing the existing code to

```java
			if (!plugin.isOptionsDialogOpen() && config.control().matches(e))
			{
				mappedKeyCode = KeyEvent.VK_CONTROL;
			}
```

and you interpreted it as why you aren't adding a new block of code which is this.

This is what I had meant https://github.com/Adam-/runelite/commit/305127855e142de0feb0a0de1fec933a1c2c5663

jbentleyh:
Hah! Ok, apologies for that - we did have a misunderstanding. Yes, that would work and I just pushed the commit - I just testing in-game and works as expected.

PR #18998, opened 2025-05-15
fix decimal point test failures on german locale

hawolt:
this PR fixes these three failing tests on a german Locale

```
[ERROR] Failures: 
[ERROR]   ChatCommandsPluginTest.testSecondsToTimeString:1074 expected:<0:03[.]60> but was:<0:03[,]60>
[ERROR]   CoordinateClueTest.testResourceAreaCosts:92
[ERROR]   CrypticClueTest.forResourceAreaCosts:84
[INFO] 
[ERROR] Tests run: 588, Failures: 3, Errors: 0, Skipped: 2
```

I have also applied auto formatting based on this wiki entry


Nightfirecat:
For what it's worth, we already set English language and US region for surefire: https://github.com/runelite/runelite/blob/4672e0b3b6a8b9e38575d6bee8f700683643c9dd/pom.xml#L198

While passing an English locale to these functions would work, we'd ideally like a solution similar to the above which would work for any and all new tests written

PR #19009, opened 2025-05-19
Update ItemChargePlugin.java to auto track charges for binding necklace

millinoid:
Add binding necklace auto tracking charges when runecrafting combination runes


cdfisher:
This issue was already fixed with 9729eaa4958e848ad977becee1db51820024cbc0

millinoid:
That change doesn't update binding necklaces

cdfisher:
> That change doesn't update binding necklaces

That's exactly what it does.

If you're saying that because you're still encountering the issue with the charges not updating in game, that's because there hasn't been a client release since that change was made.

Issue #18914, opened 2025-04-11
Bank Placeholder feature is gone


Junodude:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Since the OSRS update on April 9, 2025. The placeholder option for MES is no longer working correctly. All the placeholder items I had were replaced with the default "Withdraw 1" option. I changed them back to placeholder. On April 10, 2025, all the items I had set to placeholder the previous day were all switched to the "Withdraw all but 1" option. The option to set them as placeholder is either not there, or still sets the item to "Withdraw all but 1". Image provided

### Screenshots or videos

![Image](https://github.com/user-attachments/assets/f7e2f80b-287a-4d27-b922-9901e190701d)

### RuneLite version

Runelite v 1.11.6.1
Oldschool revision: Rev230
Launcher version: 2.6.9

zconnelly13:
Same here on...

Runelite version 2.7.2
Mac version: Sonoma 14.4.1 (23E224)

Hevenlys:
I am having the same exact issue.

bsh0021:
I am having this exact same issue with the "placeholder" menu swap being set as "Withdraw-All-but-1". I found a workaround using the  "always set placeholders " button on the bottom right of the bank menu and keeping the default "withdraw-1" option. However this does not fix the issue with the menu swap.

![Image](https://github.com/user-attachments/assets/7fecfa68-a9c2-4f8a-b090-1fb0ed9c184c)

bkohrn:
Any progress on this?

Junodude:
> Any progress on this?

No clue. I stopped using this function and now just use the lock placeholder feature that "Bsh0021" suggested on Apr 20. Even if they do fix it, it could just bug out or change again randomly and I'd have to re-placehold every specific item again.

PR #19012, opened 2025-05-19
Fix Menu Entry Swapper Bank Placeholder option

Russ36363:
Fix for https://github.com/runelite/runelite/issues/18914


abextm:
This is wrong if bank_bankops_toggle_on is enabled and the item has configurable charges. LayoutManager also has a variant of this code to support the potion store that needs to be updated I think.

PR #19014, opened 2025-05-20
Add pay-dirt to the mining skill calculator

HelixSpiral:
This adds Pay Dirt from the Motherlode Mine into the mining skill calculator


LlemonDuck:
Mlm grants additional mining xp when cleaning based on the ores received, so I can't really see this being useful as it does not (and cannot) account for that xp.

HelixSpiral:
Fair enough

PR #19019, opened 2025-05-20
clues: Update Anglerfish skill challenge with demonic worms

Plondrein:
After Yama release, one can use demonic worms instead of regular sandworms.


Frosty-J:
I was thinking the text should be "Sandworms" instead of "Worms" to make it clear that king worms (and any other types of worms) won't work. But I didn't know diabolic worms exist... tricky.

PR #19027, opened 2025-05-22
skill calculator: add diabolic worms xp modifier

Macweese:
This adds [diabolic worms](https://oldschool.runescape.wiki/w/Diabolic_worms) as a modifier to the skill calculator's fishing actions, as they give a different amount of experience compared to when anglerfish are caught with sandworms as bait.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/FishingBonus.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/FishingBonus.java
@@ -24,6 +24,8 @@
  */
 package net.runelite.client.plugins.skillcalculator.skills;
 
+import java.util.EnumSet;
+import java.util.Set;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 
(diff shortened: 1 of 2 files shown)


evilgenious448:
I feel like this should be implemented as a Skilling Bonus for Fishing since the Skilling Action is the same, it's just a modifier to the XP rate.

Previously we had 
`GOLD_BAR `
and
`GOLD_BAR_(GOLDSMITH_GAUNTLETS)`

But we changed so that Goldsmith Gauntlets was a modifier in the Smithing Boosts.

Thus, it should be `Diabolic worms (66%)` in the Fishing Bonus and it applies only to Raw Anglerfish.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/FishingBonus.java
```diff
@@ -32,8 +34,17 @@
 public enum FishingBonus implements SkillBonus
 {
 	ANGLERS_OUTFIT("Angler's Outfit", 1.025f),
+	DIABOLIC_WORMS("Diabolic worms", 0.66f),
```
evilgenious448, line 37: "Worms" should have a capital W for captialisation consistency with other Skilling Bonuses.

PR #19030, opened 2025-05-24
Skill Calculator: Add Aether Runes to RunecraftAction

evilgenious448:


jgresham:
I think the bool `ignoreBonus` should be `false`. eg. Aether runes grant bonus xp with daeyalt ess @evilgenious448

evilgenious448:
> I think the bool `ignoreBonus` should be `false`. eg. Aether runes grant bonus xp with daeyalt ess @evilgenious448

You are correct, thanks for pointing this out, I've fixed it up in the latest commit.

PR #19052, opened 2025-06-02
clues: fix incorrect clue itemIds

TheLope:
1. Item ID should be 7276
    - https://oldschool.runescape.wiki/w/Clue_scroll_(medium)_-_GOT_A_BOY
2. Item ID should be 3493
    - https://oldschool.runescape.wiki/w/Clue_scroll_(easy)_-_Search_the_chests_in_the_Dwarven_Mine

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -36,6 +36,7 @@ import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.Singular;
+import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.NPC;
 import net.runelite.api.ObjectComposition;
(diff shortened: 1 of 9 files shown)


Nightfirecat:
I've added a commit to test all the clue types with IDs defined for duplicates. This identified one more place where duplicate IDs are present, namely for the cryptic clues with the following hints:
* Search the chests in the Dwarven Mine.
* Search through some drawers in the upstairs of a house in Rimmington.

From personal experience, I believe the ID given is for the latter clue (the Rimmington drawers) however I recommend seeing that it is tested before committing to that change. Once the tests pass, this fix is good to go.

TheLope:
> I've added a commit to test all the clue types with IDs defined for duplicates. This identified one more place where duplicate IDs are present, namely for the cryptic clues with the following hints:
> 
> * Search the chests in the Dwarven Mine.
> * Search through some drawers in the upstairs of a house in Rimmington.
> 
> From personal experience, I believe the ID given is for the latter clue (the Rimmington drawers) however I recommend seeing that it is tested before committing to that change. Once the tests pass, this fix is good to go.

Confirmed and updated. Tests passing.

TheLope:
Is it intentional this isn't getting picked up by the test since its variations of the same step?
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/AnagramClue.java#L430-L439

These being separate would result in Nieve being the default even if the player only has Steve

PR #19061, opened 2025-06-09
Fix custom cursor on Linux and macOS

Vesylum:
## Summary
- support platform specific Toolkit cursor sizes
- fallback to default cursor if custom one can't be created

## Testing
- `mvn -q -DskipTests install` *(fails: Non-resolvable import POM)*


riktenx:
How many Rs are there in the word "strawberry"?

PR #19084, opened 2025-06-18
Update CoordinateClue.java

matthewfoucher:
Fixed the descriptions for several coordinate steps that implied that their respective dig spots were located in the ocean.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CoordinateClue.java
@@ -79,12 +79,12 @@ public class CoordinateClue extends ClueScroll implements LocationClueScroll
 		CoordinateClue.builder()
 			.itemId(ItemID.TRAIL_CLUE_MEDIUM_SEXTANT025)
 			.location(new WorldPoint(3184, 3150, 0))
-			.directions("South of Lumbridge Swamp.")
+			.directions("Southern Lumbridge Swamp.")
 			.build(),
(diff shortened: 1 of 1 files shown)


Felanbird:
Most of these are copied verbatim from Jagex's version of the hints, which I assume is because they think in terms of where the text is/where the map would direct you, if you typed it into the map search box

matthewfoucher:
> Most of these are copied verbatim from Jagex's version of the hints, which I assume is because they think in terms of where the text is/where the map would direct you, if you typed it into the map search box

Hold up, I'm confused. When you play on mobile, these exact incorrect descriptions appear, yes. But these RuneLite descriptions far predate the mobile clue helper. How were they copied verbatim from Jagex?

Felanbird:
I was thinking about it backwards as so many of the hints that we have actually are exactly the same as the ones Jagex has, but what you pointed out seems correct timeline wise, possibly Jagex used or copied our original hints.

edit: 
Funnily enough I think this is actually a full-circle moment, as I see that `On the summit of Trollheim.` is one of our hints, but in https://github.com/runelite/runelite/pull/18327/files#diff-05d981d47300979b2cfaabc7ef7909d3d41e8de4cb0c64eb51b1daef4aff2af4R225 I explicitly changed it from `summit` to `at the top` when I copied the new hints from Jagex, as I figured it would be less confusing for people to get an exact direction, than a correct but not very common in conversation, word, but it turns out we were probably the original cause for this

matthewfoucher:
Yeah that is what I assumed, that Jagex just copied all of the text from RuneLite. I find it unlikely that they would come up with the exact same wording for their hints.

PR #19085, opened 2025-06-18
xptracker: restore xp tracker state on rsprofile change

cdfisher:
Resolves the behavior described in #19083 in which saved XP tracker sessions are wiped when hopping between world types which use separate RuneScapeProfile types


Adam-:
fixed in 9b6ab8506ef277daeeaf4f597970696d4348aa13

PR #19105, opened 2025-06-26
Update master ca thrall timer to 2min

z-x63:
Updating master ca tier to 2min due to changes by summer sweep up today


Adam-:
This should either check gm || m or just remove the gm check. Probably remove the gm check since you cant have that without master can you?

z-x63:
Makes sense - removed the gm check

Thanks!

Ricardo5423:
Maybe i don't understand, but it seem's that it's not working... My CA timer still at 1:30 min

Nightfirecat:
The client has not yet updated since this code change was merged. Wait for us to update--possibly this week Wednesday

PR #19136, opened 2025-07-09
Allow twitch to use plugins to format messages

### LINKED ISSUE #19135: Emoji and other features non-functional when using ChatMessageManager
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

When a plugin is used to inject a chat message, such as Twitch, plugins such as Emoji are ignored.
The plugins actually do run - they run on .getValue() - but then when insde `ChatMessageManager.add()` the call tp `line.setRuneLiteFormatMessage(message.getRuneLiteFormattedMessage());` ignores the value in `message.getValue()` and reverts all changes made by plugins.
The plugins are called as a result of the `client.addChatMessage` call.

If I change the line to `line.setRuneLiteFormatMessage(message.getValue());` this works correctly, and prior to this line, the value in `message.getRuneLiteFormattedMessage()` is null.

### Screenshots or videos

Image

Image

Image

Image

### RuneLite version

RuneLite Version: 1.11.10.2

### Logs

_No response_

TapGhoul:
resolves #19135

This adds a config value which defaults to the old behavior, but allows people to opt out of the bypassing of plugins to allow for plugins such as Emoji to reformat twitch messages.


TapGhoul:
My most recent CI failure (https://github.com/runelite/runelite/actions/runs/16351991361/job/46200872986) is as a result of master failing CI, not my own plugin. It was failing before due to a checkstyle issue.

TapGhoul:
https://github.com/runelite/runelite/commit/10b6d68d381ea400224fe77fcef051c5ad21bb9d was just pushed to master, so that should solve my CI woes.

Adam-:
This should probably unconditionally use value(). Having a config option to control this is stupid.

TapGhoul:
@Adam- Sure, I can strip it. I put the config in because I wasn't sure if it could conflict with plugins made by others who may have depended on old behavior, but no fuss from my end.

The other case where it doesn't use value() is system/status messages, which I plan to leave using runeLiteFormattedMessage - this should always pass through untouched.

Fiffers:
Well, that explains a lot. Spent a bit of today trying to debug the custom-emoji plugin and get it to work with the twitch plugin, but now I know why nothing worked. Really hope this gets merged :)

TapGhoul:
@Fiffers While you're at it I'd recommend also following #19141 which fixes some of the garbage shown in chat when viewers are using 7tv's anti-antispam thing

PR #19137, opened 2025-07-09
twitch: Added filtering for 7TV ChatSpam module suffix

TapGhoul:
7TV adds a suffix ` \u{E0000}` to messages when sent twice in order to bypass duplicate message detection on twitch. This is an extremely commonly used plugin and feature on Twitch, and renders in game as ??. It's a good idea to filter it.

Before:
image

After:
image


TapGhoul:
If there's a cleaner way to do this, I'm happy to do so - java is not my primary language, so I'm happy to tweak things to be more idiomatic. As it stands though, this works.

The `.stripTrailing()` is to remove the space that's on the end. Normally, twitch strips prefixed and suffixed whitespace for us, but the duplicate message fix bypasses this on the suffix, so I've added it to compensate.

TapGhoul:
Converted to draft as there's more fixes of this nature that would be worth applying, so I'm gonna go add those.

TapGhoul:
I'm going to recreate this MR with a better fix under a different branch.

PR #19180, opened 2025-07-25
Loot tracker: Cleaning dirty arrowtips

samszotkowski:
Cleaning [dirty arrowtips](https://oldschool.runescape.wiki/w/Dirty_arrowtips) processes the entire stack at once, always converting exactly one dirty arrowtip into one clean one. If you don't have enough inventory space it will process as many as it takes to fill your inventory. They never drop onto the floor.

I more or less copied the code from the hunters sacks bit where it calls `addLoot()` directly with an `amount`, since it's kind of similar in that we want to capture multiple events happening at the same time.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -1129,6 +1129,20 @@ public class LootTrackerPlugin extends Plugin
 			{
 				onInvChange(collectInvItems(LootRecordType.EVENT, BIRDNEST_EVENT));
 			}
+			else if (event.getMenuOption().equals("Clean") && event.getItemId() == ItemID.DIRTY_ARROWTIPS)
+			{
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -1129,6 +1129,20 @@ else if (event.getMenuOption().equals("Search") && BIRDNEST_IDS.contains(event.g
 			{
 				onInvChange(collectInvItems(LootRecordType.EVENT, BIRDNEST_EVENT));
 			}
+			else if (event.getMenuOption().equals("Clean") && event.getItemId() == ItemID.DIRTY_ARROWTIPS)
+			{
+				// Entire stack gets cleaned at once, want to track how many
+				onInvChange((((invItems, groundItems, removedItems) ->
+				{
+					int itemId = event.getItemId();
```
Adam-, line 1137: the lifetime of the menu ends after this event is done, and calling getItemId() on it some time later may return the item from a completely different menu

PR #19181, opened 2025-07-25
Skill Calculator: Varlamore Part 3

evilgenious448:
The following Fletching Actions have been added:

- Full Atlatl Dart fletching process including making: (1) Atlatl Dart Shafts, (2) Headless Atlatl Darts, (3) Atlatl Dart Tips, (4) Atlatl Darts
- Redwood Hiking Staff
- Vale Totems


evilgenious448:
Vale Totems do not have any Item directly associated with them, so I used the Veil Offerings item as it was the most relevant one. Unfortunately, this required rewriting all of the enums to support `NamedSkillAction`, similar to how the `ConstructionActions.java` does it, where the Action name does not match ItemID used as the icon.

For the name I went with "[Wood type] vale totem" since I felt like it could be confused with other totems in the game completely unrelated to the minigame. 

Also, I just lumped the xp rates for building and decorating the totems into 1 enum. I figured this was best since the gameplay loop requires playesr to build totems and then decorate them and there is not really a reason to not follow that loop. The values I got are found on the [wiki](https://oldschool.runescape.wiki/w/Vale_Totems#Creation). 

And yes @Nightfirecat before you ask, I promise I spent time playing the minigame, before adding this stuff to Skill calc, unlike last time with mixology 😄. XP rates are very competitive and it is a fun alternative to the traditional fletching skilling method.

Nightfirecat:
I've fixed the xp values for the entries here (vale totem entries were mostly off by decimals, the wiki had updated values; the atlatl dart values didn't correctly account for the quantities produced)

Issue #4126, opened 2018-07-01
Feature Request: Resource tracker 


SwampyApple:
**Is your feature request related to a problem? Please describe.**
Efficiency and information are the name of the game and are the main draw of clients like Runelite. Many people enjoy farming a lot of different things, from looting bosses, grinding mobs or even things like mining. However, it can be a chore to track exactly what materials and resources you've spent and keeping track of what items you've gained when these activities are spread over multiple bank runs.

**Describe the solution you'd like**
This is why I'd like to suggest a new tab on the Runelite sidebar which is focused on tracking resources during set periods of time. How it would work is that you press a button to track your farming. This would start a timer, tracking your time spent, including a pause and reset button.

While this is enabled, any resource you consume would be listed under the 'Resources used' section. The tracking of the resource use could work with a whitelist, where you specifically enable tracking for certain items. 

Likewise an 'Resources/Loot gained' section would track all of the drops you collect during this period of an active resource tracker. A blacklist could filter out certain items you're not interested in to keep the overview tidy. 

Paired with the running timer, this could lead to an updating overview of usage/gain per hour or even averages per kill. For example the resource use section: Saradomin Brews - 25 (-14/hr) (-2/kill). Meanwhile the resource gain section could list things like: Zulrah Scales + 1450 (1200/hr) (160/kill).

Hopefully this explanation will properly show the basic concept I envisioned, but I'm happy to elaborate if certain aspects aren't clear. I do realize that the resources/kill part might be a little bit difficult, so potentially the tracking has to be locked to a certain boss for that to work.

deathbeam:
Updated issue to contain only relevant stuff, rest of requests was already implemented or is being worked on by Jagex (collection log) or was rejected

outpoints:
Is there any news on this? :(

BlankPixels:
Did this feature get added for herb runs? I don't see any herb patches on my loot tracker

PR #19214, opened 2025-07-31
Loot tracker plugin - track herb runs

jfsullivan1:
## Overview
This patch will add loot tracking for herb runs (for farming). This could be expanded later into all farming gains, but we should probably have this different section for herb runs and this is a good starting point for all farming gains. This addition should help address https://github.com/runelite/runelite/issues/4126


BlankPixels:
DId this get added?

jfsullivan1:
> DId this get added?

It did not.

BlankPixels:
was there a reason? i've been hoping this feature would be added for a long time

PR #19231, opened 2025-08-05
ItemMapping: Add Avernic treads and Eye of ayak

jakebrehm:
Added Avernic treads and Eye of ayak to ItemMapping so that their values would be reflected in the player's bank value (Bank plugin).

I decided to add Avernic treads under the *Revertible items* section and the Eye of ayak to the *Degradable/charged weaponry/armour* section.

This fix should be accurate for the Eye of ayak, but as explained [here](https://oldschool.runescape.wiki/w/Avernic_treads_(max)), reverting the Avernic treads will also return any boots that were added to it (i.e., the Primordial, Pegasian, and/or Eternal boots). This means that the value will not be completely accurate since the value of any added boots are not included. I'm new to contributing so I'm not sure if there's a pre-existing way to do this yet.


Adam-:
> This fix should be accurate for the Eye of ayak, but as explained [here](https://oldschool.runescape.wiki/w/Avernic_treads_(max)), reverting the Avernic treads will also return any boots that were added to it (i.e., the Primordial, Pegasian, and/or Eternal boots). This means that the value will not be completely accurate since the value of any added boots are not included. I'm new to contributing so I'm not sure if there's a pre-existing way to do this yet.

I don't think there is a way to represent this currently (an untradeable item breaking down into multiple tradable items). What you have is probably fine.

Nightfirecat:
Untradeable items comprised of multiple tradeable items (think ornament kits for godwords or for zenyte jewelry) need multiple entries, one for each tradeable item which makes up the nontradeable variant. eg.

```java
ITEM_AVERNIC_TREADS(ItemID.AVERNIC_TREADS, ItemID.AVERNIC_TREADS_MAX, ItemID.AVERNIC_TREADS_MAGIC, ItemID.AVERNIC_TREADS_MELEE, ItemID.AVERNIC_TREADS_RANGED, ItemID.AVERNIC_TREADS_MELEE_MAGIC, ItemID.AVERNIC_TREADS_MELEE_RANGED, ItemID.AVERNIC_TREADS_RANGED_MAGIC),
ITEM_PRIMORDIAL_BOOTS(ItemID.PRIMORDIAL_BOOTS, ItemID.AVERNIC_TRADS_MAX, ItemID.AVERNIC_TREADS_MELEE, ItemID.AVERNIC_TREADS_MELEE_MAGIC, ItemID.AVERNIC_TREADS_MELEE_RANGED),
// etc
```

I'll add the missing boots entries for this PR, but what's here so far looks good.

Nightfirecat:
Added in c878f6ddce20395d7b324fdc76883994d5ce1aaa

PR #19242, opened 2025-08-07
Fix canFilterPlayer to handle names with spaces.

Yenof:
Added Text.sanitize to the playerName and the getLocalPlayer().getName so that names with spaces (\u00A0) are handled as expected.


Yenof:
Adam did it better.

PR #19254, opened 2025-08-09
item stats: add blighted overload

Macweese:
This PR adds the blighted overloads to the Item Stats plugin, this potion is only available in DMM type worlds, and has different effects compared to overloads available in COX/NMZ.
Ref #19250 , #19047 (deleted)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -265,6 +265,8 @@ public class ItemStatChanges
 
 		// Regular overload (NMZ)
 		add(combo(SUPER_ATTACK_POT, SUPER_STRENGTH_POT, SUPER_DEFENCE_POT, superRangingPot, superMagicPot, heal(HITPOINTS, -50)), ItemID.NZONE1DOSEOVERLOADPOTION, ItemID.NZONE2DOSEOVERLOADPOTION, ItemID.NZONE3DOSEOVERLOADPOTION, ItemID.NZONE4DOSEOVERLOADPOTION);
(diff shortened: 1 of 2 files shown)


Macweese:
I've revisited this PR after receiving notifications, and upon review I found that since the commit was authored, changes have been announced to Blighted Overloads for the upcoming seasonal Deadman mode, which I want to bring to attention.

These were detailed under the section 'Miscellaneous Changes' of this post:  
https://secure.runescape.com/m=news/deadman-annihilation-overview--rewards-blog?oldschool=1

>Blighted Overloads will deal 5 damage to you (down from 25) when sipped. Your Defence level will decay down to 90% (up from 80%) of its original value.

My assumption is that the PR was merged to accommodate the upcoming event, however, I'm unsure if the announced changes will apply to all Blighted Overloads (including permanent Deadman mode) or just those for this specific seasonal event. If it is the latter then there might be distinct item IDs for those, which we aren't covering here.

Additionally, based on the post, I interpreted the behavior of overloads to be a ceiling-based calculation rather than a compounding effect, on defence drain. If that's true, the current implementation doesn't account for it.

Nightfirecat:
I would assume this will affect all blighted overloads since they don't tend to maintain different dmm items across servers. (seasonal/permanent)

Nightfirecat:
Besides adding the initial check for the capped lower bound for defense, I don't think we can really communicate that the repeated overload effect would lead to greater and greater defense drain over time. If you'd like to put in a PR to change the hitpoints drain (in advance of the upcoming seasonal), I'll be happy to merge it once that gets underway. Otherwise, I'm inclined to maybe just leave it be, since I don't think it's often that people will be drinking a blighted overload with their defense level significantly drained, or would care if it's slightly off.

PR #17726, opened 2024-04-13
loot tracker: track agility dispenser

BBeeaarr:
Closes #17715 
Implemented by converting the chat message into a loot event

Tested on the following scenarios:
Inventory completely full no looting bag --results in no extra consumable
Inventory partially full no looting bag --results in no extra consumable
Inventory not full no looting bag --results in an extra consumable

Inventory completely full open looting bag not full --results in no extra consumable 
Inventory partially full open looting bag not full --results in an extra consumable not placed in looting bag
Inventory not full open looting bag not full --results in an extra consumable not placed in looting bag

Inventory full open looting bag full --results in no extra consumable 
Inventory partially full open looting bag full --results in no extra consumable
Inventory not full open looting bag full --results in an extra consumable

PR #19274, opened 2025-08-13
Loot tracker: track wilderness agility dispenser

samszotkowski:
image

I thought it might be nice to record loot for the [Agility dispenser](https://oldschool.runescape.wiki/w/Agility_dispenser). It's a pretty popular activity, and seeing the total value you've collected so far can be useful for risk management, e.g. "I'll leave as soon as I've made 1 mil."

Everyone brings an open looting bag here, so we can't track it the nice way, but luckily there's a chat message that always says exactly what loot you got. Parsing the message and mapping the words to item ids is ugly, so I will not be at all surprised if there's a more elegant solution than what I have here.

Every time the dispenser is tagged, people get one ticket + two noted items + one unnoted item (if they have inventory space; if not, it doesn't drop to the ground, there's just no item) + sometimes also a medium clue. After removing color tags, the messages look like:

- `You have been awarded 3 x Blighted manta ray and 1 x Steel platebody, and an extra Blighted manta ray from the Agility dispenser.`
- `You have been awarded 3 x Blighted manta ray and 1 x Steel platebody from the Agility dispenser.`
- `You have been awarded a clue scroll, 3 x Blighted manta ray and 1 x Steel platebody, and an extra Blighted manta ray from the Agility dispenser.`
- `You have been awarded a clue scroll, 3 x Blighted manta ray and 1 x Steel platebody from the Agility dispenser.`

https://regex101.com/r/B56zoo/2


samszotkowski:
Duplicate of #17726 and #17777 so plenty to choose from!

samszotkowski:
I tidied up the code by using ItemManager.search() instead of writing out the map of item names -> ids. This way is much more expensive but idk that it matters considering the code will only run once every minute or so. If that's a problem just ditch the latest commit.

PR #19302, opened 2025-08-22
fix: Citric cellar clue location

KeiranY:
Was set to floor 0 when it should be 1. Related to https://github.com/KeiranY/clue-pathing-runelite-plugin/issues/14.


Nightfirecat:
The clues with coordinates set on plane 0 are done so intentionally because of a limitation in the client's hint arrow, see https://github.com/runelite/runelite/pull/16618

KeiranY:
Hmm okay, that makes sense for the hint arrows.
Just out of interest @Nightfirecat is there a list of the clue steps that have had this change made to them? I can see some on #16618 but that doesn't actually cover the steps I've got above. No worries if not, cheers for the detail

Nightfirecat:
I don't, it's fairly common for clue cleanups like this to happen across multiple commits or PRs since there's so many and it's easy to miss, or not know about, all the steps which would be affected.

Nightfirecat:
(fwiw if the ardy mill step has the wrong plane we absolutely can and should fix that; it's just the plane 0 steps which we'd like to leave as-is for now for the above-mentioned reason)

PR #19321, opened 2025-08-27
disable comma when renaming bank tags

### LINKED ISSUE #19297: Don't allow commas in tag tab names
### Discussed in https://github.com/runelite/runelite/discussions/19296

<sup>Originally posted by **aco8ogren** August 21, 2025</sup>
tl;dr

There should be a check for 'does this tag name contain commas' whenever someone renames a bank tag

full message

I just renamed some tags to include commas in the name (e.g. 'woodcutting, fletching, forestry, construction' all as one bank tag, since I was keeping them together). I should've thought it through more - in retrospect I'm not surprised that this broke those two bank tags (which I now sadly have to re-build). I think there are probably a lot of people out there who would make the same mistake, and it should be very easy to add a simple check for 'does this string contain commas or other sabotaging characters' whenever someone uses the 'rename' function. Let me know what you think & thank you for the amazing plug-in! (:</div>

hawolt:
fixes #19297 by adding a comma to the filtered characters which fully prevents entering a comma in the chat dialog that pops up

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/tabs/TabInterface.java
@@ -97,7 +97,7 @@ import net.runelite.client.util.Text;
 @Slf4j
 public class TabInterface
 {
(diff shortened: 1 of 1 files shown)


raiyni:
Needs removed from existing tabs too

PR #19336, opened 2025-09-02
Pixel-perfect ui scaler

kzwicker:
Adds new UI scaling mode "Pixel" which is almost as sharp as nearest-neighbor but does not suffer from aliasing at non-integer scaling values by using bilinear sampling at the edges of texels. Method from: https://colececil.dev/blog/2017/scaling-pixel-art-without-destroying-it/

Compare @ 1642x1080 (1080p black bars)

Nearest

Bilinear

Catmull-Rom

xBR

Pixel

Difference is most noticeable with text

@ ~1440p

Nearest

Pixel

Parity with https://github.com/117HD/RLHD/pull/706

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -1439,8 +1439,8 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 		// Set the sampling function used when stretching the UI.
 		// This is probably better done with sampler objects instead of texture parameters, but this is easier and likely more portable.
 		// See https://www.khronos.org/opengl/wiki/Sampler_Object for details.
-		// GL_NEAREST makes sampling for bicubic/xBR simpler, so it should be used whenever linear isn't
-		final int function = uiScalingMode == UIScalingMode.LINEAR ? GL43C.GL_LINEAR : GL43C.GL_NEAREST;
+		// GL_NEAREST makes sampling for bicubic/xBR simpler, so it should be used whenever linear/pixel isn't
+		final int function = uiScalingMode == UIScalingMode.LINEAR  || uiScalingMode == UIScalingMode.PIXEL ? GL43C.GL_LINEAR : GL43C.GL_NEAREST;
 		GL43C.glTexParameteri(GL43C.GL_TEXTURE_2D, GL43C.GL_TEXTURE_MIN_FILTER, function);
 		GL43C.glTexParameteri(GL43C.GL_TEXTURE_2D, GL43C.GL_TEXTURE_MAG_FILTER, function);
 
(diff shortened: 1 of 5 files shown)


zyussef:
How would I be able to get this? I want that sharpness of nearest neighbor on classic fixed mode but the game is too small for me so I want to do about a 1.5x scaling to around 1147x765 ideally. I can do that now but the only UI Scaling that looks decent is Bicubic (Catmull-Rom)

kzwicker:
> How would I be able to get this? I want that sharpness of nearest neighbor on classic fixed mode but the game is too small for me so I want to do about a 1.5x scaling to around 1147x765 ideally. I can do that now but the only UI Scaling that looks decent is Bicubic (Catmull-Rom)

If you can't wait for this to be merged, you can build runelite from my [fork](https://github.com/kzwicker/runelite/tree/pixel)

ejaz-karim:
I'm looking forward for this to be merged. One small thing i've found when reading the article you linked: Csantosbh has a second article https://csantosbh.wordpress.com/2014/02/05/automatically-detecting-the-texture-filter-threshold-for-pixelated-magnifications/ This second implementation looks good if there is non-uniformed scaling e.g. if you disable "keep aspect ratio" in the stretched mode plugin. This looks overkill for most cases, and may not always be better. However, I think you should at least take a look at this. You could even implement both solutions by having both "Pixel (Fixed)" and "Pixel (Adaptive)"

kzwicker:
> I'm looking forward for this to be merged. One small thing i've found when reading the article you linked: Csantosbh has a second article https://csantosbh.wordpress.com/2014/02/05/automatically-detecting-the-texture-filter-threshold-for-pixelated-magnifications/ This second implementation looks good if there is non-uniformed scaling e.g. if you disable "keep aspect ratio" in the stretched mode plugin. This looks overkill for most cases, and may not always be better. However, I think you should at least take a look at this. You could even implement both solutions by having both "Pixel (Fixed)" and "Pixel (Adaptive)"

My current solution already takes non-uniform scaling into account as pixelsPerTexel is a vector. However, an interesting nugget from this article is a parameter k that tunes the tradeoff between sharpness and anti-aliasing, which might be worth looking into.

Dylan-Mulder:
I hope this gets merged! Running a non-integer scale myself, but I do love the pixel look.

zyussef:
Any idea of what the timeline is for merging additions to plugins and other improvements? No rush just want a better idea of when something like this could be implemented

Adam-:
If the version of this which went into 117hd https://github.com/117HD/RLHD/pull/706 ends up working out okay, I will port it to GPU (experimental). This version of the GPU plugin is getting replaced later this month and so making further changes to it now isn't so useful since it has to be ported regardless.

Adam-:
this was added in https://github.com/runelite/runelite/commit/b6a35c8832c6b8d8c96e890e48b688ef343263d1

PR #19351, opened 2025-09-08
mining: allow configuration of rock respawn timers

xq-r:
I think there is value in exposing these configurations, e.g. when using a Motherlode Mine plugin that shows its own respawn timers.

{14C7F4BC-1825-4A4D-88C1-8C6C8DB21B25}


xq-r:
Hi, are there any problems with this pull request that I can address - or is this just a feature that RuneLite doesn't want added?

PR #19352, opened 2025-09-08
Update CoordinateClue.java

sunka184:
Coordinate clue Small Island reads (CLP) Should read (BLP)


cdfisher:
The small island fairy ring is CLP. BLP goes to Mor Ul Rek.

See here: https://oldschool.runescape.wiki/w/Fairy_rings#%22C%22_combinations

PR #19357, opened 2025-09-10
Puzzle Solver plugin - Add options to increase number of moves as dots shown for puzzle solutions

pwatts6060:
- Add option to increase number of dots shown up to 8
- Interpolate dot colors between a start and end color for easier visualization
- Outline dots with a black outline. 
- Add more size variation for dots.
- The color gradient, increased size variation, and outline make overlapping dots with > 4 moves shown much easier to comprehend.

For the image below:
1 - new default
2 - 6 moves
3 - 6 moves with identical end dot color
4 - 8 moves
image

https://github.com/user-attachments/assets/590f79bb-2b7a-499d-a815-1a9f4cb00f40

This improvement has a lot of backing from clue solvers. From my experimenting this improves puzzle solving time by 50%, a 5 second time save per puzzle.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverConfig.java
@@ -84,8 +84,8 @@ public interface PuzzleSolverConfig extends Config
 
 	@ConfigItem(
 		keyName = "dotColor",
-		name = "Dot color",
-		description = "Dot color for the dots.",
+		name = "Dot start color",
+		description = "Dot color for the first solution dot.",
 		position = 4
 	)
 	default Color dotColor()
@@ -94,12 +94,12 @@ public interface PuzzleSolverConfig extends Config
 	}
 
 	@ConfigItem(
-		keyName = "gradientColor",
-		name = "Secondary gradient color",
-		description = "The color that the dots blend to from the Dot Color.",
-		position = 6
+		keyName = "dotEndColor",
+		name = "Dot end color",
+		description = "Dot color of the last solution dot that dots blend towards.",
+		position = 5
 	)
-	default Color gradientColor()
+	default Color dotEndColor()
 	{
 		return Color.RED;
 	}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
@@ -249,7 +249,7 @@ public class PuzzleSolverOverlay extends Overlay
 									int y = puzzleBoxLocation.getY() + blankY * PUZZLE_TILE_SIZE
 											+ PUZZLE_TILE_SIZE / 2 - markerSize / 2;
 
-									Color color = ColorUtil.colorLerp(config.dotColor(), config.gradientColor(),
+									Color color = ColorUtil.colorLerp(config.dotColor(), config.dotEndColor(),
 										(double) (i - 1) / (movesToShow - 1));
 									graphics.setColor(color);
 									graphics.fillOval(x, y, markerSize, markerSize);


runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -257,7 +292,7 @@ else if (config.displayRemainingMoves())
 								int lastBlankX = currentMove.getEmptyPiece() % DIMENSION;
 								int lastBlankY = currentMove.getEmptyPiece() / DIMENSION;
 
-								// Display the next 3 steps
+								// Display the next movesToShow steps
```
alexanderhenne: movesToShow not used in the loop (probably just restore the comment since you saw that >4 arrows doesn't work)

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -235,10 +263,17 @@ else if (config.displayRemainingMoves())
...
+										Color colour = HSBGradient.getHSBStep(
+											config.dotColour(), config.gradientColour(), i - 1, moves);
+										graphics.setColor(colour);
+									}
+
 									int blankX = futureMove.getEmptyPiece() % DIMENSION;
 									int blankY = futureMove.getEmptyPiece() / DIMENSION;
 
-									int markerSize = DOT_MARKER_SIZE - i * 3;
+									int markerSize = DOT_MARKER_SIZE - i * 3 * 4 / moves;
```
alexanderhenne: Seems arbitrary to me for the circle to be larger if you have few moves left? May want to have a set size for the 1st circle and circle #`movesToShow` being the smallest

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -216,10 +218,36 @@ else if (config.displayRemainingMoves())
...
+
+										if (futureMove == null || !config.drawOverlaps() && positions.contains(futureMove.getEmptyPiece()))
+										{
+											moves = i - 1;
+											break;
+										}
+										positions.add(futureMove.getEmptyPiece());
+									}
+								}
+								moves = Math.min(4, moves); // Scale the dot sizing/color interpolation by at least 4
```
alexanderhenne: I assume this moves variable is so circle #0 is `dotColor` and the #movesToShow circle is definitely `gradientColour`? Regardless if you have many steps left or only two. Correct me if I'm wrong.

Well, this adds extra complexity. Would it be ok to always interpolate between dotColour and gradientColour in `movesToShow` steps instead? That way you could take the original for loop and just do the positions.contains early termination without having to calculate the no. of moves ahead of time.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverConfig.java
```diff
@@ -55,10 +59,70 @@ default boolean displayRemainingMoves()
...
+		max = 8
+	)
+	default int movesToShow()
+	{
+		return 4;
+	}
+
+	@ConfigItem(
+		keyName = "dotColour",
+		name = "Dot colour",
```
alexanderhenne: We use American English in the UI and code elsewhere (colour -> color)

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverConfig.java
```diff
@@ -55,10 +59,70 @@ default boolean displayRemainingMoves()
 	@ConfigItem(
 		keyName = "drawDots",
 		name = "Draw dots instead of arrows",
-		description = "Draw dots increasing in size instead of arrows for the solution."
+		description = "Draw dots increasing in size instead of arrows for the solution.",
+		position = 2
 	)
 	default boolean drawDots()
 	{
 		return false;
```
alexanderhenne: As discussed in the Discord at the least dots should probably be the default for sake of UX. Might be enough to change to true here. I don't know how defaults work in the config system - if this will change to true on next plugin start for existing users who have not manually flipped it. Or if it stays false for them. That's fine imo either way. If someone really liked arrows they can switch it off

alexanderhenne (changes requested):
I think this will be really nice for the user experience when solving clues. Good initiative! I do have some comments about the implementation.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: Could I make a static array with precise preset values to grab from with the relevant index?

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: This is what I had initially, the only benefit of the way I changed it to here was a bigger change in color gradient if there are fewer dots. e.g. 8 dots is a smooth change, but if you disable overlap which most users will, it will occasionally only show say 4 dots and it will do half the color line from black to grey instead of black to white e.g. if you set black and white as the start and end. I don't care much for the colors themselves it was more about being able to differentiate easier. That said it is very minor and happy to just let it pretend there is always movesToShow amount for the interpolation and that may have benefits of consistency in appearance as well.

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
alexanderhenne: I was thinking something like this perhaps: `int markerSize = DOT_MARKER_MAX_SIZE - i * (DOT_MARKER_MAX_SIZE - DOT_MARKER_MIN_SIZE) / ((float)config.movesToShow() - 1)`

```
private static final int DOT_MARKER_MAX_SIZE = 16;
private static final int DOT_MARKER_MIN_SIZE = 2;
```
2 might be too small for the smallest circle so change it to your liking.

This way the first one is always 16, the last one is always 2, and the in-betweens are linearly interpolated between 16 and 2 based on the configurated movesToShow

raiyni:
I think for this context, our color.lerp is probably fine

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: I linearly interpolated between a max and min dot size across the movesToShow

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -216,10 +220,12 @@ else if (config.displayRemainingMoves())
 						{
 							if (config.drawDots())
 							{
-								graphics.setColor(Color.YELLOW);
-
-								// Display the next 4 steps
-								for (int i = 1; i < 5; i++)
+								int movesToShow = config.movesToShow();
+								// positions capacity chosen so set does not reallocate when added to
+								Set<Integer> positions = new HashSet<>(4 * movesToShow / 3);
```
alexanderhenne: I think you need a Math.ceil here or the set will resize when adding the last item. See https://github.com/openjdk/jdk/blob/3eb3e0dcb0aa06fe427adeeaa40b9568c7f07ee8/src/java.base/share/classes/java/util/HashMap.java#L2565

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -230,15 +236,23 @@ else if (config.displayRemainingMoves())
...
+											(double) (i - 1) / (movesToShow - 1));
+										graphics.setColor(color);
+									}
+
 									int blankX = futureMove.getEmptyPiece() % DIMENSION;
 									int blankY = futureMove.getEmptyPiece() / DIMENSION;
 
-									int markerSize = DOT_MARKER_SIZE - i * 3;
+									int numerator = (i - 1) * (DOT_MARKER_MAX_SIZE - DOT_MARKER_MIN_SIZE);
+									int markerSize = (int) Math.round(DOT_MARKER_MAX_SIZE - numerator / (double) movesToShow);
```
alexanderhenne: `(double) (movesToShow - 1)` I think?

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: Ah I looked online and seems you do 4/3 * size + 1

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -69,7 +72,8 @@ public class PuzzleSolverOverlay extends Overlay
 	private static final int INFO_BOX_BOTTOM_BORDER = 2;
 
 	private static final int PUZZLE_TILE_SIZE = 39;
-	private static final int DOT_MARKER_SIZE = 16;
+	private static final int DOT_MARKER_MAX_SIZE = 16;
+	private static final int DOT_MARKER_MIN_SIZE = 2;
```
alexanderhenne:
```suggestion
	private static final int DOT_MARKER_MAX_SIZE = 24;
	private static final int DOT_MARKER_MIN_SIZE = 4;
```

The last circle is a bit too small with 2 imo. Here's with 4 and also a larger first circle so the size difference between circles is more obvious:

image

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverConfig.java
```diff
@@ -55,9 +59,69 @@ default boolean displayRemainingMoves()
...
+
+	@ConfigItem(
+		keyName = "useDotGradient",
+		name = "Use dot gradient",
+		description = "Whether to have a color gradient for the dots between Dot Color and Secondary gradient Color",
+		position = 5
+	)
+	default boolean useDotGradient()
+	{
+		return false;
```
alexanderhenne: This could probably be enabled by default to be honest. Makes the order even more obvious by not relying solely on the circle size

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: Yeh that looks fine

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
alexanderhenne: > 4/3 * size + 1

I think this is not necessarily correct as I understand it. For cases where 4/3 * size is an integer, so size = 6 in this case, your set will have a capacity of 9. It only needs a capacity of 8 to avoid the resizing. So it's better to use Math.ceil: `new HashSet<>(Math.ceil(4 * movesToShow / 3));`

Regardless all this may be unnecessary premature optimization as the default capacity if you don't supply it in the constructor is only 16 items. Unless you've seen that giving it a smaller capacity has a measureable effect on performance.

pwatts6060:
> Nice. You may want to update the PR description with screenshots of the current circle design

Yeah just updated the media at the top of PR

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -230,23 +234,42 @@ else if (config.displayRemainingMoves())
 
 									PuzzleState futureMove = solver.getStep(j);
 
-									if (futureMove == null)
+									if (futureMove == null || !config.drawOverlaps() && positions.contains(futureMove.getEmptyPiece()))
 									{
 										break;
 									}
 
 									int blankX = futureMove.getEmptyPiece() % DIMENSION;
```
alexanderhenne: I think this is better so the set isn't populated if we are drawing overlaps

```suggestion
									if (futureMove == null)
									{
										break;
									}
									
									if (!config.drawOverlaps())
									{
										if (positions.contains(futureMove.getEmptyPiece()))
										{
											break;
										}
										
										positions.add(futureMove.getEmptyPiece());
									}

									int blankX = futureMove.getEmptyPiece() % DIMENSION;
```

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -230,23 +234,42 @@ else if (config.displayRemainingMoves())
...
 
 									int y = puzzleBoxLocation.getY() + blankY * PUZZLE_TILE_SIZE
 											+ PUZZLE_TILE_SIZE / 2 - markerSize / 2;
 
+									Color color;
+									if (config.useDotGradient())
+									{
+										color = ColorUtil.colorLerp(config.dotColor(), config.gradientColor(),
+											(double) (i - 1) / (movesToShow - 1));
+										graphics.setColor(color);
```
alexanderhenne: You already have `graphics.setColor(color);` later

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
```diff
@@ -230,23 +234,42 @@ else if (config.displayRemainingMoves())
...
+									{
+										color = config.dotColor();
+									}
+									graphics.setColor(color);
 									graphics.fillOval(x, y, markerSize, markerSize);
+
+									graphics.setColor(Color.BLACK);
+									graphics.drawOval(x - 1, y - 1, markerSize + 1, markerSize + 1);
+
+									positions.add(futureMove.getEmptyPiece());
```
alexanderhenne: This line can be removed (or moved up really) if you go with the suggestion above

alexanderhenne:
@pwatts6060 Some improvements I saw now when taking a better look at the code

runelite-client/src/main/java/net/runelite/client/plugins/puzzlesolver/PuzzleSolverOverlay.java
pwatts6060: I removed the setColor inside the if statement that wasn't needed. I think everything now is required.

pwatts6060:
@alexanderhenne I cleaned up the config and updated the PR media and description for the current state of the plugin.

Nightfirecat (approved):

PR #19363, opened 2025-09-14
config: Combo box on new line for long config names

Skretzo:
Puts the combo box on a new line if the config name is over 70% of the config panel width.
With 70% as the width threshold this only affects 3 existing config entries in core RuneLite:
|Plugin|Before|After|
|-|-|-|
|Player Indicators|before1|after1|
|XP Tracker|before2|after2|
|XP Tracker|before3|after3|

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/ConfigPanel.java
@@ -33,6 +33,7 @@ import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Dimension;
+import java.awt.FontMetrics;
 import java.awt.event.FocusAdapter;
 import java.awt.event.FocusEvent;
 import java.awt.event.ItemEvent;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I'm not sold one way or the other on whether I think this is worth including, but I'll say that while the player indicators config panel looks noticeably better as a result of this change, XP tracker looks much worse because it's the only one in a long list of similar options which is affected:

image image

raiyni:
`On-screen tracker display mode` can probably be just rewritten or just have words lopped off

Skretzo:
> I'm not sold one way or the other on whether I think this is worth including, but I'll say that while the player indicators config panel looks noticeably better as a result of this change, XP tracker looks much worse because it's the only one in a long list of similar options which is affected:

@Nightfirecat Am I understanding you correctly that XP tracker looks much worse because it is inconsistent that only one of many long configs get the new line treatment?

I would much rather let the user themselves add the new line. Would a feature be accepted where `\n` or `<br>` in the config name is transformed into putting the combo box (or whatever other config item) on a new line?

PR #19372, opened 2025-09-16
Loot tracker: Track updated wilderness loot keys

### LINKED ISSUE #19295: PVP kills not logging loot in Loot Tracker/Logger
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Since the update on August 20, 2025, loot from player kills is no longer being logged in either the Loot Tracker or the Loot Logger.

This issue might be related to a change in the loot key interface that was part of the same update.
Previously, the trackers would log two separate events for a single player kill:

The player's name and the bones dropped on their death.

The items received after opening the loot key.

Currently, neither of these events is being logged, and the loot from player kills does not appear in the tracker at all.
Note: Loot from NPC kills is still being logged correctly.

### Screenshots or videos

_No response_

### RuneLite version

1.11.14
2.5.0

### Logs

_No response_

Sacca-1:
Fixes the support for loot keys which has been broken since Jagex adopted the deadman-style wilderness loot keys in the main game.

- Hooked the loot tracker into the updated Wilderness loot key interface so it scans the five Deadman loot containers, records each tab separately (with a legacy fallback).
- Updated the chest-reset logic to clear only after all key containers (old and new) are empty.

Fixes #19295

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -141,6 +141,21 @@ public class LootTrackerPlugin extends Plugin
 	private static final int THEATRE_OF_BLOOD_REGION = 12867;
 	private static final int THEATRE_OF_BLOOD_LOBBY = 14642;
 	private static final int BA_LOBBY_REGION = 10039;
+	// PvP loot keys (Wilderness/Deadman) reuse the Deadman loot containers, one per tab
+	private static final List<Integer> PVP_LOOT_KEY_CONTAINERS = List.of(
+		InventoryID.DEADMAN_LOOT_INV0,
+		InventoryID.DEADMAN_LOOT_INV1,
+		InventoryID.DEADMAN_LOOT_INV2,
+		InventoryID.DEADMAN_LOOT_INV3,
+		InventoryID.DEADMAN_LOOT_INV4
+	);
+	private static final List<Integer> PVP_LOOT_KEYS = List.of(
+		ItemID.WILDY_LOOT_KEY0,
+		ItemID.WILDY_LOOT_KEY1,
+		ItemID.WILDY_LOOT_KEY2,
+		ItemID.WILDY_LOOT_KEY3,
+		ItemID.WILDY_LOOT_KEY4
+	);
 
 	// Herbiboar loot handling
 	@VisibleForTesting
@@ -351,6 +366,7 @@ public class LootTrackerPlugin extends Plugin
 	private NavigationButton navButton;
 
 	private boolean chestLooted;
+	private boolean pvpKeysLooted;
 	private boolean lastLoadingIntoInstance;
 	private String lastPickpocketTarget;
 	private int ignorePickpocketLoot;
@@ -815,14 +831,12 @@ public class LootTrackerPlugin extends Plugin
 				container = client.getItemContainer(InventoryID.MACRO_CERTER);
 				break;
 			case InterfaceID.WILDY_LOOT_CHEST:
-				if (chestLooted)
+				if (pvpKeysLooted)
 				{
 					return;
 				}
-				event = "Loot Chest";
-				container = client.getItemContainer(InventoryID.LOOT_INV_ACCESS);
-				chestLooted = true;
-				break;
+				pvpKeysLooted = addPvpChestLoot();
+				return;
 			case InterfaceID.PMOON_REWARD:
 				event = "Lunar Chest";
 				container = client.getItemContainer(InventoryID.PMOON_REWARDINV);
@@ -845,11 +859,7 @@ public class LootTrackerPlugin extends Plugin
 			return;
 		}
 
-		// Convert container items to array of ItemStack
-		final Collection<ItemStack> items = Arrays.stream(container.getItems())
-			.filter(item -> item.getId() > -1)
-			.map(item -> new ItemStack(item.getId(), item.getQuantity()))
-			.collect(Collectors.toList());
+		final Collection<ItemStack> items = toItemStacks(container);
 
 		if (config.showRaidsLootValue() && (event.equals(THEATRE_OF_BLOOD) || event.equals(CHAMBERS_OF_XERIC) || event.equals(TOMBS_OF_AMASCUT)))
 		{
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Given that you can have a wilderness loot chest and a theatre of blood chest both available for looting at once, it might be good to split the checks for them into separate fields.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -1102,8 +1219,9 @@ private void countChangedItems(int itemId, Object metadata)
 	public void onItemContainerChanged(ItemContainerChanged event)
 	{
 		// when the wilderness chest empties, clear chest loot flag for the next key
-		if (event.getContainerId() == InventoryID.LOOT_INV_ACCESS
-			&& Arrays.stream(event.getItemContainer().getItems()).noneMatch(i -> i.getId() > -1))
+		if (isWildernessLootChestContainer(event.getContainerId())
+			&& Arrays.stream(event.getItemContainer().getItems()).noneMatch(i -> i.getId() > -1)
```
Adam-: this is covered by `areWildernessLootKeyContainersEmpty()`

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +894,109 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+			if (containerHasItems(client.getItemContainer(id)))
+			{
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	private boolean containerHasItems(@Nullable ItemContainer container)
```
Adam-: this can be replaced with `count() > 0`

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +894,109 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+		}
+
+		return Arrays.stream(container.getItems())
+			.filter(item -> item.getId() > -1)
+			.map(item -> new ItemStack(item.getId(), item.getQuantity()))
+			.collect(Collectors.toList());
+	}
+
+	@Nullable
+	private String getWildernessLootKeyValue(int tabIndex)
```
Adam-: This is completely useless? why is this being passed as metadata?

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +894,109 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+			{
+				addLoot("Loot Chest", -1, LootRecordType.EVENT, null, legacyItems);
+				return true;
+			}
+		}
+
+		return recordedLoot;
+	}
+
+	private Collection<ItemStack> toItemStacks(@Nullable ItemContainer container)
```
Adam-: Make the few other places which have this container stream/map/collect code then call this instead.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +894,109 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+			}
+
+			String value = getWildernessLootKeyValue(index);
+			addLoot("Loot Chest", -1, LootRecordType.EVENT, Strings.isNullOrEmpty(value) ? null : value, items);
+			recordedLoot = true;
+		}
+
+		if (!recordedLoot)
+		{
+			Collection<ItemStack> legacyItems = toItemStacks(client.getItemContainer(InventoryID.LOOT_INV_ACCESS));
```
Adam-: Is this old inventory still used?

Adam- (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
Sacca-1: This was a remnant of a different approach - now deleted

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
Sacca-1: It is not, I left it initially in case of future changes, but that is indeed not needed.

Sacca-1:
I think that should cover the requested changes, please let me know if there's anything else required.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -141,6 +141,14 @@ public class LootTrackerPlugin extends Plugin
 	private static final int THEATRE_OF_BLOOD_REGION = 12867;
 	private static final int THEATRE_OF_BLOOD_LOBBY = 14642;
 	private static final int BA_LOBBY_REGION = 10039;
+	// Wilderness loot key interface reuses the Deadman loot containers, one per tab
+	private static final int[] WILDERNESS_LOOT_KEY_CONTAINERS = {
```
Nightfirecat: I'm inclined to rename this `PVP_LOOT_KEY_CONTAINERS` since there's theoretically no reason the loot tracker would not also work for tracking Deadman loot the same way.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +887,69 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+			return Collections.emptyList();
+		}
+
+		return Arrays.stream(container.getItems())
+			.filter(item -> item.getId() > -1)
+			.map(item -> new ItemStack(item.getId(), item.getQuantity()))
+			.collect(Collectors.toList());
+	}
+
+	private boolean isWildernessLootChestContainer(int containerId)
```
Nightfirecat: If you define `private static final List<Integer> WILDERNESS_LOOT_KEY_CONTAINERS = List.of(...)` rather than an array, you can replace this method and its uses with `WILDERNESS_LOOT_KEY_CONTAINERS.contains(containerId)`.

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +887,56 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+
+		for (int containerId : PVP_LOOT_KEY_CONTAINERS)
+		{
+			final Collection<ItemStack> items = toItemStacks(client.getItemContainer(containerId));
+			if (items.isEmpty())
+			{
+				continue;
+			}
+
+			addLoot("Loot Chest", -1, LootRecordType.EVENT, null, items);
```
Nightfirecat, line 914: idk if most people would understand "Loot Chest" as a loot source; a quick check with other folks on Discord seems to indicate "Loot key" would be more recognizable (since that's the item turned in to get the loot)
```suggestion
			addLoot("Loot key", -1, LootRecordType.EVENT, null, items);
```

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -880,6 +887,56 @@ public void onWidgetLoaded(WidgetLoaded widgetLoaded)
...
+
+		for (int containerId : PVP_LOOT_KEY_CONTAINERS)
+		{
+			final Collection<ItemStack> items = toItemStacks(client.getItemContainer(containerId));
+			if (items.isEmpty())
+			{
+				continue;
+			}
+
+			addLoot("Loot key", -1, LootRecordType.EVENT, null, items);
```
Nightfirecat: Never mind, "Loot Chest" is how we recorded pvp loot before. We'll keep the name.
```suggestion
			addLoot("Loot Chest", -1, LootRecordType.EVENT, null, items);
```

onivalidator:
Good lord please merge this. We're all suffering.

Nightfirecat:
FYI the `chestLooted` resetting logic you had did not seem to work, I suspect because the loot item containers don't send an event when the interface is closed. I managed to get it working properly by introducing a separate boolean field for pvp chest looting, and resetting it when all loot keys are removed from the inventory.

PR #19387, opened 2025-09-22
screenshot: prevent notification flashes

JZomDev:
Prevent notification flashing from appearing in screenshots


rsbfox:
Been using this since October 10th. No issues at all taking screenshots. Never seen a flash in one since. LGTM

PR #19394, opened 2025-09-24
cluescrolls: Add pirate bandana eyepatches to Fight Arena emote clue

coopermor:


Nightfirecat:
Can we use ItemVariationMapping for the pirate bandana requirement?

PR #19400, opened 2025-09-25
Update CoordinateClue.java to fix master clue step

Drewsername:
Fix master clue step which is a brassican mage, incorrectly identified as ancient wizard trio. Didn't think to take a screenshot, sorry.


Nightfirecat:
Added in 7eaaf4831b5acf1a6f9a09d31ed6cec931c993a4

PR #19403, opened 2025-09-27
Added username parsing for HighScorePlugin Lookup from menuTarget

### LINKED ISSUE #19380: HiScore bug with lookup menu
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

_No response_

### Bug description

When I right click on a user and the user leaves the vicinity or local area if I click lookup on the right click menu it doesn't load the highscore for the user in the sidebar of runelite.

If I right click on a user it should work when I click lookup regardless if the user is nearby now or not.

### Screenshots or videos

_No response_

### RuneLite version

Runelite version: 1.11.17
launcher version: 2.7.5

### Logs

_No response_

Cburns1331:
In certain cases event.getMenuEntry().getPlayer() is null(such as player leaving render distance, TP'ing out, or going to a different floor), but the menu target still contains the username; this fallback ensures lookup via the menuOption works in those cases.

First PR, let me know if there are changes needed, thanks!

Fixes #19380


Nightfirecat:
#13330 exists for changing this behavior

PR #19409, opened 2025-09-28
clues: Remove unobtainable essence mine HotColdLocation

TheLope:
The removed HotColdLocation is not obtainable

"North-east of the Arceuus essence mine" is the "real" HotColdLocation for that area

Was confused with the following CoordinateClue
image


Nightfirecat:
I'm pretty sure this is not an unobtainable step; I've both gotten this step myself and seen many other people get it as well.

TheLope:
> I'm pretty sure this is not an unobtainable step; I've both gotten this step myself and seen many other people get it as well.

There is a coordinate clue that is almost the same location that is obtainable, this is specifically for the Hot Cold. I have been discussing with multiple clue hunters with thousands of master clue kc who report never having seen it.

What is the original source for the HotColdLocation?

Issue #19156, opened 2025-07-17
[Interface styles] Condense player options creates multiple submenus when right-clicking a player through an object


Ririshi:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

Right-clicking a player through an object with "Condense player options" enabled splits the first player's options into multiple submenus, where the first menu only has "Walk here". See the screenshot below, the level 95 player has two submenus. The second menu contains the rest of the options for that player. I believe the options should all be in a single menu.

Or perhaps there should be only one separate "Walk here" option that isn't in a submenu, but I don't think that works since the "Walk here" option for different players could potentially target different tiles?

I've attached logs of my regular client but it also happens in safe mode.

### Screenshots or videos

<img width="613" height="294" alt="Image" src="https://github.com/user-attachments/assets/71b64da9-cd91-4930-a58f-b5969d4f8cdf" />

### RuneLite version

RuneLite version: 1.11.11
Launcher version: 2.6.13

### Logs

[launcher.log](https://github.com/user-attachments/files/21283518/launcher.log)
[client.log](https://github.com/user-attachments/files/21283519/client.log)

RB490:
Yes! This has been bothering me for a long time. Would love to see this resolved

trs:
It looks like the `condensePlayerOptions` method assumes each target is listed together, but when right-clicking through an object that is not the case. The object options appear after the player, but before the player's "Walk here" option, creating multiple sub menus.

Brainstorming how to fix this, we'd need to keep track of each "seen" player and "push" to their sub menu rather than the index incrementing today.
Or do some sort of sorting ahead of time? But we'd need to maintain the target ordering.

Adam-:
fixed in e269b52a963c0eb87b38754be0e48c5d04d35cf5

PR #19441, opened 2025-10-14
fix: hide walk here target name with condensed options

trs:
Removes the target from the "Walk here" option when condensing options.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/interfacestyles/InterfaceStylesPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/interfacestyles/InterfaceStylesPlugin.java
@@ -154,7 +154,6 @@ public class InterfaceStylesPlugin extends Plugin
 		Menu submenu = null;
 		Player prev = null;
 		boolean changed = false;
-		boolean walkFound = false;
 
 		for (MenuEntry menuEntry : menuEntries)
 		{
@@ -190,10 +189,9 @@ public class InterfaceStylesPlugin extends Plugin
 					.setDeprioritized(deprioritized);
 				changed = true;
 			}
-			else if (!walkFound && type == MenuAction.WALK)
+			else if (type == MenuAction.WALK)
 			{
 				newMenus[newIdx++] = menuEntry.setTarget("");
-				walkFound = true;
 			}
 			else
 			{


Adam-:
Does doing `type != MenuAction.WALK_HERE` instead work too?

trs:
> Does doing `type != MenuAction.WALK_HERE` instead work too?

Yeah looks like it does, I'll change that

Adam-:
This isn't what I meant. I meant adding it to line 163.

ahmadshoaib2092-cell:
Mfgty

trs:
> This isn't what I meant. I meant adding it to line 163.

How's this?

Adam-:
> > This isn't what I meant. I meant adding it to line 163.
> 
> How's this?

Does it work if you delete all of your changes except for this? 
image

trs:
> > > This isn't what I meant. I meant adding it to line 163.
> > 
> > 
> > How's this?
> 
> Does it work if you delete all of your changes except for this?

No, we need to keep track of the first one so we can add it back to the top of the menu

ririshi:
In #19156 is mentioned the following with regards to "lifting" the walk here option out of the player submenus:
> I don't think that works since the "Walk here" option for different players could potentially target different tiles?

Did you test whether this change impacts "walk here" on multiple players targeting different tiles? Let's say your camera is at a low angle and you can right-click multiple players who aren't on the same tiles due to the angle. Does walk here always go to the same tile in vanilla and the old implementation of the plugin?

Hannah-GBS:
>  Let's say your camera is at a low angle and you can right-click multiple players who aren't on the same tiles due to the angle.

No matter how many players you catch with your right click, there is only ever one "Walk here" option, which takes you to the tile your cursor was over when you right clicked. It doesn't matter what tiles the players are on. Simply moving it out of any submenus should be fine.

Adam-:
#19156 is a bug about this: 
image

Where there is a duplicate player in the menu with just "Walk here". Doing the change I suggested changes it to this:

image

Which I think effectively fixes it. I don't understand why you are trying to reorder walk here over top of the object op. That shouldn't happen and just isn't how the game works really at all, the left click option should be the object op.

trs:
> Where there is a duplicate player in the menu with just "Walk here". Doing the change I suggested changes it to this:

That works as a fix.

> Which I think effectively fixes it. I don't understand why you are trying to reorder walk here over top of the object op. That shouldn't happen and just isn't how the game works really at all, the left click option should be the object op.

Partially was to make walk here easier to find, always being at the top. It doesn't change the left click option, but does make it so the top option isn't the left click option, which breaks convention for sure.

If we don't like the reorder (which I totally get), happy to remove it!

What are your thoughts on removing the walk here target? So it just says "Walk here" rather than "Walk here PlayerName"

Adam-:
> Partially was to make walk here easier to find, always being at the top. It doesn't change the left click option, but does make it so the top option isn't the left click option, which breaks convention for sure.

But this isn't actually the convention. Eg here is the vanilla menu:

image

Notice Walk here isn't on top.

> What are your thoughts on removing the walk here target? So it just says "Walk here" rather than "Walk here PlayerName"

This is fine

raiyni:
You do not need a walk found boolean.

PR #19451, opened 2025-10-17
Add marble wall to construction calculator

charlie-collard:
For Gridmaster, I wanted to calculate how many marble blocks I needed but couldn't do it as it's not in the construction calculator. I'm lazy and don't want to do the maths manually.


Nightfirecat:
Added in 682e787ebc631e7cbddf909285c2eadab9700839, thanks!

PR #19458, opened 2025-10-19
ItemStats: Offset the secondary heal value for hunter meats

Mitchell-Kovacs:
Closes: https://github.com/runelite/runelite/issues/18505 https://github.com/runelite/runelite/pull/18533

Hunter meats will heal an initial amount and then another after 7 ticks.

This patch offsets the secondary heal value by the initial heal amount. Assuming the player takes no damage, this will display the actual HP that the player will heal to. This has the additional benefit of colour coding the entry as expected.

Additionally, missing hunter meats have also been added.

Examples using "Cooked pyre fox" which heals 11 and then 8 on an account with 99 HP with "Show absolute", "Show relative", and "Show theoretical" checked:

| HP | Picture
| --- | --- |
| 99 | image |
|89|image|
|83|image|
|79|image|

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/itemstats/ItemStatChanges.java
@@ -130,8 +130,8 @@ public class ItemStatChanges
 		// Hunter meats
 		add(combo(food(4), offsetFood(4, 4)), ItemID.WILDKEBBIT_COOKED);
 		add(combo(food(6), offsetFood(6, 5)), ItemID.LARUPIA_COOKED);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
I like the result this PR achieves, but I don't much like the way the initial heal amount effectively needs to be duplicated for definitions, ie. it's possible to write something like `combo(food(1), offsetFood(2, 3))`, which makes no sense. I think an API where the heal amounts were both given at once (`offsetFood(14, 12)` for moonlight antelope, for instance) or where the method or class accepted multiple `Food` arguments to be "applied" in sequence would be much better.

skipple:
This really should be implemented, as the current way item-stats displays information regarding hunter food is inaccurate, bordering on bug. I was going to make a PR myself to address this, but it looks like you have already done so with the same approach I was going to take. Hopefully this can be merged.

PR #17522, opened 2024-02-23
Add player option condensing to InterfaceStylesPlugin

Thource:
This changes the bottom left and chatbox right widget snap points so that they are attached to the chatbox.

https://github.com/runelite/runelite/assets/4104309/a09c3219-2e9e-4b74-8eea-4d1fb654c2ca

PR #19466, opened 2025-10-22
Fix the chatbox anchors (snap points) to the chatbox

ASong5:
This fixes CI errors on @Thource\'s feature branch that make chatbox anchors (snap points) adjust correctly with changes to chatbox size and position.

Changes include:
- Replaced deprecated Widgets class constants with GameVal constants in OverlayRenderer
- All changes from #17522


idyet:
@ASong5 / @Thource can we get a revival on this?

PR #19476, opened 2025-10-25
Bank tags enhancement

abristow3:
For #19429

# Bank Tags Scroll Navigation Enhancement

## Summary
Added quick navigation features to bank tag scroll buttons for faster tag list navigation.

### New Features
- **Scroll to top/bottom**: Right-click scroll buttons for instant jump options
- Clicking the tag tab tab also positions scroll buttons to the chosen tab

### Code Changes (`TabInterface.java`)
1. Added 2 new constants for menu options
2. Added 3 new methods:
   - `scrollToTop()` - jump to top
   - `scrollToBottom()` - jump to bottom
   - `scrollToTag()` - scroll to specific tag index

## Demo

https://github.com/user-attachments/assets/eff0cbde-025b-4956-a4a3-ded03cb38bad


aaronenberg:
Why this is such a nice QoL for bank tag tabs for players that have a high count of bank tag tabs:

With my play style, I frequently have to update the auto layout of my bank tag tabs.

The tag tab scroll bar has the "Auto layout" options. However the "View tag tabs" UI does not have this.

At a high count of tag tabs it becomes more efficient to navigate between tabs using the "View tag tabs" UI rather than scrolling the tag tab bar. But this means you still have to scroll to the tab using the tab bar in order to use auto layout whenever the layout needs updating.

Auto-scroll would fix this accessibility because you can efficiently open a tag tab using "View tag tab" UI and then with auto-scroll you have quick access to the same tag tab in the scroll bar for auto layout options.

abristow3:
Good morning, I wanted to see if this could be reviewed again by chance. It has been about a year since I opened the MR and I think this enhancement is still being wanted by the community.

PR #19486, opened 2025-10-28
Fix flaky test in `LootTrackerPluginTest#testGrubbyChest` and `MotherlodePluginTest#testOreCounter`

cperkkk:
# Fix flaky test in `LootTrackerPluginTest#testGrubbyChest` and `MotherlodePluginTest#testOreCounter`

### Problem
Both unit test fails intermittently due to non-deterministic ordering of the items during `verify` and `assertEquals`. The proposed fix ensures the test is robust to non-deterministic behaviors and compatible with different JVM, future updates to it or related dependencies. It also makes it easier if we want to add new items without rearranging the order to make the test pass.

### Solution
The fix involves two changes to ensure consistent item ordering:

1. Use `java.util.hashset` equality which doesn't care about ordering.
2. Use `org.hamcrest.Matchers.containsInAnyOrder` which also is order agnostic.

The fix does not change any code under test.

## Reproduction of Bug

Use the [TestingResearchIllinois's NonDex](https://github.com/TestingResearchIllinois/NonDex) tool and run:

```shell
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex 
```

### Error Message

Before the fix:
image

After the fix:
image
image

## Additional Notes

Let me know if further improvements or clarifications are needed!

### DIFF
--- a/runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
+++ b/runelite-client/src/test/java/net/runelite/client/plugins/loottracker/LootTrackerPluginTest.java
@@ -420,8 +420,9 @@ public class LootTrackerPluginTest
 					new ItemStack(ItemID._2DOSEPOTIONOFSARADOMIN, 3),
 					new ItemStack(ItemID.BR_2DOSE2RESTORE, 1),
 					new ItemStack(ItemID.HOSDUN_ORANGE_EGG_SAC, 1)
-			))))
-			);
+				)))
+			)
+		);
 	}
 
 	@Test


Nightfirecat:
How are you producing that error message? These tests run in CI and do not fail on the master branch, ever: https://github.com/runelite/runelite/actions/workflows/CI.yml

cperkkk:
> How are you producing that error message? These tests run in CI and do not fail on the master branch, ever: https://github.com/runelite/runelite/actions/workflows/CI.yml

Hi there @Nightfirecat, sorry for the late reply, I was very unwell, this is mainly a simple improvement to guarantee that the test will not flake in the future or with different jvm changes, similar to https://github.com/runelite/runelite/pull/14104.

For example: in the MotherlodePlugin, it uses the HashMultiset

image
image


Which according to the documentation, will not guarantee any ordering, so this may return any ordering in different JVM / in future implementations.

image

My fix is **a mere improvement** so that it will not flake anymore, to reproduce the flakiness you can play around and use the [University of Illinois's Nondex tool](https://github.com/TestingResearchIllinois/NonDex).

```
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex 
```

Please let me know if you need further clarification on this. 

Thank you so much for your guidance!

PR #19489, opened 2025-10-29
Added Navigation Button Priority Reorder

Sieder16:
Added Navigation Button Priority to allow reordering of default runelite plugins that draw an icon on the panel

See Video for what this does, It does it to all the plugins that at favourite

(Re pull with different repo name)

https://github.com/user-attachments/assets/a91abf49-7845-49f8-8739-a8e4f468409c


raiyni:
This is better written as user drag and drop icons

tylerwgrass:
related to/duplicate of: #18720

Issue #19500, opened 2025-11-03
[Runecraft plugin] Overlay "Show pouch fills" is incorrect for Colossal pouch


gavin-lb:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Unlike the Medium, Large, and Giant pouches - which randomly lose 0-3 durability per essence stored - the Colossal pouch always loses exactly 1.

Currently, the Runecraft plugin treats the Colossal pouch like the others, and hence underestimates how many times it can be filled. A fully repaired Colossal pouch is shown as having only 5 fills left, when it [should be 8](https://oldschool.runescape.wiki/w/Colossal_pouch).

### Screenshots or videos

![Image](https://github.com/user-attachments/assets/b5850962-26d8-4d12-b056-92dbdd48e497)

### RuneLite version

<img width="159" height="50" alt="Image" src="https://github.com/user-attachments/assets/6dc71b9e-8824-40b4-8b71-84576e5b3274" />

PR #19501, opened 2025-11-03
fix: properly display Colossal pouch fills #19500

gavin-lb:
Fixing issue #19500


Felanbird:
Noting here that during the original implementation for this plugin, the colossal pouch was broken for another reason which was fixed, so it's definitely possible this change is correct.

PR #19504, opened 2025-11-04
Add Bank Watcher plugin to track bank value and item deltas

Khaled2143:
### Overview
This PR introduces the **Bank Watcher** plugin, which tracks changes in the player’s bank between sessions — including item quantity and GE price deltas — and displays top movers, gainers, and losers in a clean sidebar panel.

### Features
- Scans player bank and records value changes between logins
- Displays item deltas and profit/loss overview per item allowing whales to see what values are affecting their bank value

### Testing
- Verified correct delta calculations and bank scanning behavior in live client
- Tested multiple session transitions to confirm persistent tracking accuracy
- Validated that UI updates reflect accurate item changes and bank totals

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/bankwatcher/BankItem.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/bankwatcher/BankItem.java
@@ -1,53 +1,66 @@
 package net.runelite.client.plugins.bankwatcher;
 
-public class BankItem {
+public class BankItem
+{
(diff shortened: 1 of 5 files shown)


Nightfirecat:
Thanks for the PR, but we don't feel this would be a good fit for the core client. I recommend looking into structuring it as a submission for https://github.com/runelite/plugin-hub, where it can be downloaded and installed by users who would like that functionality.

PR #19506, opened 2025-11-04
Re-add Color Banding toggle to GPU plugin

proto-cool:
Color banding is an iconic graphical style of OSRS. I think the new GPU plugin is a step in the right direction but it sucks to have lost this option. I've re-enabled it and tested the changes locally.

Here's what it looks like, just hanging out in Lumby:

With banding _disabled:_
Screenshot 2025-11-04 at 12 58 12 PM

With banding _enabled:_
Screenshot 2025-11-04 at 12 58 46 PM

Should be fairly straightforward and hopefully not have any negative downstream effects on the new renderer (it looked structured somewhat the same as the old plugin in the shader code.)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -254,6 +254,7 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 	private int uniFogDepth;
 	private int uniDrawDistance;
 	private int uniExpandedMapLoadingChunks;
+	private int uniSmoothBanding;
 	private int uniWorldProj;
 	private static int uniEntityProj;
 	static int uniEntityTint;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/resources/net/runelite/client/plugins/gpu/hsl_to_rgb.glsl
```diff
@@ -82,4 +82,14 @@ vec3 hslToRgb(vec3 hsl) {
...
+}
+
+// Reintroduce the packed-HSL path used by the old shader.
+// This preserves banding when you do hslToRgb(int(fHsl)).
+vec3 hslToRgb(int hslPacked) {
+  int var5    = hslPacked / 128;          // top bits: hue/sat
+  int hueIdx  = var5 >> 3;                // 0..63
+  int satIdx  = var5 & 7;                 // 0..7
+  int lum     = hslPacked % 128;          // 0..127
+  return hslToRgb(vec3(float(hueIdx), float(satIdx), float(lum)));
```
Adam-: This should probably just be `vec3 hsl = vec3(abhsl >> 10 & 63, abhsl >> 7 & 7, abhsl & 127);` like how it is in vert.glsl

proto-cool: Should be updated to follow that instead - confirmed that the color banding effect still works as expected

PR #19526, opened 2025-11-07
gpu: fix wrong targetDimensions uniform value

zwyz:
The `targetDimensions` uniform is being set to the same value as `sourceDimensions` when stretched mode is not enabled. This is not correct when DPI scaling is active, and causes the "Hybrid" UI scaling mode which relies on it to effectively become the same as linear scaling.

Zoomed-in before/after picture (at 225% DPI scale):
image

Note how the text below looks crisper. Only target pixels that land on multiple source pixels are interpolated, as intended by the hybrid scaling mode.


Adam-:
This is interesting. The current intended behavior is the ui scaling option only takes effect when stretched mode is enabled. I guess you are using j2d scaling but keeping stretched mode off?

zwyz:
Yes, by changing Windows's "High DPI scaling override" setting for RuneLite to "Application". Otherwise the entire client (including the RuneLite UI and the world) looks slightly blurry and UI scaling mode has not much effect as the system simply applies a linear scale to the entire window.

PR #19529, opened 2025-11-07
rl-client: include linux-arm64 lwjgl natives

pajlada:
Tested on Fedora Asahi Linux together with https://github.com/runelite/rlawt/pull/3


Adam-:
added in ebe56b9a3b817476658981c46d7fad003daaf523

PR #19535, opened 2025-11-09
clues: Fix incorrect z plane for "With drinks and dwarves I make my way"

Skretzo:
The z plane is incorrectly set to 0 instead of 1 for the elite cryptic clue "With drinks and dwarves I make my way, upon my head a glass will stay.".
https://oldschool.runescape.wiki/w/Clue_scroll_(elite)_-_With_drinks_and_dwarves_I_make_my_way

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -1985,7 +1985,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 		CrypticClue.builder()
 			.itemId(ItemID.TRAIL_ELITE_RIDDLE_VM01)
 			.text("With drinks and dwarves I make my way, upon my head a glass will stay.")
-			.location(new WorldPoint(1432, 9584, 0))
+			.location(new WorldPoint(1432, 9584, 1))
 			.npc("Funbo")
 			.solution("Speak to Funbo inside Cam Torum's pub.")
 			.build(),


Nightfirecat:
While this is technically more correct, does this lead to the clue icon not appearing on the world map? That's the reason a handful of other clues have their locations placed on plane 0. In the long term I'd like to create some runelite-client api to replace our use of the vanilla client's hint arrow (both to fix this behavior and avoid clobbering it) but I don't have that ready yet.

Skretzo:
> While this is technically more correct, does this lead to the clue icon not appearing on the world map? That's the reason a handful of other clues have their locations placed on plane 0. In the long term I'd like to create some runelite-client api to replace our use of the vanilla client's hint arrow (both to fix this behavior and avoid clobbering it) but I don't have that ready yet.

I was not aware of that issue. I don't currently have that clue. Is there a way I can fake having the clue to test it?

Felanbird:
```
inject(net.runelite.client.eventbus.EventBus.class).post(new CommandExecuted("clue", new String[]{"With drinks and dwarves I make my way, upon my head a glass will stay."}));
```
in the shell

Skretzo:
> While this is technically more correct, does this lead to the clue icon not appearing on the world map? That's the reason a handful of other clues have their locations placed on plane 0. In the long term I'd like to create some runelite-client api to replace our use of the vanilla client's hint arrow (both to fix this behavior and avoid clobbering it) but I don't have that ready yet.

It does seem to still work with z plane 1.
image

Probably unrelated, but there was no arrow on the "Gielinor Surface" world map, so I had to correctly guess that the map should be "Cam Torum". This was the case for both z plane 0 and 1. Maybe the map selector should be highlighted to better indicate where to go?
image2

Nightfirecat:
We don't really have a good fix for indicating that a sub-map needs to be selected as of yet, but if it shows correctly on the world map with either plane, this is a definite improvement.

PR #19539, opened 2025-11-10
party: Add default avatar so users without Discord appear on the world map

jamyc:
Users without Discord installed previously appeared with an empty avatar.
before

Although this might seem harmless, it caused those users **not to appear on the World Map at all**.

This PR introduces a default user icon to ensure that every party member has an avatar, allowing all users to be properly displayed on the World Map.

after
worldmapfix

---

I wasn't entirely sure about the preferred way or location to load `BufferedImage`s, so please let me know if this should be adjusted.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/party/PartyService.java
+++ b/runelite-client/src/main/java/net/runelite/client/party/PartyService.java
@@ -64,7 +64,7 @@ import static net.runelite.client.util.Text.JAGEX_PRINTABLE_CHAR_MATCHER;
 @Singleton
 public class PartyService
 {
-    private final BufferedImage DEFAULT_USER_ICON = ImageUtil.loadImageResource(PartyPlugin.class, "user_icon.png");
+	private final BufferedImage DEFAULT_USER_ICON = ImageUtil.loadImageResource(PartyPlugin.class, "user_icon.png");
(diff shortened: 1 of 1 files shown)


raiyni:
Does the image get return the same thing for every user here? https://github.com/jamyc/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/discord/DiscordPlugin.java#L241

jamyc:
That method `onDiscordUserInfo` doesn't even trigger for users in the party that do not have Discord installed.

Nightfirecat:
Where did you source this user icon?

jamyc:
> Where did you source this user icon?

Icon is from FA: https://fontawesome.com/icons/user?f=classic&s=solid
Used Figma to convert the SVG to a 32x32 png

Nightfirecat:
That icon would be fine, though it would be nice to use some osrs-themed sprite, preferably one in the cache (such as [sprite 1707](https://abextm.github.io/cache2/#/viewer/sprite/1707)) or an item image. (there are few items resembling a player bust; [the recipe for disaster heads are close](https://oldschool.runescape.wiki/w/Interface_items#Recipe_for_Disaster) but not quite what we'd be looking for)

That aside, I'll take a look at the implementation more closely this week.

PR #19541, opened 2025-11-10
fairyrings: Make fairy ring code search case-insensitive

jamyc:
Previously, the fairy ring code search only worked when entering codes in lowercase (e.g., `aiq`), while input with uppercase (`AIQ`, `Aiq`, etc.) would not return any results.

This PR normalizes the input to lowercase, allowing both lowercase and uppercase codes to be found correctly.

---

## Before:
before1
before2

## After:
after

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
@@ -193,7 +193,7 @@ public class FairyRingPlugin extends Plugin
 				tags = ring.getTags();
 			}
 
-			var filter = client.getVarcStrValue(VarClientID.MESLAYERINPUT);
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/fairyring/FairyRingPlugin.java
```diff
@@ -193,7 +193,7 @@ else if ("fairyringFilterDbrow".equals(ev.getEventName()))
 				tags = ring.getTags();
 			}
 
-			var filter = client.getVarcStrValue(VarClientID.MESLAYERINPUT);
+			var filter = client.getVarcStrValue(VarClientID.MESLAYERINPUT).toLowerCase();
```
Adam-, line 196: While the normal tags are lowercase I don't think that the config tags are, so this seems like that would affect that?

PR #19563, opened 2025-11-16
Fix 2005 interface styles friends chat icon alignment

robmonte:
The alignment of the GIM/friends chat/grouping icons (and friend list in fixed mode) in the menu stones is off slightly in the 2005 interface style. It's more obvious when using 2005 Overhaul resource pack as it properly sizes the icons as well, particularly when the group icon is set to a equally-sized red quest icon override as you can see here. If you look carefully you will notice the red quest icon for groups is still 1 pixel placement above the friends and ignore list icons. This is due to the icon placement _without_ the 2005 resource pack being slightly different, so this is the best compromise. An example is in the next comment.

Comparison gifs below:

fixed comparison

classic resize comparison

modern resize comparison


robmonte:
For thoroughness here is a comparison without the 2005 resource pack. If you move it down 2 pixels instead, the resource pack icons are perfectly aligned but the stock Runelite interface style plugin will be a little too close to the bottom edge.

without resource pack comparison

PR #19575, opened 2025-11-19
clues: add more sailing clues

Felanbird:
master clues are abex's enemy

there is a STASH unit in the game `HH_MASTER_SAIL` https://abextm.github.io/cache2/#/viewer/obj/58132 but I think I always checked stash locations manually, so I don't know how to find them in the cache, so i've only included the clue itself without the STASH


Felanbird:
32f2f00243025b41a0d6535ec3c5514db3e02a3c

PR #19577, opened 2025-11-19
Interfacestyles: fixing bug not updating combat icon[#19511]

### LINKED ISSUE #19511: [Interface Styles] "Combat Options" icon incorrect for all gameframe styles
### Type

Incorrect behavior

### Operating System

Linux

### OS Version

Arch Linux

### Bug description

Interface Styles plugin is enabled with Gameframe set to 2005, after recent game update, the icon used for Combat Options shows the same icon as without the plugin on for all gameframe styles (2005, 2006, and 2010).

### Screenshots or videos

2005 interface:
Image

2006 interface:
Image

2010 interface (slightly off-center as well):
Image

Plugin disabled:
Image

### RuneLite version

RuneLite version: 1.11.26
Launcher version: 2.7.5

### Logs

[client.log](https://github.com/user-attachments/files/23371376/client.log)
[launcher.log](https://github.com/user-attachments/files/23371377/launcher.log)

LarsTwiNky:
Close #19511

Copied combat-icon to SideiconInterface and switched SpriteOverride from SideIcon to SideiconInterface


LarsTwiNky:
The Icons arent used consistently. Some (e.g. inventory) are only SideIcons, others are SideiconsInterface and/or SidiconsNew. Now it should be fully correct on the Side Icons and in the Settings (Controls->Keybinds)

LarsTwiNky:
in e.g. SpriteID.SideIcons there is no .PRAYER  -> The SpriteIDs are generated -> so shall I use _5 oder add the .PRAYER in that? (atm im just using _5, does the same thing)

LarsTwiNky:
allIconsCorrect

raiyni:
The original plugin didn't change the icons in the settings interface explicitly like this right? I wouldn't call that a bug if that's true

LarsTwiNky:
thats right, the settings werent set explicitly, but some unintendently (the ones who are sharing the samd spriteId) and the combat icon is the one whos buggy - I just added all other icons to complete that. The combat-icon in the original is only the settings sprite. the sprite from 2010 is even slightly off, so it cant match with the sailing icon atm. 

so the explicit bug on the sidebar is the combat-Icon

didnt really thought much about that - but I could split it in two PRs if requested/necessary

LarsTwiNky:
Thinking about it getting the settings in with it is kinda dumb of me - #19694 is the new PR for the bugfix - the settings-Icons should be a second PR

Issue #19578, opened 2025-11-19
Set Heading breaks right-click in camera plugin


Wynadorn:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

The new `Set Heading` option isn't on the whitelist for ignored right-click menu entries.
`Set Heading` should be treated the same as `Walk Here`.

### Screenshots or videos

n/a

### RuneLite version

Runelute: 1.12.5
Launcher: 2.7.2
osrs: Rev 235

### Logs

not needed

Wynadorn:
Fixed in f8a6d8e

PR #19579, opened 2025-11-19
#19578 - Fix `Set Heading` in camera plugin

Wynadorn:
Treat `Set Heading` the same as `Walk Here` menu entry options to fix the right-click rotates camera behaviour.

No enum value for the set heading action exists yet so for now it's a hard coded string, quick fix for launch day


Wynadorn:
Fixed by f8a6d8e

PR #19584, opened 2025-11-19
Add "Set heading" to ignored mouse tooltips (as it's "Walk here" for boats)

Jewelots:
Tooltip is displayed 24/7 while sailing currently


nuillin:
Was just about to make a PR request this change exactly ! Commenting here for visibility

cbenitezu:
Pleae, came for the same thing, comenting for visibility

PR #19587, opened 2025-11-19
devtools: render on player's current worldview

LlemonDuck:
also removes the distance check from objects since
1. it did not seem to cause a detrimental performance hit
2. it was preventing the rendering of top level objects when the player was in a boat (and therefore the distance check was not well-formed)


runelite-client/src/main/java/net/runelite/client/plugins/devtools/DevToolsOverlay.java
Nightfirecat, line 1: `MAX_DISTANCE` is an unused constant now, remove it.

Nightfirecat:
Can this be split into 2 commits please? One for multiple-wv rendering, another for no-max-distance

PR #19595, opened 2025-11-21
ground markers: sailing support

nucleon:
fixes the two bugs added with sailing, not being able to mark tiles while navigating the helm since the menu option changes and tiles not being drawn since the player isn't in tlwv

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerPlugin.java
@@ -78,8 +78,6 @@ import net.runelite.client.util.ColorUtil;
 public class GroundMarkerPlugin extends Plugin
 {
 	private static final String CONFIG_GROUP = "groundMarker";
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/groundmarkers/GroundMarkerOverlay.java
```diff
@@ -101,7 +101,7 @@ public Dimension render(Graphics2D graphics)
 
 	private void drawTile(Graphics2D graphics, WorldView wv, WorldPoint point, Color color, @Nullable String label, Stroke borderStroke)
 	{
-		if (wv.isTopLevel())
+		if (wv.isTopLevel() && client.getLocalPlayer().getWorldView() == wv)
```
Adam-: this should instead just be `if (client.getLocalPlayer().getWorldView().isTopLevel())`

nucleon: ok

nucleon:
Maybe it's best to keep the distance check ...

PR #19618, opened 2025-11-25
StatusBars: add Boat health status bar with contextual swapping, and ItemStats for boat repair kits

romeshdev:
Add to the base status bars plugin so that it supports boat health.
Also added a configurable option to context switch player health status bars to boat health bars when on a boat. This will only happen if one of the status bars isn't already _explicitly_ set to `BOAT_HEALTH`.

Context switching enabled: (Hitpoints bar converts to boat health when on a boat)
image
image

If one of the bars is manually set to `BOAT_HEALTH` (Hitpoints bar remains as is, disabling context switching toggle has the same effect)
image

On dry land there is no boat health, so manually configured `BOAT_HEALTH` bars are disabled:
image

Item Stats for repair kits:
image


Xortrox:
Is there any update on this? or is it simply in queue for review/adding? really nice feature

romeshdev:
> Is there any update on this? or is it simply in queue for review/adding? really nice feature


Haven't had any feedback on it so I believe this is still just awaiting review.

swirle13:
+1 for support for this addition, would be very helpful

PR #19641, opened 2025-12-01
Improved Garbage for TextComponent 

Mark7625:
Before: 

image

After:

image

Before it was making a new point every render making a new object now it just set the location using the internal `setLocation` in awt point


Adam-:
added in https://github.com/runelite/runelite/commit/c9b4714094fe02f66eaab66066e7c39084beaef4

PR #19671, opened 2025-12-07
Project Init, layout overview, milestones and documentation.

NicholasDavis5:


riktenx:
Doing a VR plugin for runescape is like guaranteed to get you banned FYI.

NicholasDavis5:
> Doing a VR plugin for runescape is like guaranteed to get you banned FYI.

I appreciate the concern. **It could.** My intention is to keep all interactions within a standard game window and provide a table-top VR representation of the game as an independent view layer.

PR #19679, opened 2025-12-09
emoji: add 100 emoji

rlm95:
add 💯 emoji to supported emojis


coopermor:
You can already do this with `:hundred_points:`

rlm95:
ah ok. wasn't on any of the documentation so wasn't aware. thanks!

PR #19692, opened 2025-12-13
Add Pay Dirt to mining skill calculator

nellydocs:
Adds Pay Dirt from the Motherlode Mine to the skill calculator plugin


evilgenious448:
This has previously been rejected, see #19014

nellydocs:
Fair enough

PR #19703, opened 2025-12-16
feat(mta): add anti-epilepsy option to disable hint arrows

quietdreamr:
Adds an anti-epilepsy toggle to the Mage Training Arena plugin. The new option (Disable flashing arrows) stops all arrow hint flashing across the three MTA overlays. Any existing hint arrow is cleared when the toggle is enabled or the plugin shuts down. 

Prompted by Reddit user /u/ZeroFyzix, who reported that the flashing arrow was triggering headaches due to photosensitive epilepsy ([link](https://www.reddit.com/r/2007scape/comments/1po4yzw/how_do_i_get_rid_of_this_flashing_arrow_in/))

I am open to suggestions, especially if there is a more obvious way to perhaps disable the arrows in the client scope instead of plugin-scoped.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mta/MTAConfig.java
@@ -28,9 +28,11 @@ import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 
-@ConfigGroup("mta")
+@ConfigGroup(MTAConfig.GROUP)
 public interface MTAConfig extends Config
 {
+	String GROUP = "mta";
+
 	@ConfigItem(
 		keyName = "alchemy",
 		name = "Enable alchemy room",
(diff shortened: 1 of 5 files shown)


LlemonDuck:
this is very niche and doesn't fix the underlying problem for any other cases where the hint arrow is used. imo it should either be globalized or done as a hub plugin (preferentially as a hub plugin)

quietdreamr:
> this is very niche and doesn't fix the underlying problem for any other cases where the hint arrow is used. imo it should either be globalized or done as a hub plugin (preferentially as a hub plugin)

@LlemonDuck Completely agree it's niche. I wanted to get something out quickly for the OP dealing with epilepsy, but I admit I'm probably not familiar enough with RL to design the broader fix. definitely open to better approaches like disabling arrows globally in the client (or a hub plugin like you mentioned) I'd be happy to take a stab at it or hand it off to someone who simply knows better.

JNCressey:
I made a plugin that removes that flashing arrow. It so far removes it for the vanilla uses by jagex in tutorials, Brimhaven agility arena, etc. But it doesn't work with the mage training arena plugin because MTA re-adds the hint arrow every tick. If MTA only called to set the hint arrow when the target changes, I think it would work.
https://github.com/runelite/plugin-hub/pull/9876

PR #19705, opened 2025-12-17
gpu: eliminate performance notification messages from interface buffer on Intel Linux

pacf531:
Wanted to figure out the source of the LWJGL performance notification messages when run in debug mode.
```
[LWJGL] OpenGL debug message
	ID: 0x1
	Source: API
	Type: PERFORMANCE
	Severity: NOTIFICATION
	Message: memory mapping a busy "buffer" BO stalled and took 3.748 ms.
```
Tracked it down to the `interfaceBuf` usage of `glMapBuffer`. Modernizing it to use `glMapBufferRange` with the invalidate bit eliminates the messages and also increases performance and I saw no adverse effects with the interface or rest of the image in normal usage and while doing resizing of the Runelite window. Here's a not really scientific informal test at the GE of the performance with farthest view distance and max settings at ~4K looking directly overhead while being at the center.

Before: ~53-55 FPS
before

After: ~61-63 FPS
after

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -1359,13 +1359,14 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 
 	private void prepareInterfaceTexture(int canvasWidth, int canvasHeight)
 	{
+		long bufferSize = canvasWidth * canvasHeight * 4L;
 		if (canvasWidth != lastCanvasWidth || canvasHeight != lastCanvasHeight)
 		{
 			lastCanvasWidth = canvasWidth;
 			lastCanvasHeight = canvasHeight;
 
 			glBindBuffer(GL_PIXEL_UNPACK_BUFFER, interfacePbo);
-			glBufferData(GL_PIXEL_UNPACK_BUFFER, canvasWidth * canvasHeight * 4L, GL_STREAM_DRAW);
+			glBufferData(GL_PIXEL_UNPACK_BUFFER, bufferSize, GL_STREAM_DRAW);
 			glBindBuffer(GL_PIXEL_UNPACK_BUFFER, 0);
 
 			glBindTexture(GL_TEXTURE_2D, interfaceTexture);
@@ -1379,7 +1380,7 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 		final int height = bufferProvider.getHeight();
 
 		glBindBuffer(GL_PIXEL_UNPACK_BUFFER, interfacePbo);
-		ByteBuffer interfaceBuf = glMapBuffer(GL_PIXEL_UNPACK_BUFFER, GL_WRITE_ONLY);
+		ByteBuffer interfaceBuf = glMapBufferRange(GL_PIXEL_UNPACK_BUFFER, 0, bufferSize, GL_MAP_WRITE_BIT | GL_MAP_INVALIDATE_BUFFER_BIT);
 		if (interfaceBuf != null)
 		{
 			interfaceBuf


Adam-:
The driver performance warnings are actually specific to each driver. For example, I get no performance notification from the existing code nor any performance boost from this patch.

pacf531:
Hello, thanks for merging the patch and sorry for the delayed response.

I should clarify what set of hardware I saw these warnings pop up. I saw this on an Intel Arc dGPU (Intel Arc A770) Linux system. However, I went back and tested this on a Nvidia dGPU (Nvidia RTX 2000 Ada)+ Intel iGPU (Meteor Lake Arc Graphics with 8 Xe Cores) laptop on Windows after the fact with the same informal benchmark as above albeit at a lower 1920x1200 resolution.

For Intel, before, it was hovering around 42-44 FPS but would occasionally dip down to low to mid 20s.
before_intel
After, we are hovering around 48 FPS without dips.
after_intel

With Nvidia, before, around 60 FPS.
before_nvidia
After, negligible change at the same 59-60 FPS.
after_nvidia

Unfortunately, I don't have any other platforms and GPUs to see if it would be a regression or not but hopefully this covers most of the market and platforms that people have. If it does cause problems, feel free to revert.

Adam-:
I found a few people with integrated intel gpus that it helped on, so decided to give it a try.

Adam-:
There is correlation between this commit going out and vm crashes in glUnmapBuffer()

image

The crashes all mostly look like this:

```
Current thread (0x0000021305568000):  JavaThread "Client" daemon [_thread_in_native, id=11460, stack(0x00000086dbc00000,0x00000086dbe00000)]

Stack: [0x00000086dbc00000,0x00000086dbe00000],  sp=0x00000086dbdff1b0,  free space=2044k

Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [atio6axx.dll+0xb6d908]

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 8325  org.lwjgl.opengl.GL15C.glUnmapBuffer(I)Z (0 bytes) @ 0x000002137507a601 [0x000002137507a5c0+0x0000000000000041]
J 12493 c2 net.runelite.client.plugins.gpu.GpuPlugin.prepareInterfaceTexture(II)V (233 bytes) @ 0x0000021375705810 [0x00000213757055c0+0x0000000000000250]
J 8323 c1 net.runelite.client.plugins.gpu.GpuPlugin.draw(I)V (229 bytes) @ 0x000002136eaef1dc [0x000002136eaee9e0+0x00000000000007fc]
J 8172 c1 bm.uk(Lbm;Ljava/lang/Object;S)V (103 bytes) @ 0x000002136ea6d4b4 [0x000002136ea6cca0+0x0000000000000814]
J 12055 c1 bm.ni(Ljava/lang/Object;)V (9 bytes) @ 0x000002136d738ab4 [0x000002136d738a40+0x0000000000000074]

```


So I will revert this. There is a somewhat aggressively worded PR #19715 claiming that this change is actually against spec, but I am not familiar enough with the spec to tell if that is true.

pacf531:
Hello,

I've been out for most of yesterday so I didn't see the whole thing with regards to being against spec or not but it seemed to have resolved itself. For the record, I was quite sure that I had coded correctly with regards to the OpenGL spec and making sure Mesa didn't error out anywhere but I myself am not a graphics programmer at all, just have done some tinkering here and there. The PR unfortunately is causing issues regardless so I'm sorry for the inconvenience and understand why you reverted it.

I'll see if I can fix this in a generic manner and will open a new PR if I do find something worth discussing about this. Thanks for the feedback.

Adam-:
All of the crashes I see are in AMD GPU drivers, so it could be a driver bug.

PR #19712, opened 2025-12-20
[GameEventManager] Post VarbitChanged for Varbits & VarPlayers

Infinitay:
Added support for resending VarbitChanged events for all VarPlayers and also Varbits. Purpose for this is to remain inline with re-sending other events when a plugin is enabled and the player is already in game. Previously, you'd have to handle re-fetching the Varbit/VarPlayer state yourself, but this helps avoid that.

- I had to suppress `MagicConstant` to avoid warnings for `Client#getVarpValue` and `Client#getVarbitValue` when passing in type `int`
- The additional checks in the if statement is probably unnecessary since we already know them to be `public static final int`, but I checked it anyways for safety
- I believe we could get by iterating through only the Varbits and storing a reference set for the VarPlayers - sending the VarbitChanged as VarPlayer changed
	- The reason I didn't is because I figured looping through currently ~3,000 ints is probably better than storing them, but I digress, I can make the change if you'd like

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/util/GameEventManager.java
+++ b/runelite-client/src/main/java/net/runelite/client/util/GameEventManager.java
@@ -24,9 +24,12 @@
  */
 package net.runelite.client.util;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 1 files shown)


Nightfirecat:
We discussed this on Discord a bit; we'd prefer plugins check var values on startup rather than expect events be posted when they start. For that reason, we're not looking to update the game event manager in this way.

PR #19715, opened 2025-12-21
Draft: Add a second pixel buffer object for the interface

kiroma:
I've been nerdsniped into looking at https://github.com/runelite/runelite/pull/19705 and thought that the issue "fix" as provided is insane and misguided at best.

The problem here is quite deeply rooted as the way the spec defines `glMapBuffer*`  is ludicrously strange considering the rest of the OpenGL interface. It is neither a synchronization point nor a command, but instead it is meant to provide a view into the buffer regardless of whether if it's being actively written to, or read from, no matter which flags are specified during the mapping.
This is of course unexpected by the vast majority of the developers who continue to use this function as if it provided either a staging buffer or perform a stall in order to avoid sync issues, and indeed this is what's happening here in the real world. Unfortunately for Intel users, it seems that instead of transparently providing a staging buffer, they instead opted to simply stall until the command queue is empty, or if you're lucky, somehow determined that the buffer is no longer in use.
The currently posted fix likely improves performance by avoiding the driver's attempt to fix the sync issue that the application is now abusing. In some edge cases I believe this would lead to some visible corruption on the screen, at least under a conformant driver.

This pull request instead proposes a better fix for the current issue by adding a second buffer that can be mapped and written to while the previous frame is still being drawn by the GPU. Please do note that I'm posting this as a draft since I am not a Runescape player, nor an Intel Arc user. Sorry.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -155,6 +155,8 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 	private int interfaceTexture;
 	private int interfacePbo;
 	private int interfaceBackPbo;
+	private long interfaceFence = 0;
+	private long interfaceBackFence = 0;
 
 	private int vaoUiHandle;
 	private int vboUiHandle;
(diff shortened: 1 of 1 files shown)


kiroma:
I am wrong.

`glMapBufferRange` can have a `GL_MAP_UNSYNCHRONIZED_BIT` specified in its access bitfield. Logic would reason that by omitting this flag the driver should perform a wait until the resource is idle. So the fix was actually correct. Tracing this in Mesa drivers affirms this.
Yet https://github.com/runelite/runelite/pull/19705 now mentions that since the change was introduced, crashes have begun to occur.
Weird.

PR #19764, opened 2026-01-05
entity hider: fix hide others for attackers

OTCompa:
From the wiki for Entity Hider:
> 1. Hide Others
(Default On) Hides other players who are not on your friends/ignore list, in your friends/clan chat, or ***in combat with you***.

I believe this implies that "Hide Others" doesn't apply to players attacking you unless "Hide attackers" is enabled too.  
On the current version, attackers are still hidden when only "Hide Others" is selected. This fix should unhide players whenever they're attacking the local player.


younghans:
I support this PR! 'Hide attackers' option unchecked should show someone attacking you, even if hide others is checked (it should override that)

I was about to make the same PR, but I saw you already made it.

Why this is important: For deadman mode breaches there are lot's of players you need to hide, but if someone starts attacking you, you want to be able to see them

PR #19765, opened 2026-01-05
entity hider: better attacker tracking

OTCompa:
Adapted from [OpponentInfoPlugin](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/opponentinfo/OpponentInfoPlugin.java).
This has a minor merge conflict with #19764, will fix if both are approved.
Currently, attackers are only hidden (with "Hide attackers" on) during the duration that they're interacting with the local player. If they stop interacting with the player temporarily like walking to another tile or eating food, they immediately stop counting as an attacker. This makes it so that they'll count as an attacker for at least 5 more seconds after their last interaction with the player.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
@@ -94,7 +94,7 @@ public class EntityHiderPlugin extends Plugin
 		NpcID.MACRO_COUNTCHECK_SURFACE, NpcID.MACRO_COUNTCHECK_UNDERWATER
 	);
 
-	private static final Duration WAIT = Duration.ofSeconds(5);
+	private static final Duration WAIT = Duration.ofMillis(9600);
 
 	@Inject
 	private Client client;


raiyni:
This seems very subjective of what you would consider an attacker.

OTCompa:
> This seems very subjective of what you would consider an attacker.

Could you please explain further? This doesn’t change anything with regards to how it initially detects an attacker, it only keeps them hidden for a little longer so attackers don’t constantly hide/unhide.

raiyni:
I would expect that behavior due to the client not knowing if the person is still attacking you, or not. Because as is, they aren't attacking you if they step under you or are chugging food.

OTCompa:
> I would expect that behavior due to the client not knowing if the person is still attacking you, or not. Because as is, they aren't attacking you if they step under you or are chugging food.

Here's an example of why this PR was made:
### Current live version

https://github.com/user-attachments/assets/0cab71df-c651-4190-b21a-8bbeaf03c09f

### This PR

https://github.com/user-attachments/assets/a7505e8e-2a16-46ff-b69a-10a303453636

You can see the attacker popping in and out on the current live version on the ticks between their attacks if they take any other action that doesn't involve interacting with you. It doesn't really make sense from a user standpoint for them to pop in and out when they are by any reasonable means still attacking you.

Semi-related, but as for why I want to hide attackers, this is actually related to my other PR so that attackers stay visible instead of popping in and out while "Hide Others" is enabled.

runelite-client/src/main/java/net/runelite/client/plugins/entityhider/EntityHiderPlugin.java
```diff
@@ -88,6 +94,8 @@ public class EntityHiderPlugin extends Plugin
 		NpcID.MACRO_COUNTCHECK_SURFACE, NpcID.MACRO_COUNTCHECK_UNDERWATER
 	);
 
+	private static final Duration WAIT = Duration.ofSeconds(5);
```
hildickethan: what do you think of aligning this value with the combat logout timer? wiki says 16t/9.6s
mostly for consistency with what the game considers no longer being attacked

varriaza:
This would be amazing. Currently a huge issue in DMM breaches. I have to trade off seeing all players or not seeing who is attacking me. Both of which result in me not being able to find my attacker.

PR #19767, opened 2026-01-06
Examine: Change Pending Deque Management

### LINKED ISSUE #19762: Examine plugin does not work when other messages are received on same tick
### Type

Incorrect behavior

### Operating System

None

### OS Version

N/A

### Bug description

The Examine plugin which comes by default with RuneLite prints a line with the GE and HA value of items you examine, just after their examine text.

However, seemingly at random it will fail to do so. 

In fact, the cause of the failure is when a chat message, even an invisible / filtered one such as spam in the "Game" tab e.g. "you chop some logs", is received on the same tick as the examine text.

This is probably caused by this logical branch, indicated by debug logs when it happens:
https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java#L126

```
2026-01-04 21:36:43 GMT [Client] DEBUG injected-client - Chat message type SPAM: You manage to mine some pay-dirt.
2026-01-04 21:36:43 GMT [Client] DEBUG n.r.c.plugins.examine.ExaminePlugin - Type mismatch for pending examine: ITEM_EXAMINE != SPAM
2026-01-04 21:36:43 GMT [Client] DEBUG injected-client - Chat message type ITEM_EXAMINE: Prospector Percy has a machine that can wash this.
```

In those lines, the plugin clears the messages it would attach if the first message it reads isn't examine text, even when the examine message would have shown up later that tick. 

I don't think this "pending" deque should be cleared the way it currently is. Some kind of length-based or time-based conditions should be used for purging pending messages instead.

Thanks.

### Screenshots or videos

_No response_

### RuneLite version

latest versions

### Logs

_No response_

BBeeaarr:
Fixes #19762

Elements from the `pending` deque are no longer removed after a type mismatch. Previously `pending` would be entirely cleared upon mismatch; causing unrelated chat messages received after the user clicks examine and before the `ITEM_EXAMINE` chat message to clear `pending`. Now only the first element is removed after confirming correct chat message type. We can also use `pending.clear()` instead of `pending.poll()` if we should be clearing`pending`.

Before
pre change example GE
The autotyper message has cleared `pending` so the plugin does nothing.

After
post change examine text in GE
Now only the corresponding `ITEM_EXAMINE` chat message can clear `pending`.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
@@ -128,7 +128,7 @@ public class ExaminePlugin extends Plugin
 			log.debug("Type mismatch for pending examine: {} != {}", pendingExamine.getResponseType(), event.getType());
 			return;
 		}
-		pending.poll();
+		pending.pop();
 
 		log.debug("Got examine type {} {}: {}", pendingExamine.getResponseType(), pendingExamine.getId(), event.getMessage());
 


runelite-client/src/main/java/net/runelite/client/plugins/examine/ExaminePlugin.java
```diff
@@ -122,13 +122,13 @@ public void onChatMessage(ChatMessage event)
...
 
-		PendingExamine pendingExamine = pending.poll();
+		PendingExamine pendingExamine = pending.peek();
 		if (pendingExamine.getResponseType() != event.getType())
 		{
 			log.debug("Type mismatch for pending examine: {} != {}", pendingExamine.getResponseType(), event.getType());
-			pending.clear(); // eh
 			return;
 		}
+		pending.poll();
```
Adam-:
```suggestion
		pending.pop();
```

PR #19772, opened 2026-01-07
slayer: use predicates for task NPC matching

### LINKED ISSUE #19740: [Slayer] Slayer plugin does not highlight Prifddinas "Guard" on "Elves" slayer task
### Type

Incorrect behavior

### Operating System

MacOS

### OS Version

macOS Tahoe 26.2 (25C56)

### Bug description

The `Slayer` plugin highlights NPCs which are on your current slayer task. When assigned [Elves](https://oldschool.runescape.wiki/w/Slayer_task/Elves), Prifddinas [Guards](https://oldschool.runescape.wiki/w/Guard_(Prifddinas)) should be recognized, but they are not.

In `Task.java`, the `ELVES` enum is defined as

`ELVES("Elves", ItemID.PICKPOCKET_GUIDE_WOODELF, "Elf", "Iorwerth Warrior", "Iorwerth Archer")`

which results in `targetNames = ["Elf", "Iorwerth Warrior", "Iorwerth Archer"]`

In this case, there are other NPCs with the name "Guard" which do not belong to the same slayer task. Therefore, as far as I can tell, there is currently no way to define the Prifddinas Guard NPC.

We should have the ability to define NPCs by ID as a fallback in this scenario.

### Screenshots or videos

_No response_

### RuneLite version

RuneLite version: 1.12.10

### Logs

_No response_

jeremybahadirli:
Closes #19740

Currently, the Slayer plugin determines slayer task targets using an array of target names, and testing whether one of the target names exists within the NPC name as a whole word. A more robust system is needed to match NPCs such as Prifddinas "Guard" on the Elves task, because there are other non-task NPCs in-game which share the same name.

This PR refactors the Slayer plugin to match NPCs using a `Predicate<NPC>`, rather than `String[] targetNames`. NPC predicates can be declaratively composed using the `NpcPredicates` utility class. `NpcPredicates` provides factory methods for predicates matching exact names, name fragments (existing behavior), combat levels, and NPC ids.

Example usage:

```java
ELVES(
	"Elves",
	ItemID.PICKPOCKET_GUIDE_WOODELF,
	NpcPredicates.byName("Iorwerth Archer", "Elf Archer", "Iorwerth Warrior", "Elf Warrior", "Mourner", "Reanimated Elf")
		.or(NpcPredicates.byName("Guard").and(NpcPredicates.byCombatLevel(108)))
),
```

This change maintains backward compatibility with the existing `Task` constructors. `String[] targetNames` are converted to `Predicate<NPC>` using `NpcPredicates.byNameContaining(String...)`, which mimics current behavior by utilizing the same regex.


runelite-client/src/main/java/net/runelite/client/plugins/slayer/NpcPredicates.java
```diff
@@ -0,0 +1,173 @@
...
+ * Utility class providing {@link Predicate predicates} for matching {@link NPC NPCs}.
+ * <p>
+ * All predicates produced by this class operate on the NPC's
+ * {@link NPC#getTransformedComposition() transformed composition} and will
+ * return {@code false} if no transformed composition is available.
+ *
+ * @see NPC
+ * @see NPCComposition
+ */
+final class NpcPredicates
```
Nightfirecat: This should be singular.

runelite-client/src/main/java/net/runelite/client/plugins/slayer/NpcPredicates.java
```diff
@@ -0,0 +1,173 @@
...
+{
+	/**
+	 * Creates a predicate that matches NPCs whose transformed {@link NPCComposition#getName()} exactly equals any of the
+	 * provided names (case-insensitive).
+	 *
+	 * @param names candidate NPC names
+	 * @return a predicate that matches NPCs by the specified names
+	 */
+	static Predicate<NPC> byName(String... names)
```
Nightfirecat: I'm not sure I like the method names here as they're not entirely intuitive. I might prefer `nameEquals`, `nameContains`, `combatLevelEquals`, and `idEquals` instead. Do you have any thoughts on changing them as such?

Nightfirecat (commented):
I apologize for letting this sit without review for as long as it did, this looks pretty good at a glance. Could you split the changes into two commits: the first introducing NpcPredicate and updating the plugin and Task ctors to use it, then the following commit fixing the elves highlighting? (While you're at it, #19956 has been opened since this PR was created; could you add another commit to address it? I think using exact name matching should fix it)

Nightfirecat:
In addition to the above feedback, I think using @dorquex-ctrl's added test cases from #19968 would also be useful to validate these changes

runelite-client/src/main/java/net/runelite/client/plugins/slayer/NpcPredicates.java
jeremybahadirli: Fixed

runelite-client/src/main/java/net/runelite/client/plugins/slayer/NpcPredicates.java
jeremybahadirli: After looking at it again, I agree that your naming is better.

jeremybahadirli:
Separated the plugin changes, new task predicates, and new tests into separate commits as requested.

Added predicates to fix the following tasks:

- Cave kraken
- Kraken boss
- Cows
- Crabs
- Elves
- Vampyres

Nightfirecat:
@jeremybahadirli I've done some touchups and commit squashing, can you check on the commits/changes in this PR once more? I think I'm happy with how they look now.

jeremybahadirli:
@Nightfirecat Looks good to me.

Issue #19786, opened 2026-01-11
[Clue Scroll] Helper for medium clue still recommends bringing rope with Kandarin diary completed


fuzxi:
### Type

Incorrect behavior

### Operating System

None

### Bug description

For [this medium clue step](https://oldschool.runescape.wiki/w/Clue_scroll_(medium)_-_09.33N_02.15E), the Clue Scroll helper states "Baxtorian Falls (Bring rope)". Completing the medium Kandarin diary adds two permanent rope objects to the area that can be used instead, making it unnecessary to bring a rope.

### Screenshots or videos

<img width="581" height="307" alt="Image" src="https://github.com/user-attachments/assets/7ac7fcc1-99c6-4f32-9348-0090d2436d40" />

### RuneLite version

RuneLite version: 1.12.11
Launcher version: 2.7.5

CraigStrathclyde:
PR Raised for review :) 

https://github.com/runelite/runelite/pull/19791

zneix:
This has been fixed in https://github.com/runelite/runelite/commit/1982b0cf4dc58658fe67959a9a1d33662b3a5a69

PR #19791, opened 2026-01-12
#19786 [Clue Scroll] Helper for medium clue still recommends bringing…

CraigStrathclyde:
Resolved with helper method to check diary completion then display either bring rope or not.


Nightfirecat:
Added in 1982b0cf4dc58658fe67959a9a1d33662b3a5a69, thank you!

PR #19800, opened 2026-01-15
Time Tracker: Show Falador patch protection with elite diary

evaan:
Show the tree protection icon on the time tracking pane if the Falador diary message is in the chat


Adam-:
added in ced8cdca0a9e2992a549934503e9c847ab7c2b87

PR #19836, opened 2026-01-30
Special Attack Counter: Filter out thrall and other non-player hitsplats

### LINKED ISSUE #19722: Special Attack Counter plugin mistakes thrall damage as eye of ayak damage
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

When using the Eye of Ayak special attack at Yama, while using thralls, the thrall damage is sometimes added to the total tracked spec damage instead. I am guessing this is happening whenever the thrall hits at the same time as the eye of ayak, and the code is somehow mixing up the two.

I am also guessing that this happens with other weapons besides the Eye of Ayak, or other bosses besides Yama, but I haven't tested it.

### Screenshots or videos

_No response_

### RuneLite version

Runelite 1.12.8
Launcher 2.7.5
Olschool Rev 235

### Logs

_No response_

JoneJeremiassen:
Fixes #19722

When using special attacks with thralls active, the plugin was sometimes counting thrall damage as weapon damage when both hit on the same tick.

This occurred because thralls produce DAMAGE_OTHER hitsplats, which the previous isMine() check didn't filter out. The fix adds an isOthers() check to exclude hitsplats from thralls and other non-player sources.

Added test case to verify thrall hitsplats are correctly ignored.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/specialcounter/SpecialCounterPlugin.java
@@ -261,7 +261,7 @@ public class SpecialCounterPlugin extends Plugin
 				}
 
 				// The weapon hitsplat is always last, after other hitsplats which occur on the same tick such as from
-				// venge or thralls.
+				// venge. Thrall hitsplats are filtered out in onHitsplatApplied.
 				Hitsplat hitsplat = hitsplats.get(hitsplats.size() - 1);
 				specialAttackHit(specialWeapon, hitsplat.getAmount(), lastSpecTarget);
 			}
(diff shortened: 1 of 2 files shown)


Adam-:
Our project does not accept AI-written pull requests—not because we object to AI use in general, but because it adds friction without adding value. If we wanted AI-generated code, we could generate it ourselves in a form that matches our expectations and constraints. When a contributor submits AI-written code, the maintainer still has to do the same work: understand the intent, verify correctness, and ensure it fits the project. In practice, this often takes more time than simply writing the correct solution directly.

Additionally, reviewing AI-generated submissions tends to introduce an inefficient feedback loop. Review comments are frequently passed verbatim back to an LLM, which produces another revision that still requires careful human verification. This extra “hop” slows down iteration and diffuses responsibility for the code’s correctness. We value contributions where the author fully understands, owns, and can directly reason about the changes they submit. For this project, human-authored code leads to faster reviews, clearer communication, and higher overall quality.

PR #19837, opened 2026-01-30
Fix/thrall damage special attack bug fix

### LINKED ISSUE #19722: Special Attack Counter plugin mistakes thrall damage as eye of ayak damage

JoneJeremiassen:
This should hopefully fix #19722


Adam-:
This code doesn't work. If you want to resubmit code you've tested I will look at that.

PR #19848, opened 2026-02-02
feat:Entity Hider: Add per-NPC hiding

albus-lucius:
## Summary
Adds a sidebar panel to Entity Hider for per-NPC visibility control. Click any NPC in the list to hide/unhide it individually.

## Features
- Sidebar panel listing nearby NPCs with name and ID
- Click any NPC to instantly toggle its visibility
- Hidden NPCs shown in **red**, visible in **white**
- Hide Vicinity / Unhide Vicinity for bulk control
- Unhide All with confirmation dialog
- Hidden state persists across sessions
- Panel can be disabled via config toggle

## How it works
- Uses existing `shouldDraw` hook — no new performance overhead
- Hidden NPC IDs stored in config as comma-separated string
- HashSet lookup O(1) per render check
- Panel updates once per game tick

### Before — NPC visible (white in panel)
Before Hide

### After — NPC hidden (red in panel, gone from game)
After Hide


albus-lucius:
Closing in favor of a new PR with improved code

PR #19854, opened 2026-02-04
slayer: fix aquanite task name

hildickethan:
fixes
- Aquanite task name was wrong and still showed enchanted gem icon

checked the other tasks added with sailing and they should be ok, not 100% on the Shellbane Gryphon one since I can't find a source on the slayer task names like "the cave kraken boss" so I assume it's ok unless it's reported otherwise


Nightfirecat:
This, and 63763c82a0735a5452cc3d11beb3bf7295cee7db, are both confirmed in the cache by referencing dbrows 9415 and 9405 respectively.

PR #19866, opened 2026-02-08
PluginItem name field truncation bugfix

abristow3:
The name JLabel does not have a min width, so if a long string is entered in the author field, the author JLabel will force the name JLabel to truncate in order to make space. 

This change adds a min width to the name and author JLabels, currently set to 85 and 60 pixels respectively (tweak as needed), and respects the 200px fixed width of the PluginPanel - icon: 48px, 5px gap, plugin name 85px, preferred gap -2px?, author 60 px

The alignment is set to left for the plugin name JLabel so it grows right naturally until truncating.
The alignment is set to right for the plugin author JLabel so it grows left naturally until truncating.

NOTE: This change only will affect the PluginItem objects made from the PluginHubPanel, meaning the core PluginItems JLabels will not see this change. If the change is appropriate and accepted, I can make the change there as well. 

Before Image:
image

After Image:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
@@ -257,6 +257,7 @@ class PluginHubPanel extends PluginPanel
 			JLabel author = new JLabel(manifest.getAuthor());
 			author.setFont(FontManager.getRunescapeSmallFont());
 			author.setToolTipText(manifest.getAuthor());
+			author.setHorizontalAlignment(SwingConstants.RIGHT);
 
 			JLabel version = new JLabel(manifest.getVersion());
 			version.setFont(FontManager.getRunescapeSmallFont());
@@ -394,8 +395,6 @@ class PluginHubPanel extends PluginPanel
 			}
 			addrm.setBorder(new LineBorder(addrm.getBackground().darker()));
 			addrm.setFocusPainted(false);
-			pluginName.setHorizontalAlignment(SwingConstants.LEFT);
-			author.setHorizontalAlignment(SwingConstants.RIGHT);
 
 			layout.setHorizontalGroup(layout.createSequentialGroup()
 				.addGroup(layout.createParallelGroup()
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/config/PluginHubPanel.java
```diff
@@ -393,6 +394,8 @@ else if (update)
 			}
 			addrm.setBorder(new LineBorder(addrm.getBackground().darker()));
 			addrm.setFocusPainted(false);
+			pluginName.setHorizontalAlignment(SwingConstants.LEFT);
```
Adam-: What are these two lines for?

abristow3: Bear with me because I am struggling to put this into words, but the way I am understanding it, since a min width greater than 0 was supplied for the labels, they now are present even when an authors string doesn't fill up the entire JLbael space, and it is aligning left by default. This makes it so you see the remaining space of the min_width is filled with an empty string, or something else, but makes it appear in the UI offset from how the alignment was originally. Whereas before since min width was 0, the author label was only as big as it ever needed to be, so the default alignment was never observed. 

Although testing it again now and thinking more on it, this only needs to be applied to the author JLabel, since the name JLabel should be aligned left by default. It is probably more appropriate to stick the author label alignment line where author JLabel is created near line 260. It also looks like a gap of a few pixels needs to be added after it as well to preserve the original right offset. I just noticed that looking at the photo in the request body, to the one I am providing below, but I added a gap of 5 and looks to be close to the same.

I commented both lines out, and below is an image where you can see the behavior when the alignment is not defined. The TOA plugin is not affected to because it is a core plugin, and this is only touching pluginhub items right now. But the sailing (pluginhub) one is affected, and the offset looks different. That is because it is filling up the remaining pixels of min_width with an empty string or whatever, and then aligning left by default.

image

Again had trouble wording this, if I need to word this differently I can try. 

tldr: 
JLabel is default align left, min width makes empty label space show up, so its offset from how it displays currently.
can remove name alignment left
move author alignment right to line 260
need to define 5 pixel gap after author so it doesn't sit on the extreme right, and looks how it does currently with the changes in this PR.

abristow3: Good morning. I wanted to check if you had a chance to read the review comment, and if these changes are appropriate. If so l can go ahead and apply them to the PluginListItem for the core plugins as well.

PR #19873, opened 2026-02-11
Slayer plugin: Show respawn timer & additional highlight options

### LINKED ISSUE #3730: Slayer Target Respawn Timer
Recent implementation of respawn timer for NPC Indicator is very nice. But I mostly use it for slayer task. However this is not implemented yet.

Add the respawn timer for slayer targets also.

Been typing in slayer monster into NPCs to Highlight option to get the respawn timer.

Hannah-GBS:
Respawn timer code is copied from NPC Indicators. I've done a few slayer tasks worth of testing but there may be some edge case NPCs that need specific testing (someone with spare slayer points on DMM could buy specific tasks I think).

I'm not very familiar with the test logic so let me know if my fix (for onGameTick now having a getWorldLocation call) is correct.

Resolves #3730

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerConfig.java
@@ -115,6 +115,39 @@ public interface SlayerConfig extends Config
 
 	@ConfigItem(
 		position = 7,
+		keyName = "highlightTrueTile",
+		name = "Highlight true tile",
+		description = "Configures whether or not the NPC true tile should be highlighted."
+	)
+	default boolean highlightTrueTile()
+	{
+		return false;
+	}
+
+	@ConfigItem(
+		position = 8,
+		keyName = "highlightSwTile",
+		name = "Highlight south west tile",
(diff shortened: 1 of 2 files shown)


Nightfirecat:
I have a couple concerns with this approach, namely:

1. Code duplication: there's no reason this code should be duplicate for two plugins besides that we have not generalized it. Going ahead with this approach would mean repeated work to bring improvements to NPC spawn highlighting to multiple sources. (I can think of a couple ways in which it could be improved--namely, not presenting respawn indicators for dynamic NPCs, ensuring respawn timers don't get overwritten/removed in unexpected scenarios, and ensuring world hops don't interfere with respawn duration observation)
2. Sort of an extension of the above, I think tracking and displaying respawn locations and times would actually be better done in NpcOverlayService, since it already handles general NPC highlighting, but that too has its deficiencies--most notably that it does not allow multiple highlights (in any combination of styles) on an NPC; the first registration of an NPC applies the only highlight on it. On top of that, I know LlemonDuck has wanted to add a more general NPC highlighting config, similar to the notification config, to unify and simplify the config interface for the increasing number of plugins using this service.

With those other shortcomings in mind, I think that we're not yet in the best position to implement this change just yet.

PR #19887, opened 2026-02-15
boss timers: add shellbane gryphon

fontivan:
- Fixes #19646
- The gryphon seems to have a non-standard death as NpcUtil.isDying() is never true for it, so instead add the unique dying animation to onAnimationChanged()
- The Shellbane Gryphon respawn timer was updated to 12 seconds on the Dec 17th update (https://secure.runescape.com/m=news/more-sailing-and-interface-changes?oldschool=1)


fontivan:
image

The health bar actually takes several seconds to disappear after the boss "dies" though the timer does work nicely with the animation detection

PR #19903, opened 2026-02-19
2005 interface fixes and improvements

debuffed:
Various fixes, adjustments, and additions using original 2005 and OSRS sprites.


raiyni:
1. Remove the changes to sprite id
2. I was under the impression that we only changed sprites that were directly in the game frame and sprites that would have been changed in 2005. Because otherwise, everything else is subjective to how the style should look.

Issue #19875, opened 2026-02-11
World Hopper Percent Loss Going Over 100%


RonnyTommy548:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 11

### Bug description

The % loss next to ping, overlayed by the World Hopper plugin, is displaying percents over 100% for me. (Btw it would be great if I could toggle this feature off since I don't want to see it at all 😄 )

### Screenshots or videos

https://github.com/user-attachments/assets/6a6d7485-cc09-48c4-901b-9e0d3410e959

### RuneLite version

<img width="172" height="50" alt="Image" src="https://github.com/user-attachments/assets/3e547135-1ba5-47a5-959d-463199bb9946" />

coopermor:
Fixed in 5413f571459d03ae9e100039acf4ca96fc75dc7a which will be in the next client release.

TheHappyPittie:
@coopermor will there be a toggle for that function as well?

hash-cloud-dev:
Please do include a setting for world hopper:

[ ] Display packet loss

So that I can have display current ping enabled but I do not want to see the packet loss (unless I check that box). 

Thanks this feather would be great for when servers have alot of packet loss but are playing fine and I don't need the anxiety of staring at the number in the top right corner but I do want the other information such as ping, basically just make Display current ping only display current ping and have a separate check box for display packet loss. Thanks for reading this, please do implement this!<3

PR #19924, opened 2026-02-26
Add option for displaying the Ping packet loss Percentage

Hiramiya:
Add an option to toggle displaying the ping packet loss percentage on the overlay

Fixes secondary request of #19875


pajlada:
Hi, there was a discussion of this recently in the `#development` channel in the RuneLite Discord if you want to read up on the reasoning behind there not being an option for this. https://discord.com/channels/301497432909414422/419891709883973642/1472907988611367047

Hiramiya:
Seems a bit odd to not include an option for something people want? Any reasoning for not having the option should then be immortalised either in the original Issue (because it should be closed if the remainder of it's request is not going to be done) or in this PR. Discord is not a good place for storing development decisions..

Psychemaster:
The crux of the discussion is that ping is a significantly less useful metric for determining connection quality once packet loss is added to the mix. The two values are both important, so having one without the other paints an incomplete picture, both for end users and the support team... much like it did before the packet loss output was added at all.

PR #19934, opened 2026-03-02
Add Brutus to Cows slayer task

mansur0108:
Brutus counts as a cow, so it should be recognized as a valid cow slayer target.


Nightfirecat:
Thanks for the PR!

PR #19936, opened 2026-03-02
Fix transformed NPCs not being tracked as slayer targets

mansur0108:
Rock Crabs and other NPCs that disguise themselves, like rocks for example were not being recognized as slayer targets because when they transformed into their attackable form, an NpcChanged event fires rather than NpcSpawned. Since there was no onNpcChanged handler, these NPCs were never added to the targets list, causing plugins that rely on getTargets() like the Slayer plugin to not recognize the slayer monster. This adds an onNpcChanged handler to handle this case.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -323,13 +323,9 @@ public class SlayerPlugin extends Plugin
 	public void onNpcChanged(NpcChanged npcChanged)
 	{
 		NPC npc = npcChanged.getNpc();
-		if (isTarget(npc))
+		if (isTarget(npc) && !targets.contains(npc))
 		{
 			targets.add(npc);
-		}
-		else
-		{
-			targets.remove(npc);
 		}
 	}
 


runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
```diff
@@ -318,6 +319,20 @@ public void onNpcDespawned(NpcDespawned npcDespawned)
...
+	{
+		NPC npc = npcChanged.getNpc();
+		if (isTarget(npc))
+		{
+			targets.add(npc);
+		}
+		else
+		{
+			targets.remove(npc);
+		}
```
Nightfirecat: What cases exist where a monster would be a target originally, then _not_ be a target after transforming?

mansur0108: What I was thinking of was the crab transforming back into a rock so I understand that point, since you can't attack a rock anyway, it arguably doesn't matter if it stays in the target list. I included the remove for cleanliness so the list accurately reflects what is actually attackable at any given time as the game no longer shows an attack option and they are classified as a rock in the menu, but I can just remove the else block if you'd prefer to keep it simpler.

Nightfirecat: One issue that will arise from this is that `targets` is a List, but some monsters can transform between forms which would both be highlighted, eg. [Wyrms](https://chisel.weirdgloop.org/moid/npc_name.html#/wyrm/). This code as written would end up adding those monsters to the list multiple times, and resulting in them remaining in the list after despawning. I recommend either removing the npc from the list before re-adding it if it is a target, or considering whether this collection can become a Set instead.

mansur0108: checked with .contains and also removed the else

PR #19966, opened 2026-03-11
worldmap: fix Heroes' Guild East mining locations

pillowfication:
these sites were swapped
image


Nightfirecat:
Those must have been wrong for quite some time... Thanks for the fix!

PR #19968, opened 2026-03-12
slayer: exclude task false positives

### LINKED ISSUE #19956: Slayer plugin highlights sailing krakens on kraken task
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

_No response_

### Bug description

When on a kraken task it will highlight the krakens from Sailing, but they do not count for the slayer task and should not be highlighted. Not sure if this happens with any other similarly named creatures.

### Screenshots or videos

Image

### RuneLite version

Runelite version: 1.12.19
Launcher version: 2.6.4

### Logs

_No response_

dorquex-ctrl:
Fix broad Slayer name-matching false positives for a few newer NPCs.

This excludes:
- Armoured/Pygmy/Spined/Vampyre krakens from Cave kraken and Kraken boss tasks
- Vampyre kraken from Vampyre tasks
- Hermit crab from Crab tasks
- Demonic Brutus from Cow tasks

These cases still need the broader positive matchers (`Kraken`, `Brutus`, and the singular task-name fallback), so removing aliases alone would break valid task targets. This patch keeps the current matching approach and adds task-specific exclusions for the known false positives.

Adds regression coverage for the excluded NPCs and for the valid counterparts that should still match.

Fixes #19956

Tests:
- `:client:test`

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/slayer/SlayerPlugin.java
@@ -40,7 +40,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Function;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Named;


Nightfirecat:
I recently reviewed #19772, which I feel is a better approach to this problem, as it also handles other cases such as highlighting elven guards (having the name "Guard", but should count towards an elf task)

dorquex-ctrl:
Thanks, that makes sense. I agree #19772 is the more general fix for this class of issue.

I opened this as a narrow patch for the confirmed false positives I found, but given your feedback I’m happy to defer to the predicate-based approach there instead of pursuing a separate exclusion-list solution.

PR #19970, opened 2026-03-14
screenshot: add delve loot

### LINKED ISSUE #19236: Screenshot - Add screenshots to collection of Doom of Mokhaiotl loot
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

_No response_

### Bug description

No screenshot found in any folder for Doom of Mokhaiotl when collecting loot from the loot interface or following successful completion of a floor. Can this be added? 

Unfortunately no unique chat message during collection. Successful floor completion as follows: 

Image

Window where you can choose to 'Claim or Leave':

Image

Final window where you can select to bank or take loot, once you've committed to claim (already banked loot in this example):

Image

### Screenshots or videos

_No response_

### RuneLite version

1:11:13.1
2.6.13

### Logs

_No response_

marknewan:
Takes a screenshot of the loot screen on claim.

closes #19236


runelite-client/src/main/java/net/runelite/client/plugins/screenshot/ScreenshotPlugin.java
```diff
@@ -97,6 +98,7 @@ public class ScreenshotPlugin extends Plugin
 	private static final Map<Integer, String> CHEST_LOOT_EVENTS = ImmutableMap.of(12127, "The Gauntlet");
 	private static final int GAUNTLET_REGION = 7512;
 	private static final int CORRUPTED_GAUNTLET_REGION = 7768;
+	private static final int DELVE_LOOT_SCRIPT = 7931;
```
Adam-: This should be in `ScriptID` and named something like `DOM_BUILD_LOOT_UI`. The last argument to the script also tells you if you have claimed the loot or not which you could use instead of checking if the widget is hidden.

marknewan: I noticed LootTracker just uses `ScriptID.DOM_LOOT_CLAIM` and I wasn't able to fire it with a different int value in-game, so it should be suffice I think.

https://github.com/runelite/runelite/blob/167cc76d5223d275bcc99a10c1d68a796a252668/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java#L965-L978

Adam-: Oh yes, this seems better. And its good to keep the plugins consistent.

Issue #19972, opened 2026-03-15
Ground items: Hidden items should override highlighted items (wildcard* usecases specifically)


A-Bomb:
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Win11

### Bug description

Example: Put "Burnt*" on highlighted items list, and "Burnt bones" on hidden items. Burnt bones should then NOT be highlighted but **EVERY OTHER** "burnt*" item should, since they weren't specifically excluded.

### Screenshots or videos

No need - read description carefully.

<img width="246" height="710" alt="Image" src="https://github.com/user-attachments/assets/ccd94204-2e41-42a7-9109-41f25dcad230" />

### RuneLite version

newest. any of them will behave like this. 1.12.20

Adam-:
Highlighted items overriding hidden items is intentionally how it works currently. Consider the inverse of your bug description:

I've hidden `*bones*` but added `Hydra bones` to highlight because I want to see those

So I think either way is valid and it's not clear that the way you are proposing is really better.

A-Bomb:
> Highlighted items overriding hidden items is intentionally how it works currently. Consider the inverse of your bug description:
> 
> I've hidden `*bones*` but added `Hydra bones` to highlight because I want to see those
> 
> So I think either way is valid and it's not clear that the way you are proposing is really better.

the point would be specificity > generality/wildcards in either scenario.. problem solved then.


**Case 1**

- Highlight: burnt*
- Hide: burnt lobster

**Result:**

- burnt lobster should stay hidden because exact hide is more specific than wildcard highlight

**Case 2**

- Highlight: burnt lobster
- Hide: burnt*

**Result:**

- burnt lobster should be highlighted because exact highlight is more specific than wildcard hide. all else should be hidden still.


not sure how it works in code right now but could just assign a priority value to wildcards vs exact matches. should be not that hard to implement really.

Exact match = specificity 2
Wildcard match = specificity 1

**Examples:**

burnt lobster
matches:

- burnt* hide -> specificity 1
- burnt lobster highlight -> specificity 2

**Winner: highlight**

burnt lobster
matches:

- burnt* highlight -> specificity 1
- burnt lobster hide -> specificity 2

**Winner: hide**

Adam-:
I've made a change to prioritize exact matches over wildcard matches, if you can test it that would be helpful https://github.com/Adam-/runelite/commit/44a2ca32fe2bec466e5476fd24f1c948adf48d91

[...]

PR #19976, opened 2026-03-16
fix(grounditems): make hidden items override wildcard highlights

DeBoogie:
Issue #19972 is caused by the ground items plugin allowing the same item to match both the highlighted and hidden lists without applying hidden as the final override.

This change makes explicit hidden matches win over explicit wildcard highlight matches for the same item and applies that rule consistently to overlay coloring, notifications and lootbeams.


Adam-:
I think having highlight>hidden is expected and I don't want to change that globally. Having this prioritize matches that are exact (non-wildcard) matches over wildcard matches would be better and would address the concerns of #19972

DeBoogie:
> I think having highlight>hidden is expected and I don't want to change that globally. Having this prioritize matches that are exact (non-wildcard) matches over wildcard matches would be better and would address the concerns of #19972

Sure. Then we ignore the change.

Issue #19975, opened 2026-03-16
[Loot Tracker] Not tracking auto-opened bird nests


Infinitay:
### Type

Incorrect behavior

### Operating System

Windows

### Bug description

When having multiple bird nests with seeds in your inventory and searching only one, the game automatically continues to search the remaining bird nests in your inventory. The problem is that RuneLite's Loot Tracker plugin only seems to track the first bird nest, and not the auto-opened subsequent ones. Manually clicking to open them has no problem (disregarding the issue when opening multiple per tick).

1. Have more than one bird nests with seeds in your inventory
2. Click to `Search` only one `Bird nest` (seeds)
3. The first one is successfully tracked within the Loot Tracker plugin
4. --> Proceeding searches that the game auto-searches for you are not tracked within the Loot Tracker plugin

### Screenshots or videos

https://github.com/user-attachments/assets/13b0ef6b-168e-43ca-be3d-860cadb28c1b

### RuneLite version

RuneLite Version: `1.12.20`
Oldschool Revision: `Rev 236`
Launcher Version: `Unknown`

_I launched RL via IDE with RL version set to `latest.release`_

### Logs

```log
[Client] DEBUG n.r.c.p.antidrag.AntiDragPlugin - Set delay to 12
[Client] DEBUG c.g.itemcharges.ChargedItemManager - OPTION | option: Search, target: <col=ff9040>Bird nest</col>, action name: CC_OP, action id: 57
[Client] DEBUG injected-client - Chat message type SPAM: You take an apple tree seed out of the bird's nest.
[Client] DEBUG c.a.runelite.watchdog.EventHandler - SPAM | : You take an apple tree seed out of the bird's nest.
[Client] DEBUG c.g.itemcharges.ChargedItemManager - MESSAGE | type: SPAM, message: You take an apple tree seed out of the bird's nest., sender: null
[Client] DEBUG c.g.itemcharges.ChargedItemManager - ITEM CONTAINER | 93
[Client] DEBUG c.scrollboxinfo.ScrollBoxInfoPlugin - Loading bank scroll flag key = hasScrollInBank_SHARED_BEGINNER -> false
[Client] DEBUG c.scrollboxinfo.ScrollBoxInfoPlugin - Saving total count key = total_SHARED_BEGINNER -> 5
[Client] DEBUG c.scrollboxinfo.ScrollBoxInfoPlugin - Loading bank scroll flag key = hasScrollInBank_SHARED_EASY -> false
[...]

PR #19977, opened 2026-03-16
fix(loottracker): Track automated bird nest searching (#19975)

Infinitay:
### fix(loottracker): Track automated bird nest searching (#19975)

- Added support for tracking the loot received from the next searched bird nest that's automated by the game itself
	- Modified LootTrackerPlugin#onItemContainerChanged to see if a bird nest was searched by the game and whether or not to keep processing inventory changes while there are searchable bird nests remaining
- Added new test case LootTrackerPluginTest#testBirdNests

The following changes were made after noticing that when you `Search` -> `Bird nest`, the following searches that are automated by the game does not trigger `MenuOptionClicked`. It'll still fire `ItemContainerChanged` obviously because the inventory is changed. However, because we are calling `#resetEvent`, we end up setting `inventoryId` to `-1` and therefore early return out before the event handler could handle  `#onInvChange`. That's why I added a `ignoreReset` check in order to skip resetting the inventory and setting up the state to skip ItemContainerChanged.

---

**Note that while I was writing the tests, I realized there is a chat message being sent when you search a nest. I'm looking into relying on `ChatMessage` events instead of the _patch-work_ within `#onItemContainer`. Perhaps it'd also allow for multiple searches in the same tick too.**

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
@@ -1172,6 +1172,11 @@ public class LootTrackerPlugin extends Plugin
 		{
 			countChangedItems(ItemID.DIRTY_ARROWTIPS, client.getBoostedSkillLevel(Skill.FLETCHING));
 		}
+
+		if (message.endsWith("out of the bird's nest."))
+		{
+			onInvChange(collectInvItems(LootRecordType.EVENT, BIRDNEST_EVENT));
+		}
 	}
 
 	private void countChangedItems(int itemId, Object metadata)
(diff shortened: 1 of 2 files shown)


Infinitay:
### revert(loottracker): Reverted ItemContainerChanged changes

- Reverted the changes made within #ItemContainerChanged

### fix(loottracker): Track automated bird nest searching onChatMessage (#19975)

- Added support for tracking the loot received from the next searched bird nest that's automated by the game itself
	- Modified LootTrackerPlugin#onChatMessage to see if loot from a searched bird nest was received and call #onInvChange
		- The message is also fired for the automated proceeding searches by the game

---

Found a simpler way to hand the automated searches. Regardless, the changes still don't enable tracking multiple searches in the same tick - but that was the case prior to this as well.

PR #19984, opened 2026-03-19
api: add animation IDs for trailblazer reloaded variants of dragon tools

rfotino:
I saw someone mining at MLM with this pickaxe https://oldschool.runescape.wiki/w/Dragon_pickaxe_(or)_(Trailblazer)#Trailblazer_Reloaded and noticed veins were not starting to deplete in the plugin. It looks like it is a separate animation id that I was able to grab with the creator's kit plugin.

After adding this to the animation id file I would update here to include it in Motherlode Mine Improved https://github.com/TicTac7x/runelite-plugins/blob/86916a10ae623c1356356b0c69144b12c880f935/src/main/java/tictac7x/motherlode/oreveins/OreVeins.java#L157

Where I grabbed the animation id from creator's kit:
Screenshot 2026-03-18 182604

Video of the bug where this character's pickaxe doesn't start the vein depleting:

https://github.com/user-attachments/assets/b3b1c34a-b945-48f7-b1bb-2621b5fc4a04

### DIFF
--- a/runelite-api/src/main/java/net/runelite/api/AnimationID.java
+++ b/runelite-api/src/main/java/net/runelite/api/AnimationID.java
@@ -45,6 +45,7 @@ public final class AnimationID
 	public static final int WOODCUTTING_GILDED = 8303;
 	public static final int WOODCUTTING_DRAGON = 2846;
 	public static final int WOODCUTTING_DRAGON_OR = 24;
+	public static final int WOODCUTTING_DRAGON_OR_TRAILBLAZER_RELOADED = 11940;
 	public static final int WOODCUTTING_INFERNAL = 2117;
 	public static final int WOODCUTTING_3A_AXE = 7264;
 	public static final int WOODCUTTING_CRYSTAL = 8324;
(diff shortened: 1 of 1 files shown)


rfotino:
I bought my own ornament kit and filled in some other missing animation ids where there were already the respective dragon tool (or) animations in AnimationID:

Mining a crashed star:
Screenshot 2026-03-19 164113

Chopping a tree:
Screenshot 2026-03-19 164233

Chopping an ent, animation id was the same so I didn't add another constant here but I could add one that's the same as the normal dragon axe (or) for completeness:
Screenshot 2026-03-19 164605

I'm sure harpoon animations etc also differ but weren't in the file so I didn't go out of my way to find them.

Nightfirecat:
These IDs are already defined in the gamevals package at `net.runelite.api.gameval.AnimationID`, specifically:

```java
public static final int HUMAN_MINING_TRAILBLAZER_RELOADED_PICKAXE_NO_INFERNAL_WALL = 11827;
public static final int HUMAN_MINING_TRAILBLAZER_RELOADED_PICKAXE_NO_INFERNAL_NOREACHFORWARD = 11839;
public static final int HUMAN_WOODCUTTING_TRAILBLAZER_RELOADED_AXE_NO_INFERNAL = 11940;
```

Moreover, we usually do not add API constants for hub plugins; they can define their own constants or just use a hardcoded value.

rfotino:
Ah somehow I missed the big @ deprecated tag in this file, I also didn't realize there was a much more complete list elsewhere.

Thanks for the pointers and appreciate you taking a look at this!

PR #19992, opened 2026-03-21
Skill Calculator: Moves Camphor crates to crafting table

FalloutFalcon:
I noticed making my first batch of crates, and then going to calc some construction stuff that this was misplaced.
image
image

I've moved it into the crafting section, this meant I had to change the args as it changed from named to item. I also marked it as a members action, the surrounding items dont seem to respect that arg (fire battle staff action is def members) so im unsure if I made a mistake there.

I haven't actually built it yet and looked, as it seemed like a small enough change, let me know otherwise.

I also dont know if there is a scheme for sorting outside of level and maybe xp so i just put it a the bottom of the 62 section.


evilgenious448:
@FalloutFalcon 
> I also dont know if there is a scheme for sorting outside of level and maybe xp so i just put it a the bottom of the 62 section.

Order goes Level -> XP -> Alphabetical

evilgenious448:
> I've moved it into the crafting section, this meant I had to change the args as it changed from named to item. I also marked it as a members action, the surrounding items dont seem to respect that arg (fire battle staff action is def members) so im unsure if I made a mistake there.

No mistake on your end, the member/non-member thing is very spotty in terms of accuracy, few contributors to the skill calc have been respecting it, so overall it needs one big revision to fix them all up. Has been like this since before I touched the skill calc for the first time.

FalloutFalcon:
mb sorry for the delay, i forgor!
should pass tests now.

runelite-client/src/main/java/net/runelite/client/plugins/skillcalculator/skills/CraftingAction.java
```diff
@@ -135,6 +135,7 @@ public enum CraftingAction implements ItemSkillAction
 	EARTH_BATTLESTAFF(ItemID.EARTH_BATTLESTAFF, 58, 112.5f),
 	YEW_BIRD_HOUSE(ItemID.BIRDHOUSE_YEW, 60, 45),
 	GREEN_DHIDE_CHAPS(ItemID.DRAGONHIDE_CHAPS, 60, 124, true),
+	CAMPHOR_CRATE(ItemID.CAMPHOR_CRATE, 62, 50, true),
```
Nightfirecat: This does not need a members override--camphor crates are members items.
```suggestion
	CAMPHOR_CRATE(ItemID.CAMPHOR_CRATE, 62, 50),
```

PR #19995, opened 2026-03-21
Mining: config to change the pie fill style of the rocks overlays

Cyxxus:
Allows disabling rock overlays independent of other plugin features.

Allows changing the angle, in degrees, where the pie fill starts. 90 is the default starting in the up direction.

Allows changing the fill direction. Defaults to counterclockwise which is how it worked before. Clockwise is how the official client has implemented this feature. Both fills in both directions at the same time like pacman's mouth.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningPlugin.java
@@ -210,13 +210,14 @@ public class MiningPlugin extends Plugin
 	 */
 	private void clearExpiredRespawns()
 	{
-		if(config.showMiningRocksOverlay())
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Adding this only to the mining plugin feels out of place. If we were to add this--and I'm not guaranteeing this is a change we want--I think this would fit better as a global option within `RuneLite > Overlay settings`, and would apply to all the circle/pie overlays in the client.

Cyxxus:
> Adding this only to the mining plugin feels out of place. If we were to add this--and I'm not guaranteeing this is a change we want--I think this would fit better as a global option within `RuneLite > Overlay settings`, and would apply to all the circle/pie overlays in the client.

I'm not sure where the Overlay settings are presented on the client. Can you point me to them either on the client or the source code?

PR #20001, opened 2026-03-23
Add Larupia heal stats to item stats

Koslas1:
Adds a combo heal for [Cooked Larupia](https://oldschool.runescape.wiki/w/Cooked_larupia) to item stats

Instantly healing 6 hitpoints, then healing another 5 after a 7 tick delay (4.2s) for a total of 11 hitpoints.

I noticed this was the only hunter meat not listed in the itemstats


Nightfirecat:
While I agree this item is missing, I'm inclined to introduce it via something like #19458 (or maybe something resembling #18533) as they would give a better indication of whether the totality of the healing provided would be fully applied or encounter hp caps.

Koslas1:
Well I did see that the other hunter meats were added this way, which are under the line I added, this was the only hunter meat missing that I could see.

I also cannot actually recall how it is displayed on RuneLite using the 
`combo(food(6), food(5))`
Which all of the others use as I'm not near my PC. I thought it showed it as 2 separate heals - but could be wrong.


I mainly thought it was just weird to be missing only a single hunter meat, when all the others are there (Even if it does show total health healed only)

PR #20018, opened 2026-03-27
party: Add !inv(ite) to allow for easier party creation/joining

Thource:
This PR adds a new !invite command, which lets the player easily create a new party with a passphrase of their choosing and send the invite to the chat, where it can be right clicked by others to join.

Here's some quick videos of the new command in action:

https://github.com/user-attachments/assets/415e58c8-cc21-4fb4-b0e7-69b14043bacf

https://github.com/user-attachments/assets/75886d87-f113-4a3f-b9a8-5817eee6ed39


SvartaHjarta:
Great idea, will make use of this when merged.

Would it be possible to handle the case of passing no argument as follows?
```
if an argument is passed
  change players party to argument

if no argument is passed and player is not in a party
    rejoin players last party

send an invite to players current party
```

Issue #20010, opened 2026-03-25
Fix various `client.getLocalPlayer()` NPEs in the client


Nightfirecat:
As of a recent client update, `client.getLocalPlayer()` is now much more likely to be null, particularly close to logging in, so we should update the client code (plugins, etc.) to ensure we're properly null-checking it where needed. Some examples are given below:

```
2026-03-25 16:35:58 CET [Client] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
java.lang.NullPointerException: null
    at net.runelite.client.plugins.npcunaggroarea.NpcAggroAreaPlugin.onLogin(NpcAggroAreaPlugin.java:467)
```

```
2026-03-25 16:35:59 CET [Client] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
java.lang.NullPointerException: null
    at net.runelite.client.plugins.loottracker.LootTrackerPlugin.onChatMessage(LootTrackerPlugin.java:1049)
    at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:70)
    at net.runelite.client.eventbus.EventBus.post(EventBus.java:223)
    at net.runelite.client.callback.Hooks.post(Hooks.java:217)
```

BBeeaarr:
Other than the NPC aggro and LootTracker plugins mentioned in this issue should all plugins that call `client.getLocalPlayer()` without some sort of null checking be updated to check that `client.getLocalPlayer() != null` as a precaution? Or should I do some digging to see where the check is needed? There are lots of occurrences of `client.getLocalPlayer().someOtherFunction()` but not all seem to be throwing a NPE.

Nightfirecat:
I'm sure some calls may still be fine, so the ones which are causing NPEs should be prioritized first

hex-agon:
Some other plugins that can throw, i've found that a quick way triggering this is logging in while in a boat
```
java.lang.NullPointerException: Cannot invoke "net.runelite.api.Player.getWorldLocation()" because "player" is null
    at net.runelite.client.plugins.timersandbuffs.TimersAndBuffsPlugin.onGameTick(TimersAndBuffsPlugin.java:1205)
    
java.lang.NullPointerException: Cannot invoke "net.runelite.api.Player.getWorldLocation()" because the return value of "net.runelite.api.Client.getLocalPlayer()" is null
[...]

PR #20040, opened 2026-04-01
Fix Various NPEs When Calling getLocalPlayer Near Login

BBeeaarr:
A bit of a naive approach but I turned on all core plugins and looked for NPEs caused by `client.getLocalPlayer().someOtherFunction()` near the login boundary. I'm sure there are others that have not been reported yet but that will require more careful testing of boundary cases. 

Other than the NPEs listed in #20010 I observed the agility plugin throwing the following error:
```
2026-03-31 23:36:00 PDT [Client] WARN  n.runelite.client.eventbus.EventBus - Uncaught exception in event subscriber
	at net.runelite.client.plugins.agility.AgilityPlugin.onStatChanged(AgilityPlugin.java:240)
	at net.runelite.client.callback.Hooks.post(Hooks.java:217)
	at ow.qh(ow.java:35667)
	at ow.aj(ow.java)
	at ow.it(ow.java)
	at client.queueChangedSkill(client.java:52003)
	at net.runelite.client.plugins.virtuallevels.VirtualLevelsPlugin.simulateSkillChange(VirtualLevelsPlugin.java:140)
	at net.runelite.client.callback.ClientThread.lambda$invoke$0(ClientThread.java:49)
	at net.runelite.client.callback.ClientThread.invokeList(ClientThread.java:119)
	at net.runelite.client.callback.ClientThread.invoke(ClientThread.java:101)
	at net.runelite.client.callback.Hooks.tick(Hooks.java:241)
	at client.io(client.java:17007)
	at client.il(client.java)
	at sw.agt(sw.java:447)
	at sw.ko(sw.java)
	at sw.run(sw.java:17786)
	at java.base/java.lang.Thread.run(Thread.java:829)
```


runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -459,12 +461,12 @@ private void saveConfig()
 		}
 	}
 
-	private void onLogin()
+	private void onLogin(Player localPlayer)
```
Nightfirecat: What is the purpose of this added parameter? Avoiding a "method value could be null" warning or something of the sort? I'd rather we leave this method alone after the changes preceding the callsite for it.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -486,7 +488,22 @@ public void onGameStateChanged(GameStateChanged event)
 				if (loggingIn)
 				{
 					loggingIn = false;
-					onLogin();
+					clientThread.invokeLater(() ->
+					{
+						if (event.getGameState() != GameState.LOGGED_IN)
+						{
+							return true;
+						}
```
Nightfirecat: What situation is this trying to avoid? This check should have a comment explaining the need for it.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
```diff
@@ -486,7 +488,22 @@ public void onGameStateChanged(GameStateChanged event)
...
+						final var localPlayer = client.getLocalPlayer();
+						if (localPlayer == null)
+						{
+							return false;
+						}
+						onLogin(localPlayer);
+						scanNpcs();
```
Nightfirecat, line 510: `scanNpcs()` wasn't being called before, why is it necessary now?

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
BBeeaarr: This was added as a precaution to avoid an infinite loop in `invokeLater` should the gamestate change from LOGGED_IN to another gamestate where `localPlayer == null`. Possibly a gamestate of `CONNECTION_LOST` occurring right after login could cause this. If this is not a concern I can remove this check.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
BBeeaarr: ~`onLogin()` calls `loadConfig()` which populates the array `safeCenters` and finally `scanNpcs()` eventually calls `generateSafeArea()` which generates the NPC aggro reset borders. With the `invokeLater` method likely delaying the call of `onLogin()` I found upon logging in the NPC aggro reset borders were not drawn even if the player did not move at all. This makes sense, we are delaying the `onLogin()` call until `client.getLocalPlayer()` is not null.~

scanNpcs() must be called after `onLogin()` in order for NPC non-aggression area to be drawn from config.

runelite-client/src/main/java/net/runelite/client/plugins/npcunaggroarea/NpcAggroAreaPlugin.java
BBeeaarr: `onLogin()`'s only callsite is within the `onGameStateChanged` function which I changed to include a `client.getLocalPlayer() != null`  check. I am reusing that variable instead of calling it again. I can revert this change if that is not necessary.

raiyni: There isn't a need to pass around a local player variable.

BBeeaarr:
PR updated to address all NPEs reported in #20010

PR #20054, opened 2026-04-08
emoji: fix :shortcode: parsing after emoticons

rsbfox:
Reset idxStart on whitespace to prevent triggers with ':' from breaking subsequent :shortcode: emojis. Reordered blocks so if a line ends with  a :shortcode:, it is processed before the end-of-line check resets idxStart.

image

image


Nightfirecat:
Can you add a test for this case?

PR #20058, opened 2026-04-09
remove <color> tag from npcID names

DamnLol-GIT:
dont mind my amazing paint skills
before
after


Felanbird:
I don't really think we would want this? If an NPC has a color tag in their name I feel like that information should be known

DamnLol-GIT:
*facepalm* makes sense, just stuck out while i was testing something else out and quickly fixed it for myself lol
ill close

Issue #20113, opened 2026-05-06
Interact highlight showing too low on NPC's


eepykuromi:
### Type

Incorrect behavior

### Operating System

Linux

### OS Version

CachyOS

### Bug description

bug occurs regardless if I'm on GPU/117 HD/or default renderer

Bug also occurs on other plugins that highlight npcs, like player outline.

Bug also occurs on a completely fresh profile and fresh install.

### Screenshots or videos

<img width="449" height="405" alt="Image" src="https://github.com/user-attachments/assets/fa1ba211-6a87-4916-9bd0-84558b97e767" />

<img width="360" height="282" alt="Image" src="https://github.com/user-attachments/assets/11c7eb78-4a73-48ce-9daa-d67b5a574e46" />

### RuneLite version

Runelite version: 1.12.26.1
Launcher version: 2.7.7

G-E-N-E:
Same issue but even worse inside the hunter's guild. All npc highlights are not in sync with npc/creature; instead they are below. Impacts tag highlights, and hunter rumours plugin so I think it's a general client issue.

<img width="403" height="405" alt="Image" src="https://github.com/user-attachments/assets/96f5b662-8570-4eb2-9249-32c520213f2a" />

Shadefang:
I _think_ I just ran across this as well, though it could be a similar but separate issue

<img width="119" height="199" alt="Image" src="https://github.com/user-attachments/assets/665fe4e7-7499-462d-9900-4083190c181c" />

I noticed it on docks specifically, but it could be anywhere that the object is on a "floor" that's offset from the terrain?

RL version: 1.12.26.1
OS: windows 10
Seen on region locker GPU plugin.

PR #20114, opened 2026-05-06
Fix NPC/player highlight offset issue

kenedii:
Fix for #20113 

Original LocalPoint-based height sampling (which did bilinear interpolation and handled bridge/scene coordinates) replaced with direct WorldView.getTileHeight() calls, bypassing interpolation and bridge-aware adjustments and producing incorrectly low, misaligned highlights.

This reverts the recent getFootprintTileHeight changes in Perspective.java and restores the original LocalPoint-based height sampling.

This brings actor highlight height calculations back in sync with the actual scene geometry, fixing overlays that were appearing too low or inside the ground on NPCs and players.


Adam-:
I think that #20113 is fixed by https://github.com/runelite/runelite/commit/8b312880e5c58937381074efd05f8601fe145ff0 already?

kenedii:
You're correct - nice work, my bad!

PR #20121, opened 2026-05-08
idle notifier: fire idle when inventory becomes full while fishing

Vinnie-Singleton:
If you're AFK fishing and the inventory fills up, the idle notification doesn't fire while the fishing spot is still there. It only fires if the spot moves.

`checkInteractionIdle` only triggers when `Player.getInteracting()` is null. The fishing spot NPC is what you're interacting with while you fish. When the inventory fills you stop fishing, but the game client still has the fishing spot as your target, so `lastInteracting` gets refreshed every tick and the timer never elapses.

Adds a check on `ItemContainerChanged` for when the inventory hits 28/28 while interacting with a fishing spot, and sets a flag. `checkInteractionIdle` now also triggers on that flag, even when the target is still the spot. The flag gets cleared by `resetTimers()`, which already covers input activity, game state changes, and new interactions, so combat, spot despawn, and world hops behave the same as before.

Mining and woodcutting use GameObjects (rocks and trees) instead of NPCs and don't have this issue.

Two tests added: full inventory fires, partial inventory doesn't.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/idlenotifier/IdleNotifierPlugin.java
@@ -36,6 +36,7 @@ import net.runelite.api.Client;
 import net.runelite.api.Constants;
 import net.runelite.api.GameState;
 import net.runelite.api.Hitsplat;
+import net.runelite.api.ItemContainer;
 import net.runelite.api.NPC;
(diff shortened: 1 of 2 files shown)


coopermor:
The getInteracting API will be fixed in the next runelite update. It was a regression from Wednesday's client/engine update.

Vinnie-Singleton:
Didn't realize this was already in the works. Closing this since the API fix will cover it.

PR #20131, opened 2026-05-12
add pvp hider to entity hider

LogicalSoIutions:
Add's the PvP Auto Entity Hider straight into Entity Hider - does not affect any existing functionality and allows anybody using RuneLite to enable this setting for a fair playing field amongst PvPers


younghans:
Supporting this core approach from runelite/plugin-hub#15743. We tested the same core behavior in-game for Deadman Mode breaches: while other players are hidden, the active attacker stays visible, and attacker memory bridges gaps between attack cycles. A 15-second grace period worked well for our use case, which this PR supports through its configurable 1–15 second setting. Moving it into core Entity Hider also makes the behavior equally available to all players.

PR #20140, opened 2026-05-16
gpu: Orphan interface PBO before mapping

hansjm10:
The UI is uploaded every frame through `GL_PIXEL_UNPACK_BUFFER`. The PBO storage is currently only reallocated when the canvas size changes, so mapping it can block if the previous frame's texture upload is still using the buffer.

Re-specifying the buffer before mapping lets the driver provide fresh storage for the next upload instead of waiting on the previous one.

In local testing this improved FPS by roughly 20% FPS while standing in the middle of Priff.

Tested on an AMD Radeon RX 6900 XT, driver 26.3.1.


Adam-:
see also #19705 - maybe @pacf531 could also look at this?

PR #20147, opened 2026-05-20
util: add FileManager

riktenx:
tested this out in https://github.com/riktenx/loot-filters/pull/209


riktenx:
closed in favor of #20483

Issue #13477, opened 2021-04-13
New overlay lock position


LLandsem:
I would like to suggest a new overlay/ infobox lock position. I find having my overlays above the inventory is much easier to manage and keep track of. If there is a way to change this myself I would be willing to try it out. I don't know much about coding but I could figure it out with some help in the right direction. 

I have attachments to illustrate the suggestion.

![Overlay Fig 1](https://user-images.githubusercontent.com/79777597/114585215-683c5300-9c51-11eb-8d14-7f50ef777b2b.png)

![Overlay Fig 2](https://user-images.githubusercontent.com/79777597/114584145-5a3a0280-9c50-11eb-80a2-d5d479da36be.png)

Hydrox6:
Duplicate of #8515

Issue #12908, opened 2020-12-10
UI Anchor point above inventory.


bluelightzero:
Right now there are 7 anchor points (when you hold alt and move overlays, indicated by #)

It would be nice to have 1 above the inventory to anchor the map and orbs to. (Indicated by @)

    ............................................................................
    . #######                     #######                 #######      ####### .
    . #######                     #######                 #######      ####### .
    . #######                     #######                 #######      ####### .
    . #######                     #######                 #######      ####### .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                          .
    .                                                                  @@@@@@@ .
    .                                                                  @@@@@@@ .
    .                                                                  @@@@@@@ .
    .                                                                  @@@@@@@ .
    . #######                     #######                 #######  ____________.
    . #######                     #######                 ####### |____________.
    . #######                     #######                 ####### |            .
    . #######                     #######                 ####### |            .
    .____________________________________                         |            .
    .                                    |                        |            .
    .                                    |                        |            .
    .                                    |                        |            .
    .____________________________________|                        |____________.
    .                                    |                        |            .
    ............................................................................

Hydrox6:
Duplicate of #8515 (See comment [here](https://github.com/runelite/runelite/issues/10793#issuecomment-587267537) for reasoning)

Love the pixel art though, honestly great.

Issue #12216, opened 2020-07-24
Add more anchor points for draggable items


Ravyd:
There are about 5 different anchor points in the runelite client that you can pin the minimap, exp, etc. onto. The main problem is that there are only a few of them, and they usually don't work for what I would like. If the plugin isn't pinned to one of them, then opening the side panel doesn't adjust accordingly resulting in problems like the minimap going behind your inventory. 

**Describe the solution you'd like**
A larger number of anchors, or movable anchors, so that we can choose where these points are anchored at. 

**Describe alternatives you've considered**
More anchor points, freely selectable anchor points.

**Additional context**

![anchor](https://user-images.githubusercontent.com/29013305/88353710-c8821200-cd1b-11ea-8939-f0d7ff958b96.png)

JZomerlei:
Duplicate of #8515

Issue #12042, opened 2020-06-30
Suggestion: Add Inventory Viewer Grid Alignment/Snap with Inventory


Elite386:
**Is your feature request related to a problem? Please describe.**
When I'm using inventory viewer I'd like to be able to align the inventory grid with where the grid of the backpack would be. This is easy to do if I don't snap it but I'd like to be able to snap it so I can resize my clients and the inventory viewer stays aligned.

**Describe the solution you'd like**
Currently the Alt-Drag snap spot to the left of the backpack/open tab places the inventory (from what I can tell) on that spot based on the bottom-right corner. I'd like for the widget snap of the inventory viewer grid to be aligned with the inventory grid of the backpack. I'd also like a widget snap added to the top of the inventory.

**Describe alternatives you've considered**
In the plugin itself there could be a preset position so no other plugins need to use a snap and the grid could align based upon a snap point defined only for this plugin.

**Additional context**
Current snap on left: https://i.imgur.com/y6ESpKf.png
Proposed snap for grid to be aligned on top: https://i.imgur.com/yZTQPzD.png
Proposed snap for grid to be aligned on left: https://i.imgur.com/qsYeMR7.png

Alexsuperfly:
duplicate of #8515

PR #20178, opened 2026-06-03
Add a default `SnapCorner` to the top right of the inventory for resizable classic and modern

gmackall:
Proposing this as a default as it has been specifically requested many times, see:
https://github.com/runelite/runelite/issues/13477
https://github.com/runelite/runelite/issues/10793
https://github.com/runelite/runelite/issues/11491
https://github.com/runelite/runelite/issues/12042
https://github.com/runelite/runelite/issues/12216
https://github.com/runelite/runelite/issues/12908

Looks like:

https://github.com/user-attachments/assets/74e9a8af-6083-4d8b-b082-1c5714e99001

https://github.com/user-attachments/assets/42a898e4-4cdf-4d83-bd31-3fbae0cc10d6

(I just desperately want to swap to burning claws, click spec, and then not have to move my mouse much to re equip my defender)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -870,6 +870,7 @@ public class OverlayRenderer extends MouseAdapter
 		{
 			snapCorners.aboveChatboxRight.setPosition(snapCorners.bottomRight.getPositionX(), snapCorners.bottomRight.getPositionY());
 			snapCorners.canvasTopRight.setPosition(snapCorners.topRight.getPositionX(), snapCorners.topRight.getPositionY());
+			snapCorners.aboveInventoryRight.setPosition(snapCorners.bottomRight.getPositionX(), snapCorners.bottomRight.getPositionY());
 		}
 	}
 
--- a/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
+++ b/runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
@@ -755,6 +755,7 @@ public class OverlayRenderer extends MouseAdapter
 					overlayPosition = OverlayPosition.TOP_RIGHT;
 					break;
 				case ABOVE_CHATBOX_RIGHT:
+				case ABOVE_INVENTORY_RIGHT:
 					overlayPosition = OverlayPosition.BOTTOM_RIGHT;
 					break;
 			}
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/ui/overlay/OverlayRenderer.java
```diff
@@ -755,6 +755,7 @@ private OverlayPosition getCorrectedOverlayPosition(final Overlay overlay)
 					overlayPosition = OverlayPosition.TOP_RIGHT;
 					break;
 				case ABOVE_CHATBOX_RIGHT:
+				case ABOVE_INVENTORY_RIGHT:
```
gmackall, line 758: This is sort of awkward - it would contribute further to https://github.com/runelite/runelite/issues/3567.

I think abstractly there are a set of regions we want to use in the non-fixed modes, that simply don't logically map to the fixed mode. Perhaps `SnapCorner` should have a `resizableModesOnly` property, and we can filter instead of stacking them?

Seems like an architectural choice I wouldn't want to simply impose in this PR as a new contributor without clarification, though.

gmackall:
The planned change to allow modifying which part of the screen the overlay lays out relative to (https://github.com/Adam-/runelite/commits/overlay-origin) helps, but I would note that this is still independently useful.

That change doesn't handle the case where the stones at the bottom stack from 1 to 2 rows, as you shrink the width of the client:
<table>
  <tr>
    <th>1 Row</th>
    <th>2 Rows</th>
  </tr>
  <tr>
    <td align="center">
      1 Row
    </td>
    <td align="center">
      2 Rows
    </td>
  </tr>
</table>

I also don't really think that that change _should_ handle this case - the fact that it doesn't is by design. But it means that a snap corner above the inventory would still be useful in a way that the other change can't solve (also, it won't handle adjusting for whether the inventory is open or closed, but that's less important).

Adam-:
> I also don't really think that that change _should_ handle this case - the fact that it doesn't is by design. But it means that a snap corner above the inventory would still be useful in a way that the other change can't solve (also, it won't handle adjusting for whether the inventory is open or closed, but that's less important).

It would be possible to make `overlay-origin` use widgets as the origin (eg position an overlay relative to the location of the inventory). I think that would achieve what you are doing here, but it would let the overlay be anywhere and not just in the one snap corner? Not sure what to do if the inventory or side panel is hidden.

gmackall:
> > I also don't really think that that change _should_ handle this case - the fact that it doesn't is by design. But it means that a snap corner above the inventory would still be useful in a way that the other change can't solve (also, it won't handle adjusting for whether the inventory is open or closed, but that's less important).
> 
> It would be possible to make `overlay-origin` use widgets as the origin (eg position an overlay relative to the location of the inventory). I think that would achieve what you are doing here, but it would let the overlay be anywhere and not just in the one snap corner? Not sure what to do if the inventory or side panel is hidden.

Yeah certainly that would solve my use case, and also probably be pretty generally useful for other people? Always hard to guess other people's gameplay habits but it feels like placing stuff relative to inv and minimap would both be generally nice (and doing so in more than just the one snap corner would be a big win). If you ended up taking that approach in the change you land (and it works as I'm imagining) I'd happily close this

Adam-:
> > > I also don't really think that that change _should_ handle this case - the fact that it doesn't is by design. But it means that a snap corner above the inventory would still be useful in a way that the other change can't solve (also, it won't handle adjusting for whether the inventory is open or closed, but that's less important).
> > 
> > 
> > It would be possible to make `overlay-origin` use widgets as the origin (eg position an overlay relative to the location of the inventory). I think that would achieve what you are doing here, but it would let the overlay be anywhere and not just in the one snap corner? Not sure what to do if the inventory or side panel is hidden.
> 
> Yeah certainly that would solve my use case, and also probably be pretty generally useful for other people? Always hard to guess other people's gameplay habits but it feels like placing stuff relative to inv and minimap would both be generally nice (and doing so in more than just the one snap corner would be a big win). If you ended up taking that approach in the change you land (and it works as I'm imagining) I'd happily close this

I'll try it and see. At the very least this PR is wrong since it needs to do something in the !resizable case to position the snapcorner.

gmackall:
Cool, happy to test it out if you're interested in feedback. I'll try to add a sensible handling of the non resizable case here

gmackall:
I stacked it on top of the bottom right, same as top right of chat does. I could also look in to doing what I mentioned in this comment https://github.com/runelite/runelite/pull/20178#discussion_r3350951160 if you agree, or also I could close if you feel you wouldn't want to accept this now that [your commit](https://github.com/runelite/runelite/commit/79b64b1c5fac502621f67084669f1aaba1fdb9e0) has landed (though I think this is still a natural default snapcorner in the resizable modes). Thanks so much for landing https://github.com/runelite/runelite/commit/79b64b1c5fac502621f67084669f1aaba1fdb9e0 by the way, very much looking forward to using it.

PR #20188, opened 2026-06-05
Add 'Wrap Direction' flipping to Infoboxes

kevetone:
(Closed and reopened pr after moving changes to its own branch)

gif1

---

gif2

---

When an Infobox Overlay has too many infoboxes to be able to fit its bounds, it traditionally wraps either by making a new column to the right, or new row below. Determined by `ComponentOrientation`, which was controlled by a `Flip` menu entry. 

Added 2 new sub menu entries for `Flip`. 
  - `Orientation` : Same behavior as before. The default `onClick` for `Flip` menu click. 
  - `Wrap Direction` : Sets new `ComponentWrapDirection` value, between `NORMAL` and `REVERSED`. 
  
  ---

`PanelComponent.java`, that was responsible for the drawing children renderables (infoboxes in this case), wrapping them, and updating bounds, now uses `Wrap Direction` when determining how much to slide over the children.

Previously, wrapping could just work by adjusting the size of the panel. However, rendering engine has point `0,0` as the top left of the drawn region. This is a problem when we want to grow leftwards, or upwards. 

With children now being drawn 'out of bounds' of the original bounds, the `anchor point` / `preferredLocation` for the panel needs to be adjusted as well. To do so, `PanelComponent` just tracks the min X/Y of what it drew, and subtracts the original X / Y to form an "offset". The minimum works in this case as anything that went upward or leftward is lower than 0,0. 

 ---

There are some other fixes that need to be accompanied when applying an offset from inside `PanelComponent.render()`. Such as the panel teleporting to wrong location on Alt Drag, or Yellow Border bounds / rendering not updating, and Shift right click hitbox being misaligned. 

My fix for all of these are not the greatest, and I am open to suggestions. This is the cleanest and quickest, I could come up with, without a larger refactor. 

The core logic is, `PanelComponent` bounds act as an `offset`. That offset is now propagated up to where it needs to go in order to fix above mentioned calculations. 

- `Overlay.class` has a new `getBoundsOffset()` that just returns an offset of `0,0` as the default. This lets most things remain unaffected.

- `OverlayPanel.render()` already hands off most of the rendering to `PanelComponent`. After rendering, `OverlayPanel` now just also uses `PanelComponent`'s bounds, to update its own bounds. 

- `OverlayPanel` also overrides `getBoundsOffset()`, and uses the bounds of its `PanelComponent` as its offset. 

These bound offsets are propagated to where they need to go, like to prevent teleporting to the wrong position on Alt Drag.


ZohiDev:
bumping this would love this feature

G-lint:
Would LOVE this!!

PR #20192, opened 2026-06-07
AttackStyles: Add option to hide NPC attack options for warned attack styles.

PureIllenium:
Adding in option under the Attack Styles plugin to hide NPC attack options when a warned style is used.


PureIllenium:
Video of how this works:
https://github.com/user-attachments/assets/afe3de1b-d5dc-4e72-acc5-5f2d952fb438

PureIllenium:
Added in external plugin to handle this in the meantime - https://github.com/runelite/plugin-hub/pull/13511

PR #20220, opened 2026-06-18
itemmapping: add demonic pacts ornament kits

marknewan:
adds the variations of

* trident of the seas / swamp
* soulreaper axe
* demonic quill / slayer helm


Nightfirecat:
These changes look good, but you missed the iban's staff kit--I've added that and merged this in commit 50590a9ff83312c460f9aa7d1ff60e0456486e5f

PR #20257, opened 2026-07-02
loot tracker: tarnished items and maggot king eggs

samszotkowski:
Empty drops are observed and recorded but don't affect the side panel or config


samszotkowski:
@Blonks-afk tested this stuff. Tarnished items work (pretty standard fare), and the maggot king messages did what we expect: messages for take-eggs supplies, take-eggs nothing, and egg pops are all recorded, but do nothing to the side panel or to website dataset.

PR #20267, opened 2026-07-05
feat(plugin,bank): add hotkey to toggle withdrawing as notes

mkatychev:
Add a keybind to allow toggling 'Enable Notes' in the banking UI:
image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+root = true
+
+[*]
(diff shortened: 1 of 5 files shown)


.editorconfig
mkatychev, line 1: I can add this [`.editorconfig`](https://editorconfig.org/) to a separate PR if need be

PR #20270, opened 2026-07-06
bank tags: Make tags work in GIM storage

Robbejj:
This change will allow users to edit and search for tags in the group ironman shared storage.

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
@@ -238,6 +238,22 @@ public class BankTagsPlugin extends Plugin implements BankTagsService
 		}
 	}
 
+	private void layoutSharedBank()
+	{
+		Widget w = client.getWidget(InterfaceID.SharedBank.ITEMS);
+		if (w != null)
+		{
+			Object[] args = w.getOnInvTransmitListener();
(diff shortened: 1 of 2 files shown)


dillydill123:
@Nightfirecat or @Adam- how likely is a feature like this to be added to Bank Tags? I'm aware enhancements/features in core are usually not accepted. The original PR was part of Inventory Setups, but I hesitate to deviate from underlying bank tags. Curious what you think.

runelite-client/src/main/java/net/runelite/client/plugins/banktags/BankTagsPlugin.java
```diff
@@ -474,6 +492,8 @@ private void editTags(MenuEntry entry)
 
 					// If a tab if active, rebuild the bank to apply the changes
 					tabInterface.reloadActiveTab();
+					// If the shared bank is open, rebuild it to reapply an active tag: search
+					layoutSharedBank();
```
Adam-: I'm a little confused what this is for / how to test it, since editing the tags on something would necessarily close the search first which means you can't edit the tags on something when a search is being performed?

Robbejj: I took another look at the code and I think you're right to be confused ':D
I did some cleanup and another test session, still feels fine after removing `layoutSharedBank()`

PR #20286, opened 2026-07-08
Boosts Information - new Str,Range,Mage option + use correct base Level inside LMS

Matsyir:
New option, similar to combat, but hide Attack+Defence, since those are rarely relevant information due to the combined popularity of super combat potions & saradomin brews.

image

Particularly annoying for me, and I'm sure many others, is that when using this plugin in LMS, if you have a maxed account, you are going to constantly see a negative boost on defence & attack. Even in PvM though, I can't really think of a situation where your attack & defence boosts would be relevant to you on a max main using super combat pots and/or brews.

Here are some example screenshots

1def LMS match, with new config vs. with default Combat config (this occurs even on max/med due to the build being 85 defence)
Screenshot 2026-07-08 133436
Screenshot 2026-07-08 133440

Random bank screenshots outside of LMS
Screenshot 2026-07-08 133348
Screenshot 2026-07-08 133340

_Note: after taking those 4 screenshots, I overrode DisplayBoosts.toString() so that it shows `Str, Range, Mage` rather than just `Str Range Mage`, as seen in the inital screenshot above_

I haven't done a PR to core repo in over 5 years atp, let me know if there's any steps I forgot to take for this PR :)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostIndicator.java
@@ -60,7 +60,7 @@ class BoostIndicator extends InfoBox
 			return String.valueOf(client.getBoostedSkillLevel(skill));
 		}
 
-		int boost = client.getBoostedSkillLevel(skill) - client.getRealSkillLevel(skill);
+		int boost = client.getBoostedSkillLevel(skill) - CombatLevelsProvider.getRealSkillLevel(client, skill);
 		String text = String.valueOf(boost);
 		if (boost > 0)
 		{
@@ -74,7 +74,7 @@ class BoostIndicator extends InfoBox
 	public Color getTextColor()
 	{
 		int boosted = client.getBoostedSkillLevel(skill),
-			base = client.getRealSkillLevel(skill);
+			base = CombatLevelsProvider.getRealSkillLevel(client, skill);
 
 		if (boosted < base)
 		{
--- a/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/boosts/BoostsConfig.java
@@ -24,6 +24,12 @@
  */
 package net.runelite.client.plugins.boosts;
 
+import com.google.common.collect.ImmutableSet;
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import net.runelite.api.Skill;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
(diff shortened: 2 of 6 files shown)


Nightfirecat:
Separately from this option (which I'd personally be fine with adding), I wouldn't mind adding the ability for the plugin to "understand" base combat levels changing within LMS so long as a maintainable solution is possible--do you know offhand whether certain vars are set for the various bracket types?

Matsyir:
Edit: In hindsight, my answer is a bit of a hacky way to detect things, there probably is some straight up varbit or whatever to tell us we're in LMS and what build it is, but I don't know it no.

Original Answer:

Simply checking `client.getBoostedSkillLevel(Skill.DEFENCE);`. I think _maybe_ even getRealSkillLevel would work too. As far as I understand/from what logical solutions explained to me, you get "placed into another player" in LMS.

In regards to detecting when an LMS match starts to run this check, I'm not too sure. Probably onRegionChanged would be the best if there's nothing more direct than that.

Here's some snippets of how we check this in pvp tracker. The FightPerformance constructor gets called when you attack another player or get attacked by another player (mix of onAnimationChanged & onInteractingChanged)

FightPerformance.java:
```Java
// constructor which initializes a fight from the 2 Players, starting stats at 0. Regular use constructor.
public FightPerformance(Player competitor, Player opponent)
{
  int defLvl = PLUGIN.getClient().getBoostedSkillLevel(Skill.DEFENCE);
  
  // determine fight type based on being at LMS areas & use def level to check for LMS builds.
  this.fightType = !PLUGIN.isAtLMS() ? FightType.NORMAL :
	  defLvl <= FightType.LMS_1DEF.getCombatLevelsForType().def ? FightType.LMS_1DEF :
	  defLvl <= FightType.LMS_ZERK.getCombatLevelsForType().def ? FightType.LMS_ZERK :
	  FightType.LMS_MAXMED;
```

FightType.java:
```Java
public enum FightType
{
	LMS_MAXMED(new CombatLevels(118, 118, 85, 112, 99, 99)),
	LMS_ZERK(new CombatLevels(97, 118, 50, 112, 99, 99)),
	LMS_1DEF(new CombatLevels(91, 118, 1, 112, 99, 99)),
	NORMAL(CombatLevels.getConfigLevels());

	private static final FightType[] LMS_TYPES = {LMS_MAXMED, LMS_ZERK, LMS_1DEF};

	private final CombatLevels combatLevelsForType;
	FightType(CombatLevels combatLevelsForType)
	{
		this.combatLevelsForType = combatLevelsForType;
	}

	public CombatLevels getCombatLevelsForType()
	{
		if (this == NORMAL)
		{
			return CombatLevels.getConfigLevels();
		}

		return combatLevelsForType;
	}

	public boolean isLmsFight()
	{
		return ArrayUtils.contains(LMS_TYPES, this);
	}
}
```

Matsyir:
So basically,
on max/med: BR_ACTIVE_BUILD_PLAYER = 0
on zerk: BR_ACTIVE_BUILD_PLAYER = 1
on pure: BR_ACTIVE_BUILD_PLAYER = 2



I just noticed in the core Prayer plugin, there's this varbit which may be helpful, though it doesn't include the build itself:

`boolean inLms = client.getVarbitValue(VarbitID.BR_INGAME) != 0;`

Going to the implementation of VarbitID.BR_INGAME, we can find a few related ones - BR_MODE_SELECTED may be relevant, but that sounds like it might just be your favorite/preferred mode selection rather than the ongoing one.

EDIT: did some testing, looks like we **can** see currently used build. If you saw my initial edit, I misread the results initially
Screenshot 2026-07-14 063316
Screenshot 2026-07-14 063636
image

Matsyir:
Added new commit to make use of varbits BR_ACTIVE_BUILD_PLAYER & BR_INGAME in order to detect LMS matches/builds, and use those stats instead of normal base stats when applicable

my initial implementation caused the plugin to crash, added an invokeLater to fix (BoostsPlugin - line 314). Other calls to CombatLevelsProvider.getRealSkillLevel may also require invokeLaters, but this seems fine from my testing so far

screenshot in ferox with "Combat" option, to show that its still working as normal, vs. in LMS on pure to show that it's no longer displaying that we are debuffed on def & attack 👍
Screenshot 2026-07-14 084307
Screenshot 2026-07-14 084349

I have tested all 3 LMS builds and it seems to be working fine.

Nightfirecat:
For the new option specifically (I'm reviewing the new config option and the LMS level display options separately), I applied some cleanup changes on top of your commit, which you can see here: <https://github.com/Nightfirecat/runelite/commit/ab7906248d50c418ce7d59aa91a0b4687d2373fe> Does that look fine to you?

Matsyir:
> For the new option specifically (I'm reviewing the new config option and the LMS level display options separately), I applied some cleanup changes on top of your commit, which you can see here: [Nightfirecat@ab79062](https://github.com/Nightfirecat/runelite/commit/ab7906248d50c418ce7d59aa91a0b4687d2373fe) Does that look fine to you?

Looks all good to me overall! Good catch on retaining the original enum name for "Both".

The one thing that maybe seems unnecessary to me, although it's a hell of a micro-optimization & nitpick, but since you asked: [BoostsPlugin.java: L210](https://github.com/Nightfirecat/runelite/commit/ab7906248d50c418ce7d59aa91a0b4687d2373fe#diff-374e4236085dd69df5e550f692cc3a221f75f6ba56aa2e94161dd26269878570L210), chooses to check whether "any skill" (from the "Both" enum) is contained in the Stat changes (onStatChanged), which technically causes a bigger loop than actually needed for on the existing config/requested display.

I imagine there was some reason to make the change though, maybe to allow the plugin to work right away if you switch back to "All" config. Or maybe there's a chance of NPE with the flow of the plugin if we don't constantly send all skills here, but if that's the case, maybe it would be better to just add an `if (skill == null)` along the way before that happens.

Either way though, hell of a micro-optimization like I said, plus it's in event context not even render() or anything, very negligible detail, looks fine overall.

Nightfirecat:
That change was intentional, rather it was done to preserve the behavior that exists in the release client. In truth, who knows if players would care if skills not shown in the overlay debuffing wouldn't fire notifications anymore, but it ought to be considered separately.

PR #20292, opened 2026-07-09
motherlode: fix broken struts overlay early return

### LINKED ISSUE #20172: Motherlode Mine - Broken struts are wrong
### Type

Incorrect behavior

### Operating System

None

### OS Version

_No response_

### Bug description

**Show broken struts** does nothing. 
**Show rocks obstacles** displays both obstacles, but also broken struts.

How to fix:
Make it so **Show broken struts** actually shows broken struts, and that **Show rocks obstacles** only shows rock obstacles

### Screenshots or videos

_No response_

### RuneLite version

2.7.5

### Logs

_No response_

connorhet:
The Motherlode overlay early return only checked veins + rocks, so with both off the whole overlay bailed even when broken struts were enabled.

Includes `showBrokenStruts` in that check. Object tracking for rocks vs struts was already correct.

Fixes #20172

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSceneOverlay.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSceneOverlay.java
@@ -81,11 +81,6 @@ class MotherlodeSceneOverlay extends Overlay
 			return null;
 		}
 
-		if (!config.showVeins() && !config.showRockFalls() && !config.showBrokenStruts())
-		{
-			return null;
-		}
-
 		Player local = client.getLocalPlayer();
 
 		renderTiles(graphics, local);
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/motherlode/MotherlodeSceneOverlay.java
```diff
@@ -76,7 +76,12 @@ class MotherlodeSceneOverlay extends Overlay
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
-		if ((!config.showVeins() && !config.showRockFalls()) || !plugin.isInMlm())
+		if (!plugin.isInMlm())
+		{
+			return null;
+		}
+
+		if (!config.showVeins() && !config.showRockFalls() && !config.showBrokenStruts())
```
Adam-: looks like you could just remove this since renderTiles checks it too

connorhet: true, thanks

PR #20306, opened 2026-07-15
item mapping: add elder venator and araxyte fang

ldavid432:


ldavid432:
Removed the venator fangs and teeth as those have just been made tradeable

PR #20309, opened 2026-07-15
idlenotifier: Added collecting bloodwood sap to the animation

Cryslacks:
Update includes the animation for when collecting bloodwood sap from the new more afk bloodwood tree.

Image of animation id when collecting the sap.

image

Tested and works locally


ldavid432:
This is a duplicate of #20308

PR #20342, opened 2026-07-28
Add hallowed flail as a valid item for the falo the bard step

Bjdufre1:
Adds hallowed flail as a valid item for the falo step.


Nightfirecat:
Thank you!

Issue #19804, opened 2026-01-17
GPU plugin memory leak issue in newest RuneLite version


tormodAase:
### Type

GPU

### Operating System

Windows

### OS Version

Windows 11

### Bug description

Hey, I'm submitting this report for a friend.

For the last 2-4 days, he has been having massive issues with RuneLite. After troubleshooting for a few days, even factory resetting his computer without any luck, we found out it is a memory leak issue. EventViewer told us a few times his RuneLite application was using 19GB of memory when his computer started to act weird. RuneLite crash logs also tells us about memory issues.

After monitoring task manager while he was doing gauntlet, we could see the memory usage starting at around 900MB and gradually going up. After about 3 minutes, the memory usage was at 9000MB.

To solve this, we replaced the GPU plugin with GPU (Legacy). That seems to have been a temporary workaround that works well.

We believe it is related to the newest RuneLite version 1.12.12, since that was released just a few days ago, when he started having issues. I'm currently on 1.12.11, so I can't verify if this is a common issue or not.

### RuneLite version

RuneLite version: 1.12.12.1

### Logs

[jvm_crash_pid_14424_r.log](https://github.com/user-attachments/files/24691208/jvm_crash_pid_14424_r.log)

Adam-:
If you could try bisecting that would be helpful in narrowing this down, if it is a real issue. That said I don't see any correlation between 1.12.12 and OOMS; though if it is limited to only eg Windows+ARM it probably wouldn't stand out enough in my data to easily tell.

tormodAase:
Hey, thanks for your response!

It could be Windows-ARM related. He is using Samsung Galaxy Book 4 Edge, which uses an ARM-based CPU.

It's hard to analyze further, since the logs doesn't give us much information prior to the out of memory exceptions. In his `client.log` file, I found these two exceptions at similar timestamps to when he notices issues:

```log
2026-01-17 15:46:09 GMT [Client] ERROR n.r.client.plugins.gpu.GpuPlugin - error swapping buffers
java.lang.RuntimeException: unable to SwapBuffers
	at net.runelite.rlawt.AWTContext.swapBuffers(Native Method)
	at net.runelite.client.plugins.gpu.GpuPlugin.draw(GpuPlugin.java:1439)
	at bm.au(bm.java:15439)
	at bm.tw(bm.java)
	at bm.run(bm.java:61256)
	at java.base/java.lang.Thread.run(Unknown Source)
```

```log
[...]

PR #20348, opened 2026-07-30
gpu: describe static xyz as a four-component short attribute

jesherman:
Static zone geometry declares its xyz position as a three-component `GL_SHORT` attribute:

```java
glVertexAttribPointer(0, 3, GL_SHORT, false, VERT_SIZE, 0);
```

D3D12/DXGI has no native three-component 16-bit vertex format, so translation layers such as OpenGLOn12 have to emulate it. On a Windows machine running through OpenGLOn12, static opaque draws dominated frame time — suppressing them alone raised the frame rate to ~100 FPS. Buffer mapping instead of `glBufferSubData`, Mesa threaded dispatch, replacing `glMultiDrawArrays`, removing per-zone uniform changes, sharing VAOs, and drastically simplifying both the vertex and fragment shaders all made no material difference.

The vertex record is already 20 bytes with a padding short after xyz (attribute 1 starts at byte offset 8), so exposing four components costs nothing in memory or bandwidth and lets the translation layer select a native 4x16 format. `vert.glsl` declares `layout(location = 0) in vec3 vertf`, so the fourth component is simply dropped by the shader and rendering is unchanged.

With this one-line change the full scene rendered at ~160 FPS even at maximum draw distance.

Verified on native OpenGL as well (NVIDIA, Windows 11 arm64 and win64): the scene renders identically with no regression, so the change is a no-op for drivers that already support 3x16 natively.

Context and the full investigation: https://github.com/runelite/runelite/issues/19804#issuecomment-5125937354

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/Zone.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/Zone.java
@@ -156,7 +156,7 @@ class Zone
 		glBindBuffer(GL_ARRAY_BUFFER, buffer);
 
 		glEnableVertexAttribArray(0);
-		glVertexAttribPointer(0, 3, GL_SHORT, false, VERT_SIZE, 0);
+		glVertexAttribPointer(0, 4, GL_SHORT, false, VERT_SIZE, 0);
 
 		glEnableVertexAttribArray(1);
 		glVertexAttribIPointer(1, 1, GL_INT, VERT_SIZE, 8);


jesherman:
removed the comments from the agent which are overbearing.

Adam-:
The context from the comment would be a good thing to place in a commit message

jesherman:
> The context from the comment would be a good thing to place in a commit message

fixed, sorry about that.

Adam-:
> > The context from the comment would be a good thing to place in a commit message
> 
> fixed, sorry about that.

Your commit message isn't very good because it's listing a bunch of irrelevant things your agent tried that don't work.

Something like this is better:

```
gpu: describe static xyz as a four-component short attribute

Static zone geometry declares its xyz position as a three-component
GL_SHORT attribute, but D3D12/DXGI has no native three-component 16-bit
vertex format, so translation layers such as OpenGLOn12 have to emulate
it.

The vertex record already has a padding short after xyz, so exposing
four components lets the translation layer pick a native 4x16 format.
```

PR #20350, opened 2026-07-30
loottracker: add golem crafting

coopermor:


runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
```diff
@@ -170,6 +170,20 @@ public class LootTrackerPlugin extends Plugin
 	// Shipwreck salvaging
 	private static final Pattern SALVAGE_PATTERN = Pattern.compile("You sort through the\\s+(?<tier>\\S+)\\s+salvage.*");
 
+	// Wyrmscraig golem crafting
+	private static final Pattern GOLEM_CRAFTING_PATTERN = Pattern.compile(
+		"^As you complete the golem it leaves a gift on the ground for you: 1 x " +
```
Nightfirecat: This needs to be updated to support gems being placed directly in the gem bag after today's update. Looks like that text is in the following format:

```
As you complete the golem it leaves a gift in your gem bag for you: 1 x <item>.
```

I only tested this with a normal gem bag, this should be tested with the gem sack as well.

Nightfirecat:
Also adding a couple test cases for this (using chat messages from your log) would be ideal

runelite-client/src/main/java/net/runelite/client/plugins/loottracker/LootTrackerPlugin.java
samszotkowski: - `<col=06600c>As you complete the golem it leaves a gift in your gem bag for you: 1 x Uncut diamond</col>.`
- `<col=06600c>As you complete the golem it leaves a gift in your gem sack for you: 1 x Uncut diamond</col>.`
- `<col=06600c>As you complete the golem it leaves a gift on the ground for you: 1 x Uncut diamond</col>.`

Messages from crowdsourcing. Don't see any "in your gem tote/pouch/satchel" so idk if those exist

Psychemaster: Golems don't award opal, jade or red topaz so the rewards will never go into the bags for those gems specifically.

PR #20352, opened 2026-07-31
boss/boss timer: add Mad Angel

fontivan:
Contains two commits:

1. boss: add Mad Angel
    - add the Mad Angel to the Boss list
    - relevant post-quest NPC IDs here are:
        - 16314: idle in world
        - 16306: idle in instance
        - 16307: waking up
        - 16305: active
        - 16308: dead

2. boss timers: update onNpcChanged to account for Mad Angel
    - the Mad Angel does not despawn but instead changes from one npc to another
    - 16306 (idle) --> 16307 (waking up) --> 16305 (active) --> 16308 (dead) --> repeat
    - as a result, it does not trigger the onNpcDespawned event and instead triggers onNpcChanged
    - update onNpcChanged to account for Mad Angel changes similar to how Hueycoatl is already done

**NPC IDs**

1. Idle in World 16314
01-mad-angel-idle-world-16314

2. Idle in Instance 16306
02-mad-angel-idle-instance-16306

3. Waking Up 16307
03-mad-angel-waking-up-instance-16307

4. Active 16305
04-mad-angel-active-instance-16305

5. Dead 16308
05-mad-angel-dead-16308


runelite-client/src/main/java/net/runelite/client/game/NpcUtil.java
```diff
@@ -170,6 +170,7 @@ public boolean isDying(final NPC npc)
 			case NpcID.XARPUS_DEATH_HARD:
 			case NpcID.NIGHTMARE_DYING:
 			case NpcID.NIGHTMARE_CHALLENGE_DYING:
+			case NpcID.MAD_ANGEL_DEAD:
```
Nightfirecat: Why is this change needed? This seems unrelated to the boss timer.

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersPlugin.java
```diff
@@ -83,9 +83,19 @@ public void onNpcChanged(NpcChanged npcChanged)
 	{
 		NPC npc = npcChanged.getNpc();
 		Boss boss = Boss.find(npc.getId());
-		if (boss == Boss.HUEYCOATL)
+		if (boss == null)
```
Nightfirecat: This null check is unnecessary; `switch (boss)` handles the `boss == null` case fine.

runelite-client/src/main/java/net/runelite/client/game/NpcUtil.java
fontivan: I added this before I realized it doesn't actually trigger onNpcDespawned, so I think you are right and this is not necessary

runelite-client/src/main/java/net/runelite/client/plugins/bosstimer/BossTimersPlugin.java
fontivan: I was getting NPE's in the log without it, and my Google-fu says this requires at least Java 18 to not cause a NPE (https://openjdk.org/jeps/420). The Intellij instructions (https://github.com/runelite/runelite/wiki/Building-with-IntelliJ-IDEA) also still say JDK 11, so while its not required I think its a good idea to avoid the NPE in the logs

fontivan: I could also just leave it as an if statement since arguably the case is a bit heavy handed for 2 statements but I liked the readability of the case better here and it leaves room for any future bosses that also require this
```
if (boss == Boss.HUEYCOATL || boss == Boss.MAD_ANGEL) {
    createTimer(npc, boss);
}
```

Adam-: `if (boss == Boss.HUEYCOATL || boss == Boss.MAD_ANGEL)` makes the most sense

fontivan:
Have addressed all the feedback, rebased, and retested :) ready to go!

PR #20356, opened 2026-08-02
Fix title bar dragging with custom chrome on Linux

rsfost:
With custom window chrome enabled on Linux, the window can only be dragged from the top-left corner where the RuneLite icon is. Forwarding the events to `FlatTitlePane` seems to fix this issue when tested on Fedora 44 / KDE Plasma.


Adam-:
I can't reproduce this on kubuntu

https://github.com/user-attachments/assets/af5b39bb-7e29-43f1-918d-08db1a44813c

rsfost:
Must be something weird in my environment then, I'll do some more testing

rsfost:
Turns out this issue is caused by the Client Resizer hub plugin

PR #20360, opened 2026-08-03
worldmap: add slepe and castle drakan teleports

ldavid432:
Screenshot 2026-08-02 215534

image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/TeleportLocationData.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/worldmap/TeleportLocationData.java
@@ -165,6 +165,8 @@ enum TeleportLocationData
 	TELEPORT_CRYSTAL_PRIFDDINAS(TeleportType.OTHER, "Teleport crystal", "Prifddinas", new WorldPoint(3264, 6065, 0), "teleport_crystal_icon.png"),
 	DRAKANS_MEDALLION_VER_SINHAZA(TeleportType.OTHER, "Drakan's medallion", "Ver Sinhaza", new WorldPoint(3649, 3230, 0), "drakans_medallion_teleport_icon.png"),
 	DRAKANS_MEDALLION_DARKMEYER(TeleportType.OTHER, "Drakan's medallion", "Darkmeyer", new WorldPoint(3592, 3337, 0), "drakans_medallion_teleport_icon.png"),
+	DRAKANS_MEDALLION_SLEPE(TeleportType.OTHER, "Drakan's medallion", "Slepe", new WorldPoint(3808, 9700, 0), "drakans_medallion_teleport_icon.png"),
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/worldmap/TeleportLocationData.java
```diff
@@ -165,6 +165,8 @@ enum TeleportLocationData
 	TELEPORT_CRYSTAL_PRIFDDINAS(TeleportType.OTHER, "Teleport crystal", "Prifddinas", new WorldPoint(3264, 6065, 0), "teleport_crystal_icon.png"),
 	DRAKANS_MEDALLION_VER_SINHAZA(TeleportType.OTHER, "Drakan's medallion", "Ver Sinhaza", new WorldPoint(3649, 3230, 0), "drakans_medallion_teleport_icon.png"),
 	DRAKANS_MEDALLION_DARKMEYER(TeleportType.OTHER, "Drakan's medallion", "Darkmeyer", new WorldPoint(3592, 3337, 0), "drakans_medallion_teleport_icon.png"),
+	DRAKANS_MEDALLION_SLEPE(TeleportType.OTHER, "Drakan's medallion", "Slepe", new WorldPoint(3808, 9700, 0), "drakans_medallion_teleport_icon.png"),
+	DRAKANS_MEDALLION_CASTLE_DRAKAN(TeleportType.OTHER, "Drakan's medallion", "Castle Drakan", new WorldPoint(3560, 3360, 0), "drakans_medallion_teleport_icon.png"),
```
Nightfirecat: This teleports you inside the castle doesn't it? Let's move the marker so it's appropriately showing up inside.

ldavid432: the inside of the castle isn't in the main world map, is it fine if its only on the minimap?

Nightfirecat: Yeah

ldavid432: Actually, unless I am missing something I don't think these even display on the minimap? So I think the option would either be on the outside of the castle or just not at all?

ldavid432: Updated to this location:

image

Confirmed coordinates in-game and on mejrs map. Not technically visible in-game though

https://mejrs.github.io/osrs?m=-1&z=5&p=0&x=3166&y=7711

Nightfirecat: Sorry if I wasn't clear; given that this instance of the castle is not visible on any world map, I meant that the marker should appear on the overworld, in the spot within the castle where you would end up teleporting, even though I am aware that you don't technically go into the "overworld" drakan's castle. Simply, I would expect the marker to be ~10 or so tiles west of where you originally had it.

ldavid432: Ah my bad, I totally misunderstood what you said. Updated the location to be in the overworld, shifted over to approximately where it would be inside the castle

image

PR #20370, opened 2026-08-05
Draw ABOVE_SCENE overlays at native resolution under stretched mode

### LINKED ISSUE #7155: Don't scale RL overlays in scaled mode
I like the scaled UIs of scaled mode but miss the sizes of the non-scaled RL overlays.

I would like there to be an option to not scale the overlays this might not be easily possible and doesn't really matter that much. I just like the non-scaled size better.

**Additional context**
This really isn't a big deal and scaled mode is really nice how it is. This is just a small detail that I think would make it better.

I am talking about stretched resizeable with gpu plugin enabled if it matters.

StaySleeping:
Draft to fix #7155


StaySleeping:
Closing while the stacked rewrite is reshaped to a more minimal single-buffer PR1. Will reopen from stack branches when ready.

PR #20372, opened 2026-08-05
chatcommands: read the adventure log owner from either menu interface

420kc:
The POH adventure log opens in one of two menu interfaces depending on the player's interface style setting: `MENU` (187) or `MENU_NEW` (947). The plugin only watches `MENU`, so for players on the newer style `pohOwner` is never captured. The boss kill log and pets paths tolerate a null owner and keep working, but the Counters personal-best parse requires an owner match, so it is silently skipped and those players record no adventure-log personal bests at all.

This watches both groups and reads the owner from whichever menu is present. The title sits at a different child index in each and is surrounded by null and non-text children, so each container is scanned for the first line matching the existing title pattern rather than read from a fixed index; `ADV_LOG_EXPLOITS_TEXT_INDEX` is no longer needed. Tags are stripped before matching, which is a no-op on the old interface.

Scope is chatcommands only. The deprecated `WidgetInfo.ADVENTURE_LOG` / `WidgetID.ADVENTURE_LOG_ID` mappings still reference the old interface and are left untouched as frozen legacy API.

Tested in-game on both settings: on the original style the owner resolves through `Menu.LJ_LAYER2`, on the newer style through `MenuNew.TITLE`, and the Counters scroll parses identically in both. The existing counters test now runs against both interfaces, with the title as a sparse dynamic child plus a decoy match after it, proving the scan stops at the first hit.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/chatcommands/ChatCommandsPlugin.java
@@ -145,8 +145,6 @@ public class ChatCommandsPlugin extends Plugin
 	private static final String CA_COMMAND = "!ca";
 	private static final String CLOG_COMMAND = "!clog";
 
-	@VisibleForTesting
-	static final int ADV_LOG_EXPLOITS_TEXT_INDEX = 1;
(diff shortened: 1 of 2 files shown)


Nightfirecat:
Thank you!

PR #20373, opened 2026-08-05
Clue Scroll: Fix coordinates for Rimmington chemist medium clue

randoxs:
The existing tile coordinate is 1 tile too far north compared to the actual center tile location.


Nightfirecat:
Does digging on the currently-marked tile advance the clue? Is it faster or easier to get to the updated location? We are only really interested in centering hot-cold clues, as the solver for them doesn't work properly without accurate coordinates.

randoxs:
Digging on the currently-marked tile does advance the clue. Though it's not faster to get to the centered tile, it's the only medium clue step to not be centered, so it is inconsistent with the rest of the steps of this tier, making it quite jarring.

Nightfirecat:
I'm quite confident we don't have all other (or even most other) medium clue steps centered...

PR #20379, opened 2026-08-06
Wintertodt: Add custom notification support

### LINKED ISSUE #19863: Wintertodt plugin notifications can't be configured
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10

### Bug description

Image

### Screenshots or videos

_No response_

### RuneLite version

RuneLite version: 1.12.16
Launcher version: 2.6.7

### Logs

_No response_

justonhart:
Close #19863

Add custom notification configuration support for each notification type, while maintaining backwards compatibility with existing configs.


Adam-:
~~You shouldn't be testing `isEnabled()` all over the place like this, look at how other plugins notifications work.~~
Oh, this is really weird since you are trying to send notifications even when the notifications are not enabled due to the way the plugin config works.

justonhart:
> ~You shouldn't be testing `isEnabled()` all over the place like this, look at how other plugins notifications work.~ Oh, this is really weird since you are trying to send notifications even when the notifications are not enabled due to the way the plugin config works.

Yeah, the original plugin didn't use Notification for custom configurations. I added the new config option to allow customization of the "damage" notifications without breaking existing user configs. In the case that the new option isn't enabled, the damage notifications are sent as they were originally.

I am new to the project, but I figure this was probably a better way to do things from ux perspective. It might be cleaner to refactor how the original notifications are sent, but I assume that could break pre-existing configs.

Adam-:
When I made the notification system I remember being confused by this plugin so I just left it alone. I think what you have here is super confusing though since you will end up sending notifications even when the notification is off. Maybe a boolean "only notify when interrupted" would be better.

justonhart:
It'd certainly be easier to follow if we replace the legacy damage notification configs with notification configs alongside your proposed boolean config. Presumably this would result in interrupt settings no longer being valid; is that acceptable?

TBH re-configuring the interruption notification will not be a laborious task for users with such configurations. There would be a loss of atomic configurability, however, unless we create an interrupt boolean per damage type. Seems like overkill to me, but there's probably one user out there who would disagree.

Adam-:
You can migrate the old configs to the new configs if you want. I'm not sure if its necessary

justonhart:
Refactored notification system as discussed - all notifications rely on Notification class for configuration, and atomic damage interrupt configs have been simplified to a single boolean option.

justonhart:
Is anything else needed from me to get this moved forward?

Adam-:
Seems okay now with the changes I just pushed

PR #20395, opened 2026-08-11
clues: add item requirements overlay to cryptic clues

### LINKED ISSUE #20362: Cryptic clue not checking inventory for rope
### Type

Incorrect behavior

### Operating System

Windows

### OS Version

Windows 10 IoT Enterprise LTSC

### Bug description

It checks for light source but not rope

### Screenshots or videos

Image

### RuneLite version

RuneLite version: 1.12.34.1
Launcher version: 2.7.7

### Logs

_No response_

MrJaspen:
Fixes #20362 

## Problem

Cryptic clues that require items don't surface those requirements in the overlay. The Nardah genie step is the reported case: its solution text says a light source and a rope are needed, but only "Requires Light Source!" appears. A player reading the overlay has no indication the rope is missing.

## Cause

`CrypticClue` has no mechanism for declaring item requirements. The only requirements displayed are `requiresLight` and `requiresSpade`, both booleans fromthe `ClueScroll` base class. There is no general item requirement support, so this is a missing feature rather than a bug or missing data on the specific Nardah clue step.

`EmoteClue` already solves this: it takes `ItemRequirement` values and renders fulfilment in its own `makeOverlayHint`.

## Fix

Add a `@Singular List<ItemRequirement> itemRequirements` to the `CrypticClue` builder and render unmet requirements in `CrypticClue.makeOverlayHint`, following the pattern in `EmoteClue`. Requirements are checked against inventory and equipment combined. Requirements fulfilled by interchangeable items use `any(...)` like in `EmoteClue`.

Requirements are populated for five clue steps:
- Nardah genie step (issue opened on this clue)
- Elemental Workshop (battered key required)
- Iceberg (clockwork suit required)
- Trollweiss Mountain (sled required)
- Desert mining camp (metal key required)

## Screenshots

These screenshots show the difference from the original clue overlay to the new one with the item missing from the player's inventory and the new overlay when the player has the item required.

Nardah genie step:
genieStepOldgenieRopeNonegenieRopeYes

Elemental Workshop step - no mention of battered key in old plugin:
batteredKeyOldbatteredKeyCleanbatteredKeyFulfilled

Iceberg step - no mention of clockwork suit in old plugin:
pengSuitOldpenSuitCleanpengSuitFulfilled

Trollweiss Mountain Step:
oldSledCluenewSledClueMissingnewSledClue

Desert Mining camp step:
metalKeyOldmetalKeyMissingmetalKeyFulfilled

## Testing
- [x] Requirement shows red when the item is missing, green when in inventory or equipped
- [x] The existing "Requires Light Source!" and "Requires Spade!" lines still render with no duplication
- [x] Clues without item requirements render unchanged
- [x] All five populated clue steps verified individually
- [x] Item requirements fulfilled by interchangeable items are correctly satisfied

## Scope

The only clues I added this feature for so far are these five. Certain cryptic clues have item requirements but I felt that for now they were too bulky to add to the overlay. For example, the Viyeldi caves step has multiple item requirements and they differ based on the player's agility level. The overlay would become bulky and for a step like this, the player likely reads the overlay closely. I wanted to focus on steps where they likely go straight to the map to see where they need to go and the overlay will help them remember a small item they may miss just glancing at the overlay popup. Item requirements can easily be added to any step desired by just populating the given step in `CrypticClue`.

## Known Limitation

Steps requiring a key that is stored on the key ring is currently not detected. Key ring contents don't appear in varbits, varps, or item containers, so `fulfilledBy` can't see them. A key on the ring would incorrectly show as missing. If anyone has a better understanding of how to detect keys on the keyrings, it would be a good addition to add!

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/CrypticClue.java
@@ -57,6 +57,13 @@ import net.runelite.client.ui.overlay.components.LineComponent;
 import net.runelite.client.ui.overlay.components.PanelComponent;
 import net.runelite.client.ui.overlay.components.TitleComponent;
 
+import net.runelite.api.Client;
+import net.runelite.api.Item;
+import net.runelite.client.plugins.cluescrolls.clues.item.ItemRequirement;
+import static net.runelite.client.plugins.cluescrolls.clues.item.ItemRequirements.item;
+import net.runelite.client.ui.FontManager;
+import static net.runelite.client.plugins.cluescrolls.clues.item.ItemRequirements.any;
+
 @Getter
 @Slf4j
 @ToString
@@ -591,6 +598,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 			.location(new WorldPoint(3289, 3022, 0))
 			.objectId(ObjectID.TOURTRAP_QIP_CRATE_SINGLE)
 			.solution("Center of desert Mining Camp. Search the crates. Requires the metal key from Tourist Trap to enter.")
+			.itemRequirement(item(ItemID.METAL_KEY))
 			.build(),
 		CrypticClue.builder()
 			.itemId(ItemID.TRAIL_HARD_RIDDLE_EXP7)
@@ -658,6 +666,7 @@ public class CrypticClue extends ClueScroll implements NpcClueScroll, ObjectClue
 			.npcRegion(13457)
 			.solution("Enter the crack west of Nardah Rug merchant, and talk to the Genie. You'll need a light source and a rope.")
 			.requiresLight(true)
+			.itemRequirement(item(ItemID.ROPE))
 			.build(),
 		CrypticClue.builder()
 			.itemId(ItemID.TRAIL_CLUE_HARD_RIDDLE010)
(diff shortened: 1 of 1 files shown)


StaySleeping:
I also think this needs to be done, but you've only covered 4 clue steps. The real work is doing it for all of them that require spades, ropes, sleds, etc. also can this not be done with the current item reqs system? or is that conflicting with the emote stash unit stuff?

MrJaspen:
> I also think this needs to be done, but you've only covered 4 clue steps. The real work is doing it for all of them that require spades, ropes, sleds, etc. also can this not be done with the current item reqs system? or is that conflicting with the emote stash unit stuff?

So this uses almost the exact same code from the emoteclue overlayhint method. I also only did 5 clue steps because i thought these were the cleanest. An example of one i consciously skipped is the vigorra step that requires ring of shadows and ghost speak amulet. It made the overlay large fast because of how much text is required. A similar issue happened for the step that requires you to wear white knight top and bottom. If the desire is to have it for every clue step, I could easily add it. Most cryptic clues don't require anything so there aren't that many steps to add stuff for. 

ringShadowsOldghostSpeakStep

StaySleeping:
Yeah, just a thought. Kinda no reason to do this unless it fixes all of them. I'm not the judge of that though so you'd have to wait for a maintainer's thoughts.

Also another thought, it should come with removing the text that says "while wearing a ring of visibility..." (or similar from other clue reqs) from the "Solution:" section.

PR #20430, opened 2026-08-23
timetracking: replace farming contract dialogue tracking with varbits

coopermor:


runelite-client/src/main/java/net/runelite/client/plugins/timetracking/farming/FarmingContractManager.java
```diff
@@ -181,41 +173,28 @@ private void handleInfoBox()
...
-
-		if (dialogText.equals(CONTRACT_REWARDED))
-		{
-			setContract(null);
-		}
-
-		Matcher matcher = CONTRACT_ASSIGN_PATTERN.matcher(dialogText);
-
-		if (!matcher.find())
+		if (contractComplete == 2)
```
Nightfirecat: Let's make this handle `!= 0` instead, so we can also remove the `onChatMessage` handler in the Plugin file. (that corresponds to the `contractComplete == 1` case)

Nightfirecat (approved):

PR #20439, opened 2026-08-24
BankTags LayoutManager: add quiver's ammo to layout & ammo plugin update

JZomDev:
If quiver is in worn or in inventory place the ammo found within the quiver into spot 2 within the Default layouting. 

image


Adam-:
The client uses a param for this:
```
// 5028
[proc,dizanas_quiver_in_inventory](inv $inv0)(int)
def_int $int1 = 0;
def_int $int2 = inv_size($inv0);
def_obj $obj3 = null;
while ($int1 < $int2) {
        $obj3 = inv_getobj($inv0, $int1);
        if ($obj3 ! null & oc_param($obj3, param_1910) = 1) {
                return(1);
        }
        $int1 = calc($int1 + 1);
}
return(0);
```

PR #20456, opened 2026-08-28
item-stats: Add missing Castlewars Brew stat changes

Bonfire:
While playing Castle Wars I found that some item stat changes weren't reflected when hovering over the [Castlewars Brew](https://oldschool.runescape.wiki/w/Castlewars_brew). Namely, the Super Restore and Stamina Potion effects that were added in [April of 2025](https://oldschool.runescape.wiki/w/Castlewars_brew#Changes). This PR aims to include these missing effects in the item stats plugin for the Castlewars Brew item.

I'm not too familiar with how the item stats plugin does its `ItemChanges`, so please feel free to suggest changes. I will note that the Super Restore prayer restoration effect didn't seem to apply correctly the first time, hence why I manually apply it [here](https://github.com/runelite/runelite/pull/20456/changes#diff-8c09b1935c2fed11e28524a2f4bc37d712c9e86e81db040aeb74d5dcaed51fc4R47). 

Thank you.


runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/CastlewarsBrew.java
```diff
@@ -0,0 +1,69 @@
...
+public class CastlewarsBrew implements Effect
+{
+	@Override
+	public StatsChanges calculate(Client client)
+	{
+		// Super Restore
+		final StatsChanges changes = new SuperRestore(.25, 8).calculate(client);
+
+		// Super Restore prayer restore (not applied above for some reason?)
+		final StatChange superRestorePrayer = new PrayerPotion(8, .25).effect(client);
```
Nightfirecat, line 47: This is happening because you're not including the super restore changes in the later call to `changes.setStatChanges()`. You should not need to calculate a separate PrayerPotion effect; instead, just use the changes you've already calculated from before instead of overwriting (thus discarding) them.

Nightfirecat: Honestly, I'd be more inclined to accept something like a new constructor to `Combo`, or a companion class to it, so that a new builder, or more straightforward class could be used, since all this class does is combine a bunch of other effects.

Nightfirecat (commented):
I cobbled together a test case for testing this PR, which you can use during development. (I might want to clean up the test case somewhat if we merge it) You'll find your PR fails it because it doesn't include the agility healing which would be expected of the super restore effect.

```java
	@Test
	public void testCastleWarsBrew()
	{
		when(client.getRealSkillLevel(any(Skill.class))).thenReturn(99);
		when(client.getBoostedSkillLevel(any(Skill.class))).then(invocation ->
		{
			final Skill skill = invocation.getArgument(0);
			switch (skill)
			{
				case ATTACK:
				case STRENGTH:
				case DEFENCE:
				case MAGIC:
				case RANGED:
					return 99;
				default:
					return 60;
			}
		});

		final StatsChanges statsChanges = new ItemStatChanges().get(ItemID._4DOSE_CASTLEWARS_SKILL_POTION).calculate(client);
		final Map<Skill, Integer> changes = Arrays.stream(statsChanges.getStatChanges())
			.filter(statChange -> statChange.getStat() instanceof SkillStat)
			.collect(Collectors.toMap(
				statChange -> Arrays.stream(Skill.values()).filter(s -> s.getName().equals(statChange.getStat().getName())).findFirst().get(),
				StatChange::getRelative));

		assertEquals(Integer.valueOf(19), changes.get(Skill.ATTACK));
		assertEquals(Integer.valueOf(19), changes.get(Skill.STRENGTH));
		assertEquals(Integer.valueOf(19), changes.get(Skill.DEFENCE));
		assertEquals(Integer.valueOf(13), changes.get(Skill.RANGED));
		assertEquals(Integer.valueOf(10), changes.get(Skill.MAGIC));
		assertEquals(Integer.valueOf(32), changes.get(Skill.PRAYER));
		assertEquals(Integer.valueOf(32), changes.get(Skill.AGILITY));
	}
```

runelite-client/src/main/java/net/runelite/client/plugins/itemstats/potions/CastlewarsBrew.java
Bonfire: Wanted to say that I plan on updating this PR with the changes you've suggested when I have time. Thank you for your review and recommendations, and the handy test case

PR #20457, opened 2026-08-28
Added the option to adjust the rendering scale of the 3D scene (see discussion #19704)

NonaNonia:
Added the option to adjust the rendering scale of the 3D scene by applying it to the dimensions of the framebuffer (Fbo) and 3D viewport (not the UI viewport) and the source dimensions during the blitting of the framebuffer. This means the 3D scene gets rendered at a scaled resolution while the output then gets stretched to the correct size.

While this could've been a separate, freely adjustable setting, I decided to add a "x0.5" and "x0.25" option to the AntiAliasingMode dropdown, below the "Disabled" to "MSAA x16" scale of options, as I think that this is the most intuitive place for it.

My motivation came from [this discussion](https://github.com/runelite/runelite/discussions/19704).

I did test it with the Stretched Mode plugin enabled and disabled and I also tested moving it from a retina to a non-retina screen.
It worked as expected. (Even preserving the bug from the release client where moving from a non-retina screen to a retina screen results in 3/4 of the screen being black.)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/gpu/GpuPlugin.java
@@ -838,8 +838,8 @@ public class GpuPlugin extends Plugin implements DrawCallbacks
 
 	@Override
 	public void preSceneDraw(Scene scene, Projection entityProjection,
-	                         float cameraX, float cameraY, float cameraZ, float cameraPitch, float cameraYaw,
-	                         int minLevel, int level, int maxLevel, Set<Integer> hideRoofIds)
+		float cameraX, float cameraY, float cameraZ, float cameraPitch, float cameraYaw,
+		int minLevel, int level, int maxLevel, Set<Integer> hideRoofIds)
 	{
 		SceneContext ctx = context(scene);
 		if (ctx == null)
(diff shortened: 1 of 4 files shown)


pirssi:
While I can't wait to see this feature be implemented (hopefully), I don't think the setting should really be placed under the 'Anti aliasing' options, since it doesn't have anything to do with that per se, and should instead be a separate option/feature. 

As well as being confusing if controlled with the same setting, adding it under the antialiasing setting would also prevent using both simultaneously, which at least for me is a option I would much prefer to have.

In addition, while maybe not as useful (at least for just replicating the look of the vanilla game with the GPU plugin disabled/unscaled on higher resolutions), I would still like the render scale to be freely adjustable instead of just two preset options for further customizability.

All in all, I think the way 117HD plugin has implemented the "game resolution" setting (along with "game scaling mode") is perfect, but I don't know how difficult implenting it like that (or this way with my requests added) would be, so I appreciate that someone has finally taken the time to add the feature in one way or another :)

NonaNonia:
Sure, I implemented this in a way where I needed to change as little code as possible to achieve my use case, but this should be easy enough as it's just adding the settings while the actual code should already supports that.

NonaNonia:
Having MSAA on in conjunction with a Game Resolution scaling != 1 was a bit more complex (see commits), and I had to do some irl stuff in between, but it works now.
Enjoy having both active at the same time:
works

pirssi:
Oh boy, just seeing that screenshot there with the game scaling working makes me very excited, as silly as it sounds! Been dreaming of this feature since I moved on from 1080p monitors years ago. And just five hours to finish it is quite quick at least in my book, so thanks!

